PR #2918, opened 2014-10-25
CONCEPTUAL: Support custom config of Highlight language support

aslakknutsen:
Map a Controller to /javascript/highlight.js so we can dynamically
return a highligh.js version based on configured language support.

Configuration of Highlight can be found under Admin > Settings > Highlight.

*DO NOT PULL UPSTREAM*

Would this approach be something to consider for custom language support in highlight configured via Discourse admin settings? With this it's up to the admin to choose specifically which of the n languages to support. 

screenshot from 2014-10-25 19 45 54

Missing currently: 
* Some sort of caching of the dynamically generated/joined highlight.js file in the controller?
* Compression of the new highlight.js parts: highlight.js + languages
   * Currently require custom source build of highlight.js


SamSaffron:
I ended up taking a slightly different approach, sorry it has taken so long to get this in, thanks for the PR 

https://github.com/discourse/discourse/commit/a82530012a0d20dfd52ab8a69afa1423f43da48c

the particular changes that If felt were important were 

- reduction of duplication using fancy intializer and amendments to site settings
- kept naming parity with what hljs has to offer
- ensured we have cdn support and caching in nginx
- made it easily updateable with a rake task 
- cleaner generation of bundle (we have a file per lang) 

I very much appreciate the effort, thanks again

aslakknutsen:
Excellent!

No worries, glad the feature is in. :)

PR #2948, opened 2014-11-06
Allow the use of DISCOURSE_REDIS_SOCKET for redis configuration

riking:
This mirrors the DISCOURSE_DB_SOCKET setting. This is intended for
multi-container setups that want to be IP-independent.

### DIFF
--- a/config/discourse_defaults.conf
+++ b/config/discourse_defaults.conf
@@ -82,6 +82,9 @@ redis_host = localhost
 # redis server port
 redis_port = 6379
 
+# redis server socket (non-empty to use)
+redis_socket =
+
 # redis database
 redis_db = 0
 
--- a/config/discourse_quickstart.conf
+++ b/config/discourse_quickstart.conf
@@ -61,6 +61,9 @@ developer_emails =
 # redis server port
 # redis_port = 6379
 
+# redis server socket, if using socket
+# redis_socket =
+
 # redis password
 # redis_password =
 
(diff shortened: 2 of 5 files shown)


coding-horror:
Should we just close this?

riking:
Actually, yes. The `link:\n  data: data` in the YML file does the same job for less configuration.

HLFH:
@coding-horror @SamSaffron So if I understand, I can't use Redis with Discourse because Discourse does not support 

In `config.php` on the ownCloud webapp that is using Redis (for transactional file locking feature), I have this kind of stuff:
```php
  'redis' =>
  array (
    'host' => '/var/run/redis/redis.sock',
    'port' => 0,
    'timeout' => 0,
    'password' => 'blablabla',
  ),
  'maintenance' => false,
```

And in `discourse.conf`, I have this kind of stuff:
```
# Redis

# redis server address
redis_host = /var/run/redis/redis.sock

# redis server port
redis_port = 0

# redis password
redis_password = blablabla
```

But then, I can't start Discourse because of the `Name or service not known issue`... And you didn't merge this PR... So... Do you confirm I should completely remove Redis configuration of my `discourse.conf`? Thanks.

swissslinky:
I would like to see support for a DISCOURSE_REDIS_SOCKET environment variable.

We're running multiple discourse instances under uwsgi (not as docker containers, I know, I've read the docs, please dont...) and for the most part things are working as expected.

However, we run multiple redis instances and I cannot configure discourse to use a socket. As such each instance is defaulting to `localhost` and they're fighting it out. We notice this when we change SSO on one instance and it affects others too. 

If you're really against the above pull request and/or support Redis sockets, how can I disable the redis cache completely?

PR #2990, opened 2014-11-21
Removed titleColSpan of uncategorized topics

rumpelsepp:
This behaviour is not consistent any more. We can toggle "suppress uncategorized badge" in the admin page which has this description: "Don't show the badge for uncategorized topics in topic lists."

So I think the intended behaviour is "show the category", or "don't show the category". Not showing the badge on pinned topics is quite confusing when it's explicitly enabled in the admin settings.

There is also a thread an meta.discourse.org: https://meta.discourse.org/t/22245


coding-horror:
https://meta.discourse.org/t/uncategorized-badge-is-not-shown-on-pinned-topics/22245/4?u=codinghorror

PR #3058, opened 2014-12-26
extends share links

fantasticfears:
- extract font awesome class from share link model
- add some chinese share services to core from plugin
- add class to allow plugin hooking into the service


coding-horror:
Probably this should be closed in light of @eviltrout recent changes here to make extensibility easier in share dialog? Follow up with him on meta.

PR #3060, opened 2014-12-27
fix onebox cannot detect relative url bug

caogecym:
With this fix, url starts with '//' will be considered as a link by `better_markdown`, and thus able to be converted to onebox preview.

### DIFF
--- a/app/assets/javascripts/discourse/dialects/autolink_dialect.js
+++ b/app/assets/javascripts/discourse/dialects/autolink_dialect.js
@@ -3,7 +3,7 @@
   a hrefs for them.
 **/
 var urlReplacerArgs = {
-  matcher: /^((?:https?:(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.])(?:[^\s()<>]+|\([^\s()<>]+\))+(?:\([^\s()<>]+\)|[^`!()\[\]{};:'".,<>?«»“”‘’\s]))/gm,
+  matcher: /^((?:https?|(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.])(?:[^\s()<>]+|\([^\s()<>]+\))+(?:\([^\s()<>]+\)|[^`!()\[\]{};:'".,<>?«»“”‘’\s]))/gm,
   spaceOrTagBoundary: true,
 
   emitter: function(matches) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/dialects/autolink_dialect.js
```diff
@@ -3,7 +3,7 @@
   a hrefs for them.
 **/
 var urlReplacerArgs = {
-  matcher: /^((?:https?:(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.])(?:[^\s()<>]+|\([^\s()<>]+\))+(?:\([^\s()<>]+\)|[^`!()\[\]{};:'".,<>?«»“”‘’\s]))/gm,
+  matcher: /^((?:https?|(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.])(?:[^\s()<>]+|\([^\s()<>]+\))+(?:\([^\s()<>]+\)|[^`!()\[\]{};:'".,<>?«»“”‘’\s]))/gm,
```
SamSaffron: this does not look right to me, the regex semantics heavily changed (and probably incorrectly) here.

vendor/assets/javascripts/better_markdown.js
```diff
@@ -335,7 +335,7 @@
       // __foo__ is reserved and not a pattern
       if ( i.match( /^__.*__$/) )
         continue;
-      var l = i.replace( /([\\.*+?^$|()\[\]{}])/g, "\\$1" )
+      var l = i.replace( /([\/\\.*+?^$|()\[\]{}])/g, "\\$1" )
```
SamSaffron: I am confused by this change, and worried that you are adding semantics that can easily break elsewhere in the parser

We have found that hacking on the md parser is probably the most complicated and error prone part of hacking on Discourse and this may break stuff like `\*hello*` I don't know.

app/assets/javascripts/discourse/dialects/autolink_dialect.js
caogecym: hmm.. Basically I'm trying to make http(s) optional, because I want to match urls that start with '//'. e.g. //i.imgur.com/ZiIKFs1.mp4

SamSaffron:
sorry, I am closing this, no longer merges cleanly and comments not quite yet addressed. I am just super terrified about changes to our md parser its a bad of bugs.

PR #3062, opened 2014-12-28
FEATURE: Add way to access all screened IP addresses through json

riking:
Expected client behavior:

 1. Get the first 200 screened IP addresses with no query parameters
 2. Take the match_count and id of the last entry
 3. Request with ?after={match_count},{id}
 4. Repeat steps 2-3 until less than 200 are returned

No UI is provided in this commit.

https://meta.discourse.org/t/ip-address-no-longer-shown-on-the-admin-users-listings/23320/8?u=riking


coding-horror:
Moderators should be able to perform that export, has been confirmed a few times, so closing this for now.

PR #3073, opened 2015-01-01
FEATURE: Change email via API without authorization email.

oblakeerickson:
This commit allows someone using the api to change a users email without
generating an email_token and sending an authorization email.

- add 'confirmed' to user_params
- return json response with user_id and updated email
- return json success when using Jobs.enqueue
- test that email validations still work
- test that email token is not created with confirmed flag set
- test that Jobs.enqueue is not called with confirmed flag set
- test that email is update with confirmed flag set
- removes parenthesis from expires_now to be consistant

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -281,7 +281,7 @@ class UsersController < ApplicationController
   end
 
   def password_reset
-    expires_now()
+    expires_now
 
     if EmailToken.valid_token_format?(params[:token])
       @user = EmailToken.confirm(params[:token])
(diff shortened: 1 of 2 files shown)


eviltrout:
Wouldn't this allow anyone to just pass `confirmed=true` and not validate their email? That seems bad.

oblakeerickson:
I would want it scoped to just administrators. I was thinking that similar functionality exists with create user when you pass in `active=true` but apparently all the does is delay when the confirmation email is sent. 

This is the response I get back:

    {"success"=>true, "active"=>true, "message"=>"Your account is activated and ready to use.", "user_id"=>19}

But when I try to login as the new user it still wants me to confirm my email and I have to click the resend confirmation button. So I'm not sure what `active=true` even does?

I was thinking that this would be a good feature to have if using SSO or something and the email has already been confirmed in another app not to have to confirm it again. If you have a better way of accomplishing something similar I'm open to that as well. If this really is just a bad idea we can just close the pull request.

PR #3076, opened 2015-01-02
Update Translations

techAPJ:
cc @nlalonde


techAPJ:
@nlalonde: I think there's an issue with Transifex translation updates.

Notice that `csv_export_succeeded` is deleted from many `server` locale files, while it's present in `server.en.yml` file. Similarly `welcome_user.text_body_template` is deleted from many locale files which is unexpected.

Please have a look here.

nlalonde:
Text in server.en.yml for csv_export_succeeded changed on 2014-12-24, so Transifex considers it untranslated again. Translation activity has been pretty low during the holidays.

techAPJ:
@nlalonde: thanks, got it!

PR #3079, opened 2015-01-05
models with rspec3 syntax

lucianosousa:


lucianosousa:
@ZogStriP models are updated! o/

PR #3080, opened 2015-01-05
Miscellaneous fixes from PR#3000

riking:
Extracted the actual bugfixes from #3000.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/static.js.es6
+++ b/app/assets/javascripts/discourse/controllers/static.js.es6
@@ -3,7 +3,9 @@ export default Em.ObjectController.extend({
 
   actions: {
     markFaqRead: function() {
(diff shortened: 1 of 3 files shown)


app/controllers/list_controller.rb
```diff
@@ -115,6 +115,7 @@ class ListController < ApplicationController
       target_user = fetch_user_from_params
       guardian.ensure_can_see_private_messages!(target_user.id) unless action == :topics_by
       list = generate_list_for(action.to_s, target_user, list_opts)
+      url_prefix = nil
```
SamSaffron: this is not needed 

```
sam@ubuntu discourse % pry
[1] pry(main)> a = 1 if false
=> nil
[2] pry(main)> a
=> nil
```

SamSaffron:
thanks

PR #3081, opened 2015-01-06
FIX: Issue with Vagrant image failing specs.

edgibbs:
Submitting this in response to a discussion thread:

https://meta.discourse.org/t/why-are-my-tests-failing/23434

It turns out the Vagrant file was using ruby-2.0.0-p0-turbo.  I updated the chef solo file for discourse to add a Ruby 2.1 image with rvm and select it.  A better solution is probably to update the discourse base box, but this works for now.


coding-horror:
This is great and much needed as we have not had a lot of time to improve
our Vagrant box! Thank you!

SamSaffron:
:heart:

PR #3083, opened 2015-01-08
UX: Add class to gutter links so you can infer if they are inbound or outbound using CSS

cpradio:
Added a second class to the gutter links to indicate if they are inbound or outbound links so sites can customize the two types differently (if they want to).

No changes were made to the default CSS, so they remain without any physical indicator.


eviltrout:
Looks good to me, thanks!

coding-horror:
:+1:

PR #3004, opened 2014-11-25
Incremental membership editing on /groups/xxx/members

jmay:
First cut at feature for https://meta.discourse.org/t/improved-group-members-management/16768/14

This is not ready for merge and there are no tests. I'm a novice with the Ember stuff, hopefully I'm not far wrong in the approach. Perhaps @eviltrout could take a look?

PR #3084, opened 2015-01-08
table & model changes for group managers with permission to edit members

jmay:
Extracted the just the back-end parts from https://github.com/discourse/discourse/pull/3004

Includes API changes on `/groups/xxx` (not `/admin/groups/xxx`)

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -7,6 +7,9 @@ class Group < ActiveRecord::Base
   has_many :categories, through: :category_groups
   has_many :users, through: :group_users
 
+  has_many :group_managers, dependent: :destroy
+  has_many :managers, through: :group_managers
+
   after_save :destroy_deletions
 
   validate :name_format_validator
@@ -275,6 +278,10 @@ class Group < ActiveRecord::Base
 
   def remove(user)
     self.group_users.where(user: user).each(&:destroy)
+  end
+
+  def appoint_manager(user)
+    managers << user
   end
 
   protected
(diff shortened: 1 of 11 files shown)


jmay:
Thanks for the feedback, all applied.

jmay:
Travis shows red, but the failure seems to be a network issue when running the javascript suite, not anything with the app. Is there a way to trigger a test rerun?

jmay:
Maybe just merging master will give it a kick?

app/models/group.rb
```diff
@@ -277,6 +280,10 @@ def remove(user)
     self.group_users.where(user: user).each(&:destroy)
   end
 
+  def appoint_manager(user)
```
ZogStriP: Not sure that method is needed since it's only used in tests.

ZogStriP:
Just added a comment. Once fixed, don't forget to squash the commits into 1 and I'll get that merged ;)

app/models/group.rb
jmay: I put that in so there is a place for the controller to invoke once there is a UI for this (I'm working on it, slowly). Not sure where the best separator is between back-end and front-end stuff, and I didn't want to touch the front-end with this patch.

ZogStriP: That's fine. Can you squash the commits into 1 and then I'll merge that ;)

jmay:
There we go. *rebase is the mind-killer. rebase is the little-death that brings total obliteration.* I will face my fear.

PR #3085, opened 2015-01-09
FEATURE: suppress slug site setting

fantasticfears:
Related topic: https://meta.discourse.org/t/disabling-english-slugfier-for-topics-url-in-korean-language-setting/23780

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -207,7 +207,7 @@ SQL
 
     if slug.present?
       # santized custom slug
-      self.slug = Slug.for(slug)
+      self.slug = Slug.safe_for(slug)
       errors.add(:slug, 'is already in use') if duplicate_slug?
     else
       # auto slug
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -713,6 +713,8 @@ uncategorized:
     client: true
     default: true
 
+  suppress_slug: false
+
   # Search
   min_search_term_length:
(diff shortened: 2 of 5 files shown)


app/models/category.rb
```diff
@@ -207,7 +207,7 @@ def ensure_slug
 
     if slug.present?
       # santized custom slug
-      self.slug = Slug.for(slug)
+      self.slug = Slug.safe_for(slug)
       errors.add(:slug, 'is already in use') if duplicate_slug?
     else
       # auto slug
```
fantasticfears: I don't think so. As the topic described, custom slug should still be available in that circumstance since this site setting is used for maximizing the ability to customized non-english slug.

config/site_settings.yml
```diff
@@ -713,6 +713,9 @@ uncategorized:
     client: true
     default: true
 
+  suppress_slug:
```
fantasticfears: resolved

SamSaffron:
this change feels odd to me, for starters, if you are totally disabling slugs why is category still getting a slug? it just feels odd

also I am really worried that all sorts of parts in the system will start collapsing if Slug.for start returning a blank string. it feels like a quick hack as opposed to a targetted change. 

can you link to the discussion on meta about this?

fantasticfears:
The discussion is on the first comment. But you are right, I can do better than that

PR #3092, opened 2015-01-12
FIX: users tab in admin panel doesn’t stay highlighted

techAPJ:
[Bug report on meta](https://meta.discourse.org/t/users-tab-in-admin-panel-doesnt-stay-highlighted/23637?u=techapj)


eviltrout:
Thanks!

PR #3095, opened 2015-01-13
FIX: handle timezone detection errors

Elberet:
Fixes https://meta.discourse.org/t/importer-for-simple-machines-2-forums/17656/51?u=elberet and provides a means to specify the timezone if PHP isn't installed.


ZogStriP:
Looks fine to me. Thanks :+1:

PR #3096, opened 2015-01-13
gzip csv export files

techAPJ:
Zip CSV export files to save bandwidth and disk space.

### DIFF
--- a/app/jobs/regular/export_csv_file.rb
+++ b/app/jobs/regular/export_csv_file.rb
@@ -288,12 +288,14 @@ module Jobs
             csv << value
           end
         end
+        # compress CSV file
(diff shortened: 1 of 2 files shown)


app/jobs/regular/export_csv_file.rb
```diff
@@ -288,12 +288,14 @@ def write_csv_file(data, header)
             csv << value
           end
         end
+        # compress CSV file
+        `gzip --best #{File.expand_path("#{UserExport.base_directory}/#{@file_name}", __FILE__)}`
```
riking: Do we really need to be doing `-best`? I think -5 is the best compromise between speed / size

coding-horror: yeah no -best just -5 is fine.

techAPJ: Ok, done. #3101

PR #3100, opened 2015-01-14
Set Turkish pluralization rules

riking:
Turkish does not have different forms for singulars next to numerals
https://meta.discourse.org/t/turkish-translation-missing-many-singular-strings/23334/12


techAPJ:
:heart:

PR #3105, opened 2015-01-17
Improve noscript handling for old browsers

riking:
Previously, if a browser needed a feature that was required for Discourse to render anything, you would get a yellow banner saying 

>Sorry, but your browser is too old to work on this Discourse forum. Please upgrade your browser.

... with a vast sea of white below it.

This is because the "fallback content" is in a `<noscript>` element, and it's impossible to fetch the HTML out of a noscript element in a compatible way across all old browsers.

So, this PR:

 1. Changes the `<noscript>` to a `<div id="noscript">`
 2. Changes the `$('noscript').remove();` to `$('#noscript').remove();` in Discourse.start()
 3. Gates the removal of the noscript element by browser-update.js running, which will set `window.HaltDiscourse = true` if the browser is old
 4. Gives the html element a CSS class of `js` or `no-js` depending on the result of browser-update.js

Additionally, dead code in the StaticPage model was rewritten: by the time the code had executed, the noscript element would be long gone. It was changed to use PreloadStore.

image

### DIFF
--- a/app/assets/javascripts/discourse.js
+++ b/app/assets/javascripts/discourse.js
@@ -100,8 +100,13 @@ window.Discourse = Ember.Application.createWithMixins(Discourse.Ajax, {
     @method start
   **/
   start: function() {
-
-    $('noscript').remove();
+    if (window.HaltDiscourse) {
+      // Browser is too old to run Discourse. Don't start.
+      // see browser-update.js
+      return;
+    } else {
+      $('#noscript').remove();
+    }
 
     // Load any ES6 initializers
     Ember.keys(requirejs._eak_seen).forEach(function(key) {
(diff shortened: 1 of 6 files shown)


riking:
@coding-horror I got you your no-js view ;)

SamSaffron:
I am not to keen on this change, you added a bunch of DOM now that is going to get parsed for every browser and CSS magic to match.

Would prefer a magic query string and an extra redirect for these old browers

EG:

http://site/latest > client detects it is screwed > redirect to http://site/latest?_html_only=1&_old_browser=1

which then gives us a simple way to test and a simple way to view html_only

coding-horror:
yeah extra redirect sounds good. We don't want to clutter the default design for IE9 or any old browsers.

PR #3106, opened 2015-01-17
Update pretty_text.rb

demil:
It was causing 500 internal server error on create posts because ruby couldnt find file because of paths.
https://meta.discourse.org/t/after-upgrade-to-0-9-9-18-unknown-error-saving-post-try-again-error-500-internal-server-error/19190
I don't know why it works on some installs and don't on others, because i am not ruby developer.

### DIFF
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -82,7 +82,7 @@ module PrettyText
     end
 
     # custom emojis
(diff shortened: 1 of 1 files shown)


riking:
I think you should switch out `Rails.root` for `__FILE__/..`.

coding-horror:
Follow up on meta.discourse -- we only support the Docker based install.

PR #3107, opened 2015-01-17
Update letter_avatar.rb

demil:
Again error with paths, and again as i am not ruby developer i don't know why on some installs works but on others don't.

### DIFF
--- a/lib/letter_avatar.rb
+++ b/lib/letter_avatar.rb
@@ -24,7 +24,7 @@ class LetterAvatar
     end
 
     def cache_path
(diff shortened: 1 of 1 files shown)


ZogStriP:
What kind of error are you getting? This is the first time we've had that report. Also, are you using our recommended Docker setup?

demil:
Error is ENOENT, no such a file or directory. No i am using manual install, redis 2.8 , postgre 9.1, ruby 2.2, CENTOS 6.4.

coding-horror:
Follow up on meta.discourse -- we only support the Docker based install.

PR #3110, opened 2015-01-18
FIX: Respect user locale for reply notifications

riking:
https://meta.discourse.org/t/mislocalization-of-x-replies-in-notifications/24030

Copied the language selection logic from application_controller


eviltrout:
Thanks :+1:

PR #3111, opened 2015-01-18
Fix profile update when 'custom_fields' is blank

Flink:
It seems when custom user fields are enabled and exist, a
'custom_fields' parameter with an empty string as its value is sent when
updating a user profile.
This commit addresses this issue by checking that 'custom_fields' is
blank instead of just nil and assigns it an empty hash as it was done
before.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -66,7 +66,7 @@ class UsersController < ApplicationController
     guardian.ensure_can_edit!(user)
 
     if params[:user_fields].present?
-      params[:custom_fields] ||= {}
+      params[:custom_fields] = {} if params[:custom_fields].blank?
       UserField.where(editable: true).each do |f|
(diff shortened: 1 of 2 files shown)


eviltrout:
Wouldn't this prevent a user from clearing out a custom field? I think the proper fix might involve not sending it unless it is updated?

Flink:
I don’t know exactly what is the use of `params[:custom_fields]` but values of custom fields are not in this parameter. They are in `user_fields`. And I did not really change the current behavior, only checking for blank instead of just nil.
But I agree with you that it’ll be better to find why this parameter is sent as an empty string thus leading to a 500 error.

SamSaffron:
I object to this change, it should be fixed on the client, it should not be sending an empty string here.

Flink:
So after all, I see my change wasn’t so useless ;)

PR #3112, opened 2015-01-19
FIX: convert UTF8 charset to UTF-8

techAPJ:


coding-horror:
Hmm I think we should make a less global replace than this...

Looks like the string is

charset="UTF8"

We should replace only that IMO

techAPJ:
@coding-horror: ok, updated.

PR #3113, opened 2015-01-19
FIX: Move vagrant box to discourse hosting

edgibbs:
This change just moves the Vagrant box to the discourse account at
Atlas.


coding-horror:
:+1:  thanks @techapj for assisting here!

techAPJ:
:heart:

PR #3119, opened 2015-01-21
FIX: user name and title were not showing on post creation

techAPJ:
Before:

screen shot 2015-01-22 at 12 14 14 am

After:

screen shot 2015-01-22 at 2 16 31 am


eviltrout:
Nice one!

ZogStriP:
:+1:

PR #3122, opened 2015-01-23
FEATURE: unbind font awesome icon in share button

fantasticfears:
- extract font awesome class from share link
- add another share services

PS: `<input>` is a void element which doesn't need to close itself in HTML5. So I removed it.

### DIFF
--- a/app/assets/javascripts/discourse/initializers/sharing-sources.js.es6
+++ b/app/assets/javascripts/discourse/initializers/sharing-sources.js.es6
@@ -14,7 +14,7 @@ export default {
 
     Sharing.addSource({
       id: 'twitter',
-      iconClass: 'fa-twitter-square',
+      faIcon: 'fa-twitter-square',
       generateUrl: function(link, title) {
         return "http://twitter.com/intent/tweet?url=" + encodeURIComponent(link) + "&text=" + encodeURIComponent(title);
       },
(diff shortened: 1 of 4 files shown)


ZogStriP:
Why did you do make the share button icon unbound? Are there any performance issues?

fantasticfears:
Simply because there are icons not existing in the font awesome. At least I need a way to use a icon when add a custom service by plugin.

I don't know how to measure the performance in this case. Ember Inspector told me they rendered under 1ms.

eviltrout:
The assumption I made is that it would be a font awesome icon which indeed is always the case in stock discourse.

I'm definitely in favor of letting people use images for icons but I think repurposing "icon" seems a bit odd.

I'd prefer if we changed icon to be `'faIcon` with the existing behavior, and then `htmlIcon` for custom HTML.

fantasticfears:
@eviltrout done

SamSaffron:
we can't be adding redit linkedin and tumblr by default that just feels wrong to me cc @coding-horror

coding-horror:
yes agreed the defaults should not change. Extensibility is the goal here, not changing the defaults.

fantasticfears:
They were removed

SamSaffron:
sorry, needs a rebase

fantasticfears:
@SamSaffron Rebased. The branch was forced updated.

ZogStriP:
:+1: Thanks

PR #3124, opened 2015-01-23
Improve the base importer

gschlager:
- FEATURE: Print the time it took for the import to finish
- FEATURE: Allow importing of unactivated users
- FIX: Remove unused variables
- FIX: Accessing an undefined variable is not a good idea


ZogStriP:
Thanks :+1:

PR #3125, opened 2015-01-23
FIX: Loading of Emoji files depended on working directory

gschlager:
The importer scripts could not be used unless the working directory was the Discourse root directory.


ZogStriP:
Thanks

PR #3126, opened 2015-01-24
Latest posts endpoint at /posts.json

riking:
It will return the posts with the **last 50 ids**, in order, with posts you can't see filtered out. You can attach a `?before=5678` to only see post IDs 5629 thru 5678. With that scheme, it should be fairly easy to make sure your service sees every post it has access to, even if more than 50 posts occur between two checks.

I think this will satisfy several people's needs for an extensibility point. If they need message bus notifications, I think we could do that too.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -5,7 +5,7 @@ require_dependency 'distributed_memoizer'
 class PostsController < ApplicationController
 
   # Need to be logged in for all actions here
-  before_filter :ensure_logged_in, except: [:show, :replies, :by_number, :short_link, :reply_history, :revisions, :latest_revision, :expand_embed, :markdown, :raw, :cooked]
+  before_filter :ensure_logged_in, except: [:show, :replies, :by_number, :short_link, :reply_history, :revisions, :latest_revision, :expand_embed, :markdown_id, :markdown_num, :cooked, :latest]
 
   skip_before_filter :check_xhr, only: [:markdown_id, :markdown_num, :short_link]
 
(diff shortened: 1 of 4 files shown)


app/controllers/posts_controller.rb
```diff
@@ -5,7 +5,7 @@
 class PostsController < ApplicationController
 
   # Need to be logged in for all actions here
-  before_filter :ensure_logged_in, except: [:show, :replies, :by_number, :short_link, :reply_history, :revisions, :latest_revision, :expand_embed, :markdown, :raw, :cooked]
+  before_filter :ensure_logged_in, except: [:show, :replies, :by_number, :short_link, :reply_history, :revisions, :latest_revision, :expand_embed, :markdown_id, :markdown_num, :cooked, :latest]
```
riking: This is correct, anons should be able to view post raws.

SamSaffron:
I am uncomfortable with a hardcoded limit 50 that should be shippable from a param with a hard limit to avoid abuse (maybe 200) 

additionally the semantics of bounding on post ids and ordering on created date feels a bit odd, the api lacks a bunch of flexibility, eg: what posts happened today? what posts happened yesterday?

riking:
The ordering by created at is so that the sort will be stable across checks, even with edits. I imagine a message bus `/edits` channel so that all edits can be caught if a consumer is interested in that.

I don't see anything more wrong with a limit of 50 than a limit of 200. Could bump it up to 100? I didn't want the responses getting too big.

Getting the posts that happened today feels like a YAGNI situation to me. Just keep requesting until you have a `created_at` that's a day ago. Maybe that's one request, maybe it's 10 on a high-traffic forum with 500 posts a day. It's certainly possible as currently designed.

My rationale for the bounding on post IDs is to make the discoverability of "Which posts are *deleted* and which are *not visible to me*?" impossible, even in the face of mistakes in the query.

SamSaffron:
I am merging this , but feel we should tune this api some, in particular if ids ever become out of order it becomes a nightmare.

stefanvaduva:
@riking, can topic title be added for context? I know we can use the topic ID to grab the title, but doing another API request just to get the title is not very efficient. And I think the title is something that will be needed a lot. 

And thank you for this addition!

PR #3128, opened 2015-01-25
Convert many buttons to `d-button`

fantasticfears:


app/assets/javascripts/admin/controllers/admin-backups-index.js.es6
```diff
@@ -10,11 +10,11 @@ export default Ember.ArrayController.extend({
 
   restoreTitle: function() {
```
eviltrout: If you're changing what a property returns (key instead of translation) I think you should rename the CP to reflect it. This is no longer the `restoreTitle` but `restoreTitleKey`

fantasticfears:
@eviltrout done

SamSaffron:
sorrry, needs a rebase

ZogStriP:
still needs a rebase. Not sure it's still relevant though.

fantasticfears:
@ZogStriP I'm closing this now. `d-button` component is good to me now. Other buttons can be updated gradually.

ZogStriP:
:+1:

PR #3129, opened 2015-01-26
Added JS to add category as a class for additional styling purposes

lagwagon:
I thought that category-based (by name) styling could be useful (specifically for my use, possibly others?). Maybe not useful for the average end user, but for those of us who wish to theme our own websites it could be useful. 

Added this in and it works well. 

categories

### DIFF
--- a/app/assets/javascripts/discourse/views/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/views/topic-list-item.js.es6
@@ -10,8 +10,17 @@ export default Discourse.View.extend(StringBuffer, {
        ],
 
   unboundClassNames: function(){
-    var classes = [];
     var topic = this.get('topic');
+    var classes = [];
+    
+    if(topic.get('category')){
+      var category, category_name, category_class;
+      category = topic.get('category');
+      category_name = category.get('name').toLowerCase();
+      category_class = category_name + '-category';
+
+      classes.push(category_class);
+    }
 
     if(topic.get('hasExcerpt')){
       classes.push('has-excerpt');
--- a/app/assets/javascripts/discourse/views/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/views/topic-list-item.js.es6
@@ -14,10 +14,9 @@ export default Discourse.View.extend(StringBuffer, {
     var classes = [];
 
     if(topic.get('category')){
-      var category,categorySlug,categoryClass;
-      category = topic.get('category');      
-      categorySlug = category.get('slug');
-      categoryClass = 'category-' + categorySlug;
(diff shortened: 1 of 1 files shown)


coding-horror:
shouldn't it be `category-foo` ? so that things sort properly?

app/assets/javascripts/discourse/views/topic-list-item.js.es6
```diff
@@ -10,8 +10,17 @@ export default Discourse.View.extend(StringBuffer, {
...
   unboundClassNames: function(){
-    var classes = [];
     var topic = this.get('topic');
+    var classes = [];
+    
+    if(topic.get('category')){
+      var category, category_name, category_class;
+      category = topic.get('category');
+      category_name = category.get('name').toLowerCase();
+      category_class = category_name + '-category';
```
ZogStriP: This should be replaced by this oneliner

```javascript
var className = "category-" + topic.get('category.nameLower');
```

SamSaffron: what if a category name has a space in it or something that needs url
encoding? I am uncertain about this change, feels like custom theme
territory.

ZogStriP: Yeah, should probably use the slug instead

lagwagon:
Hey guys - Thanks for the comments. I've made some changes, namely to use the slug, and used Jeff's suggestion for how the class name appears as it does look better. 

The slug already takes care of escaping html characters and it seems to be fairly reliable in providing something that meshes well with the CSS overall.

Right now there is no way to style posts made in certain categories in a friendly way. This is a pretty huge thing that is found in nearly every cms or board software out there, I figured it's a nice-to-have.

app/assets/javascripts/discourse/views/topic-list-item.js.es6
```diff
@@ -10,8 +10,17 @@ export default Discourse.View.extend(StringBuffer, {
...
     var topic = this.get('topic');
+    var classes = [];
+
+    if(topic.get('category')){
+      var category,categorySlug,categoryClass;
+      category = topic.get('category');      
+      categorySlug = category.get('slug');
+      categoryClass = 'category-' + categorySlug;
+
+      classes.push(categoryClass);
```
ZogStriP: `classes.push("category-" + topic.get('category.slug'))` ;)

lagwagon: Haha okay, made that change too. Thanks for the advice! I generally am not writing javascript...ever.. anymore, so every comment helps me

PR #3131, opened 2015-01-26
sso_secret: cryptographically authenticate, not encrypt.

pda:
The sso_secret is an input to HMAC, which is a hash-based message authentication code, not encryption.  Perhaps this is a minor nitpick, but I think it's important to get concepts and language correct when it comes to cryptography.  I noticed this in the screenshot at https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045

This PR corrects the English phrase, but I'm not sure what the process is to get those translations updated.


pda:
CLA signed.

coding-horror:
thanks :+1:

PR #3132, opened 2015-01-27
Perf UX : reduce page paint on scrolling, due to fixed header element union with progress bar area

lidlanca:
Should provide some performance improvement for scrolling on chrome
The fixed header and the fixed progress bar on both ends  of the screen will invoke a complete visible page redraw.

For whatever reason this css should make things better for low performance devices. only tested on chrome superficially.

### DIFF
--- a/app/assets/stylesheets/common/base/header.scss
+++ b/app/assets/stylesheets/common/base/header.scss
@@ -8,6 +8,7 @@
 
     .docked & {
       position: fixed;
+      backface-visibility: hidden;  /** do magic for scrolling performance **/
     }
 
     .contents {


SamSaffron:
this seems fine to me .. we can try it out.

lidlanca:
The only thing I noticed is that on firefox(34) on windows.
An image or possibly text that is inside a "spoiler plugin"  with bluring affect.

Under certain conditions such as scrolling the page enough. and scrolling back to see the element, it might be replaced with an empty white space. The image is not getting repainted.

on firefox 35 I experienced similar behavior, with a slight difference, the image got repainted eventually  after few seconds and some hovering.

Looks like firefox might not behave nicely with that combination, perhaps a bug with their engine

if that becomes an issue, it is possible to target only webkit chrome android and such using 
-webkit-backface-visibility: hidden;

I am not even sure if this even benefit firefox performance wise.

PR #3133, opened 2015-01-27
Disable digest emails site setting

techAPJ:


coding-horror:
If you disable it you need to remove the UI in user preferences as well.

techAPJ:
@coding-horror: Good point. I made it so.

PR #3134, opened 2015-01-27
fix add nginx key command error

JavanEu:
When add nginx key, command lost key file name.


eviltrout:
What was the error? The original command looks OK to me?

SamSaffron:
yeah closing it, my guess it that the error is that curl is not installed. but wget syntax is lot less nice.

PR #3135, opened 2015-01-27
FIX: creating a user shouldn't error when optional fields aren't given

longhotsummer:
This fixes a bug where the server would 500 if the only user fields
where optional ones, and the create_user call didn't provide any
values so that `params[:user_fields]` was nil.

Additionally, don't bother double-checking for required fields, since we
iterate over all fields and will catch any that are required and blank.


SamSaffron:
seems good to me, thanks!

PR #3140, opened 2015-01-28
Fix bug when sso_overrides_avatar is true but no avatar_url is passed

dansingerman:
If a user has a current avatar, and sso_overrides_avatar is true, but no avatar_url is
passed in the sso attributes, the current code errors, as it tries to parse a nil
as a URL. It seems to me valid that a third party system may not pass an avatar_url in
some cases (e.g. avatars may not be mandatory, so not all users may have them)

This might warrant a discussion about what should happen in this case; maybe the current
avatar in discourse should be removed? This branch merely stops the login process erroring.


SamSaffron:
this looks good, thanks!

PR #3142, opened 2015-01-29
Fix quoted links giving blank page on click

riking:
https://meta.discourse.org/t/links-in-quotes-sometimes-not-working/24540/

### DIFF
--- a/app/controllers/clicks_controller.rb
+++ b/app/controllers/clicks_controller.rb
@@ -18,8 +18,10 @@ class ClicksController < ApplicationController
 
     # Sometimes we want to record a link without a 302. Since XHR has to load the redirected
     # URL we want it to not return a 302 in those cases.
-    if params[:redirect] == 'false' || @redirect_url.blank?
-      render nothing: true
+    if @redirect_url.blank?
+      render text: 'bad link', status: 422
+    elsif params[:redirect] == 'false'
+      render nothing: true, status: 201 # 201 Created
     else
       redirect_to(@redirect_url)
     end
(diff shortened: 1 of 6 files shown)


app/controllers/clicks_controller.rb
```diff
@@ -5,21 +5,28 @@ class ClicksController < ApplicationController
...
+        params.merge!({ user_id: current_user.id }) if current_user.present?
         @redirect_url = TopicLinkClick.create_from(params)
+
+        if @redirect_url.blank? && params[:url].index('?')
+          # Check the url without query parameters
+          params[:url].sub!(/\?.*$/, '')
+          @redirect_url = TopicLinkClick.create_from(params)
+        end
       end
+    rescue => e
```
SamSaffron: hmmm, why is this needed?

SamSaffron:
can I see some examples in the wild of the problem this is solving (make a meta post)

coding-horror:
It's this

https://meta.discourse.org/t/links-in-quotes-sometimes-not-working/24540

app/controllers/clicks_controller.rb
riking: Removed

db/migrate/20150129082106_add_from_quote_to_topic_link.rb
```diff
@@ -0,0 +1,5 @@
+class AddFromQuoteToTopicLink < ActiveRecord::Migration
+  def change
+    add_column :topic_links, :from_quote, :boolean, default: false, null: false
```
SamSaffron: I dunno, this is adding a feature we don't really need to add just to fix a bug, can we just fix bug instead?

riking: There's another problem hiding here - yt-light creates links in javascript that don't show up in the cooked, so they also give the "link not found" white screen.

I now agree, this is the wrong way to fix it.

SamSaffron:
I am closing this for now, let's come up with another fix

PR #3145, opened 2015-01-29
Rerender actions on like/dislike

gdpelican:
For https://meta.discourse.org/t/like-button-and-like-it-too-undo-like-are-not-synchronized/23516/3


eviltrout:
Thanks!

PR #3152, opened 2015-01-30
Fix force_avatar_update.to_i error as force_avatar_update is a boolean

dansingerman:
If force_avatar_update is passed in sso attributes it errors on
force_avatar_update.to_i. The SingleSignOn class forces avatar_force_update
to a boolean, so it should be treated as such.


ZogStriP:
Thanks :+1:

PR #3155, opened 2015-01-31
Support for relative URLs and discourse-only docker containers

vgeshel:
This set of changes does two things. They are mostly unrelated but they were essential for us in building our Discourse instance and, I believe, they are generally useful.

- Better support for RAILS_RELATIVE_URL: we have fixed numerous places where it wasn't respected.

Running under a a relative URL (domain.com/forum/...) vs subdomain (forum.domain.com) is very useful from the SEO point of view, see, for instance, [here](http://moz.com/community/q/moz-s-official-stance-on-subdomain-vs-subfolder-does-it-need-updating#reply_217479). 

- Better support for deployments with external Postgres, Redis and nginx.

The repo contains a Dockerfile which builds an image that includes only Discourse itself. This makes it easier to deploy, e.g., in AWS, using RDS (hosted Postgres) and ElastiCache (hosted Redis). The repo also contains a `fig.yml`, which helps you use [fig](http://www.fig.sh/) to do local development using an equivalent setup (the Vagrant-based setup still works just fine). The fig setup is documented in `docs/fig.md`.


vgeshel:
Closed while I fix test failures

Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM ruby:2.1
```
riking: The Dockerfile belongs in https://github.com/discourse/discourse_docker

app/views/users/activate_account.html.erb
```diff
@@ -10,14 +10,25 @@
 </div>
 
 <%- content_for(:no_ember_head) do %>
+  <%= script "preload_store" %>
```
riking: This doesn't make sense, just switch it over to the `application` layout, don't recreate the includes in another layout.

config/database.yml
```diff
@@ -1,6 +1,8 @@
 development:
   adapter: postgresql
-  database: discourse_development
+  database: <%= ENV['DISCOURSE_DB_NAME'] || 'discourse_development' %>
```
riking: This is already done for `Rails.env.production`, just reuse that code

Dockerfile
vgeshel: This Dockerfile builds an image that contains just Discourse itself. It's common for projects that support building as docker images in this kind of simple way to include a Dockerfile.

As far as I understand, the purpose of discourse_docker is quite different: it's to let a user build and run their own, custom discourse docker image (which contains postgres, redis, etc).

app/views/users/activate_account.html.erb
vgeshel: I just needed access to the Discourse object which is not available in the no_ember layout. These changes are the minimum I could find to make that work. There are a lot more differences between the no_ember and application layouts than this, the page would look different if I changed the layout.

config/database.yml
vgeshel: Sorry, not sure what you're referring to. Bottom of which file?

app/views/users/activate_account.html.erb
riking: Or you could just do an erb `<script>` and get the base_url that way?

config/database.yml
riking: https://github.com/discourse/discourse/blob/d64d3aa2e289d99be79d8a1fa3d0294c90770c49/config/database.yml#L36

vgeshel: The current arrangement is: when running with Rails.env.production, use environment variables. Otherwise, use database.yml. Are you suggesting I change that?

riking: Hmm... I think it could be changed so that in any mode, it uses what it finds in the env.

vgeshel: Wouldn't this break people who modify database.yml for development? I'm assuming the current setup exists for a reason...

riking: Well, the idea there is that in development, they wouldn't be in the env, and it would just use what's in the file. While in production, everything is specified in the env.

app/views/users/activate_account.html.erb
vgeshel: Sure. I was trying to make this similar to all the other places where `.ajax` is called (which either use `Discourse.ajax` or `$.ajax(Discourse.getURL('/users/search/users')...` What you are suggesting would make for simpler/smaller code, but different from any other ajax call. Not sure which is better.

Not sure

config/database.yml
vgeshel: Sorry, I'm not following. In production, database.yml is ignored and only environment is used. I thought you were suggesting to do the same for development/test. Did I misunderstand?

vgeshel:
I'm closing this and submitting a new pull request with the same changes but cleaned up commit history.

https://github.com/discourse/discourse/pull/3194

PR #3158, opened 2015-02-02
FIX: Not allow TL4 user to set staff color

techAPJ:
Fixes [this issue](https://meta.discourse.org/t/toggling-staff-color-on-a-post-doesnt-get-applied-when-done-as-tl4-user/24620?u=techapj)


SamSaffron:
seems correct, thanks

PR #3161, opened 2015-02-02
FIX: uncategorized category was not loading updates in edit mode

techAPJ:
Bug originally reported [here](https://meta.discourse.org/t/creating-topics-by-email-to-users-rust-lang-org-or-internals-rust-lang-org-fails-with-email-issue-unknown-to-address/24595?u=techapj).

This bug was affecting all the fields in the `Settings` tab for "uncategorized" category. The fields were not showing as updated when opening in edit mode.

### DIFF
--- a/app/assets/javascripts/discourse/routes/application.js.es6
+++ b/app/assets/javascripts/discourse/routes/application.js.es6
@@ -132,18 +132,13 @@ var ApplicationRoute = Discourse.Route.extend({
     },
 
     editCategory: function(category) {
-      var router = this;
+      var self = this;
 
-      if (category.get('isUncategorizedCategory')) {
-        Discourse.Route.showModal(router, 'editCategory', category);
-        router.controllerFor('editCategory').set('selectedTab', 'general');
-      } else {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/routes/application.js.es6
```diff
@@ -134,16 +134,11 @@ var ApplicationRoute = Discourse.Route.extend({
     editCategory: function(category) {
       var router = this;
 
-      if (category.get('isUncategorizedCategory')) {
-        Discourse.Route.showModal(router, 'editCategory', category);
+      Discourse.Category.reloadById(category.get('id')).then(function (c) {
+        Discourse.Site.current().updateCategory(c);
```
eviltrout: While you're in there, instead of `Discourse.Site.current()` please change it to `self.site.updateCategory(c);` where `self` is defined outside of the closure

techAPJ: Done! Thanks for the optimization tip!

PR #3166, opened 2015-02-03
add table prefix constant to bbpress import script & fix user imports

fantasticfears:
- Added `DB_TABLE_PREFIX` to get rid of replace table name when importing.
- style change: uppercase SQL 
- remove some old table column in wp_users which at least not available after bbpress 2.5.4


SamSaffron:
seems good to me! thanks

PR #3172, opened 2015-02-05
FIX: check for full name in invite email subject

techAPJ:
If the user has not provided full name, the invite email is being sent out with subject:

`(techapj) invited you to join discourse.example.com`

instead, it should be:

`techapj invited you to join discourse.example.com`

### DIFF
--- a/app/mailers/invite_mailer.rb
+++ b/app/mailers/invite_mailer.rb
@@ -9,7 +9,7 @@ class InviteMailer < ActionMailer::Base
 
     # get invitee name (based on site setting)
     invitee_name = invite.invited_by.username
-    if (SiteSetting.enable_names)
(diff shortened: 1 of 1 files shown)


app/mailers/invite_mailer.rb
```diff
@@ -9,8 +9,8 @@ def send_invite(invite)
 
     # get invitee name (based on site setting)
     invitee_name = invite.invited_by.username
-    if (SiteSetting.enable_names)
-      invitee_name = "#{invite.invited_by.name} (#{invite.invited_by.username})"
+    if SiteSetting.enable_names
```
ZogStriP: `if SiteSetting.enable_names && invite.invited_by.name.present` ?

techAPJ: :+1: Done.

PR #3175, opened 2015-02-06
FIX: Remove g and m flags from autolink regex

chapel:
Fixes issue where a non-link is falsely identified as a link and given the link text of a later link.

More info in thread: https://meta.discourse.org/t/http-gets-parsed-incorrectly-in-posts/24866

The `g` and `m` flags together caused the above case. How this regex works, there isn't any reason to use `g` as the regex should match at the start of the given text being tested, and for `m` that is for matching across lines, and links are not valid with multiple lines.


ZogStriP:
Thanks ;)

PR #3177, opened 2015-02-06
Add rails error translation strings

techAPJ:


techAPJ:
Just a note for future reference:

For translating the error messages, Rails will use the following order of translations:

```
activerecord.errors.models.user.attributes.name.blank
activerecord.errors.models.user.blank
activerecord.errors.messages.blank
errors.attributes.name.blank
errors.messages.blank
```

PR #3178, opened 2015-02-08
Refactor methods in Application Controller

tiimgreen:


SamSaffron:
From a cursory look this looks fine, but we have a policy of not accepting these kind of PRs

see:

Sorry :(

PR #3179, opened 2015-02-08
updating the composer category selector style

awesomerobot:
Updating select2 dropdown styles on composer to work with dark themes

discussed here: https://meta.discourse.org/t/dark-color-theme-categories-selection-unusable/24879

screenshot 2015-02-07 23 24 53

:telescope: :crocodile:


SamSaffron:
nice, thanks!

PR #3182, opened 2015-02-08
FIX: Ember regression in read-only banner

riking:
Caused by https://github.com/discourse/discourse/commit/e1826e30


eviltrout:
:fired:

PR #3183, opened 2015-02-08
Consolidate custom exception handling

riking:
This PR started from the following goals:

 - If a client requests JSON, they should get JSON; never something that looks 'pretty close'.
 - Do NOT break existing code. It just so happens that in the majority of cases, none of the code was looking at this stuff.
    - This resulted in a few notable things: the "BAD CSRF" return is not changed, and `rescue_discourse_actions` has a hack to keep behavior for `TopicsController#show`.
 - Errors should be machine-parsable as well as human-readable. If the error is translated, that ruins machine parsing.
 - The client-side error handling is pretty terrible. This commit is one of the building blocks towards improving it.

### Overview

HTML is no longer stuffed into the response for a JSON request except when the client is expecting that.

The standard errors are now machine-parsable, and gained translations.

### Details

The new "standard" error json response now looks like this:

```
{
  errors: ["Human-Readable text", "More human text"]
  error_type: 'rate_limit'
}
```

It is both compatible with the old format ("`errors` is an array of strings") and more machine-readable due to the new error_type field.

`render_json_error` takes a second parameter, `type`, which is placed in the response as `error_type`.

The following exceptions now conform to this standard format: `RateLimiter::LimitExceeded`, `Discourse::NotLoggedIn`, `Discourse::NotFound`, `Discourse::InvalidAccess`, `Discourse::ReadOnly`.

Discourse::NotLoggedIn got a behavior change - if the request is JSON or an XHR, (in addition to the previous "not a GET" condition) it returns an error instead of redirecting to the root page.

Discourse::NotFound and Discourse::InvalidAccess also got a behavior change. *If it was not raised by TopicsController#show*, and the request is a XHR or json request, it returns a JSON error instead of returning HTML.

When the client requests `TopicsController#show`, however, it expects the 404 page body to be returned if the topic doesn't exist. To minimize the chance of breakage, that behavior is placed in `rescue_discourse_actions`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -93,14 +93,14 @@ class ApplicationController < ActionController::Base
       time_left = I18n.t("rate_limiter.hours", count: (e.available_in / 1.hour.to_i))
     end
 
-    render_json_error I18n.t("rate_limiter.too_many_requests", time_left: time_left), 'rate_limit', 429
+    render_json_error I18n.t("rate_limiter.too_many_requests", time_left: time_left), :rate_limit, 429
   end
 
   rescue_from Discourse::NotLoggedIn do |e|
     raise e if Rails.env.test?
 
     if (request.format && request.format.json?) || request.xhr? || !request.get?
-      rescue_discourse_actions('not_logged_in', 403, true)
+      rescue_discourse_actions(:not_logged_in, 403, true)
     else
       redirect_to "/"
     end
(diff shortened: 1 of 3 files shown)


SamSaffron:
can you confirm that non-json 404s get the correct 404 pages still?

riking:
Yep, I tested in my browser.

eviltrout:
I like it. I really want to move us towards more common error handling code in the JS too.

One thing I think should be done is to convert the string types to symbols in the ruby code (I am fine with them being serialized as strings to the client).

Also I don't think we should merge until after we cut the new release which is imminent as this touches a lot of major code paths.

app/controllers/application_controller.rb
```diff
@@ -93,40 +93,43 @@ class RenderEmpty < Exception; end
...
   end
 
   rescue_from Discourse::InvalidAccess do
-    rescue_discourse_actions("[error: 'invalid access']", 403, true) # TODO: this breaks json responses
+    rescue_discourse_actions(:invalid_access, 403, true)
   end
 
   rescue_from Discourse::ReadOnly do
-    render status: 405, json: failed_json.merge(message: I18n.t("read_only_mode_enabled"))
+    render_json_error I18n.t('read_only_mode_enabled'), :read_only, 405
```
SamSaffron: I don't like this api it should be fixed

```
render_json_error I18n.t('read_only_mode_enabled'), type: read_only, status: 405
```

SamSaffron:
this looks ok to me ... merging, lets see how we go.

PR #3190, opened 2015-02-09
Delete old ErrorLog, use Logster for 500 errors

riking:
We will likely need some new ignore patterns in the near future.

This should surface several TypeErrors that were being dumped in a write-only XML file.


riking:
Before:

```
<hash>
  <date type="dateTime">2015-01-10T22:10:41+00:00</date>
  <guid>428e7b2c-3c9e-4040-ad8a-ff64978f8d9b</guid>
  <user-id type="integer">84</user-id>
  <parameters>{"tags":"","topic_id":"221","_":"1420927713082","controller":"tagger/tags","action":"set_tags"}</parameters>
  <action>set_tags</action>
  <controller>tags</controller>
  <backtrace>TEN 80x24 PAGES OF BACKTRACE ELIDED</backtrace>
  <message>undefined method `valid_level?' for TrustLevel:Class</message>
  <url>https://forum.riking.org/tagger/set_tags?tags=&amp;topic_id=221&amp;_=1420927713082</url>
  <exception-class>NoMethodError</exception-class>
</hash>
```

After, Logster summary line:

    NoMethodError: undefined method `valid_level?' for TrustLevel:Class in tags#set_tags

lib/discourse.rb
```diff
@@ -26,6 +26,29 @@ def self.handle_exception(ex, context = {}, parent_logger = nil)
...
+    }
+
+    env = request.env.dup
+
+    context.each do |key, value|
+      Logster.add_to_env(env, key, value)
+    end
+
+    begin
+      Thread.current[Logster::Logger::LOGSTER_ENV] = env
```
SamSaffron: this is messy api wise, we should clean it

SamSaffron:
much better, good to be rid of that code

PR #3191, opened 2015-02-09
FIX: :sa: Pluralization rules were wrong lang keys

riking:


SamSaffron:
I guess this is ok...

PR #3194, opened 2015-02-10
Support for relative URL and smaller docker images

vgeshel:
This set of changes does two things. They are mostly unrelated but they were essential for us in building our Discourse instance and, I believe, they are generally useful.

* Better support for RAILS_RELATIVE_URL: we have fixed numerous places where it wasn't respected.
Running under a a relative URL (domain.com/forum/...) vs subdomain (forum.domain.com) is very useful from the SEO point of view, see, for instance, here.

* Better support for deployments with external Postgres, Redis and nginx.
The repo contains a Dockerfile which builds an image that includes only Discourse itself. This makes it easier to deploy, e.g., in AWS, using RDS (hosted Postgres) and ElastiCache (hosted Redis). The repo also contains a fig.yml, which helps you use fig to do local development using an equivalent setup (the Vagrant-based setup still works just fine). The fig setup is documented in docs/fig.md.


vgeshel:
This is a resubmission of https://github.com/discourse/discourse/pull/3155. Please see that for some discussion about the changes.

SamSaffron:
This has to be split up into 2 PRs, all the docker stuff should go in discourse_docker

its mixed up here with 2 distinctly different things.

vgeshel:
@SamSaffron: I can split into 2 PRs if necessary, but the docker stuff here is very different from what's in discourse_docker and I put it here for a reason (of course you may not agree).

As far as I understand, the purpose of discourse_docker is to let a user build and run their own, custom discourse docker image which contains postgres, redis, nginx, and their configuration.

The Dockerfile here is quite different: it builds an image that contains just Discourse itself and nothing specific to the particular installation. It could be uploaded to a public registry if desired. It doesn't depend on anything in discourse_docker. It's common for projects that support building docker images of this kind to include a Dockerfile in the same repo as the code. Putting it into a different repo would just make building more complicated.

boushley:
@vgeshel You may want to take another look at discourse_docker. There are samples in there for a setup where only discourse is running inside of the container. (I'm currently using this in a setup with RDS and ElastiCache as the backend.)

The docker container may still contain some pieces of postgres / redis, but the bulk of that work can be avoided by not including the postgres or redis templates in your container configuration. Look at `samples/web_only.yml`. Also, this post may be helpful: http://dev.bizo.com/2014/06/discourse-on-aws.html

vgeshel:
@boushley You're right, of course, with discourse_docker you can write a config that tells Discourse to talk to external services. But this is unnecessarily complicated. With the Dockerfile I'm proposing, you could just have a Discourse image in a public registry and use it like you would any other docker image. Since it's configured at runtime through environment vars, not at build time like with discourse_docker, it can be used with tools like Fig, Kubernetes, and many others.

boushley:
I agree that it would be nice to standardize the build system around Dockerfiles and such eventually. I'm just trying to point out that discourse_docker is the place to do that. Based on the organization as things stand discourse_docker is created for that purpose.

It would certainly be nice to have a Dockerfile based build as this would add things like caching as well.

To overcome the build time config I'm currently mounting a custom config file from the host system into the container to overwrite `/var/www/discourse/config/discourse.conf`. Definitely some hoops to jump through right now, but I think docker_discourse is the place to smooth that out.

vgeshel:
That's fine and is certainly not my call. I'm just trying to point out that there is a lot of stuff in discourse_docker, none of which is necessary to build a simple, functional, runtime-configurable discourse image.

In fact, I'm not even sure where in discourse_docker I would put the Dockerfile that I'm proposing. discourse_docker is about building and running docker images using a specific build/run mechanism. I don't see how a Dockerfile that does not use that mechanism would fit there.

boushley:
That's certainly something that @SamSaffron would have to speak to.

What you're really proposing with this Dockerfile is a different way of building images for distribution. On the meta forums the Discourse team has made it clear that they're only going to support one way of building and distributing Discourse. As such I think we should really consider proposing these as changes to the way things are done there.

Having a runtime configurable setup is a great feature ask, and the best way to do that may be to avoid using pups and the launcher and the template system built out. We should really probably create a thread on the discourse meta forum for dev talking about how to do runtime configurable instances. And propose this as an option.

SamSaffron:
this PR must be split out, the docker stuff is totally unrelated and its
confusing things.

Sure you may be removing launcher and pups but instead you take on fig, so
something has to give here. I am concerned about vetting 2 official
template systems. If we go fig and throw away launcher so be it, but I am
not keeping both around and I am not giving up any of the functionality
that ./launcher has to offer

vgeshel:
Thanks, I will split this out, I'm just trying to understand what to put into the new PRs.

I didn't explain my intentions well in the original comment. First, I did not intend to replace discourse_docker, launcher, pups, etc. I understand that those work well for many people and I don't presume to tell them to change. What I meant to propose was a simpler alternative that works better for us, and, I think, for others who'd like to use Discourse with docker deployment/orchestration tools that expect images to be configurable at runtime using environment vars, links, and volume mappings. Among other things, it allows people to avoid building docker at all: the image can be put in a public registry and you can simply `docker run` it (though you do have to have postgres and redis set up first, so this may not be such a big win). 

In any case, I understand that you don't want to have alternative ways of running docker, so I'm not sure where that leaves this change. If you won't accept this change, that's of course totally fine.

Second, I was even less clear on the fig part: it's mostly just an example. It works and can be useful, we use it for development (but not for production). But it's primarily meant as an example of how to run the standalone Discourse docker image: what environment vars, links, volume mappings it expects and provides. I think the fig.yml shows this more clearly and concisely than text would.

I'll submit two separate PRs shortly... Thanks for your feedback.

PR #3197, opened 2015-02-11
FIX: Don't expire ratelimits at midnight server time

riking:
This fixes the error message returned when you exceed the rate limit by not throwing out the action data when the date rolls over.


SamSaffron:
ok merged following a quick chat with Jeff

PR #3202, opened 2015-02-13
Disallow posting replies to deleted topic

gdpelican:
For replying to deleted topic issue described [here](https://meta.discourse.org/t/replying-to-a-deleted-topic-returns-a-cryptic-message/17035/)

**Before topic is deleted:**
screenshot from 2015-02-14 12 10 26

**After topic is deleted:**
screenshot from 2015-02-14 12 11 03

### DIFF
--- a/app/assets/javascripts/discourse/models/post-stream.js.es6
+++ b/app/assets/javascripts/discourse/models/post-stream.js.es6
@@ -453,13 +453,18 @@ const PostStream = Ember.Object.extend({
         existing = postIdentityMap.get(postId);
 
     if(existing){
-      const url = "/posts/" + postId;
-      Discourse.ajax(url).then(
+      const postUrl = "/posts/" + postId,
+            topicUrl = "/t/" + existing.topic_id;
+      Discourse.ajax(postUrl).then(
         function(p){
           self.storePost(Discourse.Post.create(p));
         },
         function(){
           self.removePosts([existing]);
+        });
+      Discourse.ajax(topicUrl).then(
+        function(t){
+          self.get('topic').updateFromJson(t);
         });
     }
   },
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -155,8 +155,8 @@ export default DiscourseController.extend({
   },
 
   disableSubmit: function() {
-    return this.get('model.loading');
-  }.property('model.loading'),
+    return this.get('model.loading') || this.get('model.topic.deleted');
+  }.property('model.loading', 'model.topic.deleted'),
 
   save: function(force) {
     var composer = this.get('model'),
--- a/app/assets/javascripts/discourse/models/composer.js
+++ b/app/assets/javascripts/discourse/models/composer.js
@@ -120,6 +120,9 @@ Discourse.Composer = Discourse.Model.extend({
     // reply is always required
     if (this.get('missingReplyCharacters') > 0) return true;
 
+    // ensure topic is not deleted
+    if (this.get('topic.deleted')) return true;
+
     if (this.get("privateMessage")) {
       // need at least one user when sending a PM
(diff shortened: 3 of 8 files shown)


riking:
Would be nice if we could get the server to return the right error message, too.

gdpelican:
Added that, although if the client side does its job, you shouldn't ever see the error message.

Screenshot!
screenshot from 2015-02-14 15 45 16

app/assets/javascripts/discourse/models/post-stream.js.es6
```diff
@@ -453,14 +453,19 @@ const PostStream = Ember.Object.extend({
...
+      const postUrl = "/posts/" + postId,
+            topicUrl = "/t/" + existing.topic_id;
+      Discourse.ajax(postUrl).then(
         function(p){
           self.storePost(Discourse.Post.create(p));
         },
         function(){
           self.removePosts([existing]);
         });
+      Discourse.ajax(topicUrl).then(
```
ZogStriP: Do we need that 2nd request?

gdpelican: I think so.

The main 'bug' in all of this is that when a topic is deleted, it calls `PostDestroyer.destroy(topic.posts.first)`
This notifies the client that the post has been deleted, but says nothing about the topic itself. So as far as the composer (and the rest of the clientside app) knows, the topic is just fine and dandy.

I think duplicating that logic on the client side (If a topic's first post is deleted, then the topic is deleted) is a bit fragile (and the Store is so abstracted away in this app that I couldn't find a clean way of doing it anyhow), and that writing a line that's like 'if topic.deleted || topic.posts[0].deleted' in several places is gross.

Happy to be wrong, but I didn't see a way out of calling the server again in some way to update the topic.

gdpelican:
I squashed this up; let me know if there's anything else I can do towards this PR @riking @ZogStriP

coding-horror:
I like this and I think we should fold it in after the 1.2 release tomorrow

coding-horror:
can you review this tomorrow @SamSaffron

SamSaffron:
I am reading through this and just not following how this sequence will work 

- Normal user navigates to topic
- Admin deletes topic
- User attempts to reply to topic

Looking at topic.deleted seems super odd to me as only admins will have access to that, normal users will get 404s 

Instead, I recommend vastly simplifying this. 

- Post creator should return an exception when posting, that is handled correctly via the UI 
- Live refresh on topic should be corrected so when the OP is deleted we display a warning in the composer and disable posting unconditionally
- We should never just remove the OP from the stream it creates a very odd state.

gdpelican:
Hm. This PR works for me in that case, same as described in the screenshots above.

I guess the part that confuses me is that 'deleted' is an admin-only property of the topic? In the ember model it's defined as 
```
  deleted: Em.computed.notEmpty('deleted_at'),
```

Where `deleted_at` is just a column on the topic which is serialized through TopicViewSerializer:
```
attributes_from_topic :deleted_at
```

I think one of the primary causes of this bug is that the current code *does* just remove the OP (the first post from the topic), without updating the topic client side, which creates the odd state we're trying to avoid.

At least as I was able to test (an admin user deleting a topic while a non-admin user is viewing a topic with the composer open), this does disable posting unconditionally, as well as returning an appropriate error message if a submission does somehow get through.

Happy to have missed something, but a bit confused by the feedback.

gdpelican:
Ah, l'retract! It does seem to be admin-only.

The server side still handles the error nicely though. So it's just a matter of finding some other way of notifying the client side of the deletion.

techAPJ:
I spent a day looking into this issue, and I think that handling error client side i.e. showing live warning in composer and disabling submit button is a very complex task to handle cleanly. It requires hacks and complex code to make it happen and I think we should first handle this error properly server side and then improve the client side error as we move forward.

I have used @gdpelican's server side code to handle errors for closed and deleted topic. Here's the PR: #3277

SamSaffron:
closing this post
https://github.com/discourse/discourse/pull/3277

PR #3203, opened 2015-02-14
FIX: respect default locale settings when seed

fantasticfears:
We seed special categories during migration so we have to set the
locale before the migration happens.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,6 +1,10 @@
-# we need to run seed_fu every time we run rake db:migrate
-task 'db:migrate' => 'environment' do
+# we should set the locale before the migration
+task 'set_locale' do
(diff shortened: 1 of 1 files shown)


fantasticfears:
I don't see why the CI failed...Segment fault is weird.

ZogStriP:
I would merge that but like @SamSaffron asked, why is that in a new rake task? Why not simply add that line before the first line of the `db:migrate` rake task?

fantasticfears:
@ZogStriP I don't know what I doesn't express effectively above there...

The situation is if I put in the `db:migrate`, then it won't be set until the migration is done. If so, the categories seeding in the migration will still be set to English locale.

ZogStriP:
I guess it's fine. Let's try it :+1:

PR #3206, opened 2015-02-16
FIX: Default sort grantable badges by name

edgibbs:
Currently has no default sort and was stated as a bug here:
  https://meta.discourse.org/t/badges-not-sorted-in-grant-badge-dropdown/23739
  (Post was recently deleted or made private?)

Not sure if you'd still like to change this to do a default sort by name, but I added a test for the ember controller and made the change to sort by name for grantable badges by default.


SamSaffron:
thanks merged it in

PR #3207, opened 2015-02-16
Update server.zh_CN.yml

chuanjin:
wrong spell 心的 -->  新的


chuanjin:
wrong Chinese words

techAPJ:
Thanks for this, but we do not accept PRs for translations. Instead, you should update the translations here:

https://www.transifex.com/projects/p/discourse-org/language/zh_CN/

PR #3208, opened 2015-02-17
UX: add user invite link on admin interface

techAPJ:
invite


coding-horror:
Love it! :heart_eyes: We get so many support Q's about this. We may need to link it somewhere else as well.

PR #3210, opened 2015-02-17
Fix regression on editing private messages

Flink:
v1.2.0beta9 has introduced a regression in edit of a private topic
(first post). Previously a check for no change in TopicsController was
made but it has been changed without considering that the topic could
be private.

By simply forcing a conversion of `topic.category_id` to integer, the case
where its value is nil is handled correctly as it was previously.


ZogStriP:
Thanks :+1:

Flink:
You’re welcome :)

PR #3211, opened 2015-02-18
Improve auto generated email reply error message

techAPJ:


spec/components/email/receiver_spec.rb
```diff
@@ -350,7 +350,7 @@ def test_parse_body(mail_string)
       let!(:reply_key) { '636ca428858779856c226bb145ef4fad' }
       let!(:email_raw) { fixture_file("emails/auto_reply.eml") }
       it "raises a TopicNotFoundError" do
```
ZogStriP: might want to change that too ;)

techAPJ: Good catch! Just did :)

PR #3217, opened 2015-02-20
FIX: link insertion dialog should not modify the inserted links

techAPJ:


eviltrout:
Seems drastic but it seems that people want what they pasted to just go through. Let's try it out.

PR #3218, opened 2015-02-20
fixing alignment in category dropdown

awesomerobot:
bar badge style text was off-centered


coding-horror:
See how you feel about a similar fix here? https://github.com/discourse/discourse/pull/3220

If that's not good please submit another, sorry, saw these out of order..

PR #3222, opened 2015-02-22
FEATURE: User preference to always include context

lake54:
This adds an option in the user preferences area that allows a user to always include context in notification emails. Currently the amount of context is the amount set by the site administrator (a default of 5), but in the future I can see this being more customisable by the user (perhaps a preset dropdown list or an input box like in the admin panel).

Added a unit test to verify functionality.

### DIFF
--- a/app/assets/javascripts/discourse/models/user.js
+++ b/app/assets/javascripts/discourse/models/user.js
@@ -186,6 +186,7 @@ Discourse.User = Discourse.Model.extend({
                                'email_digests',
                                'email_direct',
                                'email_always',
+                               'email_force_context',
                                'email_private_messages',
                                'dynamic_favicon',
                                'digest_after_days',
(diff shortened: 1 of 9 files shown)


lake54:
Sigh... and that's the reason behind submitting a PR from a branch... I'll remember for next time!

SamSaffron:
I am pretty uneasy with this change, its a very confusing setting to be adding to users. also where do you stop? allow them to specify the amount of context? allow them to specify that context is included regardless of what was previously sent? allow them to specify if context goes before or after.

Its a big change imo and needs to be discussed on meta first. 

Closing for now till we decide the outcome on meta (please link to this PR when discussing this)

lake54:
Posted as requested:

https://meta.discourse.org/t/forcing-email-context-input-requested/25577

Personally, I disagree that it's confusing, but I agree it probably needs more discussion from other users.

PR #3224, opened 2015-02-23
FIX: Don't escape test in topic excerpt for mobile.

edgibbs:
Stopped escaping topic excerpt in the mobile handlebars template to ensure things like quotes don't get escaped.  Bug if from this post:

https://meta.discourse.org/t/double-escaped-single-quotes-in-suggested-posts/24271


SamSaffron:
good catch

qrohlf:
👍 Glad this got fixed. Thanks!

PR #3226, opened 2015-02-23
FEATURE: new site setting contact_url

techAPJ:
If the Contact URL is provided, that will be shown on `/about` page, instead of Contact Email.


eviltrout:
Looks good!

PR #3227, opened 2015-02-23
FIX: Forward to SSO login automatically

sigginet:
Forward to SSO login URL automatically if SSO is enabled and login is required.

Makes it simpler for users to log in automatically.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -389,7 +389,13 @@ class ApplicationController < ActionController::Base
 
       # save original URL in a cookie
       cookies[:destination_url] = request.original_url unless request.original_url =~ /uploads/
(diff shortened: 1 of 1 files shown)


riking:
Travis failure is an unordered TopicQuery result

sigginet:
@riking .. I have nooo clue what that means .. (I started learning RoR this weekend)

It looks like Travis succeeds using Ruby 2.0 but fails in 2.1 due to this TopicQuery result (which I believe I didn't touch..?)
Is it possible that Travis is drinking at work?

riking:
Yeah, heh. I was saying that the failure was unrelated to your changes; it's due to an improperly nondeterminstic test.

SamSaffron:
I will get that test fixed :) thanks for the change

PR #3229, opened 2015-02-24
Api key should be easy to copy

iamntz:
On Windows/Chrome, double clicking on the api key will also select the label, so if you don't pay attention you will copy the _key_ part as well (and also a nice whitespace at the end):

```
Key80575586ae7514facbc68b0e4f2227508d6f595af87831d5936c0d723962bca8
```

This PR should fix this.

### DIFF
--- a/app/assets/javascripts/admin/templates/user_index.hbs
+++ b/app/assets/javascripts/admin/templates/user_index.hbs
@@ -225,7 +225,7 @@
 
     {{#if api_key}}
       <div class='long-value'>
-        {{api_key.key}}
+        <input type="text" value="{{api_key.key}}" onclick="this.select()" style="width:450px">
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/admin/templates/user_index.hbs
```diff
@@ -225,7 +225,7 @@
 
     {{#if api_key}}
       <div class='long-value'>
-        {{api_key.key}}
+        <input type="text" value="{{api_key.key}}" style="width:450px">
```
SamSaffron: this is odd on a few levels, 

1. it should not be editable
2. CSS belongs in SCSS files, not inline.

SamSaffron:
closing for now, needs to be redone cleanly, (readonly text input with css in scss files)

PR #3230, opened 2015-02-24
Avoid gzip "-k" option for wider OS compatibility

jamielinux:
The "-k" option tells gzip to keep the original files intact and is an
alternative to `gzip -c file > file.gz`. It was implemented in 2013:
http://git.savannah.gnu.org/cgit/gzip.git/commit/?id=0192f02

There are a few popular operating systems (ie, Red Hat 7, Debian Wheezy)
whose version of gzip does not have the "-k" option. Compiling assets
breaks on these operating systems. Using "-c" instead ensures that it
works even with older versions of gzip.

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -118,7 +118,7 @@ end
 
 def gzip(path)
   STDERR.puts "gzip #{path}"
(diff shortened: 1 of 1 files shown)


eviltrout:
Thanks that makes a lot of sense!

(I confirmed the CLA was signed)

PR #3232, opened 2015-02-24
FIX: Reenable console output for verbose localization

riking:
It's a little bit useless if the translators can't see what the keys actually are...


SamSaffron:
aha

PR #3234, opened 2015-02-25
Allow adding and removing members of groups by username or id

dansingerman:
This will make the API consistent for adding and removing users.

As discussed here: https://meta.discourse.org/t/discourse-gem-group-add/25668/2.


ZogStriP:
:+1: Thanks

PR #3237, opened 2015-02-25
FIX: Added 3 user event triggers

sigginet:
Created 3 user event triggers, when a user is created, updated or deleted.

Trigger 1:
user_created
 Variable - username
 Variable - email

Trigger 2:
user_updated
 Variable - username
 Variable - email

Trigger 3:
user_deleted
 Variable - username
 Variable - email

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,6 +80,13 @@ class User < ActiveRecord::Base
   after_create :create_user_profile
   after_create :ensure_in_trust_level_group
   after_create :automatic_group_membership
+  after_create do |username, email|
+    DiscourseEvent.trigger(:user_created, username, email)
+  end
+
+  after_update do |username, email|
+    DiscourseEvent.trigger(:user_updated, username, email)
+  end
 
   before_save :update_username_lower
   before_save :ensure_password_is_hashed
@@ -89,7 +96,10 @@ class User < ActiveRecord::Base
   after_save :refresh_avatar
   after_save :badge_grant
 
-  before_destroy do
+  before_destroy do |username, email|
+    # Event hook
+    DiscourseEvent.trigger(:user_deleted, username, email)
+
     # These tables don't have primary keys, so destroying them with activerecord is tricky:
     PostTiming.delete_all(user_id: self.id)
     TopicViewItem.delete_all(user_id: self.id)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,12 +80,12 @@ class User < ActiveRecord::Base
   after_create :create_user_profile
   after_create :ensure_in_trust_level_group
   after_create :automatic_group_membership
-  after_create do |username, email|
-    DiscourseEvent.trigger(:user_created, username, email)
+  after_create do
+    DiscourseEvent.trigger(:user_created, user: self)
   end
 
-  after_update do |username, email|
-    DiscourseEvent.trigger(:user_updated, username, email)
+  after_update do
+    DiscourseEvent.trigger(:user_updated, user: self)
   end
 
   before_save :update_username_lower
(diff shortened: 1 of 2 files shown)


riking:
I was recommending a `after_user_verified` hook, so that plugins wouldn't have to deal with accounts with fake emails.

sigginet:
@riking yeah .. I was looking at adding that .. so here's code that should work for that
  after_save do
    DiscourseEvent.trigger(:user_verified, user: self) if (self.active_changed? && self.active == true)
  end

sigginet:
After discussion from https://meta.discourse.org/t/making-discourse-more-pluggable-by-adding-more-callback-triggers/25579/5?u=sigurdur - the whole user object is now passed with the triggers.

A new trigger was introduced:
`user_verified`

All triggers pass the `user` object, even the `user_deleted` trigger.
The `user_deleted` callbacks are run before the user is deleted so they have the option of canceling the user deletion or access any objects with reference to the user being deleted.

Tests were created but I had a tiny problem with the `perform_account_activation` in the rspec tests but when everything seems to work properly in production..

> NOTE: These tests all work properly except for ".activate_account" > "valid token" > "trigger" > ":user_verified"

> For some reason I can't seem to figure out why UserController#perform_account_activation doesn't change the user model bool "active" to true.
> In production this happens properly, but I can't seem to get it to work in this test...

app/models/user.rb
```diff
@@ -88,8 +95,14 @@ class User < ActiveRecord::Base
   after_save :clear_global_notice_if_needed
   after_save :refresh_avatar
   after_save :badge_grant
+  after_save do
+    DiscourseEvent.trigger(:user_verified, user: self) if (self.active_changed? && self.active == true)
```
ZogStriP: I would have written 

```rails
DiscourseEvent.trigger(:user_verified, user: self) if self.active_changed? && self.active
```

app/models/user.rb
```diff
@@ -88,8 +95,14 @@ class User < ActiveRecord::Base
...
+  end
 
   before_destroy do
+    # Event hook
+    DiscourseEvent.trigger(:user_deleted, user: self)
```
ZogStriP: What happens if we somehow can't `destroy` that user?

app/models/user.rb
sigginet: @ZogStriP I thought I read somewhere that I need to keep the brackets if I had multiple conditions .. but after searching again I can't find that information - and it passed the build test in my dev env. :)

Good point on removing true ..

app/models/user.rb
sigginet: Good question.

Would it be better to move the trigger to after_destroy..?

Or better yet, rename the trigger to before_user_deleted and create another trigger called after_user_deleted?
That way you can play around with the logic if the user was actually deleted or not .. (in case you need to access the user object before `destroy`)

``` ruby
before_destroy do
  DiscourseEvent.trigger(:before_user_deleted, user: self)
  ...
end
after_destroy do
  DiscourseEvent.trigger(:after_user_deleted, user_id: self.id, self.email, self.username, self.name)
end
```

SamSaffron:
I am ok with adding extensibility where needed, but really want to make sure there is a real world use case for this, as opposed to, it would be call if... architecturally I am a bit concerned hooks are being called when there are zero consumers, if there was no cost at all to adding hooks I would be way more comfortable. 

What in particular do you need this hooks for?

Layoric:
+1 for user_created hook, good spot to check for subscriptions on an external system.

Setup is approval required based on valid subscription, this way we can automate approvals based on the web hook call and then calls back into discourse.

It would be could to add user email to the hook request as well. Noticed that the user object is missing the email in this PRs implementation.

sigginet:
@SamSaffron I need these hooks for integrating e-mail automation with Discourse.
Essentially adding users to e-mail lists, then follow up when they get badges.. (like pestering the users until they post the first post - which awards them a badge) or congratulating users when they get a badge (first like or whatever).

ZogStriP:
@sigginet can you rebase? I'll merge it once done.

eviltrout:
Ping on this one -- we're happy to merge with a rebase, otherwise we might have to close it.

hugohernani:
Was the user_created hook ever added?

PR #3239, opened 2015-02-26
FIX: TL3 could retitle restricted topics

riking:


ZogStriP:
No test? :(

SamSaffron:
I think this is legit regardless, but agree we should have a test

PR #3241, opened 2015-02-26
UX: show user name and title on about page

techAPJ:
with names enabled:

screen shot 2015-02-27 at 00 55 11

with names disabled:

screen shot 2015-02-27 at 00 54 37


downeym:
Love this! Thank you!

SamSaffron:
this is a big packet of styling changes for something so small. 

also I do not think we need any conditional logic here, just show full name there regardless and the "title" / "username" / "name" should be styled the same way it is on the posts, black for everything is confusing.

techAPJ:
Updated the "username" / "name" / "title" style to match with posts and removed conditional logic. Screenshot:

screen shot 2015-03-02 at 15 18 05

SamSaffron:
cheers

PR #3242, opened 2015-02-26
Fix CORS issue when avatar url is on S3

kennym:


kennym:
Related discussion: https://meta.discourse.org/t/avatar-upload-not-working

SamSaffron:
thanks, seems good to me.

PR #3243, opened 2015-02-27
Add translation for email-related messages

ulysses4ever:


gschlager:
I don't think this will be accepted. You should add your Translations in Transifex.
https://www.transifex.com/projects/p/discourse-org/language/ru/

ulysses4ever:
@gschlager thank you for pointing to Transifex, I'll try there.

PR #3244, opened 2015-02-27
FEATURE: Adds transliteration of German umlauts in slugs

gschlager:
https://meta.discourse.org/t/please-replace-german-umlauts-in-urls-with-their-respective-digraphs/24489/7

- Moves the already existing transliteration rules into `transliterations.en.yml` (there's no need to translate this for every language). The same goes for the stringex configuration.
- Doesn't calculate the default slug for *zh_CN* and *ja* anymore. It hasn't been used anyway since stringex is used instead.
- Removes a wrong comment from the Russion transliteration file (I hate wrong comments)


riking:
Can you add these for the Vietnamese transliterations? https://meta.discourse.org/t/vietnamese-slug-support-for-better-seo/22422/9?u=riking

gschlager:
OK, I've done that using the values from https://github.com/thangngoc89/discourse-vietnamese-slug
However, I don't think this will be usable until the Vietnamese translations are actually added to Discourse. There are no locale files for vi_VN yet.

gschlager:
Travis error doesn't seem to be related to this.

riking:
> vi_VN

Use just `vi`, as that's what's in plurals.rb.

gschlager:
Done.
BTW: The tests seem to be rather flaky at the moment.

SamSaffron:
@nlalonde something is super odd with our report specs, they always seem to fail erratically

SamSaffron:
I am fine with this change, moving transliteration out of transifex makes sense.

PR #3246, opened 2015-03-02
Remove topic_auto_close_at from PostSerializer

riking:
We should look at extracting some of the cruft from these repsonses.

This isn't used by the client, the client uses `auto_close_at` from the topic `details`.


ZogStriP:
Guess that's fine.

PR #3247, opened 2015-03-03
group manager can issue invitations from restricted topics

jmay:
See https://meta.discourse.org/t/invitations-to-group-restricted-topics/25895

Enables the Invite button on topics inside group-restricted categories, when the user is group manager for *all* groups that have access to that category. The group selector is not displayed on the Invite popup. Group membership is implicitly added to the invitation.


ZogStriP:
Looks fine. Thanks :+1:

PR #3251, opened 2015-03-05
rebuilding the category badge css

awesomerobot:
Rebuilt a lot of the category badge CSS now that we've got two (or more!) category styles. Everything should align better now and be more consistent; didn't touch the HTML structure. 

I think I checked every instance of where categories appear, but will have to :mag:  on meta.


coding-horror:
Yeah we will need to check this once it is deployed in the next 15 mins

coding-horror:
And thanks! This was a thorn in our side

PR #3252, opened 2015-03-05
Don't show blank page for external links with no click track record

riking:
Instead of showing a white screen, give an unstyled interstitial page with a click-through to the destination. This should keep the security aspect of preventing Discourse sites from being open redirects, while still achieving the goal of not landing the user on a blank page when they click a link.

I also cleaned up a dodgy bit of code in click_track where the "restore href" wasn't quite the inverse of the original operation.

### DIFF
--- a/app/assets/javascripts/discourse/lib/click_track.js
+++ b/app/assets/javascripts/discourse/lib/click_track.js
@@ -15,7 +15,7 @@ Discourse.ClickTrack = {
     @param {jQuery.Event} e The click event that occurred
   **/
   trackClick: function(e) {
-    if (Discourse.Utilities.selectedText()!=="") return false;  //cancle click if triggered as part of selection.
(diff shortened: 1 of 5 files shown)


riking:
Build failure is unrelated

coding-horror:
Why do we have to let the user continue? I agree a blank page is not great, but the whole thing is invalid, GIGO, is it not?

https://meta.discourse.org/clicks/track?url=https://www.example.com/&post_id=null&topic_id=null
https://meta.discourse.org/clicks/track?url=https://www.example.com/&post_id=12345&topic_id=26012
https://meta.discourse.org/clicks/track?url=https://www.example.com/

I feel like this solution is overly complex, adding UI and translations.

riking:
The problem is when the link actually is legit. We have a real link on the client side that the server doesn't know about. One cause of this is when you edit a link inside of a `[quote]`.

PR #3253, opened 2015-03-05
FEATURE: All preload data available under /site/

riking:
This will be essential for entirely local Discourse clients.


SamSaffron:
seems fine, thanks

PR #3254, opened 2015-03-05
"New Users" -> "Users" in the About page for consistency

camilleroux:


coding-horror:
We made this "New Users" intentionally to distinguish it from "Active Users".. thanks though

phu9Suop:
I feel that 
from https://meta.discourse.org/about
```
users/Benutzer | 25 | 264 | 1,3 T. | 63,9 T.
-- | -- | -- | -- | --
Aktive users/Benutzer | 482 | 1,2 T. | 2,3 T. | —

```
the intent of users in the first line is not clear: 
There will by mathematically seen always less active users that total users should be. This is irritating.
What is the purpose of line users ? Please add for the sake of clarity

PR #3255, opened 2015-03-05
Documentation: prefix `gem install mailcatcher` with sudo

jamesgecko:
Resolves the gem installation issue described in [this topic](https://meta.discourse.org/t/permission-denied-when-installing-mailcatcher-in-vagrant-box/25999) by installing mailcatcher with superuser permissions.


SamSaffron:
sure, its vagrant only thanks

PR #3257, opened 2015-03-06
re-rebuilding the category badge css

awesomerobot:
This will still require some customization edits for themes using the box style of categories, but it should be minimized a bit. The CSS structure is much stronger and more organized, and font/size changes shouldn't break the layout. 

* Custom widths  (static, e.g. width: 150px)  should be applied to the wrapper (.box)
* Custom padding (variable, e.g. padding: 5px)  should be applied to the innermost span (.badge-category)

:crocodile:  :palm_tree:

### DIFF
--- a/app/assets/stylesheets/common/base/_topic-list.scss
+++ b/app/assets/stylesheets/common/base/_topic-list.scss
@@ -99,10 +99,6 @@
     }
   }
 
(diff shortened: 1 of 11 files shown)


eviltrout:
Great, thanks!

PR #3258, opened 2015-03-07
Added location.present? test to user profile updating.

mcg1969:
A simple fix: when I was pulling in my askbot data I wanted to be able to map over the user-supplied location data. This fix ensures that base.rb sees the location data.


ZogStriP:
Thanks :+1:

PR #3260, opened 2015-03-07
[code gardening] fix symbolic meaning of protected for class methods

ipoval:
assuming that `protected` was used not only with symbolic intention but to hide class_methods which are not part of open API.
proper hide class methods that don't want to be open.

### DIFF
--- a/app/models/application_request.rb
+++ b/app/models/application_request.rb
@@ -79,10 +79,7 @@ class ApplicationRequest < ActiveRecord::Base
     end
   end
 
-  protected
-
   def self.req_id(date,req_type,retries=0)
-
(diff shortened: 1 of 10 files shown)


coding-horror:
I think the consensus was that this is a code style change only PR which we do not accept as a matter of policy.

ipoval:
@coding-horror , understood, but it looks that it's not just code style change, but grammar fix: private doesn't make class methods private

SamSaffron:
No we can not accept this, I am sorry, 

Please see: https://meta.discourse.org/t/discourse-development-contribution-guidelines/3823?u=sam 

There is mountains of help we need! Why not start with the bug list?

PR #3261, opened 2015-03-09
vBulletin user import missing email field

dpi:
Importers is not selecting the email field and is currently generating a random email for all imported users.


dpi:
**For anyone that has already gone through the import process and wants to fix the email address:**

Change the line in vbulletin.rb to include the email field. (See the attached change)

Then modify base.rb, where:

```ruby
        if user_id_from_imported_user_id(import_id)
          users_skipped += 1
```

add the follow code after:

```ruby
          existing_uid = @existing_users[import_id.to_s]
          existing = existing = User.where(id: existing_uid).first
          existing.email = u[:email]
          existing.save!
          puts "Updated email for existing user #{existing.username}"
```

eviltrout:
Thanks!

PR #3268, opened 2015-03-10
FIX: Revert bd61cdf, use allowUncat render option instead

riking:
site.json needs to have all categories a user can see.

This reverts https://github.com/discourse/discourse/commit/bd61cdf21c7348eeaf4246eb91889f2b7f644a1c

@SamSaffron I think that your site had `allow_uncategorized_topics=false` and `suppress_uncategorized_badge=false`.


SamSaffron:
this can not be reverted, we can not send down "uncategorized" to the client if its not allowed, it leads to a myriad of issues.

PR #3270, opened 2015-03-12
FIX: Importers should create duplicate categories if the parent category is different

gschlager:
https://meta.discourse.org/t/phpbb-3-importer-goes-slow/26182/8?u=gerhard

This changes the content of `@categories_lookup` in the base importer from `Category` objects to IDs since the category names aren't needed anymore. The lookup method has been renamed too.
All importers (except askbot which seems to do this differently) were changed to use the category ID instead of name.


nlalonde:
Thanks for this fix! All the importers just got better. :+1:

PR #3271, opened 2015-03-13
Adds a few enhancements to the base importer

gschlager:
- Allows importing of globally pinned topics
- Doesn't restore the original value of a `SiteSetting` anymore if an importer changed its value
- Allows all file extensions during the import
- Adds the ability to import bookmarks


ZogStriP:
:+1:

PR #3273, opened 2015-03-15
Fix gem load order for SFM2 importer

Elberet:
See https://meta.discourse.org/t/migrating-from-mybb/25563/8?u=elberet


ZogStriP:
Thanks

PR #3274, opened 2015-03-15
Fix broken mysql2 load order in mybb import script

dandv:
Similar to 1cc7db23707a70889dd135c0cde32ff1850013e5 and [thanks to](https://meta.discourse.org/t/migrating-from-mybb/25563/8?u=dandv) @Elberet.


ZogStriP:
Thanks

PR #3275, opened 2015-03-15
Add support for MyBB table prefix

dandv:
Add support for [table prefix](http://docs.mybb.com/1.8/install/#table-prefix), defaulting to `mybb_`.

Includes the mysql2 load order fix from PR #3274.


ZogStriP:
Looks fine :+1:

PR #3276, opened 2015-03-15
Don't scare the user re. Sidekiq if they've just migrated

dandv:
I've just migrated and have [a ton of queued jobs](https://meta.discourse.org/t/dont-trigger-digests-on-migration/26345), which I believe is normal.


ZogStriP:
I don't think this should be in the default message. A migration is a rare event and only done once. No need to burder all other users with that information.

PR #3277, opened 2015-03-16
FIX: return proper error message when replying to a closed or deleted topic

techAPJ:
Better error message when posting on closed/deleted topic.

screen shot 2015-03-16 at 00 43 22


SamSaffron:
This is better, we should show a warning right away if message bus tells us that post #1 is deleted ..

PR #3279, opened 2015-03-18
Fix gem load order for all importers

gschlager:
This fixes the gem load order for the rest of the importers.
https://meta.discourse.org/t/migrating-from-mybb/25563/8


ZogStriP:
Thanks :+1:

PR #3280, opened 2015-03-18
Imported users shouldn't get an email immediately after the import

gschlager:
This sets the `last_emailed_at` for imported users to now which postpones the sending of digest emails for the configured amount of time. Therefore, the first digest email gets sent after 1 week (the default value of `default_digest_email_frequency`).
As suggested in https://meta.discourse.org/t/dont-trigger-digests-on-migration/26345/7


ZogStriP:
Thanks :+1:

PR #3281, opened 2015-03-19
update zh locale

exherb:
add missing zh_CN&zh_TW locale


gschlager:
You should add those translations in Transifex: https://www.transifex.com/projects/p/discourse-org/

exherb:
ok

PR #3282, opened 2015-03-19
[WIP] FEATURE: Add support for two factor authentication.

fantasticfears:
https://meta.discourse.org/t/two-factor-local-login-option/12597

The functionality is there. Before I continue, I'd like to receive some advice from team. I am going to write the process in the topic above.

Something should be done:

- [ ] session_controller_spec
- [ ] refine the setting text

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -148,7 +148,8 @@ gem 'oj'
 gem 'pg'
 gem 'pry-rails', require: false
 gem 'rake'
-
+gem 'rotp'
+gem 'rqrcode'
 
 gem 'rest-client'
 gem 'rinku'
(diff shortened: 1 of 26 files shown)


siepkes:
Cool feature! We are currently using two factor authentication with Discourse and OpenAM (oauth2). One advantage of offloading the more complex authentication to an AM solution like OpenAM (or Gluu, or whatever AM solution you want to use) is that it allows for vastly more flexibility then Discourse probably ever will. So while I think its really cool what you've made I would just like to point out that it might be wise for the devs to think about how much security complexity they want to pull in to Discourse and where to draw the line and off-load it to other solutions.

SamSaffron:
closing this for now... its gone quite stale, also two factor does feel like a plugin.

PR #3283, opened 2015-03-19
single quote password in backup command

eriko:
This protects against characters like '&' in passwords. Sometimes you are assigned passwords by idiots or are and idiot that uses pronounceable passwords.  Anyways this small change protects against ruby's shell interpreter from background the pg_dump command before it has really started.


ZogStriP:
> Sometimes you are assigned passwords by idiots or are and idiot that uses pronounceable passwords.

:sob:

PR #3284, opened 2015-03-19
FEATURE: new site setting min_first_post_length

techAPJ:
[Feature request here](https://meta.discourse.org/t/how-to-set-minimum-topic-body-length/26123/17?u=techapj)

### DIFF
--- a/app/assets/javascripts/discourse/models/composer.js.es6
+++ b/app/assets/javascripts/discourse/models/composer.js.es6
@@ -40,6 +40,8 @@ const Composer = Discourse.Model.extend({
     return this.get('creatingPrivateMessage') || this.get('topic.archetype') === 'private_message';
   }.property('creatingPrivateMessage', 'topic'),
 
+  topicFirstPost: Em.computed.or('creatingTopic', 'editingFirstPost'),
+
   editingPost: Em.computed.equal('action', EDIT),
   replyingToTopic: Em.computed.equal('action', REPLY),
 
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -40,6 +40,10 @@ const Composer = Discourse.Model.extend({
     return this.get('creatingPrivateMessage') || this.get('topic.archetype') === 'private_message';
   }.property('creatingPrivateMessage', 'topic'),
 
+  topicFirstPost: function(){
```
ZogStriP: `Em.computed.or("creatingTopic", "editingFirstPost")` ;)

techAPJ: Done :)

lib/validators/post_validator.rb
```diff
@@ -25,7 +25,17 @@ def presence(post)
   end
 
   def stripped_length(post)
-    range = post.topic.try(:private_message?) ? SiteSetting.private_message_post_length : SiteSetting.post_length
+    range = if post.topic.try(:private_message?)
+      # private message
+      SiteSetting.private_message_post_length
+    elsif ( post.is_first_post? || (post.topic.present? && post.topic.posts_count == 0 && !post.topic.try(:private_message?)) )
```
ZogStriP: I don't think you need the `!post.topic.try(:private_message?)` test here since you've already rulled that's  not a PM

techAPJ: Correct! updated the code..

ZogStriP:
:+1:

PR #3285, opened 2015-03-19
user card animation when there is an active user card

fantasticfears:
https://meta.discourse.org/t/animations-in-general-and-animations-for-user-cards/25843/20


ZogStriP:
Looks like it needs a rebase.

fantasticfears:
Rebased and:

- FIX: prevent redirecting to user page when close it
- Don't cover user small avatar by adding more top space

fantasticfears:
Fixed the spec, ready to review.

PR #3286, opened 2015-03-20
FEATURE: load english translation by default as migration needs

fantasticfears:
- and move two get started docs into translation file

https://meta.discourse.org/t/move-documentations-to-transifex/26542


SamSaffron:
very uneasy about moving huge chunks of text into the yaml file, its already huge and why load it all into memory

gschlager:
I have to agree with Sam. I guess it would be better to just add the Markdown files to Transifex.

fantasticfears:
Better to merge them into Transifex as Sam suggest.

https://meta.discourse.org/t/move-documentations-to-transifex/26542/3?u=fantasticfears

PR #3287, opened 2015-03-20
FIX: Discourse.Utilities.avatarImg test on Firefox

savef:
This qUnit test was failing in Firefox 36 with "setting a property that has only a getter" because devicePixelRatio was being changed directly. The fix uses defineProperty to redefine devicePixelRatio instead.


eviltrout:
Thanks! Looks good to me.

PR #3288, opened 2015-03-20
Improve the italic and bold buttons on multiline selections.

savef:
This misuses the system that Pagedown has set up somewhat, but it has to in order to affect the content inside the selection. The end behaviour seems to me to be non-buggy and much more obvious to an end user.

Essentially, instead of performing the operation to surround characters on one line, it splits the selection into lines the loops the operation. Usually a single "chunk" is returned with some replacement text and markdown to insert before and after. Because we're replacing text inside the selections the before and after are left blank and the selection contains all the changed markdown.

Related meta thread: [link](https://meta.discourse.org/t/wysiwyg-buttons-on-multiline-text-bug/25667)

Please try the current behaviour of the WYSIWYG italic/bold buttons compared to after this commit, I think the new behaviour is clearly an improvement.

### DIFF
--- a/app/assets/javascripts/discourse/lib/Markdown.Editor.js
+++ b/app/assets/javascripts/discourse/lib/Markdown.Editor.js
@@ -1542,57 +1542,75 @@
     };
 
     commandProto.doBold = function (chunk, postProcessing) {
-        return this.doBorI(chunk, postProcessing, 2, this.getString("boldexample"));
+        return this.doSurroundLines(chunk, postProcessing, 2, this.getString("boldexample"));
     };
 
     commandProto.doItalic = function (chunk, postProcessing) {
-        return this.doBorI(chunk, postProcessing, 1, this.getString("italicexample"));
+        return this.doSurroundLines(chunk, postProcessing, 1, this.getString("italicexample"));
     };
 
-    // chunk: The selected region that will be enclosed with */**
+    commandProto.doSurroundLines = function(realChunk, postProcessing, nStars, fallbackText) {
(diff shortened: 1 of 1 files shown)


savef:
I've noticed at least 2 instances of me using the wrong spacing (I'm used to 2 space tabs, but this document uses 4). I take it I should fix mine to match the 4 spaces in the rest of the document, rather than converting it all to 2 spaces (just checking because [CONTRIBUTING](https://github.com/discourse/discourse/blob/master/CONTRIBUTING.md) says that should be the convention)?

savef:
I don't know enough about the tests to say why it failed, but I'm pretty sure it's nothing to do with my changes. Any ideas?

eviltrout:
Looks like the failure is unrelated, we can ignore that for now.

eviltrout:
To be honest, I am a little hesitant to accept this patch, since it modifies a library we use directly instead of upstream, and as @coding-horror mentioned we plan on rewriting the editor window at some point.

Is there any chance you could get this merged into pagedown first? I'd be much more amicable about it then? It seems to be more than a few lines of code.

SamSaffron:
@eviltrout  we are so diverged off pagedown at the moment that its impractical to get merged back, I nuked big chunks of code and added plenty hacks. 

I am open to accepting if this works but too want to just get rid of all this complex code and replace with something small that fits us better.

savef:
So what's the next step? :smile_cat:

eviltrout:
Let's merge it then! Thanks.

PR #3290, opened 2015-03-22
FEATURE: allow to disable stringex for symbolizing the topic and category title

fantasticfears:
I received a lot of request to remove the symbolized the slug for Chinese. They think the slug generated by stringex is not good enough but they are totally fine with the default `topic`.

As we can already set the category slug manually, it's cool to make the slug for title optional. This PR is a compromising effect since disabling slugger(`Slug.for`) may lead to other components collapsed.

I found several place use `Slug.for` which I can't understand. So I would postpone to refactor them until I know why it's there.

If the team are open to this change, I will fix the spec


SamSaffron:
no sure why you would want to disable stringex and still leave default slug algorithm around, seems odd. if you disabled I assume you simply don't want slugs

PR #3295, opened 2015-03-24
Swapped order of checks

RaceProUK:
It was pointed out by Monarch that they were the wrong way round :blush:


SamSaffron:
yeah, there is that :)

PR #3296, opened 2015-03-25
FIX: delete duplicate invites

techAPJ:
[Issue reported here](https://meta.discourse.org/t/cleanup-of-duplicate-invites/26673?u=techapj)


eviltrout:
Looks good!

PR #3299, opened 2015-03-26
FIX: Stagger daily backups by up to 10 minutes

riking:
Requested by @Supermathie :)


coding-horror:
:+1: I always wondered why we didn't have more randomization around timing in the various low-pri events

riking:
This solution doesn't scale out to every daily scheduled job - better would be to stagger `daily at:` by site. For just backups, this is more than sufficient.

PR #3304, opened 2015-03-28
Desktop notifications!

riking:
Wasn't actually too hard, but there are several rough points:

 - ~~I put the code in a controller. That feels extremely weird.~~ Fixed, now lives in lib/
 - ~~The code to get a URL from a notification is duplicated.~~ Fixed
 - ~~Currently, all tabs generate the notifications, and you get a replace animation for every tab you have open past 1. We need a protocol using localStorage and `window.on('unload')`.~~ Fixed

Other than that... it works :D


riking:
Also to-do: Notification sound is a must

balupton:
This is fantastic. What's left to do?

app/assets/javascripts/discourse/controllers/notifications.js.es6
```diff
@@ -1,6 +1,7 @@
-export default Ember.ArrayController.extend({
+const NotificationsController = Ember.ArrayController.extend({
```
eviltrout: Why bother making it a constant if you are not going to use it? I prefer the export directly in this case unless you need to reference the class somehow within this file.

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
+
+let primaryTab;
+let liveEnabled;
```
eviltrout: I would prefer it if the boolean vars are initialized to `false` rather than `null`. I know `null` is basically false but I think it helps readability when opening the file as to the type of the vars.

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
...
+    try {
+      localStorage.getItem(focusTrackerKey);
+    } catch (e) {
+      Em.Logger.info('Discourse desktop notifications are disabled - localStorage denied.');
+      return false;
+    }
+    liveEnabled = true;
+    Em.Logger.info('Discourse desktop notifications are enabled.');
+    return true;
+  }).then(function(c) {
```
eviltrout: Why do you have two `then` methods attached here? It looks like it could be together in one?

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
...
+        Em.Logger.error(e);
+      }
+    }
+  }).catch(function() {
+    liveEnabled = false;
+    //Em.Logger.debug('Discourse desktop notifications are disabled - permission denied.');
+  });
+}
+
+function init2(container) {
```
eviltrout: I am not a fan of `init2` as a function name. Perhaps `setupNotifications` or something more descriptive like that.

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
...
+  }
+  if (markAllSeen) {
+    Discourse.ajax("/notifications.json?silent=true").then(function(result) {
+      updateSeenNotificationDatesFrom(result);
+    });
+  }
+
+  notificationTagName = "discourse-notification-popup-" + Discourse.SiteSettings.title;
+
+  const messageBus = container.lookup('message-bus:main');
```
eviltrout: If the only thing that the container does is look up the `messageBus`, I think it's cleaner to just pass the `messageBus` instance into the `init` isn't it? Why give this code the ability to look up anything when it just needs one thing?

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
...
+  mbClientId = messageBus.clientId;
+
+  window.addEventListener("storage", function(e) {
+    // note: This event only fires when other tabs setItem()
+    const key = e.key;
+    if (key !== focusTrackerKey) {
+      return true;
+    }
+    if (primaryTab) {
```
eviltrout: primaryTab = !primaryTab

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
```diff
@@ -0,0 +1,245 @@
...
+      const notificationIcon = Discourse.SiteSettings.logo_small_url || Discourse.SiteSettings.logo_url;
+
+      // This shows the notification!
+      const notification = new Notification(notificationTitle, {
+        body: notificationBody,
+        icon: notificationIcon,
+        tag: notificationTagName
+      });
+
+      // if (enableSound) {
```
eviltrout: Probably can remove this comment

app/assets/javascripts/discourse/controllers/notifications.js.es6
riking: Because I did use it at one point, then removed that :P

app/assets/javascripts/discourse/lib/desktop-notifications.js.es6
riking: Nope... `primaryTab = false;`

SamSaffron:
what is left here ? cc @eviltrout

eviltrout:
I'm going to cautiously accept this now :)

PR #3305, opened 2015-03-28
persian language

Qasem-h:
Please put in discourse persian language
http://www.mediafire.com/download/bzkzkaztdwrza1y/Persian+fa_IR+lang.zip


Qasem-h:
update pull request

PR #3313, opened 2015-03-31
FEATURE: invite existing user to a topic

techAPJ:
Invite "exiting member" to a topic (only available for public topics).

One change I did in this PR is that group autocomplete field will only be visible when inviting user (via email) to a forum or private topic. That means when inviting existing/new user to a public topic, groups can't be added.


app/assets/javascripts/discourse/controllers/invite.js.es6
```diff
@@ -54,6 +54,24 @@ export default ObjectController.extend(ModalFunctionality, {
...
+    return this.get('invitingToTopic') && !this.get('isPrivateTopic');
+  }.property('invitingToTopic', 'isPrivateTopic'),
+
+  /**
+    Show Groups? (add invited user to private group)
+
+    @property showGroups
+  **/
+  showGroups: function() {
+    return this.get('isAdmin') && !this.get('allowExistingMembers');
```
SamSaffron: this makes no sense, you are stopping admins from inviting into existing groups, this should only happen AFTER they select a user that is not an email

app/assets/javascripts/discourse/templates/modal/invite.hbs
```diff
@@ -11,9 +11,13 @@
   {{else}}
 
     <label>{{inviteInstructions}}</label>
-    {{text-field value=email placeholderKey="topic.invite_reply.email_placeholder"}}
+    {{#if allowExistingMembers}}
```
SamSaffron: all this code feels very duplicate, we have it in invite into PM we need to use a single temple here to do both and a single controller

app/mailers/user_notifications.rb
```diff
@@ -188,11 +193,12 @@ def notification_email(user, opts)
     use_site_subject = opts[:use_site_subject]
     add_re_to_subject = opts[:add_re_to_subject]
     show_category_in_subject = opts[:show_category_in_subject]
+    original_username = @notification.data_hash[:original_username] || @notification.data_hash[:display_username]
```
SamSaffron: this is a bit weird, why not pass in original username?

techAPJ: Because in case of "Invite user to a Message", [we are passing](https://github.com/discourse/discourse/blob/master/app/models/topic.rb#L549) `display_username` to notifications data. So with this current code, username is not getting shown in "Invited to PM" notification emails.

Another way to handle this case would be passing `original_username` as well to notifications data.

techAPJ:
@SamSaffron: Updated the code as per your feedback. Please review.

techAPJ:
Any improvement needed here?

app/assets/javascripts/discourse/controllers/invite.js.es6
```diff
@@ -19,12 +19,12 @@ export default ObjectController.extend(ModalFunctionality, {
...
     if (this.get('saving')) return true;
-    if (this.blank('email')) return true;
-    if (!Discourse.Utilities.emailValid(this.get('email'))) return true;
+    if (this.blank('emailOrUsername')) return true;
+    if ( !this.get('invitingToTopic') && !Discourse.Utilities.emailValid(this.get('emailOrUsername')) ) return true;
     if (this.get('model.details.can_invite_to')) return false;
     if (this.get('isPrivateTopic') && this.blank('groupNames')) return true;
     return false;
-  }.property('email', 'isPrivateTopic', 'groupNames', 'saving'),
+  }.property('emailOrUsername', 'isPrivateTopic', 'groupNames', 'saving'),
```
ZogStriP: forgot to observe `invitingToTopic`

app/assets/javascripts/discourse/controllers/invite.js.es6
```diff
@@ -54,17 +54,44 @@ export default ObjectController.extend(ModalFunctionality, {
...
+  }.property('invitingToTopic', 'isPrivateTopic'),
+
+  /**
+    Show Groups? (add invited user to private group)
+
+    @property showGroups
+  **/
+  showGroups: function() {
+    return this.get('isAdmin') && (Discourse.Utilities.emailValid(this.get('emailOrUsername')) || this.get('isPrivateTopic') || !this.get('invitingToTopic'));
+  }.property('isAdmin', 'emailOrUsername'),
```
ZogStriP: shouldn't you observe `isPrivateTopic` and `invitingToTopic` too?

app/assets/javascripts/discourse/views/invite.js.es6
```diff
@@ -4,9 +4,13 @@ export default ModalBodyView.extend({
...
-           I18n.t('user.invited.create');
-  }.property('controller.invitingToTopic')
+    if (this.get('controller.isMessage')) {
+      return I18n.t('topic.invite_private.title');
+    } else if (this.get('controller.invitingToTopic')) {
+      return I18n.t('topic.invite_reply.title');
+    } else {
+      return I18n.t('user.invited.create');
+    }
+  }.property('controller.invitingToTopic', 'controller.isMessage')
```
ZogStriP: you can also observe `controller.{invitingToTopic,isMessage}` ;)

app/assets/javascripts/discourse/controllers/invite.js.es6
techAPJ: Yes, corrected now!

app/assets/javascripts/discourse/controllers/invite.js.es6
techAPJ: Oops, done!

app/assets/javascripts/discourse/views/invite.js.es6
techAPJ: Thanks for the tip! Implemented :)

ZogStriP:
Thanks :+1:

PR #3322, opened 2015-04-02
Checking for pending users every two hours

thomasschmit:
Checking for pending users every nine hours seems not being enough.


coding-horror:
I think this should be a site setting variable, I am uncomfortable with hard coding it to every 2 hours.

PR #3324, opened 2015-04-02
UX: cleaner trending up/down stats colors

wil93:
Before:
before

After:
after

The reason for this is that if, say, I have `1000 (down arrow)` all colored in red, then I'm led to believe that that stat has *decreased by 1000* (which is not necessarily the case, maybe it decreased by 1).


coding-horror:
Nice change! If you can get the CLA signed I think this is a good one to pull in!

wil93:
I signed it after @discoursebot commented, maybe he hasn't noticed yet :smile:

PR #3326, opened 2015-04-03
[WIP] FEATURE: per-language defaults for site settings

fantasticfears:
have to use `refresh!` twice to get it work for two circumstances:
- Rails booting
- `default_locale` changed. 

Where should we put the overrides? File approach may be not so helpful since the overrides is not big.

https://meta.discourse.org/t/setting-default-site-settings-and-scss-based-on-locale/25217

### DIFF
--- a/config/default_settings/zh_CN.yml
+++ b/config/default_settings/zh_CN.yml
@@ -1,0 +1,11 @@
+posting:
+  min_post_length: 8
+  min_first_post_length: 8
+  min_private_message_post_length: 4
+  min_topic_title_length: 5
+  uncategorized_description: 不需要分类的主题，或者不适用于任何分类的主题。
+  min_title_similar_length: 5
+  min_body_similar_length: 6
+
+uncategorized:
+  min_search_term_length: 2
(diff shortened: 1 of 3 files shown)


config/default_settings/zh_CN.yml
```diff
@@ -0,0 +1,11 @@
+posting:
+  min_post_length: 8
+  min_first_post_length: 8
+  min_private_message_post_length: 4
+  min_topic_title_length: 5
+  uncategorized_description: 不需要分类的主题，或者不适用于任何分类的主题。
```
gschlager: Why is this even in the site_settings? Can't we move that into `server.yml` and allow proper translation in Transifex? Wait, there is already a (different) [uncategorized_description in `server.yml`](https://github.com/discourse/discourse/blob/master/config/site_settings.yml#L352).

fantasticfears: It's a site setting. Basically, the answer is no. Uncategorized category uses a different approach. Since it's a site setting, it can be changed by admin at any time and they also should change this site setting.

PR #3331, opened 2015-04-03
Allowing feed polling with UTF-8 titles

thomasschmit:
This pull request prevents the following error when the pooled feed contains special characters:
```
Job exception: incompatible character encodings: ASCII-8BIT and UTF-8
```

Source: https://meta.discourse.org/t/embed-poll-feed-fail-with-utf8-title/25776


ZogStriP:
Thanks :

PR #3332, opened 2015-04-04
UX: add reply-to glyph in composer

techAPJ:


coding-horror:
I don't think we should hard-code the color here, should be set via CSS

techAPJ:
Ok, added the CSS.

PR #3338, opened 2015-04-06
add rtl support for persian language

Qasem-h:


SamSaffron:
sure seems fine to me, thanks!

PR #3340, opened 2015-04-07
Updating Persian translate

Qasem-h:
1: Some Update translate
2:It adds a pluralization rule to fa_IR.js.erb. The rule always returns 'other'. This is because the fa_IR translation files do not use the 'one' key for pluralization.
note : Persian does not distinguish between the singular and plural forms
of nouns in the same way as English does. For numbered elements,
always return the 'other' key from the translation file.
Q :@scossar

### DIFF
--- a/config/locales/client.fa_IR.yml
+++ b/config/locales/client.fa_IR.yml
@@ -13,7 +13,7 @@
 # Read more here: https://meta.discourse.org/t/contribute-a-translation-to-discourse/14882
 # Persian languages by http://discours.ir
 # To validate this YAML file after you change it, please paste it into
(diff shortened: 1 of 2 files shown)


ZogStriP:
Now that you've pushed the files in the repository, all the translations must be done through [transiflex](https://www.transifex.com/projects/p/discourse-org/translate/#fa_IR).

Qasem-h:
i think we need this file ! We have some error in translate file 
2:It adds a pluralization rule to fa_IR.js.erb. The rule always returns 'other'. This is because the fa_IR translation files do not use the 'one' key for pluralization

PR #3344, opened 2015-04-07
Fix italics/bold WYSIWYG bug with nothing highlighted.

savef:
This bug was reported here: https://meta.discourse.org/t/ctrl-b-selects-asterisks/27215 - It was something I broke whilst writing PR3288.

The fix checks if it is a multiline selection, if it is not (which includes blank selections) it will leave the asterisks unhighlighted.

Also fix a bug where asterisks would not be stripped if there was whitespace at the beginning of a line in a multiline selection.

Also fix styling issues I missed last time so that it matches the rest of the document. Specifically, 4 character tabs and spaces after "if"s.


ZogStriP:
Thanks :+1:

PR #3345, opened 2015-04-07
FIX: Pin icon in RTL layout was wrong

riking:


app/assets/javascripts/discourse/templates/topic-status.raw.hbs
```diff
@@ -1,5 +1,6 @@
 {{~#if view.renderDiv ~}}
```
SamSaffron: if view.renderDiv ... then ... render span ... seems very odd to me.

riking: Addressed by renaming to `addWrapper`

SamSaffron:
I do not think we need this now that I added the auto rtl stuff

PR #3347, opened 2015-04-07
[update] persian language

Qasem-h:
100% Ended client.fa_IR 
and 
some update server.fa_IR.yml


ZogStriP:
Like I said earlier, we won't accept PR that are updating existing translation files since they will be overriden whenever we fetch the translations from transiflex. Please use transiflex for all translation work from now on.

Qasem-h:
Im end the clinet file,

PR #3357, opened 2015-04-09
Make avatars in oneboxes round, too

Elberet:
Also adds a subtle shaded border around the avatar to smooth over the perceived greater gap between avatar and content.

See https://meta.discourse.org/t/round-avatars-in-topic-list/27098/106?u=elberet


coding-horror:
Thanks but per discussion on meta I think Oneboxes are ok as is

PR #3360, opened 2015-04-09
Update Translate

Qasem-h:
This is 100% translated and reviewed fa_IR file
please add this to core file because we have many translate missing


ZogStriP:
This is the **last** time I will say this: we do **not** accept pull requests updating translation files. **Everything** has to go through transifex otherwise they'll get overriden next time we pull the translations from transifex.

PR #3362, opened 2015-04-11
Please add info about the vagrant box architecture

rubo77:
if it is 32 bit or 64bit

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,6 +5,7 @@
 Vagrant.configure("2") do |config|
   config.vm.box= "discourse/discourse-0.9.9.15.box"
   config.vm.box_url = "https://vagrantcloud.com/discourse/discourse-0.9.9.15.box"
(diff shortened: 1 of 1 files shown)


rubo77:
It seems like 
https://vagrantcloud.com/discourse/discourse-0.9.9.15.box
only works for 64 bit. Can you provide an alternate box for 32bit?

rubo77:
It turned out, that the box is 32 bit already. The only problem was, that I had to enable Virtualization in my bios settings.

Had to restart my Computer to access bios and set it there, so this hint should be set in the vagrant file there somehow.

coding-horror:
Ok, you signed the CLA?

rubo77:
I signed. But this pull-request is just a kind of issue request..

The comment I added doesn't say anything yet.

Please close this PR and add a reasonable comment there instead

thx.
Rubo77

PR #3365, opened 2015-04-12
initialize discourse in vagrantfile

rubo77:
Those lines from the documentation Vagrant.md should be started right away in `Vagrantfile`

    cd /vagrant
    bundle install
    bundle exec rake db:migrate

Starting Rails:

    bundle exec rails s


rubo77:
There is a semicolon missing after `set -x;`

I am not sure if this patch is appreciated, if so, should I correct this PR with --amend? Or start a new PR?

SamSaffron:
pretty sure @eviltrout  updated our vagrant, can you confirm it is good now?

PR #3368, opened 2015-04-12
Multidirectional mixins

Qasem-h:
Adds mixins to replace all right/left css rules - float, text-align, left and right margin and padding, the margin and padding shorthand, the left and right positional rules, and border-left/border-right.

With these mixins ltr and rtl stylesheets can be generated with a minimum of code duplication. The rtl.scss file has been reduced to 21 lines. Everything seems to be working as it should.

Tests for the mixins are included at stylesheets/common/foundation/multi_directional_mixins_tests.scss. The tests aren't automated but you can run them at http://sassmeister.com/

Right now the rtl stylesheets (mobile, admin, and desktop) are hardcoded into the link tags in _discourse_stylesheets.html.erb .There needs to be a conditional function similar to the current mobile_view? one to load the correct stylesheets. The stylesheets also need to be added to the config.asset.paths. 

There are probably better ways to wire the parts up than I have done.

I understand if you don't use this - it is making a lot of changes to the core scss files. (The changes were made with search and replace on a text editor. It took about an hour.)

If you do use this idea, instead of writing float: $value; people will now have to write @include float($value);.

I'm happy to make any changes you might want.
by @scossar  and me
THis is a Online Demo :http://104.236.244.77/
it worK best


Qasem-h:
THis is a Online Demo :http://104.236.244.77/
 it worK best

Moter8:
You commit *every single file* individually?

Qasem-h:
@Moter8 i think we changed over 10000 line code ! This is normal.

PR #3373, opened 2015-04-14
Add blank alt attribute to user card avatars.

savef:
This was giving an ugly border to avatars in the user card as the full size version loaded in Firefox. 

See [here](https://meta.discourse.org/t/small-ui-bug-with-user-cards-in-firefox-on-linux/27408/2) and [here](https://meta.discourse.org/t/small-ui-bug-with-user-cards-in-firefox-on-linux/27408/3).

(It's the *broken image* border in Fx, the avatar's border-radius was making it appear as just four small lines.)


savef:
Oops, that took longer than expected. Should I squash it into one commit?

ZogStriP:
Yeah definitiely squash into 1 commit please :)

savef:
Done.

ZogStriP:
Awesome, thanks :+1:

PR #3375, opened 2015-04-14
FEATURE: invite existing users to private topic

techAPJ:
Requested [here](https://meta.discourse.org/t/user-invitation-poking-a-user-to-a-topic-is-now-live/27180/53?u=techapj)


techAPJ:
@eviltrout code updated as per your feedback.

eviltrout:
Thanks for the hard work on this one!

PR #3377, opened 2015-04-15
Auto rtl

scossar:
Allows for automatic right-to-left conversion of stylesheets based on the default locale of the forum.

This is done by adding the R2 gem to the Gemfile and updating the conditional check for rtl sites in `DiscourseSassCompiler`. Previously it checked for the state of the `GlobalSetting.rtl_css` value. That defaulted to false and did not work well when set to true. I have set it to check that:

- `SiteSettings.allow_user_locale` is false
- the locale is one of 'he ar fa_IR' (those locales are hardcoded into the method - the only place I could find them defined was in RTL.rb

For the R2 gem to work I have had to add `direction: ltr` to the html element. This is so it can be reversed by r2. I have also removed all previous styles from `rtl.scss` - these were conflicting with the rtl conversion.

To change a site's direction in production requires changing the locale and running `./launcher rebuild app`.

To change a site's direction in development requires changing the locale, stopping the server, deleting the contents of `public/uploads/stylesheet-cache` and rebooting the server.

In development it is very slow - it is processing multiple requests.

It does not style the admin section of the site. I have produced rtl styles for admin, but including them requires a bit of fiddling around with how they are loaded.

It would be great to have the option of allowing users to select their own local and switching style sheets as appropriate for the given locale. That is very doable, but it seems like doing it the way I have is a good place to start.


SamSaffron:
Lets merge this in, its much better than carrying a huge RTL extra file around :+1:

PR #3379, opened 2015-04-15
indicate that lists can be pasted if pipe-delimited

betson:
In the configuration for a list-setting component, the Select2 component is initialized with a pipe ("|") as the separator. (app/assets/javascripts/admin/components/list-setting.js.es6)

This should be communicated to the user in the event they want to paste a list of domains for the blacklist/whitelist.

Current display:
current_display


coding-horror:
seems good, thanks!

PR #3388, opened 2015-04-22
Adding ViewAction into email notifications

Mooash:
See: https://meta.discourse.org/t/integrate-viewactions-one-click-actions-schemas-into-discourse-emails/27056

This is the initial pull request. I hope to add a further `ViewAction` for signup.

### DIFF
--- a/app/views/email/notification.html.erb
+++ b/app/views/email/notification.html.erb
@@ -23,3 +23,9 @@
 %{unsubscribe_link}
 </div>
 </div>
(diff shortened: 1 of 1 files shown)


coding-horror:
also shouldn't the string be an i18n call?

Mooash:
I've tested it and it works fine in my vagrant instance. I tested it against nearly every single email alter I could try and send.

As for the i18n call, you're likely correct but I'll be honest, I have no idea how to change it to that.

eviltrout:
Looks good now, let's try it out!

PR #3390, opened 2015-04-23
Display warning that Vagrant image is out of date

wjordan:
Since the Vagrant image is currently out of date, I suggest adding this note in the documentation (at least until the image is updated and the setup process works as expected) so others don't needlessly struggle with a broken installation as I did.


eviltrout:
Thanks but the correct solutionw as for me to fix up Vagrant, so I did that!

https://github.com/discourse/discourse/commit/17dc8b8e4f2bf7b984ae279f7a0f76a810ac8dd8

wjordan:
Even better, the Vagrant setup works perfectly now. Thank you so much!

PR #3393, opened 2015-04-23
FIX: validate integer site setting

techAPJ:
[reported here](https://meta.discourse.org/t/sitesetting-numbers-not-properly-validated/25026?u=techapj)


eviltrout:
Thanks!

PR #3397, opened 2015-04-24
Bump rails to latest patch 4.1 release. (4.1.10)

arthurnn:
We should be using the latest patch release of the current supported
version.
Also this adds a more explicit version of rails on the Gemfile. This is
will help to test on rails 4.2 and master.


arthurnn:
:+1: ?

SamSaffron:
cool :+1:

PR #3398, opened 2015-04-24
Bump rails-observers to master version.

arthurnn:
https://github.com/rails/rails-observers/commit/876c522184a4ed26d564833f496ef1bfa7af654a
Has the patched applied on SamSaffron/rails-observers.
Also that version should work for rails 4> , so we should not branch it to install

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,7 +71,7 @@ end
 if rails_master?
   gem 'arel', git: 'https://github.com/rails/arel.git'
   gem 'rails', git: 'https://github.com/rails/rails.git'
-  gem 'rails-observers', git: 'https://github.com/SamSaffron/rails-observers.git'
+  gem 'rails-observers', git: 'https://github.com/rails/rails-observers.git'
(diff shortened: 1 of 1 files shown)


chancancode:
It might be easier to review if we batch them up. There are a few more related commits on the rails-42 branch



—

arthurnn:
@chancancode I am afraid to ship all at once because usually those things break apps ..
I like shipping updates one-by-one, so we can make sure things work isolated.
WDYT @SamSaffron ?

chancancode:
Not sure about this one. The only commit this gives you is the one that makes it work on Rails 5, which is not needed on Rails 4.1/4.2 mode, so we would be adding one hit dependency for the regular uses for no benefit. IMO we should keep this in the conditional since we are only using it to test against Rails master (and hopefully we will remove the observer code before Rails 5 is released, so we will probably never need this)

chancancode:
(Never need this in production, that is)

arthurnn:
@chancancode Fair point, I will leave this inside the `if` branch still.

arthurnn:
@chancancode Updated.

SamSaffron:
cool, thanks !

PR #3399, opened 2015-04-24
AMS 0.8.3 was released with the patch we want

arthurnn:
review @chancancode @SamSaffron


SamSaffron:
I need to confirm discourse bench perf does not regress with this upgrade, its a pivotal little piece :) 

can you run script/bench.rb before and after?

chancancode:
fwiw, 0.8.2 -> 0.8.3 is a pretty small diff, so we can probably deduce that by reading it :P

https://github.com/rails-api/active_model_serializers/compare/v0.8.2...v0.8.3

SamSaffron:
cool then :)

PR #3400, opened 2015-04-24
Update travis config file

arthurnn:
A few things in here:
- Test on ruby 2.2
- Allow failures in all variants of RAILS_MASTER test
- Use db:create task


SamSaffron:
thanks!

PR #3403, opened 2015-04-25
Rspec 3

arthurnn:
Main motivation: rspec 2 is not supported anymore, and some gems dont support rails 4.2 + rspec 2.
Fix: Update rspec to 3.x

I used this fantastic tool http://yujinakayama.me/transpec/ to do the heavy lift. And did a few other necessary manual changes.

review @SamSaffron @chancancode @coding-horror


spec/integrity/i18n_spec.rb
```diff
@@ -25,11 +25,11 @@
...
       client = YAML.load_file("#{Rails.root}/config/locales/client.#{locale}.yml")
-      client.count.should == 1
-      client[locale].should_not == nil
-      client[locale].count.should == 2
-      client[locale]["js"].should_not == nil
-      client[locale]["admin_js"].should_not == nil
+      expect(client.count).to eq(1)
+      expect(client[locale]).not_to eq(nil)
+      expect(client[locale].count).to eq(2)
+      expect(client[locale]["js"]).not_to eq(nil)
```
kirs: What about changing ``not_to eq(nil)` to `.to be_present`? I think it's better readable.

arthurnn: good call. will do that.

ZogStriP:
That's awesome. Would you mind rebasing and squashing into 1 commit?

myronmarston:
:+1: It's nice to see projects upgrading to RSpec 3 :).  Two questions/suggestions:

* Is there any reason you updated to RSpec 3.1 and not 3.2?  3.2 should be a drop-in replacement for 3.1 as long as you're only using RSpec's public APIs (and given that you were able to upgrade from 2.x to 3.x, I don't think you're using any of our private APIs).  3.3 is also just around the corner...
* To get the full benefits of RSpec 3 you may want to update your config to enable some new config options in RSpec 3.  See [.rspec](https://github.com/rspec/rspec-core/blob/v3.1.0/lib/rspec/core/project_initializer/.rspec) and [spec/spec_helper.rb](https://github.com/rspec/rspec-core/blob/v3.1.0/lib/rspec/core/project_initializer/spec/spec_helper.rb) for the recommended config options that 3.1 generates when you use `rspec --init` on a new project. (If you update to 3.2 you may want to consult those files for v3.2.0).

arthurnn:
@ZogStriP I didnt want to squash the commits in one, because logically they have different meaning, let me walk you through them:
-  Just ran Transpec command to migrate to new syntax
- Fixed one migrated test that didnt work
- Did the Gem upgrade
- Fixed the warnings

Of course I can squash them in one commit, however the good thing, specially about the first commit, is that it contains all the stats of the changes in the commit message.
@myronmarston I will go and move to 3.2, and apply your suggestions, thanks for reviewing it :heart:

.rspec
```diff
@@ -1 +1,2 @@
 --colour
+--require spec_helper
```
myronmarston: You actually may not want this here.  The point of putting this here (as recommended by the skeleton RSpec 3 generates) is to ensure it's always loaded w/o having to repeat `require 'spec_helper'` at the top of every spec file.  However, it doesn't look like you bothered to remove the `require 'spec_helper'` lines from the spec files.  Also, if you have this line in `.rspec`, it's important that `spec_helper` is really light weight since it will _always_ get loaded.  The `spec_helper` on the project is currently loading heavyweight dependencies like rspec-rails, rails, etc -- but presumably, some spec files aren't going to need those things loaded, and having this required is going to slow down running them.  That's the reason for this note at the top of the generated `spec_helper`:

https://github.com/rspec/rspec-core/blob/v3.2.3/lib/rspec/core/project_initializer/spec/spec_helper.rb#L7-L13

rspec-rails 3 follows this principle by keeping `spec_helper` nice and light weight, with minimal config that you'll always want, and also providing a `rails_helper` that loads rspec-rails and rails and can be required by the files that need rails loaded.

In the short term, if you want to add config like this, I'd suggest making a new `spec/minimal_spec_helper.rb` (or `spec/fast_spec_helper.rb` or whatever), moving the light-weight, non-rails config there (such as `mock_framework = :mocha` and `order = 'random'`), and updating this to require that file.  Or maybe punt on this for now until the `spec_helper` code has been split into `spec_helper` and `rails_helper`.

arthurnn: @myronmarston thanks a lot for this explanation :heart: :heart: :heart: :heart: :heart: , I am really glad you are here to catch my mistakes :smile_cat: .

For now, I will remove this, and let the test files require spec_helper. After that we can try to split that in 2 files as you suggested.
thanks again.

chancancode:
:heart:

arthurnn:
@SamSaffron good to merge this sooner rather than later otherwise it will get a bunch of conflicts with moving tests on master.

spec/components/site_setting_extension_spec.rb
```diff
@@ -3,19 +3,15 @@
...
-
-
   let :provider do
     SiteSettings::LocalProcessProvider.new
   end
 
   def new_settings(provider)
-    c = Class.new
-    c.class_eval do
+    c = Class.new do
```
SamSaffron: :)

SamSaffron:
Thanks so much !!! :heart:

PR #3404, opened 2015-04-27
FEATURE: magic login route for admin when SSO is enabled

techAPJ:
Login page screenshot:

screen shot 2015-04-27 at 16 59 54

cc @SamSaffron, @coding-horror

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,7 +6,7 @@ require_dependency 'rate_limiter'
 class UsersController < ApplicationController
 
   skip_before_filter :authorize_mini_profiler, only: [:avatar]
-  skip_before_filter :check_xhr, only: [:show, :password_reset, :update, :account_created, :activate_account, :perform_account_activation, :authorize_email, :user_preferences_redirect, :avatar, :my_redirect, :toggle_anon]
+  skip_before_filter :check_xhr, only: [:show, :password_reset, :update, :account_created, :activate_account, :perform_account_activation, :authorize_email, :user_preferences_redirect, :avatar, :my_redirect, :toggle_anon, :admin_login]
 
   before_filter :ensure_logged_in, only: [:username, :update, :change_email, :user_preferences_redirect, :upload_user_image, :pick_avatar, :destroy_user_image, :destroy, :check_emails]
   before_filter :respond_to_suspicious_request, only: [:create]
@@ -23,7 +23,8 @@ class UsersController < ApplicationController
                                                             :perform_account_activation,
                                                             :send_activation_email,
                                                             :authorize_email,
-                                                            :password_reset]
+                                                            :password_reset,
+                                                            :admin_login]
 
   def index
   end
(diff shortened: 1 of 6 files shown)


downeym:
:+1:

nlalonde:
I reviewed it and agree that it looks good.

app/controllers/users_controller.rb
```diff
@@ -344,6 +345,46 @@ def logon_after_password_reset
...
+
+      user = User.find_by(email: params[:email])
+      if user && user.admin? && user.id != Discourse::SYSTEM_USER_ID
+        email_token = user.email_tokens.create(email: user.email)
+        Jobs.enqueue(:user_email, type: :admin_login, user_id: user.id, email_token: email_token.token)
+        @message = I18n.t("admin_login.success")
+      else
+        @message = I18n.t("admin_login.error")
+      end
+    elsif !params[:token].blank?
```
ZogStriP: I tend to prefend `foo.present?` to `!foo.blank?`

app/controllers/users_controller.rb
```diff
@@ -344,6 +345,46 @@ def logon_after_password_reset
...
+      # token recieved, try to login
+      if EmailToken.valid_token_format?(params[:token])
+        @user = EmailToken.confirm(params[:token])
```
ZogStriP: So you can basically confirm *any* token using this endpoint?

techAPJ: Not exactly, this method confirms any token *generated via* `EmailToken` class.

In case of SSO enabled instance *the only way* to generate token via `EmailToken` class is using this method, and only Admin can do it.

ZogStriP: :+1:

app/controllers/users_controller.rb
techAPJ: Thanks for the tip! I made it so :)

app/controllers/users_controller.rb
```diff
@@ -344,6 +345,46 @@ def logon_after_password_reset
...
+
+    unless SiteSetting.enable_sso && !current_user
+      return redirect_to path("/")
+    end
+
+    if request.put?
+      RateLimiter.new(nil, "admin-login-hr-#{request.remote_ip}", 6, 1.hour).performed!
+      RateLimiter.new(nil, "admin-login-min-#{request.remote_ip}", 3, 1.minute).performed!
+
+      user = User.find_by(email: params[:email])
```
eviltrout: It might be better to filter the admin/system out via your your query instead of app code:

`user = User.where(email: params[:email], admin: true).where.not(id: Discourse::SYSTEM_USER_ID).first`

techAPJ: This is much better! Done! :+1:

riking:
> Not exactly, this method confirms any token generated via EmailToken class.

> In case of SSO enabled instance the only way to generate token via EmailToken class is using this method, and only Admin can do it.

There should definitely be a comment to that effect in the source, or someone isn't going to remember that later.

spec/controllers/users_controller_spec.rb
```diff
@@ -315,6 +315,57 @@
...
+
+      it 'does not enqueue mail with admin email and sso disabled' do
+        SiteSetting.enable_sso = false
+        Jobs.expects(:enqueue).never
+        put :admin_login, email: admin.email
+      end
+
+      it 'does not enqueue mail with normal user email and sso enabled' do
+        SiteSetting.enable_sso = true
+        Jobs.expects(:enqueue).never
```
SamSaffron: in general I would just test for error code here returned by controller as opposed to magic job stuff.

SamSaffron:
thanks!

PR #3407, opened 2015-04-28
Update discourse_quickstart.conf

helloravi:
It is just a spelling mistake.


SamSaffron:
the whole quickstart file should be nuked ... in fact .. doing so now

SamSaffron:
nuked that file :)

PR #3408, opened 2015-04-29
Check Rails.version instead of ENV

arthurnn:
Like that we can have code that works on multiple Rails versions, and we
dont need to mix a new method on `Kernel`.
Also, this makes easier to test have multiple versions, example 4.2 and master.


SamSaffron:
this looks fine to me, sure

PR #3411, opened 2015-04-29
Escape term for category autocomplete regex

Elberet:
See https://meta.discourse.org/t/error-when-text-entered-in-category-select-box-contains-invalid-regex-syntax/27999?u=elberet

### DIFF
--- a/app/assets/javascripts/discourse/components/category-group.js.es6
+++ b/app/assets/javascripts/discourse/components/category-group.js.es6
@@ -13,7 +13,8 @@ export default Ember.Component.extend({
       allowAny: false,
       dataSource(term){
         return Discourse.Category.list().filter(function(category){
-          const regex = new RegExp(term, "i");
+          const regex = new RegExp(term.replace(
+            /[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&"), "i");
           return category.get("name").match(regex) &&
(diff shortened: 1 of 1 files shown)


SamSaffron:
that regex is making me cringe, why not just check that the string is in the other string with indexOf or something.

riking:
Maybe we need a regexEscape() function that we can use in all the spots we need a regexp escape.

Elberet:
What's so bad about the regex? It matches a single character from character class container characters that have special meaning for JS regexes... the only reason it's hard to read is all the escaping.
Using indexOf "or something" would add two or three local variables and a loop.

DeanMarkTaylor:
I would recommend using a less escape crazy version:
```JavaScript
function escapeRegExp(string){
  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
}
```
Taken from: https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions

riking:
@Elberet it's not the regex per se, but more the fact that this function - escaping for use in a regex - is present in multiple places in Discourse, so there should be one place with the function that gets used every time.

Elberet:
Yeah, I'm on it. I'll follow up on Monday. ^^

ZogStriP:
ping @Elberet

ZogStriP:
Closing this for now. @Elberet feel free to ping us when you have time to work on it.

PR #3413, opened 2015-04-29
Add a GET route that logs a user out for Single Sign Out.

boushley:
We use this to allow our centralized authentication service to log a user out without needing to coordinate  Cross Origin POST/DELETE.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -4,7 +4,7 @@ require_dependency 'single_sign_on'
 class SessionController < ApplicationController
 
   skip_before_filter :redirect_to_login_if_required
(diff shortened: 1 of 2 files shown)


boushley:
I've filled out the CLA. Let me know if there are any other concerns.

SamSaffron:
this is a security risk people can then just add a random link on the site that logs people out.

eviltrout:
It's worse than that. Someone can add an image that links to it and anyone
who looks at that image will be logged out.

boushley:
Indeed, people could get annoying with something like this.

Would adding a token that is signed with the shared secret be more acceptable? Maybe an issued at and the user who is to be logged out?

SamSaffron:
closing this for now, why can not simply enable CORS and use a POST? nonces are possible but very complex to get right.

PR #3416, opened 2015-04-30
Simplify the pull_translations script

gschlager:
- There's no need for configuring `trans.<lang>` for each resource.
  Using the global `lang_map` in the Transifex client config is enough.
- The Transifex client can rename locales. No need to do this manually.
- Pull all translation files for each locale that has a file named
  `config/locales/client.<lang>.yml`. Up until now the script did only
  update existing files, but didn't add missing ones. That is why there
  are so many translation files for the poll plugin missing.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,92 +1,57 @@
 [main]
 host = https://www.transifex.com
+lang_map = es_ES: es, fr_FR: fr, ko_KR: ko, pt_PT: pt
 
 [discourse-org.clientenyml]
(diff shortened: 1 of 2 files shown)


ZogStriP:
:+1:

SamSaffron:
nice one, @techAPJ can you review and confirm it is all good so we can merge?

techAPJ:
Looks good, tested and works great! :+1:

PR #3417, opened 2015-04-30
FIX: flagging topic was giving error message

techAPJ:
[Reported here](https://meta.discourse.org/t/should-it-be-possible-to-flag-a-closed-issue/27876?u=techapj)


eviltrout:
:+1:

PR #3420, opened 2015-05-01
FEATURE: accept category_id as new-topic param to better support subcategory

techAPJ:


SamSaffron:
seems correct, looks like the composer model already takes care of dealing with ids

PR #3422, opened 2015-05-02
FIX: Don't ask for notification permission until first one

riking:


SamSaffron:
seems fine, lets try it out

PR #3423, opened 2015-05-02
Extract the parsing of polls

gschlager:
This makes parsing of polls from the cooked text reusable.
Importers need this change since the [base importer disables validations](https://github.com/discourse/discourse/blob/master/script/import_scripts/base.rb#L449) during the creation of posts. Otherwise it's hard to get to the correct poll options within an importer.


ZogStriP:
Thanks :+1:

PR #3424, opened 2015-05-03
Fix: @mention dialect matches too often

Elberet:
See https://meta.discourse.org/t/a-lone-gets-autocompleted-to-the-next-mention/28228


ZogStriP:
Thanks ;)

PR #3426, opened 2015-05-04
Notify on first like, too

riking:
As per https://meta.discourse.org/t/show-desktop-notifications-when-available/10183/110?u=riking


SamSaffron:
nahh I disagree with this, I omitted this on purpose.

PR #3428, opened 2015-05-04
UX: show caps lock warning on password reset page

techAPJ:
screen shot 2015-05-04 at 11 58 07


coding-horror:
looks good, but shouldn't there already be a translation for the caps lock alert somewhere since we had that feature in 1.2?

techAPJ:
@coding-horror: correct, there is a translation key for caps lock alert in `client.en.yml` file. I made the changes to use that instead.

PR #3430, opened 2015-05-04
Fixing map-marker FontAwesome typo

downeym:
The correct icon is map-marker, not map-maker.
http://fontawesome.io/icon/map-marker/


coding-horror:
oh cool thanks for this!

PR #3431, opened 2015-05-06
Update Translations

techAPJ:


gschlager:
Don't merge this yet! I've just noticed that I introduced a bug in 1023badd905e1b0c606eaa3c6435bcbf821361e6 :cry: 

The Transifex client doesn't apply the mappings from the `lang_map` to the root key in the yml files. It's using `es_ES` instead of `es`.

PR #3432, opened 2015-05-06
Fixes the language codes in the pull_translations script

gschlager:
Fixes the problem mentioned in PR https://github.com/discourse/discourse/pull/3431


techAPJ:
Thanks! :+1:

PR #3434, opened 2015-05-06
FEATURE: support category/subcategory name in new-topic route

techAPJ:
Supports creating new topic via URL like:

> https://meta.discourse.org/new-topic?title=topic%20title&body=topic%20body&category=support/wordpress


techAPJ:
Fixed the build.

PR #3435, opened 2015-05-07
FIX: display issues in the admin area for RTL forums

scossar:
admin.css is not being processed by the `R2.r2` method.  It mostly works, but there are a few display issues. Here are before and after images for what this PR fixes.

`.nav-stacked` needs to be pulled-right. `.setting-controls` and `.setting-label` need their positions reversed.
screenshot 2015-05-07 11 38 05

screenshot 2015-05-07 11 42 17

The background image for `.select2-search-choice-close` needs to be repositioned.
screenshot 2015-05-07 11 39 03

screenshot 2015-05-07 11 42 40

The admin controls in the `.current-style` editor need to be floated left.
screenshot 2015-05-07 09 16 47

screenshot 2015-05-07 09 18 00

It also adds a rule to `.rtl.scss` to force code blocks to always display LTR.

### DIFF
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -28,6 +28,11 @@
   @include rotate(180);
 }
 
+// Make code blocks always display LTR.
+.rtl code {
+  direction: ltr !important;
+}
+
 // Admin RTL styles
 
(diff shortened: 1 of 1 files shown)


scossar:
It looks like part of this was recently implemented.

PR #3444, opened 2015-05-09
Add pluralization rule to fr

Qasem-h:
It adds a pluralization rule to french

### DIFF
--- a/app/assets/javascripts/locales/fr.js.erb
+++ b/app/assets/javascripts/locales/fr.js.erb
@@ -1,3 +1,6 @@
 //= depend_on 'client.fr.yml'
 //= require locales/i18n
 <%= JsLocaleHelper.output_locale(:fr) %>
+I18n.pluralizationRules['fr'] = function (n) {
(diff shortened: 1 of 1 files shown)


SamSaffron:
why do we need this cc @ZogStriP  ?

Qasem-h:
It adds a pluralization rule to fr.js.erb. The rule always returns 'other'. This is because the fr translation files do not use the 'one' key for pluralization.
french does not distinguish between the singular and plural forms
of nouns in the same way as English does. For numbered elements,
always return the 'other' key from the translation file.

ZogStriP:
> french does not distinguish between the singular and plural forms

Yeah? Where did you get that from? As a French, I'm pretty sure we do.

Qasem-h:
@ZogStriP some one reported this in forum , franch translate is 99% but not works some strings in it

ZogStriP:
so?

Qasem-h:
So i think need this for fixing it

PR #3445, opened 2015-05-10
Don't send emails to anonymous users

lukegb:
Also changes behaviour of `User.real` to not return anonymous users.

This means user counts will no longer include them, and the
mailing list system will ignore them even if they somehow end up
with the feature turned on.


ZogStriP:
Thanks :+1:

PR #3446, opened 2015-05-11
Change Vagrantfile to default VM 1024MB of RAM

wlads:
Revert a commit that changed the default Vagrantfile to 2048MB of RAM
https://meta.discourse.org/t/discourse-as-your-first-rails-app/5751/105


ZogStriP:
Thanks

PR #3451, opened 2015-05-11
Stop sso login processing after rendering error

Urth:
This prevents a DoubleRenderError triggered on the redirect_to.


ZogStriP:
Thanks :+1:

PR #3452, opened 2015-05-11
Fix censored bullets

YesThatAllen:
Displays the actual characters `■` used to censor words, leaving `(&#9632;)` as a reference.

TODO Decide how best to handle this in RTL languages, and Dutch, which has this mid-sentence.

Topic at https://meta.discourse.org/t/should-word-censorings-site-setting-show-or-9632/28599


ZogStriP:
Translation files must be changed via transifex otherwise the changes will be overriden next time we update them. Also, I don't know if transifex correctly handle this kind of characters.

PR #3454, opened 2015-05-11
FIX: show user card on group page and fix styles

fantasticfears:
Bug: https://meta.discourse.org/t/wrong-reply-style-for-group-post-page/28343

1. Fix the wrong quote style in the post stream by removing max width constraint
2. Category is moved to right next to title.
3. Username is not added as bug page described. So weird if the username and name are the same.

screenshot 2015-05-12 01 00 15
screenshot 2015-05-12 01 00 19
screenshot 2015-05-12 01 10 01
screenshot 2015-05-12 01 10 04


fantasticfears:
@eviltrout Can you take a look at this? This also fixes the Ember deprecation which can help us from Travis failure.

ZogStriP:
Thanks, looks fine :+1:

PR #3455, opened 2015-05-11
Give Anonymous a name

pacificfils:
Simple fix to ensure that Anonymous mode can be used even when names are required.

Fixes:

    ActiveRecord::RecordInvalid (Validation failed: Name can't be blank)

### DIFF
--- a/app/services/anonymous_shadow_creator.rb
+++ b/app/services/anonymous_shadow_creator.rb
@@ -34,7 +34,7 @@ class AnonymousShadowCreator
       shadow = User.create!(
         password: SecureRandom.hex,
         email: "#{SecureRandom.hex}@anon.#{Discourse.current_hostname}",
-        name: "Anonymous",
+        name: I18n.t(:anonymous),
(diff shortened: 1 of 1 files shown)


pacificfils:
Additionally, it might be possible to use a Zero-Width Space (U+200B) to ensure that no name is displayed by the Anonymous username.

ZogStriP:
Anonymous is fine, but it'll have to be localizable. Can you make this change?

pacificfils:
@ZogStriP I am unsure how to make that change here.

How would I proceed?

gschlager:
Why don't you use the username as name? Discourse doesn't show the name when it has the same value as the username and it would still meet the requirement, wouldn't it?

SamSaffron:
I am not really sure about this.... 

anonymous21 "anoymous" ... all through the forum seems a bit odd and extra verbose. I think a better strategy is allowing mods to assign an avatar for ALL anons.

ZogStriP:
Used @gschlager's suggestion to properly fix the issue (cf. d4974b70938b07a60d8e6edbd258940a3aeb2c01)

PR #3457, opened 2015-05-12
Fix permissions of various files

darix:
Follow up to https://meta.discourse.org/t/files-with-weird-permissions-in-the-git-repository/28393


eviltrout:
Thanks!

PR #3458, opened 2015-05-12
Add Bosnian langage for discourse

Qasem-h:
Add Bosnian langage for discourse
Translated by : @Revolucionar


ZogStriP:
Thanks :+1:

PR #3459, opened 2015-05-12
Add http-ip param for mailcatcher

arturtr:
Fixes problems with port-forvarding for mailcatcher web-interface


eviltrout:
What is the problem that this fixes?

arturtr:
Whithout this params I cann't open http://localhost:4080/ - port not forwarded

eviltrout:
Weird, that definitely used to work in the old image! Thanks for the fix.

PR #3461, opened 2015-05-13
FIX: category custom slug can't be set when generation method is none

fantasticfears:


SamSaffron:
seems reasonable .. thanks

PR #3463, opened 2015-05-13
FIX: match subdomain with email domain blacklist

techAPJ:
[requested here](https://meta.discourse.org/t/email-domain-blacklist-with-wildcards/28627/9?u=techapj)

### DIFF
--- a/lib/validators/email_validator.rb
+++ b/lib/validators/email_validator.rb
@@ -17,7 +17,7 @@ class EmailValidator < ActiveModel::EachValidator
 
   def email_in_restriction_setting?(setting, value)
     domains = setting.gsub('.', '\.')
-    regexp = Regexp.new("@(#{domains})", true)
+    regexp = Regexp.new("@(.+\.)?(#{domains})", true)
     value =~ regexp
   end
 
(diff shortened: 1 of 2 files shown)


ZogStriP:
build failed?

techAPJ:
Yes, didn't noticed that :disappointed: .. fixed now

lib/validators/email_validator.rb
```diff
@@ -16,9 +16,13 @@ def validate_each(record, attribute, value)
...
-    domains = setting.gsub('.', '\.')
-    regexp = Regexp.new("@(#{domains})", true)
-    value =~ regexp
+    is_blacklisted = false
+    domains = setting.split("|")
+    email_domain = value.split("@").last if value
+    domains.each { |domain|
+      is_blacklisted = true if email_domain && (email_domain == domain || email_domain.end_with?(".#{domain}"))
+    }
+    is_blacklisted
```
ZogStriP: Hmm, I would have simply updated the regex to something like `Regexp.new("@(.+\.)?(#{domains})", true)`

techAPJ: Awesome! Updated the code.

PR #3464, opened 2015-05-13
Add a SSO logout endpoint.

boushley:
I tried adding a SSO logout endpoint via GET [before](https://github.com/discourse/discourse/pull/3413) but was told to perform a logout via CORS.

Currently there is a problem with making requests that require pre-flighting via CORS: https://meta.discourse.org/t/cors-preflight-checks-dont-get-porper-headers/28779

This works around that issue and allows for calling logout without first needing to make a request to `/session/csrf` (which the CORS preflight problem prevents us from doing).

Using _method = DELETE and a POST request this can be run without a pre-flight request. The check-xhr skip for this endpoint is to avoid setting custom headers on CORS requests, since custom headers require a pre-flight request.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -4,7 +4,8 @@ require_dependency 'single_sign_on'
 class SessionController < ApplicationController
 
   skip_before_filter :redirect_to_login_if_required
-  skip_before_filter :check_xhr, only: ['sso', 'sso_login', 'become', 'sso_provider']
+  skip_before_filter :check_xhr, only: ['sso', 'sso_login', 'sso_logout', 'become', 'sso_provider']
+  skip_before_filter :verify_authenticity_token, only: ['sso_logout']
 
   def csrf
     render json: {csrf: form_authenticity_token }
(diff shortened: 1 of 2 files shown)


eviltrout:
It seems to me this is not secure. I could trigger a user to logout as long as SSO was enabled by making a POST to a discourse forum, which I could do via a hidden form for example.

Could you allow the OPTIONS request through but not the delete one?

boushley:
@eviltrout we should definitely allow the `OPTIONS` request through. If we fix that bug, then I can adjust this to add back on the `check-xhr` filter which requires a custom header. With that in place we would only be allowing these requests from the domain allowed by CORS.

Unfortunately I'm a bit out of my depth on diagnosing problems inside the rack middleware. I can certainly keep trying to track what is going on, but if you have any pointers on what may be mangling those headers, that would be super helpful.

eviltrout:
I went ahead and added support for OPTIONS when cors is enabled on discoruse. Try out this commit and let me know how it works for you:

https://github.com/discourse/discourse/commit/963b08f0639d8c67c9272da23a49582967c3510e

PR #3467, opened 2015-05-14
Enable CORS requests to pass necessary headers.

boushley:
To fully enable session deletion over CORS we need support for passing the
`X-Requested-With` header so that these requests can pass the `check-xhr` filter.

I also allowed the `X-CSRF-Token` to enable the alternative CSRF passing syntax.


boushley:
@eviltrout Thanks for enabling the preflight requests. It turns out we need one more header to fully enable session deletion over CORS.

I've tested with this change and things work great.

eviltrout:
Sure, thanks!

PR #3468, opened 2015-05-14
FIX: unhandle show report action

fantasticfears:
Another Ember 1.11 regression.

https://github.com/discourse/discourse/compare/discourse:master...fantasticfears:fix-dashboard?expand=1#diff-9dac043915f9c7827266856b449d87c8R163

The other part is to use component to get rid of deprecation.


fantasticfears:
cc @eviltrout

eviltrout:
Awesome, thanks!

PR #3469, opened 2015-05-14
FIX: topic list header

scossar:
The topic-list `category` header does not have a set width. Toggling it to display topics with no categories causes the width to collapse to the width of the word `category`. This causes the chevron icon to display underneath the text.

screenshot 2015-05-14 10 58 57

A solution is to add a min-with rule for `.topic-list .category`.  It also needs to be center aligned to match the other topic-list headers

screenshot 2015-05-14 11 12 09

The topic-list `activity` header is set to 60px. This causes the chevron icon to display below the text.

screenshot 2015-05-14 11 06 40

The replies (`.posts`) and `views` topic list headers are set at 65px. They display with the chevron icon to their right. Setting the `activity` header to 65px solves this problem.

screenshot 2015-05-14 11 09 45

### DIFF
--- a/app/assets/stylesheets/desktop/topic-list.scss
+++ b/app/assets/stylesheets/desktop/topic-list.scss
@@ -130,6 +130,7 @@
   }
   .category {
     min-width: 80px;
+    text-align: center;
   }
 
   .post-actions {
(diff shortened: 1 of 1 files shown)


SamSaffron:
@awesomerobot does this look good ?

scossar:
I'm not sure about the `text-align: center`. It looks a *little* wrong to have the `category` heading centered , but have the `users` heading aligned left. I tried centering the `users` heading, but that looks bad. The problem with not centering the `category` heading once it has been given a minimum width, is that when you hover over it the text looks misaligned.

PR #3470, opened 2015-05-14
In the case of an expired nonce, return a 400 status code instead of 500

ahuling13:
500 status codes are for unexpected server-side error scenarios. When an expired nonce is used by the client, a 4XX-level error is more appropriate because the client has submitted a bad request (by using an expired nonce). A 500 also causes Internet Explorer to show its default 500 page which does not show the error message and leads to a bad end user experience. I am choosing 400 for the new status rather than 401 or 403 because 401 requires a WWW-Authenticate header which would be difficult to generate in an SSO scenario and a 403 implies that no re-authentication will address the failure.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -57,7 +57,7 @@ class SessionController < ApplicationController
 
     sso = DiscourseSingleSignOn.parse(request.query_string)
     if !sso.nonce_valid?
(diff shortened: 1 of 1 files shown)


SamSaffron:
I think what we want here is status 419

ahuling13:
Good call, @SamSaffron. Updated.

SamSaffron:
merging in and will fix the failing test case myself. thanks

PR #3476, opened 2015-05-15
Add SSO setting to not trust emails automatically

paulkaplan:
[Feature request on /meta](https://meta.discourse.org/t/email-activation-when-using-sso/28861)

Having an SSO flow where the provider is forced to validate emails is causing spam issues when the provider does not want to do email validation. This allows an admin to tell discourse to do it's own email validation when using SSO.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -73,6 +73,11 @@ class SessionController < ApplicationController
         if SiteSetting.must_approve_users? && !user.approved?
           render text: I18n.t("sso.account_not_approved"), status: 403
           return
+        elsif !user.active?
+          activation = UserActivator.new(user, request, session, cookies)
+          activation.finish
+          session["user_created_message"] = activation.message
+          redirect_to users_account_created_path and return
         else
           log_on_user user
         end
--- a/spec/controllers/session_controller_spec.rb
+++ b/spec/controllers/session_controller_spec.rb
@@ -193,6 +193,54 @@ describe SessionController do
       expect(logged_on_user.custom_fields["bla"]).to eq(nil)
     end
 
+    context 'when sso emails are not trusted' do
+      before do
+        SiteSetting.sso_trusts_email = false
+      end
+
+      context 'if you have not activated your account' do
+        it 'does not log you in' do
+          sso = get_sso('/a/')
+          sso.external_id = '666' # the number of the beast
+          sso.email = 'bob@bob.com'
+          sso.name = 'Sam Saffron'
+          sso.username = 'sam'
+
+          get :sso_login, Rack::Utils.parse_query(sso.payload)
+
+          logged_on_user = Discourse.current_user_provider.new(request.env).current_user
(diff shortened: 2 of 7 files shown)


coding-horror:
hmm @SamSaffron will need to review this Australia Monday

paulkaplan:
Sounds good. He recommended it as a possible solution on our forum over at [discuss.inventables.com](http://discuss.inventables.com/).

spec/controllers/session_controller_spec.rb
```diff
@@ -193,6 +193,54 @@ def sso_for_ip_specs
...
+
+      context 'if you have activated your account' do
+        it 'allows you to log in' do
+          sso = get_sso('/hello/world')
+          sso.external_id = '997'
+          sso.sso_url = "http://somewhere.over.com/sso_login"
+
+          user = Fabricate(:user)
+          user.create_single_sign_on_record(external_id: '997', last_payload: '')
+          user.stubs(:active?).returns(true)
```
SamSaffron: not a big fan of this stubbing

app/controllers/session_controller.rb
```diff
@@ -73,6 +73,11 @@ def sso_login
         if SiteSetting.must_approve_users? && !user.approved?
           render text: I18n.t("sso.account_not_approved"), status: 403
           return
+        elsif !user.active?
```
SamSaffron: how often should we hit them with activiation emails? there should be some limit, no?

SamSaffron:
we need to sort out some limiting on activation emails cause it can get a bit out of hand. 

also I wonder if this should just be part of the sso paylod, add a untrusted_email=true to the payload you send us and then we can avoid needing another site setting.

paulkaplan:
You are right. Right now it would send an activation email every time an inactive user tried to log in via SSO. I wasn't quite sure what to do that. When creating users normally, the `create` action is only called once and that is where the email is sent. With the `sso_login`, there is not a separate action for create / log in (that is handled by the model). Any suggestions?

spec/controllers/session_controller_spec.rb
paulkaplan: I'm not that familiar with the best practices for stubbing. What should it be?

paulkaplan:
Also @SamSaffron I agree that it would work as a payload switch rather than a site setting. I think that detail is more about how you want people to be using the feature. If, as @coding-horror says, Discourse is getting "blamed" for spam and the goal of this is to prevent that, then a prominent site setting seems the right way to go. But I'll leave the choice up to you guys. Let me know and I'll gladly change the implementation.

paulkaplan:
@SamSaffron I replaced the site setting with a payload attribute `require_activation`. This is actually a much more flexible solution, because while our site doesn't have email validation, there are other ways that we could auto-trust users (for example, if they have placed an order on our site).

SamSaffron:
the code path looks safe to me... lets try it out... let me know how you go

PR #3482, opened 2015-05-17
Import/Export site customizations

riking:
Finally, we can have an easy way to share customizations!

image

image

### DIFF
--- a/app/assets/javascripts/discourse/components/json-file-uploader.js.es6
+++ b/app/assets/javascripts/discourse/components/json-file-uploader.js.es6
@@ -3,6 +3,7 @@ export default Em.Component.extend({
   fileInput: null,
   loading: false,
   expectedRootObjectName: null,
+  hover: 0,
 
   classNames: ['json-uploader'],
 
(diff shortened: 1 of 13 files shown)


riking:
Test failures are unrelated.

app/assets/javascripts/discourse/components/json-file-uploader.js.es6
```diff
@@ -0,0 +1,105 @@
...
+    const self = this;
+
+    const $fileInput = $this.find('#js-file-input');
+    this.set('fileInput', $fileInput[0]);
+
+    $fileInput.on('change', function() {
+      self.fileSelected(this.files);
+    });
+
+    const $dragContainer = $this.find('.jsfu-shade-container');
```
SamSaffron: supporting drag and drop here seems like a massive amount of gold plating. not sure we need it .

config/locales/client.en.yml
```diff
@@ -143,6 +143,7 @@ en:
     every_two_weeks: "every two weeks"
     every_three_days: "every three days"
     max_of_count: "max of {{count}}"
+    alternation: "or"
```
SamSaffron: "or" will never translate right, you need to translate the whole sentence

riking: Yeah, wasn't sure about the design here.

Any suggestions? Here is where it's used:

app/assets/javascripts/discourse/templates/modal/upload-customization.hbs
```diff
@@ -0,0 +1,8 @@
+<form {{action "dummy" on="submit"}}>
+  <div class='modal-body'>
+    {{json-file-uploader value=customizationFile extension=".dcstylejson"}}
```
SamSaffron: really not a fan of dcstylejson its a real weird extension.

either 

dcstyle

or 

dcstyle.json if we must

app/assets/javascripts/discourse/components/json-file-uploader.js.es6
riking: I tried using the system and found that drag&drop was needed for a nice flow, at least in Chrome.

 1. go to meta, click download button on theme, it goes to bottom of window
 2. go to your site, click import
 3. drag file from the bottom up into the import window

app/assets/javascripts/discourse/templates/modal/upload-customization.hbs
riking: Agreed, will revert to `.dcstyle.json`

coding-horror:
we really should pull this in, maybe as soon as 1.4 is ongoing?

PR #3485, opened 2015-05-17
FIX: also inline category badge styles to digest email

fantasticfears:
cc @coding-horror

Fix the broken digest email styles introduced in https://github.com/discourse/discourse/commit/54ce66fc62ee12f231d1b880e6b09910e487a5ef


SamSaffron:
thanks

PR #3490, opened 2015-05-19
Fix RTL Customize Toggles & Dashboard

Qasem-h:
admin-cous
admin-couss

### DIFF
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -54,8 +54,8 @@ float: left;
   margin-top: 10px;
   width: 60%;
 }
-.rtl .dashboard-left {
-  float: right;
+.rtl .dashboard-right {
+  float: left;
   margin-top: 10px;
-  width: 60%;
+  width: 40%;
 }
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -49,3 +49,8 @@ float: left;
 position: static !important;
 float: left;
 }
+.rtl .dashboard-left {
+  float: right;
+  margin-top: 10px;
+  width: 60%;
+}
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -39,7 +39,7 @@
   float:right !important;
 }
 .rtl .autocomplete {
-left: 0px !important;
+left: 1 px !important;
 }
 .rtl .customize .current-style .toggle-mobile {
 position: static !important;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/common/base/rtl.scss
```diff
@@ -38,6 +38,14 @@
 .rtl .pull-left {
   float:right !important;
 }
-.rtl .autocomplete {
-left: 360px  !important;
+
+left: 0px !important;
```
ZogStriP: I don't think you want that without a selector

Qasem-h: Sorry I forget to change it

SamSaffron:
are we sure about this change to 

 .rtl .autocomplete  

sounds VERY wide to me and not specific to this page.

Qasem-h:
Im sure about that,

Qasem-h:
@ZogStriP  can you review this PR ?

Qasem-h:
dashbord 
befro 
new0

after
new1

Qasem-h:
and aotucomplite  : 
if 0px
new-p-0
and its now like this
new1

SamSaffron:
@scossar can you review?

scossar:
I'll look at it later today.

scossar:
It's good. I think there are a couple of things need to be fixed.

Adding the `!important` flag to a direction makes it so that it is NOT flipped. That what we want here.
```
.rtl .customize .current-style .toggle-mobile {
  position: static !important;
  float: left !important; // This needs the !important flag
}

.rtl .customize .current-style .toggle-maximize {
  position: static !important;
  float: left !important; // This needs the !important flag
}
```

For `.dashboard-left` and `.dashboard-right` it might be good enough to do it like this. The width has already been defined and can be left out here. The `margin-top` is not in the main Discourse styles. Maybe there is a reason to use it here.
```
.rtl .dashboard-left {
  float: right !important;
}
.rtl .dashboard-right {
  float: left !important;
}
```

 `.rtl .autocomplete` doesn't need the `!important` flag. I think `.rtl .autocomplete { left: 27px; }` (which is flipped to `right: 27px;`) would be closer to the way it works in the left-to-right version of Discourse. I don't know if `.autocomplete` is used outside of the composer. If it is, this will have some kind of effect on it.

SamSaffron:
@Qasem-h can you take care of these changes?

Qasem-h:
@SamSaffron  Im tested it in localhost its works
and i well check it again TONIGHT

Qasem-h:
I think .autocomplete left need just 1px or 0 px but im changed to 27 px

scossar:
You should make it the way that looks correct to you.

SamSaffron:
there are too many commits here, can you rebase as single commit and I will merge.

PR #3491, opened 2015-05-19
FEATURE: layout direction from user locale

scossar:
@SamSaffron here's the PR.

What it does:
- it create two new stylesheets (desktop_rtl.scss and mobile_rtl.scss). Right now they are just copies of the ltr versions

- adds a conditional check to `_discourse_stylesheet.html.erb` to load the correct stylesheet. I've commented what it is doing. The logic could probably be more elegant. I'm assuming that this needs to be rewritten as a helper function.

- adds a `rtl` option to `DiscourseStylesheets.compile`  - it sets the rtl option based on it's target file's name.

- changes the rtl conditional check in `DiscourseSassCompiles.compile` so that it is based on the rtl option

- adds the `desktop_rtl` and `mobile_rtl` to the files array in `assets:precompile:css`  (I've left `force: true` as an option - that has been removed from the current Discourse version, so should probably be removed here.)


SamSaffron:
ok complete via, thanks for the footwork 

https://github.com/discourse/discourse/commit/45bdd23689f83a90de2df5e089bab5d52af1c4d4

PR #3492, opened 2015-05-20
changing color of progress bar

awesomerobot:
also made the light blue slightly less light, and consistent throughout

screenshot 2015-05-19 21 31 35

:tophat: 
:cactus:


SamSaffron:
awesome, much prefer this, I wonder if we should lighten the bg it a bit more so it stands out a bit less? 

also confirming we are good on mobile here

PR #3496, opened 2015-05-21
Fix parsing of boolean attribute in SSO

paulkaplan:
The `require_activation` field was added and intended to be a boolean, but wasn't being parsed as one (i.e. if anything was present, including "false", it evaluated to true).


ZogStriP:
Thanks, mind squashing the commits into 1?

paulkaplan:
Squashed!

SamSaffron:
cool good catch, is this working out for you?

paulkaplan:
@SamSaffron we are about to enable it! I'll let you know.

PR #3498, opened 2015-05-24
Improve developer guide.

tgxworld:


docs/DEVELOPER-ADVANCED.md
```diff
@@ -19,7 +19,8 @@ Note: If you are developing on a Mac, you will probably want to look at [these i
 ## Before you start Rails
 
 1. `bundle install`
-2. `bundle exec rake db:create db:migrate db:test:prepare db:seed_fu`
+2. Start up Redis by running `redis-server`
+3. `bundle exec rake db:create db:migrate db:test:prepare`
```
SamSaffron: I don't think we need db:test:prepare

tgxworld: I'm not sure now. Will look into it.

Also, a few lines down the guide

> The seed data includes a pinned topic that explains how to get an admin account, so start there! Happy hacking!

I don't see it any where in the seed data. Do you still want to include it in the seed data? I can add it to the PR.

tgxworld: @SamSaffron I just tested, `db:test:prepare` is still required.

coding-horror:
seems OK then!

PR #3499, opened 2015-05-24
FIX: add missing username properties to usercard

fantasticfears:
Fix: https://meta.discourse.org/t/usercards-dont-show-shortnames-when-switched-directly/29150


ZogStriP:
Thanks :+1:

PR #3500, opened 2015-05-25
FIX: rare user card exception

fantasticfears:
- Use jquery.dotdotdot to help with block ellipsis.
- trigger the shown logic in user card view
- some tweaks to user card show logic


ZogStriP:
Thanks :+1:

PR #3501, opened 2015-05-26
FIX: do not inject nofollow in all subdomain links

techAPJ:
https://meta.discourse.org/t/add-rel-nofollow-admin-setting-is-not-working-right-its-making-parent-domain-link-as-nofollow/29226?u=techapj

Also there was bug when domain like `foo.com` is added in `exclude_rel_nofollow_domains` setting, then domain like `nofoo.com` was also being allowed to be excluded from nofollow.

### DIFF
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -234,16 +234,19 @@ module PrettyText
     domains = SiteSetting.exclude_rel_nofollow_domains
     whitelist = domains.split('|') if domains.present?
 
-    site_uri = nil
+    site_uri = URI(Discourse.base_url)
+    site_uri_domain = site_uri.host
+    site_uri_domain = PublicSuffix.parse(site_uri.host).domain if PublicSuffix.valid?(site_uri.host)
     doc.css("a").each do |l|
       href = l["href"].to_s
       begin
         uri = URI(href)
-        site_uri ||= URI(Discourse.base_url)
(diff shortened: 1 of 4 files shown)


techAPJ:
Updated the PR to use [publicsuffix-ruby](https://github.com/weppos/publicsuffix-ruby) gem.

lib/pretty_text.rb
```diff
@@ -224,11 +224,15 @@ def self.add_rel_nofollow_to_user_content(html)
       href = l["href"].to_s
       begin
         uri = URI(href)
+        uri_domain = uri.host
+        uri_domain = PublicSuffix.parse(uri.host).domain if PublicSuffix.valid?(uri.host)
         site_uri ||= URI(Discourse.base_url)
```
ZogStriP: This should be move out of the loop.

techAPJ: Done :)

ZogStriP:
Seems it needs a rebase

techAPJ:
Rebased and squashed.

SamSaffron:
lets not merge this till we see the impact of this gem on memory.

techAPJ:
@SamSaffron fixed this via: https://github.com/discourse/discourse/commit/93ab03966ea9395f86b39fded3ef8f70e274857a

PR #3503, opened 2015-05-26
Fix missing bracket in name selector in url anchor jumping

paulkaplan:
The missing bracket was causing errors in Safari and Firefox. See [this meta thread](https://meta.discourse.org/t/intra-post-anchor-links-dont-work-in-safari-firefox/29125/7).


ZogStriP:
Thanks! :+1:

PR #3504, opened 2015-05-26
FIX: add missing translation keys

techAPJ:
Reported here: https://meta.discourse.org/t/missing-english-copy/27243?u=techapj


techAPJ:
Fixed the build.

PR #3506, opened 2015-05-28
FIX: Reply to Topic shortcut no longer needs the Reply Topic Button to be available

cpradio:
Use the controller to open the reply composer window so the button doesn't have to be in the DOM.
https://meta.discourse.org/t/shift-r-only-works-sometimes/29332


ZogStriP:
Awesome! Thanks :+1:

PR #3507, opened 2015-05-28
FIX: attachment links in topic map were giving 404

techAPJ:
Reported here -- https://meta.discourse.org/t/links-to-uploaded-files-in-topic-details-are-wrong-try-discourse-org/25537?u=techapj


app/serializers/topic_link_serializer.rb
```diff
@@ -25,6 +26,10 @@ def internal
     object['internal'] == 't'
   end
 
+  def attachment
+    object['internal'] == 't' && !!(object['url'] =~ /uploads\/default/)
```
ZogStriP: Have you tested using S3 as a storage engine?

PR #3508, opened 2015-05-28
Better error message when new registration limit from an IP address is reached

techAPJ:
cc @SamSaffron

### DIFF
--- a/lib/validators/allowed_ip_address_validator.rb
+++ b/lib/validators/allowed_ip_address_validator.rb
@@ -4,9 +4,11 @@ class AllowedIpAddressValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
     if record.ip_address
-      if ScreenedIpAddress.should_block?(record.ip_address) ||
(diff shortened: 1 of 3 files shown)


lib/validators/allowed_ip_address_validator.rb
```diff
@@ -4,10 +4,12 @@ class AllowedIpAddressValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
     if record.ip_address
-      if ScreenedIpAddress.should_block?(record.ip_address) ||
-         (record.trust_level == TrustLevel[0] && SpamHandler.should_prevent_registration_from_ip?(record.ip_address))
+      if ScreenedIpAddress.should_block?(record.ip_address)
         record.errors.add(attribute, options[:message] || I18n.t('user.ip_address.blocked'))
       end
+      if record.trust_level == TrustLevel[0] && SpamHandler.should_prevent_registration_from_ip?(record.ip_address)
```
ZogStriP: Should't this be inside the other `if`?

PR #3512, opened 2015-05-30
FIX: typo at client.ja.yml

d6rkaiz:
Fixes from double-byte characters to single-byte character and spaces.


gschlager:
You should fix this in [Transifex](https://www.transifex.com/projects/p/discourse-org/language/ja/).

d6rkaiz:
Sorry, I tried to sign up in Transifex, but do not know how to use.

ZogStriP:
Like @gschlager said, you should use transifex to update translations.

PR #3513, opened 2015-06-01
UX: show meaningful tooltip for user menu

techAPJ:
Meta topic: https://meta.discourse.org/t/tooltip-for-user-menu-is-not-meaningful/29432?u=techapj


SamSaffron:
@coding-horror  I prefer 

"Your profile actions" or "My profile actions" here which is faster and better imo than 

"sam profile" which is plain odd.

coding-horror:
sure whatever you like here is fine with me, what we had before was super
bad

SamSaffron:
cool cleaning this its rather odd cause its reusing localization it should not ... fixing shortly

PR #3514, opened 2015-06-01
FIX: new task to delete old drafts weekly

techAPJ:
Meta topic: https://meta.discourse.org/t/many-drafts-of-published-posts-exist-in-the-database/27046

### DIFF
--- a/app/jobs/scheduled/clean_up_drafts.rb
+++ b/app/jobs/scheduled/clean_up_drafts.rb
@@ -1,0 +1,12 @@
+module Jobs
+  class CleanUpDrafts < Jobs::Scheduled
+    every 1.week
+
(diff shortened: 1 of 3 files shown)


coding-horror:
I don't like having two settings here, perhaps you could set the days to zero to disable the job?

And why would you want this disabled anyway, set it to 9999 days...

techAPJ:
Okay I removed the setting to disable it, setting it to something like 99999 will work for most cases (if they want to disable it).

SamSaffron:
hold your horses please, lets first find the orphan drafts, the ones we absolutely want to remove cause they were already published

also kind of opposed to carrying around yet another job to save an infinitesimal amount of db space.

PR #3519, opened 2015-06-02
Respect cookie[:destination_url] in Single Sign On

dansingerman:
When the login_required setting is true, the destination URL is dropped. This change means it will be
respected at login time

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -413,17 +413,22 @@ class ApplicationController < ActionController::Base
       raise Discourse::InvalidAccess.new unless current_user && current_user.staff?
     end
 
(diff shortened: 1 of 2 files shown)


dansingerman:
I tend to agree.

This fix is consistent with how destinations are treated in ApplicationController#redirect_to_login_if_required

I guess that should be refactored as well if we want to store the destination in a session?

SamSaffron:
yeah sounds good to me

dansingerman:
@SamSaffron updated as discussed :)

SamSaffron:
Sorry for the delay, this does look right, would not mind a few tests for it though in a separate pr

PR #3520, opened 2015-06-03
UX: decrease margin between title and post in email digest

techAPJ:


coding-horror:
meh, I was assuming there would be some other fix needed with this.. it's fine as is if we know gmail is interefering with the styles in some magical way. No need to change.

PR #3522, opened 2015-06-03
FIX: move old drafts cleanup to the dedicated method and add test

techAPJ:
Made the improvements @SamSaffron described [here](https://github.com/discourse/discourse/commit/c987a1d57103e50d787bb465848ccd696bded3a0#commitcomment-11489081).

`delete_drafts_older_than_n_days` setting now defaults to 180 days.


spec/models/draft_spec.rb
```diff
@@ -55,6 +55,13 @@
     Draft.cleanup!
 
     expect(Draft.count).to eq 1
+
+    # should cleanup drafts more than 180 days old
+    SiteSetting.stubs(:delete_drafts_older_than_n_days).returns(180)
```
SamSaffron: minor but simpler 

SiteSetting.delete_drafts_older_than_n_days = 180

no need to stub here .

PR #3524, opened 2015-06-03
UX: prevent horizontal scroll on iPhone

techAPJ:
cc @coding-horror


ZogStriP:
Which iPhone?

techAPJ:
iPhone 4 and above, here's the repro before fix: http://recordit.co/rAZmg7Ff8p

PR #3525, opened 2015-06-03
UX: post content was moving up when viewing replies

techAPJ:
Before:

before

After:

after

cc @coding-horror


riking:
Uh, doesn't this leave out moderator posts?

coding-horror:
ok well feel free to PR it

PR #3528, opened 2015-06-05
FIX: staff should be immune to max_invites_per_day setting

techAPJ:
This was regressed in [this commit](https://github.com/discourse/discourse/commit/dd91d5b02f74bf0226a5d2b7502eb6f000f72dca).

cc @SamSaffron


eviltrout:
Thanks!

PR #3529, opened 2015-06-05
FEATURE: Use created_at to remove an ip if its last_match_at is null

cpradio:
Only use created_at if the last_match_at is NULL (or not populated) to ensure all IPs are appropriately cleaned up
https://meta.discourse.org/t/look-at-created-at-when-deleting-ips-if-last-match-at-is-null/29682

### DIFF
--- a/app/jobs/scheduled/clean_up_unmatched_ips.rb
+++ b/app/jobs/scheduled/clean_up_unmatched_ips.rb
@@ -11,7 +11,7 @@ module Jobs
 
       # remove old unmatched IP addresses
       ScreenedIpAddress.where(action_type: ScreenedIpAddress.actions[:block])
-                       .where("last_match_at < ?", last_match_threshold)
+                       .where("last_match_at < ? OR (last_match_at IS NULL AND created_at < ?)", last_match_threshold, last_match_threshold)
                        .destroy_all
     end
 


app/jobs/scheduled/clean_up_unmatched_ips.rb
```diff
@@ -11,7 +11,7 @@ def execute(args)
 
       # remove old unmatched IP addresses
       ScreenedIpAddress.where(action_type: ScreenedIpAddress.actions[:block])
-                       .where("last_match_at < ?", last_match_threshold)
+                       .where("last_match_at < ? OR last_match_at IS NULL AND created_at < ?", last_match_threshold, last_match_threshold)
```
ZogStriP: An `OR` and an `AND` in the same clause without paranthesis feels weird

cpradio: I can make it use paranthesizes, fairly certain order of operations will process the AND first though. If it makes you feel better, I'll definitely add them.

ZogStriP: Yup, add them plz ;)

cpradio:
I don't think this build failure is a result of my code... so not sure what is happening.

ZogStriP:
That's travis being capricious

cpradio:
Phew! Thanks for the explanation, as I was worried my rebase somehow corrupted something outside the file I changed (even though I knew that wasn't possible). :+1:

riking:
I think you could have actually done `COALESCE(last_match_at, created_at) < ?` which is a bit smaller.

PR #3530, opened 2015-06-05
Fix typo

stripTM:
In footer show "Hay are 87 no leídos y 9 nuevos temas restantes", remove 'are'


stripTM:
CLA signed.

What is the reason for the Travis error?

ZogStriP:
I'm sorry but we only accept translations updates via https://www.transifex.com/projects/p/discourse-org/. Otherwise, we lose these fixes when we pull translations from transifex.

PR #3532, opened 2015-06-07
Adding " the" for readabilty

YesThatAllen:
See https://meta.discourse.org/t/should-any-email-subjects-from-discourse-not-be-tagged/19911/


coding-horror:
"Welcome to the Discourse Meta?"

No, I don't think this makes sense.

PR #3535, opened 2015-06-08
Add admin creation to developers guide.

tgxworld:


ZogStriP:
There is a better rake task to create an admin user `rake admin:create` ;)

tgxworld:
@ZogStriP  Should we at least remove the following line? :smile_cat: 

> The seed data includes a pinned topic that explains how to get an admin account, so start there!

PR #3537, opened 2015-06-09
Clean RSS author tag; remove email address

dfabulich:
The author tag doesn't require an email address. We should just show the user name instead.


dfabulich:
Switching to dc:creator instead, in pull request #3538

PR #3538, opened 2015-06-09
Convert author tag to dc:creator

dfabulich:
RSS spec says the author tag should be an email address, forcing us to put in a junk no-reply@example.com email. Instead, we should use dc:creator, which allows us to use any name we want for the user.


coding-horror:
seems like a good idea, I remember the author tag being a catch-22 because the spec does require it to be email, and we don't want to expose emails for regular users..

techAPJ:
Thanks! :+1:

PR #3539, opened 2015-06-10
FIX: Incorrect check when no text is selected.

tgxworld:
When I click on a post and even if no text is selected, `rangeCount` will still be `1`. Therefore, we're running unnecessary code until we check it against the buffer.

From the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Selection/isCollapsed), `isCollapsed` checks if any text has been selected.

cc/ @SamSaffron 

EDIT: Just spotted this in the Docs.
> Keep in mind that a collapsed selection may still have one (or more, in Gecko) Ranges, so Selection.rangeCount may not be zero.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/quote-button.js.es6
+++ b/app/assets/javascripts/discourse/controllers/quote-button.js.es6
@@ -27,7 +27,7 @@ export default DiscourseController.extend({
 
     const selection = window.getSelection();
     // no selections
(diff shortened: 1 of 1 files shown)


ZogStriP:
Let's try it.

tgxworld:
@ZogStriP Cool. If you set a breakpoint at the check, you'll notice that it no longers go pass it even though you've only clicked on the post.

PR #3540, opened 2015-06-10
Don't limit @mention autocomplete to latin characters

noamraph:
The userSearch() function, used for @mention autocomplete, returned an empty
list if the query string included non-latin characters or spaces. This removes
this restriction, so you can search users by any characters in their display
name, including spaces.


eviltrout:
Looks good thanks!

noamraph:
Cool! Thanks!

PR #3541, opened 2015-06-10
Remove actionpack-action_caching gem from Gemfile

camelmasa:
Discourse is not using the gem.

Please close this PR if using the gem at somewhere :ok_hand:


SamSaffron:
thanks, good catch

camelmasa:
@SamSaffron Thank you for merge :smile:

PR #3545, opened 2015-06-11
Add plugin outlet to user card metadata

ligthyear:
This PR adds a plugin-outlet in the "metadata" area of the user card, allowing plugins to easily extend the content there. 

In the example, that I am currently building, we want to show the (very important) points of reputation someone gained on their user card directly (hint, it's the _Reputation 10 Points_ part ;) ):

screen shot 2015-06-11 at 15 26 19


ZogStriP:
Seems fine.

PR #3549, opened 2015-06-14
FIX: vertically align category badge-wrapper in suggested-topics h3

scossar:
The category badges in the suggested-topics h3 have a different font size than their parent. The category-badges also have different line-heights, depending on whether they have the class .bullet, .box, or .bar. 

This adds a class to the parent h3 and sets its position to relative. Each of the category badge-wrapper classes can now have their vertical position set through their top value in `topic.scss`.

screenshot 2015-06-14 02 46 41

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -484,6 +484,3 @@ DEPENDENCIES
   uglifier
   unf
   unicorn
-
-BUNDLED WITH
-   1.10.3
--- a/app/assets/javascripts/discourse/templates/topic.hbs
+++ b/app/assets/javascripts/discourse/templates/topic.hbs
@@ -110,7 +110,7 @@
                 <div class='topics'>
                   {{basic-topic-list topics=model.details.suggested_topics postsAction="showTopicEntrance"}}
                 </div>
-                <h3>{{{view.browseMoreMessage}}}</h3>
+                <h3 class='topics-more'>{{{view.browseMoreMessage}}}</h3>
               </div>
             {{/if}}
 
--- a/app/assets/javascripts/discourse/templates/topic.hbs
+++ b/app/assets/javascripts/discourse/templates/topic.hbs
@@ -110,7 +110,7 @@
                 <div class='topics'>
                   {{basic-topic-list topics=model.details.suggested_topics postsAction="showTopicEntrance"}}
                 </div>
-                <h3 class='topics-more'>{{{view.browseMoreMessage}}}</h3>
+                <h3 class='topics-other'>{{{view.browseMoreMessage}}}</h3>
               </div>
             {{/if}}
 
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -484,6 +484,3 @@ DEPENDENCIES
   uglifier
   unf
   unicorn
-
-BUNDLED WITH
-   1.10.3
```
SamSaffron: let's not remove this :) can you update bundler?

app/assets/javascripts/discourse/templates/topic.hbs
```diff
@@ -110,7 +110,7 @@
                 <div class='topics'>
                   {{basic-topic-list topics=model.details.suggested_topics postsAction="showTopicEntrance"}}
                 </div>
-                <h3>{{{view.browseMoreMessage}}}</h3>
+                <h3 class='topics-more'>{{{view.browseMoreMessage}}}</h3>
```
SamSaffron: why do we need .topics-more when we have #suggested-topics ?

SamSaffron:
@awesomerobot  can you have a review of this? perhaps correct it without pixel pushing?

scossar:
The reason for adding the .topics-more class was to avoid having to set all
#suggested-topics h3 positions to relative. But it works just as well if
you target #suggested-topics h3, and #suggested-topics .bullet .box and
.bar and leave out the .topics-more class.

scossar:
The problem with doing it this way is that it doesn't work if the text is
pushed onto more than one line.

scossar:
It works, but it's kind of a weird way of doing it.

scossar:
I've worked through this again. It seems like the simplest approach. I haven't found a better way of moving the badge icon and the badge text as a single unit.

Adding a class to the h3 parent of the category badge makes it easier to target the correct category badge - there are other `.badge-category` badges in the `#suggested-topics` div.

I renamed the new class to `topics-other` -  that could probably be better.
I copied the styles to `mobile/topic.scss`. The badge positions can be adjusted independently there.

coding-horror:
Can we use flexbox here? The problem seems to be vertical centering, which is the holy f'ing grail of all HTML and CSS, and Flexbox is the only non-insane way to do it. Rules out IE9. Kind of works in IE10. That's the downside.

scossar:
I just tried setting the 'Browse other topics' h3 to display flex. (`flex-direction: row; align-items: baseline;`)
It works, but flexbox  removes trailing white space from strings, and there are two strings in the h3 that use trailing whitespace for formatting. We'd need to add margins in the css. The alignment for the `.badge-wrapper .bullet` still needs to be adjusted after being centered with flexbox. The css to do that is a little weird - you have to change the line-height on the category-bullet and the top-margin on the category text. 
The css in the PR is probably more maintainable/readable that doing it that way.

coding-horror:
regardless, rather than tilting at the windmill of vertical centering with hacks, we should lean on flexbox for this since it's designed for expressly this purpose (among others), a  sane vertical center in CSS. So if we have to add margins elsewhere that is not a problem.

scossar:
Will center with flexbox

PR #3552, opened 2015-06-15
[fr] better translation for prevent

cake17:
The translation of `prevent` in french is, to me, `empêcher`


eviltrout:
Sorry, we need all our translations to go through Transifex:

cake17:
Ok sorry I didn't know, I'll try with transifex

PR #3553, opened 2015-06-15
FIX: vertically align category badge

scossar:
Adds  `.other-topics`class to `topic.hbs` to make it easier to target the special case of displaying category badges within the h3 element.

Forces `.other-topics` category badges to display inline so that they align correctly with their parent text.

Adjusts the vertical height of `.other-topics .badge-wrapper.bullet`.


scossar:
I'm having some issues with updating my version of bundler. When I get it sorted out I'll fix  `Gemfile.lock`. Feel free to make your own PR with the styles from here. I'm pretty sure it's the right approach.

scossar:
I removed the `other-topics` class. These styles can be safely added to the badge-wrapper class of any h3 element in the suggested-topics div.

scossar:
The Gemfile.lock for this is messed up. It was easier to just create a new PR.

PR #3554, opened 2015-06-15
FIX: set inline category badge to display inline

scossar:
After a lot of messing around with this, the solution turns out to be very simple. The alignment problems are coming from trying to display block and inline-block elements inside of the h3 text.


coding-horror:
:+1:

SamSaffron:
thanks, looks pretty clean

PR #3556, opened 2015-06-17
FEATURE: Changing timestamp through UI

AayEmmZee:
This commit adds the option to change timestamp through the UI from two
places. The selected menu options and the post menu!

We have made the changes which @eviltrout suggested previously! (thanks!)

### DIFF
--- a/app/assets/javascripts/discourse/components/date-field.js.es6
+++ b/app/assets/javascripts/discourse/components/date-field.js.es6
@@ -1,0 +1,4 @@
+
+export default Ember.TextField.extend({
+    type: 'date'
+});
--- a/app/assets/javascripts/discourse/controllers/back-date.js.es6
+++ b/app/assets/javascripts/discourse/controllers/back-date.js.es6
@@ -1,0 +1,22 @@
+import ModalFunctionality from 'discourse/mixins/modal-functionality';
+import ObjectController from 'discourse/controllers/object';
+import { categoryLinkHTML } from 'discourse/helpers/category-link';
+
+export default ObjectController.extend(ModalFunctionality, {
+    content: {},
+    needs: ["topic"],
+    post: null,
+
+    _forwardAction(name) {
+        const date = this.get('dateValue');
+        const time = this.get('timeValue');
+        const dateTime = date + ' ' + time ;
+        this.get("controllers.topic").send(name, this.get('model'), dateTime);
+        this.send("closeModal");
+    },
+
+    actions: {
+        backDatePost() {this._forwardAction("changeTimeStamp")}
+    }
+
+});
--- a/app/assets/javascripts/discourse/controllers/back-date.js.es6
+++ b/app/assets/javascripts/discourse/controllers/back-date.js.es6
@@ -1,11 +1,10 @@
 import ModalFunctionality from 'discourse/mixins/modal-functionality';
 import ObjectController from 'discourse/controllers/object';
-import { categoryLinkHTML } from 'discourse/helpers/category-link';
 
 export default ObjectController.extend(ModalFunctionality, {
     content: {},
     needs: ["topic"],
-    post: null,
+
 
     _forwardAction(name) {
         const date = this.get('dateValue');
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -3,7 +3,6 @@ import BufferedContent from 'discourse/mixins/buffered-content';
 import SelectedPostsCount from 'discourse/mixins/selected-posts-count';
 import { spinnerHTML } from 'discourse/helpers/loading-spinner';
 import Topic from 'discourse/models/topic';
-import Post from 'discourse/models/post';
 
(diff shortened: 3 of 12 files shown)


app/assets/javascripts/discourse/components/date-field.js.es6
```diff
@@ -0,0 +1,4 @@
+
+export default Ember.TextField.extend({
+    type: 'date'
```
eviltrout: If all you're doing is changing the type of field, you can do it in the hbs template instead. We don't need the `date-field` and `time-field` components.

`{{input type="date" value=something}}`

app/assets/javascripts/discourse/controllers/back-date.js.es6
```diff
@@ -0,0 +1,21 @@
...
+export default ObjectController.extend(ModalFunctionality, {
+    content: {},
+    needs: ["topic"],
+
+
+    _forwardAction(name) {
+        const date = this.get('dateValue');
+        const time = this.get('timeValue');
+        const dateTime = date + ' ' + time ;
+        this.get("controllers.topic").send(name, this.get('model'), dateTime);
```
eviltrout: It looks like `_forwardAction` only ever calls `changeTimeStamp`. It doesn't seem like the name needs to be dynamic in this case. In fact, it doesn't seem like it needs to be a function at all. Why not just put all this location in the `backDatePost` action?

Additionally, we use two spaces for indentation here. In some of the files you stuck with two but in others like this one you used 4.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -116,6 +116,13 @@ export default ObjectController.extend(SelectedPostsCount, BufferedContent, {
     deleteTopic() {
       this.deleteTopic();
     },
+    // Change Timestamp
+    changeTimeStamp(post, dateTime) {
+      if (this.get('selectedPostsCount') === 1) {
```
eviltrout: Why is this logic here? It seems like the `post` is always passed to this method. Why would it need to be retrieved again from what is selected?

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -495,6 +502,10 @@ export default ObjectController.extend(SelectedPostsCount, BufferedContent, {
     return post.get('post_number') === 1 && post.get('topic.expandable_first_post');
   }.property(),
 
+  canChangeTimeStamp: function() {
```
eviltrout: This should be `Ember.computed.equal('selectedPostsCount', 1)`

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -283,6 +283,15 @@ const Topic = RestModel.extend({
     });
     return Discourse.ajax("/t/" + this.get('id') + "/recover", { type: 'PUT' });
   },
+  
+    // Change Timestamp
+  changeTimeStamp(post, dateTime) {
+      return Discourse.ajax("/topics/back_date/" + post.id + "/" + dateTime, { type: 'POST' }).then(function (result) {
```
eviltrout: Updates to objects should use `PUT` instead of `POST`

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -283,6 +283,15 @@ const Topic = RestModel.extend({
...
+          post.set('created_at', result.basic_post.created_at);
+          location.reload();
```
eviltrout: I believe I commented on this last time you submitted this PR, but you shouldn't reload the page just because something changed. You should just update the object that changed and handlebars will update the UI automatically.

app/assets/javascripts/discourse/routes/topic.js.es6
```diff
@@ -63,6 +63,12 @@ const TopicRoute = Discourse.Route.extend(ShowFooter, {
       this.controllerFor('modal').set('modalClass', 'feature-topic-modal');
     },
 
+    // Change Timestamp
+    backDatePost(model) {
+      showModal('back-date', { model, title: 'post.controls.backdate' });
+      this.controllerFor('back-date').setProperties( {post: model});
```
eviltrout: This shouldn't be necessary. The `back-date` controller already has the `model`

app/controllers/topics_controller.rb
```diff
@@ -414,6 +414,19 @@ def reset_new
     render nothing: true
   end
 
+  #Change Timestamp
+  def back_date
+    post = Post.find(params[:id])
+    time = params[:dateTime]
+    post.update(created_at: Time.parse(time), updated_at: Time.parse(time)) ? render_serialized(post, BasicPostSerializer) : render_json_error(post)
+    puts '------------------------------------------'
```
eviltrout: Please remove the `puts` statements from your code.

SamSaffron:
Closing this, appreciate the effort but 

1. PR was not updated with all @eviltrout's concern 
2. I would much prefer we actually agree that we need the feature before all this works gets done. This should have been discussed on meta first.

PR #3557, opened 2015-06-19
FIX: positioning of usercard in rtl layouts

scossar:
The usercard is positioned off the screen in rtl layouts.  This  PR checks the html direction attribute for 'rtl'. If the direction is rtl, it calculates the appropriate `position.right` based off of the `target.offset()` left position. 

I have created a plugin that using the same logic as a temporary fix. You can see it working on this forum: http://haxe.org.il1


scossar:
I was redeclaring the const `overage`. I have changed it two let statements in the if else block.

SamSaffron:
ok lets try this. thanks

PR #3546, opened 2015-06-11
FEATURE: Ctrl+f when topic search is active will close search.

tgxworld:
https://meta.discourse.org/t/pressing-ctrl-f-twice-on-topic-search-should-close-the-discourse-search-window/19649

PR #3559, opened 2015-06-19
FEATURE: Close search window when Ctrl+F is pressed

gschlager:
This is an alternative to #3546 and implements the expected behavior described in  https://meta.discourse.org/t/pressing-ctrl-f-twice-on-topic-search-should-close-the-discourse-search-window/19649/7

And it fixes a minor bug: Up until now all search related keyboard shortcuts enabled the "search context" since the [`selectContext` was always truthy](https://github.com/discourse/discourse/blob/bb5d27429fe12f892e62d70b0a481a39eeb2bf64/app/assets/javascripts/discourse/lib/keyboard_shortcuts.js#L158-L162).


SamSaffron:
lets try this out, thanks!

PR #3561, opened 2015-06-22
FIX: Tagging plugin was blocking composer status on smaller screens.

tgxworld:
Fixes: https://meta.discourse.org/t/saved-hidden-behind-tags-box/29576/6

In conjunction with: https://github.com/discourse/discourse-tagging/pull/34


ZogStriP:
Mind adding a before/after screenshots since it's CSS only change?

SamSaffron:
lets try this out.

PR #3563, opened 2015-06-23
fix script/setup_dev

thiesen:
hello!
i'm starting to contribute to discourse and had some trouble running `script/setup_dev` so i decided to fix that script and update the advanced developer documentation to make the setup easier.

what do you guys think about it?


eviltrout:
That setup_dev script doesn't seem to do all the steps you removed. For example `db:test:prepare` is not present. Maybe it should be removed? I'm not sure where we previously told people to use it.

thiesen:
@eviltroutshould do you think I should complete the `script/setup_dev` including the `db:test:prepare` to people start using the script or just close the PR and move on to another issue?

eviltrout:
I don't think we need this PR actually, the steps are OK to follow.

I think it might be better to just remove `setup_dev`.

thiesen:
@eviltrout is that ok?

eviltrout:
Yes, good to me, thanks :)

PR #3566, opened 2015-06-25
Fix right clicking username on profile card takes you to profile page

morashid92:
bug: https://meta.discourse.org/t/middle-right-clicking-on-username-of-profile-card-takes-you-to-the-page-you-are-viewing-not-the-profile/30404

Hey I have added the link-to helper and I believe this fixes the issue :)


eviltrout:
This is embedding a link within a link which is invalid HTML.

I fixed it in 0a9e2f54ff8f97

PR #3568, opened 2015-06-25
FIX: mark user as approved if an invite is already present

techAPJ:
If the site has `must approve users` setting enabled and admin/moderator sends an invite to an email. Then if the user signs up on the forum without clicking invite link, his account should be automatically approved.

[Bug reported here](https://meta.discourse.org/t/user-was-forced-to-wait-for-approval-despite-a-pending-invitation/30276?u=techapj).


eviltrout:
Looks good to me thanks!

PR #3569, opened 2015-06-25
FEATURE: Make abbreviated numbers (k, M) localizable

gschlager:
- Makes the decimal separator and thousands delimiter localizable
- Adds translations for all languages, otherwise this will look very ugly
- English locale: Changes the thousands delimiter from uppercase K to [lowercase k](https://en.wikipedia.org/wiki/Unit_prefix).

https://meta.discourse.org/t/is-there-a-way-to-translate-numbers-e-g-1-3k-in-the-views-column/30374

### DIFF
--- a/app/assets/javascripts/discourse/lib/formatter.js
+++ b/app/assets/javascripts/discourse/lib/formatter.js
@@ -258,14 +258,18 @@ relativeAge = function(date, options) {
 };
 
 var number = function(val) {
+  var formattedNumber;
+
   val = parseInt(val, 10);
   if (isNaN(val)) val = 0;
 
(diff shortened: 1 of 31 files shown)


riking:
LGTM, except I think that the format translations will need to be pushed to Transifex somehow - Transifex only reads the english file

gschlager:
Yeah, I know about the translations. But I thought it would be a good idea to temporarily add them here as well, otherwise there will be lots of complaints about missing translations since those numbers are used in the topic list where everyone can see them. We don't have a locale fallback yet. :wink:

eviltrout:
I like it, let's bring it in!

PR #3570, opened 2015-06-26
FIX: Middle click on user card image and name not working.

tgxworld:
Extension of https://github.com/discourse/discourse/commit/0a9e2f54ff8f97ebac800dd027715347b5f70976.

cc/ @eviltrout


eviltrout:
Looks good thanks!

PR #3571, opened 2015-06-26
UX: Dropdown not dismissing until next page renders.

tgxworld:
Address: https://meta.discourse.org/t/about-page-slow-menu-slow-to-disapear/24635.


tgxworld:
One behavior that will change here is that when we search within a topic, clicking on any results will close the dropdown as opposed to the previous behavior where the search will remain open. Was the previous behavior by design?

SamSaffron:
yes, previous behavior was by design.

tgxworld:
@sam ok! I'll update the PR to take that into account.

SamSaffron:
cool :+1:

PR #3572, opened 2015-06-26
FIX: cannot read property 'forEach' of undefined that prevents embed from working in Chrome

jqnatividad:
Fixes "Uncaught TypeError: Cannot read property 'forEach' of undefine…d" in Chrome that stops discourse embed.js from working.

See https://meta.discourse.org/t/how-can-i-troubleshoot-discourse-comments-embedding-feature/30323/10


ZogStriP:
Closing this. Feel free to comment to have it re-opened if this happens again.

PR #3573, opened 2015-06-26
FIX: Disable validation during thumbnail creation

gschlager:
This fixes problems with rebaking posts. There's no need to validate an upload during thumbnail creation.

https://meta.discourse.org/t/rebaking-posts-fails-when-attachments-arent-allowed-anymore/30429


eviltrout:

PR #3574, opened 2015-06-27
FIX: Remove invalid translation keys from polls plugin

gschlager:
This removes invalid translation keys from polls plugin and should allow Transifex to detect those keys as pluralized. And it changes the file headers since they were wrong.

https://meta.discourse.org/t/polls-labels-russian-pluralization-missing-translations/30235

### DIFF
--- a/plugins/poll/config/locales/client.en.yml
+++ b/plugins/poll/config/locales/client.en.yml
@@ -1,19 +1,27 @@
 # encoding: utf-8
+# This file contains content for the client portion of the poll plugin, sent out
+# to the Javascript app.
 #
-# Never edit this file. It will be overwritten when translations are pulled from Transifex.
-#
-# To work with us on translations, join this project:
+# To work with us on translations, see:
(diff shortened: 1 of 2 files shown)


eviltrout:
Why is `zero` causing this? It certainly is valid:

https://github.com/svenfuchs/i18n/blob/master/lib/i18n/backend/pluralization.rb#L23

gschlager:
Yes, it's valid in this implementation, but it's not part of the standard. It's even mentioned in [this comment](https://github.com/svenfuchs/i18n/blob/master/lib/i18n/backend/base.rb#L133-L135).

The [pluralization rules](http://www.unicode.org/cldr/charts/25/supplemental/language_plural_rules.html#en) for English consist only of "one" and "other". Transifex **currently** (I sent them a feature request which they may or may not implement) doesn't handle the special case of "zero", so it doesn't detect any key as pluralized if there exists anything else than "one" and "other".

SamSaffron:
I think this should be ok cause zero = other in this case.

PR #3575, opened 2015-06-27
jagged border on post summary gaps

awesomerobot:
screenshot 2015-06-27 13 26 30

:scissors:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -481,6 +481,3 @@ DEPENDENCIES
   uglifier
   unf
   unicorn
-
-BUNDLED WITH
-   1.10.3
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -481,6 +481,3 @@ DEPENDENCIES
   uglifier
   unf
   unicorn
-
-BUNDLED WITH
```
gschlager: You shouldn't remove that. Update your version of bundler instead: `gem install bundler`

coding-horror:
thanks very much! I pulled this in manually since I didn't know how to omit the bundler file :cow2: :dash:

PR #3576, opened 2015-06-29
Add option to set the newuser_spam_host_threshold.

tgxworld:
@eviltrout When importing a large set of data, I ran into the problem where some new users had posts that were missing due to the `newuser_spam_host_threshold`. I think an option to allow it to be configurable would be useful.


tgxworld:
Hmm actually there are a [bunch of settings](https://github.com/discourse/discourse/blob/master/config/site_settings.yml#L400-L410) that may potentially result in the import failing. Do you think we should override them during the import?

eviltrout:
Unfortunately the disqus import is quite old and doesn't use our new import base class which already disables all those settings. It needs to be ported over to the new format.

tgxworld:
@eviltrout Ah Ic. In that case I'll port the Disqus importer to the new format since the logic in this current script still works :smile:

PR #3582, opened 2015-06-30
FIX: combo-box should observe on nameProperty

riking:


eviltrout:
That's a lot of observing for things that haven't until this time needed to change.

Maybe we should hide observing it behind an option? `nameChanges=true` or something?

riking:
Okay, sure.

eviltrout:
:+1:

PR #3585, opened 2015-07-01
FEATURE: Reserved usernames

riking:
basically this https://twitter.com/FioraAeterna/status/614433067877273601

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -231,6 +231,10 @@ class UsersController < ApplicationController
       return fail_with("login.password_too_long")
     end
 
+    if SiteSetting.reserved_usernames.include? params[:username].downcase
+      return fail_with("login.reserved_username")
+    end
+
     user = User.new(user_params)
 
     # Handle custom fields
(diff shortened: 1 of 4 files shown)


eviltrout:
This really needs a test

riking:
Yep, seeing as I typoed the site setting name.

app/controllers/users_controller.rb
```diff
@@ -231,6 +231,10 @@ def create
       return fail_with("login.password_too_long")
     end
 
+    if SiteSetting.reserved_usernames.include? params[:username].downcase
+      return fail_with("login.reserved_username")
+    end
+
```
nlalonde: This should be validated in the model. See username_validator in user.rb.

riking: Nope, that will add problems for the existing @discourse account created on the hosted sites, and you wouldn't ever be able to add a username to the default list without causing people problems.

coding-horror:
Agreed we should pull this in

eviltrout:
Bringing it in.

PR #3586, opened 2015-07-02
Strip insignificant zeros from formatted numbers

gschlager:
https://meta.discourse.org/t/1-2k-looks-good-but-1-0k-looks-strange/30702


coding-horror:
https://meta.discourse.org/t/1-2k-looks-good-but-1-0k-looks-strange/30702/2?u=codinghorror

I just don't agree with this request, apologies.

PR #3587, opened 2015-07-02
FIX: Emoji not shown in notifications.

tgxworld:
Fixes: https://meta.discourse.org/t/topic-title-emoji-not-supported-in-notifications/30670/2.

Do let me know if this is a right fix :smile: Thanks!


SamSaffron:
looks good, probably should add a test though :+1:

tgxworld:
@SamSaffron Which parts should be tested? I usually prefer integration tests over unit tests but I'm not quite sure what the usual practice is here. :smile_cat:  Thanks in advance!

PR #3588, opened 2015-07-02
FEATURE: "Enabled" column in /admin/plugins

riking:
Also, added enabled_site_setting to the Poll plugin so it shows up properly.

### DIFF
--- a/app/assets/javascripts/admin/routes/admin-plugins.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-plugins.js.es6
@@ -4,8 +4,21 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    showSettings() {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/admin/routes/admin-plugins.js.es6
```diff
@@ -4,8 +4,21 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    showSettings() {
-      this.transitionTo('adminSiteSettingsCategory', 'plugins');
+    showSettings(plugin) {
+      const controller = this.controllerFor('adminSiteSettings');
+      this.transitionTo('adminSiteSettingsCategory', 'plugins').then(function() {
```
riking: It transitions to this first because if the site settings have never loaded, calling filterContentNow() first will fail - it doesn't have the list.

PR #3591, opened 2015-07-06
FIX: Don't root CSV export URLs

riking:
Should fix https://meta.discourse.org/t/data-exports-wrong-redirect/30774 by making the link start with `/` instead of `//`.


SamSaffron:
@techAPJ can you review

techAPJ:
Looks good!

PR #3592, opened 2015-07-07
FIX: Substrings of reserved usernames no longer treated as reserved

Xenograph:
I'm assuming that's not intended functionality. For some background, it looks like SiteSetting.reserved_usernames is a string of usernames separated by "|". The current code checks incoming usernames against being a substring of SiteSetting.reserved_usernames. This causes, for example, "strator" to be considered reserved since it is a substring of "administrator". The fix is pretty straightforward, I simply tokenized the string into an array of reserved username strings.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -231,7 +231,7 @@ class UsersController < ApplicationController
       return fail_with("login.password_too_long")
     end
 
-    if SiteSetting.reserved_usernames.include? params[:username].downcase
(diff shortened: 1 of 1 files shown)


riking:
d'oh! I thought for some reason we already had the automagic turning that into an array in advance, which is not the case. LGTM.

coding-horror:
Are there tests for this?

eviltrout:
Yes but the tests didn't include substrings. It would be good if we added one of that case.

PR #3593, opened 2015-07-08
Cache results on about page

tgxworld:
Fixes: https://meta.discourse.org/t/about-page-slow-menu-slow-to-disapear/24635

cc/ @SamSaffron

### DIFF
--- a/app/jobs/scheduled/dashboard_stats.rb
+++ b/app/jobs/scheduled/dashboard_stats.rb
@@ -1,16 +1,13 @@
 module Jobs
   class DashboardStats < Jobs::Scheduled
+    include Jobs::Stats
+
     every 30.minutes
 
     def execute(args)
-      stats = AdminDashboardData.fetch_stats.as_json
-
-      # Add some extra time to the expiry so that the next job run has plenty of time to
-      # finish before previous cached value expires.
-      $redis.setex AdminDashboardData.stats_cache_key, (AdminDashboardData.recalculate_interval + 5).minutes, stats.to_json
-
+      stats = AdminDashboardData.new.as_json
+      set_cache(AdminDashboardData, stats)
       stats
     end
-
   end
 end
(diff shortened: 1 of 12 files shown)


app/jobs/scheduled/dashboard_stats.rb
```diff
@@ -1,16 +1,7 @@
...
-    def execute(args)
-      stats = AdminDashboardData.fetch_stats.as_json
-
-      # Add some extra time to the expiry so that the next job run has plenty of time to
-      # finish before previous cached value expires.
-      $redis.setex AdminDashboardData.stats_cache_key, (AdminDashboardData.recalculate_interval + 5).minutes, stats.to_json
-
-      stats
-    end
+    include Jobs::Stats
```
SamSaffron: how is this working, is args somehow AdminDashboardData ?

tgxworld: Opps I see what you mean. I didn't realize that the job wouldn't receive the right argument when ran in the background.  I'll see if there is a way to pass it any argument. Otherwise, I'll extract the logic for `$redis.setex` into a different method which can be reused.

riking: yes, right here: https://github.com/discourse/discourse/pull/3593/files#diff-b5204f67764bba61319900088604de81R4

tgxworld: @riking But it wouldn't get any args when it is run every 30 mins. :smiley_cat:

tgxworld:
@SamSaffron Updated. :smile:

SamSaffron:
my concern with this implementation is that it can lead to some confusing situations. 

EG: 

make someone admin -> go to about page ... user is missing from admin 

I think the caching needs to be more granular

tgxworld:
> EG:
> 
> make someone admin -> go to about page ... user is missing from admin

We're only caching the stats in the table so the above scenario would still work. The PERF on the about page is because we're hitting the DB 14 times for the different count so I'm not too sure if we would want to cache the admins on the page. Not sure if I'm missing something here :smiley_cat:

SamSaffron:
this is good :+1: we just need a test here

tgxworld:
@SamSaffron Done :smile:

SamSaffron:
cool, lets pull it in then :+1:

PR #3594, opened 2015-07-08
FIX: admin settings page for small screens

scossar:
Improves the display of the admin settings page for small screens.

Adds 2 breakpoints, one for mobile, one for tablet. Values are set at the top of the file. The mobile breakpoint could probably be reduced.

Switches the order of the text-field and checkbox inputs in `site-setting.hbs` so that they display in the correct order when vertically stacked on small screens.

Sets widths of elements as percentage of screen width. The actual values could be adjusted. I've tried to make it readable on all devices.

The admin pages for Users, logs and plugins should probably have minimum widths set on their table elements. That would force users on phones to scroll to read the table, but it would keep them readable.

Ideally, the admin-nav should be an off-screen slide-out menu.


SamSaffron:
@awesomerobot what do you think about this? is there a cleaner way to take care of this?

coding-horror:
@scossar could you resolve conflicts / rebase?

scossar:
I've deleted the local branch to this in a way that doesn't make it very easy to recover. (I'm just learning some of this stuff.) The changes here can be simplified quite a bit. I'll send a new request later today or early tomorrow.

PR #3595, opened 2015-07-08
UX: Add printer-friendly.scss to make short topics print nicely.

cpradio:
UI
image

Print
faq_guidelines - staff - discourse

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -8,4 +8,5 @@
 @import "common/components/*";
 @import "common/input_tip";
 @import "common/topic-entrance";
+@import "common/printer-friendly";
 @import "common/base/*";
--- a/app/assets/stylesheets/common/printer-friendly.scss
+++ b/app/assets/stylesheets/common/printer-friendly.scss
@@ -1,0 +1,79 @@
+@media print {
+  /* remove items N/A for print
+  ** visual clutter etc. */
+  .avatar,
+  .topic-avatar,
+  header,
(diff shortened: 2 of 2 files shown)


riking:
Don't bother targeting `.ember-view` in any CSS rules - it's useless and might cause your CSS to break later.

cpradio:
Thanks @riking, references to ember-view have been removed/replaced with class names.

SamSaffron:
I thought the correct way of doing print is adding an extra stylesheet

eg:

```
<link rel="stylesheet" href="print.css" type="text/css" media="print" />
```

that way we don't load any of this stuff for desktop.

coding-horror:
yeah I'd pull this into the media="print" stylesheet

cpradio:
@SamSaffron, last I knew that was incorrect. The browser downloads both regardless, it just may defer the download by a few seconds.
https://zoompf.com/blog/2009/12/browser-performance-problem-with-css-print-media-type

Using a media query reduces the load by 1 HTTP request.

cpradio:
I also proved it works the same on Mobile. It will download both (from what I can tell), the media query approach lessens the extra HTTP request and it honors the print styling.

cpradio:
Okay, I talked with a few CSS gurus, Paul O'Brien being one of them and did a lot of article searching. In short, everyone says to use media queries over a separate print stylesheet.

A List Apart, uses the print media query for their printing needs and I believe everyone holds them up in high regards. I seriously believe the way I handled this, to be the best method for Discourse.

SamSaffron:
ok, its such a small amount of extra rules, lets merge it.

PR #3596, opened 2015-07-09
FIX: Automatically add like-count to post menu

riking:
https://meta.discourse.org/t/missing-all-the-likes-on-every-post-after-upgrading-to-latest-version-53b0b70/30898/


SamSaffron:
I want to merge this in but it should be done properly... 

all that is needed here is a single update statement

update site_settings set value = replace(value, 'like', 'like-count|like') where name = 'post_menu' and value not like '%like-count%'

or something like that

riking:
Updated and tested.

SamSaffron:
perfect

PR #3597, opened 2015-07-09
FEATURE: functional admin pages for mobile phone

scossar:
- Adds a slide-out menu to the admin settings page
- Removes admin-nav from admin plugins page
- sets a min-width on admin logs table

This works quite well. If you are interested in using it I can do some work to clean it up and work through the different admin pages. It is set with a large value for the mobile-breakpoint. It could be reduced to around 600px.

It doesn't use touch events for the menu,

The menu closes automatically when a new link is selected. It is hooked into the didTransition action - that doesn't work if the same link is selected. That needs to be fixed. I'm just learning ember.

I think a good standard for this is the mobile view for WordPress admin.

screenshot 2015-07-09 14 57 12

screenshot 2015-07-09 14 57 27

screenshot 2015-07-09 14 57 46


scossar:
I've removed the 'only show overridden' checkbox and the 8px of padding around the content - only for very small screens. The result looks a lot better.

screenshot 2015-07-08 22 38 33

screenshot 2015-07-08 22 38 46

SamSaffron:
@coding-horror this is quite a big change, how do you feel about it? 

@scossar this needs a rebase

@eviltrout  can you review the ember parts?

scossar:
If you are interested in something like this I will redo the work and improve on it a bit.

coding-horror:
I think it is fine, we have long said that admin is not really designed for mobile so have spent no time here.. as long as the risk is low, why not?

SamSaffron:
@scossar feel free to work on this, we want to include it :+1:

scossar:
I am still working on this. I don't have much time this week. I will try to finish it early next week.

coding-horror:
Is this superseded by the other PR I just merged?

scossar:
Yes, it would be better to just build on top of the other PR.

PR #3599, opened 2015-07-09
Add ace:resize app event

riking:
Needed for data-explorer to resize the editor window

### DIFF
--- a/app/assets/javascripts/admin/components/ace-editor.js.es6
+++ b/app/assets/javascripts/admin/components/ace-editor.js.es6
@@ -26,7 +26,14 @@ export default Ember.Component.extend({
       this._editor.destroy();
       this._editor = null;
     }
+    this.appEvents.off('ace:resize', this, this.resize);
   }.on('willDestroyElement'),
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/admin/components/ace-editor.js.es6
```diff
@@ -6,6 +6,8 @@ export default Ember.Component.extend({
   classNames: ['ace-wrapper'],
   _editor: null,
   _skipContentChangeEvent: null,
+  _unusedStashSelf: null,
+  stashSelf: Em.computed.alias('_unusedStashSelf'),
```
eviltrout: Why are you stashing self into a property? It seems really odd to me.

Also besides setting it to a  self and later null it seems unused?

riking: I was using it to have a way to call `aceComponent.trigger('resize')` from the view:admin-plugins-explorer. Is there a better way to do that?

SamSaffron:
qunit tests are failing

PR #3600, opened 2015-07-10
FIX: Clicking 'show more' should not dismiss search dropdown.

tgxworld:
cc/ @SamSaffron


tgxworld:
O wait... if we're not searching within a topic. Should clicking on `show more` dismiss the dropdown?

SamSaffron:
if we are not within a topic it should dismiss (it also is a standard link)
unless we click show more on category / user.

tgxworld:
@SamSaffron That should fix it. I did come across another bug though (I didn't break that :stuck_out_tongue_closed_eyes: ), if a dropdown is active and we click on a link outside of the dropdown if doesn't dismiss.

SamSaffron:
thanks, yeah looks good

PR #3602, opened 2015-07-12
Fix LocalStore.remove_file to not raise if source doesn't exist

lukegb:
FileUtils.move actually ends up raising an "unknown file type" error if the file doesn't exist instead of Errno::ENOENT.

It's possible to rescue this, but in the end it's easier to just ask move not to throw an error, since we're going to throw it away anyway.


SamSaffron:
@ZogStriP i think this is ok, should we merge?

lukegb:
I think as you mentioned in the line comment is might be worth investigating whether the `rescue` block is even necessary.

ZogStriP:
Thanks.

PR #3605, opened 2015-07-15
FEATURE: Automatically copy the share link if possible

riking:
With support for execCommand(copy) in most browsers now, we can add this to the share dialog. If the copy works, small green text that says "copied!" will fade in.

image


coding-horror:
This is really cool but I think the copy should be a tiny bit more verbose
(this is by definition desktop) and say "copied to clipboard!"

SamSaffron:
cool, while there one huge PITA is that the share dialog renders superimposed on the link, makes its particularly terrible on mobile, should render above.

PR #3607, opened 2015-07-15
FIX: Encode percent character in search terms.

tgxworld:
Fixes: https://meta.discourse.org/t/searching-gives-an-error-while-trying-to-load-page/30896

### DIFF
--- a/test/javascripts/adapaters/topic-list-test.js.es6
+++ b/test/javascripts/adapaters/topic-list-test.js.es6
@@ -1,0 +1,13 @@
+module("adapter:topic-list");
+
+import { finderFor } from 'discourse/adapters/topic-list';
+
+test("finderFor", function() {
+  // Mocking instead of using a pretender which decodes the path and thus does
+  // not reflect the behavior of an actual web server.
+  var mock = sandbox.mock(Discourse);
+  mock.expects("ajax").withArgs("/search.json?q=test%25%25");
+  var finderForFunction = finderFor('search', { q: "test%%" });
+  finderForFunction();
+  mock.verify();
+});
--- a/app/assets/javascripts/discourse/adapters/topic-list.js.es6
+++ b/app/assets/javascripts/discourse/adapters/topic-list.js.es6
@@ -1,6 +1,6 @@
 import RestAdapter from 'discourse/adapters/rest';
 
-function finderFor(filter, params) {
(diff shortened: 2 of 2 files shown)


tgxworld:
@eviltrout Could you review this? Thank you :smile:

test/javascripts/adapaters/topic-list-test.js.es6
```diff
@@ -0,0 +1,11 @@
+module("adapter:topic-list");
+
+import { finderFor } from 'discourse/adapters/topic-list';
+
+test("finderFor", function() {
+  var mock = sandbox.mock(Discourse);
+  mock.expects("ajax").withArgs("/search.json?q=test%25%25");
```
eviltrout: It would be better if you used pretender to mock the AJAX calls. I don't like using sinon mocks for this purpose.

You should be able to add this to `create-pretender` and have it intercepted automatically in test mode.

tgxworld: Ahhh I tried out pretender but it presents a different set of problems. I'll try to summarize what I've done so far. :smile_cat: 

### Problem
Malformed URI error raised by the server because we are passing it a URI that it cannot parse. Example: `/search.json?q=test%`

### What we need to test here
We need to add a test to ensure that the URL we passed to the ajax call encodes the `%` character. In other words, I'm only concerned about the parameter that `Discourse.ajax` receives.

### Problem faced when using pretender
We use `route-recognizer` to identify the route to match to in the pretender. However, [`route-recognizer` decodes the path](https://github.com/discourse/discourse/blob/master/vendor/assets/javascripts/route-recognizer.js#L622) which isn't what the actual server would do. Otherwise, I would have tried something like this.

```javascript
var params;
new Pretender(function() {
  this.get('/search.json', function(request) {
    params = request.queryParams();
  });
}); 
equal(params.q, 'test%25')
```

eviltrout: Wow thanks for going above and beyond on this one. I think in this case the mock is fair, but could you add a comment explaining why it has to be a mock? Just so someone doesn't come in and try to refactor it and go through the same pain?

tgxworld:
@eviltrout Updated :)

SamSaffron:
cool, thanks !!

PR #3609, opened 2015-07-15
FEATURE: Localization fallbacks

riking:
This patch will (unconditionally) perform localization fallback from the user locale to the SiteSetting.default_locale to english.

It is only enabled in production.

I'm also not quite sure how to test this - ideally, there are no missing translations, so no matter which translation key is chosen, it might eventually get translated.

### DIFF
--- a/lib/js_locale_helper.rb
+++ b/lib/js_locale_helper.rb
@@ -55,15 +55,14 @@ module JsLocaleHelper
   def self.load_translations_merged(*locales)
     @loaded_merges ||= {}
     @loaded_merges[locales.join('-')] ||= begin
-      # TODO - this will need to be reworked to support N fallbacks in the future
-      all_translations = locales.map { |l| JsLocaleHelper.load_translations l }
+      all_translations = {}
       merged_translations = {}
-      merged_translations[locales[0].to_s] = all_translations[0][locales[0].to_s]
-      if locales[1]
-        merged_translations[locales[1].to_s] = deep_delete_matches(all_translations[1][locales[1].to_s].dup, merged_translations[locales[0].to_s])
-      end
-      if locales[2]
-        merged_translations[locales[2].to_s] = deep_delete_matches(all_translations[2][locales[2].to_s].dup, merged_translations[locales[0].to_s], merged_translations[locales[1].to_s])
+      loaded_locales = []
+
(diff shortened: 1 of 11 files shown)


riking:
Benchmark says that the memoization should improve asset compilation times by about 0.75 seconds, despite the perf hit from combining locales.

```
kane@kane-TECRA-Z40-B:~/projects/discourse$ git checkout patch-7 && bundle exec ruby ~/projects/readygo/readygo.rb --compare script/bench/js_locale_helper_bench.rb 
................

JsLocaleHelper generating files
  Baseline: |                                                              X---|
  Current:  |                              X-                                  |
            0                                                            1.589 s
```

xfalcox:
This is awesome and really needed. How hard is to implement a custom fallback order?

Like this:

- User locale = server locale = pt-br
- Best fallback possible: pt_BR -> pt_PT -> en_US

Maybe as a list on conf file?

SamSaffron:
@eviltrout both me and you need to review this, can you try to look through it tomorrow... I will review as well

SamSaffron:
@riking I can not see a test that shows keeps working as it used to in dev. we still need the strings to help translators... also how do you disable this in prd?

lib/js_locale_helper.rb
```diff
@@ -1,30 +1,94 @@
...
+      plugin_translations = {}
+      Dir["#{Rails.root}/plugins/*/config/locales/client.#{locale_str}.yml"].each do |file|
+        plugin_translations.deep_merge! YAML::load(File.open(file))
+      end
+
+      # merge translations (plugin translations overwrite default translations)
+      translations[locale_str]['js'].deep_merge!(plugin_translations[locale_str]['js']) if translations[locale_str] && plugin_translations[locale_str] && plugin_translations[locale_str]['js']
+
+      # We used to split the admin versus the client side, but it's much simpler to just
+      # include both for now due to the small size of the admin section.
```
eviltrout: Wait, does this mean even in production users will get the admin section even as regular users? I'm not sure if I support that. What is the size difference?

riking: Hey, I didn't even notice that. Looks like the `admin_js` section starts at the 63% mark, line-wise.

eviltrout:
Merging this in optimistically, let's keep an eye on it.

PR #3611, opened 2015-07-15
FIX: Allow to change ownership on deleted users' posts

riking:


SamSaffron:
sure .. lets try it out

PR #3613, opened 2015-07-17
FIX: typo in rtl.scss

scossar:
Changes `text-aligh: left` to `text-align: left`. Adds a comment explaining how the !important declaration prevents a rule from being flipped.


SamSaffron:
thanks!

PR #3614, opened 2015-07-17
Run main tests on Travis first.

tgxworld:
Inspired by https://github.com/rails/rails/pull/17246/files.


tgxworld:
Ah ok there isn't a benefit here unless we have more parallel workers,

SamSaffron:
sure, seems fine, thanks

PR #3616, opened 2015-07-19
FIX: Add padding to container

scossar:
At screen widths between 850px and 1005px there is no left margin on the page. There are a few possible fixes. Here is one.

- puts `.list-controls` and `#topic-title` inside of `.container`
- removes padding from `#topic-title`, `.container.posts`, and `.list-controls`
- adds 10px padding to `.container`

Doing it this way allows for the option of setting  separate backgrounds  on the html and `.container` elements. Having padding as a property of the container makes sense. You can control all the padding in one place and everything will line up. 

screenshot 2015-07-19 11 08 17


scossar:
I've rethought this a bit. I think there is a more straightforward way to approach it.

PR #3618, opened 2015-07-20
FIX: padding and alignment issues

scossar:
This adds the class `.wrap` to the `#main-outlet` div. It serves to wrap all content. It defines a width, a margin relative to the browser window, and padding relative to the application content. It is set with a consistent padding left and right of 8px.  This can be adjusted. The `.container` class is redefined to just extend `.clearfix`. 

This relates to this topic on Discourse: https://meta.discourse.org/t/borders-have-a-unsweet-spot-of-window-size/31049

screenshot 2015-07-20 01 09 40

### DIFF
--- a/app/assets/javascripts/discourse/templates/application.hbs
+++ b/app/assets/javascripts/discourse/templates/application.hbs
@@ -1,6 +1,6 @@
 {{render "header"}}
 
-<div id='main-outlet'>
+<div id='main-outlet' class='wrap'>
   {{outlet}}
   {{render "user-card"}}
 </div>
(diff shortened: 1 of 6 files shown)


SamSaffron:
is this good on mobile too?

scossar:
Yes it is. It just creates a wrapper div around all the content. If the
padding on the wrap div is set to 0 for mobile, then everything should be
exactly the same as it was before. If you would like to use the wrap div to
set the outer padding (the space between the edge of the content and the
edge of the screen), then some padding and margins will have to be adjusted
for mobile.
I'll look at this again tomorrow and make sure it's working the way it's
supposed to.

SamSaffron:
@awesomerobot can you confirm this is good so I can merge?

SamSaffron:
thanks :+1:

scossar:
Great. Except there is now a bit too much padding on mobile. A quick fix is
to set .wrap {
padding: 0; } in mobile/discourse.scss. A better fix is to leave the
padding on .wrap and remove the margins  from the elements that line up
with it. I'll try to send a PR for that in the next couple of hours.

PR #3619, opened 2015-07-20
Translation for admin.permalink

Vernam:
French translation for the permalink


SamSaffron:
This should come via trasifex, not in a PR

PR #3620, opened 2015-07-21
FIX: Do automatic copy on share in click handler

riking:
I think this should fix it!

https://meta.discourse.org/t/it-says-copied-to-clipboard-but-there-is-nothing-in-my-clipboard/31244/4


SamSaffron:
lets hope it does :)

PR #3621, opened 2015-07-21
FIX: Twitter onebox styling

riking:
A previous patch added white-space: pre-wrap; but forgot to make sure the beginnings of lines were straight! This fixes that by making the tweet body a floating element so it doesn't wrap on the avatar.


SamSaffron:
anything that removes !important fills me with delight!

PR #3622, opened 2015-07-21
FIX: Provide proper source map URLs with a CDN

riking:
I tested this on my forum and it didn't break the assets compile, and I tested locally with what I think is a valid DISCOURSE_CDN_URL, but I can't tell if it's *really* correct.

Also, `load lib/global_path.rb` is pretty dang ugly.


SamSaffron:
cool, lets try it

PR #3623, opened 2015-07-21
FIX: Change post owner of deleted post, take 3

riking:
Works On My Machine™

post.recover! was trying to create TopicLink objects, but they had null users


SamSaffron:
lets see ...

PR #3625, opened 2015-07-22
FIX: Remove unnecessary padding and margins from mobile view

scossar:
Lines up all content in mobile view with padding set on `.wrap`.

Alignment of elements in mobile-view was previously set individually on each element with margins and padding. This PR removes left and right padding from elements that span the full-screen width in the mobile-view. They are all now aligned with the padding set on `.wrap`.

Instead of making these changes, it would also be possible to set the left and right padding on `.wrap`  to 0 for the mobile view.


coding-horror:
These are really great changes, thank you! :gift:

PR #3629, opened 2015-07-27
FIX: Let users open header links in new tab without dismissing dropdown.

tgxworld:
Fixes: https://meta.discourse.org/t/ctrl-clicking-a-notification-closes-the-notifications-menu/31463/4


ZogStriP:
Thanks :+1:

PR #3632, opened 2015-07-27
Fixes some minor issues with the phpBB3 importer

gschlager:
FEATURE: Adds a link to the Howto on meta
FIX: Suppresses multiple warnings about invalid votes for the same topic
FIX: Adds missing table prefixes
FIX: Removes duplicate column from query


ZogStriP:
Thanks :+1:

PR #3633, opened 2015-07-28
FEATURE: Add CommentMigration for db column comments

riking:
For documenting the database (will be used in discourse-data-explorer).

I didn't put in too many, mostly want to get a few in there as an example.

### DIFF
--- a/db/migrate/20150727230537_add_example_column_comments.rb
+++ b/db/migrate/20150727230537_add_example_column_comments.rb
@@ -1,0 +1,26 @@
+require 'comment_migration'
+
+class AddExampleColumnComments < CommentMigration
+
+  def comments_up
+    {
+      posts: {
+        _table: 'If you want to query public posts only, use the badge_posts view.',
+        post_number: 'The position of this post in the topic. The pair (topic_id, post_number) forms a natural key on the posts table.',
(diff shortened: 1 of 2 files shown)


db/migrate/20150727230537_add_example_column_comments.rb
```diff
@@ -0,0 +1,26 @@
...
+      posts: {
+        _table: 'Every post on the forums. If you want to query public posts only, use the badge_posts view.',
+        post_number: 'The position of this post in the topic. The pair (topic_id, post_number) forms a natural key on the posts table.',
+        raw: 'The raw Markdown that the user entered into the composer.',
+        cooked: 'The processed HTML that is presented in a topic.',
+        reply_to_post_number: "If this post is a reply to another, this column is the post_number of the post it's replying to. [FKEY posts.topic_id, posts.post_number]",
+        reply_quoted: 'This column is true if the post contains a quote-reply.',
+      },
+      topics: {
+        _table: "Every topic on the forums. To query public topics only: SELECT ... FROM topics t LEFT INNER JOIN categories c ON (t.category_id = c.id AND c.read_restricted = false)"
```
eviltrout: Every topic on the forum seems a bit redundant?

Same with "Every post on the forum".

What else would "posts" and "topics" be?

PR #3634, opened 2015-07-28
FIX: Draft overlaps topic counts blurb in suggested topics.

tgxworld:
fixes: https://meta.discourse.org/t/post-draft-in-progress-overlaps-unread-new-counts-at-topic-end/26659


SamSaffron:
yeah ... this should work lets try it

PR #3635, opened 2015-07-28
Create views 'likes', 'bookmarks', 'flags'

riking:
Splitting the post actions table - it adds an updatable view, so new code and old code will be seeing the same things.

I have verified with the SQL console that the likes and bookmarks views can be INSERTed into.


riking:
recreating on different branch

PR #3636, opened 2015-07-30
FEATURE: Add off button on preferences for popup notifications

riking:


riking:
image
image
image
image
image

config/locales/client.en.yml
```diff
@@ -413,6 +413,17 @@ en:
...
+      desktop_notifications:
+        label: "Desktop Notifications"
+        not_supported: "Notifications are not supported on this browser. Sorry."
+        perm_default: "Turn On Notifications"
+        perm_denied_btn: "Permission Denied"
+        perm_denied_expl: "You have denied permission for notifications. Use your browser to enable notifications, then click the button when done. (Desktop: The leftmost icon in the address bar. Mobile: 'Site Info'.)"
+        disable: "Disable Notifications"
+        currently_enabled: "(Currently Enabled)"
+        enable: "Enable Notifications"
+        currently_disabled: "(Currently Disabled)"
```
eviltrout: minor, but could you make this lower case? (currently disabled) -- title case should be used for titles.

app/assets/javascripts/discourse/components/desktop-notification-config.js.es6
```diff
@@ -0,0 +1,61 @@
...
+  classNames: ['controls'],
+
+  _version: 0,
+
+  _init: function() {
+    this.invalidate();
+  }.on('init'),
+
+  invalidate() {
+    this.set('_version', this.get('_version') + 1);
```
eviltrout: I am not a huge fan of the `invalidate` method here.

I think a better approach would be to add a computed property for `Notification.permission`. Have your `isDeniedPermission`, `isDefaultPermission` etc depend on that new property.

Then instead of calling `invalidate` call `propertyDidChange` when you need to look at it.

riking: Oh, `propertyDidChange` is the thing I was missing. OK! Will get that fixed when I get home

eviltrout:
Thanks for fixing it!

riking:
No fix here - this is all added features ;)

PR #3637, opened 2015-08-01
UX: Social login buttons alignment off on mobile.

tgxworld:
## Before

screenshot from 2015-08-01 11 50 34

screenshot from 2015-08-01 11 55 50

<hr>

## Current (Alignment off on chrome but fine on firefox)
screenshot from 2015-08-01 12 04 54

screenshot from 2015-08-01 12 04 44

<hr>

## With Fix
screenshot from 2015-08-01 12 10 12
screenshot from 2015-08-01 12 10 03

cc/ @SamSaffron I'm not sure if the increase height of the buttons was intended so I reverted back to the previous height.

### DIFF
--- a/app/assets/stylesheets/common/components/buttons.css.scss
+++ b/app/assets/stylesheets/common/components/buttons.css.scss
@@ -128,8 +128,7 @@
   &:before {
     margin-right: 9px;
     font-family: FontAwesome;
-    line-height: 1.6em;
-    font-size: 1.3em;
+    font-size: 17px;
   }
   &.google, &.google_oauth2 {
     background: $google;


tgxworld:
Hmm looks fine on firefox... I'll investogate

riking:
> I'm not sure if the increase height of the buttons was intended so I reverted back to the previous height.

We do want the taller buttons.

SamSaffron:
I do want easier to click buttons, but will merge this for now

tgxworld:
This might be interesting.....I just happened to use an iPhone 6s and the alignment wasn't off on safari mobile. might be a chrome issue so I'll look into this further

PR #3638, opened 2015-08-01
FIX: admin settings page for mobile screens

scossar:
- Sets the box-model to `border-box` for the `.admin-content` div for screen sizes up to the mobile-breakpoint. This makes it a lot easier to add padding, margins, and borders to elements without throwing off the layout.

- Fixes the admin settings page.

- Tweaks the 'logs' and custom-color-scheme pages just enough to make them workable.

It would be quite easy to move from here to a a more complete fix for mobile-admin. I've done most of the work for that already.

This could be done without changing the box model but it would be a little bit more complicated.

### DIFF
--- a/app/assets/stylesheets/common/admin/admin_base.scss
+++ b/app/assets/stylesheets/common/admin/admin_base.scss
@@ -3,6 +3,24 @@
 @import "common/foundation/mixins";
 @import "common/foundation/helpers";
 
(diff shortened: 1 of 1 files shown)


coding-horror:
Ok so this change only affects admin, yes?

scossar:
yes

coding-horror:
Let's give it a shot then...

PR #3639, opened 2015-08-03
FEATURE: Show "load more" button on tall screens

riking:
Simple workaround for https://meta.discourse.org/t/infiniscroll-not-working-if-the-last-item-of-first-batch-is-not-off-screen-part-deux/23596/12

image


SamSaffron:
why not just trigger another batch if its not on screen?

riking:
Actually wasn't as hard as I thought to do that... there we go.

ZogStriP:
:+1:

PR #3641, opened 2015-08-04
adding flexbox to header via mixins (all prefixes)

awesomerobot:
forgot safari previously; tested on iOS :+1:


coding-horror:
thanks let's give this a try! does tagging need to change as well?

PR #3642, opened 2015-08-05
Split post_actions table into views

riking:
Splitting the post actions table. Will make some of the refactorings in the future easier.

 - Adds an updatable view, so you can create a Like object and save it (but we don't, because PostAction.act is the creation method)
 - Uses ActiveRecord single-table inheritance on the post_action_type_id column
 - Flag-only class methods on PostAction were moved & forwarded

I have verified with the SQL console that the likes and bookmarks views can be INSERTed into.

~~Tests pass, but no new tests were added.~~ Added tests for the subclassing.

### DIFF
--- a/app/models/like.rb
+++ b/app/models/like.rb
@@ -1,0 +1,84 @@
+
+class Like < PostAction
+
+  self.table_name = :likes
+
+  def self.pa_type
+    PostActionType.like
+  end
+
+  def pa_type
+    self.class.pa_type
+  end
+
+  # === BEGIN CRAZY SUBCLASSING MADNESS === #
+
+  def self.sti_name
+    pa_type
+  end
+
+  def self.type_condition(table = arel_table)
+    table[:post_action_type_id].in([pa_type])
+  end
+
+  before_save do
+    self.post_action_type_id = 2
+  end
+
+  # === END CRAZY SUBCLASSING MADNESS === #
+
+  before_create do
+    raise AlreadyActed if PostAction.where(user_id: user_id)
+                            .where(post_id: post_id)
+                            .where(post_action_type_id: pa_type)
+                            .where(deleted_at: nil)
+                            .exists?
+  end
+
+  def self.lookup_for(user, topics)
+    PostAction.lookup_for(user, topics, pa_type)
+  end
+
+  def self.count_per_day(opts=nil)
+    PostAction.count_per_day(pa_type, opts)
+  end
+
+  def self.act(user, post, opts = {})
+    PostAction.act(user, post, pa_type, opts)
+  end
+
+  def self.remove_act(user, post)
+    PostAction.remove_act(user, post, pa_type)
+  end
+
+  def is_bookmark?
+    false
+  end
+
+  def is_like?
+    true
+  end
+
+  def is_flag?
+    false
+  end
+
+  def is_private_message?
+    false
+  end
+
+end
+
+# == Schema Information
+#
+# Table name: likes
+#
+#  id                  :integer          primary key
+#  post_id             :integer
+#  user_id             :integer
+#  created_at          :datetime
+#  updated_at          :datetime
+#  deleted_by_id       :integer
+#  deleted_at          :datetime
+#  post_action_type_id :integer
+#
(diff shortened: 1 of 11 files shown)


app/models/like.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def self.pa_type
+    PostActionType.like
+  end
+
+  def pa_type
+    self.class.pa_type
+  end
+
+  # === BEGIN CRAZY SUBCLASSING MADNESS === #
```
eviltrout: Firstly I don't think we need to comment this way. I don't think it adds much to the code to just call it crazy. Either explain why something is done or let's leave out the comment altogether.

Secondly, can't this come from PostAction? It seems for Bookmark a lot of the code is shared. If we can't inherit it because it's class methods, how about using a mixin to avoid repeating ourselves?

db/migrate/20150728202123_create_post_action_views.rb
```diff
@@ -0,0 +1,31 @@
...
+    execute <<SQL
+CREATE VIEW bookmarks AS
+SELECT id, post_id, user_id, created_at, updated_at, deleted_by_id, deleted_at, post_action_type_id
+FROM post_actions
+WHERE post_action_type_id = 1
+SQL
+
+    execute <<SQL
+CREATE VIEW flags AS
+SELECT *
```
eviltrout: I prefer it to be explicit rather than `SELECT *`

eviltrout:
I added a couple of comments. I would also like to see if this affects performance in any way:

- Are there any more queries when viewing a topic or performing an action on a post?
- If so, is there a difference in ms for this? is it better? worse?

riking:
> Are there any more queries when viewing a topic or performing an action on a post?
> If so, is there a difference in ms for this? is it better? worse?

According to mini-profiler, this didn't change the queries when viewing a topic at all. That's the line I put a TODO on.

Performing an action:

 - The uniqueness check is performed twice (oops)

```
-- master
SELECT  1 AS one FROM "post_actions"  WHERE ("post_actions"."deleted_at" IS NULL) AND "post_actions"."user_id" = 5696 AND "post_actions"."post_id" = 95688 AND "post_actions"."post_action_type_id" = 2 AND "post_actions"."deleted_at" IS NULL AND "post_actions"."disagreed_at" IS NULL AND "post_actions"."targets_topic" = 'f' LIMIT 1   

-- after change
SELECT  1 AS one FROM "post_actions"  WHERE ("post_actions"."deleted_at" IS NULL) AND "post_actions"."user_id" = 5696 AND "post_actions"."post_id" = 95537 AND "post_actions"."post_action_type_id" = 2 AND "post_actions"."deleted_at" IS NULL AND "post_actions"."disagreed_at" IS NULL AND "post_actions"."targets_topic" IS NULL LIMIT 1   
;
SELECT  1 AS one FROM "post_actions"  WHERE ("post_actions"."deleted_at" IS NULL) AND "post_actions"."user_id" = 5696 AND "post_actions"."post_id" = 95537 AND "post_actions"."post_action_type_id" = 2 AND "post_actions"."deleted_at" IS NULL LIMIT 1
;
```

 - This **removes** the auto_close_if_threshold_reached query (a flag check) when **liking** a post
```
-- master
SELECT post_actions.user_id, COUNT(post_id) FROM "post_actions" INNER JOIN "posts" ON "posts"."id" = "post_actions"."post_id" AND ("posts"."deleted_at" IS NULL) WHERE ("post_actions"."deleted_at" IS NULL) AND "post_actions"."disagreed_at" IS NULL AND "post_actions"."deferred_at" IS NULL AND "post_actions"."agreed_at" IS NULL AND "post_actions"."deleted_at" IS NULL AND "post_actions"."post_action_type_id" IN (3, 4, 7, 8) AND (posts.topic_id = 22517) AND ("post_actions"."user_id" != -1) GROUP BY post_actions.user_id

-- after change
```

SamSaffron:
This refactor is gigantic and I am struggling to find the tangible customer facing benefit we would get from it, it is also high risk, I worry about perf. I also worry about dynamically registered post actions. I get the desire to split stuff up but it just scares me a lot to make such a huge change to the code base.

I am shelfing this for now.

PR #3644, opened 2015-08-07
Adding two plugin outlets for notifications and their dropdowns

gtanczyk:
This change will be useful for Babble plugin: gdpelican/babble#14, so it is possible now to add another notification icon and its dropdown.


eviltrout:
Sounds good to me!

PR #3645, opened 2015-08-09
Added missing `meta_category_name` to Russian locale

MOZGIII:


ZogStriP:
I'm sorry but all translations must be done via [transifex](https://www.transifex.com/discourse/discourse-org/) otherwise they get overwritten when we pull the translations.

MOZGIII:
Got it.
Translated it via transiflex (https://www.transifex.com/discourse/discourse-org/viewstrings/#ru/serverenyml/20642552?key=meta_category_name), unfortunately, it seems like there is no progress taking the translation over to the code.
Should someone commit it or sth?

ZogStriP:
We manually pull translations from transifex about twice a week. I guess you'll have to wait a bit ;)

PR #3648, opened 2015-08-11
Adjust Brewfile to follow homebrew-bundle syntax

marlonandrade:
According to https://meta.discourse.org/t/brew-bundle-is-dead/23962
the official replacement for `brew bundle` is homebrew-bundle
(https://github.com/Homebrew/homebrew-bundle) and its syntax changed a
little bit.


SamSaffron:
sure ... lets try this out

PR #3649, opened 2015-08-13
Click reply should clear edit

tgxworld:
fixes: https://meta.discourse.org/t/reply-doesnt-clear-the-edit-post-text-if-clicked-after-edit/26012


tgxworld:
@eviltrout Could you review this?  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

SamSaffron:
hmm... what happens if you 

1. Start editing a post
2. Make some changes
3. Click "edit" on another post 

My expectation would be a "Are you sure you want to abandon your current edit"

tgxworld:
> hmm... what happens if you
> 
> Start editing a post
> Make some changes
> Click "edit" on another post
> My expectation would be a "Are you sure you want to abandon your current edit"

[This scenario is point number 3](https://meta.discourse.org/t/reply-doesnt-clear-the-edit-post-text-if-clicked-after-edit/26012/4?u=tgxworld) that I wasn't sure if it was a bug because the existing code is written to just ignore the click on edit if the Edit reply is dirty. I could update it to meet your expectations above.

SamSaffron:
nahh ignoring the click is not correct merging this in and giving it a good test

tgxworld:
@SamSaffron  Ok I'll submit a PR later to not ignore that click :smile_cat:

PR #3651, opened 2015-08-14
FIX: Broken private message search context.

tgxworld:
When using [ember-computed-decorators](https://github.com/rwjblue/ember-computed-decorators#real-world-getset-syntax), we don't have to specify the key.

cc @eviltrout


ZogStriP:
Thanks :+1:

PR #3653, opened 2015-08-14
Skip latest posts with no topic in rss

gdpelican:
For https://meta.discourse.org/t/posts-rss-returns-a-500-internal-server-error-actionview-template-error-undefined-method-title-for-nil-nilclass/32078/2


SamSaffron:
looks good to me :) thanks

PR #3655, opened 2015-08-14
Fix private message search

tgxworld:
meta: https://meta.discourse.org/t/pressing-return-when-searching-in-messages-returns-no-results/31949


SamSaffron:
sorry, sorted via 

https://github.com/discourse/discourse/commit/6086b07324d02f5d31a498790cf51cd081dbb74c

PR #3656, opened 2015-08-14
FIX: Wrong value set when using ember-computed-decorators.

tgxworld:
Follow up to https://github.com/discourse/discourse/pull/3651.

cc/ @ZogStriP @eviltrout opps found more instances.


ZogStriP:
Thanks :+1: 

cc @eviltrout

PR #3657, opened 2015-08-16
Lets see if this properly caches bundler on travis.

tgxworld:


tgxworld:
> [Bundler caching will not automatically work if you override the install step. You can instead use the arbitrary directory caching method described below:](http://docs.travis-ci.com/user/caching/#With-a-custom-install-step)

screenshot from 2015-08-17 01 15 38

Ok I think it worked maybe we could try this out?

SamSaffron:
sure, lets try it. :heart:

PR #3658, opened 2015-08-16
FEATURE: Add slide-out menu for small screens on admin settings page

scossar:
The menu is toggled with a button that is added to the admin-controls div. I haven't added any functionality to close the menu when a menu item is selected. I'm not sure how to best add an action to the link-to helper.

I have overridden the 10px padding that is set on `.wrap` by adding a -10px margin to the admin-contents div. Setting left and right padding on the outer `.wrap` div was probably a mistake (my mistake from a previous commit) - many containing elements for mobile need to extend to the edge of the screen.

screenshot 2015-08-16 11 11 13

screenshot 2015-08-16 11 12 04

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
@@ -64,6 +64,10 @@ export default Ember.ArrayController.extend({
         filter: '',
         onlyOverridden: false
       });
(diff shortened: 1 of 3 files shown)


SamSaffron:
interesting, so this is responsive based on width ... lets try it out,

SamSaffron:
thanks heaps :)

PR #3660, opened 2015-08-17
FIX: Rate limit message not shown.

tgxworld:


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -504,7 +504,9 @@ const Composer = RestModel.extend({
       return post.save(props).then(function(result) {
         self.clearState();
         return result;
-      }).catch(rollback);
```
tgxworld: @SamSaffron Was there a reason why you had to catch the error twice? 

Right now, the first catch is extracts the error and throws the error as a String. The second time we extract the error, the message gets [logger](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/lib/ajax-error.js.es6#L6-L8) and returns a generic error message

eviltrout:
This looks good to me.

PR #3661, opened 2015-08-17
Preserve user-field options when updating user-fields

b-studios:
**Problem**:
When a user field is changed (for instance by changing the position)
and the options are not retransmitted, the options of the user field
are deleted.

**Fix**:
Avoid deleting options of the user-field when no options are
transmitted.


eviltrout:
Thanks!

PR #3662, opened 2015-08-17
Use userfield serializer in json dump

b-studios:
**Problem:**
When anonymous and authentication is required, the json dump does not include user field options. This leads to the options not being shown in the registration form.

Also see discussion at [meta](https://meta.discourse.org/t/custom-user-field-enhancements/31572/4).

**Solution:**
Use userfield serializer for json dump to make sure that also the options are serialized correctly.

Warning: Please review this PR carefully to ensure no critical data is leaked.


eviltrout:
Thanks for fixing this!

PR #3663, opened 2015-08-17
FEATURE: Take advantage of Android browser features

riking:
Just tested this, and it pretty much looks good no matter what color you have in there.

Also provide the apple touch icon in a `<link rel=icon>`, because that's what android expects.


techAPJ:
:+1:

PR #3664, opened 2015-08-18
Use ajax-error in controller:topic.

tgxworld:


SamSaffron:
nice to have less code thanks!

PR #3666, opened 2015-08-18
FIX: Tighter rate-limit for post self-deletions

riking:
as per https://meta.discourse.org/t/minor-delete-undelete-needs-a-rate-limit/18027/10

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -184,6 +184,7 @@ class PostsController < ApplicationController
 
   def destroy
     post = find_post_from_params
+    RateLimiter.new(current_user, "delete_post", 3, 1.minute).performed! unless current_user.staff?
 
     if too_late_to(:delete_post, post)
       render json: {errors: [I18n.t('too_late_to_edit')]}, status: 422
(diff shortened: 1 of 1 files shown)


SamSaffron:
sure... seems fine

app/controllers/posts_controller.rb
```diff
@@ -184,6 +184,7 @@ def reply_history
 
   def destroy
     post = find_post_from_params
+    RateLimiter.new(current_user, "delete_post", 3, 1.minute).performed! unless current_user.staff?
```
tgxworld: I don't think this should be handled at the controller level but rather in [PostDestroyed#destroy](https://github.com/discourse/discourse/blob/master/lib/post_destroyer.rb). 

Also, I think we should have a SiteSettings for the cooldown time.

app/controllers/posts_controller.rb
```diff
@@ -206,6 +207,7 @@ def expand_embed
 
   def recover
     post = find_post_from_params
+    RateLimiter.new(current_user, "delete_post", 3, 1.minute).performed! unless current_user.staff?
```
tgxworld: I don't think we should set a rate limit for recover and instead allow users to recover immediately. If the `SiteSettings.post_edit_time_limit` happens to be smaller than 1 minute, I will end up with a scenario where I can no longer recover my post. WDYT? :smiley:

PR #3668, opened 2015-08-19
spoiler tag uses replaceBBCode

gwwar:
Allows the application of blur effects to both images and text in a single spoiler tag.

Fix for https://meta.discourse.org/t/spoiler-does-not-support-emoji/29902, this also requires https://github.com/discourse/discourse-spoiler-alert/pull/12 to work.

Initial:
screen shot 2015-08-18 at 5 48 40 pm

Hover:
screen shot 2015-08-18 at 5 48 54 pm

Click:
screen shot 2015-08-18 at 5 49 07 pm


ZogStriP:
Thanks :+1:

PR #3670, opened 2015-08-19
Dark theme fixes, part 3

riking:


app/assets/stylesheets/common/base/topic-post.scss
```diff
@@ -129,6 +129,11 @@ aside.quote {
   }
 }
 
+.topic-body {
```
riking: moved from the desktop/ and mobile/ folders

SamSaffron:
looks fairly safe, thanks!

PR #3671, opened 2015-08-19
CLEANUP: Coalesce repeated CSS properties

riking:
I ran the RubyMine inspections for "optimize background/margin/padding properties" and "duplicated css property".


eviltrout:
Really nice! I should start linting my CSS :)

PR #3676, opened 2015-08-19
Update Gemfile.lock

GitLaboratory:
Correction to support the version of "2.2.2" build.


eviltrout:
This is failing and we fixed it in https://github.com/discourse/discourse/commit/113e8d62bac17e65c93e5535af45d3a1d2879d86

PR #3677, opened 2015-08-19
fix for links in smtp providers

YesThatAllen:
not sure why mandrill still gets **bold** treatment but whatevs

[fixed]


gschlager:
It's bold because it's `[**Mandrill**]` instead of `[Mandrill]` (remove the `**` which is Markdown for bold).

YesThatAllen:
Thanks @gschlager. I was wondering why (as opposed to how)  it was bolded. But didn't unbold it bc it's no matter to me.

eviltrout:
Thanks!

PR #3678, opened 2015-08-20
FEATURE: Allow admin to change timestamp of topic.

tgxworld:
Reopening this... I accidentally deleted my branch on Github.


SamSaffron:
here goes nothing :) since it is super isolated code, easiest thing to do here is merge it in and have a play!

SamSaffron:
thank!

PR #3690, opened 2015-08-24
FIX: Title popup tip not positioned correctly.

tgxworld:
meta: https://meta.discourse.org/t/placement-of-the-title-is-required-dialog/30828/4

screenshot from 2015-08-24 18 18 42


ZogStriP:
Thanks :+1:

PR #3691, opened 2015-08-24
FIX: Dirty edit can toggle to another edit.

tgxworld:
Follow up to https://github.com/discourse/discourse/pull/3649#issuecomment-133227526.

Meta: https://meta.discourse.org/t/reply-doesnt-clear-the-edit-post-text-if-clicked-after-edit/26012


ZogStriP:
Looks fine. Thanks :+1:

PR #3692, opened 2015-08-25
FIX: Censored words filter not applied to title.

tgxworld:
meta: https://meta.discourse.org/t/profanity-filter-not-working-on-titles-topic-subject-line/31093


ZogStriP:
Thanks :+1:

tgxworld:
@ZogStriP Do you know if there are plans to move the dialect modules to es6? However, I don't think it will work with the V8 interpreter though.

ZogStriP:
Yeah we do have plans to move dialects to ES6. It's super hard though... We might do that at the same time we change to a better markdown engine.

tgxworld:
:+1:

PR #3695, opened 2015-08-25
FIX: Consider reserved usernames as not 'available'

riking:
https://meta.discourse.org/t/reserved-usernames-ignored-by-invites/32490


ZogStriP:
No tests? :cry:

riking:
@ZogStriP added tests :)

eviltrout:
Thanks, although I do think it's a bit odd that it'll suggest admin1 if admin is taken :)

PR #3696, opened 2015-08-25
FIX: hamburger menu transform for right-to-left layouts

scossar:
For rtl layouts the transform needs to be -330px.

### DIFF
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -59,5 +59,5 @@ float: left !important;
 
 .rtl #hamburger-menu.slideright {
   transform: translateX(-330px);
-  -webkit-transform: translateX(330px);
+  -webkit-transform: translateX(-330px);
 }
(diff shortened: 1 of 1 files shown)


eviltrout:
I am actually going to be messing with the menu today to make it display differently depending on the screen size. Is it possible you could look at my changes once committed and see if this still makes sense?
I suspect it might change it.

scossar:
Sure. Whatever translateX values you use will need to be negated for right-to-left layouts.

scossar:
It looks like the position of `.menu-panel` on larger screens is now being set with javascript. That makes it more difficult to reposition it for rtl layouts.  There is a new outer `.wrap` class that sets the outside width and margin of the application content. If you set the position of `.wrap` to `relative` then the `.menu-panel` could be positioned absolutely in relation to that without needing to use javascript.
This seems to do what you are looking for:
```
// base/discourse.scss
.wrap {
  position: relative;
}
.menu-panel.drop-down {
  position: absolute;
  right: 0;
  top: 50px;
}
```
As an aside, when I created the wrap class I added left and right padding to it. That was a mistake - it creates some problems, especially on mobile. The padding should be added to the `.container` class instead. It's an easy fix that I would be happy to sort out.

scossar:
The menu has been changed so that this no longer works.

PR #3697, opened 2015-08-26
FEATURE: Allow plugins to add admin dashboard warnings

riking:
Have wanted this in a few plugins for a while.

Need to add a helper to automatically do the plugin enabled checks, though.

### DIFF
--- a/spec/models/admin_dashboard_data_spec.rb
+++ b/spec/models/admin_dashboard_data_spec.rb
@@ -2,6 +2,32 @@ require 'spec_helper'
 
 describe AdminDashboardData do
 
+  describe "adding new checks" do
+    it 'calls the passed block' do
+      called = false
+      AdminDashboardData.add_problem_check do
+        called = true
+      end
+
+      AdminDashboardData.fetch_problems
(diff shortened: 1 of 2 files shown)


spec/models/admin_dashboard_data_spec.rb
```diff
@@ -2,6 +2,32 @@
...
+      AdminDashboardData.add_problem_check do
+        called = true
+      end
+
+      AdminDashboardData.fetch_problems
+      expect(called).to eq(true)
+    end
+
+    it 'calls the passed method' do
+      $test_AdminDashboardData_global = false
```
eviltrout: You are not tearing down the stuff you add to `AdminDashboardData` in these methods. Other tests will have them depending on when this test was added.

riking: Fixed: https://github.com/discourse/discourse/pull/3697/files#diff-3a95b4ba07b38c0b4584b4efb269cb8eR6

eviltrout:
Looks good to me, thanks!

PR #3698, opened 2015-08-26
FIX: Autocomplete and Emojis when bounded by non-word characters.

tgxworld:
meta: https://meta.discourse.org/t/two-emojis-delimited-by-a-dot-the-second-one-does-not-render/32111
          https://meta.discourse.org/t/pinging-in-parentheses-doesnt-auto-suggest/21377

The logic is a little hard to follow here but I hope I didn't break anything :stuck_out_tongue_closed_eyes: 

screenshot from 2015-08-26 17 26 04


ZogStriP:
Let's try it ;) Thanks :+1:

PR #3699, opened 2015-08-26
FIX: emoji unescape

gwwar:
Fix for https://meta.discourse.org/t/im-confused-by-o/30869/2 Unescape now checks if the emoji ends with a colon, or is an emoticon that begins with a colon. 

Title posts now only render emojis with ending colons.
screen shot 2015-08-26 at 12 05 23 pm

### DIFF
--- a/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb
+++ b/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb
@@ -30,16 +30,19 @@ for (var name in aliases) {
 }
 
 Discourse.Emoji.unescape = function(string) {
+  //this can be further improved by supporting matches of emoticons that don't begin with a colon
   if (Discourse.SiteSettings.enable_emoji && string.indexOf(":") >= 0) {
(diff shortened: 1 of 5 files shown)


gwwar:
working through some corner cases with face emojis `:)`

SamSaffron:
I like this... it is clear and clean and well tested ... I wonder though 

What about:

`a:smile:a`

or 

`b:smile:` 

In composer both cases just render as text.

gwwar:
Nice case, it looks like master also incorrectly replaces it with an emoji. I can clean up this behavior tomorrow.

screen shot 2015-08-26 at 8 04 03 pm

coding-horror:
Wait why would anyone expect a:smile:a to work? This seems fraught with danger

gwwar:
I've updated the regex so that emojis next to word-characters won't match, ie: `a:smile:a` will be text. This appears to match composer behavior. Please take another look at the additional test cases, and let me know if I'm missing some other expected behavior.

ZogStriP:
Awesome!

PR #3700, opened 2015-08-27
FIX: mobile settings page for rtl layouts

scossar:
- fixes the settings page admin-controls
- fixes the transform of the admin-detail section
- removes the transition from the admin-detail section

All changes only effect RTL layouts.


scossar:
I'll redo this.

PR #3701, opened 2015-08-27
Fix same domains links on subfolder installs

xfalcox:
isInternal link handling logic needs to test if the link is inside the same subfolder

### DIFF
--- a/app/assets/javascripts/discourse/lib/url.js.es6
+++ b/app/assets/javascripts/discourse/lib/url.js.es6
@@ -271,7 +271,7 @@ const DiscourseURL = Ember.Object.createWithMixins({
 
   // This has been extracted so it can be tested.
   origin: function() {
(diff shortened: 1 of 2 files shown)


riking:
Travis failure is unrelated.

xfalcox:
Maybe we should add tests for subfolder detection? And I signed the CLA already...

nlalonde:
Good catch. Your change removes the `http(s?)://` and adds a trailing slash at the end. Not sure if it matters, but to strictly keep the same behaviour this should do the trick:

```
window.location.origin + (Discourse.BaseUri === "/" ? '' : Discourse.BaseUri)
```

xfalcox:
Thx @nlalonde ! Fixed it and added some tests.

nlalonde:
Thanks! :+1:

PR #3702, opened 2015-08-27
FIX: prevent horizontal scrollbar from appearing at 1000px screen width

scossar:
Removes the medium-width rule that was setting the width of the `body`, `.wrap`and `.full-width` classes at 995px for screen sizes between 1000px and 1139px. Because Discourse is using `box-sizing: content-box` the padding on the outer `.wrap` class was being added to the 995px width and causing the horizontal scrollbar to appear. There is no need to set the `.wrap` width, other than with max-width. It can be allowed to fill the screen until it reaches that point.

The padding should still be removed from the `.wrap` class and added instead to `.container`.


coding-horror:
I agree, this was acting weird. Good to remove stuff.

PR #3706, opened 2015-08-29
FIX: rescale image during cooked_post_processor when only height or width is specified

gwwar:
Fix for https://meta.discourse.org/t/img-width-and-height-attribute-oddities/26891

When adding an image in the composer with only a width or height attribute, ie:

```
<img src="/uploads/default/original/1X/892174c24f4ddca8c145a8e35d6ba90cadca59ad.jpg" height="100"> 
```

The cooked_post_processor will rescale that relative to the original dimensions:
```
<img src="//localhost:3000/uploads/default/original/1X/892174c24f4ddca8c145a8e35d6ba90cadca59ad.jpg" height="100" width="200">
```

screen shot 2015-08-29 at 3 01 40 pm

screen shot 2015-08-29 at 3 01 06 pm

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -94,7 +94,20 @@ class CookedPostProcessor
 
   def get_size_from_attributes(img)
     w, h = img["width"].to_i, img["height"].to_i
-    return [w, h] if w > 0 && h > 0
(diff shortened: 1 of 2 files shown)


ZogStriP:
Thanks :+1:

PR #3708, opened 2015-08-30
FIX: Post age time not being translated correctly.

tgxworld:
Meta: https://meta.discourse.org/t/post-age-times-arent-visually-consistent/31058


ZogStriP:
Thanks :+1:

PR #3709, opened 2015-08-30
UX: Love hearts.

tgxworld:
screenshot from 2015-08-31 00 45 30


ZogStriP:
@coding-horror do we want that?

coding-horror:
Probably not on the home page per se, but Likes column is not visible by default anyway, so ... maybe it doesn't matter?

ZogStriP:
Yeah I guess it makes it consistent.

SamSaffron:
Nahh I much prefer this in monochrome here ...

PR #3710, opened 2015-08-30
FIX: use transform mixin for mobile admin settings

scossar:
The transition in `admin_base.scss` needs the webkit prefix.


ZogStriP:
Why change the mobile breakpoint?

scossar:
I've set the mobile breakpoint back to 700px. The idea behind changing it was to only have the mobile admin-setting navigation applied to layouts that are completely broken. The admin slide-out nav is an improvement over the way it worked previously, but it is still far from perfect.

SamSaffron:
ok its terrible on iphone at the moment, lets see if this fixes it :+1:

PR #3716, opened 2015-08-31
FIX: Fix the logout functionality in core

cpradio:
The logout action was missing from core, so it was throwing an error when the setting was enabled.


SamSaffron:
cool, thanks

PR #3718, opened 2015-09-01
FIX: Allow user to quote Emojis.

tgxworld:
meta: https://meta.discourse.org/t/cant-copy-and-paste-emoji-and-cant-quote-emoji-cry/32725

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js
+++ b/app/assets/javascripts/discourse/lib/utilities.js
@@ -97,7 +97,10 @@ Discourse.Utilities = {
     // Strip out any .click elements from the HTML before converting it to text
     var div = document.createElement('div');
     div.innerHTML = html;
-    $('.clicks', $(div)).remove();
+    var $div = $(div);
+    // Find all emojis and replace with its title attribute.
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/lib/utilities.js
```diff
@@ -97,6 +97,8 @@ Discourse.Utilities = {
     // Strip out any .click elements from the HTML before converting it to text
     var div = document.createElement('div');
     div.innerHTML = html;
+    // Find all emojis and replace with its title attribute.
+    $(div).find('img.emoji').replaceWith(function() { return this.title });
```
riking: should create `var $div = $(div)` and reuse

tgxworld: :+1: Thanks!

SamSaffron:
hmmm ... side effect though is that now titles on emoji are a bit uglier :( @coding-horror  ?

SamSaffron:
OK, this only impact alt, so I guess it is fine. :)

PR #3722, opened 2015-09-02
Move plugin menu outlet to live with other menus

gdpelican:


SamSaffron:
Interesting, isn't header-after-home-logo in the exact same spot though? 

Open to moving this, but 2 outlets that do the same thing seems a bit odd

gdpelican:
Hm, I did try the header-after-home outlet, but it didn't seem to appear with the other dropdowns; I'd guess because it's within that `contents clearfix` div. Is there another plugin out there using the header-after-home to display a dropdown?

SamSaffron:
no probs, lets move it and see how it goes, it seems to be in a very odd place.

PR #3723, opened 2015-09-02
FIX: Timegap only shows up for sequential posts.

tgxworld:
Meta:
* https://meta.discourse.org/t/large-time-gap-indicator-disappears-on-delete/30864/6
* https://meta.discourse.org/t/missing-x-months-ago-between-posts-when-post-numbers-are-not-continuous/32714

### DIFF
--- a/app/assets/javascripts/discourse/components/time-gap.js.es6
+++ b/app/assets/javascripts/discourse/components/time-gap.js.es6
@@ -2,6 +2,8 @@ import SmallActionComponent from 'discourse/components/small-action';
 
 export default SmallActionComponent.extend({
   classNames: ['time-gap'],
+  classNameBindings: ['hideTimeGap::hidden'],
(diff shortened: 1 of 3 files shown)


coding-horror:
sweet! this has always bugged me, even if it handles the deleted post case that would be perfect :icecream:

tgxworld:
@coding-horror This is how it'll look when the deleted post filter is applied

screenshot from 2015-09-02 17 29 37

SamSaffron:
I wonder, is it desirable to be hiding time gaps when you collapse the stream, I think its actually quite useful there cc @eviltrout  @coding-horror

coding-horror:
I am only worried about missing time gaps when posts are deleted at the moment.. are you referring to "Summarize This Topic" and filtering a topic by user?

SamSaffron:
lets try this out!

PR #3724, opened 2015-09-02
Always use locale fallback on server

gschlager:
Currently missing translations in server.yml can lead to errors that do no exist in production environment, which makes it quite hard to debug i18n issues.

This PR restores the previous locale fallback behavior in development environment. It uses the English text instead of showing en error message when a translation is missing on the server. That's the way it was before https://github.com/discourse/discourse/pull/3609.

The client side behavior is still unchanged: It uses fallback in production mode and shows the missing translation key in development mode.

Without this change:
image

With this change:
image

https://meta.discourse.org/t/always-enable-localization-fallback/32314

### DIFF
--- a/config/initializers/i18n.rb
+++ b/config/initializers/i18n.rb
@@ -22,16 +22,5 @@ class FallbackLocaleList < Hash
   end
 end
 
-class NoFallbackLocaleList < FallbackLocaleList
-  def [](locale)
-    [locale]
-  end
-end
-
-
-if Rails.env.development?
-  I18n.fallbacks = NoFallbackLocaleList.new
-else
-  I18n.fallbacks = FallbackLocaleList.new
-  I18n.config.missing_interpolation_argument_handler = proc { throw(:exception) }
-end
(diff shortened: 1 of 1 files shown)


gschlager:
Travis error seems to be unrelated.

eviltrout:
Wouldn't showing english in development mode make it harder to find keys that are missing? It's pretty clear from the error message that a key is missing there.

gschlager:
In this case it's clear. But it doesn't always work this ways. For example, try to preview the digest mail in German in development mode. You'll get an exception as described in https://github.com/discourse/discourse/pull/3734 because the format string isn't translated yet.
(It isn't translatable in Transifex, so this will fail in any language other than English. I'm working on a fix for that.)

Anyway, as a developer I'd like to work with a system that behaves the same, regardless if it is running in production mode or not.
And as a translator I honestly don't care about those "missing translation" messages. I don't need them - Transifex is enough for me. But maybe my workflow is completely different from everyone else...

PR #3728, opened 2015-09-04
FEATURE: Add a trigger for plugins to include their own post processing jobs.

tgxworld:
cc/ @SamSaffron


SamSaffron:
sure, but let's name it after_trigger_post_process

technically it should probably be named after_trigger_post_post_process... but that name is odd :)

tgxworld:
Updated :)

SamSaffron:
cool !

riking:
@tgxworld You should add another one here:
https://github.com/discourse/discourse/blob/master/lib/cooked_post_processor.rb#L24

PR #3731, opened 2015-09-05
FIX: simplify positioning of menu panel

scossar:
- Simplifies how the menu-panel is positioned by placing the menu components in their correct position in the markup and using css for positioning when possible
- menu-panel horizontal positioning is removed from the javascript and set with css
- fixes the issue with menu-contents being clipped on some mobile screens
- removes the gap that was occurring between the menu-panel and the header on mobile
- works for LTR and RTL layouts

If Discourse was using  `box-sizing: border-box` some height and width calculations would be more straightforward.


scossar:
I've added a fix for the gap that has been occurring between the slide-out panel and the header in the mobile view. The problem was that the headerHeight() function wasn't taking the padding-top on the header into account. There is 3px of padding in the desktop view and none in the mobile view, because Discourse is using content-box box-sizing this padding wasn't being accounted for.

eviltrout:
This PR looks kind of weird - can you rebase so I can review?

scossar:
Sure. It's really 3 separate fixes. One for the gap between the menu and
the header on mobile, one for the clipped content in the menu when with
window height is less than the the menu content height, and one for the
positioning of the menu so that it works for left-to-right and
right-to-left layouts. I can redo this and send them as three separate PRs.

scossar:
It is confusing, I'll close it and send as three PRs.

PR #3734, opened 2015-09-06
Remove hard-coded fallback to English locale

gschlager:
There's no need for this explicit fallback to English anymore:
- In production mode the English locale is always used as fallback.
- In development mode there is currently no fallback, so the English locale isn't loaded and the second call to `I18n.l()` fails too. PR https://github.com/discourse/discourse/pull/3724 would fix the error in development mode.


SamSaffron:
seems correct, thanks!

PR #3735, opened 2015-09-07
FIX: Discourse.BaseUri should not default to /

riking:
The new invariant is that is never ends in a slash.

as discussed with @SamSaffron


eviltrout:
This could break many things (although it looks good and safe). My call is to merge this after we cut our next stable release.

SamSaffron:
@eviltrout your call if you want to merge this in now.

eviltrout:
Let's do it and keep an eye on it.

PR #3736, opened 2015-09-07
FEATURE: Integrity hashes for scripts

riking:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

image
---
image

The `logster_javascript` partial will only take effect when there is an error before vendor.js loads and runs & puts in the real onerror handler, but this is necessary to get reports of integrity violations.

---

~~don't pull this, it breaks prod atm: https://code.google.com/p/chromium/issues/detail?id=527286~~ Worked around

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,6 @@
 require 'current_user'
 require 'canonical_url'
+require 'integrity_hash'
 require_dependency 'guardian'
 require_dependency 'unread'
 require_dependency 'age_words'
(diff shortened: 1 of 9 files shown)


eviltrout:
This is a cool idea but I feel that it should be hidden behind a feature flag at first. Can you add a site setting to enable/disable it? 

We can then try it out on various sites, see if there are issues then perhaps default it to Y at that point.

SamSaffron:
hmmm ... what exactly are we gaining here feature wise?

eviltrout:
It's more secure, as it ensures that assets haven't been tampered with. I think that's worth supporting.

riking:
Actually, according to https://github.com/rails/sprockets-rails

> Sprockets 3.x adds experimental support for subresource integrity checks

so we should probably wait for that rather than that messy code in integrity_hash that relies on the asset config being specifically set up.

PR #3738, opened 2015-09-08
FIX: add margin to allow for right and left box-shadow blur on search input

scossar:
There is an 8px blur on input[type="text"]:focus. The box-shadow is not accounted for by setting the input's box-sizing to border-box. This fix adds a 3px right and left margin to the input and then subtracts 6px from the input's width (margins are not included in border-box either.)


scossar:
I just changed it to set all margins on the input box a 0.5em - it's a minor detail.

PR #3739, opened 2015-09-08
FIX: eliminate 3px gap under mobile menu-panel

scossar:
@eviltrout 
Uses outerHeight() instead of height() to calculate header height in header.js.es6. There is padding on the header in the desktop view, but none in the mobile view. outerHeight() includes the padding in its calculation - this allows .menu-panel.slide-in to have its top position calculated more easily.


eviltrout:
This looks good thanks!

PR #3740, opened 2015-09-08
FIX: eliminate clipped menu content on small screens

scossar:
@eviltrout 
The height calculation for the slide-out menu in menu-panel.js.es6 is not taking the padding on the menu panel into account. This results in the menu-panel extending below the window when the window height is less than the menu contents. This fix is one way to do it. ~~It sets the PANEL_BODY_PADDING as a constant.~~

### DIFF
--- a/app/assets/javascripts/discourse/components/menu-panel.js.es6
+++ b/app/assets/javascripts/discourse/components/menu-panel.js.es6
@@ -2,7 +2,6 @@ import { default as computed, on, observes } from 'ember-addons/ember-computed-d
 import { headerHeight } from 'discourse/views/header';
 
 const PANEL_BODY_MARGIN = 30;
-//const PANEL_BODY_PADDING = 7;
 const mutationSupport = !!window['MutationObserver'];
 
 export default Ember.Component.extend({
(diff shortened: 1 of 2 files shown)


scossar:
It looks like I can simplify this. I'll send some changes soon.

eviltrout:
Something is a bit off with this -- the hamburger is displaying all the categories in one column instead of two on mobile mode. (Tested with chrome set to iphone5 mode)

scossar:
Yes, I see that. The problem there is the width on the category links of 45% combined with the category link margins is greater than 50% of the menu width. 

I can fix that in the next hour or so.

scossar:
I changed the margin on .category-link to padding, set the width to 50%, and the box-sizing to border-box.

SamSaffron:
yeah the clipping now is real bad, in chrome it is impossible to scroll to the bottom of menus @scossar  can you rebase? @eviltrout if a rebase is hard here maybe you should just borrow the fixes ?

scossar:
Sure, I just sent it. I can look at this in the morning (Vancouver time) if
you have any questions.

SamSaffron:
Pretty sure @eviltrout took care of this, can you review if there are any changes that need to be included and do a new PR for it? 

thanks for all your help here @scossar

scossar:
Yes you can close this now.

PR #3741, opened 2015-09-08
FIX: reorder header markup so it works in LTR and RTL layouts

scossar:
@eviltrout 
This places the menu components after the ul.icons. Doing that allows the menu panel to be positioned horizontally relative to the .d-header .panel div. The benefit of doing this is that it works for right-to-left layouts as well as LTR layouts. If this is not done you will have to test for whether or not you are in a RTL layout in _layoutComponent() and adjust the positioning there.

### DIFF
--- a/app/assets/javascripts/discourse/components/menu-panel.js.es6
+++ b/app/assets/javascripts/discourse/components/menu-panel.js.es6
@@ -28,11 +28,9 @@ export default Ember.Component.extend({
       const $buttonPanel = $('header ul.icons');
       if ($buttonPanel.length === 0) { return; }
 
-      const buttonPanelPos = $buttonPanel.offset();
-      const posTop = parseInt(buttonPanelPos.top + $buttonPanel.height() - $('header.d-header').offset().top);
-      const posLeft = parseInt(buttonPanelPos.left + $buttonPanel.width() - width);
-
-      this.$().css({ left: posLeft + "px", top: posTop + "px", height: 'auto' });
+      // These values need to be set here, not in the css file - this is to deal with the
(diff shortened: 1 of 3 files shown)


eviltrout:
I just tried this out and it's not 100% working. If I resize the browser vertically it's not sizing the dropdowns to the full height.

The bottom of the menu should always be a few pixels from the bottom of the screen. With this patch it doesn't do that.

eviltrout:
Oh wait I see there's another PR for that. Trying that out.

scossar:
This shouldn't change anything except how  `.drop-down` `top` position is calculated, and how the horizontal positioning `.menu-panel` is calculated. I might have included a part of the other PR in this one. I'll look at it again.

scossar:
I had deleted a few lines from menu-panel.scss that should only have been deleted in the other PR. I have re-included them now. This code should work exactly the same as what is in the master branch. 
The difference is that it works for RTL layouts. It still has the problem with menu-content in the slide-out menu being clipped when the menu content has a greater height than then browser window. There is a fix for that in the other PR.

In reference to that, you wrote: "The bottom of the menu should always be a few pixels from the bottom of the screen. With this patch it doesn't do that." This is the behaviour of the drop-down menu. For the slide-in menu the logic seems to say that if there is less than a 20px space between the bottom of the menu and the bottom of the window then the menu should extend to the bottom of the window.
```js
if ((menuTop + contentHeight) < ($(window).height() - 20)) {
        height = contentHeight + "px";
      } else {
        height = $(window).height() - menuTop;
      }
```

eviltrout:
Yes, sorry and thanks for the correction. I just tried it and it's working beautifully now. Thank you so much for the hard work on this.

PR #3742, opened 2015-09-09
Repurposing CONTRIBUTING.md into a link portal

erlend-sh:


eviltrout:
I feel like this removes a lot of useful information that is not necessarily obvious by the new version. For example the CLA is quite important and I'm not sure how users would quickly find it?

erlend-sh:
I'll resume discussion on Meta.

PR #3743, opened 2015-09-10
Add embed/info endpoint for TopicEmbed queries

santouras:
`TopicEmbed` requires a unique `embed_url`, however there is no API endpoint for fetching information about a url that may already exist in the system.

This means that plugins like the wordpress discourse plugin can fail if the plugin has sent information to discourse successfully but for some reason hasn't updated their own system. This can make them get into a continuous loop of sending data to discourse and getting back unhelpful error messages and never knowing what topic their url has been created and associated with.

This new `#info` endpoint will allow these plugins to initially query if their url is already in the system and get back information. Then if nothing is returned, they know they can safely request a new URL to be added.

### DIFF
--- a/app/controllers/embed_controller.rb
+++ b/app/controllers/embed_controller.rb
@@ -1,6 +1,8 @@
 class EmbedController < ApplicationController
   skip_before_filter :check_xhr, :preload_json, :verify_authenticity_token
-  before_filter :ensure_embeddable
+
+  before_filter :ensure_embeddable, except: [ :info ]
+  before_filter :ensure_api_request, only: [ :info ]
 
   layout 'embed'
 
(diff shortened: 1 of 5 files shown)


app/controllers/embed_controller.rb
```diff
@@ -35,6 +37,22 @@ def comments
     discourse_expires_in 1.minute
   end
 
+  def info
+    embed_url = params.require(:embed_url)
+    te = TopicEmbed.where(embed_url: embed_url).first
+
+    embed_info = { topic_id: 0 }
+
+    embed_info.merge!({
```
eviltrout: This should be done as a `TopicEmbedSerializer`

santouras:
updated that now @eviltrout

eviltrout:
Did you forget to add the new serializer to the commit? I don't see it in the diff and the builds are failing.

santouras:
yup sorry!

santouras:
ping @eviltrout

eviltrout:
Looks good to me, thanks!

PR #3745, opened 2015-09-10
FEATURE: Add a plugin outlet for custom controls.

tgxworld:


SamSaffron:
sure, we have user-custom-preferences, but I guess if you need a full control then we could add there

PR #3747, opened 2015-09-10
FIX: remove 'left: auto' for slide-in menu (required for RTL layouts)

scossar:
Sorry, I left this out of the previous PR. It has no effect on LTR layouts, but is required for RTL layouts to work.


eviltrout:
Thanks! :)

PR #3748, opened 2015-09-10
Category reordering fixes

riking:
- Fixed behavior: it will now swap positions and prevent you from saving if two categories have the same position (file: controllers/reorder-categories.js.es6)
 - Move "New Category" and "Reorder Categories" into a dropdown (file: components/categories-admin-dropdown.js.es6)
 - Always show "New Topic" on categories page, even for admins
 - Make category reorder modal full-height (.full-height-modal) (file: stylesheets/common/base/modal.scss)
 - Move category reorder stylesheet out of admin stylesheet


riking:
Test failure is spurious.

app/assets/javascripts/discourse/components/categories-admin-dropdown.js.es6
```diff
@@ -0,0 +1,37 @@
...
+        title: I18n.t('categories.reorder.title'),
+        description: I18n.t('categories.reorder.title_long'),
+        styleClasses: 'fa fa-random'
+      });
+    }
+    return items;
+  },
+
+  // sendAction() is pretty silly if you ask me
+  actionFor_createCategory: 'createCategory',
```
eviltrout: this is an odd JS convention.

Why not just make a const at the top of the file with a map for the values?  `const actionsFor = {}` then change the clicked to use that. I'd prefer that to the weird underscore

app/assets/javascripts/discourse/controllers/reorder-categories.js.es6
```diff
@@ -38,29 +38,22 @@ export default Ember.Controller.extend(ModalFunctionality, Ember.Evented, {
     return anyChanged;
   }.property('categoriesBuffered.@each.hasBufferedChanges'),
 
-  saveDisabled: Ember.computed.alias('showApplyAll'),
+  @computed('showApplyAll', 'showFixIndices')
+  saveDisabled(a, b) {
```
eviltrout: `a || b` isn't clear at all. I'd rename the variables to be what they represent or just use the `Ember.computed.or`CP macro

app/assets/javascripts/discourse/components/categories-admin-dropdown.js.es6
riking: I was about to say that wouldn't work, but it does: fc5e157

app/assets/javascripts/discourse/controllers/reorder-categories.js.es6
riking: doh, I thought at first I needed !or, and couldn't find the computed.* for that, so... yeah, changed to computed.or.

app/assets/javascripts/discourse/templates/modal/reorder-categories.hbs
```diff
@@ -1,36 +1,33 @@
...
-        </tbody>
-      {{/each}}
-    </table>
-    <div id="rc-scroll-bottom"></div>
-  </div>
+<div class="modal-body reorder-categories full-height-modal">
+  <div id="rc-scroll-anchor"></div>
+  <table>
+    <thead>
+    <th class="th-pos">Position</th>
```
eviltrout: Missing I18n keys here

riking:
build fixed

SamSaffron:
cool, lets try this, I really want this fix in

PR #3749, opened 2015-09-10
FIX: Don't allow editing seeded category security settings

riking:
https://meta.discourse.org/t/some-threads-no-reply-buttons-for-users/33101/
https://meta.discourse.org/t/you-should-not-be-allowed-to-change-lounge-security/17921/
https://meta.discourse.org/t/updating-software-resets-lounge-security/16753/

image

### DIFF
--- a/app/assets/javascripts/discourse/components/edit-category-security.js.es6
+++ b/app/assets/javascripts/discourse/components/edit-category-security.js.es6
@@ -7,16 +7,24 @@ export default buildCategoryPanel('security', {
 
   actions: {
     editPermissions() {
-      this.set('editingPermissions', true);
+      if (!this.get('category.is_special')) {
+        this.set('editingPermissions', true);
+      }
     },
 
     addPermission(group, id) {
-      this.get('category').addPermission({group_name: group + "",
-                                       permission: Discourse.PermissionType.create({id})});
+      if (!this.get('category.is_special')) {
+        this.get('category').addPermission({
+          group_name: group + "",
+          permission: Discourse.PermissionType.create({id})
+        });
+      }
     },
 
     removePermission(permission) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/components/edit-category-security.js.es6
```diff
@@ -7,16 +7,24 @@ export default buildCategoryPanel('security', {
...
+        this.set('editingPermissions', true);
+      }
     },
 
     addPermission(group, id) {
-      this.get('category').addPermission({group_name: group + "",
-                                       permission: Discourse.PermissionType.create({id})});
+      if (!this.get('category.is_special')) {
+        this.get('category').addPermission({
+          group_name: group + "",
```
eviltrout: Why `group + ""`? Is that to avoid calling `toString()`?

riking: Oops, looks like I triggered a format there. Same code as before.

app/serializers/category_serializer.rb
```diff
@@ -37,6 +38,11 @@ def can_delete
     true
   end
 
+  def is_special
+    [SiteSetting.lounge_category_id, SiteSetting.meta_category_id, SiteSetting.staff_category_id, SiteSetting.uncategorized_category_id]
+    .include? object.id
+  end
```
eviltrout: One thing I prefer to do to keep JSON size down is have boolean values return true always, then use `include_is_special?` with the logic. That way the attribute will not be present unless true.

riking: b706c59

SidVal:
Could it be that it is now the same in version v1.8.0.beta9 +20?

I mean that currently the lounge category does not allow adding new levels of trust?

PR #3751, opened 2015-09-11
Update some strings to meet referred options lists

ismail-arilik:
The options which changed strings are referred, were changed to lists so these strings were supposed to be generalized.


ZogStriP:
Thanks :+1:

ismail-arilik:
@ZogStriP thanks to you, too =)

PR #3753, opened 2015-09-11
FIX: Fallbacks for missing interpolation arguments

riking:
This takes effect when an interpolation is removed from a translation in
a Discourse update.

How this works:

The I18n::Backend::Fallbacks loops with a catch(:exception), so calling throw(:exception) will cause it to use the next locale, until it reaches English which is assumed to be correct.

```
        I18n.fallbacks[locale].each do |fallback|
          begin
            catch(:exception) do
              result = super(fallback, key, options)
              return result unless result.nil?
            end
          rescue I18n::InvalidLocale
            # we do nothing when the locale is invalid, as this is a fallback anyways.
          end
        end
```

Also, enable fallbacks in everything except development (#3724 for more discussion) - we should be able to test the fallbacks.

### DIFF
--- a/config/initializers/i18n.rb
+++ b/config/initializers/i18n.rb
@@ -28,8 +28,10 @@ class NoFallbackLocaleList < FallbackLocaleList
   end
 end
 
-if Rails.env.production?
-  I18n.fallbacks = FallbackLocaleList.new
-else
+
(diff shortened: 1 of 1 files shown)


gschlager:
:+1:

PR #3755, opened 2015-09-11
FIX: eliminate horizontal scrolling on category page for mobile

scossar:
Sets percentage widths on .category-list-item table cells. Moves the category borders to the edge of the screen:

screenshot 2015-09-11 11 45 22


eviltrout:
Looks good thanks!

PR #3758, opened 2015-09-13
FIX: Title input validation position.

tgxworld:
My previous fix was causing problems with private messages.
screenshot from 2015-09-14 00 31 40

Reverted back to the original design where we push the pop up right
screenshot from 2015-09-14 00 31 04


techAPJ:
:+1:

PR #3760, opened 2015-09-14
Signup cta

riking:
Oops, I never created a PR for this...


app/assets/javascripts/discourse/initializers/signup-cta.js.es6
```diff
@@ -0,0 +1,86 @@
...
+      // Requirements met.
+      session.set('showSignupCta', true);
+    }
+
+    screenTrack.set('anonFlushCallback', checkSignupCtaRequirements);
+
+    // Record a visit
+    const nowVisit = new Date().getTime();
+    const lastVisit = keyValueStore.getInt('anon-last-visit', nowVisit);
+    if (nowVisit - lastVisit > ONE_DAY) {
```
riking: this logic is bad, it's actually counting visits >24hr apart

app/assets/javascripts/discourse/lib/screen-track.js.es6
```diff
@@ -115,26 +113,52 @@ const ScreenTrack = Ember.Object.extend({
...
+          if (controller) {
+            const postNumbers = Object.keys(newTimings).map(function(v) {
+              return parseInt(v, 10);
+            });
+            controller.readPosts(topicId, postNumbers);
+          }
+        });
+      } else {
+        // Anonymous viewer - save to localStorage
+        const store = this.get('keyValueStore');
```
eviltrout: I find using the name `store` here quite confusing since the `store` means something totally different in our app. Could you replace with `keyValueStore` or `kvsStore` or something more clear?

app/assets/javascripts/discourse/templates/components/signup-cta.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="signup-cta alert alert-info">
+  {{#if session.hideSignupCta}}
+    <p>
+      {{i18n "signup_cta.hidden_for_session"}}
+      <a {{action "neverShow"}}>{{i18n "signup_cta.hide_forever"}}</a>
+    </p>
+  {{else}}
+    <p>{{i18n "signup_cta.line_1"}}</p>
```
eviltrout: I am not a fan of `line_1` and `line_2` for this. I think it would be better if you could come up with a better description for the paragraphs, or perhaps just include both together + HTML for paragraphs in one key.

eviltrout:
Cool let's try it out and see!

PR #3763, opened 2015-09-15
FEATURE: Track user profile views

tgxworld:
Meta: https://meta.discourse.org/t/is-there-a-way-to-track-profile-page-views/32385/21

### DIFF
--- a/app/models/user_profile_view.rb
+++ b/app/models/user_profile_view.rb
@@ -1,0 +1,5 @@
+class UserProfileView < ActiveRecord::Base
+  validates :user_profile_id, presence: true
+  validates :viewed_at, presence: true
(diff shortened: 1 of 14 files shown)


app/models/user_profile_view.rb
```diff
@@ -0,0 +1,47 @@
+class UserProfileView < ActiveRecord::Base
+  validates_presence_of :user_profile_id, :ip_address, :viewed_at
+
+  belongs_to :user_profile
+
+  def self.add(user_profile_id, ip, user_id=nil, at=nil, skip_redis=false)
```
tgxworld: I should probably refactor this as it shares the logic with `TopicViewItem`

PR #3766, opened 2015-09-15
Fix selected posts mobile

tgxworld:
### Input overflowing
screenshot from 2015-09-15 16 38 38

### Selected posts buttons not clickable and was being selected as part of the first post
screenshot from 2015-09-15 16 37 37

### Make category combobox take 100% width
Meta: https://meta.discourse.org/t/category-selector-too-narrow-when-moving-posts-on-mobile/33278
screenshot from 2015-09-15 16 49 44


ZogStriP:
Thanks :+1:

PR #3769, opened 2015-09-15
UX: Disable button while searching.

tgxworld:
The issue is more prominent on mobile where the connection speed is slow.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
+++ b/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
@@ -12,6 +12,7 @@ export default Ember.Controller.extend({
   selected: [],
   context_id: null,
   context: null,
+  searching: false,
 
   @computed('q')
   hasAutofocus(q) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -45,9 +46,9 @@ export default Ember.Controller.extend({
     return isValidSearchTerm(q);
   },
 
-  @computed('searchTerm')
-  isNotValidSearchTerm(searchTerm) {
-    return !isValidSearchTerm(searchTerm);
+  @computed('searchTerm', 'searching')
+  searchButtonDisabled(searchTerm, searching) {
+    return (searching || !isValidSearchTerm(searchTerm) ? true : false);
```
ZogStriP: Do you really need the ternary operator here?

SamSaffron: also ... trick if you insist on api being pristine
```
!!(searching || !isValidSearchTerm(searchTerm))

or 

!searching && isValidSearchTerm(searchTerm)

```

:)

tgxworld: > Do you really need the ternary operator here?

It was really late at night :stuck_out_tongue_closed_eyes: 



> also ... trick if you insist on api being pristine

:+1:

SamSaffron:
sure ... sounds good!

PR #3772, opened 2015-09-15
Email in the admin menu should be in the plural form

gzoritchak:
The view shows dozens or hundreds emails.

Email -> Emails

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2100,7 +2100,7 @@ en:
 
 
       email:
-        title: "Email"
+        title: "Emails"
         settings: "Settings"
         all: "All"
         sending_test: "Sending test Email..."


eviltrout:
Sorry but I don't think this is correct. By default it shows Settings. Emails > Settings is a bit weird to me. I believe as a category "Email" is good.

gzoritchak:
I agree with you Emails > Settings is a bit weird. 

To me that sections should be exactly like the Users or Logs ones. You go on it to see the activity and not the settings (which you have set before the docker build).  And clicking on it should display the "all emails" view. The email settings should be in the Settings section. It would make more sense.

I mentioned that point because it led to bad french translations of Sent and Skipped and made it incomprehensible.

But never mind, it's a minor point of detail.

PR #3773, opened 2015-09-16
FIX: Composer being abandoned when transiting between routes.

tgxworld:
Meta: https://meta.discourse.org/t/cant-dismiss-post/33137/7


eviltrout:
Thanks so much!!

PR #3775, opened 2015-09-16
Fix composer for FF14+

xfalcox:
Fix the preview for old versions of Firefox.

https://meta.discourse.org/t/composer-preview-broken-in-firefox-esr-17/33365

### DIFF
--- a/app/assets/stylesheets/desktop/compose.scss
+++ b/app/assets/stylesheets/desktop/compose.scss
@@ -369,10 +369,11 @@
 
 
     .wmd-input, .wmd-preview-scroller, .wmd-preview {
+      -moz-box-sizing: border-box;
       box-sizing: border-box;
       width: 100%;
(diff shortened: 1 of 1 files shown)


riking:
What's the min-height adjustment do?

xfalcox:
@riking Without min-height it looks like this:

discourse-without-minheight-auto

eviltrout:
I agree the `min-height` seems a bit risky. The `moz` tag seems safe but are we sure that won't break something else?

xfalcox:
@eviltrout I found it strange too, but, the selectors are pretty specific. I'm using it with CSS customization on my instance and it works well. Maybe test with CSS customization on meta?

https://bugzilla.mozilla.org/show_bug.cgi?id=308801

eviltrout:
Cool let's give it a shot! Thanks a lot.

PR #3780, opened 2015-09-17
FIX: Only cancel composer for when editing first post.

tgxworld:
Opps made a mistake in the previous PR :stuck_out_tongue_closed_eyes:


ZogStriP:
Thanks :+1:

PR #3782, opened 2015-09-17
Fix bug when adding new custom group

getabetterpic:
When adding a new group, if Delete was clicked before saving the
group, the modal would come up, but the Yes confirmation would
do nothing. This fixes that by taking an approach similar to the
Badges and checks to see if the Group is new, and if it is, just
returns to the index instead of invoking the confirm modal.

This also fixes the redirect after saving a new Group.

This was reported at https://meta.discourse.org/t/another-very-minor-admin-bug-when-creating-a-new-group/32734


eviltrout:
Thanks, looks great :)

PR #3784, opened 2015-09-18
UX: Allow users to open search in new tab.

tgxworld:
cc @SamSaffron


SamSaffron:
sure, seems good to me :)

PR #3786, opened 2015-09-18
FIX: Load fallback locales in Sidekiq jobs

gschlager:
Not sure if this is the right place for fixing this, but it seems to work...
https://meta.discourse.org/t/sidekiq-sending-email-digests-fails-missingtranslationdata/32500/8


Bahanix:
:+1:

eviltrout:
Cool thanks for the fix!

PR #3787, opened 2015-09-18
FIX: Some strings in locale files were not translatable

gschlager:
This PR contains 3 related commits:
- A few subtle changes to the locale files in order to remove duplicate keys. Ruby's `I18n` and Transifex can't merge duplicates; they use only the value from the last occurrence. After this change it's possible to translate date formats (e.g. used in the digest emails) that were not available in Transifex up until now.
- A spec that fails if it finds duplicate keys in any of the English locale files. It adds a class to the `lib` folder that is reused by the `pull_translations.rb` script. I wasn't sure where to put it, so I hope that's fine.
- Adds support for YAML anchors (e.g. `&datetime`) and aliases (e.g. `<<: *datetime`) to the `pull_translations.rb` script. The English locale files use aliases and anchors in order to prevent duplicate translations. Transifex doesn't support them. The script finds those aliases and anchors in the original locale files and adds them to the locale files pulled from Transifex.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -153,7 +153,6 @@ en:
     sign_up: "Sign Up"
     log_in: "Log In"
     age: "Age"
-    last_post: "Last Post"
     joined: "Joined"
     admin_title: "Admin"
     flags_title: "Flags"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -20,17 +20,21 @@ en:
     short_date: "D MMM, YYYY"
     long_date: "MMMM D, YYYY h:mma"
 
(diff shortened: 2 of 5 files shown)


SamSaffron:
This feels a bit high risk to be including it in 1.4, we can look at merging in a few days.

gschlager:
That's fine with me.

gschlager:
Who's brave enough to merge this in? :wink:

eviltrout:
I'm going to merge this in now. It seems fairly low risk as it only affects pulling translations. We can always roll back if we find the translations were mangled somehow. Thanks so much for the effort and patience on this one!

PR #3788, opened 2015-09-18
Fix bug where create topic is suggested when insufficient permissions

getabetterpic:
When a user is viewing a category they do not have permission to create
a topic in, they are still prompted to create a topic if they are at the
bottom of the topics. This fixes the issue, as well as disabling the
New Topic button if they don't have permission to create a topic for
that category.

screenshot from 2015-09-18 19 32 55

This is related to the ux issue [here](https://meta.discourse.org/t/creating-a-topic-where-permission-does-not-exist-assumes-uncategorised/18801) and a bug filed [here](https://meta.discourse.org/t/why-not-create-a-topic-in-category-without-topic-creating-permissions/31611)

### DIFF
--- a/app/assets/javascripts/discourse/routes/build-category-route.js.es6
+++ b/app/assets/javascripts/discourse/routes/build-category-route.js.es6
@@ -63,9 +63,15 @@ export default (filter, params) => {
 
     setupController(controller, model) {
       const topics = this.get('topics'),
-            periodId = topics.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : '');
+            periodId = topics.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : ''),
+            canCreateTopic = topics.get('can_create_topic'),
+            canCreateTopicOnCategory = model.get('permission') === Discourse.PermissionType.FULL;
 
-      this.controllerFor('navigation/category').set('canCreateTopic', topics.get('can_create_topic'));
+      this.controllerFor('navigation/category').setProperties({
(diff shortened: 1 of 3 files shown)


coding-horror:
oh nice one, I remember this has been asked about a fair bit :+1:

app/assets/javascripts/discourse/routes/build-category-route.js.es6
```diff
@@ -63,9 +63,15 @@ export default (filter, params) => {
...
       const topics = this.get('topics'),
-            periodId = topics.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : '');
+            periodId = topics.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : ''),
+            canCreateTopic = topics.get('can_create_topic'),
+            canCreateTopicOnCategory = model.get('permission') === Discourse.PermissionType.FULL;
 
-      this.controllerFor('navigation/category').set('canCreateTopic', topics.get('can_create_topic'));
+      this.controllerFor('navigation/category').setProperties({
+        canCreateTopicOnCategory: canCreateTopicOnCategory,
+        cannotCreateTopicOnCategory: !canCreateTopicOnCategory,
```
riking: You can add a property to `controller/navigation/category.js.es6` like this:

    cannotCreateTopicOnCategory: Ember.computed.not('canCreateTopicOnCategory'),

getabetterpic: Is there an advantage to using the computed property? I was leaning towards
this implementation since it puts all the related code in one place.

ZogStriP:
Thanks :+1:

PR #3789, opened 2015-09-19
Upgrade to fontawesome 4.4

gdpelican:


gschlager:
I don't think it's that easy. Take a look at @ZogStriP's commit when he updated to 4.3: https://github.com/discourse/discourse/commit/d9606bef9d286c32a55fd58ea4f0027ac5d03764

gdpelican:
Hm, I'd be surprised if it was that intensive again, considering that was a
relatively big version jump (4.0.3 to 4.3.0), but I'll go back see what
more needs to be done for this upgrade.

gschlager:
But changing *only* the version number isn't enough, is it? What about the font files? Either I'm not seeing them or you forget to add them. :wink:

gdpelican:
Wishful thinking on my part I suppose; was able to get the icon I was
looking for to show up with just that, but seems likely there'll be more to
do to get it going properly. I'll have a go at it over the next few days or
so.

SamSaffron:
@gdpelican this is the dance you need to dance :) 

https://github.com/discourse/discourse/commit/45ab42dea99d10ec519679a70a879830fcc38e8e

PR #3791, opened 2015-09-19
FIX: Slugs for badges

gschlager:
FIX: Use translated badge names for slugs
FIX: Use configured slug generater method for badges

https://meta.discourse.org/t/german-umlauts-are-not-handled-correct-for-badges/31675


ZogStriP:
Thanks :

PR #3792, opened 2015-09-19
FIX: I18n.exists? should use locale fallbacks

gschlager:
`I18n.exists?` ignored the locale fallbacks.
https://meta.discourse.org/t/search-help-doesnt-work-for-some-locales/31350

I'll try to get this into [Ruby I18n](https://github.com/svenfuchs/i18n)...

BTW: The `NoFallbackLocaleList` is really annoying during development (#3724). :frowning:

### DIFF
--- a/lib/freedom_patches/i18n_fallbacks.rb
+++ b/lib/freedom_patches/i18n_fallbacks.rb
@@ -1,0 +1,17 @@
+module I18n
+  module Backend
+    module Fallbacks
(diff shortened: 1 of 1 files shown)


SamSaffron:
cool! 

will merge this, but can you add some test cases here?

gschlager:
Yes, I'll do that. Didn't have time for it right now, but will need them for I18n anyway.

PR #3793, opened 2015-09-20
Stable

3rdwiki:
An issue would be that on the title of a discourse post ...

you cannot add a full stop to the end of a title and if you add a form of punctuation e.g. a ? or a ! 
you get either the full stop not included or the '?' is forced right next to the end of the line.

Apart from bad grammar, this is bad punctuation and I feel discourse is forcing the issue.

Of course even if you 'go back' and try to fix this - it doesn't change the issue.

screenshot 2015-09-20 at 06 05 44


SamSaffron:
hmmm something is terribly wrong with this PR, closing

3rdwiki:
Where do I raise an 'issue' with @discourse , then ? Because there's an error and I want it fixed.

PR #3795, opened 2015-09-21
FIX: relatively position slide-in menu as a work around for ios scrolling behaviour

scossar:
ios devices have a built in 'elastic scrolling' feature. This allows for scrolling past the top and bottom of the webpage. It's causing some weird behaviour with the slide-in menu. When expanded, it's remaining fixed in position while the elements around it may be changing their position (see: https://meta.discourse.org/t/elastic-scrolling-on-ipad-causes-header-to-unstick-from-top-of-window/33481). The problem is worse when there is a custom header added to the page. 

fixed-position-menu

This is a possible fix. It positions  `.menu-panel.slide-in` relative the the `.d-header.wrap` so that the header and the menu can't become separated from each other. The only risk that I can see from making this change is that it removes `position: relative` from `.d-header .contents` and adds it instead to `.d-header .wrap`. I've looked quite closely at it and it seems fine.

This isn't directly related to the PR but, in `menu-panel.scss` there is this rule for `.panel-body`:
```css
  .panel-body {
    position: absolute;
    top: 3px;
    bottom: 37px;
    width: 97%;
  }
```
I don't think that is required for anything. When there is a scrollbar on the slide-in menu it is causing .panel-body to override the padding that has been set on the slide-in menu.

### DIFF
--- a/app/assets/javascripts/discourse/components/menu-panel.js.es6
+++ b/app/assets/javascripts/discourse/components/menu-panel.js.es6
@@ -15,6 +15,7 @@ export default Ember.Component.extend({
     const $window = $(window);
     let width = this.get('maxWidth') || 300;
     const windowWidth = parseInt($window.width());
+    const $menuControls = $('.d-header .panel');
 
     if ((windowWidth - width) < 50) {
       width = windowWidth - 50;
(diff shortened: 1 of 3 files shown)


scossar:
I'll close this for now.

PR #3800, opened 2015-09-22
Import script infrastructure - batch skipping, usernames

riking:
Querying for the existence of 1000 rows at a time is much faster than 1 at a time. So let's do that.

The import scripts weren't updated to handle the username format changes.


techAPJ:
:+1:

PR #3801, opened 2015-09-22
[WIP] Hide 'boring' PMs in their own tab

riking:
The conditions for a PM to be "boring" are:

 - Messages from the site contact user are only boring to the sender until there's a reply
 - Flag -> notify moderators is always boring unless there's a non-auto-inserted reply
 - Flag -> notify user is only boring to the sender until there's a reply

The implementation needs to change - we can't have this monster SQL firing off all the time.

Proposed implementation: A column on the topic holding the 'boringness'.

0 - not boring to anyone, 1 - boring to topic creator, 2 - boring to all participants.
When a topic is created, the boringness is 0 unless specified otherwise in the PostCreator options.
Whenever a reply is made to a topic, unless the reply is an auto-inserted reply from resolving a Notify Moderators flag, the boringness is set to 0.

Messages from the site contact user, and Flag -> Notify User are boring 1.
Flag -> Notify Moderators are boring 2.

We can adapt the queries present in this WIP version of the patch to backfill the boringness for all topics in a migration.

Then, the list queries would just be `.where("topics.boringness = 0 OR (topics.boringness = 1 AND topics.user_id = :user_id)")` and `.where("topics.boringness = 2 OR (topics.boringness = 1 AND topics.user_id <> :user_id)")`.

WIP, on hold. If you want to take up this feature, notify me (leave a comment on the PR?)

### DIFF
--- a/app/assets/javascripts/discourse/controllers/user.js.es6
+++ b/app/assets/javascripts/discourse/controllers/user.js.es6
@@ -1,6 +1,7 @@
 import { exportUserArchive } from 'discourse/lib/export-csv';
 import CanCheckEmails from 'discourse/mixins/can-check-emails';
 import computed from 'ember-addons/ember-computed-decorators';
+import { propertyNotEqual } from 'discourse/lib/computed';
 
 export default Ember.Controller.extend(CanCheckEmails, {
   indexStream: false,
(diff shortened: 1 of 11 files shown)


SamSaffron:
I feel like categories is going to be a better way to fight this battle.

PR #3802, opened 2015-09-22
FIX: Correct processing of quote line in MyBB import

dandv:
Thanks @riking for [the help](https://meta.discourse.org/t/mybb-import-doesnt-process-on-quoted-posts-pid-and-date/27412/8?u=dandv).


SamSaffron:
sure ... lets try this.

PR #3803, opened 2015-09-22
FIX: keep topic admin icon from overlapping drop-down menu

scossar:
The drop-down menu inherits a z-index of 1000 from its parent's stacking context (the header). The admin wrench also has a z-index of 1000, but occurs later in the document so it appears on top of the menu. Reducing `.show-topic-admin` z-index to 999 allows the drop-down menu to appear overtop of the wrench.

I'm not sure that `.show-topic-admin` needs a z-index set on it at all.


eviltrout:
Thanks, seems to work great :)

PR #3804, opened 2015-09-22
Add tests for I18n.exists? freedom patch

gschlager:
Tests for PR https://github.com/discourse/discourse/pull/3792

### DIFF
--- a/spec/helpers/i18n_fallbacks_spec.rb
+++ b/spec/helpers/i18n_fallbacks_spec.rb
@@ -1,0 +1,56 @@
+require 'spec_helper'
+
+describe 'freedom patch for I18n::Backend::Fallbacks' do
+  before do
+    I18n.locale = :de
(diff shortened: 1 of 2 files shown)


spec/helpers/i18n_fallbacks_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+
+describe 'freedom patch for I18n::Backend::Fallbacks' do
+  before do
+    I18n.locale = :de
+    SiteSetting.stubs(:default_locale).returns(:de)
```
riking: SiteSetting.default_locale = 'de'

gschlager:
@riking Thanks for pointing me in the right direction. I fixed the tests.

I had to remove
```ruby
after do
  I18n.backend.reload!
end
```
since it broke many of the test. But I guess the `foo` and `bar` keys won't do much harm anyway if they stay loaded.

PR #3805, opened 2015-09-23
FIX: Don't subscribe to Desktop Notifications on mobile.

tgxworld:
On mobile, we're getting the follow error for trying to construct a `Notification` since it isn't supported on Android. On top of that, the `Notification` API is meant for Desktop notifications so we can avoid setting it up and subscribing to it on Mobile. 

```
Uncaught TypeError: Failed to construct ‘Notification’: Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead
```

Related issue: https://code.google.com/p/chromium/issues/detail?id=481856

There was a suggestion to check for the error in the issue but it didn't make sense to me because it assumes that the permission has not been requested before.

### DIFF
--- a/app/assets/javascripts/discourse/initializers/subscribe-user-notifications.js.es6
+++ b/app/assets/javascripts/discourse/initializers/subscribe-user-notifications.js.es6
@@ -29,10 +29,6 @@ export default {
         });
       }
 
-      bus.subscribe("/notification-alert/" + user.get('id'), function(data){
-        onNotification(data, user);
-      });
-
       bus.subscribe("/notification/" + user.get('id'), function(data) {
         const oldUnread = user.get('unread_notifications');
         const oldPM = user.get('unread_private_messages');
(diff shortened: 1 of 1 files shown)


ZogStriP:
Thanks :+1:

fbender:
Instead of Browser/Mobile sniffing, the code should feature detect the capabilities (or at least wrap it in a try…catch so the app can act on that without crashing). Unfortunately, I don't know enough of the internals to fix this myself. @tgxworld would you mind having a look into that?

Be aware that Desktop browsers may also switch to Notifications only able to be fired from ServiceWorkers and/or secure contexts, which will throw if the code stays as is.

tgxworld:
@fbender It has already been implemented in https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/lib/desktop-notifications.js.es6#L17-L55.

This PR is just skipping the setup of desktop notifications for mobile browsers.

PR #3808, opened 2015-09-23
FIX: Use translated badge name and slug within notifications

gschlager:
I had to revert some of the changes made in https://github.com/discourse/discourse/pull/3791 and move the code from `BadgeSerializer` to `Badge` in order to reuse the `displayName` and `slug` for notifications.

### DIFF
--- a/app/models/badge.rb
+++ b/app/models/badge.rb
@@ -329,11 +329,37 @@ SQL
     Badge.find_each(&:reset_grant_count!)
   end
 
+  def display_name
+    if self.system?
+      key = "admin_js.badges.badge.#{i18n_name}.name"
+      I18n.t(key, default: self.name)
+    else
+      self.name
+    end
+  end
+
+  def long_description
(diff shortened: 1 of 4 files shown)


app/models/badge.rb
```diff
@@ -329,12 +329,38 @@ def self.ensure_consistency!
...
+    if self[:long_description].present?
+      self[:long_description]
+    else
+      key = "badges.long_descriptions.#{i18n_name}"
+      I18n.t(key, default: '')
+    end
+  end
+
+  def slug
+    Slug.for(self.display_name, '-')
```
gschlager: We always need a slug for badges. If there isn't one, `- ` will be used as slug.
URLs without a slug like https://meta.discourse.org/badges/2 do not work when entered in the browser's address bar. I'm not sure if and how this can be changed so that badge URLs without slug work.

ZogStriP:
Thanks

PR #3809, opened 2015-09-23
FIX: Only enabled badges can be granted

gschlager:
There was an error when you tried to grant a disabled badge.

Bonus FIX: Sort badges by displayName


ZogStriP:
Thanks :+1:

PR #3810, opened 2015-09-24
Update server.zh_TW.yml

stephenyeong:
update login: translation


techAPJ:
Please add/update translations from Discourse Transifex project: https://www.transifex.com/projects/p/discourse-org/

PR #3812, opened 2015-09-24
FIX: allow emoji class when crawling embedded content

gwwar:
Fixes: https://meta.discourse.org/t/emoji-are-huge-when-embedding/32634

Blogs that use standard emoji libraries may display huge emoji in embedded posts. This is caused by ruby readability stripping all attributes except for `src` and `href`. Many libraries use the `class='emoji'` and css to scale down the emoji images. This fix adds an extra class name whitelist so specified class names won't be removed when embedded.

Example of a huge emoji:
screen shot 2015-09-24 at 3 38 42 pm

To reproduce embed some content using crawler settings, for example:

whitelist

Also adds rspc-html-matchers for easier testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -120,6 +120,7 @@ group :test, :development do
   gem 'simplecov', require: false
   gem 'timecop'
   gem 'rspec-given'
+  gem 'rspec-html-matchers'
   gem 'pry-nav'
   gem 'spork-rails'
   gem 'byebug', require: ENV['RM_INFO'].nil?
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -337,6 +337,9 @@ GEM
     rspec-given (3.5.4)
       given_core (= 3.5.4)
       rspec (>= 2.12)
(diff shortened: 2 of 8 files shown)


SamSaffron:
We probably want a site setting here for embed allowed classes and default it to emoji, that way we have the flexibility of allowing classes through and styling with custom css.

gwwar:
Sounds reasonable. I can update the PR to expose the allowed classes field in admin settings.

If you'd think something more generic might work better, as a first pass I could pass through options for `:tags` and `:attributes` for [ruby readability](https://github.com/cantino/ruby-readability) Unfortunately the current whitelist options in readability are rather simple. If you whitelist an attribute it's allowed for all whitelisted tags.

SamSaffron:
This seems ok to me @gwwar  can you rebase ... @eviltrout  if looks good to you can you merge in.

eviltrout:
Thanks @gwwar, this looks great!

PR #3813, opened 2015-09-25
UX: Topic link should not take full width.

tgxworld:
Meta: https://meta.discourse.org/t/reply-as-linked-topic-clickable-link-target-is-double-the-length/33631/4


tgxworld:
Failure is not related

SamSaffron:
cool lets try this, hopefully no side effects :)

PR #3815, opened 2015-09-25
FEATURE: Push Notifications for Chrome.

tgxworld:
bug_report-19

- [x] Handle config of push notifications properly
- [x] Fallback to Desktop notification when push notification is disabled.
- [ ] Handle condition when connection is not done over HTTPS


app/controllers/push_notifications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      render json: failed_json
+    end
+  end
+
+  def unsubscribe
+    PushNotifications.unsubscribe(current_user)
+    render json: success_json
+  end
+
+  def latest
```
tgxworld: Payload hasn't been released for push messaging so we have to do this. See https://code.google.com/p/chromium/issues/detail?id=486040

lib/push_notifications.rb
```diff
@@ -0,0 +1,51 @@
+class PushNotifications
+  KEY_PREFIX = "push-notification-subscription:".freeze
+  GCM_ENDPOINT = 'https://android.googleapis.com/gcm/send'.freeze
+
+  def self.registration_id(user)
+    $redis.get(redis_key(user))
+  end
+
+  def self.subscribe(user, endpoint)
+    $redis.set(redis_key(user), extract_registration_id(endpoint))
```
tgxworld: TODO: This is sharing the registration ID across browsers.

config/locales/server.en.yml
```diff
@@ -2080,6 +2082,9 @@ en:
 
         If the above link is not clickable, try copying and pasting it into the address bar of your web browser.
 
+    push:
+      body: You have %{count} new notifications.
```
gschlager: This needs pluralization (`one` and `other`).

tgxworld: hmm I could add that in but currently, we won't call the translation when we only have one notification.

gschlager: It's needed, even if it doesn't make sense for English. Other languages have other pluralization rules and Transifex doesn't detect it as pluralized unless there are `one` and `other` keys in English. Trust me. ;-)

tgxworld: Ah ic. :+1: thanks!!

lib/push_notifications.rb
```diff
@@ -0,0 +1,51 @@
...
+        body = URI.encode_www_form(
+          registration_id: registration_id,
+          collapse_key: Discourse.base_url
+        )
+
+        result = Excon.post(GCM_ENDPOINT, headers: headers, body: body)
+
+        if result.status == 200 && result.body.start_with?("Error")
+          Rails.logger.error("There was an error from Google Cloud Message: #{result.body}")
+        elsif result.status != 200
```
tgxworld: TODO: Check the response status and unsubscribe registration ids that are no longer in use.

PR #3816, opened 2015-09-25
Make the small-action width responsive on desktops

cpradio:
https://meta.discourse.org/t/small-action-class-is-not-responsive-on-desktops/33742


ZogStriP:
Thanks :+1:

PR #3817, opened 2015-09-25
FIX: Pluralizations in English locale files

gschlager:
Adds a test that checks all English locale files for invalid pluralizations and fixes all existing errors.

https://meta.discourse.org/t/internal-server-error-caused-by-i18n-invalidpluralizationdata/33573

BTW: There will be conflicts with my other PR https://github.com/discourse/discourse/pull/3787 that is still open. I'll rebase as soon as one of those PRs is merged into master.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -537,8 +537,8 @@ en:
         ok: "We will email you to confirm"
         invalid: "Please enter a valid email address"
         authenticated: "Your email has been authenticated by {{provider}}"
+        frequency_immediately: "We'll email you immediately if you haven't read the thing we're emailing you about."
         frequency:
-          zero: "We'll email you immediately if you haven't read the thing we're emailing you about."
           one: "We'll only email you if we haven't seen you in the last minute."
           other: "We'll only email you if we haven't seen you in the last {{count}} minutes."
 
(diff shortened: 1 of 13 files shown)


config/locales/client.en.yml
```diff
@@ -1510,11 +1509,6 @@ en:
             one: "1 person voted for this post"
             other: "{{count}} people voted for this post"
 
-      edits:
-        one: 1 edit
-        other: "{{count}} edits"
-        zero: no edits
```
gschlager: The `edits` [key exists for a long, long time](https://github.com/discourse/discourse/blame/master/config/locales/client.en.yml#L1513-L1516), but I couldn't find a place where it is actually used. That's why I removed it.

config/locales/client.en.yml
```diff
@@ -1737,12 +1729,10 @@ en:
           other: "Unread ({{count}})"
         help: "topics you are currently watching or tracking with unread posts"
         lower_title_with_count:
-          zero: ""
```
gschlager: `lower_title_with_count.zero` was never used.

spec/integrity/i18n_spec.rb
```diff
@@ -57,4 +57,46 @@
...
+    locale_files = ['config/locales', 'plugins/**/locales']
+                     .product(['server.en.yml', 'client.en.yml'])
+                     .collect { |dir, filename| Dir["#{Rails.root}/#{dir}/#{filename}"] }
+                     .flatten
+                     .map { |path| Pathname.new(path).relative_path_from(Rails.root) }
+
+    module Pluralizations
+      def self.load(path)
+        whitelist = Regexp.union([/messages.restrict_dependent_destroy/,
+                                  /js.filters.(category|new|latest|unread).title/])
```
gschlager: Small whitelist: AFAIK there's no reason to pluralize the filters and it was a nightmare to find all the places where those keys are used. :wink:

eviltrout:
Wasn't the issue only with server side keys? The client templates seem to be quite a bit longer with this patch and I thought `zero` was a nice feature.

gschlager:
Well, the problem exists everywhere - on client and server. Transifex doesn't detect a pluralized string as soon as you use anything other than `one` and `other`. There's no way around it.

Yes, the OP in the the linked topic did only complain about the server keys, but there were a few about the client keys as well. I can search for them if you want. :wink:

The templates are a little bit longer now, I know. :cry:  I was thinking about adding an additional argument to the `I18n.t` method for the "old" `zero` key, but decided against it since this change affects only 3 templates. Do you have a better solution in mind?

gschlager:
I fixed the merge conflict.
There are still a few keys that aren't pluralized properly, see: https://meta.discourse.org/t/js-filters-latest-title-incomplete-pluralization-in-transifex/34116/2

@eviltrout How should I fix the (remaining) issues without complicating the templates too much? Any advice?

gschlager:
@eviltrout I rebased and fixed the pluralizations for the filter keys. Can you review this please?
https://meta.discourse.org/t/js-filters-latest-title-incomplete-pluralization-in-transifex/34116/6

eviltrout:
Thanks! Now that we've cut a new beta I'm merging this in so we can try it out :)

ZogStriP:
:+1:

gschlager:
Thanks for merging it in.

Excited

PR #3819, opened 2015-09-27
FIX: Add missing pluralization keys to poll plugin locale

gschlager:
https://meta.discourse.org/t/some-keys-in-polls-contain-numbers-and-thus-needs-pluralization/33794


ZogStriP:
Thanks :+1:

PR #3820, opened 2015-09-28
save height on small screens

rubo77:
there is a lot of white space wetween each post, that annoys on small screens.

This will minimize the used space if the screen is lower than 700px in height

### DIFF
--- a/app/assets/stylesheets/desktop/topic-post.scss
+++ b/app/assets/stylesheets/desktop/topic-post.scss
@@ -1025,3 +1025,17 @@ and (max-width : 767px) {
   }
 
 }
+
+@media all
+and (max-height: 700px) {
+   .topic-body{
+      padding-bottom:0px;
+      margin-bottom:-15px;
+   }
+   .ember-view{
+     margin-bottom: 10px;
+   }
+   nav.post-controls{
+    margin-top: -18px;
+   }
+}


ZogStriP:
Can you post a before/after screenshot so that we can see what changes?

coding-horror:
Yes we need a screenshot for this one...

rubo77:
[on large screens](https://tinypng.com/site/output/u6c049knisvlbiqr.png/discourse-style-large-screen.png)

on large screens

---
 [on small screens](https://tinypng.com/site/output/vogann9ke3o796d6.png/discourse-style-small-screen.png)

on small screens

coding-horror:
Hmm images not appearing for me?

rubo77:
didn't know, that the images at http://tinypng.org/ expire after a day. Here again:

# on small screens:

[![small][1]][1]
---

# on large screens:

[![large][2]][2]

  [1]: http://i.stack.imgur.com/LcEPZ.png
  [2]: http://i.stack.imgur.com/6QBkl.png

coding-horror:
Looks OK to me, not too risky, and limited to desktop CSS. What device has a < 700px height??

rubo77:
if you have a device with a kigh resolution screen you usually enlarge all websites by 200% or more.

On my 13'' screen with 3200x1800px and a zoom level of about 300% I get a readeable size but it counts as just 600px in height

SamSaffron:
I just rolled this back it was targeting .ember-view which is pretty much every element on the page and giving it a margin, this in turn was breaking the front page 

image


I am just not sure I agree with this patch if we want to save space we should do so more consistently and definitely avoid negative margins and so on.

rubo77:
I think this will be a less intrusive anhancement then:

    .post-menu-area {
      margin-bottom: 0px;
      margin-top: -18px;
    }
    .topic-body {
      padding: 5px 11px 2px;
    }

PR #3823, opened 2015-09-30
FIX: phpBB3 importer failed to import users

gschlager:
Can we backport this to the stable branch as well?
I fear there will be quite a lot support request otherwise.

https://meta.discourse.org/t/phpbb-import-no-users-imported/33873

### DIFF
--- a/script/import_scripts/base.rb
+++ b/script/import_scripts/base.rb
@@ -198,12 +198,7 @@ class ImportScripts::Base
     return false if import_ids.empty?
 
     existing = "#{type.to_s.classify}CustomField".constantize.where(name: 'import_id')
-
-    if Fixnum === import_ids.first
-      existing = existing.where('cast(value as int) in (?)', import_ids)
-    else
-      existing = existing.where('value in (?)', import_ids)
-    end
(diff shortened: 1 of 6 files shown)


script/import_scripts/base.rb
```diff
@@ -198,12 +198,7 @@ def all_records_exist?(type, import_ids)
     return false if import_ids.empty?
 
     existing = "#{type.to_s.classify}CustomField".constantize.where(name: 'import_id')
-
-    if Fixnum === import_ids.first
-      existing = existing.where('cast(value as int) in (?)', import_ids)
-    else
-      existing = existing.where('value in (?)', import_ids)
-    end
+    existing = existing.where('value in (?)', import_ids.map(&:to_s))
```
gschlager: @riking I had to remove the cast. Some importers (like phpBB3) insert numbers and strings as `import_id` into the same custom_fields table. This change prevents cast errors during imports.

riking: Ah, that would do it. Got it.

nlalonde: Value column is always a string (text) column, so this works.

PR #3829, opened 2015-10-02
FIX: Who Liked a Post now shows UserCard when avatar is clicked

cpradio:


SamSaffron:
I wonder though if a computed attribute client side or pushing up to basic user serializer is better, but will accept this for now, thanks :)

cpradio:
You know I wondered that too, but I saw other serializers do thing, so I went with "follow an example". Granted, there are a few spots where computed properties are used in a similar way too...

cpradio:
If you decide with computed properties, let me know and I'll submit a PR undoing this and using a computed property. I definitely don't have a preference.

PR #3830, opened 2015-10-03
FIX: SSO provider redirect now works if user is initially not logged in

ntadej:
A fix for https://meta.discourse.org/t/discourse-sso-provider-login-unknown-error/28231

When logging in, we should not redirect to SSO callback URL but instead return it and then redirect using JavaScript. I'm not sure if this is the best fix. Open to suggestions and improvements.


ntadej:
If you think this is a valid fix, I can try and fix the tests.

ntadej:
This seems to be fixed now so I'm closing this.

PR #3831, opened 2015-10-03
UX: Emoji toolbar on mobile.

tgxworld:
meta: https://meta.discourse.org/t/more-discoverability-for-emoji/11616/6


coding-horror:
:+1: very excited we have a way to trigger the emoji picker on mobile now thanks to you and @eviltrout

PR #3835, opened 2015-10-04
FIX: typo in category-unread.hbs

scossar:
Typo in `category-unread.hbs` is causing category-unread.newTopics to return a title of [object][object]. This is what shows up in the tooltip.

screenshot 2015-10-04 13 21 36


ZogStriP:
Thanks :+1:

PR #3839, opened 2015-10-07
UX: Add mobile love to admin-nav for Plugin Pages

cpradio:
Before Screenshot:
image

After Screenshot:
image


SamSaffron:
lets try this, its definitely better than what it was

PR #3844, opened 2015-10-08
FIX: Pressing Ctrl+F twice should show browser find

gschlager:
https://meta.discourse.org/t/you-can-no-longer-press-ctrl-f-twice-to-use-native-find/34265


nlalonde:
Thanks!

PR #3846, opened 2015-10-09
Disable byebug when RubyMine is used for debugging

gschlager:
Debugger in RubyMine doesn't work well with byebug.
For example, debugging Sidekiq jobs doesn't work when byebug is loaded.


SamSaffron:
sure, thanks

PR #3847, opened 2015-10-11
Prepare for separate RSpec helper files

andyw8:
Since rspec-rails 3, the default installation creates two helper files:
* `spec_helper.rb`
* `rails_helper.rb`

`spec_helper.rb` is intended as a way of running specs that do not
require Rails, whereas `rails_helper.rb` loads Rails (as Discourse's
current `spec_helper.rb` does).

For more information:

https://www.relishapp.com/rspec/rspec-rails/docs/upgrade#default-helper-files

In this commit, I've simply replaced all instances of `spec_helper` with
`rails_helper`, and renamed the original `spec_helper.rb`.

This brings the Discourse project closer to the standard usage of RSpec in a Rails app.

At present, every spec relies on loading Rails, but there are likely
many that don't need to. In a future pull request, I intend to introduce a
separate, minimal `spec_helper` which can be used in tests which don't
rely on Rails.


SamSaffron:
I am afraid of this commit, a real side effect we saw in the past when we allowed these huge changes it that your name shows up everywhere when people browse the project on github, I would prefer if the team just slowly moved to this pattern each time we commit changes to a specific spec 

@eviltrout  thoughts?

eviltrout:
I actually have no objections to this commit at all. It's a rename and is probably quite safe. Doesn't bother me if someone's name shows up on github either -- what is the concern there?

andyw8:
I appreciate that wide changes such as this do somewhat pollute the git history.

I think the alternative approach @SamSaffron suggest would difficult to enforce, and it would likely take several years to move everything over, possibly causing confusion in the mean time due the two different styles present.

SamSaffron:
@andyw8  ok I will merge this in if you rebase

andyw8:
@SamSaffron I've rebased.

SamSaffron:
thanks @andyw8  appreciated

tgxworld:
@SamSaffron I think we should notify plugin authors that wrote specs about this. :smile:

gdpelican:
Yarp, I just tripped over this. Assuming I can just rename spec_helper to rails_helper like in here and everything will be fine.

PR #3848, opened 2015-10-11
Remove example CSS selectors from locale file

gschlager:
The example CSS selectors for embedding shouldn't be part of the locale file. There's no value in translating them. They are just confusing the translators.

styx31 commented on Transifex:
> Does these css selectors really need translation ? Do they concern the discourse CSS strucure or the target website ?


ZogStriP:
I agree. Thanks :+1:

PR #3849, opened 2015-10-11
FIX: admin margin

scossar:
I previously set a margin of -10px on `.admin-contents`. That margin should only have been applied to the mobile view. Applying it to the desktop view pushes `.admin-contents` out of the wrapper.

We don't want this:
screenshot 2015-10-11 12 36 34

This is how it looks with this change:
screenshot 2015-10-11 12 37 14

The negative margin on mobile is to bring some elements right up to the edge of the window:
screenshot 2015-10-11 12 43 58


scossar:
I'll close this and send it again.

PR #3850, opened 2015-10-11
Added san francisco font

fathermerry:


coding-horror:
Thanks for this! These are odd font names, is there anything cleaner to use?

SamSaffron:
Also, I really want some screen shots here of before and after. and some
feedback from @awesomerobot before changing fonts globally.

fathermerry:
There's no cleaner way to write it. Apple didn't include the new font in font book by default. I found out here myself https://css-tricks.com/i-left-my-system-fonts-in-san-francisco/

fathermerry:
I think making use of the system font ensures some kind on consistency with the operating system, and is not so much a global font change as it is _supporting_ the new El Capitan. I might be wrong though. Here are screenshots.
Before
After

coding-horror:
It's a really nasty hack in the current form.. if we could specify it as a normal font-family, maybe.. I just don't like shipping this level of nasty hack in the box. People can add it to Admin, Customize, CSS as needed. Appreciate the effort though. (I thought it would be simpler!)

fathermerry:
That’s okay. Thanks for the consideration


Cheers,Opemipo.

PR #3851, opened 2015-10-12
Fix impersonate redirect on subfolder install

xfalcox:
Impersonate was redirecting to base site, instead of Discourse when using subfolder install.


riking:
That seems odd... why call getURL() then?

ZogStriP:
Yeah I agree with @riking here. @nlalonde what do you think?

xfalcox:
Now that you say :D

I'm not 100% familiarized with Discourse code base so wasn't sure about getUrl side effects. 

So I'm changing it to / or Discourse.BaseUri

nlalonde:
I agree. The `Discourse.getURL` isn't needed there.

PR #3854, opened 2015-10-15
FIX: Race condition when wrapping `PostCreator#create` in a transaction.

tgxworld:
https://meta.discourse.org/t/automated-responses-to-flags-not-generating-notification/31151/3

### DIFF
--- a/app/jobs/regular/post_alert.rb
+++ b/app/jobs/regular/post_alert.rb
@@ -3,7 +3,7 @@ module Jobs
 
     def execute(args)
       # maybe it was removed by the time we are making the post
-      if post = Post.find_by(id: args[:post_id])
+      if post = Post.find(args[:post_id])
         # maybe the topic was deleted, so skip in that case as well
         PostAlerter.post_created(post) if post.topic
       end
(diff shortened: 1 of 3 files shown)


app/jobs/regular/post_alert.rb
```diff
@@ -3,7 +3,7 @@ class PostAlert < Jobs::Base
 
     def execute(args)
       # maybe it was removed by the time we are making the post
-      if post = Post.find_by(id: args[:post_id])
+      if post = Post.find(args[:post_id])
```
tgxworld: I think it is better to raise an error for visibility then let it silently bypass the job.

spec/models/post_action_spec.rb
```diff
@@ -516,6 +516,22 @@
       topic.reload
       expect(topic.posts.count).to eq(1)
     end
+
+    it "should create a notification in the related topic" do
```
tgxworld: adding an additional test to cover a success scenario

app/jobs/regular/post_alert.rb
riking: Actually, wouldn't this alone cause it to work? Because Sidekiq automatically retries failed jobs. It'd just inflate the failed jobs count.

tgxworld: This was one of the solution initially but it wasn't solving the root cause of the problem. It'll also introduce a delay as to when the notification is sent out.

ZogStriP:
Thanks :+1:

PR #3855, opened 2015-10-15
FIX: Fixing margin-top in .card-content element when have an image background

diegoeis:
When the user have an image background, the .card-content element have
a weird margin-top: 80px; I just remove this margin. See the evidence in image:

screen-shot-2015-10-15-at-8 45 16-pm


coding-horror:
Er.. what? That's intentional. Not sure why you're removing it?

diegoeis:
When, I saw it, I thought it was a problem in design. For me, this look like a layout error. :-\
So, if it is not a problem, we can close this PR. ;-)

PR #3856, opened 2015-10-16
FIX: Whitelist parentheses and brackets in autocomplete.

tgxworld:
https://meta.discourse.org/t/default-emoji-plugin-how-to-disable/3415/14


tgxworld:
@ZogStriP hmm just saw the post on meta. This only handles english characters :(

PR #3858, opened 2015-10-16
UX: Trigger autocompleter only for specific characters.

tgxworld:
https://meta.discourse.org/t/default-emoji-plugin-how-to-disable/3415/18

### DIFF
--- a/app/assets/javascripts/discourse/lib/autocomplete.js.es6
+++ b/app/assets/javascripts/discourse/lib/autocomplete.js.es6
@@ -6,6 +6,8 @@
 
 export var CANCELLED_STATUS = "__CANCELLED";
 
+const allowedLettersRegex = /[\s\t\[\{\(]/;
+
 var keys = {
   backSpace: 8,
   tab: 9,
@@ -287,7 +289,7 @@ export default function(options) {
     if (options.key && e.which === options.key.charCodeAt(0)) {
       caretPosition = Discourse.Utilities.caretPosition(me[0]);
       var prevChar = me.val().charAt(caretPosition - 1);
-      if (!prevChar || /[^\w\)\]]/.test(prevChar)) {
+      if (!prevChar || allowedLettersRegex.test(prevChar)) {
         completeStart = completeEnd = caretPosition;
         updateAutoComplete(options.dataSource(""));
       }
@@ -341,7 +343,7 @@ export default function(options) {
         stopFound = prev === options.key;
         if (stopFound) {
           prev = me[0].value[c - 1];
-          if (!prev || /[^\w\)\]]/.test(prev)) {
+          if (!prev || allowedLettersRegex.test(prev)) {
             completeStart = c;
             caretPosition = completeEnd = initial;
             term = me[0].value.substring(c + 1, initial);


app/assets/javascripts/discourse/lib/autocomplete.js.es6
```diff
@@ -6,6 +6,9 @@
...
...}-\u{9FCC}\u{A000}-\u{A48C}\u{A4D0}-\u{A4FD}\u{A500}-\u{A60C}\u{A610}-\u{A61F}\u{A62A}\u{A62B}\u{A640}-\u{A66E}\u{A67F}-\u{A697}\u{A6A0}-\u{A6E5}\u{A717}-\u{A71F}\u{A722}-\u{A788}\u{A78B}-\u{A78E}\u{A790}-\u{A793}\u{A7A0}-\u{A7AA}\u{A7F8}-\u{A801}\u{A803}-\u{A805}\u{A807}-\u{A80A}\u{A80C}-\u{A822}\u{A840}-\u{A873}\u{A882}-\u{A8B3}\u{A8F2}-\u{A8F7}\u{A8FB}\u{A90A}-\u{A925}\u{A930}-\u{A946}\u{A960}-\u{A97C}\u{A984}-\u{A9B2}\u{A9CF}\u{AA00}-\u{AA28}\u{AA40}-\u{AA42}\u{AA44}-\u{AA4B}\u{AA60}-\u{AA76}\u{AA7A}\u{AA80}-\u{AAAF}\u{AAB1}\u{AAB5}\u{AAB6}\u{AAB9}-\u{AABD}\u{AAC0}\u{AAC2}\u{AADB}-\u{AADD}\u{AAE0}-\u{AAEA}\u{AAF2}-\u{AAF4}\u{AB01}-\u{AB06}\u{AB09}-\u{AB0E}\u{AB11}-\u{AB16}\u{AB20}-\u{AB26}\u{AB28}-\u{AB2E}\u{ABC0}-\u{ABE2}\u{AC00}-\u{D7A3}\u{D7B0}-\u{D7C6}\u{D7CB}-\u{D7FB}\u{F900}-\u{FA6D}\u{FA70}-\u{FAD9}\u{FB00}-\u{FB06}\u{FB13}-\u{FB17}\u{FB1D}\u{FB1F}-\u{FB28}\u{FB2A}-\u{FB36}\u{FB38}-\u{FB3C}\u{FB3E}\u{FB40}\u{FB41}\u{FB43}\u{FB44}\u{FB46}-\u{FBB1}\u{FBD3}-\u{FD3D}\u{FD50}-\u{FD8F}\u{FD92}-\u{FDC7}\u{FDF0}-\u{FDFB}\u{FE70}-\u{FE74}\u{FE76}-\u{FEFC}\u{FF21}-\u{FF3A}\u{FF41}-\u{FF5A}\u{FF66}-\u{FFBE}\u{FFC2}-\u{FFC7}\u{FFCA}-\u{FFCF}\u{FFD2}-\u{FFD7}\u{FFDA}-\u{FFDC}\)\]]/u;
```
ZogStriP: I would not call that "`autoCompleteRegex`" but rather "`lettersRegex`" or something similar.

coding-horror:
I don't know if we want this giant regex.. if that's the best way, it's not worth doing

tgxworld:
@coding-horror We could use [XRegExp](https://github.com/slevithan/xregexp). :smile:

SamSaffron:
I am not following ... why not simply a regex that looks at stuff that we do not want to trigger autocomplete for anymore? [^'/\s] or something like that ... just assume that if its not space or some sort of whitelisted punctuation it is good.

tgxworld:
The regex is doing that. We don't want auto complete to trigger after any characters,  )and ].

This PR is to fix the problem where letter characters from other languages are triggering auto complete while /[a-zA-Z0-9_]/ is not. Therefore, we need to exclude letter characters from other languages as well. The other approach I thought of was to maintain a list of characters that would trigger auto complete but we would probably end up with quite a long list as well.

SamSaffron:
I am so confused, can you do a quick synopsis of what the original fix was
that cause the regression with some examples.

tgxworld:
### The original problem
Auto complete was only triggered after spaces and users were complaining about not being able to trigger autocomplete after a `.` or `(`. 

### The first fix
I submitted a PR to allow autocomplete to trigger after non-word characters.

###  Problem
As reported here: https://meta.discourse.org/t/default-emoji-plugin-how-to-disable/3415/18?u=tgxworld

bug_report-6

### The fix
We can't depend on `\w` or `\W` since it only covers `[a-zA-Z0-9_]`. Therefore, we have to exclude all letters in the Unicode letter categories. However, JavaScript doesn't support Unicode categories in regexp so we're using a regexp that covers the letter categories.

SamSaffron:
Then why no change it to allow explicitly after a `(` '[` and some other
chars? seems like an easier fix.

SamSaffron:
I think we should just revert the original fix and whitelist a few more conditions for allowing `@`

explicitly allow `@` after space, tab, newline `[` `(` and `{` I think this covers the majority of the cases and does not require a huge regex.

tgxworld:
@SamSaffron Updated :)

ZogStriP:
MUCH MUCH better :+1:

PR #3861, opened 2015-10-18
why not free software

n3k00n3:
Why not free sofware ?


n3k00n3:
http://www.gnu.org/philosophy/free-sw.html

n3k00n3:
I signed

eviltrout:
I think more people understand the "Open Source" label than free software.

PR #3862, opened 2015-10-20
FIX: Emoji toolbar too wide on mobile.

tgxworld:
This was removed in https://github.com/discourse/discourse/commit/94b60e62a263322ee7e4c0765b8d1307575f37e0#diff-d63881c9f1e1fbeeb8d94b2d881120e1L240

cc/ @eviltrout


eviltrout:
Oops, totally my fault and an accident. Thanks for the fix.

tgxworld:
@eviltrout No worries :stuck_out_tongue: Just checking if it was intended :smile:

PR #3863, opened 2015-10-20
FIX: Topic#last_posted_at was not being set when changing topic times…

tgxworld:
…tamp.

meta: https://meta.discourse.org/t/changing-timestamp-doesnt-update-topic-list-on-front-page/34684


ZogStriP:
Thanks :+1:

PR #3868, opened 2015-10-21
Fix base and vBulletin importer

gschlager:
https://meta.discourse.org/t/vbulletin-import-fails-on-import-topics/34683
https://meta.discourse.org/t/vbulletin-migration-issue/34657


eviltrout:
Thanks for the fixes :)

PR #3869, opened 2015-10-24
Add templates to resolver path for plugin admin templates

gdpelican:


app/assets/javascripts/discourse/ember/resolver.js.es6
```diff
@@ -188,7 +188,8 @@ export default Ember.DefaultResolver.extend({
       const dashed = decamelized.replace(/_/g, '-');
       return Ember.TEMPLATES[decamelized] ||
              Ember.TEMPLATES[dashed] ||
-             Ember.TEMPLATES[dashed.replace('admin-', 'admin/')];
+             Ember.TEMPLATES[dashed.replace('admin-', 'admin/')] ||
+             Ember.TEMPLATES[dashed.replace('admin-', 'admin/templates/')];
```
gdpelican: I was attempting to add an admin template to a plugin, and found that the resolver didn't pick it up because 'decamelized' here was `javascripts/admin_chats`, which means it was looking for `javascripts/admin/chats`, when the template is located at `javascripts/admin/templates/chats`

I left the existing `dashed.replace('admin-')` bit because I wasn't sure if that was a mistake or if other routes use that correctly.

eviltrout: I'm a little confused -- how is it finding our other templates then?

gdpelican: I haven't grokked the plugin assets system 100% yet, but it seems like the difference may be that plugin assets are stored under `javascripts/admin/templates/`, whereas discourse core assets are stored under just `admin/templates/`

So, for plugin assets:
tries this: `javascripts/admin/chat` (no good because `javascripts/admin/chat` isn't a thing)
and this: `admin/templates/chat` (no good because we'd need `javascripts` at the root there)

However, for normal assets:
tries this: `javascripts/admin/groups` (no good, because it's not in the javascripts/ folder)
and this: `admin/templates/groups` (works!)

I think this is unique to the admin assets, because they're a little weird, all the others will come in with dot separators (eg `common.header`), while the admin ones are camelized (eg `adminGroups`).

There's even a few hacks in the resolver already around admin routes, such as `resolver.js.es6#179`:
```
    if (decamelized === "javascripts/admin") {
      return Ember.TEMPLATES['admin/templates/admin'];
    }
```

gdpelican: Oh, maybe of note: The way I'm adding that admin route is to put these lines at the end of the `admin-route-map.js.es6` `map()` function, which seemed to be consistent with the rest of the file:

```
    this.resource('adminChats', { path: '/chats' }, function() {
      this.resource('adminChat', { path: '/:id' })
    });
```

eviltrout: Can you not follow this approach? It's our best supported interface for plugins to add admin pages:

https://meta.discourse.org/t/beginners-guide-to-creating-discourse-plugins-part-5-admin-interfaces/31761

gdpelican:
Ah, thanks for that!

It's not ideal (I was hoping to use the sidebar to display the existing chats, similar to the 'groups' admin page), but I think it's totally workable.

PR #3872, opened 2015-10-27
#compact! can return nil if no changes were made

carsonreinke:


lib/memory_diagnostics.rb
```diff
@@ -22,6 +22,7 @@ def self.compare(from=nil, to=nil)
     diff = diff.map do |id|
       ObjectSpace._id2ref(id) rescue nil
     end.compact!
```
eviltrout: Actually wouldn't this always be `nil`? 

`compact!` never returns the object it seems.

```
[1, 2, 3, 1].compact!
=> nil
```

carsonreinke:
So basically it was never running unless there were changes made.  Let me update the PR.

eviltrout:
Great thanks!

carsonreinke:
Thanks!

PR #3877, opened 2015-10-29
Update README.md

freakomaniac:


ZogStriP:
Thanks but I don't feel it's necessary to precise that point.

PR #3885, opened 2015-10-30
Digital Ocean should be DigitalOcean https://www.digitalocean.com/leg…

seth-reeser:
…al/terms/


eviltrout:
Sounds reasonable thanks!

PR #3889, opened 2015-11-03
FIX: avatar not shown correctly when there is a port in the URL

wil93:
I noticed that if the URL contains a port (e.g. `localhost:4000`, when using Vagrant), the little avatar beside a quote is not shown. Before-after screenshots:

little avatar not showing

In the Vagrant case, the computed path was `//localhost:3000/user_avatar/localhost/eviltrout/40/1_1.png`. So the port is there, is just wrong for some reason. In my real word use case the path was just `//example.com/user_avatar/...` without the port. However, now it's just: `/user_avatar/localhost/eviltrout/40/1_1.png` so I guess it's better either way?

### DIFF
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -31,7 +31,7 @@ module PrettyText
         avatar_template = user.avatar_template
       end
 
-      UrlHelper.schemaless UrlHelper.absolute avatar_template
+      Discourse.base_uri avatar_template
     end
 
     def is_username_valid(username)
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -1,6 +1,5 @@
 require 'v8'
 require 'nokogiri'
-require_dependency 'url_helper'
 require_dependency 'excerpt_parser'
 require_dependency 'post'
 


coding-horror:
Always love a PR with a great explanation and screenshot! :+1:

lib/pretty_text.rb
```diff
@@ -1,6 +1,5 @@
 require 'v8'
 require 'nokogiri'
-require_dependency 'url_helper'
```
eurushibata: teste

ZogStriP:
Discourse, like any other standard rails application, runs on port `3000` in `development` mode. The issue you're experiencing is because you're running inside Vagrant. Indeed, Vagrant is exposing port `3000` to port `4000` so that your machine can access Discourse. Since we're forcing absolute URLs on all images, they all have port `3000` in their URL which means that your browser can't access them. I'm afraid I can't accept this fix since we definitely want absolute URLs on images/avatars.

wil93:
> Since we're forcing absolute URLs on all images, they all have port `3000` in their URL

In my use case, however, discourse is run behind port 8000 (and those images don't have any port in the URL). Is there a way to use port 8000 in the URLs?

> we definitely want absolute URLs on images/avatars

Why is that?

ZogStriP:
> In my use case, however, discourse is run behind port 8000 (and those images don't have any port in the URL). Is there a way to use port 8000 in the URLs?

Have you tried setting the `port` site setting in the Developper tab?

> > we definitely want absolute URLs on images/avatars

> Why is that?

Because it makes post processing the posts much easier ;)

PR #3890, opened 2015-11-07
Clarify what happens when newuser_max_links is exceeded

dandv:
Supersedes #3845


SamSaffron:
oh my, the account will not get blocked, all that happens is that the user is not allowed to make the post and displayed a warning ... "too many links in your post" that is all.

PR #3891, opened 2015-11-07
FIX: Use selected text in the link if text is already selected, otherwise, put the URL

cpradio:


eviltrout:
This should really have a test in `d-editor-test.js.es6`

cpradio:
Test added :smile:

test/javascripts/components/d-editor-test.js.es6
```diff
@@ -193,6 +193,23 @@ testCase('link modal (simple link)', function(assert) {
   });
 });
 
+testCase('link modal (simple link) with selected text', function(assert, textarea) {
+  this.set('value', "hello world.");
```
eviltrout: If you are using the `testCase` helper it should default to `hello world.`. Also you shouldn't need an `andThen` to set the selectionStart and End, other tests don't  need it!

cpradio: https://github.com/cpradio/discourse/blob/cbde168281e2c7a37a0a4701677d8a0e0ec32344/test/javascripts/components/d-editor-test.js.es6#L98-L101

This uses andThen to select text...

cpradio:
Updated and rebased to clean up the commits

eviltrout:
Thanks!

PR #3892, opened 2015-11-09
FIX: Update Quote button to use quote_text and to insert text when nothing is selected

cpradio:


cpradio:
Going back to the drawing board on this too

PR #3895, opened 2015-11-10
Update ning.rb

rblagahit:


ZogStriP:
Thanks but changing this line is specific to every install and thus should not be commited into core.

PR #3897, opened 2015-11-12
FIX: position admin-nav absolutely

scossar:
The admin-nav needs to be positioned absolutely on small screens for it to work as a slide-out menu. It's position has been changed to static. I'm not sure if that was done for a reason. Positioning it absolutely doesn't seem to cause any problems anywhere. This PR changes it back to `position: absolute`.

How it looks now:

screenshot 2015-11-12 10 08 35

How it looks with `position: absolute`

screenshot 2015-11-12 10 09 46

screenshot 2015-11-12 10 09 56


ZogStriP:
Thanks :+1:

PR #3898, opened 2015-11-13
UX: Use localized date format in digest email

gschlager:
Use localized date format in digest email instead of the format introduced in https://github.com/discourse/discourse/commit/734c272de81baac8a70801e257a994a8a6e9f616

PR https://github.com/discourse/discourse/pull/3787 has been merged - the date formats are localizable. I liked the idea of omitting the year when it's the current year, so I left that in.
See also: https://meta.discourse.org/t/change-of-date-in-digest-email-body/34983

This is what the digest preview looks like for the English locale:
image
image

And here's the German locale:
image
image

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -42,11 +42,10 @@ class UserNotifications < ActionMailer::Base
   end
 
   def short_date(dt)
-    current = Time.now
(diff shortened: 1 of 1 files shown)


ZogStriP:
Thanks :+1:

PR #3899, opened 2015-11-13
UX: Show translated modifier keys in editor shortcuts

gschlager:
Modifier keys should be translatable. For example, in German we use
- "Umschalt" instead of "Shift"
- "Strg" instead of "Ctrl"


SamSaffron:
thanks!

PR #3900, opened 2015-11-16
migrate_to_s3: check for missing local files

uppfinnarn:
If an image is already on S3, but not available locally (eg. if they're already migrated), `path` is nil. This crashes the task, because File.exists?(nil) is not valid.


ZogStriP:
Thanks :+1:

PR #3903, opened 2015-11-17
FIX: Pressing Ctrl+F twice should close Discourse search window

gschlager:
Make it work with Mousetrap 1.5.3 that was added in bb21902954511c66b622ce36ca3aa9afbe49db4b.
Overwriting the `stopCallback` was moved to the `prototype` (https://github.com/ccampbell/mousetrap/issues/288).


eviltrout:
Thanks!

PR #3904, opened 2015-11-17
FEATURE: New setting to allow all caps posts

featheredtoast:
Adds a setting to ignore text_sentinel's check on all caps content.

### DIFF
--- a/lib/text_sentinel.rb
+++ b/lib/text_sentinel.rb
@@ -73,7 +73,7 @@ class TextSentinel
 
   def seems_quiet?
     # We don't allow all upper case content in english
-    not((@text =~ /[A-Z]+/) && !(@text =~ /[^[:ascii:]]/) && (@text == @text.upcase))
+    SiteSetting.allow_uppercase_posts || not((@text =~ /[A-Z]+/) && !(@text =~ /[^[:ascii:]]/) && (@text == @text.upcase))
   end
 
 end
--- a/spec/components/text_sentinel_spec.rb
+++ b/spec/components/text_sentinel_spec.rb
@@ -74,6 +74,11 @@ describe TextSentinel do
       expect(TextSentinel.new(valid_string.upcase)).not_to be_valid
     end
 
+    it "allows all caps topics when loud posts are allowed" do
+      SiteSetting.stubs(:allow_uppercase_posts).returns(true)
+      expect(TextSentinel.new(valid_string.upcase)).to be_valid
+    end
+
     it "enforces the minimum entropy" do
       expect(TextSentinel.new(valid_string, min_entropy: 16)).to be_valid
     end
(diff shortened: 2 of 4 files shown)


SamSaffron:
Did you confirm this works for both topic title and body? 

also let's rename to "allow_uppercase_posts" ... its easier to search for.

featheredtoast:
I traced both title and body validations to the same piece of code, and made the assumption. Testing now, just to be certain.

Meantime, I've pushed up the rename commit for review. Would you prefer this rebased/squashed after confirming?

featheredtoast:
PS, confirmed as working for both title and body.

lib/text_sentinel.rb
```diff
@@ -45,7 +45,7 @@ def valid?
     seems_meaningful? &&
     seems_pronounceable? &&
     seems_unpretentious? &&
-    seems_quiet?
+    (seems_quiet? || SiteSetting.allow_uppercase_posts)
```
ZogStriP: I would write `(SiteSetting.allow_uppercase_posts || seems_quiet?)` so that we don't even call the `seems_quiet?` method when we allow all uppercase posts ;)

ZogStriP:
> Meantime, I've pushed up the rename commit for review. Would you prefer this rebased/squashed after confirming?

Yeah, we'd :heart: that ;)

featheredtoast:
Done!

spec/components/text_sentinel_spec.rb
```diff
@@ -74,6 +74,11 @@
       expect(TextSentinel.new(valid_string.upcase)).not_to be_valid
     end
 
+    it "allows all caps topics when loud posts are allowed" do
+      SiteSetting.stubs(:allow_uppercase_posts).returns(true)
```
eviltrout: Instead of mocking here, could you just say `SiteSetting.allow_uppercase_posts = true`? We reset changes to site settings after every test.

lib/text_sentinel.rb
```diff
@@ -45,7 +45,7 @@ def valid?
     seems_meaningful? &&
     seems_pronounceable? &&
     seems_unpretentious? &&
-    seems_quiet?
+    (SiteSetting.allow_uppercase_posts || seems_quiet?)
```
eviltrout: To make it consistent with the the other `seems_` methods, could you move the `SiteSetting` check inside the `seems_quiet?` -- it'll just read better

spec/components/text_sentinel_spec.rb
featheredtoast: I don't know if I agree with that one. I'm following the stubbing conventions of what every other test in that file are doing.

featheredtoast:
Pushed `SiteSetting` check inside `seems_quiet?` method.

lib/text_sentinel.rb
featheredtoast: I agree, I like that better too. Updated.

spec/components/text_sentinel_spec.rb
eviltrout: Sure if every other test does it that way, I'm okay with it. But we prefer not to mock in the future. At some point someone should fix this file.

PR #3905, opened 2015-11-18
FIX: check for site direction before positioning d-editor-modal

scossar:
The d-editor-modal is positioned off the screen for RTL layouts. This checks the site direction and sets the appropriate direction for positioning. 

Something like this: `const dir = $('html').css('direction') === 'rtl' ? 'right' : 'left';` is also used in `views/user-card.js.es6`. It will come up anytime there is inline horizontal positioning. Maybe there is a better way to check if the layout is rtl?


SamSaffron:
sure, let's try this !

PR #3906, opened 2015-11-23
FIX: Keyboard shortcut to quote reply.

tgxworld:
meta: https://meta.discourse.org/t/q-keyboard-shortcut-not-working/35901


ZogStriP:
Thanks :+1:

PR #3907, opened 2015-11-23
sso login

zonghao:
if wrong 
   Please fix

### DIFF
--- a/lib/user_name_suggester.rb
+++ b/lib/user_name_suggester.rb
@@ -35,10 +35,13 @@ module UserNameSuggester
 
   def self.sanitize_username(name)
     name = ActiveSupport::Inflector.transliterate(name)
-    name = name.gsub(/^[^[:alnum:]]+|\W+$/, "")
-               .gsub(/\W+/, "_")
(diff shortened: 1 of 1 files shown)


ZogStriP:
I'm not sure I understand what issue are you trying to fix? Can you explain?

zonghao:
because of 
FEATURE: relax username rules to allow - and . and leading _[https://github.com/discourse/discourse/commit/262f561a877e0296d7ca7f6ec9f27edd0d30ca8e]

But 
Single sign-on (sso) user didn't work, still is '.' To '_'   (1.4.1 version)
like: zonghao.zhang to zonghao_zhang

So，I found the problem seems to be in this：
/lib/user_name_suggester.rb
in 
def self.sanitize_username(name)
name = ActiveSupport::Inflector.transliterate(name)
name = name.gsub(/^[^[:alnum:]]+|\W+$/, "")
.gsub(/\W+/, "")
.gsub(/^\+/, '')
.gsub(/[-.]{2,}/, "")
name
end

Please check this issue!
Sorry, English is not good.

PR #3908, opened 2015-12-01
Allow compat with ActionSummary for topic model

gdpelican:
I'm having trouble getting `Topic.create({})` to go in a plugin because of this line: 
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/topic.js.es6#L418

Moving ActionSummary out of the Discourse global fixes this for me; any reason not to do this?


ZogStriP:
Actually, there are [plenty of reasons to **do** it](https://meta.discourse.org/t/deprecating-es6-compatibility-layer/35821) ;)

If you fix the eslint error, I'll happilly merge that :+1:

gdpelican:
Updated!

ZogStriP:
Thanks :+1:

PR #3909, opened 2015-12-01
reused value of `Time.now` in a method

humzashah:
`Time.now` was being used to fetch the current time, twice in the
same line. This commit stores the time value in a variable and
reuses it instead of generating/fetching it for the second time.

Additionally, a guard clause in the same method is substituted by
an `if/else/end` block for clarity.


SamSaffron:
Thanks!

PR #3911, opened 2015-12-05
FIX: On clicking edit category button, notification level is set to default

abbasfaisal:
This fixes 
https://meta.discourse.org/t/cancelling-edit-category-screen-changes-following-status/31861/2.

The received value of notification_level is set to null so we need to set it up before returning category.

### DIFF
--- a/app/serializers/basic_category_serializer.rb
+++ b/app/serializers/basic_category_serializer.rb
@@ -38,4 +38,7 @@ class BasicCategorySerializer < ApplicationSerializer
     scope && scope.can_edit?(object)
   end
 
+  def notification_level
+    object.notification_level || CategoryUser.where(user: object.user, category: object).first.try(:notification_level)
+  end
 end


app/controllers/categories_controller.rb
```diff
@@ -73,6 +73,8 @@ def show
     if Category.topic_create_allowed(guardian).where(id: @category.id).exists?
       @category.permission = CategoryGroup.permission_types[:full]
     end
+    @category.notification_level = CategoryUser.where(category: @category.id).first.notification_level
```
tgxworld: You'll have to make sure that the `CategoryUser` belongs to the `current_user` as well. :smile:

tgxworld:
On a higher level, I think we should fix this at `BasicCategorySerializer`. Basically, we need to check if the category's `notification_level` has been set. If not, we have to fetch it.

abbasfaisal:
@tgxworld Done.

app/serializers/basic_category_serializer.rb
```diff
@@ -38,4 +38,7 @@ def include_can_edit?
     scope && scope.can_edit?(object)
   end
 
+  def notification_level
+    object.notification_level ? object.notification_level : CategoryUser.where(user: object.user, category: object).first.try(:notification_level)
```
tgxworld: I think this can be simplified to `object.notification_level || CategoryUser.where....` :smile:

abbasfaisal:
Done.

tgxworld:
Looks good to me. Thanks for the PR @abbasfaisal!

PR #3912, opened 2015-12-06
FIX: Admin logging should not log permissions if none has been set.

tgxworld:
Meta: https://meta.discourse.org/t/extra-admin-logs-generated-on-category-edit/36345


tgxworld:
Ah I broke the build will fix.

PR #3915, opened 2015-12-07
FIX: Avoid bang methods in Disqus importer.

tgxworld:
Meta: https://meta.discourse.org/t/disqus-importer-fails-undefined-method/36200

### DIFF
--- a/script/import_scripts/disqus.rb
+++ b/script/import_scripts/disqus.rb
@@ -151,8 +151,7 @@ class DisqusSAX < Nokogiri::XML::SAX::Document
         @threads.delete(id)
       else
         # Normalize titles
-        t[:title].gsub!(@strip, '') if @strip.present?
-        t[:title].strip!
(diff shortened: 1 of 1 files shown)


script/import_scripts/disqus.rb
```diff
@@ -151,8 +151,7 @@ def normalize
         @threads.delete(id)
       else
         # Normalize titles
-        t[:title].gsub!(@strip, '') if @strip.present?
-        t[:title].strip!
+        t[:titlet] = [:title].gsub(@strip, '').strip if @strip.present?
```
ZogStriP: `titlet`?

tgxworld: oops will fix!

yyhmsg:
Thanks @tgxworld for the quick fix.  The importer seems to work now - or at least my dry_run worked, I'll try some more tests first and then let you know if there are any issues with production.

tgxworld:
Merging this in :smile:

PR #3917, opened 2015-12-09
Change to use weak reference incase object is collected by GC

carsonreinke:
I was getting some seg faults from referencing an already collected object (I have no idea how this is even possible).  Changing to use a `WeakRef` will just raise an exception when object has already been GC'd.

Obviously, this will introduce a ton more objects if there is a big diff.

### DIFF
--- a/lib/memory_diagnostics.rb
+++ b/lib/memory_diagnostics.rb
@@ -20,7 +20,7 @@ module MemoryDiagnostics
 
     require 'objspace'
     diff = diff.map do |id|
-      ObjectSpace._id2ref(id) rescue nil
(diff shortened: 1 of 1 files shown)


coding-horror:
I am unclear why this is needed?

carsonreinke:
I am trying to create a good test case, for whatever reason, MRI would segfault after using the object from `ObjectSpace._id2ref(id)`.

carsonreinke:
I could not create a good test case and it no longer happens in MRI 2.3.0, so I give up.  I thought for a moment it was something to do with the fork recording the ObjectSpace state, but I could never prove it.  Sorry for the bother.

PR #3920, opened 2015-12-14
FIX: Emoji in Discourse onebox is wrapped in square brackets.

tgxworld:
screenshot from 2015-12-14 21 47 08

### DIFF
--- a/lib/onebox/engine/discourse_local_onebox.rb
+++ b/lib/onebox/engine/discourse_local_onebox.rb
@@ -63,7 +63,7 @@ module Onebox
             topic = post.topic
             slug = Slug.for(topic.title)
 
-            excerpt = post.excerpt(SiteSetting.post_onebox_maxlength)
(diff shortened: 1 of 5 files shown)


lib/onebox/engine/discourse_local_onebox.rb
```diff
@@ -51,7 +51,7 @@ def to_html
             topic = post.topic
             slug = Slug.for(topic.title)
 
-            excerpt = post.excerpt(SiteSetting.post_onebox_maxlength)
+            excerpt = post.excerpt(SiteSetting.post_onebox_maxlength, { keep_emoji_codes: true })
```
tgxworld: This case is a little special because we're getting the cooked version and then passing it again to the markdown engine. Because emojis have `alt=':heart:'`, the markdown engine is coverting `:heart:` to another emoji image.

PR #3921, opened 2015-12-14
Add Slovak locale

gschlager:
https://meta.discourse.org/t/slovak-translation-after-switching-to-sk-locale-discourse-home-page-is-not-loaded/36365

Translation progress of client is 99%, server is at 60%.


ZogStriP:
Ďakujem :+1:

PR #3922, opened 2015-12-14
username as from in emails

featheredtoast:
Allow username to display in from field for email notifications even when full name fields are set to display in posts.


featheredtoast:
A simple implementation for:

https://meta.discourse.org/t/is-it-possible-to-make-usernames-as-email-from-field/36704/1

SamSaffron:
Failures are unrelated, merging in, its a nice clean change. Thanks!

PR #3923, opened 2015-12-15
Have DiskSpace.{free,used} use BigDecimal instead of directly doing F…

envygeeks:
…ixnum#to_i

As it stands the way Discourse handles Disk usage and free space reporting it can lead to nearly 0 disk space being reported to Discourse, this is because when a the command is ran and returned (at first through `awk`, the latter is for consistency) it returns notation, `Fixnum#to_i` returns the number to the left rather than anything from the left and right combined and expanded, leading to something like `8.76138e+09` (what is returned on our systems with that command) to be 8, meaning we only have 8 bytes available to Discourse according to it. By using BigDecimal to wrap that and then run `Fixnum#to_i` we get the proper number `8761380000`, you can optionally also just run `#to_f.to_i` but that looks nasty compared to just using STDLib's BigDecimal.

### DIFF
--- a/lib/disk_space.rb
+++ b/lib/disk_space.rb
@@ -1,3 +1,5 @@
+require "bigdecimal"
+
 class DiskSpace
 
   extend ActionView::Helpers::NumberHelper
@@ -57,11 +59,12 @@ class DiskSpace
   protected
 
   def self.free(path)
-    `df -Pk #{path} | awk 'NR==2 {print $4 * 1024;}'`.to_i
+    number = `df -Pk #{path} | awk 'NR==2 {print $4 * 1024;}'`.strip
+    BigDecimal.new(number).to_i
   end
 
   def self.used(path)
-    `du -s #{path}`.to_i * 1024
+    number = `du -s #{path}`
+    BigDecimal.new(number).to_i * 1024
   end
-
 end


coding-horror:
Is this ok @ZogStriP ?

ZogStriP:
It is. I was waiting for the CLA

coding-horror:
ok @envygeeks can you sign the CLA?

envygeeks:
@coding-horror @ZogStriP I will review it today.

tgxworld:
@envygeeks Any updates on this :)

envygeeks:
So.... not signing the CLA means a loss of attribution?

ZogStriP:
Sorry, I forgot to add your username to the commit :(

coding-horror:
Can we reverse patch the commit? Does Git allow that? I got the impression you could change history at will in GitHub.

SamSaffron:
Apologies, I rolled back the fix @envygeeks , we can not accept it without you signing the CLA it was an error on our part to take that diff and apply it, this will not happen again. 

https://github.com/discourse/discourse/commit/7138b74110b2c5d311bdfb021b1587ac596e32e3

If you are open to signing the CLA we can reopen this and merge in your fix, otherwise we need to figure out a different way of fixing this.

envygeeks:
@SamSaffron the CLA has been signed.

SamSaffron:
thanks, I went with this

https://github.com/discourse/discourse/commit/ee823305a1b8c618ecd8676945c73375da8be665

and attributed you. 

Having a tough time reproducing the issue though. 

```
irb(main):007:0> `echo 99999213123121937123981273981273912873192873312312399999 | awk '{ print $1 * 8282722312312312312312313281231231232222227287; }'`.to_i
=> 828265713748556322186039391453181644449283843146096108129993462121255051385741022935555734106961608704
```

envygeeks:
@SamSaffron Ubuntu Cloud uses `mawk` instead of `gawk` by default, consumer Ubuntu (non-server) uses `gawk`. The result you are getting is probably from `gawk`.  This should be reproducible on your own images (unless you guys bring in a dependency that wants gawk... we do not use the official images because we scale much differently than you allow,) however you can replicate it on any system with (mawk) using:`df -Pk / | mawk 'NR==2 {print $4 * 1024;}'`

envygeeks:
@SamSaffron Here are more examples:

```
jordon ~/ % docker run --rm -it envygeeks/ubuntu:lts bash -ic "df -Pk / | \
  awk 'NR==2 {print \$4 * 1024;}'"

6.71604e+10
```

```
jordon ~/ % docker run --rm -it ubuntu:14.04.3 bash -ic "df -Pk / | \
  awk 'NR==2 {print \$4 * 1024;}'"

6.71604e+10
```

SamSaffron:
I see, oh my there are so many flavors of awk out there, it's landmine central. I wonder if I should amend our image to install gawk ...

envygeeks:
@SamSaffron Using BigDecimal like we did should solve all the differences because it can handle both normal int/float and notated int/float.  Actually we tend to do that with all numbers provided by the system for that exact reason, it's easier to expect notation.

SamSaffron:
I see, interestingly this is not happening with our image: 

```
docker run --rm -it envygeeks/ubuntu:lts  bash -ic "df -Pk / | \
  awk 'NR==2 {print \$4 * 1078324223423424324982340984320983342347424;}'"

> 2.0175e+49

docker run --rm -it discourse/base:1.0.15  bash -ic "df -Pk / | \
  awk 'NR==2 {print \$4 * 10783242234234243342347424;}'"
203848352179566282693453010698240

```

Nonetheless, I am pretty sure the fix I put in place should fix your system as well.

envygeeks:
@SamSaffron I just did an rdepends, I see why you guy have gawk now, ya'll carry a lot more dependencies than we do (which is understandable, you cater to a different audience.)

I'm happy to have a fix none-the-less, now Jekyll Talk will will be able to accept uploads!

ZogStriP:
I sincerely apologize @envygeeks. It wasn't my intention to _steal_ your code. I was too eager to have this fix merged... Won't happen again.

envygeeks:
It's alright @ZogStriP I assumed somebody had just forgotten to `git am`

PR #3925, opened 2015-12-16
FIX: Update all does not trigger callbacks.

tgxworld:
meta: https://meta.discourse.org/t/notifications-not-clearing-properly-unless-visited-from-menu/19041/6


ZogStriP:
:+1:

PR #3926, opened 2015-12-16
FIX: Selection logic error in applySurround

riking:
sel.start - hlen is guaranteed to be before the text we're talking about, so it doesn't make any sense. All the other selectText() calls use sel.start + hlen, so that's clearly the correct operation.


riking:
Nope, I read the code wrong. That's the "undo" functionality. The bug is somewhere else....

PR #3927, opened 2015-12-18
Add Discourse.getURL to the url passed into page view events

kellec:
This means that instances in a subdirectory correctly receive the subfolder string as a part of the path.

What we were getting:
image

What this will give us:
image

(The `dp` value overwrites the `dl` value, meaning the analytics end up observing the incomplete path)

cc @cpradio


cpradio:
:+1:

ZogStriP:
Thanks

PR #3928, opened 2015-12-21
Return 204 instead of 404 for favicon.ico requests

jamielinux:
This saves some network bytes (especially if the admin has a heavy custom 404 page). Also saves some disk writes.

### DIFF
--- a/config/nginx.sample.conf
+++ b/config/nginx.sample.conf
@@ -80,11 +80,9 @@ server {
     internal;
   }
 
-  # return a cheap 404 for favicon cause lots of browsers like asking for it
-  # view source and old browsers, this bypasses the rails stack at the expense of
-  # having an ugly 404 page
+  # bypass rails stack with a cheap 204 for favicon.ico requests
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am worried about this , the favicon does not exist it really is a 404, is there any backing anywhere that 204 is desirable here and has no side effects?

jamielinux:
I haven't personally seen any side effects after using this on my own sites for a long time. I've seen it recommended in many places on the Internet (but of course whether the sources are reputable or not is another question). 

https://www.nginx.com/resources/wiki/start/topics/recipes/piwik/

http://markmail.org/message/hst2v4eqhk6vs5j3#query:nginx%20favicon.ico+page:1+mid:c5nwdcpm46xhcenn+state:results

http://maksim.sorokin.dk/it/2010/12/19/handling-noisy-favicon-ico-messages-in-nginx/

coding-horror:
I say let's try it, reasons seem plausible. Thanks!

PR #3930, opened 2015-12-24
UX: Improve copy for toolbar button that triggers emojis modal.

tgxworld:
@coding-horror @eviltrout Is this change ok? The currently text seems to imply that it'll add a ~~smile~~ slightly_smile emoji.


coding-horror:
Hmm, well, the idea is that you show the keyboard shortcut, which would be :) in this case I guess. So perhaps the correct thing to do is

`Emoji :)`

PR #3931, opened 2015-12-24
FIX: Pluralization error when only a pluralization key is overridden.

tgxworld:
https://meta.discourse.org/t/cannot-split-topic-see-error-logs-in-the-topic-body/36962/5
https://meta.discourse.org/t/new-topic-500-internal-server-error/37186


eviltrout:
Thanks so much!

PR #3932, opened 2015-12-26
FIX: Update vulnerable gems.

tgxworld:
We have some vulnerable gems :stuck_out_tongue_closed_eyes: 

https://github.com/jollygoodcode/jollygoodcode.github.io/issues/15


SamSaffron:
I wonder if we even need jquery rails, we don't use that stuff

tgxworld:
hmm `ember-rails` requires it as a dependency

PR #3933, opened 2015-12-26
FIX: Correctly import users and passwords from phpBB 3.1

gschlager:
- [Ruby doesn't support the $2y$ prefix](http://stackoverflow.com/a/20981781) used for bcrypt hashes.
- The importer ignored users that didn't have optional profile fields.


eviltrout:
Thanks this looks great!

PR #3934, opened 2015-12-26
UX: Show translated badge names and badge groupings

gschlager:
UX: Show translated badge names and badge groupings in admin window
FIX: Disallow renaming of system badges
FIX: Disallow renaming/deleting of system badge groupings
FIX: The badge grouping "Other" wasn't marked as "system" group.
UX: Use "modern" button style in badge groupings dialog
UX: Use i18n for texts used in the badge groupings dialog

The admin window for badges uses the current user's locale for badge names and badge groupings instead of always showing the English text. In order for this to work editing the names of system badges and badge groupings is now disabled. AFAIK they are overwritten by DB migrations anyway and there is now the possibility to override them with Customize -> Text Content.

image

https://meta.discourse.org/t/missing-badges-translations/19116

### DIFF
--- a/app/assets/javascripts/admin/controllers/modals/admin-edit-badge-groupings.js.es6
+++ b/app/assets/javascripts/admin/controllers/modals/admin-edit-badge-groupings.js.es6
@@ -2,15 +2,13 @@ export default Ember.Controller.extend({
   needs: ['modal'],
 
   modelChanged: function(){
-
-    var grouping = Em.Object.extend({});
-
-    var model = this.get('model');
-    var copy = Em.A();
+    const model = this.get('model');
+    const copy = Em.A();
+    const store = this.store;
 
     if(model){
       model.forEach(function(o){
-        copy.pushObject(grouping.create(o));
+        copy.pushObject(store.createRecord('badge-grouping', o));
       });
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/admin/controllers/modals/admin-edit-badge-groupings.js.es6
```diff
@@ -1,16 +1,15 @@
...
-
-    var grouping = Em.Object.extend({});
-
     var model = this.get('model');
     var copy = Em.A();
 
     if(model){
       model.forEach(function(o){
-        copy.pushObject(grouping.create(o));
+        copy.pushObject(BadgeGrouping.create(o));
```
eviltrout: It would be better if you used `this.store.createRecord('badge-grouping', o)` - then you don't have to import the model.

gschlager: Good to know. I changed it.

ZogStriP:
:+1:

ZogStriP:
Thanks :+1:

camilleroux:
+1

PR #3935, opened 2015-12-27
Fix require_relative in database_3_1.rb

alx:
The path was incorrect and crashed the import process, changing the path solved this issue


gschlager:
I noticed that too. :wink: 
see PR https://github.com/discourse/discourse/pull/3933

alx:
Thanks, alreaydy fixed in #3933

PR #3938, opened 2015-12-28
Show composer messages popup overtop of preview window

scossar:
This moves the composer messages popup overtop of the composer preview window. This change affects all window heights. It would be possible, but slightly more complicated, to do a media query for the window height and only move the messages for window heights <  725px.

screenshot 2015-12-28 12 28 08

screenshot 2015-12-28 12 32 23

screenshot 2015-12-28 13 55 51


coding-horror:
sure let's try this thank you!

PR #3943, opened 2015-12-30
Allow +1 via email

gdpelican:
Happy to hear some feedback on this one.

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -1,5 +1,6 @@
 require_dependency 'new_post_manager'
 require_dependency 'email/html_cleaner'
+require_dependency 'post_action_creator'
 
 module Email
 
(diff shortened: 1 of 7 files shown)


lib/email/receiver.rb
```diff
@@ -239,6 +244,18 @@ def wrap_body_in_quote(user_email)
       @body = "[quote=\"#{user_email}\"]\n#{@body}\n[/quote]"
     end
 
+    def create_post_action(email_log, type)
+      PostAction.act(email_log.user, email_log.post, type)
```
eviltrout: Calling `act` doesn't seem to ensure the guardian calls are executed.

You'll need to make sure something like [this code](https://github.com/discourse/discourse/blob/master/app/controllers/post_actions_controller.rb#L10) is executed.

(you'll also need a test for this -- what happens when the guardian check fails?)

gdpelican: Alright, I've applied a guardian call to the create_post_action and added a few specs around it; should be good for another look @eviltrout

eviltrout:
Fantastic work, thanks!

PR #3944, opened 2015-12-30
FIX: When 410 is received, display proper error message instead of generic.

abbasfaisal:
This fixes
https://meta.discourse.org/t/clicking-a-link-to-a-private-or-deleted-topic-shows-as-a-network-error/35797.


eviltrout:
Seems good to me, thanks!

PR #3946, opened 2015-12-31
Remove the description about IP based access

omo:
The doc said you can access it using IP address if DSN isn't connfigured yet, but apparently you cannot.
I guess it is because NGINX isn't configured in that way.


omo:
Uh, I cannot sign the CLA without the employer's approval. Will come back later.

coding-horror:
Accessing site via IP works just fine, I do it on every single Digital Ocean install I set up for the self service plan.

omo:
Thanks for the comment! I saw the nginx error page, but probably that was by different cause.

PR #3947, opened 2016-01-03
Add locale for Vietnamese

khoa-le:
I have added Vietnamese locale, and also update Transifex.com translation. Please review it. Thanks!


gschlager:
I'm quite sure the locale should be `vi` instead of `vi_VN`.

khoa-le:
I followed that [link](https://meta.discourse.org/t/how-to-add-a-new-language/14970). Do I need to update `vi_VN` to `vi`?

gschlager:
Yes, unless you tested it with `vi_VN` and it worked.
I didn't test it myself, but I assume it should be `vi` since the [plurals.rb](https://github.com/discourse/discourse/blob/master/config/locales/plurals.rb#L108) uses `vi`.

khoa-le:
Sure, it worked. I deployed to [http://forum.nhadep24.com](http://forum.nhadep24.com), and can change the config to vi_VN.

ZogStriP:
I agree with @gschlager here, we should follow rails' convention and use `vi` instead of `vi_VN`. Can you make that change @khoa-le?

khoa-le:
Thank you, @ZogStriP and @gschlager. I will update it.

gschlager:
One thing is still missing. You need to add the language mapping `vi_VN: vi` to [.tx/config on line 3](https://github.com/discourse/discourse/blob/master/.tx/config#L3).

SamSaffron:
Thanks!

PR #3949, opened 2016-01-05
fixed password validator on equality with email

bgr11n:
Fixed password validator on equality to email, on elsif which check presence of username instead of password


eviltrout:
Thanks!

PR #3950, opened 2016-01-08
Enums that are used in tables need to be stable

techAPJ:
Meta topic => https://meta.discourse.org/t/enums-that-are-used-in-tables-need-to-be-stable/37622


eviltrout:
Looks good :)

techAPJ:
@eviltrout thanks for reviewing!

PR #3951, opened 2016-01-10
Add more precise regex for email address validation

tobiasfeistmantl:


eviltrout:
It looks like this is causing our test suite to fail. There might be something wrong with the regular expression?

ZogStriP:
Also, why do we need a more "precise" regex when the only effective email validation is to actually send an email? ;)

tobiasfeistmantl:
Oh!

I thought that this regex is used for email address validation. But the email validator does the validation with a better regex.

My mistake :)

PR #3954, opened 2016-01-11
save height on small screens

rubo77:
There is a lot of white space between each post, that annoys on small screens.

This will minimize the used space if the screen is lower than 700px in height.

Rework of #3820

### DIFF
--- a/app/assets/stylesheets/desktop/topic-post.scss
+++ b/app/assets/stylesheets/desktop/topic-post.scss
@@ -1026,3 +1026,16 @@ and (max-width : 767px) {
   }
 
 }
+
+@media all
+and (max-height: 700px) {
+
+   .post-menu-area {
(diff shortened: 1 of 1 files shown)


ZogStriP:
Can we get a before/after screenshot?

rubo77:
its pretty much the same like in https://github.com/discourse/discourse/pull/3820 but with another css element used

ZogStriP:
Ok, let's try this then ;)

rubo77:
These two have to be added:

```
.topic-meta-data .post-info {
  margin-top: -2px;
}
.gutter .post-links {
	margin-top: -12px;
}
```

see https://meta.discourse.org/t/username-and-timestamp-at-top-of-posts-too-high-in-latest-discourse/39643

PR #3955, opened 2016-01-12
FIX: Move store for readonly mode to DB.

tgxworld:
* The thread dies when the container is rebuilt causing the site
  to come out of readonly mode.

Not sure if there were other considerations previously.
cc/ @SamSaffron  @ZogStriP

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -932,6 +932,7 @@ en:
     github_client_id: "Client id for Github authentication, registered at https://github.com/settings/applications"
     github_client_secret: "Client secret for Github authentication, registered at https://github.com/settings/applications"
 
+    readonly_mode: "Sets the site to read-only mode."
     allow_restore: "Allow restore, which can replace ALL site data! Leave false unless you plan to restore a backup"
     maximum_backups: "The maximum amount of backups to keep on disk. Older backups are automatically deleted"
     automatic_backups_enabled: "Run automatic backups as defined in backup frequency"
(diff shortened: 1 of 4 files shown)


tgxworld:
@SamSaffron retrieval of values from `SiteSetting` is cache so no extra queries.

ZogStriP:
This was done in Redis because it allowed you to put the site in readonly mode for a little while, restore a backup and still be in readonly mode to confirm that everything works. Not sure that workflow is still useful though...

SamSaffron:
Also super concerned this is going to risk sites in the wild being stuck in
readonly does this allow for crashing restore process?

tgxworld:
>  little while

I kind of need it abit longer. I ran into this when I was migrating meta's db and upon rebuild the site came out of read-only even before I could check that it was connecting properly.

> Also super concerned this is going to risk sites in the wild being stuck in readonly

Hmm that is a good point. I could run a migration script to ensure they remain in the right state though, assuming the redis-key doesn't expire when they are rebuilding the container.

Is there a scenario that this might happen? 

> does this allow for crashing restore process?

Hmm not sure what you mean here.

SamSaffron:
I mean ...

1. backup process enables readonly mode
2. backup process runs out of memory and crashes
3. site is stuck readonly now potentially for hours cause backup in an
unattended processe

This is why the expiring keys in redis are so powerful, process crashes,
site moves out of readonly automatically.

ZogStriP:
I'm going to close this. We need to figure out something else ;)

tgxworld:
:+1:

PR #3956, opened 2016-01-12
FIX: Clashing category slug.

tgxworld:
Meta: https://meta.discourse.org/t/mentioning-categories-instead-of-groups/14697/41?u=tgxworld

### DIFF
--- a/app/models/concerns/category_hashtag.rb
+++ b/app/models/concerns/category_hashtag.rb
@@ -1,0 +1,23 @@
+module CategoryHashtag
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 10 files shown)


app/models/concerns/category_hashtag.rb
```diff
@@ -0,0 +1,23 @@
+module CategoryHashtag
+  extend ActiveSupport::Concern
+
+  SEPERATOR = ":".freeze
```
ZogStriP: SEPARATOR ;)

tgxworld: Will fix! :P

PR #3958, opened 2016-01-13
FIX: Incorrect subfolder authors urls when embedding

halfstrik:
Fix URLs for post authors and their avatars in embedded comments so they work when using subfolder installation.
See [this bug](https://meta.discourse.org/t/links-without-sub-folder-path-in-embedded-comments-continue-discussion-and-user-links/37836/3).
PS: I'm using `Discourse.base_url` here, may be it worth using `Discourse.base_uri` instead?

### DIFF
--- a/app/views/embed/comments.html.erb
+++ b/app/views/embed/comments.html.erb
@@ -16,11 +16,11 @@
       <%- end %>
 
       <div class='author'>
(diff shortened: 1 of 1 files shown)


eviltrout:
Thanks!

PR #3963, opened 2016-01-17
FIX: avatar and username should be side-by-side in post-notification email

scossar:
On Outlook 2013 the username is pushed to the right edge of the post-notification email. See: https://meta.discourse.org/t/first-link-in-post-footer-very-distracting/37726/26?u=simon_cossar

The fix is to put the avatar/username row into a nested table. I've copied and pasted the html source from a user-notification email into the Litmus email builder. This is from my local computer, so the avatar images aren't being generated.

Here are before the fix and after the fix screenshots  on Outlook 2013 and Gmail webmail:

(before) screenshot 2016-01-17 12 42 17

(after)screenshot 2016-01-17 12 33 41

(before)screenshot 2016-01-17 12 42 43

(after)screenshot 2016-01-17 12 34 21

On Outlook 2013 it displays properly with the change. On Gmail webmail it displays properly in both instances.

I'll be able to test this more thoroughly later this week.

### DIFF
--- a/app/views/email/_post.html.erb
+++ b/app/views/email/_post.html.erb
@@ -1,19 +1,25 @@
 <table class='post-wrapper <%= post.whisper? ? "whisper" : "" %>'>
   <tbody>
     <tr>
-      <td class='user-avatar'>
-        <img src="<%= post.user.small_avatar_url %>" title="<%= post.user.username%>">
-      </td>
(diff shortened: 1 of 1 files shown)


coding-horror:
seems OK let's try it -- thank you!

PR #3966, opened 2016-01-18
FIX: email button padding for Outlook

scossar:
The 'Visit Topic' button for emails currently has no horizontal padding on Outlook 2013. This is because Outlook doesn't recognize horizontal padding on the `<a>` tag. 

The general recommendation for buttons in html emails is to put the link in a table `td` element. That works, but to do it for the 'Visit Topic' link would involve changing some code - the link would need to be separated from the respond instructions in `user_notifications.visit_link_to_respond`.

Instead of doing that, this PR pads out the link with a borders. The only downside that I can see is that the border is slightly narrower on Outlook than on other clients. See https://litmus.com/blog/a-guide-to-bulletproof-buttons-in-email-design ( 3) BORDER-BASED BUTTONS) for a reference.

You might want to reduce the font-size and border widths from what I have set.

It is possible to add some extra horizontal padding to Outlook 2013 by conditionally adding `&nbsp;` characters on either side of the link, but to do this would require separating the link from the reply text.

Here are some screen shots:

screenshot 2016-01-18 14 29 36

screenshot 2016-01-18 14 26 54

screenshot 2016-01-18 14 27 57

screenshot 2016-01-18 14 28 25

### DIFF
--- a/lib/email/styles.rb
+++ b/lib/email/styles.rb
@@ -214,7 +214,7 @@ module Email
         element.css('a').each do |inner|
           # we want the first footer link to be specially highlighted as IMPORTANT
           if footernum == 0 and linknum == 0
-            inner['style'] = "font-size: 16px; font-family: Helvetica, Arial, sans-serif; background-color: #006699; color:#ffffff; border-top: 12px solid #006699; border-bottom: 12px solid #006699; border-right: 18px solid #006699; border-left: 18px solid #006699; display: inline-block;"
+            inner['style'] = "font-size: 16px; background-color: #006699; color:#ffffff; border-top: 12px solid #006699; border-bottom: 12px solid #006699; border-right: 18px solid #006699; border-left: 18px solid #006699; display: inline-block;"
           else
             inner['style'] = "color:#666;"
           end
--- a/lib/email/styles.rb
+++ b/lib/email/styles.rb
@@ -214,7 +214,7 @@ module Email
         element.css('a').each do |inner|
           # we want the first footer link to be specially highlighted as IMPORTANT
           if footernum == 0 and linknum == 0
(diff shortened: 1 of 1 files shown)


gschlager:
We should set the `font-familiy` for the whole email instead of just changing the button's font. The screenshot of Outlook looks kinda strange where everything is using a serif font except the button.

See also: https://meta.discourse.org/t/ugly-emails-in-desktop-outlook/33772

coding-horror:
Hmm, no, let's just leave font family at default for everything then.

scossar:
Right now the font is being set as `font-family:'lucida grande',tahoma,verdana,arial,sans-serif;` on some elements, but not on the main body text. The default is kind of ugly on Outlook. To change it there just needs to be a rule added for `p` elements in `Styles#format_html`

coding-horror:
It's much better to let the email client pick the fonts it wants -- I suggest removing the oddball font style from that one element.

scossar:
Here is is with the default font.

screenshot 2016-01-18 16 42 31

scossar:
I'm seeing an issue with this where it doesn't look very good if the reply instructions wrap onto a second line on mobile. Reducing the size of the button makes it a little better, but I think the best solution would be to have the button on its own line.

scossar:
I've reduced the widths on the borders. The result should be similar to the way it looks now, but with padding on the button for Outlook 2013.

Here are some screenshots. These images also have a max-width set on them. That is not a part of this pr.

screenshot 2016-01-20 11 23 25

screenshot 2016-01-20 11 23 55

screenshot 2016-01-20 11 24 38

coding-horror:
looks good to me!

PR #3967, opened 2016-01-19
Avoid loading eslint twice.

tgxworld:
Lets see if this improves 
screenshot from 2016-01-19 11-14-45


tgxworld:
I saved  a second on my machine but on travis still a second

PR #3969, opened 2016-01-20
FEATURE (WIP): add max-width and center email notifications

scossar:
Work in progress:

- adds a layout that sets a max-width of 600px (and a fixed width of 600px for Outlook) see: http://webdesign.tutsplus.com/tutorials/creating-a-future-proof-responsive-email-without-media-queries--cms-23919

- sets a max-width of 600px on images (fixed width of 600px for Outlook)

The css in the layout is written inline for now. It could be added to `Email::Styles`.

This may work with a more minimal layout that doesn't add a `head` section or doctype. I have read that there are issues with the iphone not respecting padding on content when there is no `<meta name="viewport" content="width=device-width, initial-scale=1.0">` tag. I haven't tested that though.

Using `Margin` instead of `margin` is a hack to keep it from being stripped by Outlook.

Feel free to use any of this.

screenshot 2016-01-20 13 57 04

screenshot 2016-01-20 13 57 14

screenshot 2016-01-20 13 57 39

screenshot 2016-01-20 13 58 31

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -5,6 +5,7 @@ require_dependency 'age_words'
 class UserNotifications < ActionMailer::Base
   helper :application
   default charset: 'UTF-8'
+  layout 'user_notifications'
 
   include Email::BuildEmailHelper
 
@@ -290,6 +291,7 @@ class UserNotifications < ActionMailer::Base
     else
       html = UserNotificationRenderer.new(Rails.configuration.paths["app/views"]).render(
         template: 'email/notification',
+        layout: 'layouts/user_notifications',
         format: :html,
         locals: { context_posts: context_posts,
                   post: post,
--- a/app/views/layouts/user_notifications.html.erb
+++ b/app/views/layouts/user_notifications.html.erb
@@ -1,0 +1,42 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+  <!--[if !mso]><!-->
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <!--<![endif]-->
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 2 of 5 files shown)


coding-horror:
do you feel this is safe to merge?

scossar:
No. Someone needs to look at how it's adding the layout in
`user_notifications` (it's doing it in two places) and at the change in
`styles.rb` from using `@fragment = Nokogiri::HTML.fragment(@html)` to
`@doc= Nokigiri::HTML(@html)`.

coding-horror:
ok I nominate @ZogStriP as he's on mail as a primary feature for a few more weeks..

scossar:
If you just want to use the ideas in there that's fine by me. The layout seems good. It's what's recommended in the tutorial I reference in the first comment on here, and is very close to what they teach in the CodeSchool email course. Setting the max-width on the images for Outlook (and Yahoo) was a bit of trial end error. It's set in `Styles::format_basic`
```
...
      big_images.each do |img|
        add_styles(img, 'width: 100%; height: auto; max-width: 600px;') if img['style'] !~ /max-width/
        add_attributes(img, width: 600)
      end
```

The logic is that Outlook gets its width from the `width="600"` attribute, everything else gets the width from the inline styles. If `max-width` isn't set images can get huge in Yahoo.

ZogStriP:
I like it, but you know what's missing? Some tests ;) Can you add some?

scossar:
Sure, I'd like to. It might be a couple of days before I can get it done.

scossar:
@ZogStriP I've added some tests.

I reworked `styles_spec.rb` so that it works with a Nokogiri document instead of a fragment. Under `context "format notifications"` I'm testing that I haven't broken `Styles#style`.

I had a hard time finding a way to test that that correct layout is being used for email notifications. I ended up checking for the meta tag from the layout. Maybe there's a better way. 

Any suggestions for improvement are welcome.

I have access to Litmus email testing for the next couple of weeks - possibly longer if I get some work using it. I did a quick test of a post. You can see the results here: https://litmus.com/checklist/public/b5e45d3 
I'm happy to tweak the styles if there's anything you want changed - (font-size and line height could be set for desktop clients.)

If you want to add a basic layout to notifications that don't have an html_part, (so that their width is constrained) that can be done in `renderer.rb` by wrapping `PrettyText.cook(text)` in a table.

coding-horror:
if @ZogStriP  is OK with it we can merge

ZogStriP:
Thanks :+1:

PR #3972, opened 2016-01-22
Remove Arel patch that has been merged upstream.

tgxworld:
cc/ @SamSaffron


SamSaffron:
cool if its upstreamed sure ! :)

PR #3973, opened 2016-01-25
FEATURE: AR adapter to failover to a replica DB server.

tgxworld:


tgxworld:
cc/ @SamSaffron

SamSaffron:
Did you test this on multisite, to ensure it works?

SamSaffron:
looks like this does nothing if no env is set, so merging in ... lets test on meta

lib/active_record/connection_adapters/postgresql_fallback_adapter.rb
```diff
@@ -0,0 +1,102 @@
...
+      Rails.logger
+    end
+  end
+
+  def start(existing_connection, config)
+    return if @running
+    @running = true
+    Discourse.enable_readonly_mode
+
+    timer_task = Concurrent::TimerTask.new(execution_interval: interval) do |task|
```
SamSaffron: I am a bit uneasy with adding another thread here, it feels too heavy, instead hijack when you checkout a connection like the connection reaper does

tgxworld:
> Did you test this on multisite, to ensure it works?

Multisite breaks with this because of the way it deals with establishing a connection. This failover will probably need to be built into multisite

lib/active_record/connection_adapters/postgresql_fallback_adapter.rb
```diff
@@ -0,0 +1,114 @@
...
+    @running = true
+
+    Thread.new do
+      begin
+        logger.info "#{self.class}: Checking master server..."
+        connection = ActiveRecord::Base.postgresql_connection(config)
+
+        if connection.active?
+          logger.info "#{self.class}: Master server is active. Reconnecting..."
+          ActiveRecord::Base.remove_connection
```
tgxworld: This is to ensure we remove any existing connections to the replica.

tgxworld:
@SamSaffron Updated.

lib/active_record/connection_adapters/postgresql_fallback_adapter.rb
```diff
@@ -0,0 +1,113 @@
...
+  attr_reader :running
+  attr_accessor :master
+
+  def initialize
+    @master = true
+    @running = false
+  end
+
+  def verify_master
+    return if @running && recently_checked?
```
SamSaffron: slight concurrency concern here... do you need a mutex?

lib/active_record/connection_adapters/postgresql_fallback_adapter.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        if connection.active?
+          logger.info "#{self.class}: Master server is active. Reconnecting..."
+          ActiveRecord::Base.establish_connection(config)
+          Discourse.disable_readonly_mode
+          @master = true
+        end
+      rescue => e
+        if e.message.include?("could not connect to server")
+          logger.info "#{self.class}: Connection to master PostgreSQL server failed with '#{e.message}'"
```
SamSaffron: I would warn here, a warning every 5 secs when db is on fire is not the end of the world

PR #3980, opened 2016-01-27
added plugin outlet under the header content div

emanuelet:
As I need to add a navigation bar in that position I will like to add this plugin outlet


ZogStriP:
Thanks :

emanuelet:
thanks to you

PR #3981, opened 2016-01-27
Replace uncategorized_description setting with translation

gschlager:
Moves the `uncategorized_description` site setting into a locale file.
The existing setting value gets migrated into the `translation_overrides` table.

https://meta.discourse.org/t/new-localizable-resources/21452


ZogStriP:
Thanks :+1:

PR #3982, opened 2016-01-29
Move notification layout styles inline, into the layout

scossar:
This is the easiest way that I can see to get the layout styles included for the email digest. Currently, the email digest uses the user_notification layout, but the styles for the layout are not being inlined. This is because `Styles#format_notification` is not called for the email digest. The styles cannot be included in `Styles#format_html`, because, for the notifications, that method is called after the classes have been stripped from the html.

Without this change the digest is very wide on Outlook 2013.

### DIFF
--- a/app/views/layouts/user_notifications.html.erb
+++ b/app/views/layouts/user_notifications.html.erb
@@ -13,7 +13,7 @@
   </style>
   <![endif]-->
 </head>
-<body style="Margin:0;padding:0;min-width:100%;background-color:#ffffff;">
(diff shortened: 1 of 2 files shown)


ZogStriP:
I had to revert your previous PR (#3969), can you include it in this one?

coding-horror:
This has conflicts so it can't be merged. The current state of email formatting is disastrous based on previous checkins... should I just make this change by hand?

scossar:
I think just close it. I'll make a new PR if that's ok.

PR #3987, opened 2016-01-29
Small changes to the phpBB3 importer

gschlager:


script/import_scripts/phpbb3/settings.yml
```diff
@@ -17,7 +17,8 @@ import:
   # This is the path to the root directory of your current phpBB installation (or a copy of it).
   # The importer expects to find the /files and /images directories within the base directory.
   # This is only needed if you want to import avatars, attachments or custom smilies.
-  phpbb_base_dir: /var/www/phpbb
+  # Set this to /shared/import/data if you are using the Docker based importer.
```
tgxworld: Hmm I think the comment here might not be needed since we already set it as the default. Perhaps set it to `/var/www/phpbb` if you're not using the Docker based install?

gschlager:
@tgxworld I changed the comment and updated the PR.

ZogStriP:
Thanks ;)

PR #3995, opened 2016-02-05
Reducing duplication and string allocation in Jobs::ExportCsvFile

devonestes:
I reduced duplication in the #user_list_export method by extracting several methods that were common to both branches of the logic in that method. I also changed a couple string literals to symbols in places where I knew it wouldn't make a difference, since it looked like memory usage was a concern because there is a lot of manual de-allocation of variables.

I'm not sure if this would be considered too much of a 'refactoring only' PR, but I figured I'd submit it since that's the best way for me to get a gauge of what's too minor and what isn't.

### DIFF
--- a/app/jobs/regular/export_csv_file.rb
+++ b/app/jobs/regular/export_csv_file.rb
@@ -44,7 +44,7 @@ module Jobs
     end
 
     def user_archive_export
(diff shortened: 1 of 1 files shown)


ZogStriP:
This is cool :+1: 

What we don't want is _ref*cktoring_ PR that touches lots of different files only to change the name of a variable.

devonestes:
Cool! Well then I'll have some more stuff coming now that I know these kinds of things are ok.

PR #3996, opened 2016-02-05
UX: Use i18n for date picker

gschlager:
The date picker should respect the user's locale.

German locale without this PR:
image

After:
image

- I had to upgrade moment.js to version 2.11.2 for this to work.
- I couldn't find any incompatibilities, so everything should work just fine.
- I made a small change to the `JsLocaleHelper` so that we don't need to rename the locale files anymore.


ZogStriP:
:heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

PR #3998, opened 2016-02-06
Enable block quotes to be readable on the Profile About Me section

cpradio:
https://meta.discourse.org/t/blockquotes-on-profile-page/39150


ZogStriP:
Thanks :+1:

PR #3999, opened 2016-02-06
Move from [] -> Notification.none

lcpriest:
Returning [] as an 'empty collection' generally needs more code when calling it to do with the response.
e.g. in the case of 
Replacing it with `.none` is a pretty safe bet.

http://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/none

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -148,7 +148,7 @@ class Notification < ActiveRecord::Base
         end
       end.take(count)
     else
-      []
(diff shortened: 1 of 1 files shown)


lcpriest:
Just did!

SamSaffron:
not following, this method is always returning an array? is this solving a particular bug?

lcpriest:
Ah, my mistake, I missed that!

PR #4000, opened 2016-02-06
FIX: Remove invalid 'http://' website from profiles

gschlager:
https://meta.discourse.org/t/cannot-update-preferences-website-is-invalid/39162


techAPJ:
Thank You! :+1:

PR #4004, opened 2016-02-10
Feature: (WIP) Send email notifications in user's preferred language

scossar:
This is a part of the larger feature that is outlined in this topic:

https://meta.discourse.org/t/variable-default-interface-language-for-new-users/33741/8

Currently, all email notifications are sent in the site's default language. This PR adds a `locale` option to each of the notification methods (except digest) in `message_builder.rb`, so that notifications are sent in the user's preferred locale.

I have finished the rest of the larger feature  for setting the user's locale from the accept-language headers. I'll send it in a separate PR tomorrow. Any feedback on this would be greatly appreciated.

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -11,37 +11,43 @@ class UserNotifications < ActionMailer::Base
   def signup(user, opts={})
     build_email(user.email,
                 template: "user_notifications.signup",
+                locale: user_locale(user),
                 email_token: opts[:email_token])
   end
 
   def signup_after_approval(user, opts={})
     build_email(user.email,
                 template: 'user_notifications.signup_after_approval',
+                locale: user_locale(user),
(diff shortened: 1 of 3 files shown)


ZogStriP:
:+1:

coding-horror:
Ditto can you review this Aus Monday Sam? Customer interest.

scossar:
Rebased and simplified. No longer adding locale as an option for UserNotifications that are calling `#notification_email`. It is now set once in `#send_notification_email`. I have added `locale` as an option to calls to `I18n.t` in message_builder.rb for the cases where template_args are not being passed. This allows the unsubscribe link to be translated.

There may be a better way to do this. Unless you do something, the site setting to allow users to set their locale is kind of useless.

coding-horror:
Is this safe to merge @zogstrip?

ZogStriP:
:+1:

PR #4006, opened 2016-02-11
Feature: (WIP) Set locale from Accept-Language header

scossar:
For forums that allow user's to select their locale, this PR sets the locale for anonymous users from the Accept-Language header. It also automatically adds a locale property when a new user is created. This property is set from the I18n.locale. The reason for setting the locale on signup is so that new users can begin using the site in their preferred language without having to navigate the their user preferences to change it. This functionality is very similar to how google/facebook/twitter deal with locale.

It is using the http_accept_language gem to detect the user's preferred language: https://github.com/iain/http_accept_language

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,8 @@ gem 'active_model_serializers', '~> 0.8.3'
 
 gem 'onebox'
 
+gem 'http_accept_language', '~>2.0.5'
+
 gem 'ember-rails'
 gem 'ember-source', '1.12.2'
 gem 'barber'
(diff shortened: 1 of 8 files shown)


coding-horror:
Can you review this Australia Monday, Sam? We have a customer interested in this

ZogStriP:
Looks like this needs a rebase

scossar:
I can do it later today.

scossar:
It should be good now.

coding-horror:
Ok is this safe to merge @zogstrip?

ZogStriP:
:+1:

MuthukkumarNicheSoft:
I have cloned this code and put it in discourse but no luck. could some one please tell me how to fix the issue of user language selection.

PR #4007, opened 2016-02-11
FIX: Client settings were not being published.

tgxworld:
cc/ @SamSaffron

### DIFF
--- a/spec/components/site_setting_extension_spec.rb
+++ b/spec/components/site_setting_extension_spec.rb
@@ -19,7 +19,7 @@ describe SiteSettingExtension do
       end
     end
   end
-  
+
   let :provider_local do
     SiteSettings::LocalProcessProvider.new
   end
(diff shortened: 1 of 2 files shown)


spec/components/site_setting_extension_spec.rb
```diff
@@ -140,6 +140,21 @@ def new_settings(provider)
...
+        settings.client_setting("test_setting")
+        called = false
+
+        MessageBus.subscribe('/client_settings') do |msg|
+          expect(msg.data["value"]).to eq(88)
+          called = true
+        end
+
+        settings.test_setting = 88
+        sleep 0.01
```
tgxworld: @SamSaffron Do you know if there is a way to monitor MessageBus so that I can wait until it publishes?

ZogStriP: I believe you can use `MessageBus.track_publish`.

tgxworld:
Thanks! @ZogStriP  :)

PR #4009, opened 2016-02-13
Take filename to write to as optional parameter to export_category

ebernhardson:
My discourse instance will be making regular automated public backups
of specific categories. It's preferred to be able to directly control
the path and filename of the output, rather than letting discourse
choose for me. This was already mostly supported, a filename parameter
just needed to be passed through the cli app.

### DIFF
--- a/lib/import_export/import_export.rb
+++ b/lib/import_export/import_export.rb
@@ -6,8 +6,8 @@ require "json"
 
 module ImportExport
 
-  def self.export_category(category_id)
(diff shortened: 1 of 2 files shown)


ebernhardson:
I have to send the CLA through legal on my side, will sign early next week
after hearing back.

ebernhardson:
CLA has been submitted

ZogStriP:
Thanks :+1:

PR #4011, opened 2016-02-13
Fixes for I18n and translation overrides

gschlager:
This PR includes a couple of fixed for I18n and translation overrides:
- Makes it possible to overrides texts that aren't translated in Transifex by using the fallback locale.
- It wasn't possible to use overrides for more than one language.
- Translations from fallback locales were missing after `I18n.reload!`

This should fix the problems reported [here](https://meta.discourse.org/t/customization-text-reverses-after-update/39019), [here](https://meta.discourse.org/t/please-help-all-custom-translations-lost/37559), [here](https://meta.discourse.org/t/customize-all-text-in-discourse/36092/11) and [here](https://meta.discourse.org/t/customize-all-text-in-discourse/36092/28).

I removed the `MessageBus` call that is executing `I18n.reload!` since `I18n.reload!` is already called in `translation_override.rb`. I'll put it back in if somebody can name a good reason why `MessageBus` was used. :wink:

### DIFF
--- a/lib/freedom_patches/translate_accelerator.rb
+++ b/lib/freedom_patches/translate_accelerator.rb
@@ -105,8 +105,7 @@ module I18n
 
       by_site = @overrides_by_site[site]
 
-      by_locale = nil
-      unless by_site
+      unless by_site && by_site.has_key?(locale)
         by_site = @overrides_by_site[site] = {}
 
         # Load overrides
--- a/lib/freedom_patches/translate_accelerator.rb
+++ b/lib/freedom_patches/translate_accelerator.rb
@@ -28,6 +28,7 @@ module I18n
       @overrides_by_site = {}
 
       reload_no_cache!
+      ensure_all_loaded!
     end
 
     LOAD_MUTEX = Mutex.new
--- a/lib/i18n/backend/discourse_i18n.rb
+++ b/lib/i18n/backend/discourse_i18n.rb
@@ -43,7 +43,13 @@ module I18n
       end
 
       def search(locale, query)
-        find_results(/#{query}/i, {}, translations[locale])
+        results = {}
+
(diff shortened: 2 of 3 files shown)


riking:
I assume that MessageBus was used to broadcast to all other processes that they need to reload the i18n data as well.

gschlager:
Rebased in order to fix Travis build.

@riking I couldn't find any MessageBus consumer for `i18n-flush` except the one in `100-i18n.rb`. But I'm no MessageBus expert, so I could be missing something.

config/initializers/100-i18n.rb
```diff
@@ -4,7 +4,3 @@
 I18n.backend = I18n::Backend::DiscourseI18n.new
 I18n.config.missing_interpolation_argument_handler = proc { throw(:exception) }
 I18n.init_accelerator!
```
gschlager: I'm not sure if this is intentional, but `I18n.reload!` is called twice during initialization. Once by this call to `I18n.init_accelerator!` and once by [i18n_railtie.rb](https://github.com/rails/rails/blob/7d242be5a4fbfaea2d585905d76cc6bae45cbe77/activesupport/lib/active_support/i18n_railtie.rb#L63). I thought I'd mention it since I noticed it during debugging.

SamSaffron:
@eviltrout  can you review?

eviltrout:
@gschlager the `100-i18n.rb` is the subscription that causes a reload whenever it is published. It is important for multi-process deploys to be contacted when they need to refresh. It needs to be put back in.

gschlager:
@eviltrout Makes sense...
I rebased and removed the commit that changed `100-i18n.rb`. Should work fine now.

eviltrout:
Thanks, looks good now!

PR #4015, opened 2016-02-16
FIX: topic summary description text was conflicts with reply counter

fantasticfears:
https://meta.discourse.org/t/over-1000-replies/39568


ZogStriP:
Thanks :+1:

PR #4016, opened 2016-02-17
Add instructions for creating an admin account on the CLI to the Vagrant docs

alehandrof:


alehandrof:
Not sure why the checks are failing -- I only made a change to the Vagrant docs.

SamSaffron:
hmmm ... what this RAILS_DB business here? that should only be relevant in multisite

alehandrof:
Here's how this PR came about: https://meta.discourse.org/t/how-to-get-admin-access-when-using-the-vagrant-install/39736

Perhaps it will make more sense with that context. I'll fully admit I have wrapped my head around the intricacies of Discourse yet :)

alehandrof:
I understand what you were asking and you're right, of course. I just tested the following on brand new repo and the account creation works as expected:

```
vagrant up
vagrant ssh
bundle install
bundle exec rake db:migrate
rake admin:create 
```

I'm not sure why it wasn't working on my other earlier version -- but I've been doing all sorts of weird things to it.

Since it's a tiny change, shall I close down the PR and you can add it at some point to the Markdown file?

SamSaffron:
thanks, yeah we should do a bigger review of the dev doco agree it needs some tuning

PR #4017, opened 2016-02-18
FEATURE: Admin Banner when Logster logs is getting flooded.

tgxworld:
https://meta.discourse.org/t/admin-banner-notification-of-error-floods-at-logs/39379

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -977,6 +977,9 @@ en:
     max_invites_per_day: "Maximum number of invites a user can send per day."
     max_topic_invitations_per_day: "Maximum number of topic invitations a user can send per day."
 
+    alert_admins_if_errors_per_minute: "Number of errors per minute in order to trigger an admin alert. A value of 0 disables this feature. NOTE: requires restart."
+    alert_admins_if_errors_per_hour: "Number of errors per hour in order to trigger an admin alert. A value of 0 disables this feature. NOTE: requires restart."
+
     suggested_topics: "Number of suggested topics shown at the bottom of a topic."
     limit_suggested_to_category: "Only show topics from the current category in suggested topics."
 
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -761,6 +761,14 @@ rate_limits:
   tl2_additional_likes_per_day_multiplier: 1.5
   tl3_additional_likes_per_day_multiplier: 2
   tl4_additional_likes_per_day_multiplier: 3
+  alert_admins_if_errors_per_minute:
+    client: true
+    shadowed_by_global: true
+    default: 0
+  alert_admins_if_errors_per_hour:
+    client: true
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/discourse/templates/application.hbs
```diff
@@ -6,6 +6,7 @@
       {{custom-html "top"}}
     {{/if}}
     {{global-notice}}
```
SamSaffron: Can't this be simply added into global-notice, uneasy adding another component for everyone, its more work for ember to do on every page.

config/site_settings.yml
```diff
@@ -764,6 +764,14 @@ rate_limits:
   tl2_additional_likes_per_day_multiplier: 1.5
   tl3_additional_likes_per_day_multiplier: 2
   tl4_additional_likes_per_day_multiplier: 3
+  alert_admins_if_errors_per_minute:
+    client: true
+    shadowed_by_global: true
+    default: 100
```
SamSaffron: what value is used to disable altogether? that should be default.

SamSaffron: also we got to think about how we are going to get this all going in our environment, we may need a custom plugin for meta to poll rate, cause error reports are reported from multiple sources.

app/assets/javascripts/discourse/templates/application.hbs
tgxworld: `global-notice` deals with static content that is displayed site wide for all users. 
`logs-notice` would need to keep track of a few properties to determine if it should be displayed. I don't think we should mix it into `global-notice`.

tgxworld: O wait .. let me try moving the logic into a model

config/site_settings.yml
```diff
@@ -761,6 +761,14 @@ rate_limits:
   tl2_additional_likes_per_day_multiplier: 1.5
   tl3_additional_likes_per_day_multiplier: 2
   tl4_additional_likes_per_day_multiplier: 3
+  alert_admins_if_errors_per_minute:
+    client: true
```
SamSaffron: default should be false for now till we tune this

config/locales/server.en.yml
```diff
@@ -977,6 +977,9 @@ en:
     max_invites_per_day: "Maximum number of invites a user can send per day."
     max_topic_invitations_per_day: "Maximum number of topic invitations a user can send per day."
 
+    alert_admins_if_errors_per_minute: "Number of errors per minute in order to trigger an admin alert. NOTE: requires restart."
+    alert_admins_if_errors_per_hour: "Number of errors per hour in order to trigger an admin alert. NOTE: requires restart."
```
SamSaffron: description should be fixed, to note that 0 means it is disabled

SamSaffron:
Looking good, go ahead and merge in first thing next week

PR #4018, opened 2016-02-18
Create client.vi_VN.yml

httvncoder:
Vietnamese file : translate for use

### DIFF
--- a/config/locales/client.vi_VN.yml
+++ b/config/locales/client.vi_VN.yml
@@ -1,0 +1,2291 @@
+vi_VN:
+  js:
+    number:
+      format:
+        separator: "."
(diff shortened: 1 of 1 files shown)


SamSaffron:
@techAPJ what do we do about this ?

SamSaffron:
I will just merge in for now :) thanks

techAPJ:
@SamSaffron this PR needs to be reverted. We already have Vietnamese translation file: https://github.com/discourse/discourse/blob/master/config/locales/client.vi.yml

@httvncoder Vietnamese translations needs to be updated via Transifex: transifex.com/projects/p/discourse-org/

PR #4022, opened 2016-02-21
Don't try to import invalid websites

gschlager:
Importers shouldn't try to import invalid websites and the bio shouldn't have more than 3000 chars.


ZogStriP:
Thanks :+1:

PR #4024, opened 2016-02-21
FIX: Don't suggest invalid username

gschlager:
Invalid usernames were suggested when
* the name was longer than the max length and
* the last character after truncation was an underscore.


ZogStriP:
Thanks :+1:

PR #4025, opened 2016-02-21
Allow importers to set empty names

gschlager:
Importers should be able to set the user's name to an empty string.


ZogStriP:
Thanks :+1:

PR #4026, opened 2016-02-21
Add reminder to run ./launcher restart after the restore completes

dandv:
See https://meta.discourse.org/t/restore-from-backup-does-not-restore-the-whole-state-of-my-forum/24903/14?u=dandv


SamSaffron:
This is going to be a problem with our hosted customers that have no ability to do this... @ZogStriP  not sure that we should do here.

ZogStriP:
I agree, we need to figure out what's wrong with site settings in production.

PR #4027, opened 2016-02-22
FIX: Use `UserStat#posts_read_count` for :all period on Users page.

tgxworld:
For imported users, there are no user visits which means that the Users page read will return 0 which does not match the posts read count in the user's summary page.


tgxworld:
Closing for now first while I investigate deeper

PR #4028, opened 2016-02-23
FIX: Remove existing characters when a new emoji is selected.

dtchau:
When the user choose an emoji from the "More..." option, any previously typed character should be discarded.

This is a fix for https://meta.discourse.org/t/when-inputting-emoji-by-typing-and-choosing-more-the-result-includes-the-parenthesis/39785

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -301,7 +301,16 @@ export default Ember.Component.extend({
           showSelector({
             appendTo: self.$(),
             container,
-            onSelect: title => self._addText(self._getSelected(), `${title}:`)
+            onSelect: title => {
+              // Remove the previously type characters when a new emoji is selected from the selector.
+              let selected = self._getSelected();
+              let newPre = selected.pre.replace(/:[^:]+$/, ":");
+              let numOfRemovedChars = selected.pre.length - newPre.length;
+              selected.pre = newPre;
+              selected.start -= numOfRemovedChars;
+              selected.end -= numOfRemovedChars;
+              self._addText(selected, `${title}:`);
+            }
           });
           return "";
         }


app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -301,7 +301,16 @@ export default Ember.Component.extend({
...
-            onSelect: title => self._addText(self._getSelected(), `${title}:`)
+            onSelect: title => {
+              // Remove the previously type characters when a new emoji is selected from the selector.
+              let selected = self._getSelected();
+              let newPre = selected.pre.replace(/:[^:]+$/, ":");
+              let numOfRemovedChars = selected.pre.length - newPre.length;
+              selected.pre = newPre;
+              selected.start -= numOfRemovedChars;
+              selected.end -= numOfRemovedChars;
+              self._addText(selected, `${title}:`);
```
tgxworld: Thanks for the PR @dtchau :) I don't think this is where the code broke though. Can you try a `git bisect` to and find out where it broke? Probably due to changes in autocomplete.js

dtchau: Thank you for the input. I will dig in tomorrow.

dtchau: Hi @tgxworld, it took me awhile because of my unfamiliarity with the project and some quirks serving the server (namely the migrating issue because of the rails_multisite gem).

Anyhow, the commit that introduced this problem is - 8b59f76f5999124e13ecf40c46d8721a0e605e0b and was part of PR https://github.com/discourse/discourse/pull/3831

Now that we've found the changes that caused this problem in the first place, I want to make a case for this PR. 

1. Before this bug, as soon as the "more..." option was clicked, all previously typed characters will be removed. As a result, even if the users didn't end up choosing anything from the popup, their intention/"hard work" was wiped out.  I don't think this was a right thing to do.
1. The handling of the "more..." option is only specific to itself. So I think it makes more sense to do so only on the option's `onSelect` event and not tie it directly to the entire autocomplete feature.

What do you think?

SamSaffron:
Let's try this, current behavior is really odd, cause you select an emoji and it simply breaks.

PR #4029, opened 2016-02-23
FIX[WIP]: Email category badge

scossar:
Adds module `EmailCategoryBadge` to improve the display of category badges on various email clients - especially Outlook 2013 and Gmail. Improves the alignment of the bullet with the category name for recent versions iOS.

Here's how it looks:
Bullet:
https://litmus.com/pub/aa1934f
Box:
https://litmus.com/pub/aa65977
Bar:
https://litmus.com/pub/3b0663c

For comparison here's the bullet and the box styles from the master branch:
https://litmus.com/pub/8579472
https://litmus.com/pub/76cb63e

I'll make a separate PR to deal with other alignment issues.

### DIFF
--- a/app/helpers/user_notifications_helper.rb
+++ b/app/helpers/user_notifications_helper.rb
@@ -1,3 +1,5 @@
+require_dependency 'email_category_badge'
+
 module UserNotificationsHelper
 
   def indent(text, by=2)
@@ -67,6 +69,10 @@ module UserNotificationsHelper
 
   def format_for_email(html)
     PrettyText.format_for_email(html).html_safe
+  end
+
+  def email_category_badge(category, opts=nil)
+    EmailCategoryBadge.html_for(category, opts).html_safe
   end
 
 end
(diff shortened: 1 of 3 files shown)


coding-horror:
Awesome thanks for this. The category badge has been subtly messed up for a long time so this is great!

coding-horror:
Is this safe to merge in your opinion?

scossar:
I think so. My concern was that I might not have organized the files
correctly. It leaves some unused code in category_badge.rb. It looks like
`opts[:inline_style]` (in category_badge.rb) is only there to deal with
email, but with this change it's not being used anymore.

I'm pretty sure it won't break anything.

coding-horror:
Sorry we didn't merge this yet, anything else to do here?

scossar:
Yes, look at the alignment that you are now getting for the bullet style.

scossar:
Well, I got one this morning that looked bad on gmail, but I just tested
the latest and the bullet style looks ok, except that it doesn't work on
Outlook. The problem there is that you are assigning multiple css rules to
the same element. Outlook uses the first rule and ignores the rest so you
are left with white text on a white background.
The box style still has a lot or problems - it doesn't work on Outlook 2013
or gmail. Maybe you should just not use it in emails? If you do want to
keep using it, you need to change it so that it is created by putting a
background color on a link. Look at the way the reply link-button is
created in lib/email/style.rb for reference.

PR #4030, opened 2016-02-23
Topic status event

joebuhlig:
Created an event that triggers when the status of a topic is updated.


joebuhlig:
I need to squash these commits yet but other than that it should be good to go.

joebuhlig:
Had issues consolidating these commits and rebasing. Don't ask. I created a new PR for this: https://github.com/discourse/discourse/pull/4037

PR #4031, opened 2016-02-23
fix: support for hyphens in group name

rriemann:
The group mention @ORG-team triggers notifications for the group @ORG. This fix changes the RegExp, so that the group name is correctly extracted.

see: https://meta.discourse.org/t/group-mentions-that-begin-with-the-same-characters-may-be-incorrect/39892/12?u=rriemann

### DIFF
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -73,7 +73,7 @@ module PrettyText
   @ctx_init = Mutex.new
 
   def self.mention_matcher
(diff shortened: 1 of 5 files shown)


SamSaffron:
cool I will merge this in, but want to do a little re-work don't think we need the regex at all

SamSaffron:
K amended here:

https://github.com/discourse/discourse/commit/833af461ab5d9e170306172b68541bd62b09201a

PR #4032, opened 2016-02-24
FIX: Autocomplete not triggering consistently on mobile.

tgxworld:
cc @SamSaffron


SamSaffron:
This feels a bit odd to me ... I really want to make sure that the same search is never repeated ... what is actual repro of the issue? 

also shfit looks like a typo :)

PR #4034, opened 2016-02-24
show monthly top topics on 404 page

techAPJ:
cc @SamSaffron @coding-horror

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1232,6 +1232,7 @@ describe UsersController do
           guardian = Guardian.new(user)
           guardian.stubs(:ensure_can_edit!).with(user).raises(Discourse::InvalidAccess.new)
           Guardian.stubs(new: guardian).with(user)
+          Guardian.stubs(new: guardian).with(nil)
 
           put :update, username: user.username, name: 'Jim Tom'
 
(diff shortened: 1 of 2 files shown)


coding-horror:
basically a one-liner so I think it's OK unless you have concerns @SamSaffron

spec/controllers/users_controller_spec.rb
```diff
@@ -1232,6 +1232,7 @@ def post_user
           guardian = Guardian.new(user)
           guardian.stubs(:ensure_can_edit!).with(user).raises(Discourse::InvalidAccess.new)
           Guardian.stubs(new: guardian).with(user)
+          Guardian.stubs(new: guardian).with(nil)
```
ZogStriP: Are you stubbing it with `user` or `nil`?

techAPJ: `nil`, so on 404 page monthly top topics will be loaded assuming that the user is not logged in.

ZogStriP: Then the previous line isn't necessary, isn't it?

techAPJ: Okay, I updated the spec via https://github.com/discourse/discourse/commit/6df5b38b54d265fef0e6a9749ff89839db898374

PR #4035, opened 2016-02-24
Return 404 instead 500 when plugin is disabled.

tgxworld:
https://meta.discourse.org/logs/show/1fa6f998b88d9a15e6bc2bedb8ece9f2

@SamSaffron Please review :) When a plugin is disabled, we don't want users to be seeing internal server errors.


SamSaffron:
excellent!

PR #4036, opened 2016-02-24
FIX: `Array#first` with block will return the first element if nothin…

tgxworld:
…g matches.

https://meta.discourse.org/logs/show/b396a5de4f376643e5ea7dea94a449d9
https://meta.discourse.org/logs/show/758eb7a0a1cb5601f870ba32830baac6

cc/ @ZogStriP


ZogStriP:
:+1:

PR #4038, opened 2016-02-25
FEATURE: Instagram OmniAuth login methods

markbiegel:
All methods added to include Instagram as an OmniAuth provider.


SamSaffron:
The PR looks good, but ... would prefer if gem was `require: false` can that be safely changed? 

Also, @coding-horror my call is to add instagram in, it has bigger and more diverse userbase than even github (21 million) vs instagram (400 million)

coding-horror:
OK fine by me

SamSaffron:
OK, let try this out, I will see if I can clean up the requiring

SamSaffron:
thanks!

PR #4040, opened 2016-02-25
FIX: Enable post username to be specified in embedded comments

sammyd:
Although this appeared to be supported, it seems that it only worked via RSS polling.

The JavaScript allows you to specify a `discourseUserName` parameter, which it translates into the `discourse_username` URL parameter. This updates the EmbedController to pass that parameter from the URL into the TopicRetriever, ensuring that the new topic is created with the requested username.

https://meta.discourse.org/t/discourse-username-in-embedded-comments/40103

### DIFF
--- a/app/controllers/embed_controller.rb
+++ b/app/controllers/embed_controller.rb
@@ -8,6 +8,7 @@ class EmbedController < ApplicationController
 
   def comments
     embed_url = params[:embed_url]
+    embed_username = params[:discourse_username]
 
     topic_id = nil
     if embed_url.present?
@@ -35,7 +36,7 @@ class EmbedController < ApplicationController
       end
 
     elsif embed_url.present?
-      Jobs.enqueue(:retrieve_topic, user_id: current_user.try(:id), embed_url: embed_url)
+      Jobs.enqueue(:retrieve_topic, user_id: current_user.try(:id), embed_url: embed_url, author_username: embed_username)
       render 'loading'
     end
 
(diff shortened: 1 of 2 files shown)


sammyd:
If I understand this correctly, this failed because of some time out issue on Travis?

Is there anything I need to do, or can it be triggered again?

eviltrout:
The travis failure is only 1/4 rubies so I think it's fine. Thanks for the commit :)

ellismarkf:
@sammyd, @eviltrout: I see that this feature was merged, but when I use the `discourseUserName` query parameter, it doesn't seem to override the required username for topic creation specified in the Embed settings section of the admin panel.  Is the expected behavior to override that required setting?  Thanks!

yyhmsg:
Can I bump this for @ellismarkf 

I have the exact same question, I'm trying the following (details removed) - it works fine without the discourseUserName line and doesn't seem to do anything with that line in there...

  		DiscourseEmbed = { discourseUrl: 'https://forum.mine.com/',
		discourseUserName = 'username',
        discourseEmbedUrl: 'https://mine.com" . "$page->url' };

Any clarification whether or not this should work with embedding would be much appreciated...

yyhmsg:
Can I bump this for @ellismarkf - should this be working?  If so, could the page at https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963 be updated to show this as a possible option to include in the call....

PR #4042, opened 2016-02-25
Generate proper URIs for iframes in emails

fefrei:
When iframes are turned into links for emails, so far; query strings and fragments were stripped and the `://` to separate scheme and domain was missing. This fixes both problems.

Disclaimer: I could not test this in a proper dev environment, just by playing around with a small test ruby script.

### DIFF
--- a/lib/email/styles.rb
+++ b/lib/email/styles.rb
@@ -127,8 +127,8 @@ module Email
           src_uri = URI(i['src'])
 
           # If an iframe is protocol relative, use SSL when displaying it
-          display_src = "#{src_uri.scheme || 'https://'}#{src_uri.host}#{src_uri.path}"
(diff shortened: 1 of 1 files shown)


fefrei:
After finally setting up a development environment, I can confirm you were right @SamSaffron, and have updated my commit accordingly.

fefrei:
Oops, I amended the wrong commit. Will fix shortly...

fefrei:
I have no idea why the build failed, I don't think it is related to my change.

fefrei:
Is there anything I can do about the failed builds? I think this is a case of [cry wolf syndrome](https://meta.discourse.org/t/improving-blocked-user-state/37306/19?u=fefrei).

ZogStriP:
Nope, the build's fine ;) Don't worry.

PR #4045, opened 2016-02-27
fix a few typos in the German translation

RalfJung:
This fixes two typos in the German translation.

I have to say that I find `stage_explanation` pretty incomprehensible. But I find it equally incomprehensible in German and English, so the translation is not at fault ;-) . In particular, it is unclear to me whether `can only post via email in specific topics` means that in the other topics, the user can *not post at all*, or whether in the other topics, she can post normally through the web-interface, but not via e-mail. Also, which are these `specific` topics?

---

I signed your CLA. Furthermore, I hereby declare that *anybody* has the right to use this contribution under *any* license whatsoever. This is to fight the asymmetry in the CLA, which (in a sneaky way, IMHO), turns this GPL-ed project into "GPL for everybody, BSD-like for The Company", violating the usual principle that everybody but the author has the same rights on a piece of open-source software. See <https://mjg59.dreamwidth.org/25376.html> for further explanations of this issue.

### DIFF
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -2254,7 +2254,7 @@ de:
         moderator: "Moderator?"
         admin: "Administrator?"
         blocked: "Geblockt?"
-        staged: "Insziniert?"
+        staged: "Inszeniert?"
         show_admin_profile: "Administration"
         edit_title: "Titel bearbeiten"
         save_title: "Titel speichern"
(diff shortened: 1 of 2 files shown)


RalfJung:
I found another typo; should I add a new commit to this PR or make a new PR?

gschlager:
Thanks for your help. Unfortunately translations aren't accepted as pull request.
Please join the German translation team on Transifex and help us improve the translation: https://www.transifex.com/discourse/discourse-org/

If you want to discuss the translations visit us at https://meta.discourse.org/c/dev/translations 
There's also a topic for the [German translation](https://meta.discourse.org/t/german-translation/18151). Looking forward to seeing you there. :smiley:

RalfJung:
> Thanks for your help. Unfortunately translations aren't accepted as pull request.

Ouch. I kind of expected something like that, and looked for a README in `config/locales/` that would say that these files are auto-generated or something - but there is none.

gschlager:
It's at the [beginning of every file](https://github.com/discourse/discourse/blob/master/config/locales/client.de.yml#L3) you edited. :wink: 

`# Never edit this file. It will be overwritten when translations are pulled from Transifex.`

RalfJung:
D'oh... sorry^^

PR #3822, opened 2015-09-28
FIX: Add user id in the admin user routes

fantasticfears:
https://meta.discourse.org/t/usernames-in-admin-urls-cause-routing-problems/31825

PR #4047, opened 2016-02-28
FIX: Bind action admin/users#show to GET /admin/users/list

dtchau:
I bisected the code and found commit 35142847ba3968d2a8e70357b20d1a2b116bccd1 caused this problem as part of PR https://github.com/discourse/discourse/pull/3822

Simply reverting the option `except: [:show]` won't work because that will make `admin/users/_id_.json` and `admin/users/_id_/_username_` invalid. As a result I had to be specific.

Please note that `/admin/users/list/active` has always been working.

This is a fix for https://meta.discourse.org/t/middle-click-on-users-admin-panel-fails/40289

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,7 +75,7 @@ Discourse::Application.routes.draw do
 
     resources :users, id: USERNAME_ROUTE_FORMAT, except: [:show] do
       collection do
-        get 'list' => 'users#show'
+        get 'list' => redirect('admin/users/list/active')
         get "list/:query" => "users#index"
         get "ip-info" => "users#ip_info"
         delete "delete-others-with-same-ip" => "users#delete_other_accounts_with_same_ip"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,6 +75,7 @@ Discourse::Application.routes.draw do
 
     resources :users, id: USERNAME_ROUTE_FORMAT, except: [:show] do
       collection do
+        get 'list' => 'users#show'
         get "list/:query" => "users#index"
         get "ip-info" => "users#ip_info"
         delete "delete-others-with-same-ip" => "users#delete_other_accounts_with_same_ip"


dtchau:
**Wanted** - Help with failed test.

The project passed with Ruby 2.0, 2.1 and 2.2, but failed with 2.3.
Here is the failure.
```
1) Topic change_category with a previous category increases the topic_count
     Failure/Error: @category = Fabricate(:category, user: @topic.user)
     PrettyText::JavaScriptError:
       Script Timed Out
     # ./lib/pretty_text.rb:218:in `block in markdown'
     # ./lib/pretty_text.rb:411:in `block in protect'
     # ./lib/pretty_text.rb:409:in `synchronize'
     # ./lib/pretty_text.rb:409:in `protect'
     # ./lib/pretty_text.rb:188:in `markdown'
     # ./lib/pretty_text.rb:263:in `cook'
     # ./app/models/post_analyzer.rb:12:in `cook'
     # ./app/models/post.rb:191:in `cook'
     # ./lib/post_creator.rb:181:in `before_create_tasks'
     # ./app/models/post.rb:442:in `block in <class:Post>'
     # ./app/models/category.rb:196:in `create_category_definition'
     # ./spec/models/topic_spec.rb:843:in `block (3 levels) in <top (required)>'
```

What went wrong and what do I need to do to fix it? Or is this one of those tests that would just randomly fail from time to time?

Thank you in advance for any tips/suggestions.

SamSaffron:
@eviltrout your call, URL does seem a bit weird, calling `show` `list` is really not sitting right for me... 

Should probably be called `show-user` or something like that if we must

eviltrout:
It looks like that path doesn't exist. It simply exists to redirect to `active`.

I think a better solution would be to mimic what the front end does and redirect a request to `users/list/` to `/users/list/active`

PR #4049, opened 2016-02-29
FEATURE: Support multisite in PostgreSQL fallback adapter.

tgxworld:


lib/discourse.rb
```diff
@@ -113,16 +113,19 @@ def self.activate_plugins!
   end
 
   def self.recently_readonly?
-    return false unless @last_read_only
-    @last_read_only > 15.seconds.ago
+    recently_readonly = $redis.get(recently_readonly_mode_key)
```
tgxworld: This was sneaky.. I kept getting randomly failing tests until I realized this variable is shared between multisite.

tgxworld:
@SamSaffron Please review :)

lib/discourse.rb
SamSaffron: not following ... our redis proxy is meant to namespace stuff no?

tgxworld: @SamSaffron Sorry wrong line. I meant `@last_read_only`

SamSaffron: This implementation though now means that if redis gets a readonly error it will not unconditionally do a `get` every time. so every write becomes, write + read 

I think probably better to use a hash here and key on multisite db name.

tgxworld:
@SamSaffron Updated per your comments. I'll merge once I get your :+1:

SamSaffron:
lets go for it!

PR #4050, opened 2016-02-29
FEATURE: Add importer for SourceForge discussions

gschlager:


nlalonde:
Another importer! Yay! Thanks @gschlager

erlend-sh:
Learned this too late, but for future reference if anyone in the future ever ends up dabbling with this importer:

It's possible to import users as well. Every username has a corresponding Sourceforge email account, so to populate a user's email table, just append '@users.sourceforge.net' to the user name.

PR #4052, opened 2016-03-01
UX: remove category background image after the image is deleted

techAPJ:
This PR adds a check if the category background image is present before adding the category class.

This is required because (as of now) once the category background image is deleted, the [relevant CSS](https://github.com/discourse/discourse/blob/master/lib/sass/discourse_sass_importer.rb#L57) is not removed until the hard refresh is done.

Issue reported here: https://meta.discourse.org/t/category-background-images-retain-after-being-deleted/40377?u=techapj

### DIFF
--- a/app/assets/javascripts/discourse/mixins/add-category-class.js.es6
+++ b/app/assets/javascripts/discourse/mixins/add-category-class.js.es6
@@ -11,7 +11,7 @@ export default {
 
     this._removeClass();
 
-    if (categoryFullSlug) {
+    if (categoryFullSlug && this.get("categoryBackgroundUrl")) {
       $("body").addClass("category-" + categoryFullSlug);
     }
   })),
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/mixins/add-category-class.js.es6
```diff
@@ -11,7 +11,7 @@ export default {
 
     this._removeClass();
 
-    if (categoryFullSlug) {
+    if (categoryFullSlug && this.get("categoryBackgroundUrl")) {
```
ZogStriP: If we do this, people who don't use a background image for their category won't be able to make some customisation based on the category class.

techAPJ: Good point. I agree.

PR #4053, opened 2016-03-02
FIX: Image Lightbox on Subfolder Install

xfalcox:
Since LocalStore.is_relative wasn't subfolder aware, images on subfolder install weren't getting lightboxed.

Gotta add some tests to LocalStore and Post Cook.

### DIFF
--- a/lib/file_store/local_store.rb
+++ b/lib/file_store/local_store.rb
@@ -35,7 +35,7 @@ module FileStore
     end
 
     def relative_base_url
-      "/uploads/#{RailsMultisite::ConnectionManagement.current_db}"
+      "#{Discourse.base_uri}/uploads/#{RailsMultisite::ConnectionManagement.current_db}"
     end
 
     def external?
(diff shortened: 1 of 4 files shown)


SamSaffron:
This seems like a hugely high risk area, can you add a spec here.

xfalcox:
Yeah, I'm on it.

xfalcox:
@SamSaffron What do you think? :smile:

SamSaffron:
that is a lot of stubbing :) ends up being mega fragile (if implementation changes even a touch we need touch up this spec) , can these test be performed without all that stubbing?

coding-horror:
Is this stub and mock free to your satisfaction @SamSaffron

SamSaffron:
I think this fix is good, test is very complex though... @ZogStriP  can you review, merge in and perhaps simplify the test (maybe only test store here)

nlalonde:
To be fair, this test is the same complexity as what we have for the "with large images" context. So I'll merge this, and I'll add a spec in local_store_spec.

PR #4054, opened 2016-03-02
phpbb-import-script: move bbcode_to_md to before other text processing

ryantm:
This seems to fix the issue I reported in
https://meta.discourse.org/t/import-script-phpbb/40424


SamSaffron:
@ZogStriP  your call.

ZogStriP:
I think the better fix here is to submit a PR to `bbcode_to_md` instead so that it doesn't change `https` to `http`.

ryantm:
@ZogStriP I did fix that. This is a separate problem, I think.

ZogStriP:
Great, let's merge it then ;)

PR #4056, opened 2016-03-04
FEATURE: Master-Slave Redis configuration with fallback and switch over.

tgxworld:


tgxworld:
@SamSaffron  Please review. Will need to publish a new MessageBus version :)

Gemfile
```diff
@@ -38,7 +38,8 @@ gem 'mail'
 gem 'mime-types', require: 'mime/types/columnar'
 
 gem 'hiredis'
-gem 'redis', require:  ["redis", "redis/connection/hiredis"]
+# https://github.com/redis/redis-rb/pull/591
+gem 'redis', require:  ["redis", "redis/connection/hiredis"], github: 'tgxworld/redis-rb'
```
SamSaffron: hmmm very very uneasy with this, is the a PR for it? also don't source from github publish a gem instead

SamSaffron: let's freedom patch this the test will cover that it is still working on redis upgrades. we just need to remember to remove the freedom patch so add a note there.

spec/components/discourse_redis_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  describe DiscourseRedis::FallbackHandler do
+    it 'should switch back to the master server' do
+      begin
+        fallback_handler.master = false
+        $redis.without_namespace.expects(:get).raises(Redis::CommandError.new("READONLY"))
+        Redis::Client.any_instance.expects(:call).with([:info]).returns(DiscourseRedis::FallbackHandler::MASTER_LINK_STATUS)
+        Redis::Client.any_instance.expects(:call).with([:client, [:kill, 'type', 'normal']])
+
+        current_threads = Thread.list
+        expect { $redis.get('test') }.to change { Thread.list.count }.by(1)
```
SamSaffron: this is not really a good test ... its testing for a side effect of a very particular implementation. this really should not matter.

spec/components/discourse_redis_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        (Thread.list - current_threads).each(&:join)
```
SamSaffron: this is very weird...

spec/components/discourse_redis_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        expect(fallback_handler.master).to eq(true)
```
SamSaffron: this is fine

spec/components/discourse_redis_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        expect(Discourse.recently_readonly?).to eq(false)
```
SamSaffron: this is fine

spec/components/discourse_redis_spec.rb
tgxworld: hmm why so? We need to wait for the thread to complete otherwise we'll get randomly failing tests.

tgxworld: @SamSaffron I updated the PR per your comments. Just not sure about this one.

tgxworld:
@SamSaffron I'm merging this. Let me know if the specs is ok :)

PR #4037, opened 2016-02-24
Add topic status_update change event

joebuhlig:
Had issues rebasing. Closing the first attempt.

PR #4057, opened 2016-03-04
Add spec for topic_status_update event

joebuhlig:
I'm hoping someone can give me a hand with this one. I'm attempting to add a spec for this event: https://github.com/discourse/discourse/pull/4037. I know the event works as I'm using it in a plugin, but have been unable to get a successful run on the tests.

### DIFF
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -1584,4 +1584,20 @@ describe Topic do
 
     expect(topic.message_archived?(user)).to eq(true)
   end
+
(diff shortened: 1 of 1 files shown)


SamSaffron:
What I would do is ... subscribe to event in test, do the action that makes it happen, then check I got the event. no need for mocking or stubbing here.

joebuhlig:
Thanks for the direction, Sam. This should be good to go.

SamSaffron:
perfect!

PR #4058, opened 2016-03-04
FIX: validate user locale for email notifications

scossar:
Sets the locale to `nil`for invalid locales and empty strings.
It seems unlikely that an invalid locale can exist in the database - it's not possible to log in with one, but it is possible to have "" as a locale. Setting an empty string as the locale value for `I18n.t` produces an I18n::InvalidLocale error.

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -188,7 +188,7 @@ class UserNotifications < ActionMailer::Base
   protected
 
   def user_locale(user)
-    user.respond_to?(:locale) ? user.locale : nil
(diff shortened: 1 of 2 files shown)


SamSaffron:
let's try this, only concern... is it any way possible for user to be nil?

scossar:
>
> is it any way possible for user to be nil?


user or user.locale?

The user.locale has not been being set on signup - so it's been nil until
the user set's it in their preferences. It's now  being set as a user_param
from the I19n.locale.

I don't think it's possible that the user could be nil.

scossar:
I'll double check how this is working with SSO tomorrow.

PR #4059, opened 2016-03-04
FEATURE: Add site setting for disabling mailing list mode site wide

gdpelican:
Playing around with the email settings some in anticipation of this work:
https://meta.discourse.org/t/more-granular-mailing-list-mode/38107

This change is discussed here:
https://meta.discourse.org/t/completely-disable-mailing-list-mode/37334/8

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1248,6 +1248,7 @@ en:
     default_email_private_messages: "Send an email when someone messages the user by default."
     default_email_direct: "Send an email when someone quotes/replies to/mentions or invites the user by default."
     default_email_mailing_list_mode: "Send an email for every new post by default."
+    disable_mailing_list_mode: "Disable sending an email for every new post for all users"
     default_email_always: "Send an email notification even when the user is active by default."
     default_email_previous_replies: "Include previous replies in emails by default."
 
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1248,7 +1248,7 @@ en:
     default_email_private_messages: "Send an email when someone messages the user by default."
     default_email_direct: "Send an email when someone quotes/replies to/mentions or invites the user by default."
     default_email_mailing_list_mode: "Send an email for every new post by default."
-    disable_mailing_list_mode: "Disable sending an email for every new post for all users"
+    disable_mailing_list_mode: "Disallow users from enabling mailing list mode."
     default_email_always: "Send an email notification even when the user is active by default."
     default_email_previous_replies: "Include previous replies in emails by default."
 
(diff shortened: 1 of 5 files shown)


spec/models/user_option_spec.rb
```diff
@@ -18,6 +18,28 @@
 
   end
 
+  describe ".mailing_list_mode" do
```
tgxworld: `.` is used for class methods. Since `mailing_list_mode` is an instance method, `#` should be used instead.

config/locales/server.en.yml
```diff
@@ -1248,6 +1248,7 @@ en:
     default_email_private_messages: "Send an email when someone messages the user by default."
     default_email_direct: "Send an email when someone quotes/replies to/mentions or invites the user by default."
     default_email_mailing_list_mode: "Send an email for every new post by default."
+    disable_mailing_list_mode: "Disable sending an email for every new post for all users"
```
tgxworld: I think this should be something along the lines of `Disallow users from enabling mailing list mode`

spec/models/user_option_spec.rb
```diff
@@ -18,6 +18,28 @@
...
+    let!(:forum_user) { Fabricate(:user) }
+    let!(:mailing_list_user) { Fabricate(:user) }
+
+    before do
+      forum_user.user_option.update(mailing_list_mode: false)
+      mailing_list_user.user_option.update(mailing_list_mode: true)
+    end
+
+    it "should return false when `SiteSetting.disable_mailing_list_mode` is enabled" do
+      SiteSetting.expects(:disable_mailing_list_mode).twice.returns(true)
```
tgxworld: I would try to avoid mocking the site settings here since we can easily set it and ensure we revert to the default in the test.

Any thoughts on this? @ZogStriP :)

gdpelican: Ah, I was just aping the rest of the tests in the file with the stubbing; seems like instance method test in the file (`.redirected_to_top`) is marked with a '.' as well?

ZogStriP: Yeah, we usually try to avoid mocking when it's not needed ;) 

Just use `SiteSetting.disable_mailing_list_mode = true` is enough.

gdpelican: Done!

config/locales/server.en.yml
```diff
@@ -1248,6 +1248,7 @@ en:
     default_email_private_messages: "Send an email when someone messages the user by default."
     default_email_direct: "Send an email when someone quotes/replies to/mentions or invites the user by default."
     default_email_mailing_list_mode: "Send an email for every new post by default."
+    disable_mailing_list_mode: "Disallow users from enabling mailing list mode"
```
tgxworld: Small nitpick :P Missing a `.`

gdpelican: Fixed! :zap:

tgxworld:
Looks good. Thank you @gdpelican.

PR #4060, opened 2016-03-06
Force users to select an option in required custom user dropdown fields

JSFernandes:
**What does this PR do?**

This PR forces new users to select an option for all required custom user dropdown fields. Before this PR, the first option would come selected by default and the user could ignore the dropdown. See https://meta.discourse.org/t/custom-user-fields-dropdown-have-no-please-choose-option/40554

**Notes**

This is my first PR to Discourse and I could not find any frontend testing to which I could add a test for this feature/bugfix, as I could only find Rspec tests. If you think I should add tests to this, I ask that you point me in the direction of existing tests I can use an example.

**Screenshots**

Before
before

After
after


coding-horror:
Great! Maybe @eviltrout can review?

eviltrout:
Seems to work well, thanks!

PR #4062, opened 2016-03-07
FIX: Properly downcase unicode chars

xfalcox:
Fix to https://meta.discourse.org/t/title-prettify-does-not-correctly-lowercase-non-english-characters-when-removing-all-caps/16645

This adds a dependency on Active Support Multibyte to downcase on other languages.

### DIFF
--- a/lib/text_cleaner.rb
+++ b/lib/text_cleaner.rb
@@ -34,7 +34,7 @@ class TextCleaner
     text = text.mb_chars.downcase.to_s if opts[:replace_all_upper_case] && (text =~ /[A-Z]+/) && (text == text.upcase)
     # Capitalize first letter, but only when entire first word is lowercase
     first, rest = text.split(' ', 2)
-    if opts[:capitalize_first_letter] && first == first.mb_chars.downcase
+    if first && opts[:capitalize_first_letter] && first == first.mb_chars.downcase
       text = "#{first.mb_chars.capitalize}#{rest ? ' ' + rest : ''}"
     end
     # Remove unnecessary periods at the end
--- a/lib/text_cleaner.rb
+++ b/lib/text_cleaner.rb
@@ -31,9 +31,9 @@ class TextCleaner
     # Replace ????? with a single ?
     text.gsub!(/\?+/, '?') if opts[:deduplicate_question_marks]
     # Replace all-caps text with regular case letters
-    text = text.mb_chars.downcase if opts[:replace_all_upper_case] && (text =~ /[A-Z]+/) && (text == text.upcase)
+    text = text.mb_chars.downcase.to_s if opts[:replace_all_upper_case] && (text =~ /[A-Z]+/) && (text == text.upcase)
     # Capitalize first letter, but only when entire first word is lowercase
-    text.sub!(/\A([a-z]*)\b/) { |first| first.capitalize } if opts[:capitalize_first_letter]
+    text = text.mb_chars.capitalize.to_s if opts[:capitalize_first_letter] && text.split(' ').first == text.split(' ').first.mb_chars.downcase
     # Remove unnecessary periods at the end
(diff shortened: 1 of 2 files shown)


SamSaffron:
hmmm ... this would "infect" the rest of the app with this downcase implementation in a non consistent way ... if we really need it, it should simply be included via an initializer... 

also we need a test for this :)

coding-horror:
Yes this seems hugely dangerous, also "correct" Unicode upcase/downcase is CRAZY slow and expensive.

xfalcox:
I added two specs, covering the downcase and capitalize.

> hmmm ... this would "infect" the rest of the app with this downcase implementation in a non consistent way ... if we really need it, it should simply be included via an initializer...

Since this happens inside the `mb_chars` it doens't leak right? Only if you explicitly call for string.mb_chars.downcase, or I'm wrong?

> Yes this seems hugely dangerous, also "correct" Unicode upcase/downcase is CRAZY slow and expensive.

This get's called on Topic creation and edition right? I'm using active support since it already comes with rails, and don't think it is a bad tradeoff. Today `title prettify` can't be enabled on portuguese, russian, turkish, german and more communities since it makes titles worst. :cry:

lib/text_cleaner.rb
```diff
@@ -27,9 +31,9 @@ def self.clean(text, opts = {})
     # Replace ????? with a single ?
     text.gsub!(/\?+/, '?') if opts[:deduplicate_question_marks]
     # Replace all-caps text with regular case letters
-    text.tr!('A-Z', 'a-z') if opts[:replace_all_upper_case] && (text =~ /[A-Z]+/) && (text == text.upcase)
+    text = text.mb_chars.downcase.to_s if opts[:replace_all_upper_case] && (text =~ /[A-Z]+/) && (text == text.upcase)
     # Capitalize first letter, but only when entire first word is lowercase
-    text.sub!(/\A([a-z]*)\b/) { |first| first.capitalize } if opts[:capitalize_first_letter]
+    text = text.mb_chars.capitalize.to_s if opts[:capitalize_first_letter] && (text =~ /\S+/) && text.split(' ').first == text.split(' ').first.mb_chars.downcase
```
SamSaffron: a bit confusing and I am pretty sure it is broken and capitalizing full sentence :) something like this? 

```
first, rest = text.split(' ',2)
if opts[:capitalize_first_letter] && first == first.mb_chars.downcase
   text = "#{first.mb_chars.capitalize}#{rest?" " + rest : ""}" 
end
```

spec/components/text_cleaner_spec.rb
```diff
@@ -187,6 +187,14 @@
         expect(TextCleaner.clean_title("Hello there ?")).to eq("Hello there?")
       end
 
+      it "replaces all upper case unicode text with regular unicode case letters" do
+        expect(TextCleaner.clean_title("INVESTIGAÇÃO POLÍTICA NA CÂMARA")).to eq("Investigação política na câmara")
+      end
+
+      it "capitalizes first unicode letter" do
+        expect(TextCleaner.clean_title("épico encontro")).to eq("Épico encontro")
```
SamSaffron: change spec to 

```
expect(TextCleaner.clean_title("épico Encontro")).to eq("Épico Encontro")
```

SamSaffron:
@xfalcox  any news here?

xfalcox:
Yeah, this has been a pretty busy work week, but I'm finishing this on the
weekend :D

SamSaffron:
no probs! thanks

xfalcox:
Maybe now :+1:

SamSaffron:
lots of spec failure :( we got to get that sorted

xfalcox:
Its passing now!

coding-horror:
ok @SamSaffron can you review this one more time please

SamSaffron:
cool lets try this out!

PR #4063, opened 2016-03-07
Fix mbox.rb for updated Email::Receiver API

sghebuz:
This commit https://github.com/discourse/discourse/commit/30836573587079c5e663d7b3122957fc8c70dafe broke mbox importer.it.
Update the mbox importer for the new Email::Receiver API


SamSaffron:
seems good! thanks

PR #4064, opened 2016-03-07
Use sRGB Color Profile when converting images

darix:
With the conversion done when uploading or fetching an image we also
lose any embedded color profiles. [This leads to images possibly not rendered
properly in the browsers.][1] To fix the issue we tell imagemagick to render
the image from the embedded color profile to sRGB, which is the color
space used by most browsers.

RT_sRGB.icm is taken from the [RawTherapee repository][2] and is licensed as Public Domain.

[1]: https://meta.discourse.org/t/image-embedded-color-profile/40519
[2]: https://github.com/Beep6581/RawTherapee/blob/master/rtdata/iccprofiles/output/RT_sRGB.icm


coding-horror:
:+1: looks good, but how risky is this change? It is referencing a local file, will that file be present? @ZogStriP ?

darix:
the profile file is part of the PR

ZogStriP:
How is that different from using the `-colorspace sRGB` option?

patdavid:
@ZogStriP the `-colorspace sRGB` option is a rough conversion and not really appropriate for high quality image conversions (per im docs + forums)

patdavid:
__If__:

* the image has an embedded colorspace, this will convert it to sRGB (appropriate for browser viewing safely).
* the image does _not_ have an embedded colorspace, it's highly likely to be sRGB and thus this conversion will have no effect.
* the image does _not_ have an embedded colorspace, _but it was supposed to_, then it will be obvious to the poster that their colors are off and to fix the problem... :)

ZogStriP:
Let's try it. :+1:

darix:
deployed now.

PR #4065, opened 2016-03-07
Updated highlight.js

phw:
This updates highlightjs to the latest version, which provides additional syntax definitions and fixes to existing ones.

Please note that there was a cleanup of the highlight.js CSS, see https://github.com/isagalaev/highlight.js/blob/master/CHANGES.md and the new class reference at http://highlightjs.readthedocs.org/en/latest/css-classes-reference.html. I have tried to adapt the existing stylesheets to the new rules and tried to evaluate the affected languages. Given the amount of languages slight differences in highlighting are a possibility and in some cases unavoidable (e.g. for diff there was previously a distinction between "addition" and "change", now there is only the "addition" class for both cases).


coding-horror:
thanks! @SamSaffron will need to review this.

Freso:
@phw Perhaps update to pull in latest changes for taggerscript from hljs? :)

phw:
@Freso Done, I rebased the PR against master to keep it simple.

SamSaffron:
I worry about manual work here, we have a rake task for updating it, can you update it and run it here? otherwise we are setting us up a timebomb.

phw:
The update was performed with

    bin/rake update:highlights

The only manual work was updating the Discourse specific CSS to reflect the styling changes in highlight.js 0.9 :)

SamSaffron:
cool, sounds good then!

SamSaffron:
hmmm lang-auto is playing up big time for me ... no highlighting unless a lang is specified

SamSaffron:
ok added some code to strip lang-auto cause highlightjs is being strict about it.

phw:
Interesting. I only tested with explicit languages to see the highlighting, sorry for that

PR #4068, opened 2016-03-09
FEATURE: More granular mailing list mode

gdpelican:
As discussed here:
https://meta.discourse.org/t/more-granular-mailing-list-mode/38107

As it looks now:
screen shot 2016-03-09 at 10 03 45 pm

TODO:
- Determining when / how often to send grouped emails
- Designing email summary template

Would particularly like some feedback on the formula for email guesstimation here:
https://github.com/discourse/discourse/pull/4068/files#diff-08765a4c969575258af3892abcfa35a7R69

Currently it's looking at the number of posts created over the past week / month (with the past week average weighted 3x the past month average). Could likely do better around removing private messages, whispers, other stuff not visible to the user, etc., but figured I'd start with what was laying around.


mcwumbly:
I think you should just start with the more basic algorithm of just treating the past 4 weeks equally.

If something fancier seems warranted, then it might be worth getting more historical data from a handful of busy sites.

Plot out how many posts/week there are and see what kind of variance there is, how many weeks you need to smooth over to reduce that variance, and whether there is really and trending that would justify weighting the most recent week more heavily.

Until then, just keep it simple.

riking:
Did you consider setting `disabled` on the digest settings when this option is checked?

gdpelican:
@erlend-sh For the daily email, the spec says 'Time of sendout should be some sane default, like 6am (like the daily newspaper), based on the user's selected timezone.'

I've dug as much as I could, but haven't been able to find anywhere the user is associated with a timezone. The digest emails appear to sent out solely based on 'when we last emailed you / when we last saw you', which doesn't touch on what time zone they're in. Users have a 'location' field, but as far as I can tell, it's a vanity text field (mine on meta, for instance, is set to the ever-so-helpful ¯\_(ツ)_/¯)

Easiest might be to let the admin pick a UTC timezone hour to send out the digest, although having admins navigate what UTC is and what it means for their particular locale might be a bit tricky to explain. Thoughts?

gdpelican:
Pushing this forward a bit, I've come up with a decent-looking email template

screen shot 2016-03-23 at 1 57 20 pm


html: http://u.pomf.is/xlpnyj.html
plaintext: [text_mailing_list_template.txt](https://github.com/discourse/discourse/files/185251/text_mailing_list_template.txt)

Love to hear some feedback on them.

TODO:

- Quotes are messed up - 
In app:

screen shot 2016-03-23 at 2 11 22 pm

vs in email: 

screen shot 2016-03-23 at 2 10 57 pm

- File attachments seem to work with S3 (since they're absolute URLs), but not local file storage... not 100% on what to do about that one, I'll have to dig and see if there's an example of Discourse doing that elsewhere.

Also, gently nudge about the timezone dilemma @erlend-sh

erlend-sh:
> Easiest might be to let the admin pick a UTC timezone hour to send out the digest, although having admins navigate what UTC is and what it means for their particular locale might be a bit tricky to explain. Thoughts?

Could we do something similar to the digests (aka Activity Summary) so that the emails aren't all sent out at once? E.g. use the first recorded "when we last saw you" and keep it at that.

I wish Location wasn't a vanity text field. I'll put up a feature request for it next week.

erlend-sh:
Template looks good! Should be easier to nail down tiny tweaks after we've dog fooded it for a while. Merging will probably have to wait until after v1.5 is released.

@ZogStriP Any idea about the attachments? 
 
> File attachments seem to work with S3 (since they're absolute URLs), but not local file storage... not 100% on what to do about that one, I'll have to dig and see if there's an example of Discourse doing that elsewhere.

ZogStriP:
I need more info to help out. What's broken exactly?

gdpelican:
I'll have a look at doing something with the 'last seen at' for 

I just dug out the `PrettyText.format_for_email` method, which calls `make_all_links_absolute`; I reckon that'll sort me for that one.

Really happy to have no immediate urgency on this; what's the ETA for a 1.5 release? I can aim for around then for this to be ready.

Thanks for the responses!

erlend-sh:
One tiny addition to the spec: We want a :warning: SEVERE WARNING :warning: when someone tries to enable "Send me an email for every new post". cc @coding-horror 

I suggest we put up an extra confirmation box **if** the average daily post count is 2 or higher.

> "Are you sure? Enabling this means we'll send you an email for *every single new post on this forum*, which currently averages X new posts a day"

Sidebar: How are you calculating the "Email for every new post" average?

gdpelican:
Sounds good.

'Email for every post' is currently the number of posts over the last 30
days, divided by 30.

```
Post.where('created_at > ?', 30.days.ago).count.to_f / 30
```

It's the simplest thing, but it's also a mild overestimate, since it
includes all posts, including ones in muted categories and stuff which the
user won't actually receive emails for.

erlend-sh:
Turns out we want the feature for v1.5, so yo don't have to touch this for now. This does affect your PR though, so we can resume the "SEVERE WARNING" addition post-v1.5 and see then if any change is required on your end.

lib/pretty_text.rb
```diff
@@ -371,30 +371,14 @@ def self.strip_links(string)
...
   def self.strip_image_wrapping(doc)
     doc.css(".lightbox-wrapper .meta").remove
   end
 
-  def self.format_for_email(html)
-    doc = Nokogiri::HTML.fragment(html)
-    make_all_links_absolute(doc)
-    strip_image_wrapping(doc)
-    doc.to_html
+  def self.format_for_email(html, style: nil)
```
gdpelican: We need to format all of the posts properly for the mailing list summary email, and all that logic lives in `Email::Styles`, so I've swapped it out here. I thought this would work the same way as the digest, but since the digest only deals with post excerpts (ie the first paragraph), it sidesteps a lot of the issues with rendering full posts via email. I don't think there's an existing email which fully renders multiple posts at the same time?

I was pleasantly surprised to see that calling it this way actually takes less time without a style applied (so, for all existing calls to `format_for_email`):

```
existing_format_for_email  4.000000   0.320000   4.320000 (  5.531362)
new_format_for_email       3.600000   0.280000   3.880000 (  4.845140)
```
(run with a medium-sized post, 100x)

However, adding the `notification` style (which calls `Email::Styles.format_notification`, and applies all the bells and whistles normally applied to the user notification emails, including the existing mailing list email) slows it down significantly:

```
with_notification_style 28.250000   0.950000  29.200000 ( 32.747481)
```
which is almost 10x slower and .02 seconds per post... seems like a lot if we're sending out emails with heaps of posts in them. But we do need to run it in order to get things like quotes and oneboxes looking decent.

Here's a breakdown of the things that `format_notification` is calling (again, at 100x for a medium-sized post):
```
total                            2.470000   0.090000   2.560000 (  2.670170)
style_times_13                   0.870000   0.020000   0.890000 (  0.887273) # (we call the 'style' method 13 times for different style applications)
correct_first_body_margin        0.070000   0.000000   0.070000 (  0.074445)
correct_footer_style             0.070000   0.000000   0.070000 (  0.071749)
reset_tables                     0.060000   0.000000   0.060000 (  0.060174)
onebox_styles                    0.920000   0.030000   0.950000 (  0.940647)
plugin_styles                    0.440000   0.030000   0.470000 (  0.567873)
```
tl;dr it's mostly the calls to `styles` and onebox.

Thoughts would be to either render the whole html email, and run `format_notification` once on the whole thing (might be quicker?), or have each post store an additional 'email' version of itself in the DB, alongside 'raw' and 'cooked'. Both seem somewhat hard to me, and would rather go with something easier for this PR. Maybe I'm missing something simpler?

Maybe @ZogStriP or perf-master @SamSaffron has some thoughts on this?

gdpelican:
I've updated this to send out the digest email on the day-iversary of the first time the user was seen (so, not on the first day, but each day after that); this job runs hourly, so it should catch everyone. Those interested in DB diving could change the behaviour by updating `first_seen_at` on the user (and another iteration would probably do well to pull this out into a more intuitive, customizable value)

I reckon this PR is about feature-complete now; and would be happy for a CR.

erlend-sh:
Excellent :smile: I'm pretty sure this'll have to wait until we kick off the v1.6 branch, so please nudge us about this first thing next week.

gdpelican:
I've spruced this up to incorporate the mailing list warning from 1.5, yay!

gdpelican:
Thought I'd kick the tires on this again; anything I should be doing to move this along?

ZogStriP:
It looks good. We just release 1.5 and are waiting a couple of weeks for things to stabilise before doing/merging large piece of work ;)

erlend-sh:
Is this good to go now @ZogStriP ?

ZogStriP:
Sorry for the delay 😳  Looks like this needs a rebase now :(

ZogStriP:
Weird, I wish I had been notified of your rebase. Unfortunately, it seems that the specs are not passing. Can you fix that @gdpelican? (also, please @ mention me so that I can be notified and merge as soon as possible ;) )

gdpelican:
Yep, I'm going to get this over the line in the next couple days. Sorry for
the delay. :D

gdpelican:
Ok tests passing @ZogStriP !

PR #4070, opened 2016-03-10
FEATURE: revert post to a specific revision

techAPJ:
This PR adds a "Revert" button under the bottom of every revision. Clicking on this button will revert the post to that specific revision.

screen shot 2016-03-10 at 18 10 10

Revert button will only be visible to staff, and reverting a post will create a new revision.

@eviltrout can you please review this PR?

### DIFF
--- a/app/assets/javascripts/discourse/controllers/history.js.es6
+++ b/app/assets/javascripts/discourse/controllers/history.js.es6
@@ -29,6 +29,25 @@ export default Ember.Controller.extend(ModalFunctionality, {
     Discourse.Post.showRevision(postId, postVersion).then(() => this.refresh(postId, postVersion));
   },
 
+  revert(post, postVersion) {
+    post.revertToRevision(postVersion).then((result) => {
+      this.refresh(post.get('id'), postVersion);
+      if (result.topic) {
+        post.set('topic.slug', result.topic.slug);
+        post.set('topic.title', result.topic.title);
+        post.set('topic.fancy_title', result.topic.fancy_title);
+      }
+      if (result.category_id) {
+        post.set('topic.category', Discourse.Category.findById(result.category_id));
+      }
+      this.send("closeModal");
+    }).catch(function(e) {
+      if (e.jqXHR.responseJSON && e.jqXHR.responseJSON.errors && e.jqXHR.responseJSON.errors[0]) {
+        bootbox.alert(e.jqXHR.responseJSON.errors[0]);
+      }
+    });
+  },
+
   @computed('model.created_at')
   createdAtDate(createdAt) {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/discourse/controllers/history.js.es6
```diff
@@ -29,6 +29,23 @@ export default Ember.Controller.extend(ModalFunctionality, {
     Discourse.Post.showRevision(postId, postVersion).then(() => this.refresh(postId, postVersion));
   },
 
+  revert(postId, postVersion) {
+    Discourse.Post.revertToRevision(postId, postVersion).then((result) => {
+      this.refresh(postId, postVersion);
+      if (result.topic) {
+        // TO-DO: update topic title here
```
techAPJ: @eviltrout how should I update the topic title here?

eviltrout: Do you not have access to the post object that is being updated somehow? It should have an associated `topic` that you can update if the result contains the new title.

I'd suggest calling `revert(post, postVersion)` then you could say `post.set('topic.title', newTitle)`.

eviltrout: Also then you wouldn't need `Discourse.Post`, just a member function `post.revertToRevision(postVersion).then`...

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -333,6 +333,10 @@ Post.reopenClass({
     return Discourse.ajax("/posts/" + postId + "/revisions/" + version + "/show", { type: 'PUT' });
   },
 
+  revertToRevision(postId, version) {
+    return Discourse.ajax("/posts/" + postId + "/revisions/" + version + "/revert", { type: 'PUT' });
```
eviltrout: String interpolation might be useful here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Expression_interpolation

app/controllers/posts_controller.rb
```diff
@@ -282,6 +282,68 @@ def show_revision
     render nothing: true
   end
 
+  def revert
```
eviltrout: This is a really long method to be in a controller. It looks fairly clean, but the test coverage is fairly low considering all the branches you have. 

At the very least I think it should have a few more tests, but any way you could break it up (Sevice Class, more methods) would be good too.

techAPJ: Okay, I added more tests.

techAPJ:
@eviltrout I updated the PR as per your feedback. Can you please review again?

PR #4071, opened 2016-03-11
Add an Email Services Template page

sethherr:
- Link Admin Quick Start and basic install to services page
- Provide templates for various email service providers

I think this helps make setting up email services clearer...


coding-horror:
This is a good improvement, thanks! I think I want to change the link text a bit later though.

sethherr:
Yep, definitely not perfect. But... http://blog.codinghorror.com/moving-the-block/

PR #4072, opened 2016-03-13
Fixed bug where moved post links were incorrect

delgiudices:
https://meta.discourse.org/t/when-moving-posts-links-quotes-that-were-moved-too-dont-get-updated/13999/2

### DIFF
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -91,6 +91,16 @@ class PostMover
   def move(post)
     @first_post_number_moved ||= post.post_number
 
+    post.replies.each do |reply|
+      quotes = reply.raw.scan(/\[quote[^\]]+\]/)
+      quotes.each do |quote|
+        new_quote = quote.sub(/(?<=post:)[^,"]+/, @move_map[post.post_number].to_s)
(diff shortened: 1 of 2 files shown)


app/models/post_mover.rb
```diff
@@ -91,6 +91,12 @@ def create_first_post(post)
   def move(post)
     @first_post_number_moved ||= post.post_number
 
+    post.replies.each do |reply|
+      reply.raw = reply.raw.sub(/post:[1-9]+/, "post:#{@move_map[post.post_number]}")
```
SamSaffron: it should be gsub, but really this is far too wide a catch, what if the number is missing from @move_map? what if it happens not to be a quote and just some other part of the post

delgiudices:
Okay @SamSaffron it nows replace all instances of quotes and only inside [quote ]. About the @move_map, assuming move(post) is called by move_each_post then i think @move_map will be present, how do you think i should handle that case?

SamSaffron:
Sorry, it is still not safe enough, a post can contain quotes to other topics, not only quote in the current topic.

coding-horror:
I think we will have to attack this later after CommonMark support is merged

PR #4073, opened 2016-03-13
Fix: Link count on the topic sidebar

adrapereira:
The application doesn't display more than one link with the same title in the sidebar, but the count in the "X more..." link when the sidebar is collapsed wasn't reflecting that. 

The bug was discussed [here](https://meta.discourse.org/t/beginners-guide-to-creating-discourse-plugins/30515).

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-gutter.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-gutter.js.es6
@@ -43,7 +43,7 @@ export default createWidget('post-gutter', {
     if (state.collapsed) {
       var differentTitles = [], allTitles = [];
       links.forEach(function(x) { allTitles.push(x.title); });
-      differentTitles = allTitles.filter(function(item, i, ar){ return ar.indexOf(item) === i; });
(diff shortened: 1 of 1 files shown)


SamSaffron:
The fix feels a bit fishy, it feels like we are doing the same work in 2 spots. perhaps one piece of logic can be responsible for cleaning the list of links? 

Also, big question ... if there are 10 incoming from `https://meta.discourse.org/t/beginners-guide-to-creating-discourse-plugins/30515` and another 10 incoming from `https://meta.discourse.org/t/beginners-guide-to-creating-discourse-plugins/30515?foo` do we have 20 incoming for that title? Which one do we pick?

SamSaffron:
cc @eviltrout

adrapereira:
We're more or less doing the same work in 2 spots. In the while loop we don't always iterate the whole array, which we need to do to find the duplicated titles. If we want to make it a single piece of logic, we will have to iterate the whole array every time the method is called, even if there's no "X more..." link to be shown.

In the current code we pick the first link to appear in the array, discarding the others.

eviltrout:

PR #4074, opened 2016-03-13
Always trigger clearing of saved drafts from client after save promise resolves

angusmcleod:
When a post is created the clearing of any drafts of the post is triggered from the client. However when a post is edited, the clearing of drafts is handled on the server by incrementing the draft sequence. Both types of saved content should be triggered from the client. As well as improving consistency, this fixes edge cases like that discussed here: https://meta.discourse.org/t/draft-not-cleared-properly-when-empty-edit-of-post-is-saved/40939

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -294,18 +294,15 @@ export default Ember.Controller.extend({
     });
 
     const promise = composer.save({ imageSizes, editReason: this.get("editReason")}).then(function(result) {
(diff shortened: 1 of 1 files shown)


SamSaffron:
Ok, I am uneasy with adding the extra xhr here under any conditions, in fact we should eliminate them. 

Instead I would much prefer if we "unconditionally" increment the draft sequence on the server, its a far safer practice and would kill an extra XHR on every POST

angusmcleod:
@SamSaffron Ok, makes sense. I'll update this in the next day or so.

PR #4075, opened 2016-03-13
Fix Russian grammar

meritt:


meritt:
Done

SamSaffron:
Sorry, this needs to be done via transifex https://www.transifex.com/discourse/discourse-org/ 

if we allow it in here it will simply be overwritten next time we pull translations (weekly)

PR #4076, opened 2016-03-14
FEATURE: add site setting for setting locale from header

scossar:
See https://meta.discourse.org/t/different-behaviour-of-text-direction-for-not-logged-in-visitors/40747

Adds a site setting for setting the locale from accept language headers. The new setting defaults to false and is dependent on `allow_user_locale` being set to true.

screenshot 2016-03-14 16 01 05

### DIFF
--- a/lib/validators/allow_user_locale_enabled_validator.rb
+++ b/lib/validators/allow_user_locale_enabled_validator.rb
@@ -12,7 +12,7 @@ class AllowUserLocaleEnabledValidator
   end
 
   def error_message
-    "You must first enable 'allow user locale' before enabling this setting."
+    I18n.t("site_settings.errors.user_locale_not_enabled");
   end
 
 end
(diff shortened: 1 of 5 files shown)


coding-horror:
Excellent! Can you review this tomorrow @eviltrout ?

lib/validators/allow_user_locale_enabled_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def valid_value?(val)
+    # only validate when enabling setting locale from headers
+    return true if val == "f"
+    # ensure that allow_user_locale is enabled
+    SiteSetting.allow_user_locale
+  end
+
+  def error_message
+    "You must first enable 'allow user locale' before enabling this setting."
```
ZogStriP: This needs to be translatable ;)

scossar:
The error message is now translatable.

SamSaffron:
thanks

PR #4078, opened 2016-03-15
Advance draft sequence in PostRevisor if edit contains no changes

angusmcleod:
Simplest version of the server-side fix for
https://meta.discourse.org/t/draft-not-cleared-properly-when-empty-edit-of-post-is-saved/40939


ZogStriP:
Let's try it.

PR #4079, opened 2016-03-15
update redis dependency in documentation

ryantm:
b1e0da starts using the redis ROLE command which is available in redis-server 2.8.12 and later.


SamSaffron:
thanks

PR #4081, opened 2016-03-15
FIX: Length for hashtags, increased to 101 characters.

tomasibarrab:
I increased the char length for the hashtags to 101 characters with @tgxworld help, in order to solve this bug: https://meta.discourse.org/t/hashtag-linking-to-categories-doesn-t-always-work/41003/7


tomasibarrab:
Ready to be merged @coding-horror @tgxworld @ZogStriP, can we review it?

PR #4085, opened 2016-03-16
Fix plugin links.

JaredReisinger:
Change from `https://meta.discourse.org/category/extensibility/plugin` to `https://meta.discourse.org/c/plugin` so that they actually work!  Note that `/category/...` would continue to work (it just redirects to `/c/...`), but since the site itself appears to normalize to `/c/`, it makes sense for the links to follow the same pattern.


docs/PLUGINS.md
```diff
@@ -20,6 +20,5 @@ If you just want to get some plugins for your Discourse instance, check out [the
 ### Simple Demonstration Plugins
 
 - [Manipulating text in the composer](https://meta.discourse.org/t/plugin-tutorial-1-how-to-manipulate-the-text-in-the-composer/10925)
- 
-- [Display plugin outlet locations](https://meta.discourse.org/t/plugin-outlet-locations/29589)
 
+- [Display plugin outlet locations](https://meta.discourse.org/t/plugin-outlet-locations/29589)
```
JaredReisinger: Weird... not sure what happened here... maybe a line-ending (CR vs. CRLF) normalization?

JaredReisinger:
Hmm... unexpected build failure.  It _looks_ like a completely unrelated issue, but I'll take a closer look.

PR #4087, opened 2016-03-17
Fixed anonymizer when 'full name required' setting is on

jeremylan:
When the setting 'full name required' is on the anonymizer was trying to set the user name to nil and this caused the user name and email to remain not anonymized.
Now in this scenario the user name is set to the anonymized username and the email is anonymized correctly.

### DIFF
--- a/app/services/user_anonymizer.rb
+++ b/app/services/user_anonymizer.rb
@@ -20,7 +20,7 @@ class UserAnonymizer
       @user.reload
       @user.password = SecureRandom.hex
       @user.email = "#{@user.username}@example.com"
-      @user.name = nil
+      @user.name = SiteSetting.full_name_required ? @user.username : nil
       @user.date_of_birth = nil
       @user.title = nil
       @user.uploaded_avatar_id = nil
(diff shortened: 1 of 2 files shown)


ZogStriP:
I like this, but could you add a test to ensure we don't regress here?

jeremylan:
Sure @ZogStriP, I'll do that. Do you have any links to commits with tests? That would help me as this is my first pull request!
Thanks

ZogStriP:
I don't have a link to commits with tests, but you can look at https://github.com/discourse/discourse/blob/master/spec/services/user_anonymizer_spec.rb. This might help get you started ;)

Also, don't hesitate to ask if you have any questions.

jeremylan:
Thanks @ZogStriP, I've updated the test using context, hope it's ok.

SamSaffron:
thanks

SamSaffron:
also added this, https://github.com/discourse/discourse/commit/c2fa314684dacc319acd150dd1c6336f11e64987 cause tests were not covering this correctly

PR #4088, opened 2016-03-17
Add host to nginx logs.

tgxworld:
cc @SamSaffron

### DIFF
--- a/config/nginx.sample.conf
+++ b/config/nginx.sample.conf
@@ -25,7 +25,7 @@ map $http_x_forwarded_proto $thescheme {
   https https;
 }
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
this is likely to break the log analyzer, is there a reason we need this?

tgxworld:
@SamSaffron I was debugging the CDN issue we were seeing for enterprises on a  multisite setup. Not having the host makes it really difficult to determine where the request is coming from. :)

coding-horror:
Not a big fan of the log analyzer personally, when was the last time we even looked at that data?

PR #4090, opened 2016-03-17
FIX: About posts count should exclude private

DeanMarkTaylor:
Currently the topic count only includes public topics, this change makes the post count only include public posts to match.

See: https://meta.discourse.org/t/about-page-site-statistics-doesnt-match-admin-dashboard-stats/41166


DeanMarkTaylor:
Build failures seem unrelated to me.

eviltrout:
I think the decision in that topic was to not change this statistic. If anything else we should add a new statistic for private posts.

PR #4093, opened 2016-03-18
Move autosize from admin/lib to discourse/lib

angusmcleod:
See further here: https://meta.discourse.org/t/feasible-to-move-autosize-library-to-discourse-lib/41265


eviltrout:
Sure, I can see the value of using it in a plugin, and it's only 200 lines.

PR #4103, opened 2016-03-22
App Banner Support for Android Chrome

xfalcox:
https://meta.discourse.org/t/fully-support-android-chromes-add-to-homescreen/21067/17?u=falco

### DIFF
--- a/app/assets/javascripts/discourse/initializers/android-app-banner-service-worker.js.es6
+++ b/app/assets/javascripts/discourse/initializers/android-app-banner-service-worker.js.es6
@@ -1,0 +1,16 @@
+// Android Chrome App Banner requires at least **one** service worker to be instantiate and https.
+// After Discourse starts to use service workers for other stuff (like mobile notification, offline mode, or ember)
+// we can ditch this.
+
+export default {
+  name: 'android-app-banner-service-worker',
+
+  initialize(container) {
+    const caps = container.lookup('capabilities:main');
+    const isSecure = document.location.protocol === 'https:';
+    
+    if (isSecure && caps.isAndroid && 'serviceWorker' in navigator) {
+        navigator.serviceWorker.register(Discourse.BaseUri + '/service-worker.js', {scope: './'});
+    }
+  }
+}
--- a/app/assets/javascripts/discourse/initializers/android-app-banner-service-worker.js.es6
+++ b/app/assets/javascripts/discourse/initializers/android-app-banner-service-worker.js.es6
@@ -8,9 +8,9 @@ export default {
   initialize(container) {
     const caps = container.lookup('capabilities:main');
     const isSecure = document.location.protocol === 'https:';
-    
+
     if (isSecure && caps.isAndroid && 'serviceWorker' in navigator) {
         navigator.serviceWorker.register(Discourse.BaseUri + '/service-worker.js', {scope: './'});
     }
   }
-}
+};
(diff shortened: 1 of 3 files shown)


eviltrout:
Seems a bit wasteful to instantiate a service worker that does nothing just for one platform.

Perhaps it should only be spawned on android if it is only necessary on that platform?

xfalcox:
That's a good idea @eviltrout :+1: 

There is user agent detection already set up somewhere client side?

Also, I'm reading Push Notification API for Chrome and Firefox on Android, we can re-purpose this service worker later.

eviltrout:
We do have client side detection of android, but it won't be set up in the `_discourse_javascript` file.

Instead I suggest you create a new initializer. in the initializer you can ask the container for the capabilities object and check `isAndroid`:

You can do something like this (written off the top of my head so there might be bugs, you can compare against our other initializers): 

```javascript
export default {
  name: 'android-service-work',
  after: 'sniff-capabilities',
  initialize(container) {
    const caps = container.lookup('capabilities-main');
    if (caps.isAndroid) {
      // initialize service work
    }
  }
}
```

coding-horror:
This should wait for 1.6 before merge

coding-horror:
How do you feel about this now @eviltrout

eviltrout:
I'd still prefer this to only execute on android before merging.

xfalcox:
Did the changes :fireworks:

app/assets/javascripts/discourse/initializers/android-app-banner-service-worker.js.es6
```diff
@@ -0,0 +1,22 @@
...
+
+  initialize(container) {
+    const caps = container.lookup('capabilities:main');
+    if (caps.isAndroid) {
+      if ('serviceWorker' in navigator) {
+        navigator.serviceWorker.register(Discourse.BaseUri + '/service-worker.js', {scope: './'})
+          .then(function (r) {
+            // Service Worker Success
+          })
+          .catch(function (whut) {
```
tgxworld: Perhaps rename `whut` to just `e` or `error`?

tgxworld: Actually if we're not doing anything with the response, we should leave out the calls. `then` and `catch`

xfalcox:
@tgxworld removed the unnecessary bits, and added a test for https, since it obligatory.

tgxworld:
@xfalcox Thanks! The PR needs a rebase :)

xfalcox:
Okay, I think it's rebased :smile:

xfalcox:
I guess that

eviltrout:
Actually we don't need to ask people to rebase anymore @tgxworld. When
merging github now has an option to squash commits :)

xfalcox:
Ruby 2.0 build failed, but I think it's unrelated.

tgxworld:
@eviltrout but we'll still need to rebase if there is a merge conflict.

xfalcox:
Hey @tgxworld  anything else?

coding-horror:
if this looks OK to you can we merge it @eviltrout ?

eviltrout:
Yes it looks good. Thanks for the effort and back and forth @xfalcox !

xfalcox:
Thanks everyone, it's already working on meta! :fireworks:

If you guys wanna test faster there is a flag on chrome://flags   #bypass-app-banner-engagement-checks

PR #4106, opened 2016-03-22
UPDATE: VAGRANT.md, link to become eviltrout admin.

tomasibarrab:
I updated the VAGRANT.md file to add a link to become **eviltrout** instead of only a label, as is posted on this section:

> In a few seconds, rails will start serving pages. To access them, open a web browser to [http://localhost:4000](http://localhost:4000) - if it all worked you should see discourse! Congratulations, you are ready to start working!


tomasibarrab:
Great, i did it because it was a link hard to see, because it was posted as a label. Thank you @ZogStriP, cheers!

PR #4108, opened 2016-03-22
add ability to have post-menu button in second position

stevenpslade:
This applies to buttons created with `api.addPostMenuButton`. The case `second` will place the button after index 0.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-menu.js.es6
@@ -270,6 +270,9 @@ export default createWidget('post-menu', {
               case 'first':
                 visibleButtons.unshift(button);
                 break;
+              case 'second':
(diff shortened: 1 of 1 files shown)


ZogStriP:
Maybe `position` should be an index rather than a string (`first`, `second`...)

eviltrout:
The problem with doing it that way is the hidden elements. Often plugins want to strategically position elements around that hidden content (before it, inside it, outside it). A standard index would not reflect that.

stevenpslade:
I thought of this as well. If someone wants second position, why not third, fourth, etc? Is it ridiculous to have more case statements to handle up to five? Or is it too much of a slippery slope?

stevenpslade:
@ZogStriP @eviltrout anything else needed for this?

eviltrout:
Nope looks good to me, thanks!

PR #4109, opened 2016-03-23
FIX: posts.rss route  when left clicked.

tomasibarrab:
I made a quick fix for the route **/posts.rss** when it's right clicked, the bug is that the path is not being sent to the server or being handle by Ember, it just don't send the post, but this quick fix works. We need to check out how the **handleURL** function works.

This PR fix this bug:

https://meta.discourse.org/t/left-clicking-link-to-posts-rss-does-not-work/41437

### DIFF
--- a/app/assets/javascripts/discourse/lib/url.js.es6
+++ b/app/assets/javascripts/discourse/lib/url.js.es6
@@ -99,7 +99,17 @@ const DiscourseURL = Ember.Object.createWithMixins({
   **/
 
   isInternalPathButNotEmber: function(path){
-    return path.indexOf("posts.rss") >= 0
+   /*
+   RSSwhitelist = ["posts.rss", "support.rss", "bug.rss", "feature.rss", "dev.rss", "ux.rss", "installation.rss",
+   "meta.rss", "marketplace.rss", "hosting.rss", "howto.rss", "plugin.rss", "praise.rss", "blog.rss", "faq.rss", "releases.rss"];
+
+   This array contains some of the .rss routes on Discourse, several of them are categories, and the method is
+   extracting the .rss route and verifying if the indexOf is >= to zero.
+   */
+
+   let pathStripped =  path.split("/").pop();
+   return path.indexOf(pathStripped) >= 0;
+
   },
 
   routeTo(path, opts) {
--- a/app/assets/javascripts/discourse/lib/url.js.es6
+++ b/app/assets/javascripts/discourse/lib/url.js.es6
@@ -97,10 +97,15 @@ const DiscourseURL = Ember.Object.createWithMixins({
     It contains the logic necessary to route within a topic using replaceState to
     keep the history intact.
   **/
+
+  isInternalPathButNotEmber: function(path){
+    return path.indexOf("posts.rss") >= 0
+  },
+
   routeTo(path, opts) {
     if (Em.isEmpty(path)) { return; }
 
-    if (Discourse.get('requiresRefresh')) {
+    if (Discourse.get('requiresRefresh') || this.isInternalPathButNotEmber(path)) {
       document.location.href = Discourse.getURL(path);
       return;
     }


tomasibarrab:
@ZogStriP  here it is, i closed it before, re-open now.

ryantm:
This PR might fix /posts.rss and lay a groundwork for fixing other static links, but I don't see how it would fix /bug.rss or other URLs that depend on category slugs.

tomasibarrab:
This is what i told you, is not a fix, it's a quick fix only, the bug is really weird to me actually, because the route exists, and the page is available, but something goes messy when you made a **click event** on the link, something goes wrong there, i debugged it yesterday and i found that it is in that method that something goes wrong, the */posts.rss* is lost and throws the 404 error.

riking:
I think you'd do better to insert code here:

        const transition = router.handleURL(path);

Check if the transition is going to the "not found" page, issue a HEAD request for the URL, if 2xx then assign document.location.

tomasibarrab:
I don't think that the idea from @riking  would work, because you will need to have a whitelist to check those static urls all the time, however, i will fix this in order to make a **groundwork** as @ryantm says, just let me do a list of some *.rss* pages, and i will commit that and see what's next.

Thank you guys, cheers!

tomasibarrab:
Updated PR, i did a Groundwork for the .rss routes. @ZogStriP

tomasibarrab:
@coding-horror @ZogStriP  @SamSaffron  @tgxworld  can we review what did happen?

erlend-sh:
We appreciate the pull requests @tomasibarrab, but please be patient with us. Everyone in the team keeps an eye on pending PRs, so there's no need to mention a group of people like that. v1.5 is due out in a few days, so the developers are most likely occupied for the time being.

tomasibarrab:
@erlend-sh it's ok, don't worry! cheers!

SamSaffron:
anything .rss or anything .json should simply bypass the ember router and do a full page load

tomasibarrab:
where's the problem then?

ryantm:
@tomasibarrab Doesn't your change do something way more specific than what Sam says?

tomasibarrab:
I only created a validator that checks if the value of the index is equals or greater than zero, this works for me on my locale when i left click the link.

``` js
let pathStripped =  path.split("/").pop();
return path.indexOf(pathStripped) >= 0;
```

just i'm just obtaining the **posts.rss** in a variable called ***pathStripped*** and verifying the value of the index of **posts.rss** (in this case) and if this value if equals or greater than zero, it works, i think that the value is 26 for **posts.rss**

ryantm:
Do URLs like `https://meta.discourse.org/t/left-clicking-link-to-posts-rss-does-not-work/41437` still not cause a refresh? It seems like `isInternalPathButNotEmber` should return `true` for that URL.

tomasibarrab:
It will only return ```true``` if the path after the last ```/``` is within the routes library and have an index value

coding-horror:
I would like to know how @eviltrout feels about this, non-Ember links have been a problem (but a very rarely encountered one) since day 1 of the project and it has not caused much pain, nor been much of an inconvenience..

app/assets/javascripts/discourse/lib/url.js.es6
```diff
@@ -97,10 +97,25 @@ const DiscourseURL = Ember.Object.createWithMixins({
...
+   /*
+   RSSwhitelist = ["posts.rss", "support.rss", "bug.rss", "feature.rss", "dev.rss", "ux.rss", "installation.rss",
+   "meta.rss", "marketplace.rss", "hosting.rss", "howto.rss", "plugin.rss", "praise.rss", "blog.rss", "faq.rss", "releases.rss"];
+
+   This array contains some of the .rss routes on Discourse, several of them are categories, and the method is
+   extracting the .rss route and verifying if the indexOf is >= to zero.
+   */
+
+   let pathStripped =  path.split("/").pop();
+   return path.indexOf(pathStripped) >= 0;
```
eviltrout: I don't understand how this works. You're taking the last segment from the path, but then checking that it's in the path? Wouldn't it always be true unless it's a root path?

Can you provide more examples?

I would be more comfortable with what @SamSaffron suggested and just check for an extension of `.rss` or `.json`

PR #4110, opened 2016-03-23
Fix for category stripe misalignment

gdpelican:
Before:
screen shot 2016-03-23 at 2 45 51 pm

After:
screen shot 2016-03-23 at 2 45 56 pm

### DIFF
--- a/lib/category_badge.rb
+++ b/lib/category_badge.rb
@@ -50,7 +50,7 @@ module CategoryBadge
       case (SiteSetting.category_style || :box).to_sym
       when :bar then inline_category_stripe(category.color, 'display: inline-block; padding: 1px;', true)
       when :box then inline_category_stripe(category.color, 'left: 5px; display: block; position: absolute; width: calc(100% - 5px); height: 100%;')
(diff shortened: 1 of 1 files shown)


coding-horror:
Errrr before and after look identical?

gdpelican:
Sorry, double screenshotted :)

I was testing it out using `UserNotifications.digest(User.last).deliver` and inspecting in mailcatcher.

gdpelican:
The bug was that it was applying a `&nbsp;` character to the subcategory stripe, but not the parent category stripe.

For category: (https://github.com/discourse/discourse/blob/master/lib/category_badge.rb#L41)
```
when :bullet then inline_category_stripe(parent_category.color, "<styles>")
```

For subcategory: (https://github.com/discourse/discourse/blob/master/lib/category_badge.rb#L53)
```
when :bullet then inline_category_stripe(category.color, "<styles>", true)
```

(with 'true' being the 'insert_blank' param for that method)

coding-horror:
Ok let's try it then

PR #4111, opened 2016-03-23
Update automated account blocked by staff PM text

jomaxro:
When a user's account is blocked by staff, they receive an automated PM informing them of such.  Unlike the new user too many flags PM, this PM does not inform that user what blocked means.  This PR adds two lines to the PM, one explaining what blocking means, the other informing the user to contact staff with questions.


coding-horror:
sure, better than what we had! thanks!

PR #4112, opened 2016-03-23
Feature: Merge multiple responses

adrapereira:
This feature was requested in meta [here](https://meta.discourse.org/t/add-a-merge-multiple-responses-admin-post-selection-action/14904).

It adds a button to a topic's administration menu enabling the admin to merge 2 or more posts by the same user.

**Necessary conditions to merge posts:**
* All posts must be deletable (so the main post of a topic cannot be among the selected posts)
* 2 or more posts must be selected
* All posts must be by the same user

There's no additional logic to account for posts that are replies to other responses, all posts will be merged in the same way, chronologically, with two new lines between each one.

The merged content will be edited into the last selected posts while the remaining responses will be marked as deleted.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -517,6 +517,25 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
       });
     },
 
+    mergePosts() {
+      bootbox.confirm(I18n.t("post.delete.confirm", { count: this.get('selectedPostsCount')}), result => {
+        if (result) {
+          const selectedPosts = this.get('selectedPosts');
+          const selectedReplies = this.get('selectedReplies');
+          const postStream = this.get('model.postStream');
+
+          Discourse.Post.mergePosts(selectedPosts, selectedReplies);
+          //postStream.get('posts').forEach(p => {
+          //  if (this.postSelected(p)) {
+          //    p.set('deleted_at', new Date());
+          //  }
+          //});
+
+          this.send('toggleMultiSelect');
+        }
+      });
+    },
+
     expandHidden(post) {
       post.expandHidden();
     },
@@ -691,6 +710,11 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
     if (!Discourse.User.current() || !Discourse.User.current().admin) return false;
     return this.get('selectedPostsUsername') !== undefined;
   }.property('selectedPostsUsername'),
+
+  canMergePosts: function() {
+    if (this.get('selectedPostsCount') < 2) return false;
+    return this.get('selectedPostsUsername') !== undefined;
+  }.property('selectedPostsCount'),
 
   categories: function() {
     return Discourse.Category.list();
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -522,7 +522,7 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
         if (result) {
           const selectedPosts = this.get('selectedPosts');
 
-          Discourse.Post.mergePosts(selectedPosts);
+          Post.mergePosts(selectedPosts);
           this.send('toggleMultiSelect');
         }
       });
@@ -709,12 +709,7 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
     // Must be able to delete all of the posts, this makes it so
     // that the main post of a topic can't be part of the merged posts
     const selectedPosts = this.get('selectedPosts');
-    let canDelete = true;
-    selectedPosts.forEach(function(p) {
-      if (!p.get('can_delete')) {
-        canDelete = false;
-      }
-    });
+    let canDelete = selectedPosts.every(p => p.get('can_delete'));
     if(!canDelete) return false;
     // All the posts must have the same creator
     return this.get('selectedPostsUsername') !== undefined;
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -712,7 +712,7 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
     const selectedPosts = this.get('selectedPosts');
     let canDelete = true;
     selectedPosts.forEach(function(p) {
-      if (!canDelete || !p.get('can_delete')) {
+      if (!p.get('can_delete')) {
         canDelete = false;
       }
     });
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -522,7 +522,6 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
         if (result) {
           const selectedPosts = this.get('selectedPosts');
           const selectedReplies = this.get('selectedReplies');
-          const postStream = this.get('model.postStream');
 
           Discourse.Post.mergePosts(selectedPosts, selectedReplies);
           this.send('toggleMultiSelect');
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -522,7 +522,7 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
         if (result) {
           const selectedPosts = this.get('selectedPosts');
 
-          Post.mergePosts(selectedPosts);
+          Discourse.Post.mergePosts(selectedPosts);
           this.send('toggleMultiSelect');
         }
       });
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -518,19 +518,13 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
     },
 
     mergePosts() {
-      bootbox.confirm(I18n.t("post.delete.confirm", { count: this.get('selectedPostsCount')}), result => {
+      bootbox.confirm(I18n.t("post.merge.confirm", { count: this.get('selectedPostsCount')}), result => {
         if (result) {
           const selectedPosts = this.get('selectedPosts');
           const selectedReplies = this.get('selectedReplies');
           const postStream = this.get('model.postStream');
 
           Discourse.Post.mergePosts(selectedPosts, selectedReplies);
-          //postStream.get('posts').forEach(p => {
-          //  if (this.postSelected(p)) {
-          //    p.set('deleted_at', new Date());
-          //  }
-          //});
-
           this.send('toggleMultiSelect');
         }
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -563,6 +575,23 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
 
+  canMergePosts: function() {
+    // Two or more posts selected
+    if (this.get('selectedPostsCount') < 2) return false;
+    // Must be able to delete all of the posts, this makes it so
+    // that the main post of a topic can't be part of the merged posts
+    const selectedPosts = this.get('selectedPosts');
+    let canDelete = true;
+    selectedPosts.forEach(function(p) {
+      if (!canDelete || !p.get('can_delete')) {
```
riking: You can leave out the first part of this condition here.

lib/post_merger.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def merge
+    postContent = []
+    @posts.each {|p| postContent.push(p.raw) }
+
+    post = @posts.last
+    changes = {
+      raw: postContent.join("\n\n"),
+      edit_reason: "Merging the posts"
```
riking: This should be a better message.

    Merged #{@posts.length} posts by #{@posts.first.creator}

app/controllers/posts_controller.rb
```diff
@@ -237,6 +238,29 @@ def destroy_many
...
   end
 
+  def merge_posts
+    params.require(:post_ids)
+
+    posts = Post.where(id: post_ids_including_replies).order(:id)
+    raise Discourse::InvalidParameters.new(:post_ids) if posts.blank?
+
+    # Make sure we can delete the posts
+    posts.each {|p| guardian.ensure_can_delete!(p) }
```
tgxworld: I think the logic here should be moved into `PostMerger`.

app/controllers/posts_controller.rb
```diff
@@ -237,6 +238,29 @@ def destroy_many
...
+
+    PostMerger.new(current_user, posts).merge
+
+    Post.transaction do
```
tgxworld: Same as https://github.com/discourse/discourse/pull/4112/files#r63487573

tgxworld:
Hi @adrapereira !

Thank you for the PR :)

Are you able to add some tests to make sure `PostMerger` is doing what you expect it to do? Feel free to ping me if you need any help with this? Thanks

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -320,6 +320,16 @@ Post.reopenClass({
...
   },
 
+  mergePosts(selectedPosts, selectedReplies) {
+    return Discourse.ajax("/posts/merge_posts", {
+      type: 'PUT',
+      data: {
+        post_ids: selectedPosts.map(function(p) { return p.get('id'); }),
+        reply_post_ids: selectedReplies.map(function(p) { return p.get('id'); })
+      }
+    });
```
tgxworld: It seems like we're not updating the client and handling errors on the ajax calls here.

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -320,6 +320,16 @@ Post.reopenClass({
     });
   },
 
+  mergePosts(selectedPosts, selectedReplies) {
+    return Discourse.ajax("/posts/merge_posts", {
+      type: 'PUT',
+      data: {
+        post_ids: selectedPosts.map(function(p) { return p.get('id'); }),
```
tgxworld: Nitpick here: Lets do `selectedPosts.map(p => p.get('id'))` instead

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -320,6 +320,16 @@ Post.reopenClass({
...
+        reply_post_ids: selectedReplies.map(function(p) { return p.get('id'); })
```
tgxworld: I don't see this being used anywhere in the controller :grin:

app/controllers/posts_controller.rb
adrapereira: Any part in specific? It's the same way as the other methods in the controller. Example: https://github.com/discourse/discourse/blob/master/app/controllers/posts_controller.rb#L162

app/assets/javascripts/discourse/models/post.js.es6
adrapereira: @tgxworld The client is updated. :smiley:  There's no error handling in the AJAX call since the other AJAX calls didn't have it either. Can you point me to an example of how it should be handled in Discourse?

coding-horror:
@tgxworld if this still looks good, can we merge it in?

adrapereira:
@coding-horror Not yet, I still need to add some tests.

adrapereira:
@tgxworld I've added some tests to PostMerger. Please let me know if you need anything more.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -410,6 +410,17 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
       });
     },
 
+    mergePosts() {
+      bootbox.confirm(I18n.t("post.merge.confirm", { count: this.get('selectedPostsCount')}), result => {
+        if (result) {
+          const selectedPosts = this.get('selectedPosts');
+
+          Discourse.Post.mergePosts(selectedPosts);
```
tgxworld: Let's import the `Post` model instead.
`import Post from 'discourse/models/post';`

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -563,6 +574,23 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
   }.property('selectedPostsUsername'),
 
+  canMergePosts: function() {
+    // Two or more posts selected
+    if (this.get('selectedPostsCount') < 2) return false;
+    // Must be able to delete all of the posts, this makes it so
+    // that the main post of a topic can't be part of the merged posts
+    const selectedPosts = this.get('selectedPosts');
+    let canDelete = true;
+    selectedPosts.forEach(function(p) {
```
tgxworld: You can probably do something like `selectedPosts.some(p => p.get('can_delete'))`

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -320,6 +320,13 @@ Post.reopenClass({
     });
   },
 
+  mergePosts(selectedPosts) {
+    return Discourse.ajax("/posts/merge_posts", {
+      type: 'PUT',
+      data: {post_ids: selectedPosts.map(p => p.get('id'))}
+    });
```
tgxworld: I realized we are not handling errors here :)

app/controllers/posts_controller.rb
tgxworld: Opps nevermind this is fine :+1:

lib/post_merger.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def initialize(user, posts)
+    @user = user
+    @posts = posts
+  end
+
+  def merge
+    postContent = []
+
+    @posts.each {|p| postContent.push(p.raw) }
```
tgxworld: nitpick:  Missing space after `{|p|`

app/controllers/posts_controller.rb
```diff
@@ -237,6 +238,20 @@ def destroy_many
     render nothing: true
   end
 
+  def merge_posts
+    params.require(:post_ids)
+
+    posts = Post.where(id: post_ids_including_replies).order(:id)
+    raise Discourse::InvalidParameters.new(:post_ids) if posts.blank?
```
tgxworld: Hmm we can't just check for a blank here. Probably do a `map(&:id)` to make sure the ids are equal the `post_ids`. Otherwise, there might be cases where one of the id is valid we will not know about it.

lib/post_merger.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    post = @posts.last
+    changes = {
+      raw: postContent.join("\n\n"),
+      edit_reason: "Merged #{@posts.length} posts by #{@user.name}"
+    }
+    revisor = PostRevisor.new(post, post.topic)
```
tgxworld: I think the revision should happen within the transaction as well

spec/components/post_merger_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      reply2 = create_post(topic: topic)
+      reply3 = create_post(topic: topic)
+
+      replies = []
+      replies.push(reply1)
+      replies.push(reply2)
+      replies.push(reply3)
+
+      postContent = []
+      replies.each {|p| postContent.push(p.raw) }
```
tgxworld: Hmm I'm not a fan of this test because it is duplicating the logic of what `PostMerger` does in the test. 

What I usually do is manually form the expected output, so it'll be something like

```
create_post(topic: topic, raw: 'this is the first line')
create_post(topic: topic, raw: 'this is the second line')
create_post(topic: topic, raw: 'this is the third line')

expected_output = <<~OUTPUT
this is the first line

this is the second line

this is the third line
OUTPUT
```

tgxworld:
@adrapereira Sorry for the late reply. I've left some comments :)

app/assets/javascripts/discourse/models/post.js.es6
adrapereira: Can you tell me what I should do when if there's an error? There's no error handling in any of the AJAX calls of that model.

app/controllers/posts_controller.rb
adrapereira: I'm sorry but I don't understand what you're asking, can you please specify?

app/assets/javascripts/discourse/models/post.js.es6
tgxworld: Sure :) For a start, https://github.com/discourse/discourse/blob/be3a5a56ccc284b352aa65080fc6b955f73cc72a/app/assets/javascripts/admin/controllers/admin-site-text-edit.js.es6#L13 will be fine

app/controllers/posts_controller.rb
tgxworld: If `post_ids_including_replies` is equal to `[1, 2, 3]` where only `[1, 2]` are valid ids. The check here won't catch that `3` is invalid because we are only checking if `posts` collection is blank.

SamSaffron:
@adrapereira  branch is in conflict state now, can you rebase?

app/controllers/posts_controller.rb
adrapereira: I understand why we should do it but couldn't figure out how to implement it... :worried:

tgxworld: Sorry for the delay in getting back to you :(

You can do something like `posts.pluck(:id) == params[:post_ids]`.

coding-horror:
This seems in good shape if you can rebate?

adrapereira:
I think this is ready for merge.

tgxworld:
Closing in favor of https://github.com/discourse/discourse/pull/4346

PR #4115, opened 2016-03-24
Mbox importer improvements (attachments and better subject strip)

sghebuz:
Improve the mbox importer: 

1. strip localized reply prefixes (just Italian right now but more can be easily added)
2. strip localized and standard forward prefixes (just Italian right now but more can be easily added)
3. Import attachments on the emails to the topic/reply


SamSaffron:
cool lets try this out!

PR #4118, opened 2016-03-27
Add .byebug_history to gitignore

gdpelican:
Using byebug results in a `.byebug_history` file which stores a list of the previous commands fed to byebug, which doesn't belong in the repo and can be a bit annoying to step over any time a commit is made. Should be safe to git ignore.


tgxworld:
:+1: Been running into the same thing

PR #4122, opened 2016-03-29
Added Lets Encrypt Information

djtye:
Well got to make sure forums become SSL friendly ;)


coding-horror:
I think better to add it to the links at the bottom in "bonus stuff you can do after install", instead..

djtye:
I will also if its needed modify the discourse_docker to comment it out :)

coding-horror:
Sure, having it commented out there is smart, I can support that!

djtye:
I done that, I will add it to the bonus stuff. I had trouble finding out about Lets Encrypt for Discourse, need to make people aware :)

PR #4123, opened 2016-03-29
Added Lets Encrypt to bonus stuff

djtye:
Don't want to confuse users do we now ;)


coding-horror:
I don't really like it at the top, it should be with the other HTTPS stuff. I'll make the change the way I want it to be..

djtye:
Okay :)

coding-horror:
It's a good suggestion and I just added it, but I am picky about stuff ;)

djtye:
Well you are the Lead Dev aha, I don't blame you :P

PR #4124, opened 2016-03-30
Make MySQL port number configurable.

karies:
We needed to configure the phpBB MySQL port - here's the patch. Thanks for considering!


ZogStriP:
Thanks :+1:

PR #4126, opened 2016-03-30
Romanian Plurals

iamntz:
Relevant discussion: https://meta.discourse.org/t/romanian-pluralization-rules-are-wrong/41761


SamSaffron:
thanks !

PR #4130, opened 2016-04-01
Add `after-site-map-links` plugin outlet

NuckChorris:
For hummingbird.me I was hoping to shove some non-discourse-related links into the :hamburger: menu (mostly for mobile, where our full menu bar disappears), but there's a severe paucity of plugin outlets in there and the template is far too large to feasibly override.

This adds a plugin outlet in what I think is a fairly reasonable location for many purposes: below `site-map-links` and above the categories list.  If a plugin wants to add a section to the menu, this makes that possible (previously you could only really get it _inside_ the `site-map-links` list, unless you decorated the `hamburger-categories` widget to put something before it, even if you don't affect the categories list)

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/hamburger-menu.hbs
+++ b/app/assets/javascripts/discourse/templates/components/hamburger-menu.hbs
@@ -73,6 +73,8 @@
 
     {{plugin-outlet "site-map-links-last"}}
   {{/menu-links}}
+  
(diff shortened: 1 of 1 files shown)


tgxworld:
@NuckChorris The hamburger menu has gone through a refactor.

Can you try this? https://meta.discourse.org/t/a-tour-of-how-the-widget-virtual-dom-code-in-discourse-works/40347/33?u=tgxworld

PR #4131, opened 2016-04-01
Add no_deleted option to disqus import

sethherr:
Disqus preserves deleted comments, provide an option to skip importing them.


ZogStriP:
Thanks :+1:

PR #4133, opened 2016-04-02
Test EmbedHelper.embed_post_date

compwron:
Used in https://github.com/discourse/discourse/blob/166c56ee0d870e566168f867d3d9e000145dd157/app/views/embed/comments.html.erb#L13 and not tested directly elsewhere. 

I believe that these tests will not be flaky because the duration to present is controlled by Timecop.

### DIFF
--- a/spec/helpers/embed_helper_spec.rb
+++ b/spec/helpers/embed_helper_spec.rb
@@ -1,0 +1,41 @@
+require 'rails_helper'
+
+describe EmbedHelper do
+  describe "embed_post_date" do
+    subject { embed_post_date(dt) }
+
+    context "with a date that is right now" do
+      let(:dt) { Time.now }
+      it "shows less than 1 minute because this test runs quickly" do
+        expect(subject).to eq "< 1m"
+      end
+    end
+
+    context "with a date that is right now" do
+      let(:dt) { 5.minutes.ago }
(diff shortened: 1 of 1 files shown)


spec/helpers/embed_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      it "shows less than 1 minute because this test runs quickly" do
+        Timecop.freeze(Time.new(2016, 4, 1)) do
+          expect(subject).to eq "5m"
+        end
+      end
+    end
+
+    context "with a date from more than a month ago" do
+      let(:dt) { Time.new(2016, 3, 1) }
+      it "" do
```
ZogStriP: why no description here?

spec/helpers/embed_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+describe EmbedHelper do
+  describe "embed_post_date" do
+    subject { embed_post_date(dt) }
```
tgxworld: `dt` is really ambiguous. I can't tell from one look what it means/represent.

tgxworld:
Ping @compwron :grin:

SamSaffron:
I am closing this for now, in general I prefer to avoid PRs that just add a test.

PR #4137, opened 2016-04-03
FEATURE: Add warning input to flag dialog when notifying a user

cpradio:
Added isWarning property
pass is_warning along to post_action
Added is_warning to possible arguments to receive from post_actions route
Only show warning checkbox for staff
Only permit the is_warning argument if the user is staff

### DIFF
--- a/app/assets/javascripts/discourse/controllers/flag-action-type.js.es6
+++ b/app/assets/javascripts/discourse/controllers/flag-action-type.js.es6
@@ -5,6 +5,7 @@ export default Ember.Controller.extend({
   needs: ['flag'],
 
   message: Em.computed.alias('controllers.flag.message'),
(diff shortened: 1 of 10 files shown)


cpradio:
Adding screenshot
image

coding-horror:
I think it might be better to have a red danger button addedat the bottom for staff like we do for spam flags "delete spammer" except it says official warning with the alert glyph

cpradio:
Oh, so instead of a checkbox do a second button? I could look into doing that.

cpradio:
New screenshot
image

coding-horror:
Looks good to me!

coding-horror:
Can we review and merge this @SamSaffron I think it's a good change

SamSaffron:
Sure, looks good and PR is very tidy, nice work @cpradio !

PR #4141, opened 2016-04-04
correcting errors in translation sources

frictionel:


config/locales/server.en.yml
```diff
@@ -2378,7 +2378,7 @@ en:
   email_log:
     no_user: "Can't find user with id %{user_id}"
     anonymous_user: "User is anonymous"
-    suspended_not_pm: "User is suspended, not a message"
+    suspended_not_pm: "User is suspended, not a PM"
```
fantasticfears: PM seems not used anymore. They are however appeared as `message` in any place. I didn't confirm that

frictionel: yes, seems you are right. it's handled inconsistently though. sometimes PM are referred to as 'private message' and sometimes just 'message'. I updated my commit (I think, first time I do this :)) by replacing a PM with message in another place instead of replacing the above message with PM.

ZogStriP:
Thanks 👍

PR #4143, opened 2016-04-05
altering topic-map widget's like count to show total topic like count

stevenpslade:
Fixing bug where the `topic-map-summary` was only displaying the first post's like count rather than the entire topic's like count.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/topic-map.js.es6
+++ b/app/assets/javascripts/discourse/widgets/topic-map.js.es6
@@ -81,7 +81,7 @@ createWidget('topic-map-summary', {
 
     if (attrs.topicLikeCount) {
       contents.push(h('li.secondary', [
-        numberNode(attrs.likeCount),
+        numberNode(attrs.topicLikeCount),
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/widgets/topic-map.js.es6
```diff
@@ -81,7 +81,7 @@ createWidget('topic-map-summary', {
 
     if (attrs.topicLikeCount) {
       contents.push(h('li.secondary', [
-        numberNode(attrs.likeCount),
+        numberNode(attrs.topicLikeCount),
         h('h4', I18n.t('likes_lowercase', { count: attrs.likeCount }))
```
ZogStriP: Shouldn't this also be `attrs.topicLikeCount`?

stevenpslade:
both `likeCount`s have now been changed to `topicLikeCount`

ZogStriP:
cc @eviltrout

eviltrout:
Excellent thanks!

PR #4146, opened 2016-04-06
PERF: Rendering crawler's template is expensive.

tgxworld:
@eviltrout This was what I was trying to understand just now and thank you for explaining it to me. 

Going directly to `/about` makes two requests, one to `/about.html` and `/about.json`. This particular problem only affects routes which are present in Rails and Ember.

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -9,8 +9,6 @@ class AboutController < ApplicationController
 
     respond_to do |format|
       format.html do
-        # @list = list
-        # store_preloaded(list.preload_key, MultiJson.dump(TopicListSerializer.new(list, scope: guardian)))
         render :index
       end
       format.json do
--- a/app/views/about/index.html.erb
+++ b/app/views/about/index.html.erb
@@ -1,3 +1,5 @@
+<% if crawler_layout? %>
+
 <% content_for :title do %><%=t "about" %><% end %>
 
 <section itemscope itemtype="https://schema.org/AboutPage">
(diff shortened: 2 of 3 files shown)


app/views/about/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<%- if @use_crawler_content %>
```
ZogStriP: Where is this variable used/set?

app/controllers/about_controller.rb
```diff
@@ -9,8 +9,6 @@ def index
 
     respond_to do |format|
       format.html do
-        # @list = list
-        # store_preloaded(list.preload_key, MultiJson.dump(TopicListSerializer.new(list, scope: guardian)))
```
eviltrout: Weird... it was preloading a topic? On the about route??

app/views/about/index.html.erb
tgxworld: Here https://github.com/discourse/discourse/blob/a5d8dfb07e5362f688344f576160066fd0bfaceb/app/controllers/application_controller.rb#L55 :)

app/controllers/about_controller.rb
tgxworld: Hmm not sure why it was commented out. @nlalonde Do you happen to recall why? I checked the commit for this and I didn't find anything.

app/views/about/index.html.erb
ZogStriP: It's `@use_crawler_layout` instead of `@use_crawler_content` ;)

tgxworld: ahhhhh good catch :)

tgxworld:
Updated to use the cache in the template

PR #4147, opened 2016-04-06
FIX: Consider using default top setting when user returns or enough data exists for Top Page Results

cpradio:
Return default top setting as part of best_periods_for to see if it can be used

### DIFF
--- a/spec/controllers/list_controller_spec.rb
+++ b/spec/controllers/list_controller_spec.rb
@@ -224,26 +224,60 @@ describe ListController do
   describe "best_periods_for" do
 
     it "returns yearly for more than 180 days" do
+      SiteSetting.stubs(:top_page_default_timeframe).returns("all")
       expect(ListController.best_periods_for(nil)).to eq([:yearly])
       expect(ListController.best_periods_for(180.days.ago)).to eq([:yearly])
     end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/list_controller_spec.rb
```diff
@@ -224,28 +224,62 @@
   describe "best_periods_for" do
 
     it "returns yearly for more than 180 days" do
+      SiteSetting.stubs(:top_page_default_timeframe).returns("all")
```
eviltrout: You don't need to stub SiteSettings like this. Simply do `SiteSetting.top_page_default_timeframe = 'all` and it'll reset after the test to the default value.

ZogStriP: Right I didn't catch it. CCing @cpradio since he wrote the PR ;)

cpradio: Sweet, I'll send another PR doing it the other way :)

eviltrout: Oops didn't see that it was merged :) Still should be fixed, thanks!

cpradio: PR sent :)

PR #4148, opened 2016-04-07
FIX: Don't send rejection mailer to bounced emails.

tgxworld:


spec/components/email/receiver_spec.rb
```diff
@@ -8,10 +8,6 @@
     SiteSetting.reply_by_email_address = "reply+%{reply_key}@bar.com"
   end
 
-  def email(email_name)
```
ZogStriP: pretty sure that method is used

lib/email/receiver.rb
```diff
@@ -422,6 +426,20 @@ def can_invite?(topic, user)
       !topic.topic_allowed_groups.where("group_id IN (SELECT group_id FROM group_users WHERE user_id = ?)", user.id).exists?
     end
 
+    private
+
+    def check_reply_to_auto_generated_header
```
ZogStriP: why can't you just check `@mail.header["X-Discourse-Reply-to-Auto-Generated"] == "marked"`?

tgxworld: Because the mail's header will not contain our "poison". Instead our email reply is embedded in the body and we have to extract that out.

ZogStriP: IC

lib/email/receiver.rb
```diff
@@ -422,6 +426,20 @@ def can_invite?(topic, user)
...
+      headers = Mail::Header.new(@mail.body.to_s.gsub("\r\n\r\n", "\r\n")).to_a
```
tgxworld: We grab the headers like [how `mail` gem grabs](https://github.com/mikel/mail/blob/master/lib/mail/part.rb#L118) the delivery status report

tgxworld:
@zogstrip please review :) thank you!!

PR #4150, opened 2016-04-07
Remove `has_many :views` from User

gdpelican:
Not sure what 'views' are, but they don't seem to be a model (it's actually a column on the DB), and they break `as_json` on a User, so thinking it can go?

https://meta.discourse.org/t/uninitialized-constant-user-view/25177/10


ZogStriP:
Thanks! :+1:

PR #4153, opened 2016-04-08
FIX: Show badge count of 1 on the badges page

gschlager:
The badges page didn't show a count on the badge card if the badge was granted only once.
This changes it so that on user profiles only badge counts > 1 are shown and on the badges page counts > 0 are shown.


eviltrout:
Seems good thanks!

PR #4157, opened 2016-04-09
Feature: Show location and website on user card.

mcwumbly:
Location and Website are shown on user card when not blank.

https://meta.discourse.org/t/custom-user-fields-on-usercard/22662/12?u=mcwumbly

user-card-location-website

### DIFF
--- a/app/assets/javascripts/discourse/controllers/user-card.js.es6
+++ b/app/assets/javascripts/discourse/controllers/user-card.js.es6
@@ -26,6 +26,7 @@ export default Ember.Controller.extend({
   showBadges: setting('enable_badges'),
   showMoreBadges: Em.computed.gt('moreBadgesCount', 0),
   showDelete: Em.computed.and("viewingAdmin", "showName", "user.canBeDeleted"),
+  linkWebsite: Em.computed.not('user.isBasic'),
 
   @computed('model.user_fields.@each.value')
   publicUserFields() {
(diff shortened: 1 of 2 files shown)


mcwumbly:
I realize I didn't add any tests for this... Not sure if that's something you'd normally have coverage for on this project.  If so, where would it live?

ZogStriP:
Would be an integration test, or as we call them in JavaScript-land: an acceptance test ;)

mcwumbly:
Should I add tests here in that case?

```
test/javascripts/acceptance/user-card-test.js.es6
```

Here's the full body of that file currently:

```
import { acceptance } from "helpers/qunit-helpers";
acceptance("User Card");

test("card", () => {
  visit('/');

  ok(invisible('#user-card'), 'user card is invisible by default');
  click('a[data-user-card=eviltrout]:first');

  andThen(() => {
    ok(visible('#user-card'), 'card should appear');
    
  });

});
```

ZogStriP:
Yup, we're way behind in our acceptance tests.

mcwumbly:
Serious question - does the team see value in adding additional test coverage there? (Just wondering if perhaps being "behind" on acceptance tests in actually more intentional).

Are there any areas of the Ember app that are emblematic of the kind of coverage or testing patterns that the team is aiming for and wants to maintain?

(Should we move this conversation to meta?).

eviltrout:
It's definitely not intentional to be behind on acceptance tests. Ember was really bad at testing for a long time, but the story has been excellent for the last couple of years. Generally when I have to work on an untested part of the app I add a bunch of tests. For example the vdom/widget rewrite added over 100 tests to a part of the app that was missing them.

The trick is to add useful tests, and not just "the dom has element a with child b and class c" which makes it hard to change the UX.

But yes, I would absolutely love more tests overall.

mcwumbly:
OK, cool.

I will take a stab at adding additional tests for the user card next in
that case.  I can update this PR or submit a new one if you're OK pulling
this in in the mean time.

eviltrout:
Seems good to me! Thanks

PR #4158, opened 2016-04-09
UX: Show website path in website name for all domains

mcwumbly:
discussion: https://meta.discourse.org/t/show-full-website-path-on-user-profile-regardless-of-domain/42285?u=mcwumbly

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -144,19 +144,9 @@ class UserSerializer < BasicUserSerializer
   end
 
   def website_name
-    website_host = URI(website.to_s).host rescue nil
-    discourse_host = Discourse.current_hostname
-    return if website_host.nil?
-    if website_host == discourse_host
-      # example.com == example.com
-      website_host + URI(website.to_s).path
-    elsif (website_host.split('.').length == discourse_host.split('.').length) && discourse_host.split('.').length > 2
-      # www.example.com == forum.example.com
-      website_host.split('.')[1..-1].join('.') == discourse_host.split('.')[1..-1].join('.') ? website_host + URI(website.to_s).path : website_host
-    else
-      # example.com == forum.example.com
-      discourse_host.ends_with?("." << website_host) ? website_host + URI(website.to_s).path : website_host
-    end
+    uri = URI(website.to_s) rescue nil
+    return if uri.nil? || uri.host.nil?
+    uri.host + uri.path
   end
 
   def include_website_name
--- a/spec/serializers/user_serializer_spec.rb
+++ b/spec/serializers/user_serializer_spec.rb
@@ -96,20 +96,35 @@ describe UserSerializer do
         expect(json[:website]).to eq 'http://example.com/user'
       end
 
-      context "has a website name" do
-        it "returns website host name when instance domain is not same as website domain" do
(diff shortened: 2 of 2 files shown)


mcwumbly:
Updated tests:

- Removed redundant tests for domain logic that has been removed
- Added tests to enforce spec regarding truncating query params per discussion on meta

app/serializers/user_serializer.rb
```diff
@@ -145,18 +145,8 @@ def website
...
-      # example.com == example.com
-      website_host + URI(website.to_s).path
-    elsif (website_host.split('.').length == discourse_host.split('.').length) && discourse_host.split('.').length > 2
-      # www.example.com == forum.example.com
-      website_host.split('.')[1..-1].join('.') == discourse_host.split('.')[1..-1].join('.') ? website_host + URI(website.to_s).path : website_host
-    else
-      # example.com == forum.example.com
-      discourse_host.ends_with?("." << website_host) ? website_host + URI(website.to_s).path : website_host
-    end
+    website_host + URI(website.to_s).path
```
SamSaffron: no rescue here for URI(website.to_s) this means that bad URI somewhere could just explode the serializer

mcwumbly: ok, let me take a look at that

SamSaffron: probably worth adding a spec for bad URI

mcwumbly: `URI(website.to_s)` is rescued on line 147... I'll add a spec to keep it sane

mcwumbly:
If preferred, method can be rewritten with one call to URI(website.to_s):

```
  def website_name
    uri = URI(website.to_s) rescue nil
    return if uri.nil? || uri.host.nil?
    uri.host + uri.path
  end
```

app/serializers/user_serializer.rb
SamSaffron: I see, no point double parsing though ... should rewrite it

mcwumbly: OK, done.

mcwumbly:
failure looks unrelated to me:
https://travis-ci.org/discourse/discourse/jobs/122143726#L5879

```
Failures:
  1) Scheduler::Manager per host jobs correctly schedules on multiple hosts
     Failure/Error: expect(Testing::PerHostJob.runs).to eq(3)
       
       expected: 3
            got: 2
       
       (compared using ==)
     # ./spec/components/scheduler/manager_spec.rb:98:in `block (3 levels) in <top (required)>'
Finished in 9 minutes 28 seconds (files took 8.51 seconds to load)
5069 examples, 1 failure, 9 pending
Failed examples:
rspec ./spec/components/scheduler/manager_spec.rb:75 # Scheduler::Manager per host jobs correctly schedules on multiple hosts

```

SamSaffron:
There seems to be some logic regarding stripping of first subdomain (I think intended)

If my domain is https://www.samsaffron.com, I think it is reasonable to display `samsaffron.com` so stripping first www. I think is correct. 

Going to merge this in for now, old logic seems very very strange. 

Also need to test what happens to UI when your domain name is:

`https://this-is-very-very-very-very.long.com/with-a-super-long-path`

PR #4160, opened 2016-04-09
FEATURE: Allow disabling link tracking with .no-track-link class

NuckChorris:
I was running into some trouble with link tracking breaking a signup CTA link I'd insert into a onebox at runtime, because the `.onebox-results` ancestor doesn't trigger a return until much later — after the link has been rewritten. This allows disabling link tracking on any link just by adding a `no-track-link` class.  The name is kinda weird but I wanted to match the existing `track-link` class which enables link tracking on otherwise-disabled links.


eviltrout:
Thanks!

PR #4161, opened 2016-04-10
Make Vagrantfile more resilient

hellekin:
This includes two changes in Vagrantfile:

1. Disable SSH proxy

When you're using SSH configuration extensively, you may use `ProxyCommand`, in which case `vagrant ssh` may fail.  As there's no obvious reason to proxy access to a local development machine, disable it.

2. NFSv4 support

By default `vagrant` uses NFSv3, which may be incompatible with your setup.  This adds a comment to hint about how to make it work with NFSv4.

No other functionality is affected.


SamSaffron:
I think we have had a vagrant revamp since, can you check that this issue is taken care of ?

PR #4162, opened 2016-04-10
FEATURE: email keyboard for username on mobile login

frictionel:


coding-horror:
Can you elaborate on what this does? Was it tested on iOS and Android?

PR #4164, opened 2016-04-11
Add spec for website name when url has subdomain

mcwumbly:
Follow up re: the point here about subdomains:
https://github.com/discourse/discourse/pull/4158#issuecomment-208163236

Adding spec for current behavior when URL has a subdomain.

Perhaps it was intentional to strip this before, but that's not what happens right now.  Also, there's a case to be made that it should be included (It's somewhat typical for people that have their website on a hosted service that gives out subdomains for free... eg `mysite.github.io`).

### DIFF
--- a/spec/serializers/user_serializer_spec.rb
+++ b/spec/serializers/user_serializer_spec.rb
@@ -88,16 +88,32 @@ describe UserSerializer do
     end
 
     context "with filled out website" do
-      before do
-        user.user_profile.website = 'http://example.com/user'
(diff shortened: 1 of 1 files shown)


coding-horror:
I think www. should be suppressed, as in `https://www.twitter.com/mcwumbly` should show up as `twitter.com/mcwumbly`

mcwumbly:
Sure, that makes sense.  Show subdomain, bit strip www works for me.

PR #4167, opened 2016-04-11
Update sprockets.

tgxworld:


config/application.rb
```diff
@@ -144,6 +144,7 @@ def config.database_configuration
 
     # Our templates shouldn't start with 'discourse/templates'
     config.handlebars.templates_root = 'discourse/templates'
+    config.handlebars.raw_template_namespace = "Ember.TEMPLATES"
```
tgxworld: In the latest version of `ember-handlebars-template`, the namespace was changed to default to `JST`. However, that is breaking our Ember resolver which uses `Ember.TEMPLATES`

Related commit in `ember-handlebars-template`: https://github.com/tricknotes/ember-handlebars-template/commit/b8cc2e75b9d5a45159d46540142ae50eceeec74d

lib/freedom_patches/better_handlebars_errors.rb
```diff
@@ -1,13 +1,13 @@
 module Ember
   module Handlebars
-    class Template < Tilt::Template
```
tgxworld: Tile dependency has been dropped in https://github.com/tricknotes/ember-handlebars-template/commit/66ec5ef190f72b4154905cf346e01ae389a95444

lib/freedom_patches/ember_compat_handlebars.rb
```diff
@@ -1,32 +1,47 @@
 # barber patches to re-route raw compilation via ember compat handlebars
-#
 
-module Barber
-  class EmberCompatPrecompiler < Barber::Precompiler
```
tgxworld: This is really annoying, we can no longer declare our own precompiler because `ember-handlebars-template` [assumes `Barber::Precompiler`](https://github.com/tricknotes/ember-handlebars-template/blob/263a74be2a7862455db34f0e9fdd2da1fbee1d91/lib/ember/handlebars/template.rb#L149) is the precompiler to be used.

lib/freedom_patches/ember_compat_handlebars.rb
```diff
@@ -1,32 +1,47 @@
...
-  def compile_handlebars(string)
-    "Discourse.EmberCompatHandlebars.compile(#{indent(string).inspect});"
+  include Discourse::Ember::Handlebars::Helper
+
+  # FIXME: Previously, ember-handlebars-templates uses the logical path which incorrectly
+  # returned paths with the `.raw` extension and our code is depending on the `.raw`
+  # to find the right template to use.
+  def actual_name(input)
+    actual_name = input[:name]
+    input[:filename].include?('.raw') ? "#{actual_name}.raw" : actual_name
```
tgxworld: Is there a need for our Ember template resolver to look for templates with the `.raw` extension? We will have to fix the way we lookup raw templates in Ember in order to remove this nasty monkey patch.

lib/sass/discourse_sass_importer.rb
```diff
@@ -2,156 +2,118 @@
...
+        ::Sass::Engine.for_file(full_filename, options)
       end
-    end
+  end
 
-    def engine_from_path(name, dir, options)
-      full_filename, _ = Sass::Util.destructure(find_real_file(dir, name, options))
-      return unless full_filename && File.readable?(full_filename)
+  include Sass
+  include ::Sass::Rails::SassImporter::Globbing
```
tgxworld: \o/ `sass-rails` made their code modular allowing us to do this.

test/javascripts/test_helper.js
```diff
@@ -1,7 +1,7 @@
 /*global document, sinon, QUnit, Logster */
 
 //= require env
-//= require ../../app/assets/javascripts/preload_store
```
tgxworld: This is really odd but base on what I saw in the source code, `require` requires the path to have the same root path as the current file.

tgxworld:
@SamSaffron  @eviltrout Please review before I deploy this on to meta. Thank you!

lib/freedom_patches/ember_compat_handlebars.rb
eviltrout: Yes those templates are ones that are passed to the raw handlebars renderer (ie, not Ember) for performance reasons. I'd be okay with a different naming convention if that makes it easier. For example, if the template name ends with `-raw`

eviltrout:
Wow what a ton of work! Looks good to me although I doubt I'd be able to spot an error. I trust you that if it's all working we can merge this in.

tgxworld:
@eviltrout I'm not sure if I'll be able to spot an error myself xD Thanks for looking through it :) As long as nothing looks awfully wrong then I'll have it deployed on Meta tomorrow :)

PR #4170, opened 2016-04-13
Use consistent link color for website name

mcwumbly:
before

screen shot 2016-04-12 at 11 11 39 pm

after

screen shot 2016-04-12 at 11 11 25 pm


mcwumbly:
I debated whether it needs the underline styling and decided it doesn't given the globe icon, but feel free to add it if you disagree.

PR #4171, opened 2016-04-14
FEATURE: User admin - filter groups as you type

DeanMarkTaylor:
See before accepting: https://meta.discourse.org/t/groups-autocomplete-in-admin-broken/42552/3?u=deanmarktaylor


eviltrout:
Looks good thanks!

PR #4172, opened 2016-04-15
[Do not merge] Rails master bench

tgxworld:


SamSaffron:
I am closing this as too scary for now, and we are going need to revisit this when rails 5 is out ... (keep the branch around for when we do the move)

connorshea:
@SamSaffron any updates on this, or discussions I can read through?

tgxworld:
@connorshea We basically need to upgrade to Rails 5 and run our [benchmark script](https://github.com/discourse/discourse/blob/master/script/bench.rb) to make sure we don't regress on performance.

PR #4174, opened 2016-04-15
Add 'alt' tag to logo img in digest email

NickIvanter:
The rationale for this is that the logo, as any other image, is not loaded by default by many email clients. In absence of an 'alt' tag, it shows an empty space in the email. Having the site title in there seems better.


coding-horror:
Maybe -- refer to https://meta.discourse.org/t/ui-breaks-when-letter-avatar-images-dont-load/36674 for a nuanced discussion of this sort of thing

PR #4175, opened 2016-04-16
Allow any username character in user search

danielhers:
See [discussion on meta](https://meta.discourse.org/t/mention-full-name-autocomplete-doesnt-work-with-international-characters/29819/28) - this allows matching non-ASCII names too.


coding-horror:
This seems awfully incorrect to me, since `#` will now match as a "non space character".

The correct thing to do is invert the allowed username characters. Even `\W` would be better, I think.

danielhers:
Hi Jeff, thanks for the feedback!
`\W` is equivalent to current regex, except for the `.`. It doesn't help with matching non-Latin characters.
It would be easy to break on space OR `#`, but are there other characters that should not match? Is there a list somewhere?

coding-horror:
ok let's give this a try

PR #4176, opened 2016-04-17
Remove gunk from post stream serializer mixin

gdpelican:
I couldn't help but notice this isn't doing anything helpful 🌈


eviltrout:
Thanks, looks useless.

PR #4179, opened 2016-04-18
FIX: duplicate subfolder prefix in links in digest

NickIvanter:
In subfolder installs, the subfolder prefix was getting duplicated in topic and category links in the digest. E.g. http://example.com/forum/forum/t/...


xfalcox:
I can confirm this bugs affects me too. Discourse handle the links just fine, but they are incorrect indeed.

eviltrout:
Looks good to me, thanks!

PR #4182, opened 2016-04-19
FIX: make sure the best post is not the worst

NickIvanter:
By default Postgres returns NULLs first when sorting in descending order. As a result, best_post() would often actually return the "worst" post of the topic. And it would then be included in digest emails in "popular posts".

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -245,7 +245,7 @@ class Topic < ActiveRecord::Base
   end
 
   def best_post
-    posts.order('score desc').limit(1).first
+    posts.order('score desc nulls last').limit(1).first
   end
 
   def has_flags?


ZogStriP:
That's weird, the `score` column should not be `NULL`. @NickIvanter is sidekiq running properly on your Discourse?

NickIvanter:
Sidekiq is running well. What I'm seeing is that the score is null for some time after a post is published.

Many of my users have their digests configured for 30 minute intervals and what I see happening is this:

A new topic is started. Someone responds quickly with a terse message. The digest happens to be sent after the first response is published and this response, instead of the starting post, is the one that gets included in the digest.

I've investigated this and found that the score of the response is zero at the time the digest is sent out. Then, as more people interact with the topic, the score becomes non-zero, but the damage is already done. Meaning that the wrong post got sent out in the digest.

ZogStriP:
Yeah, the `score` is only computed every week.

PR #4186, opened 2016-04-20
Fix: Suspensions and Blocks do not give enough details in logs

shaktikatare1989:
pull request for the issue https://meta.discourse.org/t/suspensions-and-blocks-do-not-give-enough-details-in-logs/39554


techAPJ:
Thanks 👍

PR #4188, opened 2016-04-20
Add memconfig to INSTALL-cloud.md

pfaffman:
`./launcher memconfig app` will copy standalone.yml to app.yml and configure memory-dependent settings. 

`mkdir /var/discourse` not needed since `git` will create it in the next line.


coding-horror:
wait because I am making other changes here

PR #4189, opened 2016-04-21
Minor spelling mistake and language consistency

DeanMarkTaylor:
"Return" instead of "Retrun".
"unactivated" changed to "deactivated".
Mostly "cannot" is currently used, changed existing "can not" occurrences.
"there was no attachments" to "there were no attachments".


tgxworld:
Thank you @DeanMarkTaylor!

PR #4192, opened 2016-04-24
Respect external link preference on Profile page

cpradio:
which includes All, Replies, Likes Given, and Bookmarks.

Related discussion
https://meta.discourse.org/t/external-links-in-history-don-t-follow-open-all-external-links-in-a-new-tab-setting/42995


cpradio:
Also before accepting this PR, it would be good to get advice at
https://meta.discourse.org/t/external-links-in-history-don-t-follow-open-all-external-links-in-a-new-tab-setting/42995/10?u=cpradio

I think there is a better way to approach this, but I want to make sure you guys agree before I pursue it.

PR #4193, opened 2016-04-24
FEATURE: new bootstrap mode settings for brand new Discourse community

techAPJ:
This PR adds new "Bootstrap Mode" _hidden_ site setting which is:

- enabled _only_ after the first admin logs in

- when this mode is on, there is a mandatory only visible to staff banner up that says:

> To make launching your new site easier, you are in bootstrap mode. All new users will be granted trust level 1 and have daily email digest updates enabled. This will be automatically turned off in {x} days.

- it makes TL1 the default for all new users

- it makes daily digest the default for all new users

- it automatically turns off after ONE WEEK (7 days) from first admin login

- when turned off, it reverses those changes (back to TL0 as new user default, back to weekly digest as new user default)

### DIFF
--- a/app/assets/javascripts/discourse/components/global-notice.js.es6
+++ b/app/assets/javascripts/discourse/components/global-notice.js.es6
@@ -17,6 +17,14 @@ export default Ember.Component.extend(StringBuffer, {
       notices.push([I18n.t("emails_are_disabled"), 'alert-emails-disabled']);
     }
 
+    if (this.currentUser && this.currentUser.get('staff') && this.siteSettings.bootstrap_mode_enabled) {
+      if (this.siteSettings.bootstrap_mode_min_users > 0) {
+        notices.push([I18n.t("bootstrap_mode_enabled", {min_users: this.siteSettings.bootstrap_mode_min_users}), 'alert-bootstrap-mode']);
+      } else {
+        notices.push([I18n.t("bootstrap_mode_disabled"), 'alert-bootstrap-mode']);
+      }
+    }
+
     if (!_.isEmpty(this.siteSettings.global_notice)) {
       notices.push([this.siteSettings.global_notice, 'alert-global-notice']);
     }
--- a/app/jobs/regular/enable_bootstrap_mode.rb
+++ b/app/jobs/regular/enable_bootstrap_mode.rb
@@ -10,15 +10,9 @@ module Jobs
       return unless user.is_singular_admin?
 
       # let's enable bootstrap mode settings
-      update_site_setting('default_trust_level', TrustLevel[1])
-      update_site_setting('default_email_digest_frequency', 1440)
-      update_site_setting('bootstrap_mode_enabled', true)
-    end
-
-    def update_site_setting(id, value)
-      prev_value = SiteSetting.send(id)
-      SiteSetting.set(id, value)
-      StaffActionLogger.new(Discourse.system_user).log_site_setting_change(id, prev_value, value) if SiteSetting.has_setting?(id)
+      SiteSetting.set_and_log('default_trust_level', TrustLevel[1])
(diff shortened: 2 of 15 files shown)


coding-horror:
Very nice! Let's let @SamSaffron have a look as he asked for the initial feature. We can make this more expansive as we go, but getting something safe in to start is a good beginning, I think.

techAPJ:
Updated the PR, here are major changes:

- Admin can now disable bootstrap mode by setting "bootstrap_mode_min_users" to 0

- Admin can control when to turn off bootstrap mode by tuning "bootstrap_mode_min_users" setting

- Added a new scheduled job that check for total number of users _daily_ and disable bootstrap mode if the total users exceeds "bootstrap_mode_min_users" setting.

- Updated the banner message to:

> To make launching your new site easier, you are in bootstrap mode. All new users will be granted trust level 1 and have daily email digest updates enabled. This will be automatically turned off when total user count exceeds {x} users.

- "Disable after n days" concept is replaced by "Disable after n users"

- Logging changes in site settings

- Added more tests

app/assets/javascripts/discourse/components/global-notice.js.es6
```diff
@@ -17,6 +17,14 @@ export default Ember.Component.extend(StringBuffer, {
       notices.push([I18n.t("emails_are_disabled"), 'alert-emails-disabled']);
     }
 
+    if (Discourse.User.currentProp('staff') && this.siteSettings.bootstrap_mode_enabled) {
```
eviltrout: We're still trying to deprecate constants in Javascript. You should use `this.currentUser && this.currentUser.get('staff')` instead.

app/jobs/regular/enable_bootstrap_mode.rb
```diff
@@ -0,0 +1,24 @@
+module Jobs
+  class EnableBootstrapMode < Jobs::Base
+    sidekiq_options queue: 'critical'
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:user_id) unless args[:user_id].present?
+      return if SiteSetting.bootstrap_mode_enabled
+
+      @current_user = User.find_by(id: args[:user_id])
```
eviltrout: I don't think `@current_user` should be an instance variable. Also it's a little confusing that it uses the same name as the function `current_user` elsewhere.  I'd suggest just calling it `user` instead.

techAPJ:
@eviltrout: Updated the PR as per your comments.

app/jobs/scheduled/disable_bootstrap_mode.rb
```diff
@@ -0,0 +1,22 @@
...
+      total_users = User.where.not(id: Discourse::SYSTEM_USER_ID).count
+
+      if SiteSetting.bootstrap_mode_min_users == 0 || total_users > SiteSetting.bootstrap_mode_min_users
+        update_site_setting('default_trust_level', TrustLevel[0])
+        update_site_setting('default_email_digest_frequency', 10080)
+        update_site_setting('bootstrap_mode_enabled', false)
+      end
+    end
+
+    def update_site_setting(id, value)
```
eviltrout: This `update_site_setting` code is the same in two places. I think a class method like `SiteSetting.change(name, val)` might be better and you wouldn't be repeating yourself.

eviltrout:
Looks good. Ping me when you've fixed all the minor issues and it should be good to merge.

techAPJ:
@eviltrout Okay I created a new method `set_and_log` via https://github.com/discourse/discourse/pull/4193/commits/762cb28c3bec076dc2d7eb78aa435d5976030757

PR should be good for final review now.

eviltrout:
Looks good, 👍

PR #4194, opened 2016-04-24
Stop using OFFSET in DB queries of phpBB3 importer

gschlager:
http://use-the-index-luke.com/no-offset


riking:
I think it's worth keeping the ASC in the sort ordes.

gschlager:
I don't see why we need `ASC`. It's the default sort order anyway and all it does is make me think. IMHO it's a lot easier to read the queries without the `ASC`.

PR #4195, opened 2016-04-24
FIX: header offset 0 not respected in google android

featheredtoast:
see https://meta.discourse.org/t/d-header-0px-not-respected-google-chrome-android-6p/43131

### DIFF
--- a/app/assets/stylesheets/common/base/header.scss
+++ b/app/assets/stylesheets/common/base/header.scss
@@ -1,7 +1,7 @@
 .d-header {
     width: 100%;
     position: absolute;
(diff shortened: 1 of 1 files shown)


ZogStriP:
Any idea where that 1px comes from?

featheredtoast:
Not certain. The closest related issue I have found is [this one](http://stackoverflow.com/questions/16574776/css-border-style-double-1px-gap-chrome).

coding-horror:
this feels like a really weird fix. Let's keep this in the meta topic for now and see what happens for future versions of Chrome?

PR #4196, opened 2016-04-24
FEATURE: Make pending unapproved posts visible to their authors

NickIvanter:
As discussed here: https://meta.discourse.org/t/how-to-make-the-queued-posts-visible-to-their-authors/39153


coding-horror:
Thanks I do think this is better if the post is clearly marked as pending in some way. It is a lot of code though can you review @eviltrout after header is in?

eviltrout:
Wow 47 files changed? This is going to take some time to review!

NickIvanter:
It was a significant effort on our part. Thanks to my colleague @nnrd . We deployed it in production over a month ago and have been using since.

script/import_scripts/base.rb
```diff
@@ -528,6 +528,52 @@ def create_bookmarks(results, opts={})
     [created, skipped]
   end
 
+  # Iterate through a list of "like" records to be imported.
+  # Takes a collection, and yields to the block for each element.
+  # Block should return a hash with the attributes for the "like".
+  # Required fields are :user_id and :post_id, where both ids are
+  # the values in the original datasource.
+  def create_likes(results, opts={})
```
xfalcox: Is this related to the feature?

NickIvanter: No, this must have gotten onto this branch by mistake. Good catch.

eviltrout:
I responded on discourse meta because I think it's a better place for discussing this for now:

https://meta.discourse.org/t/how-to-make-the-queued-posts-visible-to-their-authors/39153/16?u=eviltrout

coding-horror:
I like this change in theory but I think we may need to approach it a different way down the line?

PR #4200, opened 2016-04-27
FEATURE: Enhance click-track to know about links on the Profile page …

cpradio:
…so they can open in an external window.


cpradio:
Something looks goofy in the compares... not sure why, but it is showing changes that I didn't make, especially when I compare it to the latest click-tract.js.es6 file in the Discourse repo. Such as the adding of no-track-link.

I've got a new branch that isn't showing that oddity, should I use that for the PR instead (and close this one)?

ping @ZogStriP

cpradio:
Sorry, this PR is goofy and I don't know what I did wrong, my other one looks right.
https://github.com/discourse/discourse/pull/4201

PR #4202, opened 2016-04-29
FIX: Ensure that readonly key has an expiry.

tgxworld:


tgxworld:
Moving this to another PR

PR #4203, opened 2016-04-29
FEATURE: Extend clicking tracking to the user profile page.

cpradio:
Added qunit tests
Fixed click-track-test so it uses the correct data attribute for topic_id in its fixture


eviltrout:
Thanks for all the hard work @cpradio !

PR #4204, opened 2016-04-29
FIX: Don't diplay character reference in HTML diffs

xfix:
Before this change, HTML escaping was done before splitting text into tokens, so token splitter saw literals like "&#39;", and split them as it was normal text into parts into ["&", "#", "39", ";"]. This caused diff to display character references, as those tokens used separate HTML tags to display their insertion/deletion status.

spectacle zb8627

### DIFF
--- a/lib/discourse_diff.rb
+++ b/lib/discourse_diff.rb
@@ -261,7 +261,7 @@ class DiscourseDiff
     end
 
     def characters(string)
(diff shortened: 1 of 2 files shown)


xfix:
Thinking about it, that ``flatten`` is not needed, because parens could be removed from a regex. Oh well.

SamSaffron:
This looks good to me @ZogStriP  ?

ZogStriP:
Thanks 👍

PR #4205, opened 2016-05-02
Allow an (optional) post-creation time to be submitted.

skx:
This should allow a new post to be created with an initial date/time specified by the caller, which will be useful for people writing importers.

Some brief discussion of this omission is available here:

* https://meta.discourse.org/t/set-the-date-of-a-new-topic-via-the-api/43308/2


app/controllers/posts_controller.rb
```diff
@@ -544,7 +544,8 @@ def create_params
       :reply_to_post_number,
       :auto_track,
       :typing_duration_msecs,
-      :composer_open_duration_msecs
+      :composer_open_duration_msecs,
+      :created_at
```
oblakeerickson: thanks for making this commit! I think we want to make sure this is only permitted if the api is being used. Perhaps we can leave the permitted array as is, but afterwards check if we are using the api (`is_api?`) and then append created_at to the array: `permitted << :created_at`

skx:
My apologies for the delay, somehow I missed the notification.

I've re-synced the discourse repository and updated the pull-request to only consider `created_at` when submitted via the API.  I hope that is sufficient?

PR #4207, opened 2016-05-02
FEATURE: Apply external window setting to Revision History

cpradio:
https://meta.discourse.org/t/external-links-in-history-don-t-follow-open-all-external-links-in-a-new-tab-setting/42995/36


eviltrout:
Thanks again :)

PR #4208, opened 2016-05-05
Update Sidekiq.

tgxworld:
There is a memory leak fix  https://github.com/mperham/sidekiq/blob/master/Changes.md#412


SamSaffron:
sure .. I upgraded, let's check on meta that all is fine

PR #4209, opened 2016-05-05
Modify search drop-down options text

jomaxro:
Search options text is inconsistent between on-screen and full-page search.  "Options" seems to imply the ability to change search parameters when clicked, not simply provide a help screen.  On-screen search text changed to reflect that.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -179,7 +179,7 @@ en:
     admin_title: "Admin"
     flags_title: "Flags"
     show_more: "show more"
-    show_help: "options"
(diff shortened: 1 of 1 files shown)


coding-horror:
No, this was specifically changed based on user feedback originally.

jomaxro:
Interesting.  Do you have a link to the previous discussion?  This just came up [here](https://meta.discourse.org/t/options-for-search-not-available-in-full-page/43721)

coding-horror:
It's on bbs.boingboing.net in the meta category

PR #4210, opened 2016-05-06
FIX: tag topic list header href leading to 404

damiencaselli:
I believe commit discourse/discourse-tagging@da528ed may have introduced an
error in tag list header href, [still present after merge](https://github.com/discourse/discourse/blob/master/app/views/list/list.erb#L7)
at the moment of the present commit.

While this is not really important in the standard, js-enabled application
browsing, it generates erroneous links that are displayed for SEO crawlers
such as Googlebot. This can be verified in the source of any tag page with
the proper user agent.

tag-list-seen-as-bot

four-oh-four

This introduces 404s to the site as a whole, and may be ranking some pages
(and the site) down.

### DIFF
--- a/app/views/list/list.erb
+++ b/app/views/list/list.erb
@@ -4,7 +4,7 @@
 
 <%- if SiteSetting.tagging_enabled && @tag_id %>
   <h1>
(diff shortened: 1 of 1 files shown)


nlalonde:
Thanks @damiencaselli !

PR #4211, opened 2016-05-06
Optional 'no_user_found_return_sso_url' parameter for using Discourse as SSO

reverendpaco:
Implemented an optional 'no_user_found_return_sso_url' parameter to be called by the client when client is using Discourse as an SSO and wants Discourse to redirect back to a place of the client's choosng when a user is not found.

Currently, the Discourse as an SSO implementation checks the cookies `_t` and `_session_forum` to see if the user is registered and present in the database (`_t` is the token that is located in the users table).  If a user is not found, the current implemenation forwards to the forum's /login URL.  This behavior may be what the client wants, but it would be good to give an option to the client to send somewhere else. 

This commit allows the client to embed an optional 'no_user_found_return_sso_url' parameter in the payload, prior to base64 and URL-encoding.  If the Discoure SSO endpoint detects that this parameter is present in the payload (and has a non-empty value) the Discourse server will redirect to this new location if it does not detect the user.  If this parameter is not present, then the redirection to /login will take place as it currently does. 

Additionally, as the client may choose to use the same URL for 'no_user_found_return_sso_url' as for 'return_url', this commit introduces a new query-string name-value pair to be sent back to the client 'no_user_found_return_sso_url' location. This parameter 'user_found' will **ALWAYS** be sent back to the client, either when the user is found and 'return_url' is used or when the user is not found 'no_user_found_return_sso_url' is used (values will be 'true' and 'false' respectively).

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -35,6 +35,7 @@ class SessionController < ApplicationController
     if SiteSetting.enable_sso_provider
       sso = SingleSignOn.parse(payload, SiteSetting.sso_secret)
       if current_user
+        sso.user_found = true;
         sso.name = current_user.name
         sso.username = current_user.username
         sso.email = current_user.email
@@ -46,6 +47,10 @@ class SessionController < ApplicationController
         else
           redirect_to sso.to_url(sso.return_sso_url)
         end
+      elsif (sso.no_user_found_return_sso_url != "" && !sso.no_user_found_return_sso_url.nil? )
+        sso.user_found = false;
+        no_user_URL = sso.no_user_found_return_sso_url
+        redirect_to sso.to_url(no_user_URL)
       else
         session[:sso_payload] = request.query_string
         redirect_to path('/login')
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -35,7 +35,7 @@ class SessionController < ApplicationController
     if SiteSetting.enable_sso_provider
       sso = SingleSignOn.parse(payload, SiteSetting.sso_secret)
       if current_user
-        sso.user_found = true;
+        sso.user_found = true
         sso.name = current_user.name
         sso.username = current_user.username
         sso.email = current_user.email
@@ -47,13 +47,14 @@ class SessionController < ApplicationController
         else
           redirect_to sso.to_url(sso.return_sso_url)
         end
-      elsif (sso.no_user_found_return_sso_url != "" && !sso.no_user_found_return_sso_url.nil? )
-        sso.user_found = false;
-        no_user_URL = sso.no_user_found_return_sso_url
-        redirect_to sso.to_url(no_user_URL)
       else
-        session[:sso_payload] = request.query_string
-        redirect_to path('/login')
+        sso.user_found = false
+        if sso.return_sso_unlogged_in_url.present?
+          redirect_to sso.to_url(sso.return_sso_unlogged_in_url)
+        else
+          session[:sso_payload] = request.query_string
+          redirect_to path('/login')
+        end
       end
     else
       render nothing: true, status: 404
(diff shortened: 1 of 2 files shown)


app/controllers/session_controller.rb
```diff
@@ -35,6 +35,7 @@ def sso_provider(payload=nil)
     if SiteSetting.enable_sso_provider
       sso = SingleSignOn.parse(payload, SiteSetting.sso_secret)
       if current_user
+        sso.user_found = true;
```
fantasticfears: you don't need `;` in the ruby

app/controllers/session_controller.rb
```diff
@@ -46,6 +47,10 @@ def sso_provider(payload=nil)
         else
           redirect_to sso.to_url(sso.return_sso_url)
         end
+      elsif (sso.no_user_found_return_sso_url != "" && !sso.no_user_found_return_sso_url.nil? )
```
fantasticfears: this block of logic shall be in `else` block, and you would want to set sso.user_found to all of them even it redirects to Discourse login page. (hence a client side plugin have the knowledge to do stuffs)

app/controllers/session_controller.rb
```diff
@@ -46,6 +47,10 @@ def sso_provider(payload=nil)
...
+        sso.user_found = false;
+        no_user_URL = sso.no_user_found_return_sso_url
+        redirect_to sso.to_url(no_user_URL)
```
fantasticfears: go ahead to save one line

lib/single_sign_on.rb
```diff
@@ -1,8 +1,9 @@
 class SingleSignOn
   ACCESSORS = [:nonce, :name, :username, :email, :avatar_url, :avatar_force_update, :require_activation,
-               :about_me, :external_id, :return_sso_url, :admin, :moderator, :suppress_welcome_message]
+                :no_user_found_return_sso_url,:user_found, :about_me, :external_id, :return_sso_url,
```
fantasticfears: You have an additional blank here. The most important part, I suppose `return_sso_unlogged_in_url` or something else could be more clear and align with `return_sso_url`?

fantasticfears:
Some minor feedback continued from [Meta topic](https://meta.discourse.org/t/using-discourse-as-a-sso-provider/32974/13). I left inline comments.

IMO, those redirection url shall be in SiteSetting. But that's another story. Ignore this line 👍

reverendpaco:
@fantasticfears   -- changes made.  return_sso_unlogged_in_url is the new name of the parameter.

reverendpaco:
I'm supposing there will be a time and place for this PR to be considered.  In case the "All Checks have failed" indicator is keeping eyes off of this, as far as I can tell, this is a false alarm and not related to the patch i've submitted.

reverendpaco:
Hello,  is there anything I can do to have this PR reviewed?

SamSaffron:
Sorry for the HUUUUUGGGGEEE delay here, I just am finding it hard to think of a use case for this added complexity, why does it matter if Discourse has or does not have the user registered? You are sending through full user information anyway...

reverendpaco:
I have a use case.  My own patched version of discourse depends on this use case currently -- a patched version I haven't upgraded since I submitted this PR back in May.  

My argument for this can be found at,
https://meta.discourse.org/t/using-discourse-as-a-sso-provider/32974/8

where I said "I would rather programmatically have the option to learn that this person has not yet logged in (or registered) and give them the opportunity on **my** site to continue anonymously"   [Note the emphasis on "my" site].

Where it stood before did not allow me to proffer a beacon iframe to check if the person was registered, as failure (person-not-logged-in) would redirect to discourse and not back to my site.  

The need for this is for sites like mine which are trying to piggyback *around* discourse and offer functionality (gaming in my case) that uses the forum as a user store.  My gaming site is at wordadoplicus.com, and the forum is at forum.wordadoplicus.com.

steph643:
I have a similar use case: I would like to be notified when the user abandons the Discourse login process (by clicking the modal top-right cross or outside the modal) so that I can update my site accordingly.

PR #4212, opened 2016-05-07
add missing i18n for zh_CN

williamherry:


techAPJ:
Thanks for this! But the translations should be updated via Transifex -- https://www.transifex.com/projects/p/discourse-org/

PR #4213, opened 2016-05-07
FIX: Importing user avatar when new user login by SSO

fantasticfears:
As per: https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045/292?u=fantasticfears

This PR contains a rather big refactor. But essentially it's a fix to ensure overrides semantics. Now the code structures like setup non-persistent model, update attrs, persistent models and overrides.
Integration tests build several rounds to ensure the order.

### DIFF
--- a/app/models/discourse_single_sign_on.rb
+++ b/app/models/discourse_single_sign_on.rb
@@ -80,6 +80,7 @@ class DiscourseSingleSignOn < SingleSignOn
     user.moderator = moderator unless moderator.nil?
 
     # optionally save the user and sso_record if they have changed
+    user.user_avatar.save! if user.user_avatar
     user.save!
 
     if bio && (user.user_profile.bio_raw.blank? || SiteSetting.sso_overrides_bio)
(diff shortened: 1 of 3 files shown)


spec/controllers/session_controller_spec.rb
```diff
@@ -264,63 +278,6 @@ def sso_for_ip_specs
       expect(response.code).to eq('419')
     end
 
-    describe 'can act as an SSO provider' do
```
fantasticfears: this was just rename to .sso_provider

app/models/discourse_single_sign_on.rb
```diff
@@ -47,104 +49,111 @@ def nonce_key
...
     end
 
     user
   end
 
-  def change_external_attributes_and_override(sso_record, user)
-    if SiteSetting.sso_overrides_email && user.email != email
-      user.email = email
+  def override_user_attributes
+    if SiteSetting.sso_overrides_email && @user_record.email != email
```
fantasticfears: no `presents` check here. Don't know what's the original intention. Keep as-is.

fantasticfears:
I don't know why I forget to check in the fix for importing avatar for new user in the first place...This is included now.

Rebased and checks by integration spec.

coding-horror:
It is up to @sam to review this and pull it in

SamSaffron:
This really worries me ... why is so much code being changed its very hard to digest it all and make sure there are zero regressions

fantasticfears:
@sam SSO is quite complicated in the core... One reason is to clean up process since it's getting complicated. Some changes depend on whether a record is save or not. The other one is the bug fix. I simply failed to find where is the problem in the beginning.

I can still find a place to enforce the sso_overrides_avatar in the old code if you'd like. Or I can write many more specs to address the possible problems?

fantasticfears:
Now this PR only fixs the avatar stuff.

SamSaffron:
travis failures seem legit, can you have a look @fantasticfears ?

fantasticfears:
Fixed!

SamSaffron:
cool, we just had this reported by a customer :) merging in

PR #4216, opened 2016-05-08
fix setting page input and select width inconsist issue

williamherry:
fix 
qq20160508-0 2x

and 
qq20160508-1 2x


ZogStriP:
Thanks 👍

PR #4217, opened 2016-05-08
Permit in topic search when full name is null

Mittineague:
I wanted to write an RSpec test for this, but fear problems I'm having doing so could stall this fix for an indefinite amount of time.

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -481,7 +481,7 @@ class Search
       if @term.present?
         if is_topic_search
           posts = posts.joins('JOIN users u ON u.id = posts.user_id')
-          posts = posts.where("posts.raw  || ' ' || u.username || ' ' || u.name ilike ?", "%#{@term}%")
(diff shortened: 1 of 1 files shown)


gschlager:
Am I missing something? Why use `quote_nullable(u.name)` instead of `COALESCE(u.name, '')`?

ZogStriP:
I agree with @gschlager, `quote_nullable(NULL)` returns `NULL`.

Mittineague:
I don't know if I should / need to squash the commits. 

It looks like the change took effect, but this is my first ever PR so I don't know.

Mittineague:
Should I / need to squash the change commit. 

This is my first ever PR so I don't know

ZogStriP:
> I don't know if I should / need to squash the commits.

We used to ask for squashing but now GitHub allows us to do it automatically when we merge ;)

> It looks like the change took effect, but this is my first ever PR so I don't know.

Awesome 👍

PR #4218, opened 2016-05-09
FIX: modal was not opening after navigating back in browser

techAPJ:
Bug report here: https://meta.discourse.org/t/cant-open-history-popup-after-using-browser-back-button/43898?u=techapj

This is happening for all the modals, after using browser back button while the modal is open, the modal remains hidden, probably due [this code](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/routes/discourse.js.es6#L97).

Not sure if this a perfect fix, but it detects if the modal has content and makes it visible again.

@eviltrout can you review this PR?

### DIFF
--- a/app/assets/javascripts/discourse/routes/discourse.js.es6
+++ b/app/assets/javascripts/discourse/routes/discourse.js.es6
@@ -94,7 +94,7 @@ export function cleanDOM() {
   $(document.activeElement).not("body").not(".no-blur").blur();
 
   Discourse.set('notifyCount',0);
-  $('#discourse-modal').modal('hide');
+  Discourse.__container__.lookup('route:application').send('closeModal');
   const hideDropDownFunction = $('html').data('hide-dropdown');
   if (hideDropDownFunction) { hideDropDownFunction(); }
 


eviltrout:
Can you explain to me how this works? I am not sure I understand.

techAPJ:
@eviltrout sure, here is how it works:

- when the modal is open and user presses back button, the modal is hidden via [this code](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/routes/discourse.js.es6#L97)

- when the user open the same modal again, [this code](https://github.com/discourse/discourse/blob/be3a5a56ccc284b352aa65080fc6b955f73cc72a/app/assets/javascripts/discourse/views/modal-body.js.es6#L9) is not run, hence the modal stays hidden

- to work around this, I am checking if the `modal-body` class is present in DOM, if it's present it means the modal has content and it's hidden, so we show the modal again

eviltrout:
Can you explain why [this code](https://github.com/discourse/discourse/blob/be3a5a56ccc284b352aa65080fc6b955f73cc72a/app/assets/javascripts/discourse/views/modal-body.js.es6#L9) is not run if the modal is hidden? I think that's the part I don't understand.

techAPJ:
It's because [`didInsertElement`](https://github.com/discourse/discourse/blob/be3a5a56ccc284b352aa65080fc6b955f73cc72a/app/assets/javascripts/discourse/views/modal-body.js.es6#L6) is not getting triggered since only the CSS is changed (`display: none`).

While if the model is closed via `closeModal()` then the modal body is removed (`this.render('hide-modal', { into: 'modal', outlet: 'modalBody' })`).

eviltrout:
Good find. Maybe to remove it we should instead render the hide modal? I feel like that would fix it?

techAPJ:
Yes, I tried to do that initially but I don't think we can call `closeModal()` (or render modal explicitly) inside [`cleanDOM()`](https://github.com/discourse/discourse/blob/514c3976f05785f487d275af1e96c43533e7a478/app/assets/javascripts/discourse/routes/discourse.js.es6#L77-L104) function.

eviltrout:
Ah right. In that case, you can have the `cleanDOM` function send it to the application route:

```javascript
Discourse.__container__.lookup('route:application').send('closeModal')
```

Using that to close the modal instead of hiding it should do the trick.

techAPJ:
@eviltrout: Awesome, thanks! Good to merge now? :)

eviltrout:
Yes, thanks for the effort :)

PR #4222, opened 2016-05-15
FIX[WIP]: return correct path for uploads stored on s3

scossar:
When assets are stored on S3 without a CDN, audio and video uploads are returning the wrong url.

`http://tests.testeleven.com//testeleven-forum-assets.s3-us-west-1.amazonaws.com/original/1X/06c34a2cb749a93f3551224a283080483cb417d4.wav`

The Discourse.BaseUrl is being added to the upload url without first checking if S3 is being used.

This PR adds a method to return the uploadLocation.

Issues:

 - at the moment, I am not able to test it with a CDN.

- is it safe to always use `https:` for accessing the Amazon bucket?

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js
+++ b/app/assets/javascripts/discourse/lib/utilities.js
@@ -249,16 +249,25 @@ Discourse.Utilities = {
             .join(", ");
   },
 
+  uploadUrl: function(url) {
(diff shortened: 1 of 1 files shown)


ZogStriP:
Looks fine, let's try it 👍

PR #4226, opened 2016-05-17
FEATURE: add setting permanent_session_cookie to configure session st…

xfalcox:
…ickiness

Now admins can turn make the login cookie die after the browser is closed, so the user needs to log in everytime.

See https://meta.discourse.org/t/session-timeout/34686/17?u=falco

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -890,6 +890,7 @@ en:
     post_undo_action_window_mins: "Number of minutes users are allowed to undo recent actions on a post (like, flag, etc)."
     must_approve_users: "Staff must approve all new user accounts before they are allowed to access the site. WARNING: enabling this for a live site will revoke access for existing non-staff users!"
     pending_users_reminder_delay: "Notify moderators if new users have been waiting for approval for longer than this many hours. Set to -1 to disable notifications."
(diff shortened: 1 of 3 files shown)


coding-horror:
@samsaffron will need to review

xfalcox:
I'm running this in production for a month now, it's pretty useful in a shared computers corporate environment. And with SSO it's painless.

SamSaffron:
This is fine, one thing we always wanted to add in this department was a concept of "session lifetime"

coding-horror:
Er.. what? I'm not following, how is it a good thing to never use cookies? So users have to log in every time they visit the site? At the very least this needs much clearer copy.

xfalcox:
> This is fine, one thing we always wanted to add in this department was a concept of "session lifetime"

Yes, but this would be a minutely job that check if the user is idle for more than X minutes and uses the force log out  feature. 

> Er.. what? I'm not following, how is it a good thing to never use cookies? So users have to log in every time they visit the site? At the very least this needs much clearer copy.

If an admin turn off the setting the cookie is still there, but it doesn't survive a computer restart or closing  all the tabs. 

If the user turn off a pc, in some environments, he expects to be logged out of everything. Like public libraries, pc bangs and companies pics. 

Also, I recommend using this alongside SSO so the user doesn't even know what's happening.

PR #4232, opened 2016-05-19
Capital S for Suche nach Beiträgen...

maikzimmermann1:
Grammatical change


ZogStriP:
I'm sorry but all translation changes has to go through https://www.transifex.com/ otherwise they would get erased next time we pull them from transifex.

PR #4235, opened 2016-05-23
Enhance, refactor populate.thor

pthomas551:
Enhance lib/tasks/populate.thor to allow for the creation of seed users and multiple seed topics.

Includes refactoring and use of ffaker gem to populate users and post data.

### DIFF
--- a/lib/tasks/populate.thor
+++ b/lib/tasks/populate.thor
@@ -1,10 +1,8 @@
+# Generates posts and topics
 class Populate < Thor
-
-  MAX_ERRORS = 5
-
-  desc "posts", "Generate posts in a topic"
+  desc "posts", "Generate posts"
   long_desc <<-LONGDESC
-    Create a topic with any number of posts, or add posts to an existing topic.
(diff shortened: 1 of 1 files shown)


coding-horror:
Didn't you work on the 999 reply post generation @nlalonde ? How does this look to you?

nlalonde:
Is it possible to do this without adding a new gem to the Gemfile? I would rather avoid adding a gem dependency for this, which is why I originally made the `hipster_words`, `image_posts`, and `wikipedia_posts` methods to populate data. We already have the [Fabrication gem](https://www.fabricationgem.org/), so maybe it can be used to generate users and other data.

pthomas551:
@nlalonde OK, will rework to avoid gem dependencies.

pthomas551:
Now reworked, gem dependency should be removed now. Also for future reference the hipster_words method generates amusing emails ;)

nlalonde:
Thanks, looks good! 👍

PR #4238, opened 2016-05-26
FIX: broken onebox avatar image

techAPJ:
This commit fixes the broken avatar image on internal oneboxes, as seen here:

https://meta.discourse.org/t/how-to-open-reply-window-via-url/44781/2?u=techapj

The issue is: in above case the `size` is `tiny`, and `size.to_i` converts it to zero, causing broken image.

@SamSaffron can you review this please? Thanks!


coding-horror:
Excellent the busted onebox avatars were making me crazy

PR #4240, opened 2016-05-26
FIX import of private messages from phpBB3

gschlager:
phpBB3 allows sending PMs to multiple users and it's possible to change the recipients of each message even if they belong to the same root message.

| msg_id | root_msg_id | author_id | to_address     | subject |
|--------|-------------|-----------|----------------|---------|
| 100    | 0           | 17        | u_65:u_73:u_34 | Foo     |
| 101    | 100         | 73        | u_17:u_65:u_34 | Re: Foo |
| 102    | 100         | 17        | u_73:u_34      | Re: Foo |
| 103    | 100         | 34        | u17:u_73       | Re: Foo |
| 104    | 100         | 65        | u_17:u_73:u_34 | Re: Foo |

In the above example messages with ID 102 and 103 should only be visible to some of the users involved in the conversation. Up until now all imported messages were visible to every user that was involved in a private conversation. That's quite bad.

This PR tries to fix that by comparing the users of every message. The importer creates a new topic if the users of a message do not match the users of the root message. In the above example the importer would create two topics:
- Topic with title "Foo" consisting of `msg_id` 100, 101 and 104
- Topic with title "Re: Foo" consisting of `msg_id` 102 and 103

The new import method also _automatically_ fixes PMs that were previously migrated from phpBB2 to phpBB3. Those messages always have a `root_msg_id` of 0 which normally would result in topics consisting of just one message.

| msg_id | root_msg_id | author_id | to_address     | subject |
|--------|-------------|-----------|----------------|---------|
| 100    | 0           | 17        | u_65:u_73:u_34 | Foo     |
| 101    | 0           | 73        | u_17:u_65:u_34 | Re: Foo |
| 102    | 0           | 17        | u_73:u_34      | Re: Foo |
| 103    | 0           | 34        | u17:u_73       | Re: Foo |
| 104    | 0           | 65        | u_17:u_73:u_34 | Re: Foo |

They are now grouped by subject and involved users which results in the same two topics as mentioned above.

### DIFF
--- a/script/import_scripts/phpbb3/database/database_3_0.rb
+++ b/script/import_scripts/phpbb3/database/database_3_0.rb
@@ -161,82 +161,39 @@ module ImportScripts::PhpBB3
       SQL
     end
 
-    def count_messages(use_fixed_messages)
-      if use_fixed_messages
-        count(<<-SQL)
-          SELECT COUNT(*) AS count
-          FROM #{@table_prefix}_import_privmsgs
-        SQL
-      else
-        count(<<-SQL)
-          SELECT COUNT(*) AS count
-          FROM #{@table_prefix}_privmsgs
-        SQL
-      end
+    def count_messages
+      count(<<-SQL)
+        SELECT COUNT(*) AS count
+        FROM #{@table_prefix}_privmsgs m
+        WHERE NOT EXISTS ( -- ignore duplicate messages
+            SELECT 1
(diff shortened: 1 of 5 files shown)


coding-horror:
@techapj if this looks good to you, can you pull it in?

techAPJ:
Looks good to me! 👍

PR #4241, opened 2016-05-26
Change default for mailing list mode frequency

gdpelican:
Make it so new user options are set to 'individual' by default

Means that we preserve existing mailing list user's settings. :D


coding-horror:
We should get this mergificated since it is causing problems for us on live customer sites

SamSaffron:
This fixes all brand new accounts created but does nothing to fix the history, hmmmm

SamSaffron:
I am just going to have to do a once off migration to set this to 1 and break a few people, we released beta so some people are going to be upset

SamSaffron:
yeah I think the default of 0 is correct, what we need here is a once-off query to fix up all old records so they work as expected... adding now

PR #4245, opened 2016-06-03
FIX: Repair one: in client.ru.yml and client.uk.yml for case 21,31,... when it is possible.

alvir:


ZogStriP:
Thanks be all translation changes must be done via [transifex](https://www.transifex.com/discourse/discourse-org/) otherwise we lose all the changes when we pull the translations from transifex ;)

PR #4249, opened 2016-06-03
UX: Make room for larger buttons in timeline 

gschlager:
Don't wrap buttons in timeline if there is enough room. This seems to work fine on all resolutions.
https://meta.discourse.org/t/required-css-layout-tweaks-for-new-vertical-timeline-in-1-6/45102/6

### DIFF
--- a/app/assets/stylesheets/desktop/topic-timeline.scss
+++ b/app/assets/stylesheets/desktop/topic-timeline.scss
@@ -15,9 +15,11 @@
 
   @media all and (min-width: 1140px) {
     margin-left: 820px;
+    width: $large-width - 820px;
   }
   @media all and (min-width: 1250px) {
     margin-left: 900px;
+    width: $large-width - 900px;
   }
 
   position: fixed;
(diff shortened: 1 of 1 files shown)


coding-horror:
wait a second, wouldn't this mean the page expands in width for languages where Reply is a very long word? That'd be a deal breaker on iPad, as the viewport *has* to fit what's available; it could get _clipped_ but it can't expand more than about 20px to the right without impacting iPad scaling width.

coding-horror:
Perhaps the best thing to do is limit size of the _actual reply button itself_ instead to 150px?

gschlager:
That's what it looks like with a long button label on my iPad.

image

I can't see any obvious problems (except that the button looks kinda ugly when it gets clipped).

coding-horror:
That whole column should clip when text is too long; perhaps a better
approach is to tie this to media queries based on width? If the media
queries are telling us there is more width, sure, I can support this.

coding-horror:
Why not this?

```
.timeline-container .topic-timeline .timeline-footer-controls button {
    margin-right: 0.5em;
    overflow: hidden;
    white-space: nowrap;
}
```

Which results in

image

gschlager:
Ok, I tried it with media queries:

On iPad and screen sizes below 1140px the buttons have a `max-width` of `140px`.
image

On larger screens the buttons can use the available space:
image

PR #4250, opened 2016-06-05
FIX: Show eye-slash icon when able to mark a topic as Unlisted, and e…

cpradio:
…ye icon when able to mark it Listed

Meta Discussion
https://meta.discourse.org/t/unlisted-listed-icon-reversed/45264


techAPJ:
Thanks! 👍

PR #4254, opened 2016-06-08
FEATURE: Add public type to polls.

tgxworld:


ZogStriP:
LGTM 👍

tgxworld:
@ZogStriP Thanks for reviewing :)

plugins/poll/assets/javascripts/views/poll.js.es6
```diff
@@ -3,17 +3,12 @@ import { on } from "ember-addons/ember-computed-decorators";
...
-  attributeBindings: ["data-poll-type", "data-poll-name", "data-poll-status"],
+  attributeBindings: ["data-poll-type", "data-poll-name", "data-poll-status", "data-poll-public"],
 
   poll: Em.computed.alias("controller.poll"),
 
   "data-poll-type": Em.computed.alias("poll.type"),
   "data-poll-name": Em.computed.alias("poll.name"),
   "data-poll-status": Em.computed.alias("poll.status"),
-
-  @on("didInsertElement")
```
tgxworld: @ZogStriP Btw I'm not sure why this was required and the commit goes all the way back to 2014. I'm removing it because it is causing problems on mobile and doesn't seem to be used on desktop at all.

ZogStriP: I agree it's a weird-looking piece of code... Not sure why it was in there :( I guess we'll quickly see if it was working.

PR #4256, opened 2016-06-09
Webhooks Milestone1

fantasticfears:


app/assets/javascripts/admin/routes/admin-web-hook.js.es6
```diff
@@ -0,0 +1,15 @@
+import WebHook from 'admin/models/web-hook';
+
+export default Discourse.Route.extend({
+  serialize(model) {
+    return { web_hook_id: model.get('id') };
+  },
+
+  model(params) {
```
tgxworld: Let's use custom store instead, it'll save you from alot of the glue that you have to write.

https://meta.discourse.org/t/upgrading-our-front-end-models-to-use-a-store/27837

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,32 @@
+import { popupAjaxError } from 'discourse/lib/ajax-error';
+import { default as computed, on, observe } from 'ember-addons/ember-computed-decorators';
+
+export default Ember.Controller.extend({
+  @computed('model.isSaving')
+  savingStatus(isSaving) {
+    return isSaving ? I18n.t('saving') : '';
+  },
+
+  wildcardEvent: Ember.computed('model.wildcard_event', {
```
tgxworld: This can be written as https://github.com/rwjblue/ember-computed-decorators#real-world-getset-syntax using `ember-computed-decorators`

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -1,24 +1,52 @@
 import RestModel from 'discourse/models/rest';
+import computed from 'ember-addons/ember-computed-decorators';
+import { registerHelper } from 'discourse/lib/helpers';
 
-const WebHook = RestModel.extend({
-  content_types: ['application/json', 'application/x-www-form-urlencoded']
-});
+export default RestModel.extend({
+  content_types: Em.A([{
```
tgxworld: camelCase for Javascript files :)

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,52 @@
+import RestModel from 'discourse/models/rest';
+import computed from 'ember-addons/ember-computed-decorators';
+import { registerHelper } from 'discourse/lib/helpers';
+
+export default RestModel.extend({
+  content_types: Em.A([{
```
tgxworld: we use camelCase for javascript files :)

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Admin::WebHooksController < Admin::AdminController
+
+  def index
+    webhooks = WebHook.all
```
tgxworld: You'll need to paginate this :)

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,32 @@
+module Jobs
+  class EmitWebHookEvent < Jobs::Base
+    # https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
+    RETRIES = false # TODO: this gives around 13 mins (5)
```
tgxworld: I'm guessing the user will have to manually retry this?

lib/validators/web_hook_validator.rb
```diff
@@ -0,0 +1,15 @@
+module Validators; end
+
+class Validators::WebHookValidator < ActiveModel::Validator
+
+  def validate(web_hook)
+    web_hook.errors.add(:payload_url, :blank, options) if web_hook.payload_url.blank?
```
tgxworld: Is there a reason why we can't use a `validates :payload_url, presence: true` here?

fantasticfears: No. The only thing I really need is checking as least one event is set. But I collects all those presence check in the validator anyways

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    webhook = WebHook.find(params[:id])
+
+    if webhook.update_attributes(webhook_params)
+      render_serialized(webhook, AdminWebHookSerializer, root: 'web_hooks', rest_serializer: true)
+    else
+      render_json_error webhook.errors.full_messages
+    end
+  end
+
+  def webhook_params
```
tgxworld: This method is usually made private

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -0,0 +1,46 @@
+class Admin::WebHooksController < Admin::AdminController
+
+  def index
+    data = {
+      web_hook_event_types: WebHookEventType.all,
+      web_hooks: WebHook.all # TODO: Pagination?
+    }
+    render_serialized(OpenStruct.new(data), AdminWebHooksSerializer, rest_serializer: true)
```
tgxworld: hmm perhaps just pass in the `WebHook.all` here and add association for `WebHookEventType` in the serializer?

fantasticfears: I was thinking I could sideload all relevant event types like `Badge` models. Yes, I would go for a static list for Ember now.

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,42 @@
...
+
+  actions: {
+    save() {
+      this.set('saved', false);
+
+      const model = this.get('model');
+      return model.save().then(() => {
+        this.set('saved', true);
+        this.get('controllers.adminWebHooks').get('model').addObject(model);
+      }).catch(popupAjaxError);
```
tgxworld: The state of `saved` is not handled if it errors

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,42 @@
...
+    },
+
+    destroy() {
+      return bootbox.confirm(I18n.t("admin.web_hooks.delete_confirm"), I18n.t("no_value"), I18n.t("yes_value"), (result) => {
+        if (result) {
+          const model = this.get('model');
+          model.destroyRecord().then(() => {
+            this.get('controllers.adminWebHooks').get('model').removeObject(model);
+            this.transitionToRoute('adminWebHooks');
+          }).catch(popupAjaxError);
```
tgxworld: Ditto: https://github.com/discourse/discourse/pull/4256/files?diff=unified#r68527613

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,100 @@
...
+  // FIXME: When enter event details from list, this observes are not used
+  // which results in an empty input box
+  @observes('category_ids')
+  updateCategoryFilters() {
+    this.set("categoryFilters", Discourse.Category.findByIds(this.get('category_ids')));
+  },
+
+  @computed('web_hook_event_types.[]')
+  existWildcardEvent: {
+    get() {
```
tgxworld: generally we just define it as `existWildcardEvent()` if there is not `set()`

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -0,0 +1,50 @@
+class Admin::WebHooksController < Admin::AdminController
+
+  def index
+    # TODO: Pagination
+    render_serialized(WebHook.all, AdminWebHookSerializer, root: 'web_hooks')
+  end
+
+  def show
+    webhook = WebHook.find(params[:id])
```
tgxworld: Usually I will extract this into a `before_filter` callback

Gemfile
```diff
@@ -94,7 +94,7 @@ gem 'pry-rails', require: false
 gem 'r2', '~> 0.2.5', require: false
 gem 'rake'
 
-
+gem 'faraday_middleware'
```
tgxworld: Is there a reason why you need `faraday_middleware` here? 

We already have [Excon](https://github.com/excon/excon) :)

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,77 @@
+require 'faraday_middleware'
+
+module Jobs
+  class EmitWebHookEvent < Jobs::Base
+    # https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
```
tgxworld: No need for this comment.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,77 @@
...
+    RETRIES = 5
```
tgxworld: No need for this constant if it isn't being used anywhere else

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,77 @@
...
+      conn.post do |req|
+        req.body = body
+        req.headers[HEADER_EVENT_NAME] = @opts[:event_name]
+        if @web_hook.secret.present?
+          req.headers[HEADER_SIGNATURE] = OpenSSL::HMAC.hexdigest("sha256", @web_hook.secret, body.to_s)
+        end
+      end
+    end
+
+    private def build_web_hook_body
```
tgxworld: Nitpick: 

```
private

def ....

end
```

app/models/web_hook.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def self.content_types
+    @content_types ||= Enum.new('application/json' => 1,
+                                'application/x-www-form-urlencoded' => 2)
+  end
+
+  def self.find_by_type(type)
+    WebHookEventType.where('name = ? OR name = ?', 'wildcard', type.to_s)
+                    .map(&:web_hooks)
+                    .flatten
```
tgxworld: Let's use `flat_map` here.

app/models/web_hook.rb
```diff
@@ -0,0 +1,48 @@
...
+                    .uniq
+                    .keep_if(&:active)
```
tgxworld: I think within the map we can do `web_hooks.where(active: true)`?

db/migrate/20160615235959_add_web_hooks.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      t.timestamps
+    end
+
+    create_join_table :web_hooks, :web_hook_event_types
+    create_join_table :web_hooks, :groups
+    create_join_table :web_hooks, :categories
+
+    add_index :web_hook_event_types, :name
+    add_index :web_hook_event_types_hooks, [:web_hook_event_type_id, :web_hook_id], name: 'index_web_hook_event_types_hooks_on_ids'
```
tgxworld: I think this needs to be unique?

db/migrate/20160615235959_add_web_hooks.rb
```diff
@@ -0,0 +1,25 @@
...
+      t.boolean :active, default: false, null: false
+
+      t.timestamps
+    end
+
+    create_join_table :web_hooks, :web_hook_event_types
+    create_join_table :web_hooks, :groups
+    create_join_table :web_hooks, :categories
+
+    add_index :web_hook_event_types, :name
```
tgxworld: Does this have to be unique?

spec/models/web_hook_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    it "'url_encoded' (application/x-www-form-urlencoded) should be at 2st position" do
+      expect(@content_types['application/x-www-form-urlencoded']).to eq 2
+    end
+  end
+
+  context 'web hooks' do
+    before(:all) do
+      @post_hook = Fabricate(:web_hook)
+      @topic_hook = Fabricate(:web_hook)
```
tgxworld: can we use `let` instead?

tgxworld: `before(:all)` will share the particular instance between all the test groups

spec/models/web_hook_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    describe '#find_by_type' do
+      it 'find relevant hooks' do
+        expect(WebHook.find_by_type(:post)).to eq([@post_hook])
+        expect(WebHook.find_by_type(:topic)).to eq([@topic_hook])
+      end
+
+      it 'excludes inactive hooks' do
+        @post_hook.active = false
+        @post_hook.save!
```
tgxworld: `update_attributes!` works too

spec/models/web_hook_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+      after(:all) { @wildcard_hook.destroy }
+
+      describe '#find_by_type' do
+        it 'can find wildcard hooks' do
+          expect(WebHook.find_by_type(:wildcard)).to eq([@wildcard_hook])
+        end
+
+        it 'can include wildcard hooks' do
+          expect(WebHook.find_by_type(:post)).to include(@post_hook)
```
tgxworld: Can we assert against the actual output of `WebHook.find_by_type(:post)` instead? I think the output we want here is to make sure we get `[@post_hook, @topic_hook]`

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
+require_dependency 'jobs/base'
+
+describe Jobs::EmitWebHookEvent do
+  before(:all) do
+    @post_hook = Fabricate(:web_hook)
```
tgxworld: Use `let` here

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    @inactive_post_hook = Fabricate(:inactive_web_hook)
+    @category = Fabricate(:category)
+    @topic_hook = Fabricate(:web_hook)
+    @topic_hook.categories = [@category]
```
tgxworld: You can do something like

`let(:topic_hook) { Fabricate(:web_hook, categories: [@category]) }`

spec/jobs/emit_web_hook_event_spec.rb
tgxworld: Read up on https://www.relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let

spec/models/web_hook_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+
+      it 'excludes inactive hooks' do
+        @post_hook.active = false
+        @post_hook.save!
+
+        expect(WebHook.find_by_type(:post)).to eq([])
+
+        @post_hook.active = true
+        @post_hook.save!
```
tgxworld: ditto https://github.com/discourse/discourse/pull/4256/files?diff=unified#r68528718

app/assets/javascripts/admin/components/admin-web-hook-event-chooser.js.es6
```diff
@@ -0,0 +1,46 @@
+import computed from 'ember-addons/ember-computed-decorators';
+
+export default Ember.Component.extend({
+  tagName: '',
```
tgxworld: is this being used anywhere?

fantasticfears: There is a `float`. So the wrapper tag is removed here

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
fantasticfears: Not sure whether it's needed at all. `saved` is false before the callback. As long as it's false, no text will be shown by `saveButtonText`. The text will show saving status after all

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+
+  contentTypes: Em.A([{
+    id: 1,
+    name: 'application/json'
+  }, {
+    id: 2,
+    name: 'application/x-www-form-urlencoded'
+  }]),
+  content_type: 1, // prefer json,
+  web_hook_event_types: Em.A([{ // default event type
```
fantasticfears: Setting those defaults on routes or here?

app/serializers/web_hooks_post_serializer.rb
```diff
@@ -0,0 +1,21 @@
+class WebHooksPostSerializer < PostSerializer
+  def include_yours?
```
fantasticfears: Use `include_` to exclude some UI specific fields or building from basic_serializer?

app/assets/javascripts/admin/components/admin-web-hook-event-chooser.js.es6
```diff
@@ -0,0 +1,46 @@
...
+  },
+
+  @computed('typeName')
+  details(typeName) {
+    return I18n.t(`admin.web_hooks.${typeName}_event.details`);
+  },
+
+  @computed
+  eventTypeExists: {
+    get() {
```
tgxworld: Nitpick: If we're not using `set()`, we should follow the rest of the convention and just define the computed property without `get()`

app/assets/javascripts/admin/templates/web-hooks-show.hbs
```diff
@@ -0,0 +1,78 @@
...
+  {{fa-icon 'arrow-left'}}
+  {{i18n 'admin.web_hooks.go_back'}}
+{{/link-to}}
+
+<div class='web-hook-container'>
+  <p>{{i18n 'admin.web_hooks.detailed_instruction'}}</p>
+  <form class='web-hook form-horizontal'>
+    <div>
+      <label for='payload-url'>{{i18n 'admin.web_hooks.payload_url'}}</label>
+      {{text-field name="payload-url" value=model.payload_url placeholderKey="admin.web_hooks.payload_url_placeholder"}}
```
tgxworld: lets add a just in time validation for the input fields that require it. `input-validation` should do that trick. Also, I noticed that we are not making sure it is a valid URL

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  def update
+    if @webhook.update_attributes(webhook_params)
+      render_serialized(@webhook, AdminWebHookSerializer, root: 'web_hook')
+    else
+      render_json_error @webhook.errors.full_messages
+    end
+  end
+
+  def destroy
+    @webhook.destroy
```
tgxworld: Prefer `destroy!` instead

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+    id: 3,
+    name: 'post'
+  }]),
```
tgxworld: Ah ok I think I didn't get what you mean previously. Hardcoding all the type here is bad.

You can probably do something like this in your controller

```
    render json: {
      web_hooks: serialize_data(WebHook.all, AdminWebHookSerializer)
      web_hook_event_type: serialize_data(WebHookEventType.all, some_serializer)
      other models that you need
    }
```

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+  categoryFilters: Em.A(),
```
tgxworld: Becareful here, this array will be shared across all instance of the component

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+
+  // FIXME: When enter event details from list, this observes are not used
+  // which results in an empty input box
+  @observes('category_ids')
+  updateCategoryFilters() {
+    this.set("categoryFilters", Discourse.Category.findByIds(this.get('category_ids')));
```
tgxworld: `import Category from ...` instead of using `Discourse.Category`

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+      return event.name === 'wildcard';
+    });
+  },
+
+  @computed('existWildcardEvent')
+  wildcardEvent: {
+    get(existWildcardEvent) {
+      return existWildcardEvent;
+    },
+    set(_value, existWildcardEvent) {
```
tgxworld: `_` infront of a variable name is usually used to indicate that the variable is not going to be used.

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+      const value = (_value === true || _value.toString().toLowerCase() === 'true');
```
tgxworld: just always `toString`. Not sure if we need a `toLowerCase` here

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+  // which results in an empty input box
+  @observes('category_ids')
+  updateCategoryFilters() {
+    this.set("categoryFilters", Discourse.Category.findByIds(this.get('category_ids')));
+  },
+
+  @computed('web_hook_event_types.[]')
+  existWildcardEvent() {
+    return this.get('web_hook_event_types').any(event => {
+      return event.name === 'wildcard';
```
tgxworld: Instead of using the hardcoded string `'wildcard'`, prefer what you do in https://github.com/discourse/discourse/pull/4256/files#diff-c8e99453565042747d57cd4408ec83b3R63 instead

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,98 @@
...
+
+  // FIXME: When enter event details from list, this observes are not used
+  // which results in an empty input box
+  @observes('category_ids')
+  updateCategoryFilters() {
+    this.set("categoryFilters", Discourse.Category.findByIds(this.get('category_ids')));
+  },
+
+  @computed('web_hook_event_types.[]')
+  existWildcardEvent() {
```
tgxworld: I'll probably rename this to `isWildcardEvent`

app/assets/javascripts/admin/templates/web-hooks-show.hbs
```diff
@@ -0,0 +1,78 @@
...
+
+  <div class='controls'>
+    <button {{action 'save'}} disabled={{model.isSaving}} class='btn btn-default'>{{saveButtonText}}</button>
+    {{#unless model.isNew}}
+      {{d-button class="btn-danger" label="admin.web_hooks.destroy" action="destroy"}}
+    {{/unless}}
+    <span class='saving'>{{savingStatus}}</span>
+  </div>
+
+  <table class='web-hook-events'>
```
tgxworld: do we need this?

app/controllers/topics_controller.rb
```diff
@@ -343,6 +343,7 @@ def destroy
     PostDestroyer.new(current_user, first_post, { context: params[:context] }).destroy
 
     DiscourseEvent.trigger(:topic_destroyed, topic, current_user)
+    WebHook.enqueue_hooks(:topic, topic_id: topic.id, user_id: current_user.id, category_id: params[:category_id]&.to_i)
```
tgxworld: I will use the DiscourseEvent hooks instead for this instead of adding more code into the controller

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,75 @@
+module Jobs
+  class EmitWebHookEvent < Jobs::Base
+    sidekiq_options retry: 5, queue: 'critical'
```
tgxworld: probably put this on the low priority queue.

app/models/web_hook.rb
```diff
@@ -0,0 +1,46 @@
...
+  validates_presence_of :web_hook_event_types
+
+  def self.content_types
+    @content_types ||= Enum.new('application/json' => 1,
+                                'application/x-www-form-urlencoded' => 2)
+  end
+
+  def self.find_by_type(type)
+    WebHookEventType.where("name = 'wildcard' OR name = ?", type.to_s)
+                    .flat_map { |type| type.web_hooks.where(active: true) }
```
tgxworld: you're probably going to get a N+1 query here

app/models/web_hook_event_type.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  validates :name, presence: true, uniqueness: true
+end
+
+# == Schema Information
+#
+# Table name: web_hook_event_types
+#
+#  id   :integer          not null, primary key
+#  name :string
```
tgxworld: Missing unique index

lib/post_creator.rb
```diff
@@ -220,8 +220,12 @@ def build_post_stats
   end
 
   def trigger_after_events(post)
-    DiscourseEvent.trigger(:topic_created, post.topic, @opts, @user) unless @opts[:topic_id]
+    unless @opts[:topic_id]
+      DiscourseEvent.trigger(:topic_created, post.topic, @opts, @user)
+      WebHook.enqueue_hooks(:topic, topic_id: post.topic.id, user_id: @user.id, category_id: post.topic.category&.id)
+    end
     DiscourseEvent.trigger(:post_created, post, @opts, @user)
+    WebHook.enqueue_hooks(:post, post_id: post.id, topic_id: post.topic.id, user_id: @user.id, category_id: post.topic.category&.id)
```
tgxworld: We should make use of the existing triggers ehre

app/assets/javascripts/admin/components/admin-web-hook-event-chooser.js.es6
fantasticfears: That must be slicked among rebase...

app/assets/javascripts/admin/templates/web-hooks-show.hbs
fantasticfears: Mind a placeholder here? I will follow this with a table of historical webhook events.

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,72 @@
...
+import { default as computed, observes } from 'ember-addons/ember-computed-decorators';
+
+export default RestModel.extend({
+  contentTypes: Em.A([{
+    id: 1,
+    name: 'application/json'
+  }, {
+    id: 2,
+    name: 'application/x-www-form-urlencoded'
+  }]),
```
tgxworld: Can we pass this down from the serializer as well?

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,72 @@
...
+  content_type: 1, // prefer json
+  wildcard_web_hook: false,
+  verify_certificate: true,
+  active: false,
+  web_hook_event_types: Em.A([{ // default event type
+    id: 2,
+    name: 'post'
+  }]),
```
tgxworld: Same as https://github.com/discourse/discourse/pull/4256/files#r69886639.

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,72 @@
...
+
+  init() {
```
tgxworld: ahh remember to call `this._super()` when you're overriding `init()`

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,72 @@
...
+  },
+
+  @observes('category_ids')
+  updateCategoryFilters() {
+    this.set('categoryFilters', Category.findByIds(this.get('category_ids')));
+  },
+
+  @computed('wildcard_web_hook', 'web_hook_event_types.[]')
+  description(isWildcardWebHook, types) {
+    var desc = '';
```
tgxworld: Nitpick: we prefer `let`

app/assets/javascripts/admin/templates/web-hooks-show.hbs
```diff
@@ -0,0 +1,80 @@
...
+  {{i18n 'admin.web_hooks.go_back'}}
+{{/link-to}}
+
+<div class='web-hook-container'>
+  <p>{{i18n 'admin.web_hooks.detailed_instruction'}}</p>
+  <form class='web-hook form-horizontal'>
+    <div>
+      <label for='payload-url'>{{i18n 'admin.web_hooks.payload_url'}}</label>
+      {{text-field name="payload-url" value=model.payload_url placeholderKey="admin.web_hooks.payload_url_placeholder"}}
+      {{input-tip validation=urlValidation}}
```
tgxworld: :+1:

app/assets/javascripts/admin/templates/web-hooks-show.hbs
```diff
@@ -0,0 +1,80 @@
...
+                  name="content-type"
+                  nameProperty="name"
+                  valueAttribute="id"
+                  value=model.content_type}}
+    </div>
+
+    <div>
+      <label for='secret'>{{i18n 'admin.web_hooks.secret'}}</label>
+      {{text-field name="secret" value=model.secret placeholderKey="admin.web_hooks.secret_placeholder"}}
+      {{input-tip validation=secretValidation}}
```
tgxworld: :+1:

app/assets/javascripts/admin/templates/web-hooks-show.hbs
```diff
@@ -0,0 +1,80 @@
...
+      {{/unless}}
+      <div>
+        {{radio-button class="subscription-choice" name="subscription-choice" value="wildcard" selection=model.webHookType}}
+        {{i18n 'admin.web_hooks.wildcard_event'}}
+      </div>
+    </div>
+
+    <div class='filters'>
+      <div>
+        <label><span class="icon fa fa-circle tracking"></span> {{i18n 'admin.web_hooks.category_filters'}}</label>
```
tgxworld: There is an extra space in this line

tgxworld: we have a component for creating icons `{{fa-icon "plus"}}`

app/assets/javascripts/admin/templates/web-hooks.hbs
```diff
@@ -0,0 +1,34 @@
...
+  <table class='web-hooks'>
+    <thead>
+      <tr>
+        <th>{{i18n 'admin.web_hooks.payload_url'}}</th>
+        <th>{{i18n 'admin.web_hooks.description'}}</th>
+        <th>{{i18n 'admin.web_hooks.controls'}}</th>
+      </tr>
+    </thead>
+    <tbody>
+      {{#each model as |k|}}
```
tgxworld: I don't know what `k` is :crying_cat_face:

app/assets/javascripts/discourse/lib/utilities.js
```diff
@@ -101,6 +101,12 @@ Discourse.Utilities = {
     return re.test(email);
   },
 
+  urlValid: function(url) {
+    // see:  @imme_emosol https://mathiasbynens.be/demo/url-regex
+    var re = /(https?|ftp):\/\/(-\.)?([^\s/?\.#-]+\.?)+(\/[^\s]*)?$/i;
```
tgxworld: hmmm can you actually POST to a ftp service?

app/serializers/web_hooks_topic_serializer.rb
```diff
@@ -0,0 +1,2 @@
+class WebHooksTopicSerializer < TopicViewSerializer
```
tgxworld: Do we need this serializer if it doesn't add any new methods?

config/application.rb
```diff
@@ -164,6 +164,7 @@ def config.database_configuration
     config.ember.handlebars_location = "#{Rails.root}/vendor/assets/javascripts/handlebars.js"
 
     require 'auth'
+    require 'discourse_web_hook'
```
tgxworld: Can the events be declared in the `WebHook` class?

config/locales/client.en.yml
```diff
@@ -2362,6 +2362,49 @@ en:
         all_users: "All Users"
         note_html: "Keep this key <strong>secret</strong>, all users that have it may create arbitrary posts as any user."
 
+      web_hooks:
+        title: "Webhooks"
+        none: "(EXPERIMENTAL) There are no webhook right now."
```
tgxworld: You can remove experimental :wink:

config/locales/server.en.yml
```diff
@@ -189,6 +189,8 @@ en:
     one: "Sorry, new users can only put one link in a post."
     other: "Sorry, new users can only put %{count} links in a post."
 
+  no_web_hook_event: "Sorry, you have to set at least one Webhook event."
```
tgxworld: better to nest this under `web_hooks: `

spec/models/web_hook_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        it 'accepts additional parameters' do
+          Jobs.expects(:enqueue).with(:emit_web_hook_event, web_hook_id: post_hook.id, post_id: 1, event_name: 'post')
+          Jobs.expects(:enqueue).with(:emit_web_hook_event, web_hook_id: wildcard_hook.id, post_id: 1, event_name: 'post')
+
+          WebHook.enqueue_hooks(:post, post_id: 1)
+        end
+      end
+    end
+
+
```
tgxworld: Extra line :scissors:

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,81 @@
...
+                                  'application/json'
+                                end
+
+      body = build_web_hook_body
+
+      if @web_hook.secret.present?
+        headers['X-Discourse-Event-Signature'] = OpenSSL::HMAC.hexdigest("sha256", @web_hook.secret, body)
+      end
+
+      conn.post(headers: headers, body: body)
```
tgxworld: Hmm we need to be checking if the response is 404 here.

app/assets/javascripts/admin/components/admin-web-hook-event-chooser.js.es6
```diff
@@ -0,0 +1,44 @@
...
+    set(value) {
+      const type = this.get('type');
+      const model = this.get('model');
+      // add an association when not exists
+      if (value !== this.get('eventTypeExists')) {
+        if (value) {
+          model.addObject(type);
+        } else {
+          model.removeObjects(model.filter(eventType => {
+            return (eventType.name || eventType.get('name')) === (type.name || type.get('name'));
```
tgxworld: `eventType.name || eventType.get('name')` is a little odd to me. Is `eventType` and `type` not always an Ember object?

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,100 @@
...
+  contentTypes: Em.computed.alias('controllers.adminWebHooks.contentTypes'),
+
+  @computed('model.isSaving', 'saved', 'saveButtonDisabled')
+  savingStatus(isSaving, saved, saveButtonDisabled) {
+    if (isSaving) {
+      return I18n.t('saving');
+    } else if (!saveButtonDisabled && saved) {
+      return I18n.t('saved');
+    }
+    this.set('saved', false);
```
tgxworld: Why does this property need to be set to `false`? Are you leaking state somewhere?

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,100 @@
...
+  @computed('model.payload_url')
+  urlValidation(url) {
+    // If blank, fail without a reason
+    if (Ember.isEmpty(url)) {
+      return Discourse.InputValidation.create({
+        failed: true
+      });
+    }
+
+    if (!Discourse.Utilities.urlValid(url) || url.indexOf(' ') !== -1) {
```
tgxworld: `url.indexOf(' ') !== -1` Can `Discourse.Utilities.urlValid(url)` catch this instead? Otherwise, we should trim the input

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,100 @@
...
+
+      const model = this.get('model');
+      return model.save().then(() => {
+        this.set('saved', true);
+        this.get('controllers.adminWebHooks').get('model').addObject(model);
+      }).catch(popupAjaxError);
+    },
+
+    destroy() {
+      return bootbox.confirm(I18n.t('admin.web_hooks.delete_confirm'), I18n.t('no_value'), I18n.t('yes_value'), (result) => {
```
tgxworld: Nitpick: `result => {  }` works too :) No need paranthesis

app/assets/javascripts/admin/controllers/admin-web-hooks.js.es6
```diff
@@ -0,0 +1,15 @@
+import { popupAjaxError } from 'discourse/lib/ajax-error';
+
+export default Ember.Controller.extend({
+  actions: {
+    destroy(webhook) {
+      return bootbox.confirm(I18n.t('admin.web_hooks.delete_confirm'), I18n.t('no_value'), I18n.t('yes_value'), (result) => {
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/4256/files#r70645559

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -0,0 +1,63 @@
...
+    const categories = this.get('categoryFilters');
+
+    let webhook = {
+      payload_url: this.get('payload_url'),
+      content_type: this.get('content_type'),
+      secret: this.get('secret'),
+      wildcard_web_hook: this.get('wildcard_web_hook'),
+      verify_certificate: this.get('verify_certificate'),
+      active: this.get('active'),
+      web_hook_event_type_ids: Ember.isEmpty(types) ? [null] : types.map(type => type.id),
```
tgxworld: `[null]` is really odd. Any reasons why we can't return an empty array?

app/assets/javascripts/admin/templates/web-hooks-show.hbs
tgxworld: Sure that is fine

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -0,0 +1,100 @@
...
+  urlValidation(url) {
+    // If blank, fail without a reason
+    if (Ember.isEmpty(url)) {
+      return Discourse.InputValidation.create({
+        failed: true
+      });
+    }
+
+    if (!Discourse.Utilities.urlValid(url) || url.indexOf(' ') !== -1) {
+      return Discourse.InputValidation.create({
```
tgxworld: Ah you missed some spots while rebasing `import InputValidation from 'discourse/models/input-validation';`

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -0,0 +1,81 @@
+module Jobs
+  class EmitWebHookEvent < Jobs::Base
+    sidekiq_options retry: 5
```
tgxworld: Hmm github basically let it fail and log it. Not sure if we need to retry it

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  it 'emits normally with sufficient arguments' do
+    Jobs::EmitWebHookEvent.any_instance.expects(:web_hook_request).once
+    subject.execute(web_hook_id: post_hook.id, event_name: 'post', post_id: post.id)
+  end
+
+  context 'with category filters' do
+    let!(:category) { Fabricate(:category) }
+    let!(:topic) { Fabricate(:topic) }
+    let!(:topic_with_category) { Fabricate(:topic, category_id: category.id) }
+    let!(:topic_hook) { Fabricate(:topic_web_hook, categories: [category]) }
```
tgxworld: I personally dislike `let!` because it hides the side effects for future developers adding tests to this context. Instead of "give me everything", I prefer "let me choose what I need for this test".

app/assets/javascripts/admin/models/web-hook.js.es6
fantasticfears: This is indeed a strange part. If `null` isn't there, it won't be present at payload which will result in failing to destroy associations. But I don't see other associations related properties have similar problem.

app/assets/javascripts/admin/components/admin-web-hook-event-chooser.js.es6
fantasticfears: They should always be an Ember object. `model`(array) and `type`(an event) are both Ember objects. After `filter`, `eventType` is not an Ember object any more. I can simply use the simple `.` here since there isn't an computed property.

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
fantasticfears: I am using the side effect of validation process to reset promoting text. If the model is saved once, the `Saved!` can be clear after some change is made in the form.

spec/jobs/emit_web_hook_event_spec.rb
fantasticfears: Indeed! While you ask me to change tests to use `let`, all tests are tend to be much clear as well as less side effect to other tests.

app/assets/javascripts/admin/models/web-hook.js.es6
fantasticfears: EDIT: catch this in https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/user.js.es6#L195-L197
Either we can change something based on `_ids` on ajax function or something else. This can be done after this pull request :P

PR #4257, opened 2016-06-09
PERF: Paginate public polls.

tgxworld:


tgxworld:
@eviltrout Can you review this? Thank you so much!

plugins/poll/assets/javascripts/components/poll-results-number-voters.js.es6
```diff
@@ -3,29 +3,14 @@ import User from 'discourse/models/user';
...
+  canLoadMore(voters, pollsVoters) {
+    return pollsVoters.length < voters;
   },
 
-  @computed("pollsVoters", "numOfVotersToShow")
-  showMore(pollsVoters, numOfVotersToShow) {
-    return !(Object.keys(pollsVoters).length < numOfVotersToShow);
+  @computed("poll.options", "offset")
+  voterIds(options) {
+    const ids = [].concat(...(options.map(option => { return option.voter_ids; })));
```
eviltrout: Trivial but you don't need the `{ return x }` in this case:

`const ids = [].concat(...(options.map(option => option.voter_ids)));`

plugins/poll/assets/javascripts/components/poll-voters.js.es6
```diff
@@ -1,13 +1,49 @@
...
 export default Ember.Component.extend({
   layoutName: "components/poll-voters",
   tagName: 'ul',
   classNames: ["poll-voters-list"],
   isExpanded: false,
-  numOfVotersToShow: 20,
+  numOfVotersToShow: 0,
+  offset: 0,
+  loading: false,
+  pollsVoters: [],
```
eviltrout: Initialzing a value to an array like this is probably not doing what you'd expect: Every instance of the `Component` class will have the same array!

In your case you are replacing it after the ajax call completes so bugs are unlikely, but it's a footgun and I think you should replace it with `pollsVoters: null`. If `pollsVoters` is only used in a template ember should deal with `null` safely, but if you need it to be an empty array put it in an `init()` function.

plugins/poll/assets/javascripts/components/poll-voters.js.es6
```diff
@@ -1,13 +1,49 @@
...
+    });
+  },
+
+  _getIds(ids) {
+    const numOfVotersToShow = this.get("numOfVotersToShow");
+    const offset = this.get("offset");
+    return ids.slice(numOfVotersToShow * offset, numOfVotersToShow * (offset + 1));
+  },
+
+  @on("didInsertElement")
```
eviltrout: I used to prefer `on('didInsertElement')` because it means you don't have to call `this._super()` but after talking with ember core I'm told it can't be optimized as well, so I've gone back to using `didInsertElement() { this._super(); .... `

plugins/poll/plugin.rb
```diff
@@ -214,11 +214,26 @@ def toggle_status
...
+    def voters
+      user_ids = params.require(:user_ids)
+
+      begin
+        users = User.where(id: user_ids).map do |user|
+          UserNameSerializer.new(user).serializable_hash
+        end
+
+        render json: { users: users }
+      rescue StandardError => e
```
eviltrout: This is a bit odd -- why would a serializer need a rescue?

plugins/poll/assets/javascripts/components/poll-results-number-voters.js.es6
tgxworld: :+1:

tgxworld:
Thanks for reviweing @eviltrout :+1:

PR #4258, opened 2016-06-09
Add notes to date and number formats to help translators

gschlager:
The comments show up in Transifex and will help translators finding the right format strings.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -18,7 +18,9 @@ en:
   js:
     number:
       format:
+        # symbol used to separate the integer part from the fractional part of a number
         separator: "."
+        # symbol for the thousands separator used in digit grouping
         delimiter: ","
       human:
         storage_units:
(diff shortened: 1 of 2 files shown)


config/locales/client.en.yml
```diff
@@ -35,18 +37,31 @@ en:
...
       full_with_year_no_time: "MMMM Do, YYYY"
+      # Use Moment.js format string: http://momentjs.com/docs/#/displaying/format/
       long_date_with_year: "MMM D, 'YY LT"
+      # Use Moment.js format string: http://momentjs.com/docs/#/displaying/format/
       long_date_without_year: "MMM D, LT"
+      # Use Moment.js format string: http://momentjs.com/docs/#/displaying/format/
       long_date_with_year_without_time: "MMM D, 'YY"
+      # Use Moment.js format string: http://momentjs.com/docs/#/displaying/format/
       long_date_without_year_with_linebreak: "MMM D <br/>LT"
+      # Use Moment.js format string: http://momentjs.com/docs/#/displaying/format/
```
tgxworld: Perhaps there is some where central we could place the notes? I don't think we should duplicate the same note so many times. :thought_balloon:

gschlager: I tried. Unfortunately that's not possible. The [Transifex documentation](http://docs.transifex.com/formats/yaml/) states:

> Transifex also supports developer comments in YAML files. A valid YAML comment preceding any of the elements will be saved as comment for the source string extracted from the element.

tgxworld:
Ok lets try this then :+1:

PR #4259, opened 2016-06-10
Update dependencies

tgxworld:


SamSaffron:
sure, it is correct to drop 2.0 support

tgxworld:
@SamSaffron I was wondering if we need to add a Changelog or something for this. People not on the docker related install might start running into this wall.

SamSaffron:
hmmm actually ... this is going to break UI rebuild on old versions ... we got to get that fixed asap @tgxworld 

UI rebuild should be banned on old versions of ruby, can you push that?

PR #4260, opened 2016-06-11
Replace certain uses of 'gsub' with 'tr' / 'chomp' for a speed improvement

jamescook:
String#tr is faster than String#gsub for single character substitutions: https://github.com/JuanitoFatas/fast-ruby#stringgsub-vs-stringtr-code

String#chomp is faster than String#[g]sub for replacing end of string characters: https://github.com/JuanitoFatas/fast-ruby#stringsub-vs-stringchomp-code


tgxworld:
Thanks @jamescook !

PR #4262, opened 2016-06-12
TopicStatusView: Convert statuses to property.

caugner:
Allows plugins (esp. `discourse-solved`) to add custom statuses to the topic list in the categories overview.

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-status.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-status.js.es6
@@ -25,35 +25,49 @@ export default Ember.Component.extend(StringBuffer, {
     return Discourse.User.current() && !this.get('disableActions');
   }.property('disableActions'),
 
-  renderString(buffer) {
-    const self = this;
+  statuses: function() {
+    const results = [];
 
-    const renderIcon = function(name, key, actionable) {
-      const title = Discourse.Utilities.escapeExpression(I18n.t(`topic_statuses.${key}.help`)),
-            startTag = actionable ? "a href" : "span",
-            endTag = actionable ? "a" : "span",
-            iconArgs = key === 'unpinned' ? { 'class': 'unpinned' } : null,
-            icon = iconHTML(name, iconArgs);
-
-      buffer.push(`<${startTag} title='${title}' class='topic-status'>${icon}</${endTag}>`);
+    const addStatus = (name, key, actionable) => {
+      results.push({
+        name: name,
+        key: `topic_statuses.${key}.help`,
+        actionable: actionable
+      });
     };
 
-    const renderIconIf = function(conditionProp, name, key, actionable) {
-      if (!self.get(conditionProp)) { return; }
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am not against this, but we need some tests here that at least exercise the new API and demo how it works.

caugner:
@SamSaffron I am demonstrating how it works in https://github.com/discourse/discourse-solved/pull/29 (which depends on this PR), but I'm willing to provide tests. Could you direct me to an existing test whose pattern I could adapt? I looked in `test/javascripts/components/`, but couldn't find any that tests a Ember component like this.

My idea would be the following:
```js
import TopicStatusComponent from 'discourse/components/topic-status';

TopicStatusComponent.reopen({
  statuses: function() {
    const results = this._super();
      results.push({
        name: 'bell',
        key: 'topic.bell',
      });
      return results;
  }.property()
});

// TODO Somehow call TopicStatusComponent.renderString().
// TODO Verify that result contains the bell icon.
```

caugner:
@SamSaffron Any ideas regarding my previous question?

coding-horror:
@eviltrout any suggestions here?

eviltrout:
I am not a fan of depending on plugins to reopen a class to add functionality. If a plugin has an `import` statement of one of our classes and we rename it for example, it will break the plugin in a hard way.

Instead this functionality should be added to the `pluginApi` as a method like:

```javascript
api.addTopicStatus(topic => {
  if (topic.get('alarm')) {
    return { name: 'bell', key: 'topic.bell' };
  }
});
```

caugner:
@eviltrout I agree that it would be better if the `pluginApi` would support this. I might be wrong, but doesn't this mean that topic status handling would need to be properly centralised for this to work (so that the same status icons appear everywhere, i.e. not only in lists – where it "works" – and in overviews – where my PR aimed at –, but also in other places, such as search results and email digests).

Then [this spot](https://github.com/discourse/discourse-solved/blob/master/assets/javascripts/discourse/initializers/extend-for-solved-button.js.es6#L196-L209) in the *discourse-solved* plugin could hopefully be refactored as well (which hooks in [topic-status.js.es6](https://github.com/discourse/discourse/blob/be3a5a56ccc284b352aa65080fc6b955f73cc72a/app/assets/javascripts/discourse/views/topic-status.js.es6#L5-L61) in the same way as my PR).

eviltrout:
Probably a good idea to centralize that logic, yes! Although we could start with making it work for your case and then add it for others later (using the API)

coding-horror:
Should this be closed as it is nearly 6 months old now?

caugner:
@coding-horror Sorry for the delay, I'm currently too busy to follow up on this, so closing is fine.

coding-horror:
We just upgraded to Ember 2.10 (latest, from 18 month old Ember 1.12) so a _lot_ of stuff changed on our end as well particularly w/r/t plugins.

PR #4263, opened 2016-06-12
usercard background now matches theme background (default = white)

awesomerobot:
screen shot 2016-06-12 at 7 47 34 pm
screen shot 2016-06-12 at 7 48 51 pm

⬜


coding-horror:
excellent! let's pull it in and try it out

coding-horror:
This is way nicer. Much less jarring than the black box popping up, so I agree with your instincts on this change.

PR #4264, opened 2016-06-13
Poll UI Builder.

tgxworld:
screenshot from 2016-06-14 16 26 54
screenshot from 2016-06-14 16 26 46
screenshot from 2016-06-14 16 26 38

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-editor.js.es6
@@ -362,7 +362,7 @@ export default Ember.Component.extend({
       this._resetUpload(true);
     },
 
-    showOptions() {
+    showOptions(toolbarEvent) {
       // long term we want some smart positioning algorithm in popup-menu
       // the problem is that positioning in a fixed panel is a nightmare
       // cause offsetParent can end up returning a fixed element and then
@@ -388,9 +388,8 @@ export default Ember.Component.extend({
         left = replyWidth - popupWidth - 40;
       }
 
-      this.sendAction('showOptions', { position: "absolute",
-                                       left: left,
-                                       top: top });
+      this.sendAction('showOptions', toolbarEvent,
+        { position: "absolute", left, top });
     },
 
     showUploadModal(toolbarEvent) {
@@ -420,7 +419,7 @@ export default Ember.Component.extend({
         sendAction: 'showUploadModal'
       });
 
-      if (this.get('canWhisper')) {
+      if (this.get("popupMenuOptions").some(option => option.condition)) {
         toolbar.addButton({
           id: 'options',
           group: 'extras',
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -42,6 +42,12 @@ function loadDraft(store, opts) {
   }
 }
 
+const _popupMenuOptionsCallbacks = [];
+
+export function addPopupMenuOptionsCallback(callback) {
+  _popupMenuOptionsCallbacks.push(callback);
+}
+
 export default Ember.Controller.extend({
   needs: ['modal', 'topic', 'application'],
   replyAsNewTopicDraft: Em.computed.equal('model.draftKey', Composer.REPLY_AS_NEW_TOPIC_KEY),
@@ -55,6 +61,20 @@ export default Ember.Controller.extend({
   isUploading: false,
   topic: null,
   linkLookup: null,
+
+  init() {
+    this._super();
+    const self = this
+
+    addPopupMenuOptionsCallback(function() {
+      return {
+        action: 'toggleWhisper',
+        icon: 'eye-slash',
+        label: 'composer.toggle_whisper',
+        condition: "canWhisper"
+      };
+    });
+  },
 
   showToolbar: Em.computed({
     get(){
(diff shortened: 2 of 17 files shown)


ZogStriP:
Looks good.

LGTM 👍

tgxworld:
Thanks :) still missing some cases :)

tgxworld:
@eviltrout Can you review this? Sorry that the PR is alittle large but half of it is taken up by test cases.

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -388,9 +388,8 @@ export default Ember.Component.extend({
         left = replyWidth - popupWidth - 40;
       }
 
-      this.sendAction('showOptions', { position: "absolute",
-                                       left: left,
-                                       top: top });
+      this.sendAction('showOptions', toolbarEvent,
+        { position: "absolute", left: left, top: top });
```
eviltrout: trivial but in ES6 you can just do `left, top` instead of `left: left, top:top`

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -56,6 +62,19 @@ export default Ember.Controller.extend({
...
 
+  init() {
+    this._super();
+
+    addPopupMenuOptionsCallback(function(controlller) {
+      return {
+        action: 'toggleWhisper',
+        icon: 'eye-slash',
+        label: 'composer.toggle_whisper',
+        condition: controlller.get("canWhisper")
```
eviltrout: I am a little worried to be honest about passing the entire `controller` through to the API.

I understand why you did this, but it means in the future to support this API we have to support passing the entire controller through, which means also letting people call actions on it or whatever crazy thing plugin authors decide to do.

I'd prefer if it was called with an object that contains the attributes the plugin author would want to use. 

For example `{ canWhisper: true}`

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -0,0 +1,156 @@
...
+  @computed("isNumber", "isMultiple")
+  showMinMax(isNumber, isMultiple) {
+    return isNumber || isMultiple;
+  },
+
+  @computed("pollOptions")
+  pollOptionsCount(pollOptions) {
+    if (pollOptions.length === 0) return 0;
+
+    var length = 0;
```
eviltrout: minor but this should be a `let`

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -0,0 +1,156 @@
...
+
+  @computed("isNumber", "pollMax")
+  pollStepOptions(isNumber, pollMax) {
+    if (!isNumber) return;
+    return this._comboboxOptions(1, parseInt(pollMax) + 1);
+  },
+
+  @computed("isNumber", "showMinMax", "pollName", "pollType", "publicPoll", "pollOptions", "pollMin", "pollMax", "pollStep")
+  pollOutput(isNumber, showMinMax, pollName, pollType, publicPoll, pollOptions, pollMin, pollMax, pollStep) {
+    var pollHeader = '[poll';
```
eviltrout: again, prefer `let`

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -0,0 +1,156 @@
...
+  },
+
+  @computed("pollOptionsCount", "isNumber")
+  disableInsert(count, isNumber) {
+    return isNumber ? false : (count < 2);
+  },
+
+  @computed("disableInsert")
+  minNumOfOptionsValidation(disableInsert) {
+    var options = { ok: true };
```
eviltrout: `let` or `const` here, the decision is up to you!

plugins/poll/assets/javascripts/discourse/templates/modals/poll-ui-builder.hbs
```diff
@@ -0,0 +1,58 @@
...
+        <label>{{i18n 'poll.ui_builder.poll_options.label'}}</label>
+        {{input-tip validation=minNumOfOptionsValidation}}
+        {{d-editor value=pollOptions}}
+      </div>
+    {{/unless}}
+  </form>
+</div>
+
+<div class="modal-footer">
+  {{d-button action="insertPoll" class='btn btn-primary' label='poll.ui_builder.insert' disabled=disableInsert}}
```
eviltrout: `d-button` shouldn't need the `btn` class. It has it by default. `primary` is still required though.

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -92,6 +111,25 @@ export default Ember.Controller.extend({
...
+        if (option.condition) {
+          option.condition = self.get(option.condition);
+        } else {
+          option.condition = true;
+        }
+
+        return option;
+      });
+    }
+  },
```
tgxworld: @eviltrout I updated to follow this pattern instead :) I'm not longer exposing the entire controller except for `self.get` since we need to allow plugin to add custom conditions as well.

```
Composer.reopen({
  canShow() { return true; }
})

api.addPopupMenuOptionsCallback(function() {
      return {
        action: 'someAction',
        icon: 'eye-slash',
        label: ''some label",
        condition: "canShow"
      };
});
```

eviltrout: Sure that seems better. In this case you don't need to use `self` though because `this` is set by the fat arrow `=>` syntax

tgxworld: O yes missed that one :) thank you!

PR #4265, opened 2016-06-13
Poll UX: Add icons to poll options.

caugner:
Goal: Make it more explicit that the options are clickable and not just text.


tgxworld:
See https://github.com/discourse/discourse/pull/4266#issuecomment-229237081. Thanks for the PR anyway! :+1:

caugner:
For the record: I have explained and referenced this PR here.

PR #4266, opened 2016-06-13
Poll UX: Highlight cast votes button and add icon.

caugner:
Goal: Make it clearer that one has to click the button to cast one's votes (in polls of `type=multiple`).

PR #4267, opened 2016-06-13
Poll UX: Remove danger from close poll button.

caugner:
Goal: Make the close poll button less distracting.

Justification: The close poll button is much more prominent than the cast votes button, even though it doesn't really pose any danger (which would justify its red colour), as closed polls can be re-opened.


tgxworld:
See https://github.com/discourse/discourse/pull/4266#issuecomment-229237081. Thanks for the PR though!

caugner:
For the record: I have explained and referenced this PR [here](https://meta.discourse.org/t/polls-ux-problems-and-solutions/46722).

PR #4268, opened 2016-06-14
enable args MODULE and FILTER for qunit:test rake task

rimian:
This enables running filters specs via `rake`

Like so:

`rake qunit:test MODULE='Acceptance: About'`
`rake qunit:test FILTER=foo`

(can be a combination)


SamSaffron:
Sure, looks good! :)

PR #4269, opened 2016-06-14
exit if RETRY is false

rimian:
It doesn't make a lot of sense to retry when you want to run filtered tests via rake. I created this flag. There is room for improvement but it works :)


SamSaffron:
usually I go for 0 for these kind of settings, but sure

PR #4272, opened 2016-06-16
FEATURE: Add details to popup menu options.

tgxworld:
out-12


app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -461,7 +461,7 @@ export default Ember.Component.extend({
 
         this.set('value', `${pre}${contents}${post}`);
         if (lines.length === 1 && tlen > 0) {
-          this._selectText(sel.start + hlen, contents.length - hlen - hlen);
+          this._selectText(sel.start + hlen, sel.value.length);
```
tgxworld: I'm not sure what the code was doing previously when selecting the text for apply surround. `- hlen - hlen` seems abit too magical and only works when the heal and tail is of equal length.

PR #4275, opened 2016-06-18
Support saml idp initiated sso

BrianAdams:
The current auth complete callback view assumes that either your doing SSO through Facebook or that you are doing SSO through a pop-up window from the main discourse login flow.  

There is another flow, at least using SAML, that allows the idp to have a registry of applications such that a user can click the icon and the browser opens a new window that transparently signs the user directly in to the application. Ultimately the SSO service is still calling the auth complete callback page, but in this case that page is not in a pop-up from the original discourse web page.

This change simply updates the auth complete view with a check to see if it is actually in a pop-up window, and if not, similar to the Facebook flow, simply set the local storage with the auth token and then redirect the page to the main discourse site.

The allows customers to setup discourse with SAML such that they can do idp-initiated SSO.

### DIFF
--- a/app/views/users/omniauth_callbacks/complete.html.erb
+++ b/app/views/users/omniauth_callbacks/complete.html.erb
@@ -26,9 +26,12 @@
 
        // On facebook browser, just redirect and don't close
        var ua = navigator.userAgent || navigator.vendor || window.opera;
-       if ((ua.indexOf("FBAN") > -1) || (ua.indexOf("FBAV") > -1) || (window.opener.Discourse == undefined)) {
+       if ((ua.indexOf("FBAN") > -1) || (ua.indexOf("FBAV") > -1)) {
          localStorage.setItem('lastAuthResult', JSON.stringify(authResult));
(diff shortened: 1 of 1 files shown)


SamSaffron:
I would like @eviltrout to review (he worked on the facebook hack afaik), but my call here is to nuke the facebook specific hack altogether and just use the general `if (!window.opener)` check for both cases.

eviltrout:
I think @SamSaffron's suggestion is a good one. If window.opener is not present it should always be safe to do that localStorage hack.

BrianAdams:
So... would you like me to do anything to help this move forward?  The facebook hack adds the authComplete=true flag that seems to trigger a forced page refresh that looked for state information that is unavailable outside of the facebook flow which is why I did not use the same redirect URL in the above patch.  (If needed I will fire up a test environment again to figure out what that was)

coding-horror:
There were some other changes to this per a meta topic. Look at recent @eviltrout mentions.

eviltrout:
Yeah we ended up making roughly the same change in master. Thanks for your input though!

PR #4279, opened 2016-06-20
FIX: get urlWithCDN before appending protocol

scossar:
Currently if the 's3 cdn url' is set audio and video uploads are being returned with 2 protocols (`http:http://cdn.example.com`)

This PR gets `Discourse.getURLWithCDN(url)` before appending the protocol. I've tested this with s3 uploads, with and without setting the `S3CDN` at constructivelearningspace.com.

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js
+++ b/app/assets/javascripts/discourse/lib/utilities.js
@@ -251,7 +251,8 @@ Discourse.Utilities = {
 
   uploadLocation: function(url) {
     if (Discourse.CDN) {
-      return Discourse.CDN.startsWith('//') ? "http:" + Discourse.getURLWithCDN(url) : Discourse.getURLWithCDN(url);
+      url = Discourse.getURLWithCDN(url);
+      return url.startsWith('//') ? 'http:' + url : url;
     } else if (Discourse.SiteSettings.enable_s3_uploads) {
       return 'https:' + url;
     } else {


app/assets/javascripts/discourse/lib/utilities.js
```diff
@@ -251,7 +251,8 @@ Discourse.Utilities = {
 
   uploadLocation: function(url) {
     if (Discourse.CDN) {
-      return Discourse.CDN.startsWith('//') ? "http:" + Discourse.getURLWithCDN(url) : Discourse.getURLWithCDN(url);
+      url = Discourse.getURLWithCDN(url);
+      return url.startsWith('//') ? 'http:' + url : url;
```
ZogStriP: Is it always "http"? Might want to check `use_https` site setting

scossar: I was going to ask about that :)

scossar:
> Might want to check use_https site setting

Is that available on the front end? It returns 'undefined' when I try it on meta. If not, is it ok to use `window.location.protocol`?

ZogStriP:
Yeah `window.location.protocol` is fine here.

SamSaffron:
regardless this is an improvement, I would also like to deprecate `//` as a CDN prefix longterm cause HTTP/2 is just too important.

PR #4280, opened 2016-06-20
Update Sidekiq command in DEVELOPER-ADVANCED

zachGlasgow:
We need to pass the queue names otherwise Sidekiq will use the 'default' name.
However Sidekiq will need to process Critical and Low tasks as well.


coding-horror:
:ok_hand:

PR #4281, opened 2016-06-20
FEATURE: add m4a to list of audio file extensions

scossar:
Allow m4a files to be oneboxed - will need to be added to `audio_onebox.rb` as well. Doing this makes it easy to record, upload and playback audio from a phone.


techAPJ:
Thanks! :+1:

PR #4282, opened 2016-06-21
Fix topic admin menu bottom offset. Calculate from #main's height.

nikdavis:
This was causing a small bug in short topics where #main doesn't span the whole document. The difference between #main's bottom and the document's bottom were pushing the menu above/away from the wrench.

You should be able to repro on a very short topic with a tall window. Just click on the admin wrench at the footer. Let me know if I'm off here.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/topic-admin-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/topic-admin-menu.js.es6
@@ -73,7 +73,7 @@ export default createWidget('topic-admin-menu', {
     const position = attrs.fixed ? 'fixed' : 'absolute';
 
     if (attrs.openUpwards) {
-      const bottom = $(document).height() - top - $('#main').offset().top;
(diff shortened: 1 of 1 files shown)


eviltrout:
I think I'm going to choose #4284 instead for this one - it seems to be a little more robust. Thanks though!

nikdavis:
👍

As an aside, if #main is intended to be the offset parent then would it make more sense to stop accounting for document vs #main top and bottom differences and just use #main's height? I realize I didn't stop accounting for the top's difference in my PR. I'm just curious as I'm sure I've something to learn here.

Feel free to close the PR, just didn't want the question to seem rhetorical.

tgxworld:
In the new PR, I used the difference between `document`  and `#main` to decide if we need to apply an offset for the menu. Otherwise, the menu will be overlapped by the header.

Thank you for looking into this :) I didn't see this PR when I submitted mine sorry! :crying_cat_face:

PR #4284, opened 2016-06-21
FIX: Topic admin menu item not positioned properly on short topics.

tgxworld:
cc @eviltrout 
166n99


ZogStriP:
There's also #4284 about this

PR #4285, opened 2016-06-21
FIX: Incorrect model for embedded post widget.

tgxworld:


tgxworld:
@eviltrout Please review :)

app/assets/javascripts/discourse/widgets/embedded-post.js.es6
```diff
@@ -27,6 +28,8 @@ export default createWidget('embedded-post', {
   buildKey: attrs => `embedded-post-${attrs.id}`,
 
   html(attrs, state) {
+    this.model = Post.create(attrs);
```
eviltrout: could you not call `getAncestorModel` here?

If that doesn't return a post and you have to make one, you should at least use `this.store.createRecord('post', attrs)` instead of importing the `Post` class.

tgxworld: `getAncestorModel` returns the post of the widget that the `embedded-post` widget is attached to.  Therefore, we end up getting the wrong model in cases where we call `helper.getModel()` in `decorateCooked` callbacks.

app/assets/javascripts/discourse/widgets/embedded-post.js.es6
```diff
@@ -27,6 +27,10 @@ export default createWidget('embedded-post', {
   buildKey: attrs => `embedded-post-${attrs.id}`,
 
   html(attrs, state) {
+    if (state.above) {
```
tgxworld: I updated this to only set the model when we embed post above a current one.

eviltrout: 👍

tgxworld:
Thanks for reviewing @eviltrout

PR #4288, opened 2016-06-25
Stop URLs from being censored

markwingerd:
URLs that contained a censored word were being altered by
censored-words.js and ulimately this broke the links. As an example
www.expertsexchange.com would get censored when it would link to a
legitimate website.  This URL blocking functionality should be
handled through other settings.


SamSaffron:
@eviltrout  your call here. 

On one hand I worry that this will introduce regressions cause regex based parsing of markdown is bound to have missing edge cases (there are like 3 ways of expressing a link and this is applied in pre processor)

On the other hand I really hate this feature and don't particularly care if it is perfect or not :)

eviltrout:
Seems fine to me!

markwingerd:
Awesome, thanks!

PR #4291, opened 2016-06-25
Set trim_trailing_whitespace false for markdown

sijad:


tgxworld:
What is the motivation behind this?

sijad:
in markdown two trailing space means a line break. beside that `trim_trailing_whitespace` already set as `true` in `.editorconfig`, redefining it with same value make no sense.

SamSaffron:
yeah this change is correct, our `.md` files are all visible on GitHub and it expects double space at the end for a newline

PR #4292, opened 2016-06-27
Rename `SiteSetting#use_https` to `force_https`.

tgxworld:


tgxworld:
@SamSaffron  Please review :grin:  Thanks :)

db/migrate/20160627104436_use_https_name_change_in_site_settings.rb
```diff
@@ -0,0 +1,13 @@
+class UseHttpsNameChangeInSiteSettings < ActiveRecord::Migration
+  def up
+    SiteSetting.transaction do
+      old_setting = SiteSetting.find_by(name: 'use_https')
```
SamSaffron: nahh 

you don't need the transaction there and do the whole change using raw sql, we get bitten long term with these kind of migrations

db/migrate/20160627104436_use_https_name_change_in_site_settings.rb
```diff
@@ -0,0 +1,8 @@
+class UseHttpsNameChangeInSiteSettings < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+    UPDATE site_settings SET value = (SELECT value FROM site_settings WHERE name = 'use_https') WHERE name = 'force_https';
+    DELETE FROM site_settings WHERE name = 'use_https';
```
tgxworld: @SamSaffron  Updated :) Please review the query in case just in case. Thanks!

db/migrate/20160627104436_use_https_name_change_in_site_settings.rb
```diff
@@ -0,0 +1,8 @@
+class UseHttpsNameChangeInSiteSettings < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+    UPDATE site_settings SET value = (SELECT value FROM site_settings WHERE name = 'use_https') WHERE name = 'force_https';
```
SamSaffron: update site_settings set name = 'use_https' where name = 'force_https' 

:)

PR #4294, opened 2016-06-28
PERF: Remove N+1 queries on user messages page.

tgxworld:
cc @SamSaffron

### DIFF
--- a/lib/topic_query.rb
+++ b/lib/topic_query.rb
@@ -355,10 +355,10 @@ class TopicQuery
       options = @options
       options.reverse_merge!(per_page: per_page_setting)
 
-      result = Topic
+      result = Topic.includes(:tags)
 
       if type == :group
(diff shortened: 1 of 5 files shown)


lib/topic_query.rb
```diff
@@ -355,10 +355,10 @@ def private_messages_for(user, type)
       options = @options
       options.reverse_merge!(per_page: per_page_setting)
 
-      result = Topic
+      result = Topic.includes(:tags)
 
       if type == :group
-        result = result.includes(:allowed_groups)
```
SamSaffron: hmm don't we need allowed_groups?

tgxworld: Hmm let me check. It was getting flagged as unused eager_loading

tgxworld: Yea we don't need it. I don't see it being used anywhere in the serializer

PR #4295, opened 2016-06-28
UX: Collapse users when composing a private message.

tgxworld:
https://meta.discourse.org/t/pm-with-many-users-hides-topic-input-field/6107/9

out-21
out-20


SamSaffron:
:heart:

PR #4296, opened 2016-06-28
Prevent creation of empty entry in _connectorCache for raw templates.

acshi:
The bug this fixes is described in:
https://meta.discourse.org/t/raw-template-plugin-outlet-use-can-lead-to-assertion-failure/46569


eviltrout:
This makes sense but shouldn't the same line be added before the `removeObject` call too in the `if` statement above?

acshi:
I had thought that the use of removeObject had already made the assumption that _connectorCache[outletName] was not empty, but I suppose when uniqueName appears in different outlets, it is possible for that assumption to be false. Change made.

eviltrout:
Thanks! I appreciate it.

PR #4297, opened 2016-06-29
Handle user enabled readonly mode

tgxworld:
cc @SamSaffron Did a second pass on this. It turns out that we require an expiring readonly mode more often than a permanent readonly mode.

### DIFF
--- a/app/controllers/admin/backups_controller.rb
+++ b/app/controllers/admin/backups_controller.rb
@@ -95,7 +95,7 @@ class Admin::BackupsController < Admin::AdminController
 
   def readonly
     enable = params.fetch(:enable).to_s == "true"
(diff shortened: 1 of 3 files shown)


config/initializers/005-site_settings.rb
```diff
@@ -7,6 +7,7 @@
   RailsMultisite::ConnectionManagement.each_connection do
     begin
       SiteSetting.refresh!
+      Discourse.activate_readonly
```
SamSaffron: hmmm ... can you explain this?

tgxworld: opps left over code from the previous branch. Basically, we create a thread to keep the site in readonly mode. On startup, we have to start the thread again. Otherwise, the site will fall out of readonly mode.

PR #4299, opened 2016-06-30
Add URL blocking functionality

markwingerd:
Admins can now add websites to a blocked list or add words that
should be blocked. Blocked sites can be added through the Settings >
Posting > Blocked URLs section.

### DIFF
--- a/app/assets/javascripts/discourse/dialects/censored_dialect.js
+++ b/app/assets/javascripts/discourse/dialects/censored_dialect.js
@@ -1,3 +1,4 @@
 Discourse.Dialect.addPreProcessor(function(text) {
+  text = Discourse.BlockedUrls.blockUrl(text);
   return Discourse.CensoredWords.censor(text);
 });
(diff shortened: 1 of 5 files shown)


coding-horror:
Can you add some screenshots of the UI here, also what it looks like to the user when their post is blocked due to URLs?

markwingerd:
blocked url 1
banned url 2

The first picture shows how you add urls or words to a blocked url list.  The second shows what the users will see along with the html that's generated.  The users see a link but no address pops up in the lower left hand corner. Discourse just adds a blank anchor tag.  

It would probably be more helpful adding some text to indicate that it is a blocked address.

Also, should I be concerned that the Travis CI build failed?

coding-horror:
1. This is a pure server side check, then? That'd be best.

2. The user has to see that the post was explicitly blocked after submitting, otherwise it'll generate "why isn't my link working" support requests. So, submit post, server checks blacklist, then pop a UI dialog explaining what's going on and why the post can't be saved.

SamSaffron:
closing this for now, agree with @coding-horror to accept a feature like this user needs some feedback that they are blocked from posting a URL, just gsubbing it to blank is very surprising. 

additionally this is operating on raw html as opposed parsed html so it has huge risk of false negatives.

PR #4300, opened 2016-06-30
Log RecordInvalid when verbose_sso_logging enabled

NuckChorris:
I find this to be *extremely* helpful in debugging.

Turns out we had some accounts whose emails hadn't been synced recently, and they were causing SSO to fail due to a uniqueness constraint on that column.

Wouldn't have figured that out if not for this patch.


SamSaffron:
sure seems good!

PR #4304, opened 2016-07-03
Add badge slug to class

Mittineague:
I used slug instead of name so it would be dasherized


SamSaffron:
this seems fine to me failures seem unrelated

PR #4308, opened 2016-07-05
FEATURE: Add keydown hook for widgets

gdpelican:
I'm looking to ensure that 'enter' isn't registered by a text area before submitting a form. So keyUp just won't doo!

cc @eviltrout


eviltrout:
Looks good to me thanks!

PR #4312, opened 2016-07-06
FEATURE: Add options to inject class / page into emoji toolbar

gdpelican:
I'm looking for a slightly more flexible emoji toolbar, which will let me inject some options to allow for limited emojisets for the [retort](www.github.com/gdpelican/retort) plugin.

Any objections to these?


eviltrout:
LGTM!

PR #4316, opened 2016-07-07
FIX: use strict_encode64 in sso

Sinewyk:
This avoids weird `\n` padding that all other languages do not do in their basic decode/encode 64 implementations.
possible ref https://stackoverflow.com/questions/2620975/strange-n-in-base64-encoded-string-in-ruby

Could be breaking if SSO consumers were looking for line feed(s) that MUST be here
for stripping them or something.

### DIFF
--- a/lib/single_sign_on.rb
+++ b/lib/single_sign_on.rb
@@ -30,7 +30,7 @@ class SingleSignOn
       end
     end
 
-    decoded = Base64.decode64(parsed["sso"])
+    decoded = Base64.strict_decode64(parsed["sso"])
     decoded_hash = Rack::Utils.parse_query(decoded)
 
     ACCESSORS.each do |k|
@@ -85,7 +85,7 @@ class SingleSignOn
   end
 
   def payload
-    payload = Base64.encode64(unsigned_payload)
+    payload = Base64.strict_encode64(unsigned_payload)
     "sso=#{CGI::escape(payload)}&sig=#{sign(payload)}"
   end
 


Sinewyk:
After making it work WITHOUT this PR. I want to say that this should be a non breaking change but would still be cool to merge because of testing reasons.

Depending how you test your controller or whatever, you can get bitten by the `\n` being interpreted (correctly) as a line feed (which get stripped from whatever decode64 implementation you use) or 2 characters `\` and `n` which breaks the signature check.

This difference of behavior disappear if there is no more linefeed in the output and we can test in whatever way we prefer without jumping through hoops.

PS: I signed the CLA, I don't know if you have a bot for that but ... it's done.

coding-horror:
Any thoughts on this @SamSaffron

eviltrout:
@Sinewyk Could you add some tests? They are really needed when changing functionality around logging in.

Sinewyk:
I wouldn't really know where to start on this. I'm very open to suggestions. Or even a patch I could apply :smile:

eviltrout:
Unfortunately it looks like there are no tests that pass in base64 data. @SamSaffron can you confirm and advise?

SamSaffron:
there are some end to end tests in place afaik, will double check tomorrow and add one, @Sinewyk  curious if you can give me an example of a payload that breaks without this fix?

Sinewyk:
It should never be a matter of "break". More like consistency. Apparently the base64 encode that adds the `\n` character every 60 characters AND at the end of the payload is legacy stuff.

It does not pose any problem because by default; all decode implementations (should) ignore characters outside the base64 alphabet.

So, it's for consistency and ease of testing. While testing in shell I was copy/pasting some payload generated by discourse and directly using it to query the backend. And `\n` was then interpreted as 2 characters and everything broke. This would never had happened had the payload been more simple. That's it ^^.

If you feel like this is too dangerous a change for consistency's/testing sake, please close and simply add some notes in the SSO documentation somewhere about base64 decode encode and discourse using the legacy interfaces generating `\n` characters that poses problem when passing these around.

NB: which is why actually using strict_decode may have been a mistake haha. We should just decode and not throw if we find something outside the alphabet, if a client is also using a legacy interface to encode the return payload.
> strict_decode64(str): Returns the Base64-decoded version of str. This method complies with RFC 4648. ArgumentError is raised if str is incorrectly padded or contains non-alphabet characters. Note that CR or LF are also rejected.

coding-horror:
I suggest adding comments to the documentation here instead of code, as you suggested

PR #4317, opened 2016-07-07
Put 👍 and 👎 emojis in emotional order

tnorthcutt:
Inspired by [Slack](https://twitter.com/SlackHQ/status/472141845703565312)


tnorthcutt:
I'm not 100% sure if this change will result in the desired behavior – if not, my apologies. I actually spent a good 30 minutes trying to get a local dev environment set up to test with, but ran into some roadblocks there, and realized the time I was devoting was disproportionate to the improvement... 😕 

But, hopefully the matcher just runs in the order the emoji names are defined in the file and this will result in the desired behavior 😄

eviltrout:
Lol @ emotional order :)

tnorthcutt:
@eviltrout oh wow, thanks for the quick merge! So was my assumption about the matching order correct, then?

PR #4318, opened 2016-07-08
UX: group pages should not show Messages tab to unauthorised users

techAPJ:
Meta topic: https://meta.discourse.org/t/group-pages-should-not-show-messages-tab-to-unauthorised-users/46946?u=techapj

Can you please review this PR @eviltrout?

### DIFF
--- a/app/assets/javascripts/discourse/controllers/group.js.es6
+++ b/app/assets/javascripts/discourse/controllers/group.js.es6
@@ -16,6 +16,13 @@ var Tab = Em.Object.extend({
 export default Ember.Controller.extend({
   counts: null,
   showing: 'members',
+  tabs: [
+    Tab.create({ name: 'members', active: true, 'location': 'group.index' }),
+    Tab.create({ name: 'posts' }),
+    Tab.create({ name: 'topics' }),
+    Tab.create({ name: 'mentions' }),
+    Tab.create({ name: 'messages', requiresMembership: true })
+  ],
 
   @observes('counts')
   countsChanged() {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/controllers/group.js.es6
```diff
@@ -34,11 +41,13 @@ export default Ember.Controller.extend({
...
-    Tab.create({ name: 'members', active: true, 'location': 'group.index' }),
-    Tab.create({ name: 'posts' }),
-    Tab.create({ name: 'topics' }),
-    Tab.create({ name: 'mentions' }),
-    Tab.create({ name: 'messages' }),
-  ]
+  @computed('model.is_member')
+  getTabs(isMember) {
+    if (isMember) {
+      if (this.get('tabs').length === 4) this.get('tabs').push(Tab.create({ name: 'messages' }));
```
eviltrout: I don't think checking the length of the tabs array and mutating it here is the safest approach. If we change the tabs these numbers would need to be updated and I think it would be easy for a developer to miss it.

Instead why not something like this:

```javascript
return this.get('tabs').filter(t => isMember || t.get('name') !== 'members');
```

Or even better, add `requiresMembership: true` to the `Tab.create` for members, then:

```javascript
return this.get('tabs').filter(t => isMember || !t.get('requiresMembership'));
```

That way we don't have to hard code the name.

techAPJ: ❤️ ❤️ ❤️ ❤️

eviltrout: Looks good to me now!

techAPJ: Thanks, merging in now. :)

PR #4319, opened 2016-07-10
FIX: backup was failing on large instances

techAPJ:
Backup was failing on an instance with ~10GB SQL dump. This PR fixes that issue. Weird fix, but works for me.™

@ZogStriP: Please review?


ZogStriP:
That is weird, so it means that it's better to first archive bigger files and then smaller files?

techAPJ:
Yes, that seems to be the case on my local dev machine.

PR #4322, opened 2016-07-12
Update server.ru.yml

BadAllOff:
Misleading translation, corrected.


erlend-sh:
Thanks for the fix! But I'm gonna have to ask you to submit it via Transifex:

https://www.transifex.com/discourse/discourse-org/translate/#ru/serverenyml/13698461?q=minimum

We update our translations in bulk from Transifex, so all translations must be submitted there.

PR #4325, opened 2016-07-13
FEATURE: Admins should be able to create polls even when plugin is di…

tgxworld:
…sabled.


eviltrout:
I'm not sure I agree with this. In terms of plugins "disabled" should mean "not available on this instance of Discourse." and not "available to some types of users." 

Instead it should be enabled but with a new setting like "Only staff can create polls"

PR #4328, opened 2016-07-14
FIX: poll builder should ignore empty lines

dachary:
Although pollOptionsCount skips empty lines, pollOutput inserts empty
lines. Skip them instead.

Signed-off-by: Loic Dachary <loic@dachary.org>


tgxworld:
This is awesome! Thanks!

PR #4331, opened 2016-07-14
FEATURE: pagination for webhooks

fantasticfears:


app/controllers/admin/web_hooks_controller.rb
```diff
@@ -2,9 +2,16 @@ class Admin::WebHooksController < Admin::AdminController
   before_filter :fetch_web_hook, only: %i(show update destroy)
 
   def index
-    # TODO: Pagination
+    limit = (params[:limit] || 50).to_i
```
tgxworld: params[:limit].to_i || 50

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -2,9 +2,16 @@ class Admin::WebHooksController < Admin::AdminController
...
+    offset = params[:offset].to_i + params[:limit].to_i
```
tgxworld: `limit` instead of `params[:limit]` ?

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -2,9 +2,16 @@ class Admin::WebHooksController < Admin::AdminController
...
+
+    total = WebHook.count
```
tgxworld: Don't think we need to assign this to a variable

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -2,9 +2,16 @@ class Admin::WebHooksController < Admin::AdminController
...
+    web_hooks = WebHook.limit(limit).offset(offset)
```
tgxworld: Same as https://github.com/discourse/discourse/pull/4331/files#r71067435

fantasticfears:
I fixed some magic trick and eager-loads related associations. It should be a bit less bizarre :)

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -2,9 +2,18 @@ class Admin::WebHooksController < Admin::AdminController
   before_filter :fetch_web_hook, only: %i(show update destroy)
 
   def index
-    # TODO: Pagination
+    limit = params[:limit].to_i
+    limit = 50 if limit <= 0 || limit > 200
```
tgxworld: hmmm why isn't the limitfixed?

fantasticfears: Shall I make it an constant at 50? I can only imagine an API client would try a different number of `limit` so I set an upper bound here

tgxworld: Yea just make it `50`

tgxworld:
👍 Thanks

PR #4334, opened 2016-07-16
Anchor is not necessary for avatars, change to span

idansh:
I recommend replacing the anchor '<a>' tag to a span '<span>' tag because anchor tells Google to pass page rank to user pages, which are not crawled anyways and those links are not linked to the user page, but used to trigger an overlay with user info.

I also removed the href property when applicable. In fact, some avatar links like the small ones in the topic first post don't even have an href property.

The event will still work because it is triggered on the 'trigger-user-card' selector.

I highly recommend that we make the change, a big boost for SEO with more importance to the topic pages, no page rank goes to waste.


SamSaffron:
closing cause the issue crawler view.

PR #4336, opened 2016-07-18
FEATURE: Add webhook event to track deliveries

fantasticfears:


app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -23,9 +46,14 @@ def execute(args)
...
     def web_hook_request
-      uri = URI(@web_hook.payload_url)
+      web_hook_event = WebHookEvent.create!(web_hook_id: @web_hook.id)
 
-      conn = Excon.new(uri.to_s, ssl_verify_peer: @web_hook.verify_certificate)
+      uri = URI(@web_hook.payload_url)
+      conn = Excon.new(uri.to_s,
+                       ssl_verify_peer: @web_hook.verify_certificate,
+                       retry_limit: 0,
+                       instrumentor: ExconWebHookEventInstrumentor,
```
tgxworld: I'm not really a fan of this implementation as it couples the logs to Excon. If we eventually decide to not use Excon, this portion of the code will break. I rather we check the response and create an entry in the log.

fantasticfears:
@tgxworld Did a rebase. Continue from your comment: Took some time in this. Instrumentor gives us full life cycle in the implementation details. Without the instrumentor, we can't find out the precise headers sent by Excon. But response itself is returned as an object. Extracting response is easy by all means. I don't want to lose the information which might be useful for debugging. In this case, the benefit may not be significant.

db/migrate/20160708235959_add_web_hook_event.rb
```diff
@@ -0,0 +1,18 @@
+class AddWebHookEvent < ActiveRecord::Migration
+  def change
+    enable_extension 'uuid-ossp' unless extension_enabled?('uuid-ossp')
+
+    create_table :web_hook_events, id: :uuid, default: 'uuid_generate_v1mc()' do |t|
```
tgxworld: Hmm Why do we need to use an `uuid` here?

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -54,4 +54,19 @@
                       category_id: topic_with_category.category.id)
     end
   end
+
+  describe '.web_hook_request' do
+    before(:all) { Excon.defaults[:mock] = true }
```
tgxworld: `before` will probably be sufficient?

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -54,4 +54,19 @@
...
+    after(:all) { Excon.defaults[:mock] = false }
```
tgxworld: Same as https://github.com/discourse/discourse/pull/4336/files#r71501442

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -54,4 +54,19 @@
...
+    after(:each) { Excon.stubs.clear }
+
+    it 'uses excon instrumentor to create delivery event record' do
```
tgxworld: No more instrumentor? :stuck_out_tongue:

spec/models/web_hook_event_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+describe WebHookEvent do
+  let(:web_hook) { Fabricate(:web_hook) }
+  let(:event) { WebHookEvent.new(web_hook_id: web_hook.id, response_code: 200) }
```
tgxworld: hmm can't we use a fabricator for this?

spec/jobs/emit_web_hook_event_spec.rb
fantasticfears: oops, pushed wrong branch..

db/migrate/20160708235959_add_web_hook_event.rb
fantasticfears: no lock?

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -46,6 +50,22 @@ def web_hook_request
...
+
+      _, id, event_name = name.split('.')
+      web_hook_event = WebHookEvent.find(id)
+
+      case event_name
+        when 'request'
+          web_hook_event.header = params[:headers].to_s
+          web_hook_event.payload = params[:body].to_s
+        when 'response', 'error'
+          web_hook_event.response_code = params[:status].to_i
```
tgxworld: `update_attributes!` will allow you to avoid all the `web_hook_event` repetition

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -46,6 +50,22 @@ def web_hook_request
       end
 
       conn.post(headers: headers, body: body)
+
+      _, id, event_name = name.split('.')
```
tgxworld: Hmm I'm not sure where `name` is declared.

db/migrate/20160708235959_add_web_hook_event.rb
tgxworld: I'm not following... what lock? :thought_balloon:

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -46,6 +50,22 @@ def web_hook_request
...
+      web_hook_event = WebHookEvent.find(id)
```
tgxworld: Why do we create an instance of `web_hook_event` at the start and retrieve it later? I think we should only create the `web_hook_event` after `POST`ing the results

db/migrate/20160708235959_add_web_hook_event.rb
fantasticfears: I expect several web hooks may work at the same time with a long period. `UUID` may be easier to distinguish with each others. It can also be generated in parallel

tgxworld: hmm we have alot of posts being saved in parallel as well. Don't think we need a uuid in this case. I've never used this so I might be missing something important :stuck_out_tongue:

db/migrate/20160708235959_add_web_hook_event.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.belongs_to :web_hook, null: false, index: true
+      t.string     :headers
+      t.string     :payload
+      t.integer    :response_code
+      t.string     :response_headers
+      t.string     :response_body
+      t.integer    :duration_in_milliseconds, default: 0
```
tgxworld: `duration` will be fine here :)

db/migrate/20160708235959_add_web_hook_event.rb
```diff
@@ -0,0 +1,18 @@
+class AddWebHookEvent < ActiveRecord::Migration
+  def change
+    enable_extension 'uuid-ossp' unless extension_enabled?('uuid-ossp')
+
+    create_table :web_hook_events, id: :uuid, default: 'uuid_generate_v1mc()' do |t|
+      t.belongs_to :web_hook, null: false, index: true
+      t.string     :headers
+      t.string     :payload
```
tgxworld: `string` might be too short to store the payload?

fantasticfears: ehh, I'm used to varchar in PostgreSQL. But I will change that to text anyway

db/migrate/20160708235959_add_web_hook_event.rb
fantasticfears: Not sure how it's been done. I expect `INSERT` needs to at least lock sequence to query the next viable id. Maybe it's cached.

Go back to our topic. Say, with a daily active users 500, a wildcard webhook could yield 1500 times. A numeric id is not dramatic different than others. The table in the webhook page shows a subsequent requests, in which UUID looks better than an numeric value as they looks different and there aren't numbered as a sequence. If there is a need for scrub, old id won't be just disappeared.

Technically, there aren't much different than an numeric incremental ID. It has a little to none chance of collision thanks to secure hash function. And `uuid_generate_v1mc` is time based so it could be used for index.

After that, speaking to Discourse instance:

UUID: good visual, don't need a sequence
Number: bad visual, leave a gap after scrub

spec/fabricators/web_hook_event_fabricator.rb
```diff
@@ -0,0 +1,5 @@
+Fabricator(:web_hook_event) do
+  web_hook
+  status 200
+end
+
```
tgxworld: :scissors:

spec/jobs/emit_web_hook_event_spec.rb
tgxworld: I think the description needs to be change here.

tgxworld:
@fantasticfears tests are failing on ESLint errors

fantasticfears:
Just checked in some changes. Still missing formatting stuff / some direction button, but the feature is there. By the way, do we have formatting helpers for JSON?

I renamed `event_name` to `event_type` as it's much precise. I feel it may be even more useful to include exact hook name in the headers. Though I didn't figure the exact way to do that.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -24,28 +25,43 @@ def execute(args)
...
-      body = build_web_hook_body
+      headers['X-Discourse-Event'] = @opts[:event_name] if @opts[:event_name].present?
 
       if @web_hook.secret.present?
-        headers['X-Discourse-Event-Signature'] = OpenSSL::HMAC.hexdigest("sha256", @web_hook.secret, body)
+        headers['X-Discourse-Event-Signature'] = "sha256=" + OpenSSL::HMAC.hexdigest("sha256", @web_hook.secret, body)
       end
 
-      conn.post(headers: headers, body: body)
+      now = Time.now
```
tgxworld: `Time.zone.now` instead.

tgxworld:
>  do we have formatting helpers for JSON?

What kind of formatting do you need?

fantasticfears:
Would be enough to get this from raw string. http://jsfiddle.net/unlsj/

tgxworld:
O you mean on the client side? If so we probably don't have any

spec/models/web_hook_event_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+describe WebHookEvent do
+  let(:event) { Fabricate(:web_hook_event) }
+  let(:failed_event) { Fabricate(:web_hook_event, status: 400) }
+
+  it 'update last delivery status for associated WebHook record' do
+    event.save!
```
tgxworld: Why do we need to `save!` here? There isn't any modification to the object.

spec/fabricators/web_hook_fabricator.rb
```diff
@@ -28,3 +28,9 @@
     web_hook.web_hook_event_types = [transients[:topic_hook]]
   end
 end
+
+Fabricator(:web_hook_event) do
+  web_hook
+  status 200
+end
+
```
tgxworld: :scissors:

tgxworld:
Thanks @fantasticfears I'm merging this in to test it out

PR #4339, opened 2016-07-20
Remove list serializers and obselete web_hooks field

fantasticfears:


db/migrate/20160720095915_delete_status_from_web_hook.rb
```diff
@@ -0,0 +1,5 @@
+class DeleteStatusFromWebHook < ActiveRecord::Migration
+  def change
+    remove_column :web_hooks, :status
```
tgxworld: Since we're still developing on this branch, Just remove the `add` statement in the original migration

tgxworld:
Thanks @fantasticfears  :)

PR #4340, opened 2016-07-22
update lang vi-VN

quoctrung117:


coding-horror:
is this the correct place to put this @techAPJ ?

techAPJ:
@quoctrung117 Vietnamese locale files are already present in Discourse source code.

If you want to update the translations, please do it from [Transifex](https://www.transifex.com/discourse/discourse-org/language/vi_VN/).

PR #4341, opened 2016-07-22
FEATURE: Work with compressed version of `pg_dump` during backup and …

tgxworld:
…restore.

### DIFF
--- a/lib/backup_restore/backuper.rb
+++ b/lib/backup_restore/backuper.rb
@@ -1,6 +1,7 @@
 module BackupRestore
 
   class Backuper
+    include BackupRestore::Utils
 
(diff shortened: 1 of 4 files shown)


lib/backup_restore/backuper.rb
```diff
@@ -192,81 +191,58 @@ def pg_dump_command
...
       ].join(" ")
     end
 
-    def update_dump
-      log "Updating dump for more awesomeness..."
-
-      `#{sed_command}`
-    end
-
-    def sed_command
```
tgxworld: Let's not mess with the valid copy of the `pg_dump` file. We'll do the `sed`ing when restoring.

PR #4342, opened 2016-07-22
Allow localhost as an embeddable host

acshi:
Hostnames containing localhost do not currently pass embeddable host validations. This prevents the user from embedding comments into a page hosted on localhost.

This fix causes these hostnames to pass validation.


eviltrout:
This should also have a spec to ensure localhost works.

acshi:
How does that simple spec look? I notice that the existing spec is not very thorough.

Also, is there any reason the regex are set to accept paths when those are already being stripped from the host before validation?

acshi:
@eviltrout Any thoughts on this or need for further changes?

eviltrout:
No, thank you very much!

PR #4343, opened 2016-07-22
FEATURE: configure session time via site setting for all the users

techAPJ:


coding-horror:
are the database changes related? they don't look related?

techAPJ:
@coding-horror they are just annotated, using [annotate](https://github.com/ctran/annotate_models) gem.

There is a minor database change as per @SamSaffron's feedback [here](https://meta.discourse.org/t/session-timeout/34686/24?u=techapj).

coding-horror:
:+1:

SamSaffron:
We need to improve this some. 

Expiration by default should be a "sliding window" thing, not an absolute time from login, its hugely weird to log off a user in the middle of posting, will cause lots of support issues. 

If a user is active on the site window should slide.

coding-horror:
If we do add that feature, we can reduce default cookie life to 30 days, so the user will be "logged off" 30 days after the time we saw them last.

NyroTheCat:
I'am very sad that the `permanent_session_cookie` got removed. We are running many public workstations in our reading room (library). so if the user closes the browser and leaves the workstation the session will be still active and available for the following user.

This was an important site setting as @xfalcox mentioned in #4226 
>>Er.. what? I'm not following, how is it a good thing to never use cookies? So users have to log in every time they visit the site? At the very least this needs much clearer copy.
>
> If an admin turn off the setting the cookie is still there, but it doesn't survive a computer restart or closing all the tabs.
>
>If the user turn off a pc, in some environments, he expects to be logged out of everything. Like public libraries, pc bangs and companies pics.
>
>Also, I recommend using this alongside SSO so the user doesn't even know what's happening.

xfalcox:
Hey @NyroTheCat I recommend stating a topic on meta.discourse.org so we can track this feature better.

PR #4345, opened 2016-07-25
Fix typos in descriptions of server source file

asdofindia:
I used hunspell backend with linter-spell plugin of atom to find these typos. There are several capitalizations issues that I've left untouched (pop3, POP3, etc)

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1063,7 +1063,7 @@ en:
     avatar_sizes: "List of automatically generated avatar sizes."
 
     external_system_avatars_enabled: "Use external system avatars service."
-    external_system_avatars_url: "URL of the external system avatars service. Allowed substitions are {username} {first_letter} {color} {size}"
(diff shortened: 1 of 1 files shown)


config/locales/server.en.yml
```diff
@@ -1128,7 +1128,7 @@ en:
     body_min_entropy: "The minimum entropy (unique characters, non-english count for more) required for a post body."
     allow_uppercase_posts: "Allow all caps in a topic title or a post body."
 
-    title_fancy_entities: "Convert common ASCII characters to fancy HTML entities in topic titles, ala SmartyPants http://daringfireball.net/projects/smartypants/"
```
ZogStriP: `ala` is right here.

asdofindia: Okay, let me amend the commit then. But, what does "ala" mean here?

asdofindia: Ah, "ala" as a colloquial form of "[a la](https://en.wiktionary.org/wiki/a_la#Alternative_forms)". Hmm.

coding-horror:
I reviewed and these look good -- thanks!

PR #4351, opened 2016-07-28
FEATURE: Allow poll results to be sorted.

tgxworld:
out-2


coding-horror:
I don't think we need a button here.. less UI is better.. when you click/tap to view the results, always re-order by the winners at the top.

PR #4355, opened 2016-07-29
Suggest change 'cause' to 'because' for the reason that it is better English grammar

pacharanero:
Apologies if this seems a trivial pull request, but I recently noticed that the text in the footer of email I am getting as Admin of my Discourse instance says "You are receiving these emails cause you have enabled mailing list mode". I would suggest that this should say "You are receiving these emails because you have enabled mailing list mode", which is regarded as better English grammar and more professional sounding. Here is the amendment as a pull request. I have signed the Discourse CLA.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -2184,7 +2184,7 @@ en:
     To unsubscribe from these emails, [click here](%{unsubscribe_url}).
 
   unsubscribe_mailing_list: |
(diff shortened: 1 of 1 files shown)


coding-horror:
I like this but I wanted to update copy a bit more so I made this edit:

> You are receiving this because you enabled mailing list mode.

For example this is what GitHub says:

PR #4364, opened 2016-08-03
FEATURE: Group filters for webhook

fantasticfears:
We already have a `category-chooser` so that I unified them to `xx_filter`.


fantasticfears:
Found `group-selector`. It seems not working. I will rename related components to `xx_selector`

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -22,8 +24,18 @@ export default RestModel.extend({
...
-  updateCategoryFilters() {
-    this.set('categoryFilters', Category.findByIds(this.get('category_ids')));
+  updateCategoriesFilter() {
+    this.set('categoriesFilter', Category.findByIds(this.get('category_ids')));
+  },
+
+  @observes('group_ids')
+  updateGroupsFilter() {
+    const groupIds = this.get('group_ids');
+    this.set('groupsFilterInName', Discourse.Site.currentProp('groups').filter(g => groupIds.includes(g.id)).map(g => g.name));
```
tgxworld: `filter` and `map` can probably be reduced to just a single loop with a condition

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -40,20 +52,20 @@ export default RestModel.extend({
...
       payload_url: this.get('payload_url'),
       content_type: this.get('content_type'),
       secret: this.get('secret'),
       wildcard_web_hook: this.get('wildcard_web_hook'),
       verify_certificate: this.get('verify_certificate'),
       active: this.get('active'),
       web_hook_event_type_ids: Ember.isEmpty(types) ? [null] : types.map(type => type.id),
-      category_ids: Ember.isEmpty(categories) ? [null] : categories.map(c => c.id)
+      category_ids: Ember.isEmpty(categories) ? [null] : categories.map(c => c.id),
+      group_ids: Ember.isEmpty(groupNames) ? [null] : Discourse.Site.currentProp('groups').filter(g => groupNames.includes(g.name)).map(g => g.id)
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/4364/files#r73994684

app/assets/javascripts/discourse/components/category-group.js.es6
```diff
@@ -1,37 +0,0 @@
-import { categoryBadgeHTML } from 'discourse/helpers/category-link';
```
tgxworld: Hmm I don't quite follow this change. Why does it have to be renamed?

app/assets/javascripts/discourse/templates/components/group-selector.hbs
```diff
@@ -1 +0,0 @@
-<input class='ember-text-field group-names' type="text" placeholder={{placeholder}} name="groups">
```
tgxworld: Why is the template removed?

app/assets/javascripts/discourse/components/category-group.js.es6
fantasticfears: I saw other autocomplete components are using `-selector` so I think it's better to keep them consistent. Others are `user-selector` and `group-selector`

app/assets/javascripts/admin/models/web-hook.js.es6
fantasticfears: Switched to a `reduce`

app/assets/javascripts/discourse/templates/components/group-selector.hbs
fantasticfears: It's a one line template. So I replaced it with `StringBuffer`

tgxworld: I will keep the template since it already exists plus it is easier to add stuff to a template than a StringBuffer

fantasticfears: Fixed! None of CSS targets original class names so it's safe to change those

fantasticfears:
Looks like a random error in redis failed CI

app/assets/javascripts/admin/models/web-hook.js.es6
```diff
@@ -9,7 +10,8 @@ export default RestModel.extend({
   verify_certificate: true,
   active: false,
   web_hook_event_types: null,
```
tgxworld: Hmm just realized these are using snake case. Can you fix the rest as well?

fantasticfears: Those properties are from serializer. So I guess it can be just like this now

PR #4365, opened 2016-08-03
Don't halt notification emails for those on daily mailing list mode

gdpelican:
As discussed here: https://meta.discourse.org/t/daily-updates-option-for-mailing-list-mode/45029


eviltrout:
Thanks, looks good!

PR #4366, opened 2016-08-04
Print Support

xfalcox:
As per https://meta.discourse.org/t/print-long-topic-to-pdf-redux-again/44639

Please review, and be kind :smile:


lib/topic_view.rb
```diff
@@ -37,14 +41,19 @@ def initialize(topic_id, user=nil, options={})
...
 
     options.each do |key, value|
       self.instance_variable_set("@#{key}".to_sym, value)
     end
 
     @page = 1 if (!@page || @page.zero?)
-    @chunk_size = options[:slow_platform] ? TopicView.slow_chunk_size : TopicView.chunk_size
+    @chunk_size = case
+                    when options[:slow_platform] then TopicView.slow_chunk_size
+                    when @print then 1000
```
ZogStriP: Shouldn't this be `when @print then TopicView.print_chunk_size`?

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -134,6 +134,10 @@ const Topic = RestModel.extend({
     return this.get('url') + (user ? '?u=' + user.get('username_lower') : '');
   }.property('url'),
 
+  printUrl: function(){
+    return this.get('url') + '/print';
+  }.property('url'),
```
tgxworld: We're prefer to use `@computed('url')` :)

app/assets/javascripts/discourse/views/topic-footer-main-buttons.js.es6
```diff
@@ -7,6 +7,7 @@ export default ContainerView.extend({
   @on('init')
   createButtons() {
     const mobileView = this.site.mobileView;
+    const { siteSettings } = this;
```
tgxworld: Not sure if this variable is needed right now since we're only fetching from siteSettings in a single place.

lib/topic_view.rb
```diff
@@ -5,13 +5,17 @@
...
 
-  attr_reader :topic, :posts, :guardian, :filtered_posts, :chunk_size
+  attr_reader :topic, :posts, :guardian, :filtered_posts, :chunk_size, :print
   attr_accessor :draft, :draft_key, :draft_sequence, :user_custom_fields, :post_custom_fields
 
   def self.slow_chunk_size
     10
   end
 
+  def self.print_chunk_size
```
tgxworld: This and the surround methods should be a constant IMO.

tgxworld:
Tests? :stuck_out_tongue_closed_eyes: k_out_tongue:

app/controllers/topics_controller.rb
```diff
@@ -58,8 +59,18 @@ def show
     username_filters = opts[:username_filters]
 
     opts[:slow_platform] = true if slow_platform?
+    opts[:print] = true if params[:print].present?
     opts[:username_filters] = username_filters.split(',') if username_filters.is_a?(String)
 
+    if opts[:print]
+      raise Discourse::InvalidAccess unless SiteSetting.allow_print_page
+      begin
+        RateLimiter.new(current_user, "print-topic-per-hour", 10, 1.hour).performed! unless current_user.admin?
```
tgxworld: Hmm 10 prints per hour should probably be a site setting?

tgxworld: `current_user.admin?` should be `@guardian.is_admin?`

xfalcox: What about replacing the enable site setting by a numeric one and 0 = disabled?

tgxworld: Ah I meant a site setting for the rate limit haha

xfalcox: Exactly this! A site setting for rate limit (max prints per hour, default 0) and if max == 0 it disables the feature and removes the button.

xfalcox:
Please review again :smile:

app/controllers/topics_controller.rb
```diff
@@ -58,8 +59,18 @@ def show
     username_filters = opts[:username_filters]
 
     opts[:slow_platform] = true if slow_platform?
+    opts[:print] = true if params[:print].present?
     opts[:username_filters] = username_filters.split(',') if username_filters.is_a?(String)
 
+    if opts[:print]
+      raise Discourse::InvalidAccess unless SiteSetting.max_prints_per_hour_per_user > 0
+      begin
+        RateLimiter.new(current_user, "print-topic-per-hour", 10, 1.hour).performed! unless @guardian.is_admin?
```
tgxworld: I think 10  needs to be replaced by `max_prints_per_hour_per_user`?

config/site_settings.yml
```diff
@@ -885,6 +885,9 @@ rate_limits:
     client: true
     shadowed_by_global: true
     default: 0
+  max_prints_per_hour_per_user:
+    default: 0
```
tgxworld: Hmm not sure if this is a good default because it'll raise `Discourse::InvalidAccess`

lib/topic_view.rb
```diff
@@ -113,22 +122,26 @@ def next_page
...
+    "#{Discourse.base_url}#{relative_url}"
   end
 
   def relative_url
-    @topic.relative_url
+    if !@print
+      @topic.relative_url
+    else
+      @topic.relative_url + '/print'
+    end
```
tgxworld: `"#{@topic_relative_url}#{@print ? '/print' : ''}"`? xD

spec/controllers/application_controller_spec.rb
```diff
@@ -84,6 +84,28 @@ def set_accept_language(locale)
 
   end
 
+  describe "print" do
+    render_views
+
+    context "when the SiteSetting is enabled" do
+      before do
+        SiteSetting.stubs(:max_prints_per_hour_per_user).returns(10)
```
tgxworld: You can set the site settings directly without having to stub :)

spec/controllers/topics_controller_spec.rb
```diff
@@ -738,6 +738,22 @@ def topics_controller_show_gen_perm_tests(expected, ctx)
...
+      user = Fabricate(:user)
+      get :show, topic_id: topic.id, slug: topic.slug, print: true
+
+      expect(response).to be_forbidden
+    end
+
+    it 'renders the print view when enabled' do
+      SiteSetting.max_prints_per_hour_per_user = 10
+
+      user = Fabricate(:user)
```
tgxworld: You can probably extract this to `let(:user) { Fabricate(:user) }` so that it can be shared between both tests.

spec/controllers/application_controller_spec.rb
```diff
@@ -84,6 +84,28 @@ def set_accept_language(locale)
...
+      end
+
+      it "uses the application layout when there's no param" do
+        get :show, topic_id: topic.id, slug: topic.slug
+        expect(response).to render_template(layout: 'application')
```
tgxworld: `render_template` is still supported by RSpec but deprecated in Rails 5 though. The recommended way is to write an integration test and just assert against the response's body directly.

coding-horror:
So,the final state is, printing is a keyboard shortcut exclusively? Maybe we should just go with that

xfalcox:
Rebased and good to go!

PR #4370, opened 2016-08-06
FEATURE: Allow keyboard shortcuts for topic list to start from last viewed topic

cpradio:
Relevant Discussion
https://meta.discourse.org/t/keyboard-navigation-shortcuts-dont-remember-topic-list-position-on-return/48138?u=cpradio

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-list-item.js.es6
@@ -113,11 +113,12 @@ export default Ember.Component.extend(StringBuffer, {
     }
   },
 
-  highlight() {
+  highlight(isLastViewedTopic = false) {
     const $topic = this.$();
     const originalCol = $topic.css('backgroundColor');
     $topic
       .addClass('highlighted')
+      .attr('data-islastviewedtopic', isLastViewedTopic)
       .stop()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -128,7 +129,7 @@ export default Ember.Component.extend(StringBuffer, {
     // highlight the last topic viewed
     if (this.session.get('lastTopicIdViewed') === this.get('topic.id')) {
       this.session.set('lastTopicIdViewed', null);
-      this.highlight();
+      this.highlight(true);
```
eviltrout: I've grown to dislike booleans like this as parameters. Without looking at the signature it's not obvious what the `true` is for.

I think it would be better to add an opts object like:

`this.highlight({ isLastViewedTopic: true });`

(as an aside, if the function was like `setGreen(true)` I'd have no problem with it. It's obvious it means that it's `green` in that case!)

cpradio:
Ah, good idea. I'll use the opts approach. :)

cpradio:
Using opts now added :)

eviltrout:
Thanks, got it!

PR #4371, opened 2016-08-07
FEATURE: ping for testing webhooks

fantasticfears:


app/controllers/admin/web_hooks_controller.rb
```diff
@@ -92,6 +92,38 @@ def redeliver_event
...
+      'Host' => uri.host,
+      'User-Agent' => "Discourse/" + Discourse::VERSION::STRING,
+      'X-Discourse-Event-Id' => 'null',
+      'X-Discourse-Event-Type' => 'ping'
+    }
+
+    if @web_hook.secret.present?
+      headers['X-Discourse-Event-Signature'] = "sha256=" + OpenSSL::HMAC.hexdigest("sha256", @web_hook.secret, body)
+    end
+    conn.post(headers: headers, body: body)
```
tgxworld: Hmm it doesn't seem like we are checking the response to see if the ping fails?

app/assets/javascripts/admin/controllers/admin-web-hooks-show-events.js.es6
```diff
@@ -1,7 +1,19 @@
...
+import { popupAjaxError } from 'discourse/lib/ajax-error';
+
 export default Ember.Controller.extend({
   actions: {
     loadMore() {
       this.get('model').loadMore();
+    },
+
+    ping() {
+      return bootbox.confirm(I18n.t('admin.web_hooks.events.ping_instruction'), I18n.t('no_value'), I18n.t('yes_value'), result => {
```
tgxworld: A ping event is pretty safe so I don't think we need to ask for confirmation.

tgxworld:
Aside from some minor comments. This PR looks really good to me

app/controllers/admin/web_hooks_controller.rb
fantasticfears: Nope, it should be used to test whether the configuration/network is functional.

tgxworld: If it isn't functional.. How will they know whether it is their configuration or a bug in the code?

fantasticfears: Fixed 👍

PR #4372, opened 2016-08-07
FEATURE: Add PluginApi#addTopicStatusIcon

caugner:
Note: This is a follow-up of PR #4262.

**Rationale**: A Discourse plugin may add additional state to topics. In order to visualize this additional state with an icon, there needs to be a way for the plugin to make Discourse render an additional status icon for a given topic. This should be achievable programmatically through the plugins API.

**Example**: The plugin [discourse-solved](https://github.com/discourse/discourse-solved) adds the notion of a topic having a solution (or not), which is symbolized by a check icon.

### DIFF
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -12,6 +12,7 @@ import { h } from 'virtual-dom';
 import { addFlagProperty } from 'discourse/components/site-header';
 import { addPopupMenuOptionsCallback } from 'discourse/controllers/composer';
 import { emojiUrlFor } from 'discourse/lib/text';
+import { escapeExpression } from 'discourse/lib/utilities';
 
 class PluginApi {
   constructor(version, container) {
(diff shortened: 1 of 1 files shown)


eviltrout:
I like this!

Two things though:

- This seems to share a fair bit of code with `addPosterIcon` - is there any way it could be made more DRY? Perhaps a shared function to do the bulk of the work that is the same.

- We'll need to bump the plugin api version too as this function is new. You can do this by changing `0.5` to `0.6` in `getPluginApi` and having your plugin use `0.6`.

caugner:
> This seems to share a fair bit of code with addPosterIcon - is there any way it could be made more DRY? Perhaps a shared function to do the bulk of the work that is the same.

Yes, this would be possible, if we also support `titleKey` for in `addPosterIcon`.
Where should that shared function be, maybe in `discourse/lib/utilities`?

> We'll need to bump the plugin api version too as this function is new. You can do this by changing 0.5 to 0.6 in getPluginApi and having your plugin use 0.6.

How exactly does this work out for existing plugins: Will all of these need to adopt the new api version? If not, will there be warnings or what's the consequence of using a different api version than the one present in a Discourse instance?

eviltrout:
> Yes, this would be possible, if we also support titleKey for in addPosterIcon.
> Where should that shared function be, maybe in discourse/lib/utilities?

I would just put the function in the plugin-api file for now, since it's only called from in there. No need to prematurely export it anywhere else.

> How exactly does this work out for existing plugins: Will all of these need to adopt the new api > version? If not, will there be warnings or what's the consequence of using a different api 
> version than the one present in a Discourse instance?

The version check in `plugin-api` is a `<=` so if you make it `0.6` it'll return the same object that was used before. This is what we do when adding versions, as people who code for `0.6` will be guaranteed to get a version of the object with the method.  Nobody who is coding against lower versions will have to change anything.

I'm not sure about the vagrant download issue but hopefully someone will reply.

SamSaffron:
@eviltrout  will leave this to you, perhaps just hack it in yourself to taste?

eviltrout:
I think I'll close it, since it's unfinished, hasn't been updated since August and nobody is clamouring for it. Feel free to reopen @caugner if you have time to address the issues, thanks!

PR #4377, opened 2016-08-09
FEATURE: warn localhost for webhooks

fantasticfears:


tgxworld:
Hmm I'm not sure if I understand the motivation for this. If the user is entering a URL that is local, aren't they intentionally doing it? :smile: The PR looks good to me though :+1:

fantasticfears:
@tgxworld I updated the PR. It will also check whether it points to the same instance. PR is simply a precaution

app/assets/javascripts/admin/controllers/admin-web-hooks-show.js.es6
```diff
@@ -80,12 +80,24 @@ export default Ember.Controller.extend({
...
-        this.get('controllers.adminWebHooks').get('model').addObject(model);
-      }).catch(popupAjaxError);
+      const saveWebHook = () => {
+        return model.save().then(() => {
+          this.set('saved', true);
+          this.get('controllers.adminWebHooks').get('model').addObject(model);
+        }).catch(popupAjaxError);
+      };
+
+      if (url.indexOf('localhost') !== -1 || url.match(/192\.168\.\d+\.\d+/) || url.match(/127\.\d+\.\d+\.\d+/) || url === Discourse.BaseUrl) {
```
tgxworld: hmm I'm not getting this. The URL can be anything. Why does it have to be `===` to `Discourse.BaseUrl`?

fantasticfears: I believe `Discourse.BaseUrl` is the domain name even for multisite, isn't it?

tgxworld: Correct. But a webhook can be posted to any URL right? I think I might be missing something here :sweat_smile:

fantasticfears: > But a webhook can be posted to any URL right?

Yep, but it doesn't matter. I already extract the domain from the url...Probably the name `url` confuses you?

tgxworld: hmmm if the domain is extracted from the url, why will `localhost` or the `IP`s be checked for here?

Can you post some examples of the URLs being checked here? I'm working on smth now so I'll be hard for me to pull down the branch and check

fantasticfears: Sure. The check is all the way before validation so the url is starting with scheme. `extractDomainFromUrl` is the helper. It gets rid of path, scheme and port.

It yields:

- `https://meta.discourse.org:90/random` -> `meta.discourse.org`: domain name
- `http://192.168.0.1:80/random` -> `192.168.0.1`
- `http://localhost:21/` -> `localhost`

Then I performed `===` for both possible local url.

PR #4379, opened 2016-08-09
FIX: Disable pasting hack for Firefox 50+

caugner:
**Background**:
There has been an image pasting hack in Discourse to allow pasting images in Firefox which did not implement the Clipboard API so far. As mentioned on [Meta](https://meta.discourse.org/t/firefox-50-makes-image-pasting-hack-superfluous/48017/3), FF 50 and later now [implement this](https://bugzilla.mozilla.org/show_bug.cgi?id=906420) though.

**Impact**:
This change in Firefox renders the workaround obsolete and more precisely requires it to be disabled for these versions. Otherwise pasted images are uploaded and inserted twice (*).

**Solution**:
Extend the existing version constraint to only enable the workaround in FF versions prior to 50.

(*) Note: There is currently still a bug in the FF 50 Clipboard API implementation which causes pasted images to be uploaded twice (one PNG and one JPEG) or three times respectively (with the pasting hack enabled), but [a fix for this is on the way](https://bugzilla.mozilla.org/show_bug.cgi?id=1290688) and disabling the pasting hack is necessary in either case.

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-editor.js.es6
@@ -296,7 +296,13 @@ export default Ember.Component.extend({
   // Believe it or not pasting an image in Firefox doesn't work without this code
   _firefoxPastingHack() {
     const uaMatch = navigator.userAgent.match(/Firefox\/(\d+)\.\d/);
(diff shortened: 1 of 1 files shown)


coding-horror:
:+1: very nice, thank you!

eviltrout:
Thanks!

PR #4380, opened 2016-08-10
FIX: Revised post not updated correctly when merging posts.

tgxworld:
Calling `PostRevisor` in a transaction results in the old record being fetched by the client when it is notify of changes as the transaction has not been completed.

### DIFF
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -137,6 +137,7 @@ class PostRevisor
     Post.transaction do
       revise_post
 
+      yield if block_given?
       # TODO: these callbacks are being called in a transaction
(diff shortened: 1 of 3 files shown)


lib/post_revisor.rb
```diff
@@ -137,6 +137,7 @@ def revise!(editor, fields, opts={})
     Post.transaction do
       revise_post
 
+      yield if block_given?
```
tgxworld: cc @eviltrout  @ZogStriP  Any thoughts on this extension?

ZogStriP: I ❤️  that

PR #4382, opened 2016-08-10
Fix Plural Strings That are Not Marked as Plural

SafaAlfulaij:
Some strings aren't marked as plural. This fixes them.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1028,7 +1028,9 @@ en:
       similar_topics: "Your topic is similar to..."
       drafts_offline: "drafts offline"
 
-      group_mentioned: "By mentioning {{group}}, you are about to notify <a href='{{group_link}}'>{{count}} people</a> – are you sure?"
+      group_mentioned:
+        one: "By mentioning {{group}}, you are about to notify <a href='{{group_link}}'>1 person</a> – are you sure?"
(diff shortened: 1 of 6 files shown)


config/locales/client.en.yml
```diff
@@ -1263,7 +1265,9 @@ en:
       unsubscribe:
         stop_notifications: "You will now receive less notifications for <strong>{{title}}</strong>"
         change_notification_state: "Your current notification state is "
-      filter_to: "{{post_count}} posts in topic"
+      filter_to:
+        one: "1 posts in topic"
```
gschlager: nitpick: "1 post" :wink:

config/locales/client.en.yml
```diff
@@ -1928,6 +1942,8 @@ en:
     posts: "Posts"
     posts_lowercase: "posts"
     posts_long: "there are {{number}} posts in this topic"
+
+    # keys ending with _MF use message format, see /spec/components/js_local_helper_spec.rb for samples
```
gschlager: I guess it would be a good idea to link to https://meta.discourse.org/t/message-format-support-for-localization/7035 instead. Maybe update the existing comment that you copied too. The topic on meta is a lot more helpful than the spec.

tgxworld:
Looks good to me. Thanks @SafaAlfulaij

PR #4387, opened 2016-08-12
FEATURE: Tags intersection page

gdpelican:
As discussed here:
https://meta.discourse.org/t/tag-intersections-page/47952

*Tags page with single tag (as existed before):*
screen shot 2016-08-12 at 3 49 41 pm

*Tags page with intersection (this is new):*
screen shot 2016-08-12 at 3 49 27 pm

NB that this exposes a route,
`tags/intersection/:tag_a/:tag_b`, which will allow access to an intersection of tags, but does not present that route in the interface anywhere.


bedhed:
:+1:

PR #4388, opened 2016-08-12
Always use --force when pulling translations from Transifex

gschlager:
Also, this makes it possible to pull only selected languages which is handy during development and for adding a new language.


nlalonde:
Agreed that we should always be using --force. Also, thanks for the language selection. :+1:

PR #4389, opened 2016-08-12
UX: Allow translation of composer icons

gschlager:
The icons for Bold, Italic and Heading can be replaced by a custom translation. The Font Awesome icons are only used if there is no translation or it the translated string equals the defaults (`B`, `I`, `H`).

This PR also changes the Font Awesome icon for Heading from `fa-font` to `fa-header`.

Screenshot of English toolbar with Font Awesome icons:
image

Screenshot of German toolbar with text:
image

https://meta.discourse.org/t/internationalize-b-i-and-a-in-the-message-composer-toolbar/48369

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -21,6 +21,11 @@ function getHead(head, prev) {
   }
 }
 
+function getButtonLabel(labelKey, defaultLabel) {
+  // use the Font Awesome icon if the label matches the default
+  return I18n.t(labelKey) === defaultLabel ? null : labelKey;
+}
+
 const OP = {
   NONE: 0,
   REMOVED: 1,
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -162,6 +168,11 @@ class Toolbar {
       g.buttons.push(createdButton);
     }
   }
+
+  getButtonLabel(labelKey, defaultLabel) {
```
eviltrout: This is minor but it seems `getButtonLabel` never uses `this` so it doesn't need to be an object method. You can hoist it up outside of the object declaration and it should still work.

gschlager:
@eviltrout Thanks for the tip. I moved the function outside the class.

eviltrout:
Thanks for this! Looks good to me.

PR #4391, opened 2016-08-13
Include colon in localized strings

guillaumekln:
https://meta.discourse.org/t/punctuation-not-correctly-localized/48484

### DIFF
--- a/app/assets/javascripts/discourse/models/topic.js.es6
+++ b/app/assets/javascripts/discourse/models/topic.js.es6
@@ -67,8 +67,8 @@ const Topic = RestModel.extend({
   }.property('bumped_at', 'createdAt'),
 
   bumpedAtTitle: function() {
-    return I18n.t('first_post') + ": " + longDate(this.get('createdAt')) + "\n" +
-           I18n.t('last_post') + ": " + longDate(this.get('bumpedAt'));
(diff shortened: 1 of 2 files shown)


gschlager:
`last_post` is reused multiple times, with and without the colon. We need one string with colon and one without.

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/templates/group-index.hbs#L14
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/templates/user-card.hbs#L84
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/templates/mobile/components/basic-topic-list.hbs#L33

`fist_post` is used only once, so that change is fine.

ZogStriP:
It would be better to create a new string for `bumpedAtTitle`, that way, everyone is happy and we don't risk messing with existing translations.

guillaumekln:
Thanks for the review.

I'm just not sure what the naming should be. The colon is needed when `last_post` is used in the `title` attribute. Is `last_post_title` approriate then? Or `last_post_tooltip`?

SamSaffron:
closing this for now, ideally i would include the date in the localization that way you can cleanly localize 

last_post_with_date ... etc...

PR #4395, opened 2016-08-15
FIX: correct regex for username

frictionel:
The username was not correctly detected for non-word chars, so I reuse the regex from:

https://github.com/discourse/discourse/blob/3b792054f2e3e3611ea9c45326aaa8cd604759d0/config/routes.rb#L10


tgxworld:
Thanks!

PR #4396, opened 2016-08-16
FIX: group names may be null

fantasticfears:


test/javascripts/lib/utilities-test.js.es6
```diff
@@ -27,10 +27,13 @@ test("urlValid", function() {
   ok(urlValid('http://meta.discourse.org/'), "allows http url");
   ok(urlValid('http://meta.discourse.org'), "allows url without trailing slash");
   ok(urlValid('https://meta.discourse.org/'), "allows https url");
+  ok(urlValid('http://142.42.1.1:8080/'), "allows ip with port");
+  ok(urlValid("http://-.~_!$&'()*+,;=:%40:80%2f::::::@example.com"));
```
tgxworld: This is valid?

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -76,8 +76,8 @@ export function emailValid(email) {
 }
 
 export function urlValid(url) {
-  // see: https://mathiasbynens.be/demo/url-regex
-  var re = /^https?:\/\/(-\.)?([^\s/?\.#-]+\.?)+(\/[^\s]*)?$/i;
+  // adapt from npm: url-regex
+  var re = /^(?:(?:http|https):\/\/)(?:\S+(?::\S*)?@)?(?:(?:(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[0-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))|localhost)(?::\d{2,5})?(?:(\/|\?|#)[^\s]*)?$/i;
```
tgxworld: This regexp is really complicated. Why is this necessary though?

test/javascripts/lib/utilities-test.js.es6
fantasticfears: Yes, it is, your browser should redirect you to `example.com`. Those characters are just user information. I think I can just change them to a dashed domain name. Would be easier to understand

app/assets/javascripts/discourse/lib/utilities.js.es6
fantasticfears: It's like parsing html by regexp which is a really bad practice any how. But other parsers are at least 50 lines or even more. (We need at least domain name parser, IPv4 parser and IPv6 parser) What would be your call here?

S-YOU: What kind of ipv6 url would it match? 
I tried some like http://2607:f8b0:4000:80a::200e, http://[1080:0:0:0:8:800:200C:417A]/index.html, it match nothing.
url-regex site don't even mention anything about ipv6 either.

test/javascripts/lib/utilities-test.js.es6
tgxworld: Holy that is really valid :stuck_out_tongue:

app/assets/javascripts/discourse/lib/utilities.js.es6
tgxworld: Can you split this into another PR since it isn't related to the PR's title. Also, I'm not sure if we should even support this regexp since the additional cases are correct but pretty rare IMO.

fantasticfears: @S-YOU No, it only matches the IPv4 url. I mean any validation should go with state machine to parse that accordingly. Regexp are always a hack but it's simple.

I'll split them.

tgxworld:
Opps I rebased the base branch against master. Can you rebase as well? thanks!

fantasticfears:
It's OK. Done

tgxworld:
Thanks!

PR #4397, opened 2016-08-17
FEATURE: Allow backups to be encrypted using GnuPG before uploading to S3.

tgxworld:


jscinoz:
Is there any possibility of this PR being merged? It would be great to see support for encrypted backups in Discourse.

tgxworld:
@jscinoz I'm reopening this for me to track but I think this was closed because we needed to add gpg to the base image before we can shift this feature.

SamSaffron:
I am generally supportive of this, let's just make sure we add this to the docker base image and then merge it in/

app/models/backup.rb
```diff
@@ -50,8 +50,23 @@ def s3
...
+    if SiteSetting.s3_enable_encryption
+      if (s3_gnupg_public_key = SiteSetting.s3_gnupg_public_key).blank?
+        raise Discourse::SiteSettingMissing.new("s3_gnupg_public_key")
+      end
+
+      encrypted_filename = "#{@filename}.gpg"
+      encrypted_path = "/tmp/#{encrypted_filename}"
+
+      `echo '#{s3_gnupg_public_key}' | gpg --import`
+      gnupg_user_id = `echo '#{s3_gnupg_public_key}' | gpg --list-keys --with-colons | awk -F: '/^pub:/ { print $5 }'`.chomp
```
tgxworld: @mpalmer Can I get your thoughts on this? This is very hacky to me but I'm not quite sure if there are other ways around this. Thank you!

mpalmer: I think you're doomed to hack around gpg's UI limitations... it is a fairly brutal tool.  The way you're doing things here, though, is likely to end in tears sooner or later, just because someone will change the GPG key used for encrypting their backups and your output will then not represent a single key...

You'll need to use a tempdir for storing the GPG config (or at least the public keyring), and then use `--no-default-keyring` and `--keyring /some/tmp/file` to ensure you *only* pick up the keyring file that contains the key you want to use for this session.

app/models/backup.rb
```diff
@@ -50,8 +50,23 @@ def s3
...
+      `gpg --encrypt --output #{encrypted_path} --batch --yes --verbose --recipient #{gnupg_user_id} #{@path}`
```
mpalmer: I'd greatly prefer it if gpg was used in "pipe" mode (send data in on stdin, read data back on stdout, or send only the encrypted data to disk), as it saves diskspace and disk I/O, and doesn't end up with unencrypted data laying around on disk.

app/models/backup.rb
```diff
@@ -50,8 +50,23 @@ def s3
...
+
+      File.open(encrypted_path) { |file| s3.upload(file, encrypted_filename) }
+      File.delete(encrypted_path)
```
mpalmer: Please to be using `ensure`, because having upload-failed GPG-encrypted backups laying around makes Baby Sysadmin Jesus cry.

SamSaffron:
I am closing this, we can't have it be open forever, @tgxworld  if you feel like bringing this back feel free.

PR #4398, opened 2016-08-17
FIX: Tags on topic crawler view

xfalcox:
- Remove tag object id
- Remove duplicate tag list
- Don't display tags when tagging is disabled

### DIFF
--- a/app/helpers/topics_helper.rb
+++ b/app/helpers/topics_helper.rb
@@ -15,7 +15,7 @@ module TopicsHelper
       breadcrumb.push url: category.url, name: category.name
     end
 
-    if (tags = topic.tags).present?
+    if SiteSetting.tagging_enabled && (tags = topic.tags).present?
       tags.each do |tag|
         url = "#{Discourse.base_url}/tags/#{tag.name}"
         breadcrumb << {url: url, name: tag.name}
--- a/app/views/topics/show.html.erb
+++ b/app/views/topics/show.html.erb
@@ -18,16 +18,6 @@
 </div>
 <% end %>
 
-<%- if SiteSetting.tagging_enabled && @topic_view.topic.tags.present? %>
-  <div class='tags'>
-    <%= t 'js.tagging.tags' %>:
-
-    <%- @topic_view.topic.tags.each do |t| %>
-      <%= t %>
-    <%- end %>
-  </div>
-<% end %>
-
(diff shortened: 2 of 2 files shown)


SamSaffron:
hmm but then there is no semantic diff between tag and category ... not so sure about this, perhaps should be in another section.

xfalcox:
> hmm but then there is no semantic diff between tag and category ... not so sure about this, perhaps should be in another section.

Yeah, I removed the section because it was strange having tags on the breadcumb and just after in a div right above.

----

On Google results tags are show as subcategories, so it's great:

image

The cached page, shows the duplication (ignore the tags id problem for now)

http://webcache.googleusercontent.com/search?q=cache:fmUl4lje4LMJ:meta.discourse.org/t/discourse-server-slower-after-update/36857+&cd=1&hl=en&ct=clnk&gl=br

I really don't think the separation is needed for the crawler view.

tgxworld:
Hmm I kind of agree with @SamSaffron here. We should probably remove the tags from breadcrumb and  keep the existing tags loop that we're doing.

SamSaffron:
I don't want to keep this open forever, but I am really having trouble following what this solves:

We now display this in Google

```
Discourse server slower after update - dev - Discourse Meta
meta.discourse.org › dev › docker › sso
```

The PR is attempting to make it display? 

```
Discourse server slower after update - dev - Discourse Meta
meta.discourse.org › dev
```

regrading hiding tags if disabled, absolutely agree with that

SamSaffron:
I am closing this for now ...

tgxworld:
> The PR is attempting to make it display?

> Discourse server slower after update - dev - Discourse Meta
meta.discourse.org › dev

Not really,  the PR is keeping tags in the breadcrumbs and removing the extra tags list which is incorrectly displaying the object ids of the tags as shown by http://webcache.googleusercontent.com/search?q=cache:https://meta.discourse.org/t/automatically-set-topics-to-watching-after-posting/21230

SamSaffron:
ahh I see

xfalcox:
Exactly @tgxworld <3

I think it's the correct thing to do.

tgxworld:
Hmm I was holding on merging this because of . @SamSaffron 's comments in https://github.com/discourse/discourse/pull/4398#issuecomment-240325032. Having tags appear in the breadcrumb is really odd. :thinking:

PR #4399, opened 2016-08-17
FIX: Bug fixes for tags intersection page

gdpelican:
As noted here: https://meta.discourse.org/t/tag-intersections-page/47952/8


gdpelican:
Hang on, it looks like Neil has been doing some work on this as well; let me merge in master and confirm this is still right.

PR #4400, opened 2016-08-18
FIX: improve url validation regexp

fantasticfears:


coding-horror:
I do not feel we should support this, as it is dangerous:

`http://userid:password@example.com:8080/`

also, I am unclear, why do we need to support arbitrary numeric IP addresses here?

fantasticfears:
@coding-horror Sure, I can leave user information part aside.

The numeric IP addresses is still legitimate URL. Webhook is such an advanced feature. So admin should be able to point that to an internal service (IP address) if they insist. Either way, an actual parser should definitely help if we want to restrict them in various of situations.

tgxworld:
@fantasticfears You'll have to open the PR against the `master` branch :)

fantasticfears:
@tgxworld Github now supports changing target branch. Are you sure about changing the branch to master? I think you are asking another rebase here

tgxworld:
@fantasticfears O is the url checker added as part of the webhook project?

fantasticfears:
Yes

tgxworld:
@eviltrout  What do you think about this change? We need a URL validator here and instead of maintaining a giant RegExp. I advised @fantasticfears  to pull in another library instead.  :)

eviltrout:
@tgxworld is it necessary to have such complicated validation client side? I'm wondering if we can keep the basic validation then do the complicated validation server side? Otherwise it's an extra thousand lines of code to carry around

fantasticfears:
Using Ruby `uri` module for validation now. It's more like a parser instead of validator. Anyway, the delivery status would be apparent if any error.

tgxworld:
@fantasticfears  I don't see the validation in the changes?

fantasticfears:
@tgxworld Because it's already there 😉  See https://github.com/discourse/discourse/blob/gsoc_webhooks/app/models/web_hook.rb#L10

tgxworld:
:+1:

PR #4402, opened 2016-08-19
Fix typo Vietnamese language

x0rm3r:


ZogStriP:
I'm sorry but we only accept changes to the locales using [Transifex](transifex.com/discourse) otherwise they would be lost the next time we pull translations.

PR #4409, opened 2016-08-22
FIX: Update Nabble importer to use Email.Receiver new API

Mogztter:
Email.Receiver API changed in 30836573587079c5e663d7b3122957fc8c70dafe

### DIFF
--- a/script/import_scripts/nabble.rb
+++ b/script/import_scripts/nabble.rb
@@ -116,12 +116,12 @@ class ImportScripts::Nabble < ImportScripts::Base
   end
 
   def parse_email(msg)
-    receiver = Email::Receiver.new(msg, skip_sanity_check: true)
+    receiver = Email::Receiver.new(msg)
     mail = Mail.read_from_string(msg)
     mail.body
 
-    selected = receiver.select_body(mail)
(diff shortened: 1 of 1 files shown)


ZogStriP:
Thanks 👍

Mogztter:
I'm glad to contribute to this *awesome* project :wink: 

I think `mbox` importer needs to be updated as well: https://github.com/discourse/discourse/blob/master/script/import_scripts/mbox.rb#L296

I would like to add specs on importer, do you have any recommandations/rules to follow ?

Thanks

ZogStriP:
> I would like to add specs on importer, do you have any recommandations/rules to follow ?

Unfortunately, this is one of the area we're lacking tests. There are a [few](https://github.com/discourse/discourse/tree/master/script/import_scripts/socialcast) [importers](https://github.com/discourse/discourse/tree/master/script/import_scripts/quandora) which have tests. You can use them as inspiration if you want.

mcwumbly:
@Mogztter while I've added some tests on the two importers I wrote, I'd be
happy to kick around ideas for different approaches. If you want to discuss
any of this is more detail, feel free to create a topic on meta.

PR #4411, opened 2016-08-23
filter out api key

rimian:
I think it's a good idea to filter this from the logs


tgxworld:
`api_username` as well? :)

rimian:
@tgxworld done! :)

tgxworld:
Thanks!

PR #4413, opened 2016-08-23
FIX: Nabble importer, attachments can be not found

Mogztter:
On my Nabble instance the following query sometimes returns an empty result:

```sql
SELECT content 
FROM file_node 
WHERE name='#{basename}' AND node_id = #{postid}
```

I think importer should not throws an exception when this occurred. 
Maybe we should add a `Rails.logger.warn("Unable to find attachment content with name '#{name}' and post id #{postid}")` ?

### DIFF
--- a/script/import_scripts/nabble.rb
+++ b/script/import_scripts/nabble.rb
@@ -199,28 +199,30 @@ class ImportScripts::Nabble < ImportScripts::Base
   def process_attachments(txt, postid)
     txt.gsub!(/<nabble_img src="(.*?)" (.*?)>/m) do |match|
       basename = Regexp.last_match[1]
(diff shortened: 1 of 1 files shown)


Mogztter:
Anything wrong with my contribution ?

techAPJ:
Looks good to me! Thanks! 👍  (Apologies for delay in reviewing PR!)

> Maybe we should add a `Rails.logger.warn("Unable to find attachment content with name '#{name}' and post id #{postid}")`?

Sure, sounds good! Just a minor nitpick that instead of `Rails.logger.warn` we should simply use `puts` to notify user while import script is runnning. Can you create a new PR for that? :)

PR #4415, opened 2016-08-24
FIX: “Your search term is too short” message when no search term is entered at all

vinothkannans:
PR for https://meta.discourse.org/t/your-search-term-is-too-short-message-when-no-search-term-is-entered-at-all/49070

### DIFF
--- a/app/assets/javascripts/discourse/widgets/search-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/search-menu.js.es6
@@ -97,14 +97,15 @@ export default createWidget('search-menu', {
     const results = [this.attach('search-term', { value: searchData.term, contextEnabled }),
                      this.attach('search-context', { contextEnabled })];
 
-    if (searchData.term)
-    if (searchData.loading) {
-      results.push(h('div.searching', h('div.spinner')));
-    } else {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/widgets/search-menu.js.es6
```diff
@@ -97,6 +97,7 @@ export default createWidget('search-menu', {
     const results = [this.attach('search-term', { value: searchData.term, contextEnabled }),
                      this.attach('search-context', { contextEnabled })];
 
+    if (searchData.term)
```
eviltrout: You need to update this to use braces. `if` statements without them are so error prone for refactoring later on. Additionally, I'm surprised this works without the additional check of `searchData.term.length > 0`

vinothkannans: I didn't used braces because I don't want to touch line below for aligning. Anyway I will do it now.

No need for that. If length is `0` then `if (searchData.term)` will always return `false`

vinothkannans:
@eviltrout it's done now

eviltrout:
Thanks!

PR #4416, opened 2016-08-24
FIX: missing post_count value and typo in de-locales

phaldan:


ZogStriP:
Thanks but we only accept updates to translations via [Transifex](https://www.transifex.com/discourse/discourse-org/) otherwise they get overwritten the next time we pull the translations.

PR #4419, opened 2016-08-25
FEATURE: Prioritize recent posts in Rake task.

tgxworld:
cc @ZogStriP  @techAPJ  Please review :)

Related to https://github.com/discourse/discourse/commit/5da768b4fec40cbea05df30c75f4f0b897f05e9e

### DIFF
--- a/lib/tasks/posts.rake
+++ b/lib/tasks/posts.rake
@@ -18,7 +18,7 @@ task 'posts:fix_letter_avatars' => :environment do
   rebaked = 0
   total = search.count
 
-  search.find_each do |post|
+  find_each_with_recent(search) do |post|
     rebake_post(post)
     print_status(rebaked += 1, total)
   end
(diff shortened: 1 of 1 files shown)


lib/tasks/posts.rake
```diff
@@ -134,3 +134,16 @@ task 'posts:remap', [:find, :replace] => [:environment] do |_,args|
   end
   puts "", "#{i} posts remapped!", ""
 end
+
+def find_each_with_recent(given_scope, recent_date: 6.months.ago)
```
tgxworld: `6.months.ago` is magical to me and what works well for a not so active site might not work well for a super active site. Any thoughts on how this can be improved?

ZogStriP: Note sure, maybe add a parameter? That way people can tell the rake task what "recent" mean?

SamSaffron:
not following ... why do we need 2 scopes, why not simply order by created_at descending?

tgxworld:
> NOTE: It's not possible to set the order. That is automatically set to ascending on the primary key (“id ASC”) to make the batch ordering work. This also means that this method only works when the primary key is orderable (e.g. an integer or string).

> NOTE: You can't set the limit either, that's used to control the batch sizes.

That is because ordering doesn't work with [`find_each`](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html) :)

SamSaffron:
Then the solution here is a find_each that does not suck :) not this hack

PR #4422, opened 2016-08-26
Fix SQLite datetime issues

steinwaywhw:
The raw datetime string from raw emails doesn't conform to SQLite standard. Therefore it can't be sorted. This fix will make it conforming and sort-able by SQLite.


coding-horror:
if this looks OK @eviltrout  can you merge it?

eviltrout:
Thanks!

PR #4424, opened 2016-08-29
Adjust example order to match description

bradleypriest:
Sorry for the ridiculousness of this change, just an over-eager eye.

This adjusts the order of the formatting examples in the activation email to match the order they are described in the previous line.

HTML -> BBCode  -> Markdown.


tgxworld:
Thanks @bradleypriest!

PR #4427, opened 2016-08-30
[WIP] Add text on last-visit line

xfalcox:
Just gotta add the i18n string on the tr data-last-visit-text attribute, any tips?

Probably around here:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/topic-list-item.js.es6#L51


coding-horror:
Does this work for variable title height, e.g. a long topic title that is 2 lines wrapped, or a topic title with a tag underneath?

cpradio:
It does now, since it is being applied to the TR following the one that has the class. Discourse Community FTW on this one! :+1:

coding-horror:
ok let's try it!

PR #4428, opened 2016-08-31
Add text to new messages line

xfalcox:


coding-horror:
If this looks ok to you @tgxworld maybe merge?

tgxworld:
Hmm the implementation in this PR and https://github.com/discourse/discourse/pull/4427 doesn't look right to me. One problem is that the text is not centered properly and using the session to decide when to add the data attribute to the next topic list item feels abit like a hack to me.

I gave it a shot while I was at it and came up with

https://github.com/tgxworld/discourse/commit/5647408d822af3dda29815c030583b6aa9ae5ec7

cc @SamSaffron

coding-horror:
@SamSaffron can you review and see what you think?

app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -54,6 +55,19 @@ export default Ember.Component.extend(StringBuffer, {
     return classes.join(' ');
   }.property(),
 
+  lastVisitMessage: function() {
+    if (this.get('lastVisitedTopic') == this.get('topic')) {
+      this.session.set('lastVisitedTopicAbove', true);
```
SamSaffron: all this session stuff is really strange, its a computed property ... 

also ... I don't think this would work properly if you drilled into a category cause the line can move.

config/locales/client.en.yml
```diff
@@ -1221,6 +1221,7 @@ en:
     current_user: 'go to your user page'
 
     topics:
+      new_messages_marker: "New Messages"
```
gschlager: Shouldn't this be `new_topics_marker: "New Topics"`?

coding-horror: It should be neither, it should be "Last Visit"

coding-horror:
We need this resolved @xfalcox

xfalcox:
Rebased

tgxworld:
Merged in https://github.com/discourse/discourse/commit/16aad189fb50527e87e7b4ea2af993b41d175d56

PR #4430, opened 2016-08-31
check the plugin mobile dir for templates

angusmcleod:
The resolver does not check the mobile template directory in plugins for templates, meaning it is not possible to override mobile templates from a plugin. This adds a check for mobile templates in plugins to the resolver.


eviltrout:
Thanks, looks good :)

PR #4432, opened 2016-09-02
FIX: Make `Jobs::MigrateUploadScheme` a scheduled job.

tgxworld:
This reverts https://github.com/discourse/discourse/commit/9d8db11c

If the job fails during execution or if `SiteSetting.migrate_to_new_scheme`
has been set to `false`, the job will be considered as executed even
though the migration has not been completed. As a result, the job
will have to be executed manually which is not desirable.


tgxworld:
@eviltrout Do you recall why the change in https://github.com/discourse/discourse/commit/9d8db11c was made? I hope I'm not introducing a regression when reverting the commit.

PR #4434, opened 2016-09-02
fix order of plugin template resolution

angusmcleod:
See further here: https://meta.discourse.org/t/how-to-override-an-existing-handlebars-template-from-plugin-part-ii/12245/28?u=angus


eviltrout:
👍

PR #4435, opened 2016-09-06
User Profile tabs plugin-outlet

zierg:
Add "user-profile-tabs" plugin-outlet which allows you to add tabs to the user profile page.


SamSaffron:
I don't think this works cause if you have multiple it will wrap the `li` in a div breaking the list.

SamSaffron:
Sorry, closing, we need to place the outlet outside of the list

chenyingyin:
default

PR #4436, opened 2016-09-09
Rubygems upgrade needed for mailcatcher

AhmadFCheema:
Recent version of mailcatcher giving error for the default rubygems version. The proposed solution is as mentioned in this [mailcatcher issue](https://github.com/sj26/mailcatcher/issues/267).

The updated rubygems version was 2.6.6. Although I was able to execute the rails server after the update, as changing the version might possibly have wide spread unwanted consequences, I will leave it for the maintainers to decide whether to accept this pull request or not.


SamSaffron:
I will be so happy when @JaredReisinger  change is all merged in and we make docker the official dev env :)

PR #4437, opened 2016-09-12
FIX: show event name in webhook headers

fantasticfears:
@tgxworld


tgxworld:
Thanks @fantasticfears

PR #4438, opened 2016-09-13
Fix %{} expressions in right-to-left languages

danielhers:
Namely, Hebrew, Arabic and Farsi.
Some of these expressions had the % sign AFTER the {} by mistake, and in rendering, these would come out as the literal expression rather than localized:
screenshot


tgxworld:
@techAPJ  Do you know if this has to be done it Transifex?

techAPJ:
Yes, these changes needs to be made on [Transifex](https://transifex.com/projects/p/discourse-org/) as they will be overwritten the next time we update translations.

PR #4439, opened 2016-09-13
FEATURE: Webhook for user creation and approval

fantasticfears:


coding-horror:
Excellent can you review @tgxworld?

app/models/web_hook.rb
```diff
@@ -67,6 +67,15 @@ def self.enqueue_topic_hooks(topic, user)
       )
     end
   end
+
+  %i(user_created user_approved).each do |event|
+    DiscourseEvent.on(event) do |user|
+      WebHook.enqueue_hooks(:user_creation,
+                            user_id: user.id,
```
tgxworld: NITPICK: I usually prefer that we don't flush the parameters all the way even though we do it in our Ember code. The downside of flushing the parameters is that if you change a line that changes the number of spaces it is indented by, all the following lines have to be updated and that pollutes the commit history.

tgxworld:
Merged with my changes in c463cf63d4a40375519af56dd01d22654c149063

tgxworld:
Thanks @fantasticfears!

PR #4440, opened 2016-09-15
Realtime events for webhooks and headers

fantasticfears:


tgxworld:
@fantasticfears Looks good so far to me on first review. I'll do another pass tomorrow morning as it is getting alittle late here

tgxworld:
Thanks @fantasticfears. Merged in https://github.com/discourse/discourse/commit/00d5facf365ae67dce960d4b5b98695fccb2b1c7. I made some change which will be good if you can have a look :)

PR #4442, opened 2016-09-15
KUNENA3 importer: parameterize database and fields, link to attachments

pfaffman:
Three changes, really, but I just couldn't deal with 3 PRs. Sorry.

UX: The `jos_` prefix changes in different imports due to unknown variables. I added a `KUNENA_PREFIX` parameter to avoid global replacements across the importer.

UX: In one database I imported, the `parent_id` field was called `parent`. I added a parameter `PARENT_FIELD` to deal with this.

FEATURE: Previous version left in `[attachment]` bbcodes. I instead provide URLs to the attachments.

### DIFF
--- a/script/import_scripts/kunena3.rb
+++ b/script/import_scripts/kunena3.rb
@@ -1,17 +1,23 @@
 require "mysql2"
 require File.expand_path(File.dirname(__FILE__) + "/base.rb")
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
seems fine, thanks!

PR #4444, opened 2016-09-16
FIX: Backup validation wasn't escaping hyphens

cpradio:


cpradio:
Build seems to have failed due to an unrelated qunit test that was later fixed by @nlalonde

app/controllers/admin/backups_controller.rb
```diff
@@ -119,7 +119,7 @@ def upload_backup_chunk
 
     return render status: 415, text: I18n.t("backup.backup_file_should_be_tar_gz") unless /\.(tar\.gz|t?gz)$/i =~ filename
     return render status: 415, text: I18n.t("backup.not_enough_space_on_disk")     unless has_enough_space_on_disk?(total_size)
-    return render status: 415, text: I18n.t("backup.invalid_filename") unless !!(/^[a-zA-Z0-9\.-_]+$/ =~ filename)
+    return render status: 415, text: I18n.t("backup.invalid_filename") unless !!(/^[a-zA-Z0-9\._-]+$/ =~ filename)
```
ZogStriP: Should probably escape the `-` like we do for the `.` in order to prevent the issue from happening again.

cpradio: Technically you don't have to when it ends with it, but you want it escaped anyway? As the test passes. I intentionally updated the test so it would fail if someone altered it and failed to escape the hyphen.

cpradio:
I've responded to your comment. The escaping isn't needed when the hyphen is at the end. I also updated the test/spec so it would fail if someone improperly moved the hyphen to not be at the end, thus that spec would fail.

app/controllers/admin/backups_controller.rb
ZogStriP: I know it isn't needed but I think it's a good practice to always escape the `-` in a regular expression in order to avoid this kind of bug.

coding-horror: yes I agree simpler to "always escape" being at the end is a bit of a technicality that not everyone will understand

cpradio:
Okay, it is escaped.

tgxworld:
Woops Thanks for fixing @cpradio!

pfaffman:
I'm still figuring this stuff out, but I think this needs a version bump because people with latest version (on `tests-passed` and `stable`) can't upload until they get this update.

coding-horror:
If that's the case they never would have upgraded yet, so they would have
neither the bug, nor the fix.

pfaffman:
Ah. Now I get it. It's just people who were unlucky enough to do an install after 

https://github.com/discourse/discourse/commit/512922d7767c84615b8b77089e34ede244a88c7f

but before this fix got accepted. As far as I know, that's just [two people](https://meta.discourse.org/t/cannot-upload-a-downloaded-backup/50250/2).

coding-horror:
Yes exactly -- only people who were already pulling "beyond the weekly
release" would be subject to the bug and even then only when they picked up
that buggy version in particular.

PR #4445, opened 2016-09-17
Update INSTALL-cloud.md

kathiedart:
Default to UTC because that's what it's there for. Not just some random American city.


coding-horror:
Oh, sorry, this is referring to the region of the datacenter the droplet will be hosted in, not the timezone of the operating system.

PR #4446, opened 2016-09-19
Unify API keys and web hooks into a single admin nav header.

tgxworld:
@techAPJ Can you help me review this? Thanks! :smile:


app/assets/javascripts/admin/templates/api-keys.hbs
```diff
@@ -0,0 +1,34 @@
...
+        {{#if k.user}}
+          {{#link-to 'adminUser' k.user}}
+            {{avatar k.user imageSize="small"}}
+          {{/link-to}}
+        {{else}}
+          {{i18n 'admin.api.all_users'}}
+        {{/if}}
+      </td>
+      <td>
+        <button class='btn' {{action "regenerateKey" k}}><i class="fa fa-undo"></i>{{i18n 'admin.api.regenerate'}}</button>
```
techAPJ: Minor nitpick: Here we should use `d-button` component instead.

```
{{d-button action="regenerateKey" actionParam=k icon="undo" label="admin.api.regenerate"}}
```

techAPJ:
Looks good to me! 👍

app/assets/javascripts/admin/templates/api-keys.hbs
tgxworld: Good catch! :smile:

app/assets/javascripts/admin/adapters/web-hook-event.js.es6
```diff
@@ -0,0 +1,7 @@
+import RESTAdapter from 'discourse/adapters/rest';
+
+export default RESTAdapter.extend({
+  basePath() {
+    return '/admin/api/';
```
ZogStriP: Just wondering: does this work when using subfolder? (Might have to replace this with `Discourse.getUrl("/admin/api/")`)

tgxworld: We usually don't have to worry about subfolder paths unless we're specifying the full URL. Most of the time, `Discourse.ajax` takes cares of appending the subfolder path :smile:

PR #4447, opened 2016-09-19
FEATURE: add "Approve new topics unless user level" setting

pmusaraj:


lib/new_post_manager.rb
```diff
@@ -73,6 +73,7 @@ def self.user_needs_approval?(manager)
 
     (user.trust_level <= TrustLevel.levels[:basic] && user.post_count < SiteSetting.approve_post_count) ||
     (user.trust_level < SiteSetting.approve_unless_trust_level.to_i) ||
+    (manager.args[:title].present? && user.trust_level < SiteSetting.approve_new_topics_unless_trust_level.to_i) ||
```
ZogStriP: This looks good but definitely needs a test.

pmusaraj: Sure. Are you looking for tests in Ruby or javascript? (I'm not too familiar with tests for either, but definitely less familiar with Ruby testing.)

ZogStriP: Ruby tests. You'll need to add a test case in [spec/components/new_post_manager_spec.rb](https://github.com/discourse/discourse/blob/master/spec/components/new_post_manager_spec.rb) ;)

pmusaraj: Tests added (and they helped me find another needed fix). It's my first rails test, so please check carefully.

SamSaffron:
Sure... lets pull this in, thanks @pmusaraj  !

PR #4450, opened 2016-09-20
Improve the "develop inside Docker" experience

JaredReisinger:
This PR is intended to work in concert with the discourse_docker changes in https://github.com/discourse/discourse_docker/pull/292.  I have used those changes to build a local "discourse_dev" image, and then _**these**_ changes to easily spin up and work on Discourse and plugin functionality.  It's working well for me, but of course YMMV.

Add `--init` and `--patch` options to bin/docker/boot_dev to make it easier to spin up a container running Discourse out of a development directory.

Add `bin/docker/README.md` to explain how to go about using the docker command-line tools.

Tweak the Docker command-line tools to use the `-u` option to specify user rather than changing to the user "inside" the container via `chpst` (This way, we don't have to explicitly specify the HOME environment variable, either.)

Add `bin/docker/shell` command to make it easy to jump inside the running container.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ public/plugins/*
 public/tombstone/*
 
 # Ignore bundler config
-/bin
 /.bundle
 /.vagrant
 /.vagrantfile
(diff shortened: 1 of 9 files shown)


coding-horror:
You should review this @SamSaffron

SamSaffron:
very happy the templates are gone, added a bunch of comments and I am not sure where they are gone so will do review again

SamSaffron:
I am merging this now (prior to the discourse docker merge which is in progress) mainly to help clear off the queue here :) 

Thanks @JaredReisinger  !

PR #4453, opened 2016-09-21
mbox: move configuration to ENV, add default user level

pfaffman:
Users can now copy this block to a shell to configure the script without having to edit it.
```bash
export MBOX_SUBDIR="messages" # subdirectory with mbox files
export LIST_NAME=LIST_NAME
export DEFAULT_TRUST_LEVEL=1
export DATA_DIR=~/data/import
export SPLIT_AT="^From " # or "^From (.*)"
```

There is also means to set a default user level, added in this PR because it's also a ENV variable.

Since MBOX files are all different, it is improbable that anyone would ever be able to do an MBOX import and not tweak the script, but it'll still make committing changes a bit easier since these lines won't need to change.

### DIFF
--- a/script/import_scripts/mbox.rb
+++ b/script/import_scripts/mbox.rb
@@ -1,16 +1,41 @@
 require 'sqlite3'
 require File.expand_path(File.dirname(__FILE__) + "/base.rb")
 
+# TODO: ignore ~ emacs backup files
+# DONE: sort filenames before processing
+
+# Paste these lines into your shell before running this:
+
+=begin
+export MBOX_SUBDIR="messages" # subdirectory with mbox files
+export LIST_NAME=LIST_NAME
+export DEFAULT_TRUST_LEVEL=1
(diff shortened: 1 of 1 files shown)


script/import_scripts/mbox.rb
```diff
@@ -234,7 +259,7 @@ def create_user_indices
   def clean_title(title)
     title ||= ""
     #Strip mailing list name from subject
-    title = title.gsub(/\[[^\]]+\]+/, '').strip
+    title = title.gsub(/\[#{LIST_NAME}\]/, '').strip
```
ZogStriP: You should use `Regexp.escape` here

pfaffman: Good catch, @ZogStriP !  Thanks.

pfaffman: I submitted a commit with that change. Do I need to make a new PR, or have you just not noticed y et?

Thanks.

PR #4454, opened 2016-09-21
Don't look for external facts when benchmarking

JaredReisinger:
When running `scripts/bench.rb` via the discourse/discourse_bench docker container, Facter throws an EACCES error attempting to load external puppet facts.  Since the benchmark script ends up ignoring everything except for seven specific keys, we can just avoid loading any external fact logic entirely.


SamSaffron:
cool!

PR #4455, opened 2016-09-22
FIX: Form CDN URL client side instead.

tgxworld:
@ZogStriP I think this is a better way of getting the category logo to be served via the CDN. Let me know you thoughts? :smile_cat:


ZogStriP:
Indeed, much betterer 👍

PR #4456, opened 2016-09-22
Update INSTALL-cloud.md

juliangamble:
Adding an example - as api doco is non-intuitive for `app.yml` file nested in `containers` directory when the configuration is simply `app`


coding-horror:
Hmm, I don't feel a single example is particularly useful or illustrative here, particularly of "stop"...

PR #4457, opened 2016-09-22
Add support for email whitelist/blacklist to GitHub auth

JaredReisinger:
If a site is configured for GitHub logins, _**and**_ has an email domain whitelist, it's possible to get in a state where a new user is locked to a non-whitelist email (their GitHub primary) even though they have an alternate email that's on the whitelist.  In all cases, the GitHub primary email is attempted first so that previously existing behavior will be the default.

- Add whitelist/blacklist support to GithubAuthenticator (via EmailValidator)

- Add multiple email support GithubAuthenticator

- Add test specs for GithubAuthenticator

- Add authenticator-agnostic "none of your email addresses are allowed" error message.


SamSaffron:
Thanks heaps! 

btw, team is away on yearly meetup next week so we are going to be a bit slow to respond to PRs

PR #4463, opened 2016-09-26
FIX: New Emoji had missing group info

xfalcox:


ZogStriP:
@xfalcox did you do that manually? I wonder if there's a way to automate that.

PR #4467, opened 2016-09-27
FEATURE: Advanced Search UI

cpradio:
Forum Topic:
https://meta.discourse.org/t/expandable-advanced-search-ui-panel/48440

There are some outstanding issues, but I wanted to get the ball rolling on feedback as I'm close to being done (see topic for known issues).

### DIFF
--- a/app/assets/javascripts/discourse/components/badge-selector.js.es6
+++ b/app/assets/javascripts/discourse/components/badge-selector.js.es6
@@ -6,7 +6,7 @@ export default Ember.Component.extend({
     return placeholderKey ? I18n.t(placeholderKey) : '';
   },
 
-  @observes('groupNames')
+  @observes('badgeNames')
   _update() {
     if (this.get('canReceiveUpdates') === 'true')
       this._initializeAutocomplete({updateData: true});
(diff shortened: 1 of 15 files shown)


cpradio:
Tests are now added and passing. :+1: (except for the tags input, that input is downright evil -- I wasted so much time trying to get tests to work on that input)

SamSaffron:
are these autocomplete changes tested against the other instances (emoji/pm to field, tag, category?)

overall the change look totally fine to me, my biggest issue is that now we are loading up all this extra code in application.js which is already gigantic, but that is an issue for a different topic :)

cpradio:
Yes, I went through and tested all of the admin UI and the front-end UI to ensure it didn't break anything. In fact, the canReceiveUpdates was to prevent issues with the user-selector, category-selector, and group-selector at other UI implementations.

The autocomplete this.single changes were fully tested in all locations too, and none of the other components are utilizing that option (and the ones that were, were not trying to give it an initial value -- otherwise, they would have seen the issue I was correcting).

coding-horror:
Hopefully we can review and merge this today? Sam?

SamSaffron:
OK lets try it out ! thanks @cpradio  for all your hard work!

cpradio:
Sweet. I'll follow up with any bug reports that escape my own testing. Looking forward to seeing this live!

PR #4469, opened 2016-09-28
Uses higher resolution pictures when importing avatars from Facebook

xfalcox:


xfalcox:
We are using 50x50 pictures when importing photos from Facebook atm, and the result is very ugly.

On my tests, using 480 results in better pictures.

ZogStriP:
Is `480` **always** available?

xfalcox:
@ZogStriP Facebook will always return the nearest possible, so both 99999 and 1 work as height:

https://developers.facebook.com/tools/explorer/145634995501895/?method=GET&path=me%2Fpicture%3Fheight%3D99999&version=v2.7

https://developers.facebook.com/tools/explorer/145634995501895/?method=GET&path=me%2Fpicture%3Fheight%1&version=v2.7

xfalcox:
Also, a photo before:

image

And after:

image

xfalcox:
Ok, some people insist on taking picture vertically, and there's no stop :cry: 

I'm making then square by 5bdaaca so we get a face.

PR #4470, opened 2016-09-28
Accept longer avatar_url on SSO records

xfalcox:


SamSaffron:
unlimited seems a bit on the high number ... give it some upper limit for securities sake :)

ZogStriP:
FYI: we use `limit: 1000` on other URL fields

SamSaffron:
done per:

https://github.com/discourse/discourse/commit/eacfdf92cd4686ab591a71405db01a60148b2af3

PR #4471, opened 2016-09-28
FIX: Plugin "admin_js" translations bundle was not fetched.

tgxworld:
@eviltrout Can you review this and see if the fix is OK? :smile_cat:


eviltrout:
Sure looks good to me!

PR #4472, opened 2016-09-28
Move purge_unactivated to PurgeInactive where it belongs

sethherr:
The logic for `purge_unactivated` really doesn't need to be in the `User` class

As a side benefit there is one fewer untested class and there isn't a confusing naming thing with `inactive` vs `unactivated`.

... This commit doesn't include any new code, it's just a minor refactor.


tgxworld:
@sethherr  Thank you for taking the time to submit a PR. However, I don't think that the method names are confusing enough to justify this refactor and will not be accepting this PR. :crying_cat_face:

PR #4473, opened 2016-09-29
FIX: Using ActionDispatch::Request instead of Rack::Request

darfux:
This provides a correct behavior when handling client ip based on reverse proxy, such as Nginx. If we using `Rack::Request`, the internal network clients' IP will not be handled properly as Rack's trusted proxies are hard coded in method [trusted_proxy?](https://github.com/rack/rack/blob/2.0.1/lib/rack/request.rb#L419). While the ActionDispatch's trusted proxies can be configured by config.action_dispatch.trusted_proxies.

For example, with a client address **10.128.129.193** and *config.action_dispatch.trusted_proxies* set to **[/^(127\.0\.0\.1|::1)$/]**, `ActionDispatch::Request.remote_ip` is:

> 10.128.129.193

But the `Rack::Request.ip` is:
> 127.0.0.1


SamSaffron:
however rack::request appears to be cached on ENV very cleanly, reading through [this](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/request/session.rb) it appears there is a lot more magic going on and I am unsure if this is cached or what will happen if we look up ip before rails gets its hands on it in the middleware stack. 

It appears @tenderlove and @raggi are open to making it configurable but have not found a suitable approach per: 

https://github.com/rack/rack/pull/1001

My call would be to allow what @raggi described in the PR:

```
Rails.application.config.rack.env = { 'rack.trusted_proxies' => [..]}
```

In the mean time for Discourse usage we can inherit off Rack::Request and use that for the cases that we need an IP early in the middleware stack ... or even monkeypatch Rack::Request to give it configurable env based trusted_proxy list

darfux:
Thanks for your reply! I'd like to try a workaround of the original Rack::Request then :)

PR #4475, opened 2016-09-29
FIX: Last visit must keep working when changing categories

xfalcox:


SamSaffron:
seems ok, though I feel it may need a bit of a cleanup ...

PR #4477, opened 2016-09-30
FEATURE: Add notification level user preference when replying to a topic

cpradio:
Relevant Discussion
https://meta.discourse.org/t/automatically-set-topics-to-watching-after-posting/21230

and
https://meta.discourse.org/t/automatically-set-topics-to-watching-after-posting/21230/3?u=cpradio

I didn't setup Regular as today, it defaults to Tracking, and figured leaving it defaulting to Tracking is best. If you want me to add Regular, I think that should be simple enough.


cpradio:
Seems the CI failure isn't related to my commit and was marked as SKIP in a later checkin

SamSaffron:
cool lets try this out

PR #4479, opened 2016-10-03
Remove route to action that doesn't exist

oblakeerickson:
Removed the `post "t"` route from the routes file because the
`topics#create` action doesn't exist in the topics controller. New
topics are created in the posts controller.


ZogStriP:
Good finding! 👀

PR #4481, opened 2016-10-07
Fixing apple touch icon and favicon mixup.

LizardM4:
Firefox would load the last `<link rel="icon">` containing the apple touch icon.
Removed also all unnecessary mime type attributes.

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -1,13 +1,10 @@
 <meta name="author" content="">
 <meta name="generator" content="Discourse <%= Discourse::VERSION::STRING %> - https://github.com/discourse/discourse version <%= Discourse.git_version %>">
 <%- if SiteSetting.favicon_url.present? %>
-<link rel="icon" type="image/png" href="<%=SiteSetting.favicon_url%>">
+<link rel="icon" href="<%=SiteSetting.favicon_url%>">
 <%- end %>
 <%- if SiteSetting.apple_touch_icon_url.present? %>
-<link rel="apple-touch-icon" type="image/png" href="<%=SiteSetting.apple_touch_icon_url%>">
-<%- end %>
-<%- if (SiteSetting.apple_touch_icon_url != "/images/default-apple-touch-icon.png") && SiteSetting.apple_touch_icon_url.present? %>
-<link rel="icon" sizes="144x144" href="<%=SiteSetting.apple_touch_icon_url%>">
+<link rel="apple-touch-icon" href="<%=SiteSetting.apple_touch_icon_url%>">
 <%- end %>
 <meta name="theme-color" content="#<%= ColorScheme.hex_for_name('header_background') %>">
 <% if mobile_view? %>


riking:
Hmm, I think that was the point of that code: to get Firefox to use the high resolution version. Are your favicon and touch icon significantly different?

LizardM4:
@riking yes. In particular, they will often be. But I think there are also several other points in favor of these changes.

1. Favicon are often designed with transparency, while transparency is not used in the touch icons (it's replaced with a [black background](https://realfavicongenerator.net/blog/apple-touch-icon-turns-black/)). Design guidelines for touch icons and favicons are in general different (among which, touch icons can be precompose with effects and will generally fill the whole square area, to allow the client device to cut corners, apply shadows etc.).
2. Firefox *does not* use the high resolution version. Firefox uses the *last* `<link rel="icon">` that finds in the source (if you swap the order, it will display favicon rather than hi-res).
3. This doesn't apply to other browsers that just fetch the first icon found. Safari and Chrome will also prever a .ico over a .png.
4. In any case, one should provide a high resolution version of the favicon, not the touch icon, or modify the `rel` attribute. `rel="icon"` will qualify the object as a favicon.
5. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types) states that in case of ties for the favicon, the browser should choose the last matching item in the tree, which would make the presence of a previous favicon in general useless (if all the browsers were to comply).
6. Mobile devices change quickly in resolution, so hardcoding the size of the touch icon is not useful, especially if then the admin uploads picture of a size that doesn't match 144x144; also, the `sizes` attribute helps the browser deciding which resource to load, but in this case, only one resource is available (or at least with an explicit `sizes` attribute). Same applies to the mime type `image/png`; .ico files are still used.
7. The favicon target sizes are always 16x16 (tab, bookmark). This becomes 32x32 for hi-dpi displays. As far as I know, all the other usage cases will compose a touch icon on the desk; so in this case, it doesn't make much sense to provide a high resolution image for downsampling.

coding-horror:
So this is all about Firefox? This seems like a bit of a risky change to me.

LizardM4:
@coding-horror The other way around, risky is the code that this PR fixes. This is a bugfix for [this](https://meta.discourse.org/t/apple-touch-instead-of-favicon-displayed-as-favicon-firefox/32484/11) and [this](https://meta.discourse.org/t/regression-apple-touch-instead-of-favicon-displayed-again-as-favicon-firefox/34151). Moreover, I've tested this.

In short: Discourse on Firefox displays the apple touch icon instead of the favicon (which are different). This fixes it.

@riking argued that this is intended behavior, but I think it doesn't make sense. First of all, the current implementations relies on the correct order of the `<link>` nodes, which *is* risky. Secondly, it is a useless functionality because it provides a "high resolution" variant for an icon used only at 16x16. Third, it's broken, because it doesn't provide the favicon, but the apple touch icon. Fourth, this code "works" only on Firefox, and thanks to a detail of the standard which is adopted only by Firefox, and may be dropped as far as we know, which is even *more* risky. Fourth: if the two icon types were intended to be one the high-res version of the other (which is not correct), why there are two separate settings in Discourse in the first place?

I say: let the favicon be the favicon and the apple touch icon be the apple touch icon. This change will just make Firefox less confused with two icons with different purposes and will make it behave like every browser and like the admin would expect it to.

coding-horror:
Fav icon as .ico is incompatible with all known CDNs, so that's already a
regression here.

SamSaffron:
My concern here @LizardM4 is that we used to give you an option that allowed you to specify a 144x144 icon... this change would kill that feature... 

So... for example ... if you add a site to homescreen on android you will no longer see an icon

LizardM4:
@coding-horror I don't get this. If you are referring to the removal of the `image/png` mime type, I see only advantages with it. You're removing a soft constraint. One day maybe webp or svg will be supported and that's going to be obsolete anyway.

@SamSaffron that is a good point. I tried on Android 5.1.1 and it uses the apple touch icon, but [apparently](https://developer.chrome.com/multidevice/android/installtohomescreen) it is going to be supported only for a short time. Also, it didn't precompose the icon, which means that the apple touch icon will look ugly on an Android device. As a side note regarding hard-coding the sizes, now the recommended size is 192x192.

This increases the number of icons to be produces to three. Favicon, 32x32, Apple touch icon, precomposed or not, >=150x150, and in case it differs, an Android icon at 192x192.

coding-horror:
I am open to restructuring this but it needs to be clear and defensible.
Remember we only support ie11 and newer, android 4.3 and newer, iOS 7/8 and
newer. If we can simplify it and ensure support for modern browsers and
devices that is fine.

PR #4482, opened 2016-10-07
Mbox checking and reporting

pfaffman:
mbox.rb: Improve error checking (e.g., if files not found, say so) and reporting (e.g., more "x per second" messages).


SamSaffron:
sure ... thanks!

PR #4483, opened 2016-10-07
Update DEVELOPMENT-OSX-NATIVE.MD

bekicot:
Changes:
- Re Order the steps
- Remove postgres 9.2 comment on postgres installation via homebrew (homebrew will always install the latest stable version of postgres)
- Remove trailing whitespace
- Update clone instruction
- Adding `rake db:create` before any other database action (as it assume new development)
- Remove therubyracer error fixes as we are not currently using it

There is additional setup which is i think is obsolete. The one related to therubyracer gem. are we still/ever using that ?

### DIFF
--- a/docs/DEVELOPMENT-OSX-NATIVE.md
+++ b/docs/DEVELOPMENT-OSX-NATIVE.md
@@ -14,6 +14,7 @@ If you don't already have a Ruby environment that's tuned to your liking, you ca
 2. Clone the Discourse repo and cd into it.
 3. Run `script/osx_dev`.
 4. Review `log/osx_dev.log` to make sure everything finished successfully.
+5. Jump To [Setting up your Discourse](#setting-up-your-discourse)
 
 Of course, it is good to understand what the script is doing and why. The rest of this guide goes through what's happening.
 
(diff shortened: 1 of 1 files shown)


docs/DEVELOPMENT-OSX-NATIVE.md
```diff
@@ -263,28 +251,44 @@ make install
...
 
-### Prepare your database
+```sh
+brew tap homebrew/versions
+brew uninstall v8
+brew install v8-315
+gem uninstall -a libv8
+gem uninstall -a therubyracer
+gem install libv8 -v '3.16.14.13' -- --with-system-v8
+gem install therubyracer -v '0.12.2' -- --with-v8-dir=$(brew --prefix v8-315)
```
SamSaffron: this can be removed and the libv8 stuff ... we don't use therubyracer anymore and libv8 works fine on mac these days anyway.

bekicot:
@SamSaffron Thanks for the review, i've removed `therubyracer` error fixes.

SamSaffron:
thanks!

SamSaffron:
sure, let's try this out! thanks @bekicot

PR #4485, opened 2016-10-11
FEATURE: Use the top period default for users who have been inactive or are new

cpradio:


ZogStriP:
❤️

cpradio:
@ZogStriP did my changes cause a ton of failures in other tests? I'm seeing a bunch of failures, that I didn't see prior, but I'm not sure 100% why. I think I need to figure that out

PR #4486, opened 2016-10-11
Revert "FEATURE: Use the top period default for users who have been inactive or are new"

ZogStriP:
Reverts discourse/discourse#4485


ZogStriP:
@cpradio seems like this broke the build. Can you have a look?

cpradio:
Yeah, I'm looking into it, trying to figure it out. Not sure why it is happening yet.

PR #4487, opened 2016-10-11
FEATURE: Use the top period default for users who have been inactive or are new

cpradio:
Round 2.

Don't merge yet, I am still checking to see what other tests may be viable

### DIFF
--- a/spec/models/site_setting_spec.rb
+++ b/spec/models/site_setting_spec.rb
@@ -73,29 +73,29 @@ describe SiteSetting do
 
     context "has_enough_top_topics" do
 
-      SiteSetting.topics_per_period_in_top_page = 2
-      SiteSetting.top_page_default_timeframe = 'daily'
-
       before do
-        SiteSetting.expects(:min_redirected_to_top_period).returns(:daily)
+        SiteSetting.topics_per_period_in_top_page = 2
+        SiteSetting.top_page_default_timeframe = 'daily'
+        TopTopic.stubs(:topics_per_period).with(:daily).returns(3)
+        TopTopic.refresh!
       end
 
       it "should_return_a_time_period" do
-        expect(SiteSetting.min_redirected_to_top_period(1.days.ago)).not_to eq(nil)
+        expect(SiteSetting.min_redirected_to_top_period(1.days.ago)).to eq(:daily)
       end
 
     end
 
     context "does_not_have_enough_top_topics" do
 
-      SiteSetting.topics_per_period_in_top_page = 20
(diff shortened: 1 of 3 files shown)


cpradio:
@ZogStriP Okay, the tests pass and my additional tests around the site_setting that triggers when to go to top is passing correctly too.

One thing I was attempting to do but failing (probably for good reason), was to stub `TopTopic.topics_per_period` so the logic within site_setting.rb for `min_redirected_to_top_period` would be fully tested, but I've yet to figure out how to properly do that (not sure if is it really necessary though either).

ZogStriP:
> One thing I was attempting to do but failing (probably for good reason), was to stub TopTopic.topics_per_period so the logic within site_setting.rb for min_redirected_to_top_period would be fully tested, but I've yet to figure out how to properly do that (not sure if is it really necessary though either).

What/how did you try?

cpradio:
Similar to this
https://github.com/discourse/discourse/blob/5dbd6a304bed5400be481d71061d3e3ebb4d6785/spec/models/top_topic_spec.rb#L23-L41

I attempted to fabricate topics and then refresh the TopTopic stats, but I have a feeling that is really about two steps outside of what I really needed to test. Instead forcing the SiteSetting to return what I'd expect when there are enough TopTopics for a period is likely the best solution. (at least this is the conclusion I keep coming back to when I think over this)

spec/models/site_setting_spec.rb
```diff
@@ -69,6 +69,40 @@
...
+
+      SiteSetting.topics_per_period_in_top_page = 2
+      SiteSetting.top_page_default_timeframe = 'daily'
+
+      before do
+        SiteSetting.expects(:min_redirected_to_top_period).returns(:daily)
+      end
+
+      it "should_return_a_time_period" do
+        expect(SiteSetting.min_redirected_to_top_period(1.days.ago)).not_to eq(nil)
```
ZogStriP: I think this is only testing that the stubbing works properly ;) If you comment out the content of the `min_redirected_to_top_period` method, are the tests failing?

cpradio: That is sort of my point. I have to somehow create Topics that will be returned by TopTopic, but all of my attempts failed, it always returned 0 for the count. I've tried setting it up similar to how the TopTopic specs do it, but for whatever reason it doesn't work here. Any ideas on how I can get TopTopic.where.count to return a specific value?

cpradio:
@ZogStriP got it! I finally figured out how to stub the TopTopics method to return what I needed to prove the logic works.

ZogStriP:
Looks good, thanks! 👍

PR #4494, opened 2016-10-13
FIX: Handle Post Time better, deleting search-terms character by char…

cpradio:
…acter doesn't cause search terms to remove entire filter term

More info on this. With the latest corrections applied, if you search for

after:10

Then delete the 0, followed by the 1, the 1 was left in the Advanced Search UI, these changes solve that. Happened with nearly all fields.


cpradio:
@tgxworld this is probably of interest to you.

cpradio:
Going to re-think this. As now updating the UI doesn't remove the filter text.

PR #4495, opened 2016-10-13
FIX: Improve removing advanced filters

cpradio:
Prior:
Entering `test after:5` and then removing the 5 via the search text field would result in the UI not updating

After:
UI updates after half a second

Removing it from the UI, removes it from the search field immediately.


cpradio:
@tgxworld, okay, I've improved the removing of search filter values. I also increased the debounce to 500 so if someone keys in after: they have half a second to enter the value associated to that key before it gets removed in the UI update

cpradio:
Okay, I've altered a lot. I put debounce back to 250, but I changed how findSearchTerm works. This way it works more like search.rb and separates the entire query into blocks to be managed. If we ever add plugin outlets to this template, we can let them register callbacks to let them handle their own search filters.

tgxworld:
@cpradio This is alot better :+1: Good work on the acceptance tests as well :smile_cat:

tgxworld:
@cpradio Please help me to squash your commits as well. Thank you!

cpradio:
@tgxworld sqaushed :+1: 

This should help me set the ground work to support multiple `in:` in the UI, so if this can get merged, it will give me more flexibility on implementing that

tgxworld:
Got a couple of comments to be addressed :grin:

cpradio:
Sorry, can you point out the comments you are referring to? I'm not seeing them...

tgxworld:
Ooo I have to submit the changes for review

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +74,175 @@ export default Em.Component.extend({
...
+      this.set(key, null);
+    }
   },
 
-  findUsername(searchTerm) {
-    const match = this.findSearchTerm(REGEXP_USERNAME_PREFIX, searchTerm);
+  setSearchedTermValueForCategory() {
+    const match = this.findSearchTerm(REGEXP_CATEGORY_PREFIX);
+    if (match.length !== 0) {
+      const existingInput = _.isArray(this.get('searchedTerms.category')) ? this.get('searchedTerms.category')[0] : this.get('searchedTerms.category');
```
cpradio: IIRC, there is something bizarre with how the category-chooser and other select2 items set their values. It can become a string versus an array with one item. I'll try removing it, but that is why the badge and group logic had something like this too.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +74,175 @@ export default Em.Component.extend({
...
+    const match = this.findSearchTerm(regexPrefix);
 
-    return searchTerm.substring(expressionPosition, nextExpressionPosition + expressionPosition + 2).trim().split(' ')[0];
+    if (match.length !== 0) {
+      const userInput = match[0].replace(regexPrefix, '');
+      if (this.get(key) !== userInput) {
+        this.set(key, userInput);
+      }
+    } else {
+      this.set(key, null);
```
cpradio: That makes sense

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Okay, so you are correct. It seems it was the category-selector that gave me different types of results. The category chooser seems to always return the id of the selected category.

tgxworld:
@cpradio  Looks good to me. I've have a closer look on Monday but I want to get this merged so you don't get blocked

tgxworld:
Can you do one last squash? Thank you!

cpradio:
Okay, they are squashed now. Sorry, wish I saw this earlier.

cpradio:
@tgxworld actually, if you want, I can just use this branch for the `in:` changes and just keep refactoring it down to ensure it fits. That way it doesn't block me. It isn't like these refactorings are solving anything critical at this exact moment. I'll leave it up to you though.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +86,204 @@ export default Em.Component.extend({
...
     if (!searchTerm)
-      return "";
+      return [];
 
-    const expressionPosition = searchTerm.search(EXPRESSION);
-    if (expressionPosition === -1)
-      return "";
+    let result = [];
+    const blocks = searchTerm.match(REGEXP_BLOCKS);
+    _.each(blocks, function(block) {
```
tgxworld: We don't have to depend on `underscore.js` if we do something like 

`blocks.forEach(block => { if (block.length !== 0) result.push(block) });`

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +86,204 @@ export default Em.Component.extend({
...
 
-    return searchTerm.substring(expressionPosition, nextExpressionPosition + expressionPosition + 2).trim().split(' ')[0];
+    this.set('searchTermBlocks', result);
   },
 
-  findUsername(searchTerm) {
-    const match = this.findSearchTerm(REGEXP_USERNAME_PREFIX, searchTerm);
+  filterBlocks(regexPrefix) {
+    let result = [];
+    _.each(this.get('searchTermBlocks'), function(block) {
```
tgxworld: `forEach` here too :)

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +86,204 @@ export default Em.Component.extend({
...
+        result.push(block);
+    });
 
-    if (match.length !== 0) {
-      let userInput = match.replace(REGEXP_USERNAME_PREFIX, '');
+    return result;
+  },
+
+  setSearchedTermValue(key, replaceRegEx, matchRegEx = null) {
+    matchRegEx = matchRegEx || replaceRegEx;
```
tgxworld: Perhaps `setSearchedTermValue(key, matchRegEx = replaceRegEx)`

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +86,204 @@ export default Em.Component.extend({
...
-        this.set('searchedTerms.username', userInput);
+    if (match.length !== 0) {
+      const userInput = match[0].replace(replaceRegEx, '');
+      if (this.get(key) !== userInput) {
+        this.set(key, userInput);
       }
-    } else {
-      this.set('searchedTerms.username', null);
+    } else if(this.get(key) !== null) {
+      this.set(key, null);
```
tgxworld: should probably set it back to `''` instead of `null`

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -74,107 +86,204 @@ export default Em.Component.extend({
...
     }
   },
 
-  @observes('searchedTerms.username')
-  updateUsername() {
-    let searchTerm = this.get('searchTerm');
-    const match = this.findSearchTerm(REGEXP_USERNAME_PREFIX, searchTerm);
-    const userFilter = this.get('searchedTerms.username');
+  setSearchedTermSpecialInValue(key, replaceRegEx, matchRegEx = null) {
+    matchRegEx = matchRegEx || replaceRegEx;
```
tgxworld: Ditto: Perhaps `setSearchedTermSpecialInValue(key, matchRegEx = replaceRegEx)`

test/javascripts/acceptance/search-full-test.js.es6
```diff
@@ -190,19 +193,55 @@ test("update category through advanced search ui", assert => {
...
+  click('.search-advanced-options .in-likes');
 
   andThen(() => {
-    assert.ok(exists('.search-advanced-options #s2id_in .select2-choice .select2-chosen:contains("I liked")'), 'has "I liked" populated');
+    assert.ok(exists('.search-advanced-options .in-likes:checked'), 'has "I liked" populated');
     assert.equal(find('.search input.full-page-search').val(), "none in:likes", 'has updated search term to "none in:likes"');
   });
 });
 
+test("update in:private filter through advanced search ui", assert => {
```
tgxworld: Great job on the acceptance tests :+1: It makes refactoring so much easier :smile:

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Um... I don't think that would do what it needs to do. As that would only give you one regex to use in that area, and I definitely need two for one of the conditions. Now if you meant that for `setSearchedTermSpecialInValue`, then yea, I see where I could remove that extra parameter, that was a copy/paste blunder.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Okay, but we still have the dependency due to _.isArray checks also in this file.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Okay, but we still have the dependency due to _.isArray checks also in this file.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: This was actually a mistake. Will remove the third parameter.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Can do... but you actually originally changed searchedTerms.username to be set to null...
https://github.com/discourse/discourse/commit/4c2c21ee2b7cbb9c9f2915f94774372634c875e7

Even though they all default to '' in the _init.

test/javascripts/acceptance/search-full-test.js.es6
cpradio: I added a new test based on visiting the search page without a search query and seeing errors. Should help prevent that going forward.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
tgxworld: Ahh ok ignore this xD

cpradio:
@tgxworld seems the qunit tests timed out. They run great locally, so likely a resource issue with Travis CI.

PR #4496, opened 2016-10-13
FIX: Shrink tag input and use date type for Post Time

cpradio:


tgxworld:
Thank you @cpradio! :smile:

PR #4498, opened 2016-10-13
FIX: Properly name tag intersection route

gdpelican:
Looks like I named the intersection route the same as the tags show route, which makes it impossible to properly reference the existing tags:show route, like this:

```
{{#link-to 'tags.show' tag}}
```


gdpelican:
PS existing tag links seem to be constructed more manually, like this:

```
const href = tagName === "a" ? " href='" + Discourse.getURL("/tags/" + tag) + "' " : "";
```

So this mistake hasn't affected any part of the existing repo.

tgxworld:
Thanks! @gdpelican

gdpelican:
Hey @tgxworld would you mind reverting this? It seems to break the tag intersections page and I've written a little helper function to workaround the named route issue in my plugin.

tgxworld:
@gdpelican  Done :)

PR #4506, opened 2016-10-18
FIX: Videos and audio files were not associated to the post.

tgxworld:
https://meta.discourse.org/t/mp4-being-cleaned-up-even-when-referenced-in-a-topic/51555

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -60,16 +60,8 @@ class CookedPostProcessor
   def keep_reverse_index_up_to_date
     upload_ids = Set.new
 
-    @doc.css("a[href]").each do |a|
-      href = a["href"].to_s
-      if upload = Upload.get_from_url(href)
-        upload_ids << upload.id
-      end
-    end
-
-    @doc.css("img[src]").each do |img|
-      src = img["src"].to_s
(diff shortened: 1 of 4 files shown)


lib/cooked_post_processor.rb
```diff
@@ -60,16 +60,8 @@ def grant_badges
...
-      href = a["href"].to_s
-      if upload = Upload.get_from_url(href)
-        upload_ids << upload.id
-      end
-    end
-
-    @doc.css("img[src]").each do |img|
-      src = img["src"].to_s
-      if upload = Upload.get_from_url(src)
+    @doc.css("a/@href", "img/@src").each do |media|
```
ZogStriP: 👍

riking: Wouldn't this, in combination with the other change, make it so we're checking URLs on domains entirely unrelated to the forum?

tgxworld: This change is a refactor and I don't see how it is changing the original behavior. But you're right that the other change would force us to check all other URLs. :thinking: We can match it against the current `Discourse.base_url`. I also noticed that we can actually just do a `Upload.where(url: [<urls>])` instead of executing many individual `find_by`s here

PR #4507, opened 2016-10-18
Remove user from a group by user email

ming-relax:


SamSaffron:
I am fine with this, but let's call the param `email` not `user_email`

ming-relax:
@SamSaffron I was using `use_email` because `GroupsController#add_members` is using `user_emails` as the parameter to add user to the group. I think it's better to keep the same convention.

tgxworld:
Is the use case for this to remove a user via the API?

ming-relax:
@tgxworld yes

spec/controllers/groups_controller_spec.rb
```diff
@@ -224,6 +224,13 @@
         user.reload
         expect(user.primary_group_id).to eq(nil)
       end
+
+      it "removes by user_email" do
+        xhr :delete, :remove_member, id: group.id, user_email: user.email
```
tgxworld: I usually prefer that we assert that `expect(group.users.count).to eq(1)`. Otherwise if no users are created for some reason, the test will still pass

tgxworld:
Thanks @ming-relax  :+1:

PR #4508, opened 2016-10-18
FEATURE:'No Echo' option for mailing list mode.

kstaikov:
Mailing list mode now includes the 'no echo' option: to only receive emails of posts not created
by you.  If you reply to an email thread in mailing list mode, your reply will not then be echoed
back to you in a duplicate email by the system.

This feature PR is in response to this thread on meta: https://meta.discourse.org/t/getting-email-after-my-own-reply/21158/4


kstaikov:
This feature PR is in response to this thread on meta: https://meta.discourse.org/t/getting-email-after-my-own-reply/21158/4

PR #4509, opened 2016-10-19
Added resolveService function to custom resolver

davidtaylorhq:
Added function which enables Ember services to be used in discourse plugins. See discussion on Meta here: https://meta.discourse.org/t/using-ember-services-in-discourse-plugins/51713/6?u=david_taylor


eviltrout:
Thank you!

PR #4511, opened 2016-10-21
FIX: If the tag is first used in a deleted topic the tag shouldn't be stuck

xfalcox:
This fixes https://meta.discourse.org/t/tag-gets-broken-if-the-first-topic-to-use-it-is-deleted/51701?u=falco

But is only works on Active Record > 4.2.7.1 that hasn't been released yet.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -13,7 +13,7 @@ class Tag < ActiveRecord::Base
   has_many :tag_groups, through: :tag_group_memberships
 
   def self.tags_by_count_query(opts={})
-    q = TopicTag.joins(:tag, :topic).group("topic_tags.tag_id, tags.name").order('count_all DESC')
+    q = Topic.unscoped { TopicTag.joins(:tag, :topic).group("topic_tags.tag_id, tags.name").order('count_all DESC') }
     q = q.limit(opts[:limit]) if opts[:limit]
     q
   end


app/models/tag.rb
```diff
@@ -13,7 +13,7 @@ class Tag < ActiveRecord::Base
   has_many :tag_groups, through: :tag_group_memberships
 
   def self.tags_by_count_query(opts={})
-    q = TopicTag.joins(:tag, :topic).group("topic_tags.tag_id, tags.name").order('count_all DESC')
+    q = Topic.unscoped { TopicTag.joins(:tag, :topic).group("topic_tags.tag_id, tags.name").order('count_all DESC') }
```
tgxworld: Is there a `default_scope` that we don't want on `Topic`?

tgxworld:
Tests are missing 😉

app/models/tag.rb
xfalcox: Yeah, I want to use `deleted` topics

xfalcox:
> Tests are missing :wink:

I can't test because it wont work :stuck_out_tongue_closed_eyes: 

This need a newer version of Active Record 4.2 that hasn't been released yet.

app/models/tag.rb
tgxworld: Hmm can you point me to the default scope? I can't seem to find it xD

tgxworld:
Then we probably shouldn't use a feature that hasn't been released yet for the fix :)

xfalcox:
> Then we probably shouldn't use a feature that hasn't been released yet for the fix :)

Yeah, I found it wasn't released after making the fix :cry: 

I think that a release is on the way so we can just wait for it.

tgxworld:
> I think that a release is on the way so we can just wait for it.

I don't think so. We don't know when the Rails team will release the next minor patch. Is using `unscoped` the only way to fix the bug? I still don't see where the `default_scope` is coming from though.

xfalcox:
>  I still don't see where the default_scope is coming from though.

https://github.com/discourse/discourse/blob/master/app/models/concerns/trashable.rb#L5

> Is using unscoped the only way to fix the bug?

We can always drop to raw SQL, I just didn't think it was urgent enough to justify it, sorry.

app/models/tag.rb
riking: I think the proper name for this is `.with_deleted`, no?

nlalonde: It turns out this case should be handled, but `counter_caches` rarely work.

```
    unused_tags = DiscourseTagging.filter_allowed_tags(
      Tag.where(topic_count: 0),
      guardian,
      { for_input: params[:filterForInput], term: params[:q], category: category, selected_tags: params[:selected_tags] }
    )
```

Tags on deleted topics still have a topic_count of 1. I want to remove the topic_count column from tags.

When a topic is deleted, maybe it should lose its tags. Thoughts?

xfalcox: > When a topic is deleted, maybe it should lose its tags. Thoughts?

I don't think so. That can be a bad surprise for someone. And can't be (easily) undone.

But I understand that this is complicated to implement.

I think we should move to a topic on Discourse to discuss this :smile:

nlalonde:
I made a fix that also will return unused tags from that query in all cases. https://github.com/discourse/discourse/commit/8c9d390cacafce80c622f94ec00661d8a5278bd0

PR #4512, opened 2016-10-23
Makes the composer window shorter to fit Chrome + Gboard on iOS

vibol:
The composer window buttons render below the top of the keyboard, cutting off the buttons and making them difficult to click. On Chrome for iOS, scrolling the composer window does not work. The workaround is to close the keyboard which hides the composer, then swipe back up to reveal the composer and buttons.

This change makes the composer window slightly shorter, which renders the buttons above the keyboard.

Note: As I do not have an Android device, this has not been tested on Android. This has been tested on an Apple iPhone 6s Plus and iPhone 7 Plus.


SamSaffron:
We need some screenshots here of before/after

Gemfile
```diff
@@ -43,7 +43,7 @@ gem 'redis-namespace'
 
 gem 'active_model_serializers', '~> 0.8.3'
 
-gem 'onebox'
+gem 'onebox', git: 'https://github.com/vibol/onebox.git', branch: 'feature/amazon-instant-video'
```
ZogStriP: Also, this should not have been commited ;)

vibol: Ahh yes. Long story short, building an image from non-master didn't work for me. I'll resubmit the pull from a branch with screenshots and the specific commit.

PR #4515, opened 2016-10-26
Allow to process imported content. (add additional callbacks)

mrded:
Hello,

I'm trying to import content into Discourse using `/script/import_scripts/base.rb`, and I found that after creating a post `created_post(post)` is called. 

    def created_post(post)
      # override if needed
    end

That is very nice, because I can trigger some post-migration things.

Unfortunately you have it only for posts. Can I ask you to implement something similar for `Users`, `Topics`, `Categories`, etc?

Thanks.

### DIFF
--- a/script/import_scripts/base.rb
+++ b/script/import_scripts/base.rb
@@ -146,6 +146,10 @@ class ImportScripts::Base
     admin
   end
 
+  def created_group(group)
+    # override if needed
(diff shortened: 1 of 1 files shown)


nlalonde:
Thanks!

mrded:
Wow, so quick! :)

PR #4516, opened 2016-10-26
UX: Rename search "options" to "Advanced Search"

gschlager:
This renames the "options" link in the search popup to "Advanced Search" since it opens the Advanced Search instead of showing the old search help text.

Before:
image

After:
image


coding-horror:
We want it to be "options", we don't want "Advanced Search" here.

gschlager:
OK, sorry. As a non-native speaker I just found it kinda odd...

PR #4522, opened 2016-10-28
FEATURE: Allow posting a link with the topic

fantasticfears:
Tracking on https://meta.discourse.org/t/merge-discourse-links-category-into-discourse/51483

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -282,4 +282,15 @@ module ApplicationHelper
     result.html_safe
   end
 
+  def topic_featured_link_domain(link)
+    begin
+      uri = URI.encode(link)
+      uri = URI.parse(uri)
+      uri = URI.parse("http://#{uri}") if uri.scheme.nil?
+      host = uri.host.downcase
+      host.start_with?('www.') ? host[4..-1] : host
+    rescue
+      ''
+    end
+  end
 end
--- a/app/serializers/topic_list_item_serializer.rb
+++ b/app/serializers/topic_list_item_serializer.rb
@@ -10,7 +10,8 @@ class TopicListItemSerializer < ListableTopicSerializer
              :pinned_globally,
              :bookmarked_post_numbers,
              :liked_post_numbers,
-             :tags
+             :tags,
+             :featured_link
 
   has_many :posters, serializer: TopicPosterSerializer, embed: :objects
   has_many :participants, serializer: TopicPosterSerializer, embed: :objects
@@ -70,6 +71,14 @@ class TopicListItemSerializer < ListableTopicSerializer
 
   def tags
     object.tags.map(&:name)
+  end
+
+  def include_featured_link?
+    SiteSetting.topic_featured_link_enabled
+  end
+
+  def featured_link
+    object.featured_link
   end
 
 end
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1074,6 +1074,7 @@ en:
       title_placeholder: "What is this discussion about in one brief sentence?"
       edit_reason_placeholder: "why are you editing?"
       show_edit_reason: "(add edit reason)"
+      topic_featured_link_placeholder: "Enter link shown with title."
       reply_placeholder: "Type here. Use Markdown, BBCode, or HTML to format. Drag or paste images."
       view_new_post: "View your new post."
       saving: "Saving"
@@ -1875,6 +1876,7 @@ en:
       tags_allowed_tag_groups: "Tag groups that can only be used in this category:"
       tags_placeholder: "(Optional) list of allowed tags"
       tag_groups_placeholder: "(Optional) list of allowed tag groups"
+      topic_featured_link_allowed: "Restricts editing the topic featured link in this category. Require site setting topic_featured_link_enabled is checked."
       delete: 'Delete Category'
       create: 'New Category'
       create_long: 'Create a new category'
--- a/lib/guardian/category_guardian.rb
+++ b/lib/guardian/category_guardian.rb
@@ -68,4 +68,9 @@ module CategoryGuardian
   def topic_create_allowed_category_ids
     @topic_create_allowed_category_ids ||= @user.topic_create_allowed_category_ids
   end
+
+  def topic_featured_link_allowed_category_ids
+    @topic_featured_link_allowed_category_ids = CategoryCustomField.where(name: "topic_featured_link_allowed", value: "true")
+                                                                   .pluck(:category_id)
+  end
 end
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -5,6 +5,7 @@ require_dependency 'topic_creator'
 require_dependency 'post_jobs_enqueuer'
 require_dependency 'distributed_mutex'
 require_dependency 'has_errors'
+require_dependency 'discourse_featured_link'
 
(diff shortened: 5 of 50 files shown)


app/helpers/application_helper.rb
```diff
@@ -262,4 +262,11 @@ def server_plugin_outlet(name)
     result.html_safe
   end
 
+  def topic_featured_link_domain(link)
+    uri = URI.encode(link)
+    uri = URI.parse(uri)
+    uri = URI.parse("http://#{uri}") if uri.scheme.nil?
```
tgxworld: I believe `URI.parse/encode` raises an error if the URL is not valid.

app/serializers/topic_list_item_serializer.rb
```diff
@@ -72,4 +75,12 @@ def tags
     object.tags.map(&:name)
   end
 
+  def include_featured_link?
+    SiteSetting.topic_featured_link_enabled
+  end
+
+  def featured_link
+    object.custom_fields[DiscourseFeaturedLink::CUSTOM_FIELD_NAME]
```
tgxworld: Lets extract this `.custom_fields[DiscourseFeaturedLink::CUSTOM_FIELD_NAME]` into a method under the model

lib/guardian/category_guardian.rb
```diff
@@ -68,4 +68,10 @@ def allowed_category_ids
   def topic_create_allowed_category_ids
     @topic_create_allowed_category_ids ||= @user.topic_create_allowed_category_ids
   end
+
+  def topic_featured_link_allowed_category_ids
+    @topic_featured_link_allowed_category_ids = CategoryCustomField
+                                                    .where(name: "topic_featured_link_allowed", value: "true")
+                                                    .pluck(:category_id)
```
tgxworld: The indentation here is off.

lib/post_revisor.rb
```diff
@@ -95,6 +96,13 @@ def self.track_topic_field(field, &block)
     end
   end
 
+  track_topic_field(:featured_link) do |tc, featured_link|
```
tgxworld: I would avoid using variable names like `tc` because it degrades the readability of the code.

spec/components/guardian_spec.rb
```diff
@@ -2280,4 +2280,25 @@
...
+    before { SiteSetting.topic_featured_link_enabled = true }
+    let(:guardian) { Guardian.new }
+
+    it 'returns true if no category restricts editing link' do
+      expect(guardian.can_edit_featured_link(nil)).to eq(true)
+      expect(guardian.can_edit_featured_link(5)).to eq(true)
+    end
+
+    it 'returns true if the category is listed in the list' do
+      guardian.expects(:topic_featured_link_allowed_category_ids).returns([1,2,3]).at_least_once
```
tgxworld: Lets setup the factories here instead of stubbing.

spec/components/validators/post_validator_spec.rb
```diff
@@ -5,6 +5,21 @@
   let(:post) { build(:post) }
   let(:validator) { Validators::PostValidator.new({}) }
 
+  context "post body validation" do
+    it "adds an error for empty raw" do
+      post.raw = ""
+      validator.post_body_validator(post)
+      expect(post.errors.count).to eq(2)
```
tgxworld: I would assert for the right error messages here instead of just the count otherwise it may result in false positives.

spec/models/topic_spec.rb
```diff
@@ -1724,4 +1724,30 @@ def build_topic_with_title(title)
...
+    before { SiteSetting.topic_featured_link_enabled = true }
+
+    it 'can validate featured link' do
+      topic = Fabricate.build(:topic, featured_link: ' invalid string')
+
+      expect(topic).not_to be_valid
+      expect(topic.errors[:featured_link]).to be_present
+    end
+
+    it 'can properly save the featured link' do
```
tgxworld: I don't think this test is actually saving the record?

nlalonde: Looks like it is on line 1742.

fantasticfears: I used `valid?` there but it's begin overwritten.

config/locales/client.en.yml
```diff
@@ -1861,6 +1862,7 @@ en:
       tags_allowed_tag_groups: "Tag groups that can only be used in this category:"
       tags_placeholder: "(Optional) list of allowed tags"
       tag_groups_placeholder: "(Optional) list of allowed tag groups"
+      topic_featured_link_allowed: "Only allow posting a featured link with topics in the enabled category"
```
nlalonde: I don't understand what this sentence means. Also, I tried enabling and disabling it in a category, but couldn't figure out what it does.

fantasticfears:
So it should be tuned with new Ember code now. And I added a new site setting `topic_featured_link_style` to replace an old site setting (was in this pr) which was used to allow empty post body. It basically enables post with an extra link; only link which will hide and collapse composer; the last one is showing the onebox which overrides content with the onebox and also hide and collapse the composer. That is all 3 possible options in the plugin.

As of the onebox style, it basically copies the link into the post raw. In which, it will show two links in the digest. One is with title as this PR adds. The other is the post body, a plain link. I would like to just remove this repeated link in the body. @nlalonde, I tried to look into `email_excerpt`. It invokes `PrettyText.format_for_email` in which trigger `reduce_cooked`. I found `email_excerpt` may gets called with `html='', posts_count=nil` which is weird. And I can't find where it's called...Could you give me a hint so I could expose the post and html for the `reduce_cooked`?

Another really weird parts could be 1) have to change and rebake post in the post_revisor; 2) client test, well, it's quite complicated when category restrictions are also taken into account.

config/locales/client.en.yml
fantasticfears: `SiteSetting.topic_featured_link_enabled` operates per site. Once a `topic_featured_link_allowed` is checked, it becomes per category.

fantasticfears: Rephrased as "Restricts editing the topic featured link in this category. Require site setting topic_featured_link_enabled is checked."

nlalonde: Looks better, thanks! Jeff is good with words, so he can try to improve it later.

nlalonde:
The `email_excerpt` code is only called from the [digest html erb](https://github.com/discourse/discourse/blob/master/app/views/user_notifications/digest.html.erb). I tested it and it seems to work fine in the digest email.

lib/post_creator.rb
```diff
@@ -366,6 +366,9 @@ def update_topic_auto_close
 
   def setup_post
     @opts[:raw] = TextCleaner.normalize_whitespaces(@opts[:raw] || '').gsub(/\s+\z/, "")
+    if SiteSetting.topic_featured_link_enabled && SiteSetting.topic_featured_link_style == 'onebox'
+      @opts[:raw] = DiscourseFeaturedLink.cache_onebox_link(@opts[:featured_link])
```
nlalonde: In a category that doesn't allow featured link topics, and the post raw is one line with a link to a jpg (for example), this line will set raw to nil and exception will be thrown.

fantasticfears:
I'll move find_category logic in the beginning of PostCreator. I'll comment when I done.

fantasticfears:
@nlalonde The pull request is ready.

PR #4523, opened 2016-10-30
FIX: Auto-link URLs that are inside parentheses

mcwumbly:
https://meta.discourse.org/t/links-in-parentheses-are-not-hyperlinked/24121


mcwumbly:
Hmm... looks like I wasn't running all the right tests locally... will take a look at these failures...

mcwumbly:
ok, we're green and ready to review ✅ 🔍

ZogStriP:
Thanks 👍

PR #4524, opened 2016-10-31
PERF: `NOT IN` query is really inefficient when there are a lot of re…

tgxworld:
…cords.

@SamSaffron @ZogStriP Would love to have more eyes on this. The query currently goes through all the rows in the `uploads` table making it very expensive to run.

`Array#-` under the hood will still have to iterate through all the entire array but it coverts the second array into a hash making lookups fast.

### DIFF
--- a/app/jobs/scheduled/clean_up_uploads.rb
+++ b/app/jobs/scheduled/clean_up_uploads.rb
@@ -5,31 +5,29 @@ module Jobs
     def execute(args)
       return unless SiteSetting.clean_up_uploads?
 
-      ignore_urls  = []
-      ignore_urls |= UserProfile.uniq.where("profile_background IS NOT NULL AND profile_background != ''").pluck(:profile_background)
-      ignore_urls |= UserProfile.uniq.where("card_background IS NOT NULL AND card_background != ''").pluck(:card_background)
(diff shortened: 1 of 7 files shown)


ZogStriP:
Could that instead be solved with a view or a temp table?

tgxworld:
@ZogStriP Hmm I'm not seeing how a view or a temp table will solve the problem here since the query will still have to be executed.

ZogStriP:
They can be indexed? ;)

tgxworld:
I think I'm missing something here. How will the index help with the query? :thinking:

ZogStriP:
That was just a non-☕️  idea :P

xfalcox:
Looks like this query can benefit greatly from a index on created_at field.

tgxworld:
> Looks like this query can benefit greatly from a index on created_at field.
@xfalcox  I tried this out locally and it seems like Postgres will still decide to do a seq scan even with an index on `created_at`

xfalcox:
> I tried this out locally and it seems like Postgres will still decide to do a seq scan even with an index on created_at

A `seq_scan` will always be used if the filter is targeting too much of the table (IE >30%). Did you had a copy from a big forum or just dummy data? It's strange because this is supposed to scan so little of trhe table.

tgxworld:
@xfalcox dump of a large forum

PR #4525, opened 2016-10-31
FEATURE: include post image in OpenGraph image tag

techAPJ:
Currently when sharing a Discourse post link on Twitter/Facebook the poster avatar image shows up. This PR makes it so that if the post contains image, that image will show up instead. In case of multiple images first image will be picked.

### DIFF
--- a/app/models/color_scheme.rb
+++ b/app/models/color_scheme.rb
@@ -153,4 +153,6 @@ end
 #  version      :integer          default(1), not null
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
+#  via_wizard   :boolean          default(FALSE), not null
+#  theme_id     :string
 #
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -84,7 +84,7 @@ class CookedPostProcessor
       convert_to_link!(img)
     end
 
-    update_topic_image
+    update_post_image
   end
 
   def extract_images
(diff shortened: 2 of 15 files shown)


app/models/color_scheme.rb
```diff
@@ -153,4 +153,6 @@ def dump_hex_cache
 #  version      :integer          default(1), not null
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
+#  via_wizard   :boolean          default(FALSE), not null
```
ZogStriP: This doesn't look like it's related to that commit. Can you add them in another commit instead?

lib/cooked_post_processor.rb
```diff
@@ -285,10 +285,11 @@ def create_span_node(klass, content=nil)
     span
   end
 
-  def update_topic_image
-    if @post.is_first_post?
-      img = extract_images_for_topic.first
-      @post.topic.update_column(:image_url, img["src"][0...255]) if img["src"].present?
+  def update_post_image
+    img = extract_images_for_post.first
```
ZogStriP: Shouldn't you iterate through all images until at least one as a valid "src" attribute?

techAPJ: The `extract_images_for_post` method filters the image src as per this code: https://github.com/techAPJ/discourse/blob/382803cb052d5251e2a5be2d2b4f18c2c3a72002/lib/cooked_post_processor.rb#L103-L112

That should suffice right?

ZogStriP: You're right. Not sure the `if img["src"].present?` is useful then?

techAPJ: The `if img["src"].present?` condition is required to verify that image is present in topic/post before updating `image_url` field.

Also, that condition was already present in previous code for topics, I was just making sure that it applied to posts as well 😄

PR #4526, opened 2016-10-31
Add outlet after navigation bar

gdpelican:
We're using it for Babble like this:

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/navigation-bar.hbs
+++ b/app/assets/javascripts/discourse/templates/components/navigation-bar.hbs
@@ -1,4 +1,5 @@
 {{#each navItems as |navItem|}}
   {{navigation-item content=navItem filterMode=filterMode}}
 {{/each}}
+{{plugin-outlet "after-navigation-bar" tagName="li"}}
 {{custom-html "extraNavItem"}}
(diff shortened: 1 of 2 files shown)


SamSaffron:
the issue there though is that it will wrap the `li` in a div no? maybe add the outlet after the ul?

gdpelican:
Hm, that's not the behaviour I'm seeing:

screen shot 2016-10-31 at 5 07 40 pm

Although to make it go I did have to make my component into an `a` tag, rather than `div` or `li`:

```
// connector for after-navigation-bar
{{chat-nav action="toggleChat"}}
```
```
// chat-nav.js.es6
export default Ember.Component.extend({
  tagName: 'a',
  classNameBindings: ['active']
})
```

SamSaffron:
try adding 2 of these ? it has got to wrap it in a div ...

gdpelican:
Hmmm, ick. So it's not feasible to expand out 2 plugins in the same outlet to

```
[outlet]
<li class="one"></li>
<li class="two"></li>
[/outlet]
```

and needs to be

```
[outlet]
<div>
  <li class="one"></li>
  <li class="two"></li>
</div>
[/outlet]
```
?

Come to think of it, I have a similar issue in the admin panel with Lattice and Babble installed at the same time:
screen shot 2016-10-31 at 5 35 49 pm


From my end, I think it's trouble to orchestrate things like the `active` class from outside the `ul`, and will have to duplicate styles and such there.

Anyway maybe I'll just need to follow [this conversation](https://meta.discourse.org/t/adding-items-to-navigation-bar/26830/8) with interest (and would happy to contribute some code towards it)

SamSaffron:
I think best thing to do here is just add the outlet after the list, then you can CSS it into position.

SamSaffron:
actually you are right, just using navigation item direct and adding to the list at the right time may do the trick

gdpelican:
I just got this to go without using outlets. Thanks for the nudge!

PR #4527, opened 2016-10-31
FIX: Advanced Search to utilize Category Selector so it can distingui…

cpradio:
…sh uncategorized from no category selected

### DIFF
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -60,7 +60,7 @@ export default Em.Component.extend({
     this.setProperties({
       searchedTerms: {
         username: '',
-        category: null,
+        category: '',
         group: [],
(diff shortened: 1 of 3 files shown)


cpradio:
@tgxworld, you will notice, I changed searchedTerms.category to '' instead of null, as you can see, username does that too, it seems autocomplete likes empty string over null, otherwise, it kicks all sorts of errors and I didn't want to mess with the category selector too much in fear of breaking other usages.

cpradio:
Should the Time Out be increased for qunit's rake task? I feel like all of my checkins keep hitting it.

cpradio:
Ah, is that a way to trick it to re-run the tests? I've been wondering how that can be done :+1:

tgxworld:
Yup :)

tgxworld:
@cpradio  Looks good to me :+1:  Thank you for the work :smile:

PR #4529, opened 2016-11-01
new importer: Mylittleforum

pfaffman:
This is an importer for [my little forum](http://www.mylittleforum.net/); here is the [repo](https://github.com/ilosuna/mylittleforum) for mylittle forum.


pfaffman:
Oops. Let me commit again without a zillion debug versions.

PR #4530, opened 2016-11-01
new importer: my little forum

pfaffman:
This is an importer for [my little forum](http://www.mylittleforum.net/)  ([github](https://github.com/ilosuna/mylittleforum)).

It imports users, categories, topics, posts, and creates permalinks for users, posts, and categories.


xfalcox:
Thanks @pfaffman !

PR #4531, opened 2016-11-03
Improve summing of poll percentages to 100

rizka10:
Makes a function that forces poll percentages to sum to 100 to make better choices.


ZogStriP:
Thanks 👍

PR #4536, opened 2016-11-04
FIX: missing post and topic edited webhooks

fantasticfears:


ZogStriP:
No test?

fantasticfears:
Sorry, fixed!

lib/post_revisor.rb
```diff
@@ -405,6 +405,7 @@ def is_last_post?
   def plugin_callbacks
     DiscourseEvent.trigger(:before_edit_post, @post)
     DiscourseEvent.trigger(:validate_post, @post)
+    DiscourseEvent.trigger(:before_edit_topic, @topic, @editor) if topic_changed?
```
tgxworld: I think `:before_edit_post` is sufficient here and we don't need another event trigger. We can easily check if it is a topic by checking if the `post_number` is one.

tgxworld: I think `:before_edit_post` is sufficient here and we don't need another event trigger. We can easily check if it is a topic by checking if the `post_number` is one.

fantasticfears: Somehow it's needed to limit the scope of webhook, `post_number` is not only one condition, but also a `topic_changed?`. I'll need some way to check whether topic is changed. But `PostRevisor` is a huge object, I don't see any spots sending the whole object. So I would only send the single value in the event. Though it looks a little bit suspicious comparing others.

app/models/web_hook.rb
```diff
@@ -57,18 +61,16 @@ def self.enqueue_topic_hooks(event, topic, user)
...
-        user_id: user&.id,
-        category_id: post.topic&.category&.id,
-        event_name: event.to_s
-      )
+      WebHook.enqueue_post_hooks(event, post, user)
     end
   end
 
+  DiscourseEvent.on(:before_edit_post) do |post, topic_changed|
+    WebHook.enqueue_post_hooks(:before_edit_post, post)
```
tgxworld: hmm shouldn't the event name be like `post edited` or `topic edited`?

fantasticfears: I was worried they may interested in where the event initially happens. But it seems they just need to distinguish them..so goes for `x_edited`

app/models/web_hook.rb
```diff
@@ -57,18 +61,16 @@ def self.enqueue_topic_hooks(event, topic, user)
...
-        topic_id: post&.topic&.id,
-        user_id: user&.id,
-        category_id: post.topic&.category&.id,
-        event_name: event.to_s
-      )
+      WebHook.enqueue_post_hooks(event, post, user)
     end
   end
 
+  DiscourseEvent.on(:before_edit_post) do |post, topic_changed|
```
tgxworld: Hmm sorry I just realized that this event is called before the transaction has been committed. We probably want to add event triggers for after the post has been committed to the DB

fantasticfears:
New DiscourseEvent `post_edited` added.

app/models/web_hook.rb
```diff
@@ -40,10 +40,14 @@ def self.enqueue_hooks(type, opts = {})
     end
   end
 
-  def self.enqueue_topic_hooks(event, topic, user)
+  def self.enqueue_topic_hooks(event, topic, user=nil)
     WebHook.enqueue_hooks(:topic, topic_id: topic.id, user_id: user&.id, category_id: topic&.category&.id, event_name: event.to_s)
   end
 
+  def self.enqueue_post_hooks(event, post, user=nil)
+    WebHook.enqueue_hooks(:post, post_id: post.id, topic_id: post&.topic&.id, user_id: user&.id, category_id: post&.topic&.category&.id, event_name: event.to_s)
```
tgxworld: `post&.topic&.category&.id` similarily this can be `post&.topic.category_id`

PR #4538, opened 2016-11-06
A new guard for changing post timestamps

kgish:
Currently the guard used within the topics controller for handling `/ /t/:topic_id/change-timestamp` requests uses the `can_change_post_owner?` guard.

In order to make this more flexible and allow specific guarding of only timestamp changes, I've introduced a new guard called `can_change_post-timestamps?`


SamSaffron:
Sure this seems fine, thanks

PR #4544, opened 2016-11-11
FIX: open login modal fails because of missing parameters

fantasticfears:
This may fail when auto login starts. Introduced from the refactor: https://github.com/discourse/discourse/commit/56f07529bb318aefbb582f8a317d50d07e347ee2#diff-1919f3193c69594c86ee146d84448c23R10

### DIFF
--- a/app/assets/javascripts/discourse/components/login-buttons.js.es6
+++ b/app/assets/javascripts/discourse/components/login-buttons.js.es6
@@ -1,4 +1,5 @@
 import { findAll }  from 'discourse/models/login-method';
+import { getOwner } from 'discourse-common/lib/get-owner';
 
 export default Ember.Component.extend({
   elementId: 'login-buttons',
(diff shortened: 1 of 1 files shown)


tgxworld:
@fantasticfears  I don't seem to be able to reproduce the bug. Can you let me know what the steps are? Thank you!

fantasticfears:
1. Setup a twitter oauth2 login
2. SiteSetting.invite_only=true
3. Once twitter end is authorized, the login modal won't appear since `findAll` fails.

`findAll` in `login-method.js.es6` makes use of capabilities property. I didn't add a check there because in the `application.js.es6` a call just provide the sufficient parameters. Also, `controllers/login.js.es6` seems to have an unused property `hasAtLeastOneLoginButton` using the `findAll`.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/0fbb3fb02bdc852c82d10f07222418c21531a3d0 

Thanks @fantasticfears

PR #4545, opened 2016-11-11
Add “pop3_polling_openssl_verify” setting to turn off TLS server certificate verification

amotl:
We needed a new setting like “smtp_openssl_verify_mode”, only for controlling the same thing for fetching emails using POP3 when running a mail server with self-signed certificates.
The new setting defaults to “true”, so it does not change the current behavior.


ZogStriP:
Thanks 👍

PR #4519, opened 2016-10-27
display only 1 trust level badge on user card

LeoMcA:
Discussion: https://meta.discourse.org/t/is-it-possible-to-display-only-one-trust-level-badge-on-the-user-card/27714/8?u=leomca

PR #4549, opened 2016-11-14
FIX: display only 1 trust level badge on user card

LeoMcA:
https://github.com/discourse/discourse/pull/4519#issuecomment-260461982


xfalcox:
Thanks!

PR #4550, opened 2016-11-15
FEATURE: Notify user when mention can't see the reply they were mentioned in

cpradio:
FIX: Group Mention Notifications

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-editor.js.es6
@@ -137,9 +137,10 @@ export default Ember.Component.extend({
   },
 
   _renderUnseenMentions($preview, unseen) {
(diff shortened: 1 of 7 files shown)


cpradio:
Seems to have failed due to qunit timing out again.

cpradio:
Also, I should mention that this only works for replies to an existing topic/PM. I had a PM conversation with @erlend-sh and @coding-horror about this, and the complexity greatly increases when you expand to include the initial topic.

For regular topics, it means you have to search the category permissions compared to the groups the user mentioned is in (and you have to do that each time they change the category while composing the topic). For PMs, it is a matter of comparing against the recipient/user list (not nearly as complex, but considering you can see that list while composing... seems redundant to check against it)

I can invite anyone who wants to be a part of that discussion to it on Meta, just let me know.

PR #4554, opened 2016-11-16
FIX: Track first notification read using Redis.

tgxworld:


app/models/user.rb
```diff
@@ -343,9 +343,32 @@ def saw_notification_id(notification_id)
...
   end
 
+  TRACK_FIRST_NOTIFICATION_READ_DURATION = 1.week.to_i
+
+  def self.first_notification_read_key(user)
+    "#{user.id}:first-notification-read"
+  end
+
+  def mark_first_notification_read
+    first_notification_read_key = User.first_notification_read_key(self)
```
SamSaffron: nahh this needs to go to the database on redis miss. otherwise when you junk redis there are problems.

PR #4558, opened 2016-11-17
FIX: activy column title shows the non-i18n datetime

fantasticfears:
`longDate()` utilizes a constant format string `MMMM D, YYYY h:mma` which can't be changed. The title of the activity column in topic list item will be strange in some language.


ZogStriP:
👍

PR #4561, opened 2016-11-18
Add keyPress hook to widgets

gdpelican:
I'm specifically looking for keyPress, as opposed to keyUp or keyDown, because I want to listen only for character inputs.

screen shot 2016-11-18 at 3 11 16 pm


tgxworld:
Key press event isn't supported on Android Chrome. Plus the specification states that key press is deprecated as well.

> The keypress event type is defined in this specification for reference and completeness, but this specification deprecates the use of this event type. When in editing contexts, authors can subscribe to the beforeinput event instead.

https://www.w3.org/TR/DOM-Level-3-Events/#event-type-keypress

PR #4562, opened 2016-11-19
Widget hook for 'input' event

gdpelican:
Hm, okay; how about the 'input' event, which seems to have wider support?
https://developer.mozilla.org/en-US/docs/Web/Events/input

Is 'input' too generic of a function name to be pointing the widget event to? (I wish 'beforeinput' were actually a thing.. seems to have fallen of the face of the internet.)


gdpelican:
Bump? @eviltrout is this a thing worth doing, or do you have thoughts on how to listen for keyboard input in widgets?

eviltrout:
Sorry, I missed this because I was confused between this and the PR that was closed. Can you tell me exactly why you need this new event?

gdpelican:
Sure. I'm listening for input on the Babble composer from clients, so that I can tell the other clients who is typing at any given moment.

'keyup' broadcasts _all_ key events, including shift, tab, alt, and other not-actually-typing things. That leads to annoying things like 'it says you're typing when actually you were just in the composer and alt-tabbed away'

'input', on the other hand, only triggers when the input value itself has changed, which is the behaviour I'm looking for out of the box. Maybe I'm missing an obvious solution here?

eviltrout:
Couldn't you use keyUp then compare the value to the previous value?

I'm just a little weary of adding so many event bindings.

PR #4563, opened 2016-11-19
Added flatten func for applyDecorators

vinothkannans:
flatten function added for applyDecorators of footerLinks same like [generalLinks](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/widgets/hamburger-menu.js.es6#L110). Else custom footerLinks not working in some cases.


ZogStriP:
Thanks 👍

PR #4564, opened 2016-11-22
FEATURE: Add job to clean up Redis keys that are not used.

tgxworld:
@SamSaffron Can you review this? Thank you!

I feel like MessageBus needs to have a TTL on all the keys that it sets such that keys don't stay in the db forever.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -1001,6 +1001,8 @@ developer:
   wizard_enabled:
     default: true
     hidden: true
+  clean_up_redis_keys:
+    default: true
 
 embedding:
   feed_polling_enabled:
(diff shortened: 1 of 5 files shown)


config/site_settings.yml
```diff
@@ -1001,6 +1001,8 @@ developer:
   wizard_enabled:
     default: true
     hidden: true
+  clean_up_redis_keys:
```
ZogStriP: You're missing the description of that site setting in `server.en.yml` ;)

SamSaffron:
I don't know, I don't think this belongs so deep in Discourse, no need for a site setting here or job. 

Just make a rake task for this and we can schedule it externally on our multisites. This change is particular to discourse hosters, no need to confuse admins with this.

coding-horror:
Should,this PR be closed then?

tgxworld:
Moved to a Rake task in https://github.com/discourse/discourse/commit/ce36f54dcd570145cd0d34be425692af1771c873

PR #4568, opened 2016-11-24
Add support for Sprockets 4

sgrif:
Note: This PR shouldn't be merged until the changes proposed to `ember-rails` can get merged upstream and the Gemfile can stop pointing at my fork.

We are continuing to use the Discourse special ES6 transpiler, as Babel
no longer supports decorators pending a spec change, and the baked in
support from Sprockets 4 doesn't allow the level of customization of the
babel installation that is required to get the Discourse JS code to
compile.

Beyond that, the migration was mostly just updating some interfaces to
use the Sprockets 4 API (the definitions of `self.call` on the `Tilt`
templates was directly lifted from the `LegacyTiltHandler` in Sprockets
3).

The jQuery files were renamed, as Sprockets appears to be having issues
with file names that contain a dot. I'm unsure if this is intentional or
a bug.


app/assets/config/manifest.js
```diff
@@ -0,0 +1,21 @@
+//= link_tree ../images
```
SamSaffron: This is super weird, putting images, css and fonts all into a `.js` manifest. 

Is the api in sprockets right? should there be multiple manifest files?

app/assets/javascripts/jquery_include.js.erb
```diff
@@ -1,7 +1,7 @@
 <%
 if Rails.env.development? || Rails.env.test?
-  require_asset ("jquery.debug.js")
```
SamSaffron: to me this restriction looks like a bug in sprockets @schneems

config/application.rb
```diff
@@ -64,16 +64,7 @@ def config.database_configuration
     # Allows us to skip minifincation on some files
     config.assets.skip_minification = []
 
-    # explicitly precompile any images in plugins ( /assets/images ) path
-    config.assets.precompile += [lambda do |filename, path|
-      path =~ /assets\/images/ && !%w(.js .css).include?(File.extname(filename))
```
SamSaffron: by moving this to manifest we lost some of the explicitness here ... this is only looking at non js/css files

SamSaffron:
can you review precompilation still works:

```
RAILS_ENV=production bin/rake assets:precompile
``` 

There are lots of hacks here: https://github.com/discourse/discourse/blob/master/lib/tasks/assets.rake

app/assets/config/manifest.js
sgrif: Yes, this is an API in sprockets. I don't think multiple manifest files are a thing

app/assets/config/manifest.js
```diff
@@ -0,0 +1,21 @@
...
+//= link_tree ../fonts
+//= link vendor.js
+//= link common.css
+//= link desktop.css
+//= tink mobile.css
```
sgrif: I BELIEVE (and also have no clue how I missed that)

app/assets/javascripts/jquery_include.js.erb
sgrif: I agree

config/application.rb
sgrif: Sure, technically we will link js/css files in `/assets/images`, but is that really a concern?  We can add explicit extensions if needed.

sgrif:
Hm. The CI failure is one that I've seen locally before, but haven't been able to reproduce lately. I believe it is order dependent. I will investigate further.

SamSaffron:
let me know once this is ready, I want to get it merged in as soon as you feel it is safe.

schneems:
I'm going to be free in the mornings next week. I'll be in SF for an onsite but I can still chat PST. Want to do a hangout?

sgrif:
That would be amazing. I could use your help with the failures I'm seeing on https://github.com/emberjs/ember-rails/pull/532 as well

coding-horror:
Any movement on this @sgrif ?

sgrif:
:soon: :tm: (Will have some movement tomorrow)

sgrif:
This should be good to merge assuming CI passes. `RAILS_ENV=production bundle exec rake assets:precompile` appears to work for me locally. This is still pointing at my fork of ember-rails, but their CI failures are going to take a while to pin down. Their CI is green on Rails 4.2 and later with my changes, so I think it's fine to just use the fork until we can get it merged upstream.

SamSaffron:
@sgrif my huuuuge concern is memory, we have people on 1GB droplets already freaking out big time about swap size and some of them can no longer use the web UI to upgrade Discourse anymore cause sprokets 3 is so memory heavy, if 4 becomes more memory heavy we are in super huge trouble.

sgrif:
Yes, I will try to improve the situation. There might be an easy win by disabling source maps, but I suspect it will require more in-depth work. It's also worth noting that for "app level javascript" (as opposed to "sprinkles"), Rails is moving to using webpack instead of the asset pipeline in 5.1. Might be worth exploring that as an option as well. (via https://github.com/rails/webpacker)

schneems:
I think source maps are only generated in development. But I could be wrong.

SamSaffron:
Just a quick update here, @sgrif has been super busy with a baby Ruby :) (super duper congrats) 

Discussed this PR with him in Australia and he plans to look at it some next month.

Internally there is now huge level of urgency for cause we are probably going to move off sprockets later this year (sorry @schneems ) 

Going to close this for now for @sgrif to reopen when the perf issue on sprockets 4 are worked out.

PR #4570, opened 2016-11-25
FEATURE: reply as new message to the same recipients

LeoMcA:
Related meta post: https://meta.discourse.org/t/reply-as-new-message-to-same-recipients/47463/4?u=leomca

### DIFF
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -650,17 +650,14 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
       this.send('deselectText');
 
       var options;
-
       if (this.get('model.isPrivateMessage')) {
+        let users = this.get('model.details.allowed_users');
+        let groups = this.get('model.details.allowed_groups');
 
-        var users = this.get('model.details.allowed_users');
-        var groups = this.get('model.details.allowed_groups');
-
-        var usernames = users.map(user => {
-          return user.username;
-        }).concat(groups.map(group => {
-          return group.name;
-        })).join();
+        let usernames = [];
+        users.forEach(user => usernames.push(user.username));
+        groups.forEach(group => usernames.push(group.name));
+        usernames = usernames.join();
 
         options = {
           action: Composer.PRIVATE_MESSAGE,
@@ -668,15 +665,12 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
           draftKey: Composer.REPLY_AS_NEW_PRIVATE_MESSAGE_KEY,
           usernames: usernames
         };
-
       } else {
-
         options = {
           action: Composer.CREATE_TOPIC,
           draftKey: Composer.REPLY_AS_NEW_TOPIC_KEY,
           categoryId: this.get('model.category.id')
         };
-
       }
 
       composerController.open(options).then(() => {
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -649,11 +649,37 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
       const quotedText = Quote.build(post, quoteState.get('buffer'));
       this.send('deselectText');
 
-      composerController.open({
-        action: Composer.CREATE_TOPIC,
-        draftKey: Composer.REPLY_AS_NEW_TOPIC_KEY,
-        categoryId: this.get('model.category.id')
-      }).then(() => {
+      var options;
+
+      if (this.get('model.isPrivateMessage')) {
+
+        var users = this.get('model.details.allowed_users');
+        var groups = this.get('model.details.allowed_groups');
+
+        var usernames = users.map(user => {
+          return user.username;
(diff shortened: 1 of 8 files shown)


ZogStriP:
This looks ok, but it's sensitive. I'll merge it after the version bump.

ZogStriP:
Looks like it's not passing ESlint. Can you have a look @LeoMcA?

LeoMcA:
Oops, my bad, it should pass now.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -649,11 +649,37 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
-      composerController.open({
-        action: Composer.CREATE_TOPIC,
-        draftKey: Composer.REPLY_AS_NEW_TOPIC_KEY,
-        categoryId: this.get('model.category.id')
-      }).then(() => {
+      var options;
+
+      if (this.get('model.isPrivateMessage')) {
+
+        var users = this.get('model.details.allowed_users');
```
tgxworld: We prefer to use `let` instead of `var` since `let` will ensure that the variable is only available within the block scope.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -649,11 +649,37 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
+        var groups = this.get('model.details.allowed_groups');
+
+        var usernames = users.map(user => {
+          return user.username;
+        }).concat(groups.map(group => {
+          return group.name;
+        })).join();
```
tgxworld: Instead of joining at the end, I think we can just push the usernames into a string as we iterate each array and thus avoiding creating a new array.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -649,11 +649,37 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
 
-      composerController.open({
-        action: Composer.CREATE_TOPIC,
-        draftKey: Composer.REPLY_AS_NEW_TOPIC_KEY,
-        categoryId: this.get('model.category.id')
-      }).then(() => {
+      var options;
+
+      if (this.get('model.isPrivateMessage')) {
+
```
tgxworld: I'll prefer if we :scissors: these empty lines. A good rule of thumb is to follow the style of the surrounding code.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -649,11 +649,37 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
...
+      } else {
+
+        options = {
+          action: Composer.CREATE_TOPIC,
+          draftKey: Composer.REPLY_AS_NEW_TOPIC_KEY,
+          categoryId: this.get('model.category.id')
+        };
+
+      }
```
tgxworld: We'll need a controller test for this :wink:

app/assets/javascripts/discourse/controllers/topic.js.es6
LeoMcA: This is what I came up with:
```js
let usernames = '';
users.forEach(user => {
  usernames += user.username + ', ';
});
groups.forEach(group => {
  usernames += group.name + ', ';
});
usernames = usernames.slice(0, -2);
```

app/assets/javascripts/discourse/controllers/topic.js.es6
tgxworld: Opps yea you're right.

app/assets/javascripts/discourse/controllers/topic.js.es6
tgxworld: Opps sorry. Looks like my suggestion is more complicated because we now have to deal with commas. Perhaps something like

```
let usernames = []

users.forEach(user => usernames.push(user));
groups.forEach(group => usernames.push(group));

usernames = usernames.join();
```

At least we won't have two intermediate arrays.

LeoMcA:
@tgxworld I think I've done everything you wanted, aside from the acceptance test. I've written one for the existing reply as new topic functionality, however it doesn't look like there's any pretend pm I can navigate to, to click the reply as new message button on. Some advice on where to go from here (like how to add one) would be 👌

Is it just a matter of dumping the json from a private message in `test/javascripts/fixtures/topic.js.es6` and then creating an appropriate route in `test/javascripts/helpers/create-pretender.js.es6`? I expect it'd be nicer if you guys could provide that json so it has familiar usernames like "eviltrout" in it, rather than "leo_test2"...

tgxworld:
@LeoMcA  Sorry for dropping this :( 

I'll have a look at this real soon.

LeoMcA:
I'm waiting to hear back from the client I did this for about a new contract, so no immediate rush

tgxworld:
Merged in https://github.com/discourse/discourse/commit/c76f6856ea5b52bbda1e0e7b9b35bf0c46858026

@LeoMcA Thank you for your contribution! Overall the PR was a pleasure to review. :+1:

PR #4571, opened 2016-11-26
FEATURE: Hide Profile Text from non-staff if user is suspended

cpradio:
Relevant discussion
https://meta.discourse.org/t/hide-about-me-for-suspended-users/53383


ZogStriP:
Thanks 👍

PR #4573, opened 2016-11-27
FEATURE: Add notification and session webhooks

fantasticfears:


tgxworld:
Was this discussed on meta? 🤔 I'm not sure if there are alot of valid use cases for session and notification webhooks.

fantasticfears:
Some of them are in the topic in `howto`.. 

Here is one for session: https://meta.discourse.org/t/setting-up-webhooks/49045/10?u=fantasticfears
Here is one for badge: https://meta.discourse.org/t/webhooks-feedback/49046/26?u=fantasticfears
Here is one for notification: https://meta.discourse.org/t/setting-up-webhooks/49045/6?u=fantasticfears

Actually, I'd like to introduce user badge and notification webhooks. But it seems most user badges are granted by `backfill` in BadgeGranter. Thus the DiscourseEvent in `UserBadge` isn't that helpful as well as a user badge webhook. But notification webhooks does get pushed.

tgxworld:
I'm not sure about all these hooks. The need for them seems to be too specific to a few requests and their needs are probably not very common. I'll avoid adding these webhooks to core and if they really need them, they can write a plugin for it.

fantasticfears:
No problem, I can extracted them into a plugin later. https://meta.discourse.org/t/setting-up-webhooks/49045/13

tgxworld:
@fantasticfears Sorry about that. Thank you for the work on webhooks :)

PR #4575, opened 2016-11-28
new rake task uploads:create_record

techAPJ:


techAPJ:
@tgxworld can you review this PR please?

lib/tasks/uploads.rake
```diff
@@ -650,3 +650,86 @@ task "uploads:analyze", [:cache_path, :limit] => :environment do |_, args|
...
+    paths.each_with_index do |path, index|
+      filename = File.basename(path)
+
+      printf("%9d / %d (%5.1f%%)\n", (index + 1), max, (((index + 1).to_f / max.to_f) * 100).round(1))
+
+      Post.where("raw LIKE ?", "%#{filename}%").each do |post|
+        doc = Nokogiri::HTML::fragment(post.raw)
+        updated = false
+
+        doc.css("img[src]").each do |img|
```
ZogStriP: Only works for images? Not attachments?

lib/tasks/uploads.rake
```diff
@@ -650,3 +650,86 @@ task "uploads:analyze", [:cache_path, :limit] => :environment do |_, args|
...
+              File.open(existing_path) do |file|
+                new_upload = Upload.create_for(Discourse::SYSTEM_USER_ID, file, File.basename(url), File.size(file))
+
+                if new_upload.persisted?
+                  printf "Restored into #{new_upload.url}\n"
+                  DbHelper.remap(url, new_upload.url)
+                  updated = true
+
+                  # move existing image (without upload record) to tombstone
+                  move_to_tombstone(url)
```
ZogStriP: Why are you moving the file to the tombstone here?

lib/tasks/uploads.rake
```diff
@@ -650,3 +650,86 @@ task "uploads:analyze", [:cache_path, :limit] => :environment do |_, args|
...
+
+        post.rebake! if updated
+      end
+    end
+  ensure
+    SiteSetting.max_image_size_kb = original_setting
+  end
+end
+
+def move_to_tombstone(url)
```
ZogStriP: Did you *need* to extract these method from the `local_store`?

tgxworld:
This task repeats alot of the logic from [recover from tombstone rake task](https://github.com/discourse/discourse/blob/master/lib/tasks/uploads.rake#L392-L458), I feel like we can reuse alot of the code being written here.

techAPJ:
This rake task was specific for fixing a case where I copied the images to default uploads directory while the database record for copied uploads did not exist.

I will fix that site manually. The better approach here would be to copy the images to tombstone directory and then use `uploads:recover_from_tombstone` task.

PR #4577, opened 2016-11-30
FEATURE: Allow group owners to edit group name and avatar flair.

tgxworld:
screenshot from 2016-11-30 16-15-30

screenshot from 2016-11-30 16-09-47


tgxworld:
cc @eviltrout Can you review this? Thank you!

chenyingyin:
default

SamSaffron:
not sure about allowing them to edit group name, seems like it is risky to me.

tgxworld:
Hmm that is true. All existing mentions will have to be changed. Strange we don't have that warning in the admin panel.

PR #4584, opened 2016-12-03
imporoving vb4 importer

mabras:


script/import_scripts/vbulletin.rb
```diff
@@ -469,15 +473,15 @@ def close_topics
     sql = <<-SQL
       WITH closed_topic_ids AS (
         SELECT t.id AS topic_id
-        FROM #{TABLE_PREFIX}post_custom_fields pcf
```
techAPJ: Why are you removing `#{TABLE_PREFIX}` here (and in 3 more places below)?

script/import_scripts/vbulletin.rb
```diff
@@ -718,22 +722,20 @@ def suspend_users
...
-        user.suspended_at = parse_timestamp(user["bandate"])
-        user.suspended_till = 200.years.from_now
-
-        if user.save
-          StaffActionLogger.new(system_user).log_user_suspend(user, "banned during initial import")
-          banned += 1
-        else
-          puts "Failed to suspend user #{user.username}. #{user.errors.try(:full_messages).try(:inspect)}"
-          failed += 1
+    User.find_each do |user|
```
techAPJ: Looping through all the users does not seem like a better approach here.

I know that current `suspend_users` method is broken and I have a (single line) fix for this in place which I will commit soon. Can you please omit this change?

script/import_scripts/vbulletin.rb
mabras: This query is for postgresql database (Discourse database) not for mysql (VB4 database).

techAPJ: Aha, I see now! 👍

script/import_scripts/vbulletin.rb
mabras: Ok, I will omit this.

mabras:
I have omitted the changes you asked for.

I just want to ask you, is okay to replace all the `gsub` method with `gsub!` to save to memory?

techAPJ:
Sure, go ahead! :+1:

mabras:
I did replace the `gsub` method. 

In addition to that the `@closed_topic_ids` always return an empty array duo to that `topic["open"]` return `Fixnum` object and we `==` against `String` object.  This broke the `close_topics` method.

techAPJ:
Thanks!

PR #4585, opened 2016-12-04
Add initial Urdu support with RTL direction

ibnesayeed:
This is an initial commit to make sure that the Urdu locale is added in the locale settings in admin control panel. Actual translations are yet to be added. Currently all the YAML files are copied from English and the root key of each file is changed from `en` to `ur`. Additionally, since Urdu is an RTL language, a small change is made in the RTL class to reflect that.

### DIFF
--- a/app/assets/javascripts/locales/ur.js.erb
+++ b/app/assets/javascripts/locales/ur.js.erb
@@ -1,0 +1,3 @@
+//= depend_on 'client.ur.yml'
+//= require locales/i18n
+<%= JsLocaleHelper.output_locale(:ur) %>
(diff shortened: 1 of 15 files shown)


ibnesayeed:
The build was failing because of missing Urdu locale in moment.js. I have created [a pull request](https://github.com/moment/moment/pull/3641) there to add Urdu locale support. Once the PR is accepted in moment.js, required files will be available in their build.

If needed, I can commit the moment.js Urdu locale file in discourse repo manually. Should I?

SamSaffron:
Thank you!!!

PR #4587, opened 2016-12-05
FIX: simplify CSV file upload

techAPJ:


app/models/invite.rb
```diff
@@ -265,8 +265,12 @@ def self.base_directory
...
   end
 
-  def self.chunk_path(identifier, filename, chunk_number)
-    File.join(Invite.base_directory, "tmp", identifier, "#{filename}.part#{chunk_number}")
+  def self.create_csv(file, name)
+    extension = File.extname(file.original_filename)
+    path = "#{Invite.base_directory}/#{name}#{extension}"
+    FileUtils.mkdir_p(Pathname.new(path).dirname)
+    File.open(path, "wb") { |f| f << file.tempfile.read }
+    return path
```
ZogStriP: You don't need the `return` here ;)

techAPJ: Removed :)

techAPJ:
Thanks for the review @ZogStriP! ❤️

PR #4589, opened 2016-12-05
FIX: Don't raise error when admin access is invalid.

tgxworld:


SamSaffron:
not in a rescue from on application controller?

tgxworld:
@SamSaffron It is already in https://github.com/discourse/discourse/blob/master/app/controllers/application_controller.rb#L154-L156.

The error we're seeing  here is happening in `AdminConstraint`

PR #4590, opened 2016-12-05
Use CSS animations for post and topic highlights

xfalcox:


coding-horror:
if this is safe and doesn't break anything, we should merge it.

xfalcox:
I tested it locally, but would like a review from someone.

We win CSS animations (they should be faster and more customizable) and I'm adding one event listener. The trade-off looks good for me, but please correct me if I'm wrong.

SamSaffron:
Is animationend event safe on all browsers? cc @eviltrout

ZogStriP:
http://caniuse.com/#search=animationend doesn't mention `onanimationend` so I would say 👍

SamSaffron:
Let's test this on android first, I read somewhere that it was less evenly supported there.

SamSaffron:
also, we may need to prefix http://stackoverflow.com/questions/25629096/onanimationend-webkitanimationend-oanimationend-only-fires-once

xfalcox:
It was indeed broken on Mobile, but it was because of the CSS.

Anyway tested on Chrome Stable on Android and working as of e9fa936389a748d57bab15d2cb335c1bde67d6cf

PR #4591, opened 2016-12-06
FIX: Return 403 when in reaodnly mode.

tgxworld:
@mpalmer Does this look OK to you? thank you!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -156,7 +156,7 @@ class ApplicationController < ActionController::Base
   end
 
   rescue_from Discourse::ReadOnly do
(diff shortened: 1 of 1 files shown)


mpalmer:
I think a 5xx code (specifically, "503 Service Unavailable": "The server is currently unavailable (because it is overloaded or down for maintenance). Generally, this is a temporary state.") would be more appropriate.  My rule of thumb is that if the client could conceivably change something about the request to make it work, then a 4xx response is appropriate, otherwise a 5xx is the right thing to do.  In this, all the client can do is wait for the server to come out of read-only, so 503 is better.

tgxworld:
@mpalmer Thank you for the explanation and review :+1:

PR #4592, opened 2016-12-07
FEATURE: Public groups

tgxworld:
https://meta.discourse.org/t/improving-the-groups-page-for-1-7/53347?u=tgxworld

screenshot from 2016-12-07 15-07-01
screenshot from 2016-12-07 15-06-54


app/models/topic_link.rb
```diff
@@ -274,14 +274,14 @@ def self.duplicate_lookup(topic)
...
 #  crawled_at    :datetime
 #  quote         :boolean          default(FALSE), not null
 #
 # Indexes
 #
-#  index_topic_links_on_link_post_id_and_reflection  (link_post_id,reflection)
-#  index_topic_links_on_post_id                      (post_id)
-#  index_topic_links_on_topic_id                     (topic_id)
-#  unique_post_links                                 (topic_id,post_id,url) UNIQUE
+#  index_forum_thread_links_on_forum_thread_id                      (topic_id)
```
SamSaffron: this annotation is not right, I think you should be annotating off the test database, not dev

db/migrate/20161208064834_create_group_histories.rb
```diff
@@ -0,0 +1,15 @@
+class CreateGroupHistories < ActiveRecord::Migration
+  def change
+    create_table :group_histories do |t|
+      t.integer :group_id, index: true, null: false
+      t.integer :acting_user_id, index: true, null: false
```
SamSaffron: ouch, `index: true` though sparse worries me, ideally I think we should create indexes in an extra step and name them while creating them.

app/models/topic_link.rb
tgxworld: Ah good catch. I had a database restored while running the annotation. Thank you!

PR #4593, opened 2016-12-07
FEATURE: show the reply title and the reply icon in the minimized editor

techAPJ:
@eviltrout can you review this PR please? Thanks!

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -6,6 +6,8 @@ import { default as computed, observes } from 'ember-addons/ember-computed-decor
 import { relativeAge } from 'discourse/lib/formatter';
 import InputValidation from 'discourse/models/input-validation';
 import { getOwner } from 'discourse-common/lib/get-owner';
+import { escapeExpression } from 'discourse/lib/utilities';
+import { emojiUnescape } from 'discourse/lib/text';
 
 function loadDraft(store, opts) {
   opts = opts || {};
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -195,6 +195,11 @@ const Composer = RestModel.extend({
     return canEditTopicFeaturedLink ? !this.siteSettings.topic_featured_link_onebox : true;
   },
 
+  @computed("topic")
+  draftTitle(topic) {
```
eviltrout: I don't think you should add formatting to the model. The model should be where the data lives, not its presentation.

Couldn't you do all this in the template?

```handlebars
{{#if model.topic}}
  {{fa-icon "reply"}} {{model.topic.title}}
{{else}}
  {{i18n "composer.saved_draft"}}
{{/if}}
```

eviltrout: That is also wrong though :)

If you have to emoji the title, put the property in the controller instead of the model. You can leave the icon in the template.

app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -195,6 +195,11 @@ const Composer = RestModel.extend({
...
+    return topic ? " <i class='fa fa-reply'></i> " + emojiUnescape(escapeExpression(topic.get('title'))) : I18n.t('composer.saved_draft');
```
tgxworld: Aww no :heart:  for [Template Literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals)? :stuck_out_tongue:

app/assets/javascripts/discourse/models/composer.js.es6
techAPJ: @eviltrout Okay, done. Thanks for review!

app/assets/javascripts/discourse/models/composer.js.es6
techAPJ: No need for it anymore :)

PR #4597, opened 2016-12-11
RTL direction assigned to Urdu language

ibnesayeed:
An array of RTL languages is present in two places, one in the RTL model class and the other in the application helper class. Previously, Urdu was only added in the RTL model class which was not taking affect on the interface. This PR should fix it.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -135,7 +135,7 @@ module ApplicationHelper
   end
 
   def rtl?
-    ["ar", "fa_IR", "he"].include? I18n.locale.to_s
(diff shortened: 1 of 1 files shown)


ibnesayeed:
@SamSaffron this minor PR is ready for the review and to be merged.

ibnesayeed:
@SamSaffron this PR is a complement to #4585 and is necessary to change the direction of the site. I would highly urge to get it merged unless there is something blocking, in which case I would be willing to make necessary changes.

To understand why it was missed in the earlier PR, you can read https://meta.discourse.org/t/why-is-rtl-class-user-dependent-instead-of-locale/54077 and https://meta.discourse.org/t/urdu-language-support-added/53877/3?u=ibnesayeed

PR #4598, opened 2016-12-13
mailcatcher dependency added for dev environment

ibnesayeed:
In reference to the https://github.com/discourse/discourse_docker/pull/317 the gem is added in the Gemfile and inline install command cleaned up from the mailcatcher binstub.


SamSaffron:
nahh I prefer to keep it in the dev image specifically, will amend shortly

PR #4600, opened 2016-12-13
FEATURE: Split Reply-button into 'Reply Post' and 'Reply Topic'

jsilvanus:
This feature should enable giving a different label for '_Reply Topic_' and '_Reply Post_'. I hope I didn't forget anything.

I also considered moving the button label to `controls.reply`, but that was already taken for the tooltip for said button, whereas label for the button is taken from `topic.reply.title`, so I just elaborated on the current state.


jsilvanus:
I must admit that I did not have a dev discourse running, so I didn't do a test on this... Will try to do it later!

config/locales/client.en.yml
```diff
@@ -1542,8 +1542,12 @@ en:
         remove_banner: "Remove Banner Topic"
 
       reply:
-        title: 'Reply'
-        help: 'begin composing a reply to this topic'
+        message:
+          title: 'Reply'
+          # help i.e. tooltop is at controls.reply
```
jsilvanus: I added a small comment here to clarify, where the actual tooltip is found.

config/locales/client.en.yml
```diff
@@ -1542,8 +1542,12 @@ en:
...
+        thread:
```
jsilvanus: Hmm. Now that I think of it, I think that "topic" instead of "thread" would follow the naming convention...

jsilvanus:
@ZogStriP, changed this to follow naming convention: now post & topic instead of message & thread. :)

jsilvanus:
Few thoughts:
1) I considered long and hard, whether the tooltip should be moved from `controls.reply` to `post.reply.post.help`, but decided not to go for this.

2) Should I go through all the locales and create fixes for them?

ZogStriP:
I will merge **after** the next version bump, that way, translators will have time to fix the strings.

ZogStriP:
Sorry for the delay @jsilvanus. Thanks for the PR 👍

coding-horror:
This PR caused a lot of damage -- broken translations in every language, including English. We are in the process of reverting it. 

Do not submit this again; it will not be accepted.

jsilvanus:
I thought I made clear I had another PR that has the changes for ymls... I was told it wasn't needed as the ymls were handled differently. I'm sorry for the trouble.

coding-horror:
We are still dealing with the fallout from this disastrous change:

https://meta.discourse.org/t/turkish-translation-needs-some-love/57101

jsilvanus:
@coding-horror, can I ask what was wrong with the PR? Was the initial idea of splitting the two buttons (which have different functionality but have the same string attached to them) wrong, or was my code somehow disastrous? I had all the translations in another branch, but I was told not to include them.

ZogStriP:
It was because this changed the translation of a very important feature and our client-side translation code wasn't properly falling back to English when the translation was unavailable. This basically breaking non-English Discourses. 

As for other translations, they would have been overwritten on our next pull from transifex anyway. So they would not have been useful.

jsilvanus:
Thank you, this clarifies a bit more what happened. I was in a week long training - and thus unable to participate in the discussion - while I suddenly got a notification on my mobile that my PR had caused enormous trouble and I was all "OMG WHAT DID I DO"...

PR #4601, opened 2016-12-13
Removed the fallback mailcatcher install command

ibnesayeed:
Since `apt-get install -y libsqlite3-dev && gem install mailcatcher` is now part of the Docker dev image, this fallback installation command is not needed.


tgxworld:
Thanks! @ibnesayeed

PR #4602, opened 2016-12-14
FIX: Only quote `.cooked` text.

tgxworld:
@ZogStriP Can I check if there are any scenarios where we might be quoting text outside `.cooked`? I'm worried that I might be introducing a regression here. Thank you!

### DIFF
--- a/app/assets/javascripts/discourse/components/quote-button.js.es6
+++ b/app/assets/javascripts/discourse/components/quote-button.js.es6
@@ -28,6 +28,9 @@ export default Ember.Component.extend({
     let firstRange, postId;
     for (let r = 0; r < selection.rangeCount; r++) {
       const range = selection.getRangeAt(r);
+
+      if ($(range.endContainer).closest('.cooked').length === 0) return;
+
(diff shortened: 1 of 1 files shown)


ZogStriP:
`.contents` is there to ensure we show the quote button even when you select the last paragraph (which somehow select an element which isn't a child of `.cooked`)

tgxworld:
Hmm if their selecting an element which isn't a child of `.cooked`, we shouldn't be showing the quote reply button right?

ZogStriP:
I agree, but due to this quirk we have to fix it another way

tgxworld:
Sorry for the late reply. What quirk are we referring to?

ZogStriP:
The one in my previous post ;)

> `.contents` is there to ensure we show the quote button even when you select the last paragraph

tgxworld:
@ZogStriP Can you review again? :smile:  This should fix the being able to quote the `likes` button bug

tgxworld:
Merging this in first.

PR #4604, opened 2016-12-15
Fix: close topics for vb4 importer

mabras:
In the current implementation, the `close_topics` method is somehow broken. Importing topics keeps track of the ids of closed topics in instance variable, but this make the importer not resumable. If the importer stop (memory issues...) the ids are gone.

The current implementation is good for small migration, but for big one is not ideal, and it can produces wrong results (not closing topics that needs to be close).


eviltrout:
Any chance you could sign the CLA? Otherwise we will have to reject this.

mabras:
Done

PR #4605, opened 2016-12-17
Return user_fields and custom_fields on directory_items api call.

kstaikov:
The goal here was to address the request to have custom fields in the user directory: https://meta.discourse.org/t/can-we-display-custom-fields-in-user-directory-list/35506.  

Added user_fields and custom_fields to the user_name_serializer.
This results in the directory_items api call returning the fields,
which allows us to easily include custom fields in the user directory
by lightly customizing some html if we anyone so desires.

### DIFF
--- a/app/serializers/user_name_serializer.rb
+++ b/app/serializers/user_name_serializer.rb
@@ -1,3 +1,3 @@
 class UserNameSerializer < BasicUserSerializer
-  attributes :name, :title
+  attributes :name, :title, :custom_fields, :user_fields
 end


kstaikov:
I could also just make a new serializer and inherits from this one and is called something like DirectoryUserInfoSerializer, whichever is preferred

tgxworld:
Hi thank you for your contribution. However, it doesn't look like we received a valid use case for adding the following fields. Sorry!

kstaikov:
So then if I or anyone else wants to include custom fields in the user directory, how else are we supposed to do it without having a front end plugin that calls the api for each user - which would result in as many extra api calls as there are users?

tgxworld:
Plugins can always add a new serializer and a new API endpoint

PR #4606, opened 2016-12-17
Fix Arabic Pluralization Rules for Client

mabras:
According to [Language Plural Rules](http://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html#ar) this how `many` should be calculated.


SamSaffron:
I trust that you know what you are doing 😄  , thanks!

PR #4609, opened 2016-12-18
FEATURE: Category setting to make all topics wikis

joebuhlig:


joebuhlig:
Related meta post: https://meta.discourse.org/t/category-setting-new-topics-are-wiki-by-default/16618/23?u=joebuhlig

SamSaffron:
I am ok with most of this except for the job that retroactively changes all topics to wiki in a category. 

Instead, the feature should simply change the default going forward and not touch history. Then you can make a #howto on meta explaining how to retroactively apply wiki to all old topics. As it stands it is super risky cause it can make changes that are very hard to revert.

jomaxro:
@joebuhlig Sam's request is a-OK with me - the site I'm looking to use this on already had a rake task run to make everything wiki, retroactive change isn't necessary in this feature (at least in my case).  I do tend to agree that making a change to every existing topic in a category is dangerous.

joebuhlig:
@SamSaffron and @jomaxro, I pulled the job and all the corresponding pieces. I'll put together a #howto for it on meta. It'll likely be a rails console process.

SamSaffron:
looks good to me 👍

jomaxro:
Awesome, thanks @SamSaffron.  It looks like this was merged into master.  What's the procedure to get it on tests-passed?

PR #4613, opened 2016-12-19
Tweak avatar UI on topic list

chrisbeach:
Of the avatars shown, make any previous posters semi-transparent, so only the latest poster is shown prominently. Previously, if the first avatar was the latest poster to the topic, a blue border was used as a highlight around the avatar. This is no longer necessary. Overall the topic list page looks less "busy" after this change.

discourse-topic-list-avatars-tweak


coding-horror:
Can't say I am a fan.. it suppresses everyone except the last poster.. a lot. This is better as a local CSS change.

PR #4614, opened 2016-12-19
[WIP] FEATURE: Deferred migrations that run 30 mins after core migrations.

tgxworld:


lib/backup_restore/restorer.rb
```diff
@@ -339,7 +339,6 @@ def switch_schema!
     def migrate_database
       log "Migrating the database..."
       Discourse::Application.load_tasks
-      ENV["VERSION"] = @current_version.to_s
```
tgxworld: Not sure why we needed this since `current_version` is always point to the latest migration.

lib/tasks/db.rake
```diff
@@ -5,6 +5,14 @@ end
 
 # we need to run seed_fu every time we run rake db:migrate
 task 'db:migrate' => ['environment', 'set_locale'] do
+  if Rails.env.production?
+    # Run deferred migrations 30 mins later
+    Jobs.enqueue_in(1800, :run_deferred_migrations)
```
tgxworld: I need to write code to figure out which deferred migration has already been run and run those migration.

tgxworld:
@SamSaffron  Can I get your thoughts on this?

PR #4617, opened 2016-12-19
Fix: Forced trailing slash on messageBus.baseUrl

claygriffiths:
If the `long polling base url` setting doesn't have a trailing slash then the message bus will make requests to the wrong path.

This simply ensures there's a trailing slash on the Message Bus base URL.


claygriffiths:
https://meta.discourse.org/t/message-bus-broken-with-discourse-relative-url-root/54191

SamSaffron:
I am concerned this is not the right spot for this change, should it not simply be a restriction on the site setting?

claygriffiths:
I thought about that too.

However, this makes it work even if they change the database using a hook in the YAML file and don't include the trailing slash there.

SamSaffron:
ok this is small enough I will merge it in, thanks

PR #4618, opened 2016-12-21
FIX: Don't allow invalid email to be saved.

tgxworld:
@eviltrout Can you review this? Thank you!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,6 @@
 require_dependency 'email'
 require_dependency 'email_token'
+require_dependency 'email_validator'
 require_dependency 'trust_level'
 require_dependency 'pbkdf2'
 require_dependency 'discourse'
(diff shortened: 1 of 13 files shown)


tgxworld:
ping @eviltrout

eviltrout:
Looks really good to me. Clear code and lots of tests. Merge when ready.

tgxworld:
@eviltrout Thanks for reviewing :+1:

app/models/user.rb
```diff
@@ -76,6 +77,7 @@ class User < ActiveRecord::Base
   validates_presence_of :username
   validate :username_validator, if: :username_changed?
   validates :email, presence: true, uniqueness: true
+  validates :email, format: { with: EmailValidator.email_regex }, if: :email_changed?
   validates :email, email: true, if: :should_validate_email?
```
gschlager: Is the validator (`email: true`) still needed?
And the new validator ignores the `skip_email_validation` flag. Is this intentional or should it be `if: :should_validate_email?` instead of `if: :email_changed?`?

PR #4620, opened 2016-12-22
Resolve raw templates in 'javascripts/mobile' and 'javascripts' on mobile

angusmcleod:
Further to [this discussion](https://meta.discourse.org/t/upgrading-ember-to-2-10/52724/69?u=angus), the same changes need to be made for mobile raw templates.

I've added both ``javascripts/mobile`` and ``javascripts`` as options as some raw mobile templates reference raw non-mobile templates, [see e.g.](https://github.com/angusmcleod/discourse-topic-previews/blob/master/assets/javascripts/discourse/templates/mobile/list/topic-list-item.raw.hbs)


eviltrout:
Ah yes, good call!

PR #4621, opened 2016-12-23
Toolbar action replaceText

kgish:
Should be possible to replace the current contents with some other text.


SamSaffron:
this looks good to me @eviltrout ?

PR #4622, opened 2016-12-23
FEATURE: Append tags bulk action for topics

dmacjam:
Related meta forum topic: https://meta.discourse.org/t/bulk-actions-add-tags-in-addition-to-change-tags/40059

I added new button for appending tags to topics in bulk action and renamed old button to replace tags as suggested in https://meta.discourse.org/t/multi-select-change-tags-breaks-stuff/43727/6

### DIFF
--- a/lib/discourse_tagging.rb
+++ b/lib/discourse_tagging.rb
@@ -4,7 +4,7 @@ module DiscourseTagging
   TAGS_FILTER_REGEXP = /[\/\?#\[\]@!\$&'\(\)\*\+,;=\.%\\`^\s|\{\}"<>]+/ # /?#[]@!$&'()*+,;=.%\`^|{}"<>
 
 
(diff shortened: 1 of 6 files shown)


jomaxro:
@SamSaffron any progress with this?  Would love to use this on my site!

lib/discourse_tagging.rb
```diff
@@ -4,7 +4,7 @@ module DiscourseTagging
   TAGS_FILTER_REGEXP = /[\/\?#\[\]@!\$&'\(\)\*\+,;=\.%\\`^\s|\{\}"<>]+/ # /?#[]@!$&'()*+,;=.%\`^|{}"<>
 
 
-  def self.tag_topic_by_names(topic, guardian, tag_names_arg)
+  def self.tag_topic_by_names(topic, guardian, tag_names_arg, append=false)
```
dmacjam: Yes, you identified it right. Thanks for the feedback.

PR #4623, opened 2016-12-24
Don't try and update rails-observers

maclover7:
Was removed from Discourse via c531f4ded5604fe0335284f12149322c16931cb7.


SamSaffron:
aha good point! thanks!

PR #4628, opened 2016-12-26
FIX: Move Welcome topic to I18n

caugner:
See: https://meta.discourse.org/t/translating-welcome-topics/54792?u=claas


ZogStriP:
Thanks 👍

PR #4629, opened 2016-12-27
FIX: Translate all boolean values in user index

caugner:
See: https://meta.discourse.org/t/user-administration-some-boolean-values-not-translated/54823?u=claas

### DIFF
--- a/app/assets/javascripts/admin/templates/user-index.hbs
+++ b/app/assets/javascripts/admin/templates/user-index.hbs
@@ -175,13 +175,7 @@
 
   <div class='display-row'>
     <div class='field'>{{i18n 'admin.users.active'}}</div>
-    <div class='value'>
-      {{#if model.active}}
-        {{i18n 'yes_value'}}
-      {{else}}
-        {{i18n 'no_value'}}
-      {{/if}}
-    </div>
+    <div class='value'>{{i18n-yes-no model.active}}</div>
     <div class='controls'>
       {{#if model.active}}
         {{#if model.can_deactivate}}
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/admin/templates/user-index.hbs
```diff
@@ -221,7 +221,13 @@
 
   <div class='display-row'>
     <div class='field'>{{i18n 'admin.user.admin'}}</div>
-    <div class='value'>{{model.admin}}</div>
+    <div class='value'>
+      {{#if model.admin}}
```
eviltrout: You know what would be better here to avoid repeating yourself? An ember helper that would display the boolean? For example, something like: `{{show-boolean model.admin}}`

Would be a lot more DRY.

ZogStriP:
I think it would be better to make a helper so that we can do something like `{{yes-no model.isSuspended}}`

SamSaffron:
it is extra weird that "Active" field gets a "Yes" / "No" and the rest get "true"/"false" 

Plus the UI is kind of weird

Blocked? false
Staged? false

and so on. 

Blocked? No
Staged? No 

Is a bit better, but I feel there is something better we can do. 

For example:

```
Staged? false  A staged user can only post via email in specific topics.
```

What is  the value of displaying this line if user is not staged? 

Instead I support using yes-no

app/assets/javascripts/admin/templates/user-index.hbs
caugner: I have extracted a helper `i18n-yes-no` in https://github.com/discourse/discourse/pull/4629/commits/e17bd6174c3497ccb349d1a697fb013156f10213. The prefix `i18n` announces that translation is taking place and `yes-no` (as suggested by @ZogStriP) is less technical than `boolean`.

caugner:
> Is a bit better, but I feel there is something better we can do.

In general, I could imagine showing these user statuses in a tag-style fashion near the username. Adding icons and/or color might also aid usability.

> What is the value of displaying this line if user is not staged?

Should I hide the *Staged?* line (if `false`) in the scope of this PR then? At least the *Blocked?* line right now provides a button to block a user.

PR #4632, opened 2017-01-02
FEATURE: Opt-in native Discourse app install banner

xfalcox:


app/controllers/metadata_controller.rb
```diff
@@ -29,5 +29,17 @@ def default_manifest
         }
       ]
     }
+    if SiteSetting.native_app_install_banner
```
tgxworld: I prefer to leave a blank line before/after a multi-line code block to make it easier to read.

tgxworld:
Looks good to me :+1:

PR #4635, opened 2017-01-04
FIX: Add emoji-custom class in custom emojis list.

caugner:
Context: https://meta.discourse.org/t/configure-custom-emoji/23365/10?u=claas


caugner:
Note: Check failed, because master is unstable.

ZogStriP:
Thanks 👍

PR #4636, opened 2017-01-05
FIX: Translate titles of email templates

caugner:
See: https://meta.discourse.org/t/email-template-titles-not-translated/55191?u=claas (part 1)


ZogStriP:
Thanks 😘

PR #4640, opened 2017-01-06
Allow for a custom hub server

krainboltgreene:
Seems weird to only allow a singular source of truth for updates, so I've just mimed development/test.


eviltrout:

elityty:
За мен също,благодаря!

PR #4641, opened 2017-01-07
FIX: Let skip_validations bypass guardian at post creation

martineriksson:
https://meta.discourse.org/t/guardian-bypassed-through-skip-validations-in-topiccreator-but-not-in-postcreator/55341

### DIFF
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -108,7 +108,7 @@ class PostCreator
       return false unless skip_validations? || validate_child(topic_creator)
     else
       @topic = Topic.find_by(id: @opts[:topic_id])
-      if (@topic.blank? || !guardian.can_create?(Post, @topic))
+      if (@topic.blank? || (!guardian.can_create?(Post, @topic) && !skip_validations?))
         errors[:base] << I18n.t(:topic_not_found)
         return false
       end


tgxworld:
Changes look OK to me but can you add a test case for it? Thank you for taking the time to submit a PR :+1:

martineriksson:
@tgxworld Thanks for getting back to me about this! Unfortunately, I am an amateur with almost no experience of testing, so my ability to help with this is very limited. Even after looking carefully at the test which failed I can not figure things out...

If I understand correctly, the failing test has to do with creating a moderator post from a user which has been blocked. To begin with, I do not understand the relevance of this test – if a user is blocked, how can it be the author of a moderator post in the first place? Perhaps I am missing something but should not the appropriate test be something like letting a blocked user do the actual moderator action which in turn generates the moderator post?

In any case, the problem seems to be that the method add_moderator_post (app/models/topic.rb, line 548) includes the skip_validations option when creating the post (line 558). So as I understand it, the test is strange to begin with since basically anything should be allowed when skipping validations. In other words, where the test says that the count should not be incremented, we should actually expect it to be incremented due to validations being skipped when creating moderator posts. (Except of course for the inconsistency which my pull request fixes.)

Given that add_moderator_post should use skip_validations (I do not know why it is needed but I suppose that there is some reason), we can rewrite the test to expect the post count to increase, i.e. expect the moderator post to be created despite the user being blocked. This would make the test meaningless as far as I can tell, i.e. the section under "when moderator post fails to be created". Alternatively, we could do something else to make it fail despite skipping validations, like passing an invalid user object but I suppose that would not be very meaningful.

So the best way I can think of to solve the problem with the failing test is to simply remove that part of the test, since it does not make much sense to me in the first place. But I am probably missing something here. Git blame tells me that you – @tgxworld – added this test so perhaps you can explain the reasoning behind it?

If we were to write a new test for this specifically, I don't know what that might be. The reason I made the change was importing posts to private categories, so perhaps we could do something like this:

1. A user creates a new topic in a particular category. This is expected to work.

2. Make the category private (restricted to some group which the user is not a member of).

3. User creates a new topic in the same category. This is expected to fail.

4. User creates a new topic in the category, but this time while including the skip_validation option. With the current state of the code, this should fail. After my update it should succeed.

We could perhaps call it "when creating post in restricted category without group permission" or some such, but I am now way out of my depth due to having very limited experience with both Ruby testing and Discourse development. That being said I am happy to help if I can and interested in learning more – and I also have time to spend on this. So if you give me some explanations and pointers I might be able to fix this.

tgxworld:
Thanks for digging into this. I believe the order of your condition is incorrect. 

You need to check whether validation has been skipped first before checking if the guardian can create a Post.

`!skip_validations? || !guardian.can_create?(Post, @topic))`

tgxworld:
Hi @martineriksson Any updates here?

deivid-rodriguez:
@martineriksson Hi! I proposed a fix for a very similar issue in #4691. I guess we might want to close this PR, since my PR is more complete.

> If I understand correctly, the failing test has to do with creating a moderator post from a user which has been blocked. To begin with, I do not understand the relevance of this test – if a user is blocked, how can it be the author of a moderator post in the first place? Perhaps I am missing something but should not the appropriate test be something like letting a blocked user do the actual moderator action which in turn generates the moderator post?

I agree that the test itself does not make a lot of sense. Git annotate information points to 9a0797204ab1ff84fe5a81d78209cfcf8a669010. That commits seems to be fixing another form of this same issue: when a user is blocked, `PostCreator#create` does not return an actual `Post`, so the code will crash. The commit fixes it by checking for presence on the object before using it, but that's not a good fix in my opinion. Since the (`nil`) `new_post` object is actually returned by the method, any further code that don't guard for presence on that object will lead to a crash (like it happened to us). In any case, it explains the reason for the test. It might not make a lot of sense from a high level perspective, it makes sense as a unit test to reproduce the issue.

> In any case, the problem seems to be that the method add_moderator_post (app/models/topic.rb, line 548) includes the skip_validations option when creating the post (line 558). So as I understand it, the test is strange to begin with since basically anything should be allowed when skipping validations. In other words, where the test says that the count should not be incremented, we should actually expect it to be incremented due to validations being skipped when creating moderator posts. (Except of course for the inconsistency which my pull request fixes.)

Yes. In my opinion, the better fix for the issue is to actually ensure that the method always returns a post when `skip_validations` is passed. If you see [my commit](https://github.com/discourse/discourse/pull/4691/commits/6f34665df517697b6656fc77a2ae41e1095da0f6), which fixes it in the same way as yours, I actually changed the expectations to something that better fits what you expect here.

> Given that add_moderator_post should use skip_validations (I do not know why it is needed but I suppose that there is some reason), we can rewrite the test to expect the post count to increase, i.e. expect the moderator post to be created despite the user being blocked.

One reason reason to use `skip_validations` for moderator posts is actually explained by the message of another commit in my PR: https://github.com/discourse/discourse/pull/4691/commits/ce27f8e39e2e4ebec8d6710fcd2fdce7ca1ac770. I guess that applies to other cases, automatic moderator comments have a very specific shape and can be considered valid.
 
> If we were to write a new test for this specifically, I don't know what that might be. The reason I made the change was importing posts to private categories, so perhaps we could do something like this:
>  1. A user creates a new topic in a particular category. This is expected to work.
>  2. Make the category private (restricted to some group which the user is not a member of).
>  3. User creates a new topic in the same category. This is expected to fail.
>  4. User creates a new topic in the category, but this time while including the skip_validation option.
>
> With the current state of the code, this should fail. After my update it should succeed.

I could try adding a regression test to my PR  for this case, although I think [the regression test](https://github.com/deivid-rodriguez/discourse/blob/6f34665df517697b6656fc77a2ae41e1095da0f6/spec/components/post_creator_spec.rb#L742-L745) I already added might be enough?

@tgxworld 

> You need to check whether validation has been skipped first before checking if the guardian can create a Post.
> ```
> (!skip_validations? || !guardian.can_create?(Post, @topic))
> ```

I can add this to my PR. This is just for performance, right?

martineriksson:
@deivid-rodriguez Thank you for your comments and for the PR which obviously makes the one I made redundant, since it includes the one and only change I was proposing and also updates tests etc. I will close this PR now and cross my fingers that yours get accepted!

deivid-rodriguez:
@martineriksson Thanks! Let's see what the maintainers say, but since @tgxworld was positive about your changes, I'm optimistic :) Are you able to confirm that my patch fixes the issue you were having as well?

martineriksson:
@deivid-rodriguez Since the change to line 111 in post_creator.rb is almost identical and has the same effect, I am confident that your patch solves my problem.

deivid-rodriguez:
@martineriksson After rebasing the linked PR on top of current master, I found out that the issue is no longer reproducible. My PR changed, although still contains the same change to fix a different issue (https://github.com/discourse/discourse/pull/4691/commits/3c01426b82bf0a80b0f84b717352a92272aaa03e).

Are you able to check whether your problem is still reproducible and (if yes) whether it's still fixed by my PR?

PR #4643, opened 2017-01-09
FIX: Handle img src starting with "//" in pull_hotlinked_images job

LeoMcA:


ZogStriP:
No tests? :'(

LeoMcA:
I'm not sure I know where to even begin writing a test like this (which seems to rely on so many external things, like externally hosted images)... but I'll give it a go!

LeoMcA:
@ZogStriP how's that? 😸

LeoMcA:
Is there anything on my end blocking this from being merged?

ZogStriP:
I was waiting for the build to pass and then completely forgot... Sorry.

Thanks for the work 👍

PR #4647, opened 2017-01-11
Bbpress missing display name

pfaffman:
Apparently, sometimes display_names are missing. This patch uses user_nicename if there is no display_name.


script/import_scripts/bbpress.rb
```diff
@@ -84,7 +84,7 @@ def import_users
           id: u["id"].to_i,
           username: u["user_nicename"],
           email: u["user_email"].downcase,
-          name: u["display_name"],
+          name: u["display_name"].length > 0 ? u["display_name"] : u['user_nicename'],
```
ZogStriP: rails has a `.presence` helper that return the string only if it has some content

```rails
name: u["display_name"].presence || u["user_nicename"]
```

pfaffman: Fancy. 

OK. I did that and committed again. Is that what I was supposed to do?

PR #4648, opened 2017-01-13
FEATURE: Better error message when incoming e-mail is missing a Date: header

mpalmer:
As per https://meta.discourse.org/t/straightforward-direct-delivery-incoming-mail/49487/38

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -573,6 +573,10 @@ module Email
 
       # ensure posts aren't created in the future
       options[:created_at] ||= @mail.date
+      if options[:created_at].nil?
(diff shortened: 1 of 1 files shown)


lib/email/receiver.rb
```diff
@@ -573,6 +573,10 @@ def create_post(options={})
 
       # ensure posts aren't created in the future
       options[:created_at] ||= @mail.date
+      if options[:created_at].nil?
```
ZogStriP: I think it's fine to just use today's date instead.

```ruby
options[:created_at] ||= @mail.date || DateTime.now
```

PR #4653, opened 2017-01-14
FIX: RSS Language should use a dash/hypen instead of underscore

cpradio:
https://meta.discourse.org/t/helping-with-rss-issue/55767


ZogStriP:
Thanks 👍

PR #4654, opened 2017-01-16
FIX: Do not update user stats like counts for private messages.

tgxworld:


tgxworld:
Thanks @ZogStriP  :+1:

PR #4655, opened 2017-01-16
escape censored words

rimian:
This prevents regular expression errors for some special characters in censored words. 

For example "Co(onut" fails

Some of the character substitution is a little verbose but it works. Escaping and replacing with single backslashes is tricky :(

### DIFF
--- a/lib/validators/censored_words_validator.rb
+++ b/lib/validators/censored_words_validator.rb
@@ -1,6 +1,6 @@
 class CensoredWordsValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    if !SiteSetting.censored_words.blank? && value =~ /#{escape_censored_words}/i
+    if !SiteSetting.censored_words.blank? && value.match(escape_censored_words)
       record.errors.add(
         attribute, :contains_censored_words,
         censored_words: SiteSetting.censored_words
@@ -16,12 +16,6 @@ class CensoredWordsValidator < ActiveModel::EachValidator
   private
 
   def escape_censored_words
-    # TODO escape with single slashes
-    SiteSetting.censored_words
-                              .gsub('*', '\*')
-                              .gsub(')', '\)')
-                              .gsub('(', '\(')
-                              .gsub('[', '\[')
-                              .gsub(']', '\]')
+    Regexp.new(SiteSetting.censored_words.split('|').map { |w| Regexp.escape(w) }.join('|'), true)
   end
 end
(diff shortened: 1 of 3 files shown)


coding-horror:
What is the purpose of this change? What problem does it solve? I am unclear.

rimian:
Censored words are validated as a regular expression without being escaped. Since the latest 1.8.0 beta release, we're experiencing 500 errors when creating new posts.

coding-horror:
Your censored words have regex-valid characters in them? May I ask why?

rimian:
That's a good question! My site moderators don't understand regular expressions . I think there's a good use case for escaping them or evaluating them as plain old strings. I also think my PR is only part of the solution.

coding-horror:
Sorry I still don't understand. Why do censored words need characters like / or $ in them?

rimian:
I guess they don't. But the site breaks if someone enters one in. :)

lib/validators/censored_words_validator.rb
```diff
@@ -12,4 +12,16 @@ def validate_each(record, attribute, value)
       )
     end
   end
+
+  private
+
+  def escape_censored_words
```
rimian: Unfortunately not. It adds two backslashes. :(

tgxworld:
@rimian Perhaps you can add a validation to only allow words, space and numbers?

rimian:
That's probably better. Do you mean validate the words that can be added to the censor list?

rimian:
This validates instead #4659

lib/validators/censored_words_validator.rb
tgxworld: Is that not valid?

rimian: Actually, I think you're correct and my specs aren't right. Let me see and get back to you. I have another PR open than enforces alpha numeric characters. Obviously one of them is going to have to be closed at some point. :)

rimian: @tgxworld I think I fixed it. I'm going to check it on my staging server...

rimian:
The title works but the body does not. I've obviously missed something.

screen shot 2017-01-18 at 2 23 48 pm

tgxworld:
This looks good to me :) 

There isn't a validation for censored words in the body. It is automatically hidden via markdown :)

rimian:
@tgxworld it's breaking in markdown! If there is an escaped word: it breaks the others

screen shot 2017-01-18 at 3 05 54 pm
screen shot 2017-01-18 at 3 05 42 pm

rimian:
co(onut pineapple

never thought I would post that in a PR :)

tgxworld:
Merged in https://github.com/discourse/discourse/commit/25516874b59e5a605425371684ab5ee72d9547c4. Thank you @rimian

tgxworld:
There was also a client side bug where we were not escaping valid Regexp characters.

PR #4657, opened 2017-01-16
Add reopenWidget method

gdpelican:
This is a little bit of mad science that I needed for Babble; I figured I'd offer it up; you may or may not want to support it in core.

This allows us to add or override existing methods / properties on widgets, similar to Ember's `reopenClass` method. I found myself needing it to overwrite the `url()` method in `notification-item`, because I want to point chat notifications to the full page chat. I call it like this:

```
import { reopenWidget } from 'discourse/widgets/widget'

reopenWidget('notification-item', {
  url() {
    console.log("I am a custom function!")
  }
})
```

Supporting `this._super()` as well would be ideal here, but Ember's got a whole crazy scheme for supporting that, which I don't really want to reimplement (I was able to get the super method in other ways)

### DIFF
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -5,7 +5,7 @@ import { addButton } from 'discourse/widgets/post-menu';
 import { includeAttributes } from 'discourse/lib/transform-post';
 import { addToolbarCallback } from 'discourse/components/d-editor';
 import { addWidgetCleanCallback } from 'discourse/components/mount-widget';
-import { createWidget, decorateWidget, changeSetting } from 'discourse/widgets/widget';
+import { createWidget, reopenWidget, decorateWidget, changeSetting } from 'discourse/widgets/widget';
 import { onPageChange } from 'discourse/lib/page-tracker';
 import { preventCloak } from 'discourse/widgets/post-stream';
 import { h } from 'virtual-dom';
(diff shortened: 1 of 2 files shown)


eviltrout:
This seems good to me, but I do think it should be a PluginAPI and exposed that way, rather than plugins just requiring the method and reopening themselves.

gdpelican:
Okay, I've updated this to be called this way:

```
    withPluginApi('0.1', api => {
      api.reopenWidget('notification-item', {
        url() {
          console.log('some new implementation')
        }
      })
    })
```

app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -308,6 +308,16 @@ class PluginApi {
   }
 
   /**
+   * Exposes the widget update ability to plugins. Updates the widget
+   * registry for the given widget name to include the properties on args
+   * See `reopenWidget` in `discourse/widgets/widget` from more ifo.
+  **/
+
+  reopenWidget(name, args) {
```
eviltrout: You'll also need to bump the API version, otherwise people don't have a version to tie this new function to.

gdpelican: Oh, don't quite follow that yet. Is this the current version here?

```
function getPluginApi(version) {
  version = parseFloat(version);
  if (version <= 0.6) {
```

eviltrout: Yes! If you just update that to 0.7 then anyone who uses 0.7 will be guaranteed to have your new function.

gdpelican:
Cool, I've popped that up to 0.7 now. Let me know if there's anything else? @eviltrout

eviltrout:
Thanks, it looks great!

PR #4659, opened 2017-01-17
Validate censored words

rimian:
This prevents admins from entering reserved characters in censored words which causes 500 errors when creating topics. 

screen shot 2017-01-17 at 1 01 54 pm

### DIFF
--- a/lib/site_setting_validations.rb
+++ b/lib/site_setting_validations.rb
@@ -5,6 +5,10 @@ module SiteSettingValidations
     raise Discourse::InvalidParameters.new(I18n.t("errors.site_settings.#{key}"))
   end
 
+  def validate_censored_words(new_val)
+    validate_error :invalid_character if new_val.match(/^\w*$/).nil?
+  end
+
   def validate_min_username_length(new_val)
     validate_error :min_username_length_range if new_val > SiteSetting.max_username_length
     validate_error :min_username_length_exists if User.where('length(username) < ?', new_val).exists?


rimian:
@coding-horror, @tgxworld what about this?

I am missing translations and specs. There doesn't seem to be any specs for this I can extend. Any advice? 

I'll need help with the translations :)

lib/site_setting_validations.rb
```diff
@@ -5,6 +5,10 @@ def validate_error(key)
     raise Discourse::InvalidParameters.new(I18n.t("errors.site_settings.#{key}"))
   end
 
+  def validate_censored_words(new_val)
+    validate_error :invalid_character if new_val.match(/^\w*$/).nil?
```
discoursebuild: Perhaps `/\w+/` is sufficient here? :thinking:

tgxworld: Perhaps `/\w+/` is sufficient here? Also we need a test case for this :)

rimian: @tgxworld I couldn't find any test cases for validating site settings. Maybe I am missing something?

tgxworld: Looks like no one wrote one yet :P

Perhaps start a new spec in `site_setting_validations_spec.rb`?

tgxworld:
Thanks for the PR here @rimian. After looking into this, I realized people might want to set words like `a**le` and other variants as censored words. As such, we can't really restrict censored words to just `\w`.

PR #4660, opened 2017-01-17
add more SERVER_SIDE_ONLY routes; use pathname in matching

ryantm:
Following up on https://github.com/discourse/discourse/pull/4645 here are some route fixes. I also changed it to match on the pathname instead of the full path, since it seems like that was what was intended anyway (judging by the "^/posts" regex).


eviltrout:
Awesome, thanks!

PR #4661, opened 2017-01-17
add phpBB urls to SERVER_SIDE_ONLY regexes

ryantm:
We migrated our forum from phpBB to nginx. We use nginx in front of
Discourse to redirect the phpBB URLs to the Discourse URLs. Without
this PR, any link to a phpBB URL from inside Discourse does not work
by left clicking on it.

We could fix this by adding an option to Discourse to allow custom
SERVER_SIDE_ONLY regexes, but it seems easier to fix the problem by
adding them here.

### DIFF
--- a/app/assets/javascripts/discourse/lib/url.js.es6
+++ b/app/assets/javascripts/discourse/lib/url.js.es6
@@ -15,6 +15,8 @@ const SERVER_SIDE_ONLY = [
   /^\/posts\/\d+\/raw/,
   /\.rss$/,
   /\.json$/,
+  /^\/viewtopic.php/,
+  /^\/viewforum.php/,
 ];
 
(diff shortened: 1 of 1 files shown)


eviltrout:
I don't think this should be part of Discourse core, as it is specific to your deployment. I would accept a patch to register SERVER_SIDE_ONLY routes via plugin API though.

ryantm:
@eviltrout Would you accept a PR to add a option to Discourse core to add SERVER_SIDE_ONLY routes?

eviltrout:
@ryantm you mean like a site setting? what did you have in mind if not a plugin api?

ryantm:
@eviltrout Yes, I meant a site setting. I was thinking maybe a site setting that is a list of regexs.

PR #4662, opened 2017-01-18
Fix localized group name change

tgxworld:


eviltrout:
One thing to watch out with this is that if we rename a group and `@oldgroupname` exists in the post but is NOT a group, for example within backticks or part of an email address, it will be updated.

Seems quite unlikely to be reported though, we can probably defer on it, as using the proper group name parser for the replacement is much more complicated.

tgxworld:
Thanks for reviewing @eviltrout  :+1:

PR #4663, opened 2017-01-18
PERF: Deferred loading of suggested topics.

tgxworld:
cc @SamSaffron 

For most topics, the suggested topics don't come into view until the user reaches the end of the topic. The query to load suggested topics is hurting us on perf for the initial load as shown by the mini-profiler results below.

screenshot from 2017-01-18 16-48-32

The approach taken here is to only load the suggested topics when the user reaches the end of the topic stream. Even for shorter topics, it has the added benefit of reducing our first load time.


tgxworld:
Unlikely since our footers are rendered only when all the items on the page has been loaded :)

tgxworld:
:tada: CLA Bot is back

PR #4668, opened 2017-01-22
FIX: Embedded Youtube videos lacking margin at bottom

gotens1211:
The markdown was not adding margin in the embedded videos when there was more then one new line. Changed the cook function arguments for youtube videos now it is working fine.

https://meta.discourse.org/t/double-spaced-youtube-embeds-lack-bottom-padding/55748

### DIFF
--- a/app/models/post_analyzer.rb
+++ b/app/models/post_analyzer.rb
@@ -12,36 +12,10 @@ class PostAnalyzer
     @found_oneboxes
   end
 
-#to add margin between youtube links
-def newlinereplace(*args)
-    sp = args[0].split(/\n+/)
-    a = 0
-    x = 0
-    replacem = []
- while a<sp.length do
-   if sp[a].length > 0
-    replacem[x] =sp[a]
-      x+=1
-   end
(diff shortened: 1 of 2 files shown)


tgxworld:
Hmm perhaps we can just add the margins using CSS?

coding-horror:
just CSS is preferred

gotens1211:
Actually the method i used first was conserving the behaviour of the markdown and also the files of lazyYt plugin for youtube videos used by discourse said update if necessary 😅. The video divs are separated with a br tag when there is a newline present in the markdown but if there are more then 
one newlines the br tag is absent which is causing the problem.

I have updated the css file of lazyYt to give some margin to the video container, should i push it ?

ZogStriP:
> I have updated the css file of lazyYt to give some margin to the video container, should i push it ?

Yes, please! 👍

ZogStriP:
Can you remove the cosmetic changes you've made to the `post_analyser.rb` file?

ZogStriP:
Thanks 👍

PR #4669, opened 2017-01-23
FIX: Emoji sets should contain all available emojis for the set.

tgxworld:
* Emojis not available in a particular set is simply not loaded.

### DIFF
--- a/lib/tasks/emoji.rake
+++ b/lib/tasks/emoji.rake
@@ -40,7 +40,10 @@ task "emoji:update" => :environment do
 
     code = cells[1].at_css("a")["name"]
 
-    next unless emojis[code]
+    unless emojis[code]
+      code = code.gsub(/_fe0f/, "")
+      next unless emojis[code]
+    end
 
     apple = cell_to_image(cells[4])
     google = cell_to_image(cells[5])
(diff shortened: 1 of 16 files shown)


lib/tasks/emoji.rake
```diff
@@ -113,7 +114,13 @@ def write_emojis(emojis, aliases, style, folder)
...
 def write_emoji(path, emoji)
-  open(path, "wb") { |f| f << emoji }
-  `pngout #{path}`
-  putc "."
+  begin
+    open(path, "wb") { |f| f << emoji }
+    `pngout #{path}`
+    putc "."
+  ensure
+    if !File.size?(path)
```
ZogStriP: Why not 

```ruby
raise "Failed to write emoji: #{path}" unless File.size?(path)
```

tgxworld: `unless` throws me off most of the time haha

ZogStriP:
```ruby
raise "Failed to write emoji: #{path}" if !File.size?(path)
```

is fine too

tgxworld: O do you mean why not one line it?

tgxworld:
Closing this first since we're only going to support the intersection of emojis.

PR #4675, opened 2017-01-26
FIX: Polls permanently broken if post requires approval.

tgxworld:
https://meta.discourse.org/t/poll-posts-may-not-survive-approval-queue/56278/8

cc @ZogStriP Can you help me to review this? Thank you! :heart:

### DIFF
--- a/plugins/poll/plugin.rb
+++ b/plugins/poll/plugin.rb
@@ -247,11 +247,11 @@ after_initialize do
     end
   end
 
-  validate(:post, :validate_polls) do
+  validate(:post, :validate_polls) do |force=nil|
     return if !SiteSetting.poll_enabled? && (self.user && !self.user.staff?)
 
     # only care when raw has changed!
-    return unless self.raw_changed?
+    return unless self.raw_changed? || force
 
     validator = DiscoursePoll::PollsValidator.new(self)
     return unless (polls = validator.validate_polls)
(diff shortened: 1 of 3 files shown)


plugins/poll/plugin.rb
```diff
@@ -247,11 +247,11 @@ def voters
     end
   end
 
-  validate(:post, :validate_polls) do
+  validate(:post, :validate_polls) do |force=nil|
```
ZogStriP: Is the `force` argument used somewhere?

plugins/poll/plugin.rb
```diff
@@ -268,6 +268,29 @@ def voters
     true
   end
 
+  NewPostManager.add_handler(1) do |manager|
```
ZogStriP: Can't this be moved to the `on(:approved_post)` event?

plugins/poll/plugin.rb
tgxworld: Yup, used [here](https://github.com/tgxworld/discourse/blob/bed11dfa0e6319be5228687e78de4dacd92d3a93/plugins/poll/plugin.rb#L290).

plugins/poll/plugin.rb
tgxworld: We have to validate before the `QueuedPost` record is created otherwise we run the risk of having a post that fails validation being stucked in the queue.

PR #4677, opened 2017-01-28
other keyword is missing in locale, it causes broken link

mehmetkemalbayer:
missing keyword "other" in tr locale at "read_more_MF" causes broken link problem. This pull request only adds missing keyword to locale.
In fact the problem is funny. Word other means "diğer" in Turkish. Person translates the word other but forgets the keyword in (true{} false{} other{}) structure.


techAPJ:
Hi @mehmetkemalbayer,

Thanks for this PR but we accept changes in locale files only via [Transifex](https://www.transifex.com/discourse/discourse-org/). Please make the same change there and it will be pulled in this Wednesday.

mehmetkemalbayer:
Hi,
I just added a suggestion at transifex. Please check it.
Regards.

Envoyé de mon iPhone

PR #4678, opened 2017-01-30
Update client.vi.yml

quocbao9996:
Update some missing Vietnamese translation.


ZogStriP:
Thanks for the PR but we only accept translations updates through [transifex](transifex.com/discourse).

PR #4683, opened 2017-02-02
FIX: Update Topic#relative_url to Return Relative URL

oldbonsai:
# TL;DR

Removed `Discourse.base_uri` from the output of `Topic#relative_url` to make the method do what its name implies.

# Details

The following changes to the codebase are proposed:

* Remove `Discourse.base_uri` from `Topic#relative_url`.
* Alter the permalink model to use `Topic#url` instead of `Topic#relative_url`. Preserves the output (full URL) and makes the methods consistent within `Permalink#target_url`.
* Modify app/views/list/list.erb to use the full topic URL for consistency with the full category URL and to eliminate a potentially confusing or erroneous reference to Topic#relative_url.

### DIFF
--- a/app/models/permalink.rb
+++ b/app/models/permalink.rb
@@ -76,7 +76,7 @@ class Permalink < ActiveRecord::Base
   def target_url
     return external_url if external_url
     return "#{Discourse::base_uri}#{post.url}" if post
-    return topic.relative_url if topic
+    return topic.url if topic
     return category.url if category
     nil
   end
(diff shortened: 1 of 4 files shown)


ZogStriP:
Hmm, I think this breaks subfolder installs. `Discourse.base_uri` returns the subfolder name when there is one.

oldbonsai:
_facepalm_ I didn’t see the difference between `Discourse.base_uri` and `Discourse.base_url`. Thanks for calling that out. PR retracted.

PR #4684, opened 2017-02-03
Fix reset context when context type changes

tgxworld:


tgxworld:
@eviltrout Can you review this? Thank you!

PR #4686, opened 2017-02-03
FIX: Show groups that user is owner of on groups page.

tgxworld:
cc @techAPJ Can you review this for me? Thank you!

https://meta.discourse.org/t/group-owners-cannot-see-group-if-group-is-visible-to-all-users-is-off/56688/10


techAPJ:
LGTM :+1:

PR #4689, opened 2017-02-05
Pwdreset

huero:


SamSaffron:
also, this is a UX change, we need to see some screenshots here.

SamSaffron:
HA ... what is going on here. this is all @nlalonde's work, why are you submitting a PR?

PR #4691, opened 2017-02-06
Moving posts issues

deivid-rodriguez:
Hi!

We've noticed two issues when moving posts in our discourse instance.

* We get a crash when moving a post to a closed topic. The crash looks like this error https://meta.discourse.org/t/error-when-moving-post-to-existing-topic/38790 and I think the first commit here should fix it.

* We also get a crash when moving a post to a topic in the case where the autogenerated moderator post ("a post was moved to topic [link-to-topic]") does not meet our configured min post length validation. The second commit here should fix that.

Let me know your thoughts! :)

### DIFF
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -155,8 +155,6 @@ class PostMover
       action_code: "split_topic"
     )
 
-    return unless moderator_post
-
     # We want to insert the moderator post where the previous posts were
     # in the stream, not at the end.
     moderator_post.update_attributes!(post_number: @first_post_number_moved, sort_order: @first_post_number_moved) if @first_post_number_moved.present?
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -146,15 +146,20 @@ class PostMover
   def create_moderator_post_in_original_topic
     move_type_str = PostMover.move_types[@move_type].to_s
 
-    original_topic.add_moderator_post(
+    moderator_post = original_topic.add_moderator_post(
       user,
       I18n.t("move_posts.#{move_type_str}_moderator_post",
              count: post_ids.count,
              topic_link: "[#{destination_topic.title}](#{destination_topic.relative_url})"),
       post_type: Post.types[:small_action],
-      action_code: "split_topic",
-      post_number: @first_post_number_moved
+      action_code: "split_topic"
     )
+
+    return unless moderator_post
+
+    # We want to insert the moderator post where the previous posts were
+    # in the stream, not at the end.
+    moderator_post.update_attributes!(post_number: @first_post_number_moved, sort_order: @first_post_number_moved) if @first_post_number_moved.present?
   end
 
   def posts
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -157,7 +157,7 @@ class PostMover
 
     # We want to insert the moderator post where the previous posts were
     # in the stream, not at the end.
-    moderator_post.update_attributes!(post_number: @first_post_number_moved, sort_order: @first_post_number_moved) if @first_post_number_moved.present?
+    moderator_post.update_columns(post_number: @first_post_number_moved, sort_order: @first_post_number_moved) if @first_post_number_moved.present?
   end
 
   def posts
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -652,9 +652,6 @@ SQL
 
     if (new_post = creator.create) && new_post.present?
       increment!(:moderator_posts_count) if new_post.persisted?
-      # If we are moving posts, we want to insert the moderator post where the previous posts were
-      # in the stream, not at the end.
-      new_post.update_attributes!(post_number: opts[:post_number], sort_order: opts[:post_number]) if opts[:post_number].present?
 
       # Grab any links that are present
       TopicLink.extract_from(new_post)
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -650,13 +650,13 @@ SQL
                               skip_validations: true,
                               custom_fields: opts[:custom_fields])
 
-    if (new_post = creator.create) && new_post.present?
-      increment!(:moderator_posts_count) if new_post.persisted?
+    new_post = creator.create
 
-      # Grab any links that are present
-      TopicLink.extract_from(new_post)
-      QuotedPost.extract_from(new_post)
-    end
+    increment!(:moderator_posts_count) if new_post.persisted?
+
+    # Grab any links that are present
+    TopicLink.extract_from(new_post)
+    QuotedPost.extract_from(new_post)
 
     new_post
   end
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -639,7 +639,6 @@ SQL
 
   def add_moderator_post(user, text, opts=nil)
     opts ||= {}
-    new_post = nil
     creator = PostCreator.new(user,
                               raw: text,
                               post_type: opts[:post_type] || Post.types[:moderator_action],
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -107,8 +107,8 @@ class PostCreator
       topic_creator = TopicCreator.new(@user, guardian, @opts)
       return false unless skip_validations? || validate_child(topic_creator)
     else
-      @topic = Topic.find_by(id: @opts[:topic_id])
-      if (@topic.blank? || !guardian.can_create?(Post, @topic))
+      @topic = Topic.with_deleted.find_by(id: @opts[:topic_id])
+      if (@topic.blank? || !guardian.can_create?(Post, @topic)) && !skip_validations?
         errors[:base] << I18n.t(:topic_not_found)
         return false
       end
@@ -362,13 +362,15 @@ class PostCreator
   end
 
   def create_topic
-    return if @topic
-    begin
-      topic_creator = TopicCreator.new(@user, guardian, @opts)
-      @topic = topic_creator.create
-    rescue ActiveRecord::Rollback
-      rollback_from_errors!(topic_creator)
+    unless @topic
+      begin
+        topic_creator = TopicCreator.new(@user, guardian, @opts)
+        @topic = topic_creator.create
+      rescue ActiveRecord::Rollback
+        rollback_from_errors!(topic_creator)
+      end
     end
+
     @post.topic_id = @topic.id
     @post.topic = @topic
     if @topic && @topic.category && @topic.category.all_topics_wiki
--- a/spec/models/post_mover_spec.rb
+++ b/spec/models/post_mover_spec.rb
@@ -64,6 +64,12 @@ describe PostMover do
         topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
         expect(topic.posts).to include(a_moderator_post(post_number: 2, sort_order: 2))
       end
+
+      it "adds a moderator post even if it doesn't meet length requirements" do
+        SiteSetting.stubs(:min_post_length).returns(999)
+        topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
+        expect(topic.reload.posts).to include(a_moderator_post)
+      end
     end
 
     context "errors" do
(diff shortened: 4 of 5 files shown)


deivid-rodriguez:
Rebased this PR and added a couple of simplifications to the code.

app/models/topic.rb
```diff
@@ -655,16 +654,17 @@ def add_moderator_post(user, text, opts=nil)
...
 
-      # Grab any links that are present
-      TopicLink.extract_from(new_post)
-      QuotedPost.extract_from(new_post)
-    end
+    increment!(:moderator_posts_count) if new_post.persisted?
+
+    # If we are moving posts, we want to insert the moderator post where the previous posts were
+    # in the stream, not at the end.
+    new_post.update_columns(post_number: opts[:post_number], sort_order: opts[:post_number]) if opts[:post_number].present?
```
SamSaffron: this feels super risky to me, how do we ensure we don't have 2 posts with the same post_number? I think this logic belongs in post mover not in add_moderator_post

deivid-rodriguez: @SamSaffron Do you mean the fact of updating the post number here or the fact that I'm removing the validation?

If it's the latter, I removed the validation because it was getting in the middle for us. Validator posts shouldn't follow the same validation rules as other posts. For example, they shouldn't have a minimum length, which was making moving some posts crash for us. But you're totally right about post number uniqueness, not sure if that could happen, but it's definitely something to be aware of.

deivid-rodriguez: Hei @SamSaffron. I had a closer look at this and try to address your concern in 3f99265.

I also rebased the PR and got the tests green again. https://github.com/discourse/discourse/commit/477eb0591e55a6d89d6b1928d6ddd8575a3ced73 added a couple of new tests that had uncovered some more problems with the old patch.

Let me know.

deivid-rodriguez: Ah, I forgot. I also added one extra simplification in https://github.com/discourse/discourse/pull/4691/commits/6e14eb33f7a6ea59abbedfcd9eeb149d248c4932. The idea is that by checking `ignore_validations?` at a single place we won't have any more bugs similar to the one fixed by https://github.com/discourse/discourse/pull/4691/commits/aa5b98ed09a384074deb0b192e889d2f48aa5782.

tgxworld:
@deivid-rodriguez  Is this PR ready to be reviewed again?

deivid-rodriguez:
It was ready 25 days ago. Not sure if any recent changes in master affect this.

deivid-rodriguez:
I just rebased this PR against latest master and, funny enough, there's one failing test which is the test I added in my other PR (https://github.com/discourse/discourse/pull/4701).

I'll try to investigate what's going on and to check if https://meta.discourse.org/t/error-when-moving-post-to-existing-topic/38790 can still be reproduced via unit test.

deivid-rodriguez:
@tgxworld I reviewed this today. Conclusions:

* PR rebased and tests green again.

* I can no longer reproduce either https://meta.discourse.org/t/error-when-moving-post-to-existing-topic/38790 or the crash I was getting when moving a post to a closed topic. Note that even if the crashes happened in the same place, the bugs had different reproductions. Some recent changes must have fixed those.

* The second bug in the PR description is still reproducible and fixed by this PR.

* There's another bug fix that I had detected while working on this PR. It's https://github.com/discourse/discourse/pull/4691/commits/3c01426b82bf0a80b0f84b717352a92272aaa03e.

Let me know!

SamSaffron:
This is just way too big imo, what is the minimal amount of code needed to fix the specific bug? 

Can you submit one PR per fix that just fixes the specifics? Surely large chunks of post creator, post mover and topic don't need to be re-written to fix the issue. 

I am super concerned when such an enormous change is made to our post pipeline, I worry that it will break under all sorts of conditions.

deivid-rodriguez:
@SamSaffron I'll split this into two separate PR's, one for each bug, so this is easier to review. Ok?

deivid-rodriguez:
Actually I just removed the last commit which was refactoring a duplicated check out of a block so it involved changing a lot of indentation affecting the diff size. Is this good for review or you want it smaller?

app/models/topic.rb
```diff
@@ -650,16 +649,13 @@ def add_moderator_post(user, text, opts=nil)
                               skip_validations: true,
                               custom_fields: opts[:custom_fields])
 
-    if (new_post = creator.create) && new_post.present?
-      increment!(:moderator_posts_count) if new_post.persisted?
-      # If we are moving posts, we want to insert the moderator post where the previous posts were
-      # in the stream, not at the end.
-      new_post.update_attributes!(post_number: opts[:post_number], sort_order: opts[:post_number]) if opts[:post_number].present?
+    new_post = creator.create
```
tgxworld: Are we sure `new_post` will never be `nil`?

app/models/post_mover.rb
```diff
@@ -146,15 +146,18 @@ def enqueue_notification_job
...
              topic_link: "[#{destination_topic.title}](#{destination_topic.relative_url})"),
       post_type: Post.types[:small_action],
-      action_code: "split_topic",
-      post_number: @first_post_number_moved
+      action_code: "split_topic"
     )
+
+    # We want to insert the moderator post where the previous posts were
+    # in the stream, not at the end.
+    moderator_post.update_columns(post_number: @first_post_number_moved, sort_order: @first_post_number_moved) if @first_post_number_moved.present?
```
tgxworld: hmm `add_moderator_post` is used in other places as well so I don't think you can extract this logic into `PostMover` only.

lib/post_creator.rb
```diff
@@ -362,13 +362,15 @@ def find_category_id
   end
 
   def create_topic
-    return if @topic
-    begin
-      topic_creator = TopicCreator.new(@user, guardian, @opts)
-      @topic = topic_creator.create
-    rescue ActiveRecord::Rollback
-      rollback_from_errors!(topic_creator)
+    unless @topic
```
tgxworld: This refactor here is a cosmetic change since I don't think it changes any existing logic.

spec/models/post_mover_spec.rb
```diff
@@ -47,16 +55,21 @@
     context 'success' do
 
       it "enqueues a job to notify users" do
-        topic.stubs(:add_moderator_post)
+        Jobs.stubs(:enqueue)
```
tgxworld: We generally try to avoid stubbing methods without the parameters since it makes the test brittle if we need some other job to be en-queued in the future.

spec/models/post_mover_spec.rb
```diff
@@ -47,16 +55,21 @@
...
         Jobs.expects(:enqueue).with(:notify_moved_posts, post_ids: [p2.id, p4.id], moved_by_id: user.id)
         topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
       end
 
       it "adds a moderator post at the location of the first moved post" do
-        topic.expects(:add_moderator_post).with(user, instance_of(String), has_entries(post_number: 2))
         topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
+        expect(topic.posts).to include(a_moderator_post(post_number: 2, sort_order: 2))
```
tgxworld: I don't think this assertion alone is sufficient since the test may be a false positive if `topic.posts`already contains the moderator post.

spec/models/post_mover_spec.rb
```diff
@@ -1,6 +1,14 @@
 require 'rails_helper'
 
 describe PostMover do
+  RSpec::Matchers.define :a_moderator_post do |attributes|
```
tgxworld: Hmm I'm not really a fan of ths matcher here. I prefer to have less custom rspec magic and just be explicit in our tests.

spec/models/post_mover_spec.rb
```diff
@@ -47,16 +55,21 @@
...
       end
 
       it "adds a moderator post at the location of the first moved post" do
-        topic.expects(:add_moderator_post).with(user, instance_of(String), has_entries(post_number: 2))
         topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
+        expect(topic.posts).to include(a_moderator_post(post_number: 2, sort_order: 2))
       end
 
+      it "adds a moderator post even if it doesn't meet length requirements" do
+        SiteSetting.stubs(:min_post_length).returns(999)
```
tgxworld: No nee to stub SiteSettings here. `SiteSetting.min_post_length = 999` will be sufficient.

spec/models/post_mover_spec.rb
```diff
@@ -193,7 +205,6 @@
         end
 
         it "moving all posts will close the topic" do
-          topic.expects(:add_moderator_post).twice
```
tgxworld: If this is removed, I think we need to add tests to ensure that two moderator posts have been added?

spec/models/topic_spec.rb
```diff
@@ -599,18 +599,14 @@ def build_topic_with_title(title)
...
-    context "when moderator post fails to be created" do
-      before do
-        user.toggle!(:blocked)
-      end
-
-      it "should not increment moderator_posts_count" do
-        expect(topic.moderator_posts_count).to eq(0)
-
-        topic.add_moderator_post(user, "winter is never coming")
+    it "creates a moderator post ignoring validations" do
```
tgxworld: Hmm this feels like a cosmetic only change to me. Not sure if I am missing something.

app/models/topic.rb
deivid-rodriguez: Yes, that's the idea. After this set of changes, `PostCreator` when instantiated with the `skip_validations` parameter set to `true  should always allow instantiating an actual post and never return `nil`.

app/models/post_mover.rb
deivid-rodriguez: Yes, but the only usage passing the `post_number` option was this one, which was the previous condition to apply this logic. So Sam's suggestion of moving this code here makes sense to me.

lib/post_creator.rb
deivid-rodriguez: Nope, it's needed for the commit the change belongs to. It would be just cosmetic if the `unless` block wrapped the whole method, but it's not the case.

spec/models/post_mover_spec.rb
deivid-rodriguez: I disagree, but I can remove it to fit your taste :)

spec/models/post_mover_spec.rb
deivid-rodriguez: But the whole point of this test is to check that another job (`:notify_moved_posts`) has been enqueued, so any other jobs should be ignored. That's basically the same the old stub did but I had to change it because the new code ensures that `add_moderator_post` returns an actual post and uses the return value, so leaving that stub there would crash. So I just stubbed whatever jobs are enqueued from inside `add_moderator_post`.

spec/models/post_mover_spec.rb
deivid-rodriguez: The data setup for this test does not include any moderations, so it should not include any moderator posts. Do you want me to check that before the `move_posts` call as well?

spec/models/post_mover_spec.rb
deivid-rodriguez: Mmmm, ok, so I should set it back at the end of the test or you have other mechanisms to not leak state?

spec/models/post_mover_spec.rb
deivid-rodriguez: That's right, my bad! I'll re-add those checks.

spec/models/topic_spec.rb
deivid-rodriguez: Nope. After these changes, moderator posts never fail to be created. So instead of asserting that the post is not created, we assert that it is created.

deivid-rodriguez:
@tgxworld Thanks for the review! After having a look at your comments, I think it's probably better that I split this into one bugfix per PR so it's easier to review... Should I do that?

deivid-rodriguez:
I did it in #4863.

spec/models/topic_spec.rb
deivid-rodriguez: @tgxworld I missed your point here, sorry. Even if this actually modifies the expectation of this test, it also includes some cosmetic changes. I removed the comestic changes in #4863!

PR #4693, opened 2017-02-07
Update INSTALL-cloud.md

pfaffman:
Perhaps I live in a fantasy world, but maybe this will help people include whether they even know about these instructions when they post questions like [this one](https://meta.discourse.org/t/installing-discourse-for-the-first-time-using-vultr-hosting/57016?u=pfaffman).


coding-horror:
I appreciate the sentiment, but I don't think this is necessary.

PR #4694, opened 2017-02-07
copyedit - remove curly quotes

jomaxro:
When looking at @coding-horror's copyedit commit on Saturday I noticed that he'd replaced the straight quotes with curly quotes in addition to the textual changes.  This PR reverts those changes, as well as removing all other instances of curly quotes on the English client and server YAML files.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -3384,7 +3384,7 @@ en:
 
       invites:
         add_user: "add"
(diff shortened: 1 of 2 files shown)


coding-horror:
Why would this be removed? It is valid.

jomaxro:
Their usage seemed inconsistent - most copy seemed to use straight quotes, so I thought consistency would be good.

PR #4696, opened 2017-02-08
Webhooks features

fantasticfears:
This PR is made of two commits. @tgxworld 

The first commit makes the webhook types easier to be customized.

The other is about payload size. There are many requests asking to reduce the payload in the webhooks feedback section. https://meta.discourse.org/t/webhooks-feedback/49046 So serveral customized webhook serializers are created for that reason.

### DIFF
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -8,19 +8,6 @@ module Jobs
 
       args = args.dup
 
-      if args[:topic_id]
-        args[:topic_view] = TopicView.new(args[:topic_id], Discourse.system_user)
-      end
-
-      if args[:post_id]
-        # deleted post so skip
-        return unless args[:post] = Post.find_by(id: args[:post_id])
-      end
-
-      if args[:user_id]
-        return unless args[:user] = User.find_by(id: args[:user_id])
-      end
-
       web_hook = WebHook.find(args[:web_hook_id])
 
       unless args[:event_type] == 'ping'
@@ -29,6 +16,10 @@ module Jobs
           !web_hook.group_ids.include?(args[:group_id]))
         return if web_hook.category_ids.present? && (!args[:category_id].present? ||
           !web_hook.category_ids.include?(args[:category_id]))
+
+        model = args[:event_type].to_s
+        record_id = "#{model}_id".to_sym
+        return unless args[model] = WebHookEventType.const_get("#{model.classify}Type").load_record(args[record_id])
       end
 
       web_hook_request(args, web_hook)
@@ -36,8 +27,28 @@ module Jobs
 
     private
 
-    def web_hook_request(args, web_hook)
+    def build_web_hook_body(args, web_hook)
+      body = {}
+      guardian = Guardian.new(Discourse.system_user)
 
+      if args[:event_type] == 'ping'
+        body[:ping] = 'OK'
+      else
+        model = args[:event_type].to_s
+        klass = WebHookEventType.const_get("#{model.classify}Type").serializer
+        body[model] = klass.new(args[model], scope: guardian, root: false).as_json
+      end
+
+      raise Discourse::InvalidParameters.new if body.empty?
+
+      new_body = Plugin::Filter.apply(:after_build_web_hook_body, self, body)
+
+      Rails.logger.debug("Blank web hook body: #{args}") if new_body.blank?
+
+      MultiJson.dump(new_body)
+    end
+
+    def web_hook_request(args, web_hook)
       uri = URI(web_hook.payload_url)
       conn = Excon.new(uri.to_s,
                        ssl_verify_peer: web_hook.verify_certificate,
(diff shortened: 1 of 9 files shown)


app/serializers/web_hook_user_serializer.rb
```diff
@@ -0,0 +1,33 @@
+class WebHookUserSerializer < UserSerializer
+  def can_edit
+    false
+  end
+
+  def include_email?
```
tgxworld: I wonder if we should just be explicit here and include what we want instead of excluding since the payload size is the concern. Overtime, people will add more stuff to `UserSerializer` and it'll cause the web hook payload to bloat without anyone knowing about it

app/serializers/web_hook_topic_view_serializer.rb
```diff
@@ -0,0 +1,221 @@
+require_dependency 'pinned_check'
```
tgxworld: This serializer is huge, how much of the payload does it actually reduce as compared to `TopicViewSerializer`?

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -2,23 +2,30 @@
...
-        # deleted post so skip
-        return unless args[:post] = Post.find_by(id: args[:post_id])
-      end
-
-      if args[:user_id]
-        return unless args[:user] = User.find_by(id: args[:user_id])
+      Jobs::EmitWebHookEvent.web_hook_serialization_properties.each do |property|
+        type_id = "#{property[:model]}_id".to_sym
+        next unless args[type_id]
+        return unless (args[property[:model]] = (property[:load_record] || ->(id) { property[:model].to_s.classify.constantize.find_by(id: id.to_i) }).call(args[type_id]))
```
tgxworld: I'm not sure if I like the use of a `lambda` here. It makes the code really hard to read.

app/serializers/web_hook_topic_view_serializer.rb
fantasticfears: It's a copy from `TopicViewSerializer` without the post stream so it's much less (could be 1/3 of original size).

app/serializers/web_hook_user_serializer.rb
fantasticfears: Yes, I made a copy of code from those serializers so that they can be updated and controlled separately for the webhook events.

app/jobs/regular/emit_web_hook_event.rb
fantasticfears: I've made the `lambda` explicit in the meta array so anyone who intends to add new model will need to write the lambda there. And this part simply just executes the lambda.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -2,23 +2,47 @@
...
   class EmitWebHookEvent < Jobs::Base
+    # Required option:
+    #   model                   - The symbol represents the model class
+    #   load_record             - A lambada for loading the records
+    #
+    # Acceptable options:
+    #   serializer_name         - A serializer class name otherwise `<Model>Serializer` is used.
+    def self.web_hook_serialization_properties
+      @@web_hook_serialization_properties ||= [
+        { model: :topic, load_record: ->(topic_id) { TopicView.new(topic_id, Discourse.system_user) }, serializer_name: "WebHookTopicViewSerializer" },
```
tgxworld: I'll prefer if we define these a individual methods. Much easier to test if we want to.

app/serializers/web_hook_post_serializer.rb
```diff
@@ -0,0 +1,200 @@
+class WebHookPostSerializer < ApplicationSerializer
```
tgxworld: > Yes, I made a copy of code from those serializers so that they can be updated and controlled separately for the webhook events.

Oops I'm sorry but I didn't know we were still keeping so many attributes. Duplicating all these logic is definitely a no-no. Can I get an idea of how many attributes we actually want to exclude from each serializer?

app/jobs/regular/emit_web_hook_event.rb
fantasticfears: Sure.

app/serializers/web_hook_post_serializer.rb
fantasticfears: For topic, it doesn't serialize post stream and ~10 attributes. For posts and users, ~10 less at least.

tgxworld:
@fantasticfears Sorry for the delay on this. Looks like the build is failing?

tgxworld:
Thank you for your work :+1:

fantasticfears:
No problem :) The PR is fixed.

Now there is only 1 object in the webhook payload. Now it only cuts down some huge attributes. And the rules are separate in the methods encapsulated in their own modules.

I think the important lesson for me as of working on webhooks is that focusing on data/interface is rather important cause I have many ideas to make it much clear... Anyway, the PR should be small and incremental changes based on the feedback is important at this moment.

tgxworld:
Closing in favor of https://github.com/discourse/discourse/pull/4755

PR #4697, opened 2017-02-09
using '==' instead of '=' in conditionals

sessionstop:
using '==' instead of '=' in conditionals


ZogStriP:
Thanks for your PR but using `=` is actually valid since we're not testing equality but setting the value of the `counts` variable to `counts[post.id]`.

PR #4698, opened 2017-02-09
Create correct topic path for REST adapter

jannolii:
Updating topic from my plugin using...

`topic.update(props);`

... I discovered that PUT was sent to incorrect URL (was getting 404 because topics were sent to /topics/182 not /t/my_topics_slug/182). And also the params that were created for topics controller had multiple levels and would be ignored even when fields were added to serializer

### DIFF
--- a/app/assets/javascripts/discourse/adapters/topic.js.es6
+++ b/app/assets/javascripts/discourse/adapters/topic.js.es6
@@ -1,7 +1,35 @@
 import { ajax } from 'discourse/lib/ajax';
 import RestAdapter from 'discourse/adapters/rest';
+import Topic from 'discourse/models/topic';
+
+export function Result(payload, responseJson) {
+  this.payload = payload;
+  this.responseJson = responseJson;
+  this.target = null;
+}
 
 export default RestAdapter.extend({
+  update(store, type, id, attrs) {
+    const data = {};
+    const typeField = Ember.String.underscore(type);
+    Object.assign(data, attrs);
+    return this.pathFor(store, type, id).then(function(path) {
+      return ajax(path, { method: 'PUT', data }).then(function(json) {
+        const result = new Result(json[typeField], json);
+        console.log(result);
+        return result;
+      });
+    });
+  },
+  pathFor(store, type, findArgs) {
+    var self = this;
+    return Topic.find(findArgs, {}).then(function(result) {
+      if (result) {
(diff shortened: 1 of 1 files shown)


SamSaffron:
@eviltrout this one is yours :)

app/assets/javascripts/discourse/adapters/topic.js.es6
```diff
@@ -1,7 +1,35 @@
...
 
 export default RestAdapter.extend({
+  update(store, type, id, attrs) {
+    const data = {};
+    const typeField = Ember.String.underscore(type);
+    Object.assign(data, attrs);
+    return this.pathFor(store, type, id).then(function(path) {
+      return ajax(path, { method: 'PUT', data }).then(function(json) {
+        const result = new Result(json[typeField], json);
+        console.log(result);
```
tgxworld: stray console.log here :)

app/assets/javascripts/discourse/adapters/topic.js.es6
```diff
@@ -1,7 +1,35 @@
...
+        return result;
+      });
+    });
+  },
+  pathFor(store, type, findArgs) {
+    var self = this;
+    return Topic.find(findArgs, {}).then(function(result) {
```
tgxworld: you could use the arrow function here `result =>` and remove the need for `var self = this;`

app/assets/javascripts/discourse/adapters/topic.js.es6
```diff
@@ -1,7 +1,35 @@
...
+      if (result) {
+        self.set('path', self.basePath(store, type, findArgs) + 't/' + result.slug);
```
tgxworld: Can be written as `${self.basePath(store, type, findArgs)}/t/${result.slug}` with ES6's [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals).

eviltrout:
This seems incorrect to me. The big red flag is that `pathFor` is not an asynchronous method but you're calling asynchronously in it to set the path. 

That method could even be called many times, triggering lots of rogue AJAX requests in the background.

Could you not just use the `Topic.update` method that other pieces of the Discourse codebase use?

jannolii:
@eviltrout , you're right. Was thinking in totally wrong direction here. Though strange that such inconsistency with path exists for topic (and for category too).

PR #4700, opened 2017-02-10
fix replay btn position on mobile and desktop

gsambrotta:
Move replay button at the bottom of topic when the user is log out.
Move the button on the right on desktop and central on mobile.

Related to: https://meta.discourse.org/t/position-of-reply-button-to-the-right/49573/2


coding-horror:
Thanks this seems fine! Helps anon users.

PR #4704, opened 2017-02-13
remove personal email from fixtures

rimian:
I'm writing acceptance tests against the user fixture in my plugin. It would be a good idea to use a generic email address as I'd like to respect eviltrout's privacy and not have the email leak out into other repositories :)


SamSaffron:
@eviltrout since it is your email :) note email is still there in the commit messages so I am not sure how much this is really important

rimian:
@SamSaffron here's my test: https://github.com/choiceaustralia/choice-discourse/blob/user-page/test/javascripts/acceptance/choice-discourse-test.es6

I was checking the email address which means I need to hard code it in a third party repo. It would be good not to have to worry about it. 

Many thanks!

PR #4705, opened 2017-02-13
new: server plugin outlet for indexable robots.txt

vinothkannans:
As per @SamSaffron's comment below I added a server plugin outlet.

> There is no built in extensibility to robots.txt generation, this needs to be added so it does not overwrite it completely (instead plugin should inject the instructions). Otherwise, each time we touch robots.txt plugin needs to be updated.

https://meta.discourse.org/t/discourse-sitemap-plugin/40348/32?u=vinothkannans

### DIFF
--- a/app/views/robots_txt/index.erb
+++ b/app/views/robots_txt/index.erb
@@ -15,3 +15,5 @@ Disallow: /search
 Disallow: /search/
 Disallow: /tags
 Disallow: /tags/
+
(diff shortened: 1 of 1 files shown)


coding-horror:
can you review @SamSaffron

SamSaffron:
Seems fine to me, I will backport this to stable and beta so we can junk the override from the plugin

PR #4706, opened 2017-02-13
FEATURE: Add user name in search results 

gsambrotta:
I add a user original name next to username in search reults.

Based on #pr-welcome issue: 
https://meta.discourse.org/t/in-search-result-drop-box-add-given-name-next-to-username/48735/4

QUESTION:
I try to search if there is any test to fix but i didn't find any. Akthiugh i'm pretty new to contributing to discourse so if you have any suggestion about being sure to find test if written and being sure that each test passed, please share with me

### DIFF
--- a/app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
+++ b/app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
@@ -46,7 +46,7 @@ function postResult(result, link, term) {
 }
 
 createSearchResult('user', 'path', function(u) {
-  return [ avatarImg('small', { template: u.avatar_template, username: u.username }), ' ', u.username ];
+  return [ avatarImg('small', { template: u.avatar_template, username: u.username }), ' ', h('span.user-results', h('b', u.username)), ' ',  h('span.user-results', u.name ? u.name : '') ];
 });
 
 createSearchResult('topic', 'url', function(result, term) {
--- a/app/serializers/search_result_user_serializer.rb
+++ b/app/serializers/search_result_user_serializer.rb
@@ -1,0 +1,3 @@
+class SearchResultUserSerializer < BasicUserSerializer
+  attributes :name
+end
--- a/app/assets/stylesheets/common/base/menu-panel.scss
+++ b/app/assets/stylesheets/common/base/menu-panel.scss
@@ -184,6 +184,10 @@
       display: block;
       padding: 5px;
       transition: all linear .15s;
+
+      .user-results {
+        color: dark-light-choose(scale-color($primary, $lightness: 30%), scale-color($secondary, $lightness: 70%));
+      }
     }
 
     &:hover a:not(.badge-notification) {
--- a/app/serializers/grouped_search_result_serializer.rb
+++ b/app/serializers/grouped_search_result_serializer.rb
@@ -1,6 +1,6 @@
 class GroupedSearchResultSerializer < ApplicationSerializer
   has_many :posts, serializer: SearchPostSerializer
-  has_many :users, serializer: BasicUserSerializer
+  has_many :users, serializer: SearchResultUserSerializer
   has_many :categories, serializer: BasicCategorySerializer
   attributes :more_posts, :more_users, :more_categories
 end


xfalcox:
> QUESTION:
I try to search if there is any test to fix but i didn't find any. Akthiugh i'm pretty new to contributing to discourse so if you have any suggestion about being sure to find test if written and being sure that each test passed, please share with me

There are some tests in the links below:

https://github.com/discourse/discourse/blob/df8f365d999e113615c51efee4cfa3115c267783/spec/components/search_spec.rb#L103

https://github.com/discourse/discourse/blob/4332f0dde1e06abfeec091ef3f930d9254408b87/spec/models/user_search_spec.rb

gsambrotta:
This PR is on hold till issue [explained here](https://meta.discourse.org/t/in-search-result-drop-box-add-given-name-next-to-username/48735/22) is fixed.

app/serializers/search_result_user_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+      User.avatar_template(user[:username], user[:uploaded_avatar_id])
+    else
+      user.try(:avatar_template)
+    end
+  end
+
+  def user
+    object[:user] || object
+  end
+
```
fantasticfears: You will still need `include_name?` method here if you build this new serializer. Sam refers to another thing on the Meta.

gsambrotta: @fabianoleittes i reintegrated `include_name?` and test it in different part of the app. Looks everything works properly. Thanks

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
```diff
@@ -46,7 +47,8 @@ function postResult(result, link, term) {
 }
 
 createSearchResult('user', 'path', function(u) {
-  return [ avatarImg('small', { template: u.avatar_template, username: u.username }), ' ', u.username ];
+  console.log(u);
+  return [ avatarImg('small', { template: u.avatar_template, username: u.username }), ' ', h('span.user-results', h('b', u.username)), ' ',  h('span.user-results', u.name ? u.name : '') ];
```
fantasticfears: Now you have a different case where `name` might be not available. For example, the setting `enable_name` is disabled.

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
```diff
@@ -21,6 +21,7 @@ class Highlighted extends RawHtml {
 function createSearchResult(type, linkField, fn) {
   return createWidget(`search-result-${type}`, {
     html(attrs) {
+      console.log(attrs);
```
fantasticfears: Dangling debugging stuff, should be removed

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
```diff
@@ -46,7 +47,8 @@ function postResult(result, link, term) {
 }
 
 createSearchResult('user', 'path', function(u) {
-  return [ avatarImg('small', { template: u.avatar_template, username: u.username }), ' ', u.username ];
+  console.log(u);
```
fantasticfears: Same

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
gsambrotta: ops! thanks!

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
gsambrotta: :/ thank you again!

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
gsambrotta: Here i'm checking if `name` is present in the attributes i got from the serializer. If is present it will show, if is not, it will not.
If `enable_name` is disable, server or client side, it will not be passed as attribute which will result as `false` for `u.name`.  
I have tested with `enable_name` enable/disable from client and it works correctly.
I also tested if user provide or not the name, at it also works correctly.
Do you maybe mean another case?

gsambrotta:
I have read the Contribution Guidelines and my understandig is that i have to squash smaller commit and keep just the most important ones:
```
Before submitting a pull-request, clean up the history, go over your commits and squash together minor changes and fixes into the corresponding commits.
```
Isn't it? Is allow just one commit per PR?

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
fantasticfears: You are right. Sorry, I didn't pay attention to that.

fantasticfears:
There is some flexibility here. But your two commits won't work without each other which makes sense to squash them together. And It's OK to have only one commit. As of your PR, it would be easier to find one commit instead of two when refers to git.

app/assets/javascripts/discourse/widgets/search-menu-results.js.es6
gsambrotta: no problem :)

gsambrotta:
@fantasticfears commites squash and consol log deleted.
Can you please review again?

tgxworld:
Thanks @gsambrotta  :+1:

PR #4710, opened 2017-02-15
FIX: admin dashboard shouldn't complain when using IAM profile for S3 access

ento:
Fixes https://meta.discourse.org/t/setting-up-file-and-image-uploads-to-s3/7229/62?u=ento

Previous code wasn't working as intended because it was parsed as

```ruby
    (bad_keys = (access_key or secret_key)) and !use_iam_profile
```

because of Ruby's operator precedence: `=` binds more eagerly than `and`.

http://ruby-doc.org/core-2.3.1/doc/syntax/precedence_rdoc.html


ZogStriP:
LGTM 👍  And it has tests! 😍 

Leaving it to @nlalonde because he wrote most of that file.

nlalonde:
Awesome, thank you!

PR #4715, opened 2017-02-21
FEATURE: new site setting for max logins per ip per hour/minute

techAPJ:
Meta topic: https://meta.discourse.org/t/admin-configuration-for-rate-limit/57686

@SamSaffron can you please review this PR?

### DIFF
--- a/spec/controllers/session_controller_spec.rb
+++ b/spec/controllers/session_controller_spec.rb
@@ -659,6 +659,23 @@ describe SessionController do
         end
       end
     end
+
+    context 'rate limited' do
+      it 'rate limits login' do
+        SiteSetting.max_logins_per_ip_per_hour = 2
+        RateLimiter.stubs(:disabled?).returns(false)
+        RateLimiter.clear_all!
+
+        2.times do
+          xhr :post, :create, login: user.username, password: 'myawesomepassword'
+          expect(response).to be_success
+        end
+        xhr :post, :create, login: user.username, password: 'myawesomepassword'
+        expect(response).not_to be_success
+        json = JSON.parse(response.body)
+        expect(json["error_type"]).to eq("rate_limit")
+      end
+    end
   end
 
   describe '.destroy' do
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -158,8 +158,8 @@ class SessionController < ApplicationController
       return
     end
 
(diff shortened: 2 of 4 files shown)


techAPJ:
@SamSaffron ping.

spec/controllers/session_controller_spec.rb
```diff
@@ -659,6 +659,22 @@ def post_login
         end
       end
     end
+
+    context 'rate limited' do
+      it 'rate limits login' do
+        SiteSetting.max_logins_per_ip_per_hour = 2
+        RateLimiter.stubs(:disabled?).returns(false)
```
tgxworld: I think it is better that we avoid stubbing here. Maybe we can add a method into `RateLimiter` that will allow us to enable the rate limit when we want to?

spec/controllers/session_controller_spec.rb
```diff
@@ -659,6 +659,22 @@ def post_login
...
+        RateLimiter.clear_all!
+
+        ActionDispatch::Request.any_instance.stubs(:remote_ip).returns('10.0.0.1')
```
tgxworld: Does the `remote_ip` change in the test env? It'll be good if we can avoid stubbing here.

spec/controllers/session_controller_spec.rb
```diff
@@ -659,6 +659,22 @@ def post_login
...
+        3.times do
+          xhr :post, :create, login: user.username, password: 'myawesomepassword'
+        end
+        expect(response).not_to be_success
```
tgxworld: I think we need to be checking the response each time we make a `post` request.

tgxworld:
@techAPJ  Looks good to me :+1: One small note, controller specs are going to be deprecated in Rails 5 so it is generally better to write new tests as integration specs.

spec/controllers/session_controller_spec.rb
techAPJ: Okay, it's been used in many other places. I'll look into it and push a new commit for it.

spec/controllers/session_controller_spec.rb
techAPJ: Seems like it does not. Removed the stubbing.

spec/controllers/session_controller_spec.rb
techAPJ: Okay, done.

techAPJ:
Thanks for the review, @tgxworld! ❤️

PR #4716, opened 2017-02-22
FEATURE: Allow checking the raw response of a bounced email

xfalcox:


app/assets/javascripts/admin/controllers/modals/admin-incoming-email.js.es6
```diff
@@ -12,6 +12,10 @@ export default Ember.Controller.extend(ModalFunctionality, {
 
   load(id) {
     return IncomingEmail.find(id).then(result => this.set("model", result));
+  },
+
+  load_from_bounced(id) {
```
ZogStriP: In JavaScript, we use camelCame (ie. `loadFromBounced`)

xfalcox:
Fixed the Camel/snake problems @ZogStriP :+1:

app/controllers/admin/email_controller.rb
```diff
@@ -88,6 +88,18 @@ def incoming
     render_json_dump(serializer)
   end
 
+  def incoming_from_bounced
+    params.require(:id)
+
+    bounced = EmailLog.find(params[:id].to_i)
```
tgxworld: This is going to raise an error if the ID is invalid. We probably want to use `find_by` and raise `Discourse::InvalidParameters` instead so that the error message is clearer.

app/controllers/admin/email_controller.rb
```diff
@@ -88,6 +88,18 @@ def incoming
...
+    email_local_part, email_domain = SiteSetting.notification_email.split('@')
+    bounced_to_address = "#{email_local_part}+verp-#{bounced.bounce_key}@#{email_domain}"
+
+    incoming_email = IncomingEmail.find_by(to_addresses: bounced_to_address)
```
tgxworld: We need to consider the case where `incoming_email` returns `nil`

xfalcox:
Invalid id for bounced and missing incoming are handled now, thanks @tgxworld

PR #4717, opened 2017-02-22
FEATURE: Autocomplete support on search

cpradio:
- Improve advanced search parameter matching using ^ and $ where applicable
- Added support for tag hash syntax in advanced search
- Enable autocomplete of users, categories and tags in Advanced Search and the Search Menu

### DIFF
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -50,8 +50,8 @@ export default Em.Component.extend({
 
   init() {
     this._super();
+    this._init();
     Ember.run.scheduleOnce('afterRender', () => {
-      this._init();
       this._update();
     });
   },
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -12,7 +12,7 @@ const REGEXP_IN_PREFIX             = /^in:/ig;
 const REGEXP_STATUS_PREFIX         = /^status:/ig;
 const REGEXP_MIN_POST_COUNT_PREFIX = /^min_post_count:/ig;
 const REGEXP_POST_TIME_PREFIX      = /^(before|after):/ig;
-const REGEXP_TAGS_SUFFIX           = /::tag\s?$/ig;
+const REGEXP_TAGS_REPLACE          = /(^(tags?:|#(?=[a-z0-9\-]+::tag))|::tag\s?$)/ig;
 
 
 const REGEXP_IN_MATCH                 = /^in:(posted|watching|tracking|bookmarks|first|pinned|unpinned)/ig;
@@ -230,7 +230,7 @@ export default Em.Component.extend({
 
     if (match.length !== 0) {
       const existingInput = _.isArray(tags) ? tags.join(',') : tags;
-      const userInput = match[0].replace(REGEXP_TAGS_PREFIX, '').replace(REGEXP_TAGS_SUFFIX, '');
+      const userInput = match[0].replace(REGEXP_TAGS_REPLACE, '');
 
       if (existingInput !== userInput) {
         this.set('searchedTerms.tags', (userInput.length !== 0) ? userInput.split(',') : []);
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -3,24 +3,26 @@ import { escapeExpression } from 'discourse/lib/utilities';
 
 const REGEXP_BLOCKS                = /(([^" \t\n\x0B\f\r]+)?(("[^"]+")?))/g;
 
-const REGEXP_USERNAME_PREFIX       = /(user:|@)/ig;
-const REGEXP_CATEGORY_PREFIX       = /(category:|#)/ig;
-const REGEXP_GROUP_PREFIX          = /group:/ig;
-const REGEXP_BADGE_PREFIX          = /badge:/ig;
-const REGEXP_TAGS_PREFIX           = /tags?:/ig;
-const REGEXP_IN_PREFIX             = /in:/ig;
-const REGEXP_STATUS_PREFIX         = /status:/ig;
-const REGEXP_MIN_POST_COUNT_PREFIX = /min_post_count:/ig;
-const REGEXP_POST_TIME_PREFIX      = /(before|after):/ig;
+const REGEXP_USERNAME_PREFIX       = /^(user:|@)/ig;
+const REGEXP_CATEGORY_PREFIX       = /^(category:|#)/ig;
+const REGEXP_GROUP_PREFIX          = /^group:/ig;
+const REGEXP_BADGE_PREFIX          = /^badge:/ig;
+const REGEXP_TAGS_PREFIX           = /^(tags?:|#(?=[a-z0-9\-]+::tag))/ig;
+const REGEXP_IN_PREFIX             = /^in:/ig;
+const REGEXP_STATUS_PREFIX         = /^status:/ig;
+const REGEXP_MIN_POST_COUNT_PREFIX = /^min_post_count:/ig;
+const REGEXP_POST_TIME_PREFIX      = /^(before|after):/ig;
+const REGEXP_TAGS_SUFFIX           = /::tag\s?$/ig;
 
-const REGEXP_IN_MATCH                 = /in:(posted|watching|tracking|bookmarks|first|pinned|unpinned)/ig;
-const REGEXP_SPECIAL_IN_LIKES_MATCH   = /in:likes/ig;
-const REGEXP_SPECIAL_IN_PRIVATE_MATCH = /in:private/ig;
-const REGEXP_SPECIAL_IN_WIKI_MATCH    = /in:wiki/ig;
 
-const REGEXP_CATEGORY_SLUG            = /(\#[a-zA-Z0-9\-:]+)/ig;
-const REGEXP_CATEGORY_ID              = /(category:[0-9]+)/ig;
-const REGEXP_POST_TIME_WHEN           = /(before|after)/ig;
+const REGEXP_IN_MATCH                 = /^in:(posted|watching|tracking|bookmarks|first|pinned|unpinned)/ig;
+const REGEXP_SPECIAL_IN_LIKES_MATCH   = /^in:likes/ig;
+const REGEXP_SPECIAL_IN_PRIVATE_MATCH = /^in:private/ig;
+const REGEXP_SPECIAL_IN_WIKI_MATCH    = /^in:wiki/ig;
+
+const REGEXP_CATEGORY_SLUG            = /^(\#[a-zA-Z0-9\-:]+)/ig;
+const REGEXP_CATEGORY_ID              = /^(category:[0-9]+)/ig;
+const REGEXP_POST_TIME_WHEN           = /^(before|after)/ig;
 
 export default Em.Component.extend({
   classNames: ['search-advanced-options'],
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -228,7 +230,7 @@ export default Em.Component.extend({
 
     if (match.length !== 0) {
       const existingInput = _.isArray(tags) ? tags.join(',') : tags;
-      const userInput = match[0].replace(REGEXP_TAGS_PREFIX, '');
+      const userInput = match[0].replace(REGEXP_TAGS_PREFIX, '').replace(REGEXP_TAGS_SUFFIX, '');
```
tgxworld: Hmm can this be combined into a single replace using `/(REGEXP_TAGS_PREFIX|REGEXP_TAGS_SUFFIX)/`?

app/assets/javascripts/discourse/components/search-text-field.js.es6
```diff
@@ -15,5 +21,46 @@ export default TextField.extend({
     // at the top of the page
     $(window).scrollTop(0);
     this.$().focus();
+  },
+
+  @on("didInsertElement")
+  applyAutoComplete() {
+    this._super();
```
tgxworld: I don't think we need this here :thought_balloon:

tgxworld:
Hi @cpradio, thank you for your work on this :+1: There are some problems I noted while reviewing this PR: 

1. If I autocomplete a category, it doesn't update the advanced search category fields
2. Closing and opening the advanced search panel removes the current category/user that I have in my search term.

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -340,6 +350,36 @@ export default createWidget('header', {
...
+
+  applyAutocomplete($searchInput) {
+    const siteSettings = this.siteSettings;
+
+    $searchInput.autocomplete({
+      template: findRawTemplate('category-tag-autocomplete'),
+      key: '#',
+      width: '100%',
+      treatAsTextarea: true,
+      transformComplete(obj) {
```
tgxworld: I think we can try to extract the logic here into a function so that we don't have to repeat the same logic twice.

cpradio:
> If I autocomplete a category, it doesn't update the advanced search category fields

It does that on Meta today... I'm not 100% sure why, but my changes didn't invoke this.

> Closing and opening the advanced search panel removes the current category/user that I have in my search term.

This also happens on Meta today.

> Some minor CSS changes needed here as well. The input field is obstructed.

Hmm, I could make it appear below, but when I do that, it won't line up with the character input location. At least with the way Autocomplete already handles placement, otherwise, there may be regressions in other places autocomplete gets used if I tell it to drop down Ypx...

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: Not really... as that isn't how REGEXP_TAGS_PREFIX gets used and to set it up that way would mean special cases for it in all locations used. Plus, it is attempting to accomplish two separate goals targetting two separate parts of the string.

To make it work that way, I'd have to change REGEXP_TAGS_PREFIX and REGEXP_TAGS_SUFFIX to always be part of a concatenation so it would put the `/(` and `|` and `)/ig` in it when applicable. Versus calling replace chained.

The only alternative I could see is creating another const that contains both expressions for replace. I could look into that.

app/assets/javascripts/discourse/widgets/header.js.es6
cpradio: Sounds like a good idea. I'll get on that today.

app/assets/javascripts/discourse/components/search-text-field.js.es6
cpradio: Sounds like a good idea. I'll get on that today.

app/assets/javascripts/discourse/widgets/header.js.es6
cpradio: This is done.

app/assets/javascripts/discourse/components/search-text-field.js.es6
cpradio: This is done.

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
cpradio: I implemented a REGEXP_TAGS_REPLACE that handles matching the prefix and suffix in one shot.

cpradio:
CSS changes are now also complete :+1:

cpradio:
By the way, I do want to dig into both
> If I autocomplete a category, it doesn't update the advanced search category fields
> Closing and opening the advanced search panel removes the current category/user that I have in my search term.

But I see them as a separate issue and would look to submit a separate PR to fix those. I have a feeling it has to do with some of the optimizations that were done a while back and how the `searchedTerms` object gets initialized.

cpradio:
@tgxworld, I was looking into
> Closing and opening the advanced search panel removes the current category/user that I have in my search term.

It seems that was introduced by this change
https://github.com/discourse/discourse/commit/e03d5e2140a0fcdefe7a5c7860c5bd95d4db4bd0

When I move _init() back to outside of the "afterRender" event, the behavior fixes itself.

The Category "prefilling on key-entry"  is still not working though.

tgxworld:
@cpradio Another small bug. In the header search menu, if I hit enter when selecting a category from the  autocomplete, I end up going to the full page search but the category isn't selected

cpradio:
Oh goodness, that was a doozie. Took me a while to figure out a solution to that one, but I like it (the solution that is).

tgxworld:
Cool 👍  Thank you so much for your work on this :) Can you squash your commits and I'll merge tomorrow.

cpradio:
Yup and hopefully Travis is behaving by tomorrow too! :+1:

cpradio:
And squashed. Thanks for the review.

tgxworld:
> Oh goodness, that was a doozie. Took me a while to figure out a solution to that one, but I like it (the solution that is).

Oops it looks like it isn't working for user autocomplete

cpradio:
That is due to something else underlying. keyDown is not getting called consistently on the widget... I have no idea why. But I can easily see with adding some console.log, that keyUp on search-menu-controls is not being invoked.

Simply putting `console.log('keyDown');` and `console.log('keyUp');` outlines this very easily. Put those in the respective spots, type `testing @`, you will see keyDown and keyUp entries correctly, press `Enter`, `keyUp` fires, but `keyDown` does not.

In fact, it seems to be related to having to use the SHIFT key to invoke autocomplete. Once shift is pressed, it doesn't register another keyDown even unless, 1) you type another character, 2) press SHIFT again.

I'm not sure what (if any) sort of fix can be made for this. But it doesn't seem relevant to the feature at hand.

cpradio:
I've tried countless ways to hack around this issue and all of them still leave it open to go to full page search before the selection is completed. I can check in the hack I have that makes it harder to do at the very beginning of auto-complete, but if you backspace and hit enter to "select" the item again, it will just redirect to full page search. Is that acceptable?

I can't seem to find a foolproof way of handling this.

tgxworld:
@cpradio I spent some time looking into this too and I get what you mean. It seems like username autocomplete closes when it updates the search results unlike category autocomplete which keeps it open. Hmm I'll need to spend more time on this

cpradio:
Okay, but that still doesn't fix the issue with the keyDown not firing after using SHIFT until a "normal" character is entered or SHIFT is used again. My last "hack" handles that scenario decently (unless you are a fast typer).

I have a feeling that keyDown issue goes all the way down to Ember event handling, as I think that is what is feeding it into the widgets (if I traced it back correctly).

If you figure something out, let me know, I've run out of ideas, as both Categories and Usernames are implementing Autocomplete the same way. I can reproduce the issues for both if I type fast enough of if I backspace into autocomplete (and hit enter fast enough again).

Also, just an FYI, this is something @coding-horror hired me to do, so if you figure out something and can point me in a direction, I'd have no problems fixing it, but these edge cases have me perplexed as it seems to be the underlying framework causing them, not Discourse, or my logic.

cpradio:
Thanks @tgxworld, looks good on try.discourse.org! :+1:

tgxworld:
@cpradio Hmm category autocomplete is broken in the search-menu. Can you have a look?

cpradio:
in the search-menu? Can you be more specific?

tgxworld:
hehe I was referring to the `search-menu` widget.

screenshot from 2017-03-07 12-10-16

cpradio:
Ah, yeah, that seems to be related to some error in search-menu-results, possibly. It only happens if `#` is the very first character you type, which I find bizarre. Something to do with it not displaying the "your term is too short" like it does for `@`

cpradio:
@tgxworld So far, I've discovered that it is `widget.scheduleRerender()` in search-menu.js.es6 that is causing the display issue. When that gets invoked, it causes the following JavaScript error
```
Uncaught DOMException: Failed to execute 'appendChild' on 'Node': This node type does not support this method.
    at insertNode (virtual-dom:543:20)
    at applyPatch (virtual-dom:506:20)
    at applyPatch (virtual-dom:693:23)
    at Object.patchRecursive [as patch] (virtual-dom:675:20)
    at applyPatch (virtual-dom:521:31)
    at applyPatch (virtual-dom:700:19)
    at Object.patchRecursive [as patch] (virtual-dom:675:20)
    at applyPatch (virtual-dom:521:31)
    at applyPatch (virtual-dom:700:19)
    at Object.patchRecursive [as patch] (virtual-dom:675:20)
```

I'm unsure as to why that is, but it has something to do with the rerendering of the header parent widget. And what when that occurs, it screws up the output of the autocomplete result, removing its parent div that has the class `autocomplete`, thus the display issue.

eviltrout:
One thing to be careful about is that if you asynchronously schedule rendering (say after AJAX or Ember.run), the element might not exist when your renderer gets triggered. We usually check to see if the element is still there, but it's possible something in your code is not doing this.

cpradio:
Well, it would be something in autocomplete. I didn't create any new UI, I just implemented what exists already. What is bizarre, is it only happens if the `#` is the very first character you type, otherwise, if you type `a #`, it works fine. I'll dig into the autocomplete lib again, but I couldn't see anywhere that would make it remove its parent div. It simply loads the template, and the template has the parent div in it...

cpradio:
Okay, so I set a breakpoint on where autocomplete.js.es6 loads the template, and that only gets called once. When scheduleRerender occurs, it never calls that part again, so something else is manipulating the DOM in there in a way that isn't expected and stripping the autocomplete parent div.

I just haven't a clue what, where, or how, other than the what is the rerendering of said widgets.

Is it due to where the autocomplete loading is placed?
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/search-text-field.js.es6#L15

cpradio:
Okay, we've all been looking at this the wrong way. We all assumed it was due to Ajax loading the results, in fact, it is the opposite. This issue only exists when Ajax DOESN'T load the results.

Category.search will attempt to load only categories, if it can, for the autocomplete result using `Discourse.Site.currentProp('sortedCategories')` at https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/category.js.es6#L212

It ends up if you have less than 5 categories, when you press `#` it will do an ajax call to get tags to fill in the remaining slots in autocomplete and then it loads perfectly fine. But if you have more than 5 categories, it doesn't need to perform an ajax call and that is when the issue occurs.

I still have no idea why that matters. Why the widget.scheduleRerender behaves differently when the data is loaded from an ajax request versus from a property. But it obviously does. I can only fathom it is due to the request being super quick and thus it is already rendered before scheduleRerender fires and when it does fire, it fails to load it back properly. Whereas, the ajax call occurs after scheduleRerender fires and therefore loads fine.

Either way, I thought this was an interesting observation.

PR #4718, opened 2017-02-23
FEATURE: further restrict downloading of backups

oblakeerickson:
- send email to logged in admin when they press the "download" button
- show pop-up that email was sent
- update tests
- create email template

### DIFF
--- a/app/controllers/admin/backups_controller.rb
+++ b/app/controllers/admin/backups_controller.rb
@@ -1,4 +1,5 @@
 require "backup_restore/backup_restore"
+require "email_backup_token"
 
 class Admin::BackupsController < Admin::AdminController
 
@@ -44,14 +45,33 @@ class Admin::BackupsController < Admin::AdminController
     render json: success_json
   end
 
-  # download
-  def show
+  def email
     if backup = Backup[params.fetch(:id)]
+      token = EmailBackupToken.set(current_user.id)
+      download_url = "#{url_for(controller: 'backups', action: 'show')}?token=#{token}"
+      Jobs.enqueue(:download_backup_email, to_address: current_user.email, backup_file_path: download_url)
+      render nothing: true
+    else
+      render nothing: true, status: 404
+    end
+  end
+
+  def show
+
+    if !EmailBackupToken.compare(current_user.id, params.fetch(:token))
+      @error = I18n.t('download_backup_mailer.no_token')
+    end
+    if !@error && backup = Backup[params.fetch(:id)]
+      EmailBackupToken.del(current_user.id)
       StaffActionLogger.new(current_user).log_backup_download(backup)
       headers['Content-Length'] = File.size(backup.path).to_s
       send_file backup.path
     else
-      render nothing: true, status: 404
+      if @error
+        render layout: 'no_ember', status: 422
+      else
+        render nothing: true, status: 404
+      end
     end
   end
 
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1762,6 +1762,14 @@ en:
 
       (If the link above has expired, choose "I forgot my password" when logging in with your email address.)
 
+  download_backup_mailer:
+    title: "Download Backup Mailer"
+    subject_template: "[%{site_name}] Backup Download"
+    text_body_template: |
+      Click this link to download your backup file: %{backup_file_path}
+    no_token: |
+      Sorry, this backup download link has already been used or has expired.
+
   test_mailer:
     title: "Test Mailer"
     subject_template: "[%{site_name}] Email Deliverability Test"
--- a/lib/email_backup_token.rb
+++ b/lib/email_backup_token.rb
@@ -1,0 +1,28 @@
+class EmailBackupToken
+
+  def self.key(user_id)
+    "email-backup-token:#{user_id}"
+  end
+
+  def self.generate
+    SecureRandom.hex
+  end
+
+  def self.set(user_id)
+    token = self.generate
+    $redis.setex self.key(user_id), 1.day.to_i, token
+    token
+  end
+
+  def self.get(user_id)
+    $redis.get self.key(user_id)
+  end
+
+  def self.del(user_id)
+    $redis.del self.key(user_id)
+  end
+
+  def self.compare(user_id, token)
+    token == self.get(user_id)
+  end
+end
--- a/spec/controllers/admin/backups_controller_spec.rb
+++ b/spec/controllers/admin/backups_controller_spec.rb
@@ -90,25 +90,69 @@ describe Admin::BackupsController do
     describe ".show" do
 
       it "uses send_file to transmit the backup" do
-        path = File.join(Backup.base_directory, backup_filename)
-        File.open(path, "w") { |f| f.write("hello") }
+        begin
+          token = EmailBackupToken.set(@admin.id)
+          path = File.join(Backup.base_directory, backup_filename)
+          File.open(path, "w") { |f| f.write("hello") }
 
-        Backup.create_from_filename(backup_filename)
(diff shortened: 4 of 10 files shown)


lib/email_backup_token.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  KEY = "email_backup_token"
+
+  def self.generate
+    SecureRandom.hex
+  end
+
+  def self.set
+    token = EmailBackupToken.generate
+    $redis.set KEY, token
```
tgxworld: I think the key needs to be uniq per admin user.

oblakeerickson:
Thanks for the feedback! :) I'll make those changes. Not sure if you noticed but I'm not hiding the filename anymore.

spec/controllers/admin/backups_controller_spec.rb
```diff
@@ -90,25 +90,43 @@
...
 
+      it "return 404 when token is bad" do
+        path = File.join(Backup.base_directory, backup_filename)
+        File.open(path, "w") { |f| f.write("hello") }
+
+        Backup.create_from_filename(backup_filename)
+
+        get :show, id: backup_filename, token: "bad_value"
+
+        File.delete(path) rescue nil
```
tgxworld: This needs to be run in an `ensure` block and not be rescued. Otherwise, it may silently fail and we end up not cleaning up after the test

app/controllers/admin/backups_controller.rb
```diff
@@ -44,9 +45,22 @@ def cancel
...
 
-  # download
-  def show
+  def email
     if backup = Backup[params.fetch(:id)]
+      token = EmailBackupToken.set
+      hostname = Discourse.current_hostname
+      base_url = Discourse.base_url
+      download_url = "#{Discourse.base_url}/admin/backups/#{backup.filename}?token=#{token}"
+      Jobs::DownloadBackupEmail.new.execute(to_address: current_user.email, backup_file_path: download_url)
```
tgxworld: This should probably be executed in a background job.

app/controllers/admin/backups_controller.rb
```diff
@@ -44,9 +45,22 @@ def cancel
...
   end
 
-  # download
-  def show
+  def email
     if backup = Backup[params.fetch(:id)]
+      token = EmailBackupToken.set
+      hostname = Discourse.current_hostname
+      base_url = Discourse.base_url
+      download_url = "#{Discourse.base_url}/admin/backups/#{backup.filename}?token=#{token}"
```
tgxworld: Are we able to generate the path from `admin_backups_path` instead of hardcoding the path here?

app/controllers/admin/backups_controller.rb
```diff
@@ -44,9 +45,22 @@ def cancel
...
+      Jobs::DownloadBackupEmail.new.execute(to_address: current_user.email, backup_file_path: download_url)
+      render nothing: true
+    else
+      render nothing: true, status: 404
+    end
+  end
+
+  def show
```
tgxworld: We should probably generate a different view if the token is invalid or not present. Currently we're just returning a 404 which may be confusing to users.

app/models/backup.rb
```diff
@@ -24,6 +24,27 @@ def self.[](filename)
     end
   end
 
+  def self.get_file_from_filename(filename)
```
tgxworld: Based on https://github.com/discourse/discourse/pull/4718#issuecomment-282888181, I think the following methods are not required anymore?

lib/backup_restore/backuper.rb
```diff
@@ -84,6 +84,7 @@ def initialize_state
       @dump_filename = File.join(@tmp_directory, BackupRestore::DUMP_FILE)
       @archive_directory = File.join(Rails.root, "public", "backups", @current_db)
       @archive_basename = File.join(@archive_directory, "#{SiteSetting.title.parameterize}-#{@timestamp}-#{BackupRestore::VERSION_PREFIX}#{BackupRestore.current_version}")
+      @shortname = "#{SiteSetting.title.parameterize}-#{@timestamp.split("-")[0..2].join("-")}"
```
tgxworld: I don't think this variable is being used anywhere.

lib/email_backup_token.rb
```diff
@@ -0,0 +1,29 @@
+class EmailBackupToken
+
+
```
tgxworld: Style nitpick: There are two empty blank lines here.

lib/email_backup_token.rb
```diff
@@ -0,0 +1,29 @@
...
+    "email-backup-token:#{user_id}"
+  end
+
+  def self.generate
+    SecureRandom.hex
+  end
+
+  def self.set(user_id)
+    token = EmailBackupToken.generate
+    $redis.set EmailBackupToken.key(user_id), token
```
tgxworld: It seems like we're not expiring the token here.

lib/email_backup_token.rb
```diff
@@ -0,0 +1,29 @@
...
+    token
+  end
+
+  def self.get(user_id)
+    $redis.get EmailBackupToken.key(user_id)
```
tgxworld: We can just call `self.key(user_id)`

spec/controllers/admin/backups_controller_spec.rb
```diff
@@ -90,25 +90,49 @@
...
+      end
+
+      it "return 404 when token is bad" do
+        begin
+          path = File.join(Backup.base_directory, backup_filename)
+          File.open(path, "w") { |f| f.write("hello") }
+
+          Backup.create_from_filename(backup_filename)
+
+
```
tgxworld: Style nitpick: Two blank lines here.

app/models/backup.rb
```diff
@@ -2,7 +2,7 @@ class Backup
   include ActiveModel::SerializerSupport
 
   attr_reader :filename
-  attr_accessor :size, :path, :link
+  attr_accessor :filename, :size, :path, :link, :name
```
tgxworld: I believe the changes here are not used anymore as well.

config/locales/server.en.yml
```diff
@@ -1759,6 +1759,14 @@ en:
 
       (If the link above has expired, choose "I forgot my password" when logging in with your email address.)
 
+  download_backup_mailer:
+    title: "Download Backup Mailer"
+    subject_template: "[%{site_name}] Backup Download"
+    text_body_template: |
+      Click this link to download your backup file: %{backup_file_path}
+    no_token: |
+      Sorry, that backup download link has already been used or has expired.
```
tgxworld: Hmm probably `this backup link has already been used or has expired?` because the error message is shown when the user has visited the link.

spec/controllers/admin/backups_controller_spec.rb
```diff
@@ -90,25 +90,49 @@
...
-        expect(response.headers['Content-Disposition']).to match(/attachment; filename/)
+          expect(response.headers['Content-Length']).to eq("5")
+          expect(response.headers['Content-Disposition']).to match(/attachment; filename/)
+        ensure
+          File.delete(path)
+          EmailBackupToken.del(@admin.id)
+        end
+      end
+
+      it "return 404 when token is bad" do
```
tgxworld: We need a test case when the route returns 422 as well.

app/controllers/admin/backups_controller.rb
```diff
@@ -44,14 +45,33 @@ def cancel
     render json: success_json
   end
 
-  # download
-  def show
+  def email
```
tgxworld: @oblakeerickson I just noticed that we didn't add a test case for this.

tgxworld:
@oblakeerickson Can you also squash your commits. Merge away once you're done :+1:

PR #4719, opened 2017-02-23
fix malformed qunit url

rimian:
this removes the space between the query string and the URL

```
$ MODULE='Acceptance: Search' bundle exec rake qunit:test\[20000\]
....
Running: {"module":"Acceptance: Search"}
... http://localhost:60099/qunit?module=Acceptance%3A%20Search 20000
```

The timeout value seems to work fine.

Note the back slashes are only needed for zsh shell. Otherwise run:

```
$ MODULE='Acceptance: Search' bundle exec rake qunit:test[20000]
```


tgxworld:
Thanks @rimian  :+1:

PR #4720, opened 2017-02-24
Fix a minor typo on a mailer template

rizka10:
"with" is supposed to be "wish" on 'unsubscribe_mailer.text_body_template'


tgxworld:
Thanks @rizka10 :+1:

PR #4721, opened 2017-02-25
FEATURE: Add order logic to admin users controller

oblakeerickson:
Added order and direction parameters for sorting admin user pages. This
commit only includes backend api changes.

https://meta.discourse.org/t/make-admin-users-list-sortable-suggestion/47649

Now you can pass in `order` and `asc` parameters to the
`/admin/users/list/<query>.json` endpoint.

Example:

`/admin/users/list/active.json?&order=post_count` which defaults to desc

and

`/admin/users/list/active.json?order=post_count&asc=true`

### DIFF
--- a/app/jobs/scheduled/pending_users_reminder.rb
+++ b/app/jobs/scheduled/pending_users_reminder.rb
@@ -7,7 +7,7 @@ module Jobs
 
     def execute(args)
       if SiteSetting.must_approve_users && SiteSetting.pending_users_reminder_delay >= 0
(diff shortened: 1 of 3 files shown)


SamSaffron:
looks ok to me, have you confirmed there is no N+1 lurking in the admin users page?

oblakeerickson:
I didn't notice any as I was working on it. It just uses the users and the user_stats table and does an include.

SamSaffron:
K ... lets try this out :)

PR #4724, opened 2017-02-26
FEATURE: Socialcast Importer: Set category and tags based on group

mcwumbly:
If desired, edit the `socialcast_message.rb` file to set the category
and tags for each topic based on the name of the Socialcast group it was
originally posted to.

You must create categories with the same names first in your site.

All topics will get the `DEFAULT_TAG` at minimum.

Topics posted to a group that matches any group name in the `TAGS_AND_CATEGORIES`
map will get the associated category and tags.

Other topics will be tagged with the original groupname and placed in the
`DEFAULT_CATEGORY`.


mcwumbly:
We recently used the importer with these changes to migrate users and content from a Socialcast instance to Discourse.

nlalonde:
:+1: Thanks

PR #4725, opened 2017-02-27
Remove references to elder

jomaxro:
The use of the TL4 name Elder was changed in 2014 to Leader.  The spec function was changed to `:trust_level_4`, but the it statement was not changed.


tgxworld:
Thanks @jomaxro

PR #4726, opened 2017-02-27
FEATURE: Admin user list sortable

gsambrotta:
Added sortable functionality to admin users list using api `order` and `ascending` in four different columns: `Seen` `Topics Viewed` `Posts Read` and `Read Time`

Related meta issue:
https://meta.discourse.org/t/make-admin-users-list-sortable-suggestion/47649

### DIFF
--- a/app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
+++ b/app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
@@ -4,13 +4,7 @@ import { bufferedRender } from 'discourse-common/lib/buffered-render';
 export default Ember.Component.extend(bufferedRender({
   tagName: 'th',
   classNames: ['sortable'],
-  attributeBindings: ['title'],
-  rerenderTriggers: ['order', 'asc'],
-
-  title: function() {
-    const labelKey = this.get('field');
-    return I18n.t(labelKey + '_long', { defaultValue: I18n.t(labelKey) });
-  }.property('field'),
+  rerenderTriggers: ['order', 'ascending'],
 
   buildBuffer(buffer) {
     const icon = this.get('icon');
@@ -18,23 +12,21 @@ export default Ember.Component.extend(bufferedRender({
       buffer.push(iconHTML(icon));
     }
 
-    const field = this.get('field');
-    buffer.push(I18n.t(field));
+    buffer.push(I18n.t(this.get('i18nKey')));
 
-    if (field === this.get('order')) {
-      buffer.push(iconHTML(this.get('asc') ? 'chevron-up' : 'chevron-down'));
+    if (this.get('field') === this.get('order')) {
+      buffer.push(iconHTML(this.get('ascending') ? 'chevron-up' : 'chevron-down'));
     }
   },
 
   click() {
-    const currentOrder = this.get('order'),
-          field = this.get('field');
-    console.log(this);
+    const currentOrder = this.get('order');
+    const field = this.get('field');
 
     if (currentOrder === field) {
-      this.set('asc', this.get('asc') ? null : true);
+      this.set('ascending', this.get('ascending') ? null : true);
     } else {
-      this.setProperties({ order: field, asc: null });
+      this.setProperties({ order: field, ascending: null });
     }
   }
 }));
--- a/app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
+++ b/app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
@@ -1,0 +1,40 @@
+import { iconHTML } from 'discourse-common/helpers/fa-icon';
+import { bufferedRender } from 'discourse-common/lib/buffered-render';
+
+export default Ember.Component.extend(bufferedRender({
+  tagName: 'th',
+  classNames: ['sortable'],
+  attributeBindings: ['title'],
+  rerenderTriggers: ['order', 'asc'],
+
+  title: function() {
+    const labelKey = this.get('field');
+    return I18n.t(labelKey + '_long', { defaultValue: I18n.t(labelKey) });
+  }.property('field'),
+
+  buildBuffer(buffer) {
+    const icon = this.get('icon');
+    if (icon) {
+      buffer.push(iconHTML(icon));
+    }
+
+    const field = this.get('field');
+    buffer.push(I18n.t(field));
+
+    if (field === this.get('order')) {
+      buffer.push(iconHTML(this.get('asc') ? 'chevron-up' : 'chevron-down'));
+    }
+  },
+
+  click() {
+    const currentOrder = this.get('order'),
+          field = this.get('field');
+    console.log(this);
+
+    if (currentOrder === field) {
+      this.set('asc', this.get('asc') ? null : true);
+    } else {
+      this.setProperties({ order: field, asc: null });
+    }
+  }
+}));
--- a/app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
@@ -7,7 +7,7 @@ import { observes } from 'ember-addons/ember-computed-decorators';
 export default Ember.Controller.extend({
   query: null,
   queryParams: ['order', 'ascending'],
-  order: '',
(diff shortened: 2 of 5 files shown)


gsambrotta:
This PR cannot be merge!
Need to wait for [PR 4721](https://github.com/discourse/discourse/pull/4721)

gsambrotta:
This PR is ready to be review and then be merged

app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
```diff
@@ -0,0 +1,33 @@
...
+    const field = this.get('field');
+
+    if (field === this.get('order')) {
+      buffer.push(iconHTML(this.get('ascending') ? 'chevron-up' : 'chevron-down'));
+    }
+  },
+
+  click() {
+    const currentOrder = this.get('order'),
+          field = this.get('field');
```
tgxworld: This isn't specified anywhere but we prefer 

```
const current_order = this.get('order');
const field = this.get('field');
```

app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
```diff
@@ -43,7 +48,19 @@ export default Ember.Controller.extend({
     var self = this;
     this.set('refreshing', true);
 
-    AdminUser.findAll(this.get('query'), { filter: this.get('listFilter'), show_emails: this.get('showEmails') }).then(function (result) {
+    AdminUser.findAll({ filter: this.get('listFilter'), show_emails: this.get('showEmails') }).then(function (result) {
```
tgxworld: We prefer using [arrow functions](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions) now :)

app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
```diff
@@ -43,7 +48,19 @@ export default Ember.Controller.extend({
...
+      self.set('model', result);
```
tgxworld: You'll be able to use `this` instead of `self` for arrow functions because it doesn't bind its own `this`

app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
```diff
@@ -43,7 +48,19 @@ export default Ember.Controller.extend({
...
+      self.set('refreshing', false);
+    });
+  },
+
+  @observes('order', 'ascending')
+  sortUsers: function() {
+    var self = this;
+    this.set('refreshing', true);
+
+    AdminUser.findAll(this.get('query'), {order: this.get('order'), ascending: this.get('ascending')}).then(function (result) {
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/4726/files#r103611174

app/assets/javascripts/discourse/templates/group-index.hbs
```diff
@@ -20,7 +20,9 @@
               {{/user-info}}
             </td>
             <td>
-              <span class="text">{{bound-date m.last_posted_at}}</span>
+              <span class="text">
+                {{bound-date m.last_posted_at}}
```
tgxworld: I would consider this a cosmetic change which we should avoid.

app/assets/javascripts/admin/templates/users-list-show.hbs
```diff
@@ -22,62 +22,87 @@
...
-          <td>
-            {{#if user.admin}}<i class="fa fa-shield" title="{{i18n 'admin.title'}}"></i>{{/if}}
-            {{#if user.moderator}}<i class="fa fa-shield" title="{{i18n 'admin.moderator'}}"></i>{{/if}}
-          </td>
-        </tr>
-      {{/each}}
-
+            <td>
+              {{#if user.admin}}
+                <i class="fa fa-shield" title="{{i18n 'admin.title'}}"></i>
```
tgxworld: This is probably old code but we can use the `{{fa-icon}}` component now

app/assets/javascripts/admin/templates/users-list-show.hbs
```diff
@@ -22,62 +22,87 @@
...
+              {{/if}}
+              {{#if user.moderator}}
+                <i class="fa fa-shield" title="{{i18n 'admin.moderator'}}"></i>
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/4726/files#r103611435

app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
```diff
@@ -0,0 +1,33 @@
...
+  rerenderTriggers: ['order', 'ascending'],
+
+  buildBuffer(buffer) {
+    const icon = this.get('icon');
+    if (icon) {
+      buffer.push(iconHTML(icon));
+    }
+
+    buffer.push(I18n.t(this.get('i18nKey')));
+    const field = this.get('field');
```
gsambrotta: Before i used two times but then i rewrite the function so agree, we dont need to set a var now.

app/assets/javascripts/admin/components/admin-directory-toggle.js.es6
gsambrotta: ok

app/assets/javascripts/admin/templates/users-list-show.hbs
gsambrotta: @tgxworld yep is old code and it look like i cannot use `{{fa-icon shield}}` here. If i do i got an error saying i cannot use hbs helper function in form elements

gsambrotta: @tgxworld any feedback about that?

mikechristopher: I may be wrong but looking through the code elsewhere it uses the following: {{fa-icon "shield"}} can you not use the same?

gsambrotta: @mikechristopher look like i cannot. The problem is not with `{{fa-icon "shield"}}` but with using handlerbars helpers in form elment. 
Did you see `{{fa-icon "shield"}}` used in form tag?

mikechristopher: not so much within a form tag but within another table as per the dashboard template - 

https://github.com/discourse/discourse/blob/269f6e8c3095f24e71191fb55a87bd379de0cf93/app/assets/javascripts/admin/templates/dashboard.hbs

gsambrotta:
@mikechristopher Thanks for the hint! 
I was actually using in the wrong way and the message error divert me

gsambrotta:
@tgxworld what do you think, is now ready to be merged?

tgxworld:
@gsambrotta  Looks like some extra padding has been added? It is making the page look weird

screenshot from 2017-03-14 12-25-59

tgxworld:
Also can we update the styles for `th` to follow the `/users` page? There is currently no indication that each header is clickable.

app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
```diff
@@ -40,13 +45,23 @@ export default Ember.Controller.extend({
...
-      self.set('refreshing', false);
+    AdminUser.findAll( { filter: this.get('listFilter'), show_emails: this.get('showEmails') }).then( (result) => {
+      this.set('model', result);
+    }).finally( () => {
+      this.set('refreshing', false);
+    });
+  },
+
+  @observes('order', 'ascending')
+  sortUsers: function() {
```
tgxworld: This function seems to be a duplication of `_refreshUsers` above. Is there a reason why can't pass the order into `_refreshUser` as well? :thinking:

gsambrotta: mh yeah you are right, i think i can just use `_refreshUser`

tgxworld:
Merged in https://github.com/discourse/discourse/commit/576982484d5ef64d750ee6eb958067dac070eacd.

Thanks @gsambrotta

PR #4728, opened 2017-02-27
Add support for username regex.

nbianca:
Hello,

I have added support for [Reserved username wildcard](https://meta.discourse.org/t/reserved-usernames-wildcard/41968/12).

First of all, I have extended the support to more than just a wildcard ~~, but to support any regex~~.

~~However, there is an issue with the username suggester. For example, if `i42-.*` is one of those restrictitions and I try to create a user called `i42-test`, the username suggester will try to suggest `i42-test1`, `i42-test2`, `i42-test3`, etc. and will not stop. I added another condition to stop suggesting after 10 attempts, however this is just a cheapfix. I would like to know how would the team want to fix this issue.~~

I look forward to hearing from you.

Thanks,
Bianca Nenciu

### DIFF
--- a/lib/user_name_suggester.rb
+++ b/lib/user_name_suggester.rb
@@ -21,10 +21,14 @@ module UserNameSuggester
     name = fix_username(name)
     i = 1
     attempt = name
-    until attempt == allow_username || User.username_available?(attempt)
+    until attempt == allow_username || User.username_available?(attempt) || i > 100
       suffix = i.to_s
       max_length = User.username_length.end - suffix.length - 1
       attempt = "#{name[0..max_length]}#{suffix}"
+      i += 1
+    end
+    until attempt == allow_username || User.username_available?(attempt) || i > 200
+      attempt = SecureRandom.hex[0..SiteSetting.max_username_length]
       i += 1
     end
     attempt
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -475,6 +475,22 @@ describe User do
       expect(User.username_available?('DonKey')).to eq(false)
       expect(User.username_available?('test')).to eq(false)
     end
+
+    it 'should not allow usernames matched against an expession' do
+      SiteSetting.reserved_usernames = 'test)|*admin*|foo*|*bar|abc.def'
+
+      expect(User.username_available?('test')).to eq(true)
+      expect(User.username_available?('abc9def')).to eq(true)
+
+      expect(User.username_available?('admin')).to eq(false)
+      expect(User.username_available?('foo')).to eq(false)
+      expect(User.username_available?('bar')).to eq(false)
+
+      expect(User.username_available?('admi')).to eq(true)
+      expect(User.username_available?('bar.foo')).to eq(true)
+      expect(User.username_available?('foo.bar')).to eq(false)
+      expect(User.username_available?('baz.bar')).to eq(false)
+    end
   end
 
   describe 'email_validator' do
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -162,7 +162,7 @@ class User < ActiveRecord::Base
     lower = username.downcase
 
     User.where(username_lower: lower).blank? &&
-      !SiteSetting.reserved_usernames.split("|").any? { |reserved| reserved.casecmp(username) == 0 }
+      SiteSetting.reserved_usernames.split("|").all? { |reserved| !lower.match('^' + Regexp.escape(reserved).gsub('\*', '.*') + '$') }
   end
 
   def self.plugin_staff_user_custom_fields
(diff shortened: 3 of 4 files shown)


SamSaffron:
This changes semantics though

for example `sam.saffron` will now ban `sam9saffron` as well. 

Additionally, 

- We need some tests here to confirm behavior
- The if careful someone can add a site setting that will cause the code to explode, eg: `sam)`

SamSaffron:
I think the best way forward is simply to expand semantics here without full blown regex. 

Just add support for `*` explosion. So `*admin*` would be used to ban everyone with the word admin. This mean it will be a lot harder to get this wrong.

SamSaffron:
>  the username suggester will try to suggest i42-test1, i42-test2, i42-test3, etc. and will not stop.  

Yeah this is a security issue, it must be also addressed with the PR, if a * test is the one that fails then use a generic username maybe something like 

- first try usernameRANDOMNUMBER +1 +2 etc ... up to 100 times 
- then just give user random letters and try up to 100 times

at 200 bail out 

otherwise people will be able to crash any discourse instance with this * expansion enabled

nbianca:
Hello Sam,

I made the changes per your review:
* Regex is no longer supported. The only "special character" is the wildcard symbol. I escape any symbol that might be part of a regex, so `foo.bar` will not restrict `foo9bar`.
* I added tests for username availability.
* I also added generation of random usernames consisting of random characters after 100 fails. (I am generating a new username of the same length as the initial one.) Do you think I should generate a random username only if a wildcard restriction is applied? Also, I was thinking of creating random usernames like `adjective.noun.random_number` but I am not sure how this would be affected by localization.

SamSaffron:
```
attempt = rand(36 ** name.length).to_s(36)
```

looks somewhat odd to me... 

Perhaps simply:

```
attempt = SecureRandom.hex
```

SamSaffron:
`[0..name.length]` needs better bounds, cause if you pass a name that is 1000 chars long that will simply fail 1000 times. I would keep in simple and set length fixed to SiteSetting.max_username_length

SamSaffron:
I think all we have left is to use SiteSetting.max_username_length for username fallback length, except for that we should be all good

nbianca:
Sorry, I have been busy with university. I just pushed my changes. Tests should pass, so this should be ready to merge indeed.

SamSaffron:
@nbianca no worries at all an no rush! thanks so much for contributing to Discourse.

lib/user_name_suggester.rb
```diff
@@ -21,12 +21,16 @@ def self.find_available_username_based_on(name, allow_username = nil)
...
     i = 1
     attempt = name
-    until attempt == allow_username || User.username_available?(attempt)
+    until attempt == allow_username || User.username_available?(attempt) || i > 100
       suffix = i.to_s
       max_length = User.username_length.end - suffix.length - 1
       attempt = "#{name[0..max_length]}#{suffix}"
       i += 1
     end
+    until attempt == allow_username || User.username_available?(attempt) || i > 200
```
tgxworld: Based on the discussion in 

https://github.com/discourse/discourse/pull/4728#issuecomment-282862287. I think this should be `i > 100`?

spec/models/user_spec.rb
```diff
@@ -475,6 +475,22 @@ def assert_good(username)
       expect(User.username_available?('DonKey')).to eq(false)
       expect(User.username_available?('test')).to eq(false)
     end
+
+    it 'tests when a username is reserved by an expression with wildcard' do
```
tgxworld: The description here doesn't sound right to me. Usually the `it` block describes what the expected result should be. In this case, I would go with something like:

```
context 'when reserved usernames contains an expression" do
  it "should return the right value when username is matched against the expresssion" do
  end
end
```

spec/models/user_spec.rb
```diff
@@ -475,6 +475,22 @@ def assert_good(username)
...
+      SiteSetting.reserved_usernames = 'test)|*admin*|foo*|*bar|abc.def'
```
tgxworld: We'll need to update the description for the site settings to clearly indicate that Regexp expressions are allowed.

lib/user_name_suggester.rb
nbianca: The counter `i` is not reset before this so it is going to be 100 from the previous block and another 100 from this block.

spec/models/user_spec.rb
nbianca: You are right. I changed the description.

spec/models/user_spec.rb
nbianca: This is also correct. I have updated that [here](https://github.com/discourse/discourse/pull/4728/files#diff-aeb8651199334c1cb831e884ff85773aR1043).

tgxworld:
Thanks @nbianca

PR #4731, opened 2017-03-01
Files for Greek locale

vasilisvla:
I have pushed the translated files for the Greek locale from Transifex here: https://github.com/vasilisvla/discourse-el_GR


nlalonde:
Greek was added in this commit https://github.com/discourse/discourse/commit/d8489163571a1d24a7664997a23986f93e22750c

PR #4735, opened 2017-03-02
FIX: start sidekiq process for all queues in docker dev

LeoMcA:
https://meta.discourse.org/t/docker-dev-only-starting-critical-sidekiq-queue/58242


SamSaffron:
this is confusing you are removing all the queues? how does this even work

SamSaffron:
this looks good to me!

PR #4736, opened 2017-03-02
FIX: grant trust level when bulk adding users to group

techAPJ:
Bug report: https://meta.discourse.org/t/trust-level-change-with-bulk-add-to-group/58217?u=techapj

### DIFF
--- a/app/jobs/regular/bulk_grant_trust_level.rb
+++ b/app/jobs/regular/bulk_grant_trust_level.rb
@@ -1,0 +1,17 @@
+module Jobs
+
+  class BulkGrantTrustLevel < Jobs::Base
+
+    def execute(args)
+      trust_level = args[:trust_level]
+      user_ids = args[:user_ids]
+
+      raise Discourse::InvalidParameters.new(:trust_level) if trust_level.blank?
+      raise Discourse::InvalidParameters.new(:user_ids) if user_ids.blank?
+
+      User.where(id: user_ids).find_each do |user|
+        TrustLevelGranter.grant(trust_level, user)
+      end
+    end
+  end
+end
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -414,6 +414,10 @@ class Group < ActiveRecord::Base
       if self.title.present?
         User.where(id: user_ids).update_all(title: self.title)
       end
+
+      if self.grant_trust_level.present?
+        Jobs.enqueue(:bulk_grant_trust_level, user_ids: user_ids, trust_level: self.grant_trust_level)
+      end
     end
     true
   end
--- a/app/models/group_user.rb
+++ b/app/models/group_user.rb
@@ -55,11 +55,7 @@ class GroupUser < ActiveRecord::Base
 
   def grant_trust_level
     return if group.grant_trust_level.nil?
-    if user.trust_level < group.grant_trust_level
-      user.change_trust_level!(group.grant_trust_level)
-      user.trust_level_locked = true
-      user.save
-    end
+    TrustLevelGranter.grant(group.grant_trust_level, user)
   end
 end
 
--- a/app/services/trust_level_granter.rb
+++ b/app/services/trust_level_granter.rb
@@ -1,0 +1,18 @@
+class TrustLevelGranter
+
+  def initialize(trust_level, user)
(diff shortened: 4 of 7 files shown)


app/models/group.rb
```diff
@@ -414,6 +414,16 @@ def bulk_add(user_ids)
       if self.title.present?
         User.where(id: user_ids).update_all(title: self.title)
       end
+
+      if self.grant_trust_level.present?
+        User.where(id: user_ids).find_each do |user|
```
ZogStriP: How does this scale? What if `user_ids` has 1M items?

techAPJ: That is a valid concern. What should be the correct approach here? A new background job for changing trust level in bulk?

ZogStriP: Can't this be done in raw SQL?

ZogStriP: Or like this? `User.where(id: user_ids).where("trust_level < ?", grant_trust_level).update_all(trust_level: grant_trust_level, trust_level_locked: true)`

techAPJ: I looked into that approach (raw SQL) but it seems like `user.change_trust_level!` calls `Promotion.new(self).change_trust_level!` which is [doing many things](https://github.com/discourse/discourse/blob/5dbd6a304bed5400be481d71061d3e3ebb4d6785/lib/promotion.rb#L39-L78), so it might not be possible to accommodate all that in raw SQL?

The same concern applies for using `update_all` directly.

app/models/group.rb
```diff
@@ -414,6 +414,16 @@ def bulk_add(user_ids)
...
+          if user.trust_level < self.grant_trust_level
+            user.change_trust_level!(self.grant_trust_level)
+            user.trust_level_locked = true
+            user.save
```
tgxworld: Unless we're checking for the return value, a good practice is to use `save!` or `update!`. Otherwise, this can fail silently without us know about it.

app/models/group.rb
```diff
@@ -414,6 +414,16 @@ def bulk_add(user_ids)
       if self.title.present?
         User.where(id: user_ids).update_all(title: self.title)
       end
+
+      if self.grant_trust_level.present?
+        User.where(id: user_ids).find_each do |user|
+          if user.trust_level < self.grant_trust_level
+            user.change_trust_level!(self.grant_trust_level)
+            user.trust_level_locked = true
```
tgxworld: Hmm we do we need to lock their trust level?

app/models/group.rb
techAPJ: 👍  done.

app/models/group.rb
techAPJ: We're already [doing this](https://github.com/discourse/discourse/blob/5dbd6a304bed5400be481d71061d3e3ebb4d6785/app/models/group_user.rb#L60) when adding individual user to a group.

tgxworld: The code isn't DRY here :wink:

app/models/group.rb
techAPJ: @ZogStriP I added a job to run bulk granting of trust level process in background. Looks good now?

app/models/group.rb
techAPJ: This code is now extracted out in a background job. Does it makes sense to use that background job to change trust level of an _individual_ user as well?

app/jobs/regular/bulk_grant_trust_level.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def execute(args)
+      trust_level = args[:trust_level]
+      user_ids = args[:user_ids]
+
+      raise Discourse::InvalidParameters.new(:trust_level) if trust_level.blank?
+      raise Discourse::InvalidParameters.new(:user_ids) if user_ids.blank?
+
+      User.where(id: user_ids).find_each do |user|
+        if user.trust_level < trust_level
```
tgxworld: Maybe just extract this logic below and https://github.com/discourse/discourse/blob/master/app/models/group_user.rb#L56-L63 into a concern or Service object?

techAPJ: Okay, I extracted the logic out of both places into a new Service object.

PR #4737, opened 2017-03-03
FIX: approve invited user

oblakeerickson:
This commit fixes the case where invited users who typed in a password
would not be approved by default. Because we moved the user create logic
for an invited user there was a clash with the `save` in the user model
and the `save` in the invite_redeemer class.

- added approve logic into invite_redeemer class.
- added tests to verify that the user is approved

### DIFF
--- a/app/models/invite_redeemer.rb
+++ b/app/models/invite_redeemer.rb
@@ -36,6 +36,12 @@ InviteRedeemer = Struct.new(:invite, :username, :name, :password) do
       user.password = password
     end
 
+    if !SiteSetting.must_approve_users? || (SiteSetting.must_approve_users? && invite.invited_by.staff?)
+      user.approved = true
+      user.approved_by_id = invite.invited_by_id
+      user.approved_at = Time.zone.now
+    end
+
     user.moderator = true if invite.moderator? && invite.invited_by.staff?
     user.save!
 
@@ -49,11 +55,11 @@ InviteRedeemer = Struct.new(:invite, :username, :name, :password) do
   end
 
   def process_invitation
+    approve_account_if_needed
     add_to_private_topics_if_invited
     add_user_to_invited_topics
     add_user_to_groups
     send_welcome_message
-    approve_account_if_needed
     notify_invitee
     send_password_instructions
     delete_duplicate_invites
@@ -109,7 +115,9 @@ InviteRedeemer = Struct.new(:invite, :username, :name, :password) do
   end
 
   def approve_account_if_needed
-    invited_user.approve(invite.invited_by_id, false)
+    if get_existing_user
+      invited_user.approve(invite.invited_by_id, false)
+    end
   end
 
   def send_password_instructions
--- a/spec/models/invite_redeemer_spec.rb
+++ b/spec/models/invite_redeemer_spec.rb
@@ -9,6 +9,7 @@ describe InviteRedeemer do
       expect(user.name).to eq('Walter White')
       expect(user).to be_active
       expect(user.email).to eq('walter.white@email.com')
+      expect(user.approved).to eq(true)
     end
 
     it "can set the password too" do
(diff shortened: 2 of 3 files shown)


app/models/invite_redeemer.rb
```diff
@@ -36,6 +36,10 @@ def self.create_user_from_invite(invite, username, name, password=nil)
       user.password = password
     end
 
+    user.approved = true
+    user.approved_by_id = invite.invited_by_id
+    user.approved_at = Time.now
```
tgxworld: Don't we have to check that the invite was sent by a staff before automatically approving the user? Otherwise, a normal approved user can get another user in without approval.

oblakeerickson: Yes, good point!

Maybe we should only check for staff if "Staff must approve..." is checked?

app/models/invite_redeemer.rb
```diff
@@ -36,6 +36,18 @@ def self.create_user_from_invite(invite, username, name, password=nil)
...
+    if SiteSetting.must_approve_users?
+      if invite.invited_by.staff?
+        user.approved = true
+        user.approved_by_id = invite.invited_by_id
+        user.approved_at = Time.now
+      end
+    else
+      user.approved = true
+      user.approved_by_id = invite.invited_by_id
+      user.approved_at = Time.now
```
tgxworld: We're duplicating code here which is not ideal. 

I think we can just switch the condition to 

`if !SiteSetting.must_aprrove_users? || (SiteSetting.must_approve_users? && invite.invited_by.staff?)`

app/models/invite_redeemer.rb
```diff
@@ -36,6 +36,12 @@ def self.create_user_from_invite(invite, username, name, password=nil)
       user.password = password
     end
 
+    if !SiteSetting.must_approve_users? || (SiteSetting.must_approve_users? && invite.invited_by.staff?)
+      user.approved = true
+      user.approved_by_id = invite.invited_by_id
+      user.approved_at = Time.now
```
tgxworld: `Time.zone.now` is preferred in Rails project. This is a good read as well :) http://danilenko.org/2012/7/6/rails_timezones/

coding-horror:
The way it works in .NET is `DateTime.UtcNow` which is quite clear

oblakeerickson:
@tgxworld thanks for reviewing and for the suggestions!! :smiley:

app/models/invite_redeemer.rb
```diff
@@ -109,7 +115,9 @@ def send_welcome_message
   end
 
   def approve_account_if_needed
-    invited_user.approve(invite.invited_by_id, false)
+    if get_existing_user
```
oblakeerickson: Somewhere there is a password_reset test  that checks if the user is approved or not and since its a password reset the user already exists and doesn't go into `create_user_from_invite`.

I'll have to check again but I added the `get_existing_user` check because if its an invite it should not go to the User model to approve the user.

app/models/invite_redeemer.rb
```diff
@@ -49,11 +55,11 @@ def invited_user
   end
 
   def process_invitation
+    approve_account_if_needed
```
oblakeerickson: If we don't move it up this test fails:

rspec ./spec/models/invite_redeemer_spec.rb:55 # InviteRedeemer#redeem should redeem the invite if invited by non staff but not approve

because the user gets created and then it does the approve on the user model but the user model doesn't check for the invited by staff.

app/models/invite_redeemer.rb
oblakeerickson: this is the test I was referring to:

rspec ./spec/models/email_token_spec.rb:147 # EmailToken#confirm confirms the token and redeems invite marks the user as approved

PR #4738, opened 2017-03-06
less restrictive to allow plugin outlets

rimian:
for extra-nav-items


tgxworld:
Lets test this out :) As long as it doesn't break stuff :package:

PR #4740, opened 2017-03-06
Custom filters for Topic View

davidtaylorhq:
Added functions to TopicView which allow for plugins to add custom filters. This lets plugins make use of the "view x hidden replies option". For example, discourse-solved could use it for hiding any posts between the question and the answer as described [here](https://meta.discourse.org/t/option-to-move-accepted-answer-to-the-top/58386)

The implementation is based on the TopicQuery custom filters added recently. A test has also been added to the spec file for TopicView.

Example usage:
screen shot 2017-03-06 at 20 04 35

### DIFF
--- a/lib/topic_view.rb
+++ b/lib/topic_view.rb
@@ -37,6 +37,27 @@ class TopicView
     wpcf.flatten.uniq
   end
 
+  def self.add_custom_filter(key, &blk)
+    @custom_filters ||= {}
+    @custom_filters[key] = blk
+  end
+
+  def self.remove_custom_filter(key)
+    @custom_filters.delete(key)
+    @custom_filters = nil if @custom_filters.length == 0
+  end
+
+  def self.apply_custom_filters(results, topic_view)
+
+    if @custom_filters
+      @custom_filters.each do |key,filter|
+        results = filter.call(results, topic_view)
+      end
+    end
+    results
+
+  end
+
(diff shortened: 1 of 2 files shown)


lib/topic_view.rb
```diff
@@ -37,6 +37,27 @@ def self.whitelisted_post_custom_fields(user)
...
   end
 
+  def self.add_custom_filter(key, &blk)
+    @custom_filters ||= {}
+    @custom_filters[key] = blk
+  end
+
+  def self.remove_custom_filter(key)
+    @custom_filters.delete(key)
+    @custom_filters = nil if @custom_filters.length == 0
```
tgxworld: I think we can just set `@custom_filters` to a `{}` instead.

davidtaylorhq: Yeah that seems reasonable - there's probably no need for line 47 at all.

This stuff is mostly [taken from TopicQuery](https://github.com/discourse/discourse/blob/master/lib/topic_query.rb#L78) so I wonder if @SamSaffron had a reason for setting it to `nil`?

SamSaffron:
I just feel uncomfortable about adding this extensibility I feel there are way better ways of skinning this particular cat

davidtaylorhq:
That's fine. I'll keep it to one side and if anyone feels the need for it in future it can be dug out :)

PR #4741, opened 2017-03-08
FIX: Allow user to remove bookmark from posts as long as bookmark is …

tgxworld:
…present.

https://meta.discourse.org/t/bookmark-issue-when-access-to-topic-is-lost-pms/51993


SamSaffron:
looks fine to me

PR #4742, opened 2017-03-08
FIX: latest-version tag is shown by git-describe

davidtaylorhq:
Adds the -match "v[0-9]*" parameter to git describe, this means that only version tags will be used.

See https://meta.discourse.org/t/latest-version-bug/58165/35


SamSaffron:
cool, lets try it! thanks

PR #4743, opened 2017-03-08
FIX: Category autocomplete breaks when search menu widget rerenders.

tgxworld:
https://github.com/discourse/discourse/pull/4717#issuecomment-28491458

As the default category autocomplete search results has already been cached on page load, rendering the category autocomplete is fast. Being too fast here causes the element to be included when VDOM runs its diffing algorithm to figure out what has changed. For some reason that is beyond my understanding of the VDOM library, the autocomplete popup is being replaced with the `search-menu-results` div but results in an error in the process. This PR fixes the issue by ensuring that we do not append the autocomplete popup with the div that is being re-rendered by instead append it in a parent div that is will not re-render as the search term is changed.

@eviltrout Can you review this? Thank you so much!

cc @cpradio

screenshot from 2017-03-08 20-12-43

### DIFF
--- a/app/assets/javascripts/discourse/lib/autocomplete.js.es6
+++ b/app/assets/javascripts/discourse/lib/autocomplete.js.es6
@@ -261,7 +261,11 @@ export default function(options) {
       left: "-1000px"
     });
 
-    me.parent().append(div);
+    if (options.appendSelector) {
+      me.parents(options.appendSelector).append(div);
+    } else {
+      me.parent().append(div);
+    }
 
     if (!isInput && !options.treatAsTextarea) {
       vOffset = div.height();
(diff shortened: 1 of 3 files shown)


cpradio:
Ah, you beat me to it. I figured I was still a day or two away from completely figuring out what you deciphered (based on every time I took a step away and came back, I uncovered something new about the problem). That is a very interesting solution to the problem. 

Looks like ti will still work for the Advanced Search page too, as options should be null in that case and thus not supplying a special selector to append the autocomplete results to.

tgxworld:
@cpradio Thanks for the initial debugging :) Your tip about `widget.scheduleRerender` helped me to track it down pretty easily.

PR #4745, opened 2017-03-08
Improving search result blurb

jumbojett:
# Problem

Discourse _blurb_ does not correlate with the Postgres methods used to index and search data. 

# Solution

This PR does the following:

- Removes Ruby methods that build "blurbs" in search results. In place, harnessing powerful Postgres `ts_headline` to build better blurbs.
- Removes unnecessary index data from raw_data fields (topic title) in posts, and topic indexes leaving search_data fields in current format
- Improves result processing by using preprocessed raw_data
- Removes blurb-length search option due to irrelevance in blurb calculations

### DIFF
--- a/app/assets/javascripts/discourse/templates/full-page-search.hbs
+++ b/app/assets/javascripts/discourse/templates/full-page-search.hbs
@@ -107,9 +107,7 @@
             </span>
 
             {{#if result.blurb}}
-              {{#highlight-text highlight=q}}
-                {{{unbound result.blurb}}}
-              {{/highlight-text}}
+              {{{unbound result.blurb}}}
             {{/if}}
           </div>
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -12,8 +12,7 @@ class SearchController < ApplicationController
     search_args = {
       type_filter: 'topic',
       guardian: guardian,
-      include_blurbs: true,
-      blurb_length: 300
+      include_blurbs: true
     }
 
     context, type = lookup_search_context
--- a/app/controllers/similar_topics_controller.rb
+++ b/app/controllers/similar_topics_controller.rb
@@ -10,9 +10,6 @@ class SimilarTopicsController < ApplicationController
 
     attr_reader :topic
 
-    def blurb
-      Search::GroupedSearchResults.blurb_for(@topic.try(:blurb))
-    end
   end
 
   def index
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -459,8 +459,9 @@ class Topic < ActiveRecord::Base
 
     return [] unless candidate_ids.present?
 
-    similar = Topic.select(sanitize_sql_array(["topics.*, similarity(topics.title, :title) + similarity(topics.title, :raw) AS similarity, p.cooked as blurb", title: title, raw: raw]))
+    similar = Topic.select(sanitize_sql_array(["topics.*, similarity(topics.title, :title) + similarity(topics.title, :raw) AS similarity, COALESCE(ts_headline(#{Search.query_locale}, post_search_data.raw_data,#{ts_query}, 'StartSel=\"<span class=search-highlight>\",StopSel=\"</span>\",MaxWords=15,MinWords=10,MaxFragments=2'), '') AS blurb", title: title, raw: raw]))
                      .joins("JOIN posts AS p ON p.topic_id = topics.id AND p.post_number = 1")
+                     .joins("JOIN post_search_data ON post_search_data.post_id = p.id")
                      .limit(SiteSetting.max_similar_results)
                      .where("topics.id IN (?)", candidate_ids)
                      .where("similarity(topics.title, :title) + similarity(topics.title, :raw) > 0.2", raw: raw, title: title)
--- a/app/serializers/search_post_serializer.rb
+++ b/app/serializers/search_post_serializer.rb
@@ -4,6 +4,6 @@ class SearchPostSerializer < BasicPostSerializer
   attributes :like_count, :blurb, :post_number
 
   def blurb
-    options[:result].blurb(object)
+    object.blurb.to_str
   end
 end
--- a/app/serializers/similar_topic_serializer.rb
+++ b/app/serializers/similar_topic_serializer.rb
@@ -8,7 +8,7 @@ class SimilarTopicSerializer < ApplicationSerializer
   end
 
   def blurb
-    object.topic.blurb
+    object.topic['blurb']
   end
 
   def url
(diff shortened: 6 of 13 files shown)


SamSaffron:
can you include some specific examples of how this improves things?

jumbojett:
> specific examples of how this improves things

Sure @SamSaffron ! Specifically:

- Now blurbs directly reflect the original `ts_vector` query. This means the extracted text more relevant.
- Terms are hi-lighted with respect to stemming. If a user searches for _computing_ and the search engine matches on _computer_ then this term will be hi-lighted. The Ruby and Javascript methods cannot detect this.
- Blurbs are created from a lexical analysis and contain whole words rather than cutting up strings

# Example
They say a picture is worth a thousand words. 
## Before
image

## After
image

SamSaffron:
hmmm that sample is showing a blurb that has no indication that it is taking part of the sentence, can you add the ellipsis on both ends if it is a partial quote?

jumbojett:
> can you add the ellipsis on both ends if it is a partial quote?

@SamSaffron Yes! Ellipsis for partials are currently rendered for results that have more than one fragment. `ts_header` appends ellipsis for multiple fragments. The number of fragments are limited to three and generated based on relevance to the vector query. Here's another example where ellipses appear in blurbs.

image

SamSaffron:
I follow, but what about the leading and trailing ellipsis? Should it not be? 

...think we should add a new tag to the system that isn't included ... know by replying down below. Suggest new tags Here...

SamSaffron:
Also, I wonder what will happen in other locales, especially Chinese where the tokenizing is all messed up

jumbojett:
> what about the leading and trailing ellipsis

Postgres doesn't give a way to control these at the moment. I suppose a '...' could be prepended after the fact if a capital letter is not detected? 

> I wonder what will happen in other locales, especially Chinese where the tokenizing is all messed up

I'm not sure. This PR is a direct reflection of the current `ts_vector` query. In other words, if your current search functions with Chinese locales, then the blurbs _should_ automatically reflect it.

SamSaffron:
The problem with CJK languages is that PG support is super uneven. So we do the tokenizing in ruby. 

In Chinese, for example: asentencecanlooklikethis

We go ahead and split it up and store: "a sentence can look like this" 

If you we followed this blurb scheme in Chinese the blurbs would look incredibly confusing.

jumbojett:
> We go ahead and split it up and store: "a sentence can look like this"

Ah I understand - Are there any Chinese users who could help test? I'm thinking worst case is that a blurb is not returned. 

Could the PR be accepted in the meantime and file an open issue to look into this? I left the `blurb_for` method in place for the event we need to have an alternative game-plan for Chinese text.

SamSaffron:
I can't really break Chinese here, that is a non starter, @fantasticfears  may be able to help you test this

jumbojett:
@fantasticfears if you can help in this area it would be greatly appreciated. 😊 

👌  @SamSaffron I have an up-and-coming PR that builds off of this one. If there's anything else I can do to increase the chances of its acceptance let me know.

SamSaffron:
@jumbojett If you made this optional, default off, via a site setting I would feel much more comfortable accepting it. 

Then we can experiment with it

jumbojett:
Site settings sounds like a possible path, but I'm afraid of the complexity it might introduce when trying to build upon this revision.

SamSaffron:
can you expand on why your next PR needs this change?

jumbojett:
> expand on why your next PR needs this change?

Next PR will be for expanded searches. Searches that span topic-wide. (Similar to how google would search a discourse instance.) This requires `raw_data` to be specifically formatted according to this PR.

fantasticfears:
Sure I can help you with testing in Chinese. I'll come back with testing results later. The picture looks awesome!

Would you mind drawing some lines about your plans on [Meta](https://meta.discourse.org/c/dev)?
Sam has to keep maintaining the support for CJK search. And I'd also like to work on improving searching Chinese for Discourse later on. It would be great for the community to discuss.

jumbojett:
Great! Thanks for your help @fantasticfears. 

> Would you mind drawing some lines about your plans on Meta?

Sure thing. I'll be sure to detail plan as soon as it's fully formulated.

fantasticfears:
I may run into some locale problem which I can't reproduce the production search with the local docker. Anyway, it's better to just tell you how the local tests with your PR. It breaks the sentences with blanks where it shouldn't be. It looks bad but the quality seems nice. Great work!

And Chinese search appears having some problems back in months and it's not the index problem. Your PR seems fixing it somehow...

image

As of the dev process, if I can submit proposal in GSoC this year, I'll have the opportunity to address CJK searches from the ground up. It's quite hard to reproduce or address the searches on my end :(

jumbojett:
Thank you for helping us test @fantasticfears !  

> if I can submit proposal in GSoC this year, I'll have the opportunity to address CJK searches from the ground up

That would be great.

@SamSaffron Thoughts moving forward?

SamSaffron:
I feel like we have to fix the spacing issue in chinese before accepting this, I think the best way to move forward here is to make this a site setting

jumbojett:
@fantasticfears I found the following Postgres extension. If this were included as part of the base discourse docker image, could it address white-space formatting?

https://github.com/amutu/zhparser

fantasticfears:
@jumbojett My bad. I think my backup may override the index. Now I checkout the version before your PR and restore. Then I rebuild to your branch. The space issue is a false negative. Sorry for the confusion.

image

But the search blurb may not show the search term. `raw_data` isn't being tokenized. The difference is that only the markdown marks are stripped.
And it seems bold text (style) are not filtered by the method.

https://github.com/amutu/zhparser is an tokenizer which breaks the sentences into words. The tokenization is done in Ruby now. It is tempting but harder for @SamSaffron to maintain if not stable enough.

SamSaffron:
I still don't get the reluctance to make this optional for the time being, its quite simple to do the switching cause the blurbs can be generated server side and highlighting done server sides. So the whole difference ends up being weather you select an extra column or not.

jumbojett:
> I still don't get the reluctance to make this optional for the time being

@SamSaffron apologies for the reluctance. I'm new to ruby and find it intimidating to some extent.

> highlighting done server sides

I welcome assistance. I'm not entirely sure the best way to approach this in ruby.

tgxworld:
@jumbojett Just wanted to check if there are any updates on this?

jumbojett:
Hi @tgxworld . Thanks for checking in. I've been unable to tend to this PR. I'm not sure when I'll be able to update it. I'll add you as a collab on my fork in the event you'd like to finalize it based on @SamSaffron 's advice. How does that sound?

PR #4747, opened 2017-03-10
Show disable delete btn cat

gsambrotta:
It show disabled button when category cannot be deleted.
On hover or on tap of the button, it will show explanatory tooltip.
 Steps to reproduce:
- go to a category main page. Category must have some old posts inside
- Edit the category
- Corner right of the modal edit-category, you should now see a buttons instead of text.
- Hover or tap on this button to see more info.

Reference isssue: https://meta.discourse.org/t/show-disabled-delete-button-when-category-cant-be-deleted/58199/2

screen shot 2017-03-10 at 14 02 39

### DIFF
--- a/app/assets/javascripts/discourse/controllers/edit-category.js.es6
+++ b/app/assets/javascripts/discourse/controllers/edit-category.js.es6
@@ -8,6 +8,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
   saving: false,
   deleting: false,
   panels: null,
+  hiddenTooltip: true,
 
   _initPanels: function() {
     this.set('panels', []);
@@ -16,6 +17,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
   onShow() {
     this.changeSize();
     this.titleChanged();
+    this.set('hiddenTooltip', true);
   },
 
   changeSize: function() {
@@ -101,6 +103,13 @@ export default Ember.Controller.extend(ModalFunctionality, {
           self.set('deleting', false);
         }
       });
+    },
+
+    toggleDeleteTooltip() {
+      // check if is touch device
+      if ( $('html').hasClass('discourse-touch') ) {
+        this.toggleProperty('hiddenTooltip');
+      }
     }
   }
 
--- a/app/assets/stylesheets/mobile/modal.scss
+++ b/app/assets/stylesheets/mobile/modal.scss
@@ -38,7 +38,7 @@
   width: 95%;
 }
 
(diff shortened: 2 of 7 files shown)


coding-horror:
Looks really good, visually at least!

app/assets/javascripts/discourse/controllers/edit-category.js.es6
```diff
@@ -101,6 +101,19 @@ export default Ember.Controller.extend(ModalFunctionality, {
           self.set('deleting', false);
         }
       });
+    },
+
+    showTooltip() {
```
SamSaffron: I think a much cleaner way of doing this would be to bind the classname to .cannot_delete_reason, then the action can simply toggle the class name, perhaps something along these lines

```
class="{{if deleteHidden "hidden" ""}}"
```

As a rule we prefer only to reach in to $ if we have no other option

app/assets/stylesheets/mobile/modal.scss
```diff
@@ -135,6 +135,35 @@
   }
 }
```
SamSaffron: there seems to be a fair bit in common with mobile and desktop here, can any of the duplicate stuff be moved to modal common or something?

gsambrotta: Sure! I didn't notice there was `common/modal.scss`

app/assets/javascripts/discourse/controllers/edit-category.js.es6
```diff
@@ -101,6 +102,18 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+    },
+
+    showDeleteTooltip() {
+      // check if is touch device
+      if ($('html').hasClass('discourse-touch')) {
+
+        if(this.get('hiddenTooltip')){
+          this.set('hiddenTooltip', false);
+        } else {
+          this.set('hiddenTooltip', true);
```
gsambrotta: @SamSaffron I improve the actions as you suggest. I notice my code wasn't following the rest but i needed little guide on how to improve it. So thank you for the hint.
Let me know if now looks better

SamSaffron: Thanks heaps, yes, much better

Can you try? 

```
this.toggleProperty('hiddenTooltip') ;
```

gsambrotta:
@SamSaffron thanks again for the tip. What about now?

SamSaffron:
looks good, thanks!

PR #4750, opened 2017-03-13
FEATURE: Redirect to groups page and apply group actions upon login/s…

tgxworld:
…ign up.

https://meta.discourse.org/t/is-there-a-working-group-registration-flow/56798

@ZogStriP I'm not sure if my design using localstorage is ideal. Can you review this PR for me? Thank you :heart:

### DIFF
--- a/app/assets/javascripts/discourse/components/group-membership-button.js.es6
+++ b/app/assets/javascripts/discourse/components/group-membership-button.js.es6
@@ -22,30 +22,25 @@ export default Ember.Component.extend({
     }
   },
 
-  @computed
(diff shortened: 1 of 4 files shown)


ZogStriP:
> @ZogStriP I'm not sure if my design using localstorage is ideal. Can you review this PR for me? Thank you ❤️

Why is that information stored client-side rather than server-side (eg. in the session?)

tgxworld:
> Why is that information stored client-side rather than server-side (eg. in the session)?

We don't need the data server side plus it is much easier to trigger the JS actions by fetching from a local store.

PR #4751, opened 2017-03-13
Added `ssl` parameter for action_mailer

hyunjong-lee:
append `ssl` option for `action_mailer` to deal with the following error.
- http://stackoverflow.com/questions/6579865/in-rails-3-and-actionmailer-is-it-possible-to-send-email-using-tls-over-ssl-no

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -32,7 +32,8 @@ Discourse::Application.configure do
       user_name:            GlobalSetting.smtp_user_name,
       password:             GlobalSetting.smtp_password,
       authentication:       GlobalSetting.smtp_authentication,
-      enable_starttls_auto: GlobalSetting.smtp_enable_start_tls
+      enable_starttls_auto: GlobalSetting.smtp_enable_start_tls,
+      ssl:                  GlobalSetting.smtp_enable_ssl
     }
 
     settings[:openssl_verify_mode] = GlobalSetting.smtp_openssl_verify_mode if GlobalSetting.smtp_openssl_verify_mode
--- a/config/discourse_defaults.conf
+++ b/config/discourse_defaults.conf
@@ -80,6 +80,9 @@ smtp_enable_start_tls = true
 # to disable, set to 'none'
 smtp_openssl_verify_mode =
 
+# mode for enable ssl in stmp settings
+smtp_enable_ssl = true
+
 # load MiniProfiler in production, to be used by developers
 load_mini_profiler = true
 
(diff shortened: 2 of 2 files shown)


config/environments/production.rb
```diff
@@ -32,7 +32,8 @@
       user_name:            GlobalSetting.smtp_user_name,
       password:             GlobalSetting.smtp_password,
       authentication:       GlobalSetting.smtp_authentication,
-      enable_starttls_auto: GlobalSetting.smtp_enable_start_tls
+      enable_starttls_auto: GlobalSetting.smtp_enable_start_tls,
+      ssl:                  GlobalSetting.smtp_enable_ssl
```
tgxworld: Hi @hyunjong-lee, Is this a new setting in Rails 5 because the docs doesn't seem to indicate that `ssl` is a valid option.

screenshot from 2017-03-14 09-44-11

tgxworld:
Hi @hyunjong-lee, please sign the CLA in order for us to accept this PR :)

hyunjong-lee:
Hi @tgxworld, I accepted CLA :)

tgxworld:
@hyunjong-lee Thanks :) Did you see my comments in https://github.com/discourse/discourse/pull/4751#discussion_r105812100 ?

config/environments/production.rb
hyunjong-lee: @tgxworld I think the below line indicates the option that I included.
- https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/base.rb#L409

While I am not familiar with ruby & rails, I am trying to find source codes that contains the option.

tgxworld:
@hyunjong-lee Looks like this config is [only added in Rails 5.1](https://github.com/rails/rails/commit/7cffd6b372369e93ef74c01229a0507d4086e169) we it wouldn't work for Discourse until we upgrade Rails. Thanks for contributing though.

hyunjong-lee:
@tgxworld 
I found some codes that using `ssl`, and the oldest one had committed 7 years ago.
- https://github.com/rails/rails/commit/92f49b5f1ebf42514c58e1fda87c0b8a1b33d08f#diff-53e65e5b02bb6a167c2270e40aa780fdR11

And the relative question in stackoverflow was asked 6 years ago.
- http://stackoverflow.com/questions/6579865/in-rails-3-and-actionmailer-is-it-possible-to-send-email-using-tls-over-ssl-no

I have no idea why the option was not included in official documents, it actually worked since 7 years ago. So, can you consider this PR once again?

PR #4752, opened 2017-03-13
FEATURE: Add "Show tracked topics" link to User Prefs

ElTipejoLoco:
All this pull request does is what it says on the tin. ?state=tracking is already supported in Discourse, and all this will do is add a link to the user preferences page that lets people click to see the topics they're currently tracking (as opposed to the ones they're watching or have muted).

Mostly I'm just testing the waters a little since I'd like to see if I can figure out a good, lightweight way to track unread posts by topic creators/owners only, and set that up.


ElTipejoLoco:
Removing to re-add, supposing that combination of discoursebot message not showing CLA was signed, unsquashed commit, and feature not being in wanted list is cause for delay in merging.

PR #4753, opened 2017-03-13
DEV: Add plugin-outlet to allow SSO sites to put a message on the User Preferences page

cpradio:
Relevant Discussion:
https://meta.discourse.org/t/add-link-to-external-sso-profile-to-profile-page/59027/7?u=cpradio


SamSaffron:
thanks, sure seems fine

PR #4754, opened 2017-03-14
FIX: Improve checks for non-human users.

tgxworld:


ZogStriP:
Looks 👍. Have you checked for `-1` too? Sometimes we don't use the constant.

tgxworld:
@ZogStriP  Good point let me check too :+1:

tgxworld:
Merged in https://github.com/discourse/discourse/commit/5943543ec30fe5194b0126bd05477e7969b20036

PR #4759, opened 2017-03-15
FIX: Allow ports for embed host IPs

kennym:
See https://meta.discourse.org/t/specifying-port-for-embeddable-host-given-an-ip-results-in-error/59209 for details


eviltrout:
Thanks, looks good to me!

PR #4760, opened 2017-03-16
Fixes duplicated title in header after edit

nbianca:
Fixes the issue described [here](https://meta.discourse.org/t/topic-title-duplicated-in-header/58599).

### DIFF
--- a/app/assets/javascripts/discourse/components/discourse-topic.js.es6
+++ b/app/assets/javascripts/discourse/components/discourse-topic.js.es6
@@ -64,6 +64,24 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
     this.appEvents.on('post:highlight', postNumber => {
       Ember.run.scheduleOnce('afterRender', null, highlight, postNumber);
     });
+
+    this.appEvents.on('header:update-topic', topic => {
+
+      if (topic === null) {
+        this._lastShowTopic = false;
+        this.appEvents.trigger('header:hide-topic');
+        return;
+      }
+
+      const offset = window.pageYOffset || $('html').scrollTop();
+      this._lastShowTopic = this.showTopicInHeader(topic, offset);
+
+      if (this._lastShowTopic) {
+        this.appEvents.trigger('header:show-topic', topic);
+      } else {
+        this.appEvents.trigger('header:hide-topic');
+      }
+    });
   },
 
   willDestroyElement() {
--- a/app/assets/javascripts/discourse/controllers/topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic.js.es6
@@ -913,7 +913,7 @@ export default Ember.Controller.extend(SelectedPostsCount, BufferedContent, {
       if (data.reload_topic) {
         topic.reload().then(() => {
           this.send('postChangedRoute', topic.get('post_number') || 1);
-          this.appEvents.trigger('header:show-topic', topic);
+          this.appEvents.trigger('header:update-topic', topic);
         });
       } else {
         if (topic.get('isPrivateMessage') &&


SamSaffron:
@nbianca really appreciate this, but I feel the fix is not in the right spot. 

I think it should always be safe to trigger the `header:show-topic` event it should be responsible for ensuring that it is in the correct mode prior to rendering. This logic of deciding if we should render the topic header would be in 2 places, duplicated with this change. 

@eviltrout what do you thing?

nbianca:
@SamSaffron By saying that it should be always safe to trigger `header:show-topic` do you suggest that `header:show-topic` and `header:hide-topic` should be combined? Maybe something like `header:update-topic`? The event handler would check whether the topic title should be displayed or not and act accordingly.

SamSaffron:
I see, 

My recommendation would be maybe to add an extra event called "header:update" that can be used whenever anyone feels like updating the header, then the logic can be centralized

eviltrout:
`header:update-topic` is probably a better name for the event. I think it should handle the logic of choosing to display the header depending on the scrollTop.

SamSaffron:
@nbianca keep in mind the logic is a bit more subtle than your test was. 

See: components/discourse-topic.js

What I would like is for all the logic to be centralized there, discourse-topic already knows how to make the call here so it should be the one handling this new event.

app/assets/javascripts/discourse/components/discourse-topic.js.es6
```diff
@@ -117,16 +135,11 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
...
     const showTopic = this.showTopicInHeader(topic, offset);
     if (showTopic !== this._lastShowTopic) {
-      this._lastShowTopic = showTopic;
-
-      if (showTopic) {
-        this.appEvents.trigger('header:show-topic', topic);
-      } else {
-        this.appEvents.trigger('header:hide-topic');
-      }
+      this.appEvents.trigger('header:update-topic', topic);
```
nbianca: I did this to avoid unnecessary updates and event triggers.

app/assets/javascripts/discourse/components/discourse-topic.js.es6
```diff
@@ -117,16 +135,11 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
 
     this.set('hasScrolled', offset > 0);
 
+    // Trigger a header (topic title) update, but only if it is requried.
     const topic = this.get('topic');
     const showTopic = this.showTopicInHeader(topic, offset);
     if (showTopic !== this._lastShowTopic) {
```
nbianca: I did this to avoid unnecessary updates and event triggers.

app/assets/javascripts/discourse/components/discourse-topic.js.es6
```diff
@@ -64,6 +64,24 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
...
+    this.appEvents.on('header:update-topic', topic => {
+
+      if (topic === null) {
+        this._lastShowTopic = false;
+        this.appEvents.trigger('header:hide-topic');
+        return;
+      }
+
+      const offset = window.pageYOffset || $('html').scrollTop();
+      this._lastShowTopic = this.showTopicInHeader(topic, offset);
```
nbianca: Basically this forces an update.

nbianca: I also do not that think that `offset` should be a parameter and it would be better if it were moved inside `showTopicInHeader`.

app/assets/javascripts/discourse/components/discourse-topic.js.es6
```diff
@@ -117,16 +135,11 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
...
     const topic = this.get('topic');
     const showTopic = this.showTopicInHeader(topic, offset);
     if (showTopic !== this._lastShowTopic) {
-      this._lastShowTopic = showTopic;
-
-      if (showTopic) {
-        this.appEvents.trigger('header:show-topic', topic);
-      } else {
-        this.appEvents.trigger('header:hide-topic');
-      }
```
SamSaffron: I think some of the duplication here is fine, cause otherwise you have to call showTopicInHeader twice, for no great gain

nbianca: Done.

app/assets/javascripts/discourse/components/discourse-topic.js.es6
SamSaffron: agree we can further refine this, but I think this fix for now is good

SamSaffron:
Thanks for the fix @nbianca !

PR #4761, opened 2017-03-16
FIX: Add route for user TL3 requirements page

joebuhlig:
For reference: https://meta.discourse.org/t/user-tl3-requirements-page-missing-on-refresh/59261


nlalonde:
Thanks for the fix! This was my bad... :blush:

PR #4762, opened 2017-03-16
FIX: Do not show User Cannot See Mention dialog when crafting a new topic or a new message

cpradio:
Use the model/composer topic.id to determine what users cannot be mentioned and what message to show as a warning.

Relevant Discussion:
https://meta.discourse.org/t/bug-when-composing-topic-while-viewing-pm-incorrect-mentioned-user-will-not-be-notified-warning/59270


tgxworld:
Merged in https://github.com/discourse/discourse/commit/aeb169bd0ef2e77cb205d00c6fbca91f91bac56b

Thank you @cpradio :+1:

PR #4768, opened 2017-03-20
Delete disabled Category button: remove tooltip on hover on desktop

gsambrotta:
Remove hover effect on desktop for tooltip. 
Fix requets after testing and discovering not smooth behavour of tooltip.

Issue: https://meta.discourse.org/t/show-disabled-delete-button-when-category-cant-be-deleted/58199

Fix Requested:
https://meta.discourse.org/t/show-disabled-delete-button-when-category-cant-be-deleted/58199/5


SamSaffron:
thanks!

PR #4769, opened 2017-03-20
add style to show admin/users header as clickable

gsambrotta:
Adding style to admin/users/list/active to look like the one in /users
- add pointer to mouse on hover to sortable columns
-  add grey background on hover to sortable columns
- remove top border on title's columns to replicate /users look
- give some top margin to the table to replicate /users look

Issue: https://meta.discourse.org/t/admin-users-list-does-not-indicate-which-columns-are-sortable/59268


tgxworld:
Thank you @gsambrotta  :+1:

PR #4770, opened 2017-03-20
Change name from admin page.

nbianca:
Adds a button to change the name from admin page.

Fixes [this](https://meta.discourse.org/t/change-a-users-name-from-the-admin-page/59425).

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-user-index.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-user-index.js.es6
@@ -4,6 +4,8 @@ import { propertyNotEqual, setting } from 'discourse/lib/computed';
 import { userPath } from 'discourse/lib/url';
 
 export default Ember.Controller.extend(CanCheckEmails, {
+  editingUsername: false,
+  editingName: false,
   editingTitle: false,
   originalPrimaryGroupId: null,
   availableGroups: null,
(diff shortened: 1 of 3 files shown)


SamSaffron:
there is a slight consistency issue here. 

- Change username vs. Edit Name. We should just use one word here, not too, it should probably simply be "Edit Username" 

- Change username appears to work differently, this edits in place, change username edits on its own page. I would recommend changing "edit username" to be done the same way you did it for this field.

nbianca:
@SamSaffron The "change" vs "edit" issue is found quite often. There are many fields using "change" (e.g. Change Username, Change About Me, Change Email, etc.) and about the same number of fields using "edit" (e.g. Edit Title, Edit Badges, etc.). I will change the "Change Username" to "Edit Username".

YesThatAllen:
@SamSaffron which is a better match here... `Change Username` or `Edit Title` down below?

screenshot_4498

nbianca:
This is how it looks now:

image

SamSaffron:
cool, is there a leftover template and route for the previous way of editing username?

nbianca:
That is still being used in user's preferences. Should that one be changed as well? Because if it is going to, there are many more fields that should be changed (email, user card badge, etc.)

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -54,6 +56,44 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+    saveUsername() {
+      const self = this;
+
+      return ajax(`/users/${this.get('model.username').toLowerCase()}/preferences/username`, {
+        data: { new_username: this.get('userUsernameValue') },
+        type: 'PUT'
+      }).catch(function(e) {
+        bootbox.alert(I18n.t("generic_error_with_reason", {error: "http: " + e.status + " - " + e.body}));
+      }).finally(function() {
+        self.set('model.username', self.get('userUsernameValue'));
```
SamSaffron: this is not really correct, if it fails to save finally will still be called and username will be changed incorrectly 

instead the pattern you should follow. 

1. amend model username
2. issue the PUT request
3. if there is an error, revert model username
4. finally toggle the edit

SamSaffron:
@coding-horror  copy wise, do you prefer "Edit name/username" or "Change name/username" ?

nbianca:
Made the changes after review. Would you like me to squash the commits?

SamSaffron:
Sure, go for it, let's wait on @coding-horror to confirm what copy he would prefer here.

coding-horror:
I actually feel _no_ copy is better here, just go with the pencil glyph on both, less to translate, less words on the screen for everyone to process in their mind grapes..

SamSaffron:
@coding-horror what about "Edit Title" ... just a pencil there as well ?

coding-horror:
Yeah just edit pencil button, much easier.

nbianca:
After you click `Edit`, the input and a button labeled `Save {x}` appears. Should that be a glyph as well or should I keep the text in there?

nbianca:
Done. The save button now simply says `Save`, not `Save {x}`.

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -55,21 +57,65 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+
+    saveUsername() {
+      const self = this;
+      const old_username = self.get('model.username');
+
+      self.set('model.username', self.get('userUsernameValue'));
+      return ajax(`/users/${old_username.toLowerCase()}/preferences/username`, {
+        data: { new_username: this.get('userUsernameValue') },
+        type: 'PUT'
+      }).catch(function(e) {
```
tgxworld: We prefer to use [Arrow functions](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in our code base.

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -55,21 +57,65 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+    },
+
+    toggleNameEdit() {
+        this.set('userNameValue', this.get('model.name'));
+        this.toggleProperty('editingName');
+    },
+
+    saveName() {
+      const self = this;
+      const old_name = self.get('model.name');
```
tgxworld: Style: We use camelCase for our variable names for JS files.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/9065df76ade33f763ef9b7dc5d225f874433d4f4

tgxworld:
Aside from the minor comments, the PR looks good. Thanks @nbianca  :+1:

PR #4775, opened 2017-03-21
abstract out some complexity in create

rimian:
There are a few levels of abstraction here. This makes the method easier to digest.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -296,15 +296,15 @@ class UsersController < ApplicationController
       return fail_with("login.new_registrations_disabled")
     end
 
-    if params[:password] && params[:password].length > User.max_password_length
+    if params[:password] && password_too_long?
       return fail_with("login.password_too_long")
     end
 
-    if params[:email] && params[:email].length > 254 + 1 + 253
+    if params[:email] && email_too_long?
       return fail_with("login.email_too_long")
     end
 
-    if SiteSetting.reserved_usernames.split("|").include? params[:username].downcase
+    if reserved_username?
       return fail_with("login.reserved_username")
     end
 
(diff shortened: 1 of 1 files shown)


tgxworld:
I don't quite agree with the changes here because I don't feel like the complexity of the methods is being reduced. It makes the code easier to read but introduces overhead when debugging because you now have to look for another method. I don't think the changes here are worth it unless the methods are used in more than one spot.

rimian:
Sure, it reduces complexity. The create method no longer calculates how long an email should be, it doesn't know or care how to split site settings and doesn't need to perform the calculation on the password length. Too many layers of abstraction.

Also: Password length checking occurs two other times [here](https://github.com/discourse/discourse/blob/master/app/controllers/session_controller.rb#L182) and [here](https://github.com/discourse/discourse/blob/master/app/models/user.rb#L991).

I agree my change is not perfect and also respect the decision you guys make :)

tgxworld:
> Too many layers of abstraction.

Actually these changes increases the layers of abstraction because you've extracted certain logic into another method. 

Regarding password length, I can agree with the change if `params[:password].length > User.max_password_length` is extracted into a method in the `User`'s class and that method is reused in the other controllers. Otherwise, we're not really benefiting much from extracting those logics into private methods in the controllers.

rimian:
It's feature envy. The controller has handling the data as well as deciding what to do with it.

Anyway... I think the user should validate itself like you suggest. Would creating a `User` object then asking if it's valid be a better idea?

tgxworld:
> It's feature envy. The controller is handling the data as well as deciding what to do with it.

I don't follow how extracting those logic into private methods in the controller helps here because the controller is still handling the data and deciding what to do with it. 

> Anyway... I think the user should validate itself like you suggest. Would creating a User object then asking if it's valid be a better idea?

I'm not sure if this is possible but it is a better approach :+1:

rimian:
A healthy debate always keeps me on my toes and forces me to check what I say is correct. :)

I find it useful to ask myself if my methods are concerned with more than one thing. Those private methods should not be in the controller, yes. But at least they're factoring out making calculations on data into methods that do one thing. 

I'll see if I can get this to validate. Another thing... when I ran controller tests, I tried to force a fail so I knew I was editing the right line of code that the specs were testing. I couldn't get anything to fail! Maybe it was me...

tgxworld:
@rimian Im OK with these changes if we extract the logic into their respective models because it allows those method to be reused elsewhere.

tgxworld:
@rimian  I'm cleaning up our PRs so going to be closing this first. Feel free to open a new PR with the changes.

PR #4776, opened 2017-03-23
FEAUTRE: Add `SiteSetting` overrider option.

tgxworld:


eviltrout:
This seems good to me, but I think `overrider` sounds a bit awkward.

I'd suggest changing it to `overriden_by`. It's a lot clearer to me what the class is doing in that case.

SamSaffron:
I would actually prefer a simpler more general implementation

use the env var: `DISCOURSE_HIDE_SETTING_NAME=1` to hide a site setting. Then enable this on all the multisites. 

I think it is fine to require manual intervention for all restores, it is safer for our customers on multisites and only very rarely should customers be messing with restore.

tgxworld:
Closing this based on our discussion on dev.

PR #4777, opened 2017-03-23
Allow plugins to add panels to header more easily

gdpelican:
Hey Discourse!

I'm looking to add an additional icon & panel to the header widget. Discourse has a very nice way of adding icons:

```
api.decorateWidget('header-icons:before', function(helper) {
  helper.attach('header-dropdown', { icon: 'smiley' })
})
```

but adding panels along with those icons has proven a challenge. (Maybe I'm missing something obvious?) Before, I was doing this by (very nastily) traversing through the resultant vDom result from the `html` method:

```
const _html = queryRegistry('header').prototype.html
const _super  = _html.call(this, attrs, state)
let panels = _super.children[0].children[1].children
panels.push(this.attach(myWidget))
// ^^ this is really hard and bad!
```

Predictably, this was fragile and has broken. It would be much easier for me to hook into the html method where the panels are being added, so that I could append them there, rather than trying to dig for them afterwards. :D

### DIFF
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -347,7 +347,7 @@ class PluginApi {
    * and returns a hash of attrs to pass to the widget
    **/
    addHeaderPanel(name, toggle, attrFn) {
-     queryRegistry('header').prototype.attachAdditionalPanel(name, toggle, attrFn)
+     queryRegistry('header').prototype.attachAdditionalPanel(name, toggle, attrFn);
    }
 
   /**
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -18,9 +18,10 @@ import { addTagsHtmlCallback } from 'discourse/lib/render-tags';
 import { addUserMenuGlyph } from 'discourse/widgets/user-menu';
 import { addPostClassesCallback } from 'discourse/widgets/post';
 import { addPostTransformCallback } from 'discourse/widgets/post-stream';
+import { queryRegistry } from 'discourse/widgets/widget';
 
 // If you add any methods to the API ensure you bump up this number
-const PLUGIN_API_VERSION = '0.8.5';
+const PLUGIN_API_VERSION = '0.8.6';
 
 class PluginApi {
   constructor(version, container) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -224,6 +226,11 @@ export default createWidget('header', {
     return h('div.wrap', h('div.contents.clearfix', contents));
   },
 
+  // override to allow plugins to append additional panels
+  additionalPanels() {
```
tgxworld: This design pattern will only allow one plugin to override the panels at anyone time. The way that I'll do it is to expose a new API to add panels in [plugin-api](https://github.com/discourse/discourse/blob/747f4812e4323fa303bc2bfc84fcc61f715d7e4a/app/assets/javascripts/discourse/lib/plugin-api.js.es6).

gdpelican:
Alright, I have updated this to add a `addHeaderPanel` method to the plugin api

jorgtron:
When can this get merged? :) @tgxworld @eviltrout? Need this fix to get Babble working again on lots of forums.

app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -334,6 +335,22 @@ class PluginApi {
...
+   * Example:
+   * ```javascript
+   * api.addHeaderPanel('widget-name', 'widgetVisible', function(attrs, state) { return {}; });
+   * ```
+   * note that 'toggle' is an attribute on the state of the header widget,
+   * and the attrFn receives the current attrs and state of the header as arguments,
+   * and returns a hash of attrs to pass to the widget
+   **/
+   addHeaderPanel(name, toggle, attrFn) {
+     queryRegistry('header').prototype.attachAdditionalPanel(name, toggle, attrFn);
```
eviltrout: Rather than messing with the registry and prototypes, why not make `additionalPanels` a private local variable in the `header.js.es6` file?

Then export the function `attachAdditionalPanel` and just call it in here. That is much more in line with how we usually do things.

Also, `attrFn` is not super clear to me. It seems to me like it's more of a transformer. Maybe call it `transformAttrs` instead, and update the example so it shows why you might want to do that?

gdpelican:
Ok done! @eviltrout

eviltrout:
Looks great now, thanks!

PR #4780, opened 2017-03-27
FEATURE: Send anonymized usage statistics to Discourse if "Discourse Hub" can't reach the site

techAPJ:


lib/discourse_hub.rb
```diff
@@ -4,18 +4,27 @@
...
   end
 
+  def self.stats_fetched_at=(time_with_zone)
+    $redis.set stats_fetched_at_key, time_with_zone.as_json
+  end
 
   private
 
+  def self.get_payload
+    if SiteSetting.send_anonymized_statistics_to_discourse && stats_fetched_at < 7.days.ago
```
techAPJ: @nlalonde is `7.days.ago` fine here or should we increase this to 14 days?

mcwumbly:
Is there an existing discussion about this feature on meta?  I searched briefly but didn't see anything.

lib/discourse_hub.rb
nlalonde: 7 days looks fine. 👍

nlalonde:
I think the default value of the new setting **for existing sites** should be the same as their `share_anonymized_statistics` setting. A migration like this one could do it.

https://github.com/discourse/discourse/blob/master/db/migrate/20170221204204_add_show_subcategory_list_to_categories.rb#L5-L8

lib/discourse_hub.rb
```diff
@@ -56,4 +65,13 @@ def self.referer
...
   end
 
+  def self.stats_fetched_at
+    t = $redis.get(stats_fetched_at_key)
+    t ? Time.zone.parse(t) : 1.year.ago
+  end
+
+  def self.stats_fetched_at_key
+    'stats_fetched_at'
+  end
```
tgxworld: If this value doesn't change, we should use a constant for it.

spec/components/discourse_hub_spec.rb
```diff
@@ -9,4 +9,57 @@
       expect(DiscourseHub.discourse_version_check).to eq(hub_response)
     end
   end
+
+  describe '#version_check_payload' do
```
tgxworld: You probably wanted `.version_check_payload` since `#version_check_payload` is for instance methods.

lib/discourse_hub.rb
```diff
@@ -4,18 +4,27 @@
 module DiscourseHub
 
   def self.version_check_payload
-    {
-      installed_version: Discourse::VERSION::STRING
-    }.merge!( Discourse.git_branch == "unknown" ? {} : {branch: Discourse.git_branch})
+    get_payload.merge!( Discourse.git_branch == "unknown" ? {} : {branch: Discourse.git_branch})
```
tgxworld: Style: Extra ` ` before `Discourse.git_branch`

lib/discourse_hub.rb
```diff
@@ -4,18 +4,27 @@
...
   end
 
   def self.discourse_version_check
     get('/version_check', version_check_payload)
   end
 
+  def self.stats_fetched_at=(time_with_zone)
+    $redis.set stats_fetched_at_key, time_with_zone.as_json
```
tgxworld: Would it be possible to just store the time as `Time.zone.now.to_i`?

lib/discourse_hub.rb
```diff
@@ -4,18 +4,27 @@
...
+  end
 
   private
 
+  def self.get_payload
+    if SiteSetting.send_anonymized_statistics_to_discourse && stats_fetched_at < 7.days.ago
+      About.fetch_cached_stats.symbolize_keys.merge!({ installed_version: Discourse::VERSION::STRING })
+    else
+      { installed_version: Discourse::VERSION::STRING }
```
tgxworld: We can probably move `{ installed_version: Discourse::VERSION::STRING }` into `.version_check_payload` as `default_payload`.

spec/components/discourse_hub_spec.rb
```diff
@@ -9,4 +9,57 @@
...
+      expect(json["post_count"]).not_to be_present
+      expect(json["user_count"]).not_to be_present
+      expect(json["like_count"]).not_to be_present
+      expect(json["likes_7_days"]).not_to be_present
+      expect(json["likes_30_days"]).not_to be_present
+      expect(json["installed_version"]).to be_present
+      expect(json["branch"]).to be_present
+    end
+
+    it 'should not include stats if Discourse Hub has not fetched stats since past year and site setting send_anonymized_statistics_to_discourse is disabled' do
```
tgxworld: This line is getting a little too long :P Generally you want to keep the lines to 80 chars so that we don't have to scroll while reading through code. Perhaps we can break this line down to:

```
describe 'when send_anonymize_stats is disabled' do
   describe 'when Discourse Hub has not fetched stats for the past year' do
     it 'should not..'
   end
end
```

techAPJ:
@tgxworld Updated the PR per your feedback.

PR #4781, opened 2017-03-28
Convert /users/ to /u/

eviltrout:
This is quite a big change to support changing the path from `/users/` to `/u/`.

Server and client side routes for `/users/` paths should still work for backwards compatibility, but going forward all new requests will be to `/u/`.


test/javascripts/lib/url-test.js.es6
```diff
@@ -25,3 +25,9 @@ test("isInternal on subfolder install", function() {
   not(DiscourseURL.isInternal("http://eviltrout.com/tophat"), "a url on the same host but on a different folder is not internal");
   ok(DiscourseURL.isInternal("http://eviltrout.com/forum/moustache"), "a url on the same host and on the same folder is internal");
 });
+
+test("userPath", assert => {
+  assert.equal(userPath(), '/u');
+  assert.equal(userPath('eviltrout'), '/u/eviltrout');
+  assert.equal(userPath('hp.json'), '/u/hp.json');
```
tgxworld: Maybe a test case for subfolder paths as well since `Discourse.getUrl` is extracted into the function?

config/routes.rb
```diff
@@ -304,10 +297,91 @@
   get "signup" => "static#show", id: "signup", constraints: { format: /(json|html)/ }
   get "login-preferences" => "static#show", id: "login", constraints: { format: /(json|html)/ }
 
+  get "my/*path", to: 'users#my_redirect'
+  get "user_preferences" => "users#user_preferences_redirect"
+
+  # New /u/ routes
```
eviltrout: The problem is many routes are `put` or `post`, or have aliases. I'd basically have to build a mini hash structure to account for all of that, which I'm not sure is much better than just copying and pasting with a comment.

config/routes.rb
```diff
@@ -304,10 +297,91 @@
...
+  get "u/:username/activity/pending" => "users#show", constraints: {username: USERNAME_ROUTE_FORMAT}
+  delete "u/:username" => "users#destroy", constraints: {username: USERNAME_ROUTE_FORMAT}
+  # The external_id constraint is to allow periods to be used in the value without becoming part of the format. ie: foo.bar.json
+  get "u/by-external/:external_id" => "users#show", constraints: {external_id: /[^\/]+/}
+  get "u/:username/flagged-posts" => "users#show", constraints: {username: USERNAME_ROUTE_FORMAT}
+  get "u/:username/deleted-posts" => "users#show", constraints: {username: USERNAME_ROUTE_FORMAT}
+
+  get "u/:username/topic-tracking-state" => "users#topic_tracking_state", constraints: {username: USERNAME_ROUTE_FORMAT}
+
+  # Old User routes, will be removed when /u/ is everywhere
```
eviltrout: I thought about redirects but the issue is all client libraries have to respect them. Our jquery ajax stuff should, but what about discourse-api? I didn't want a breaking change.

test/javascripts/lib/url-test.js.es6
eviltrout: Great idea will add that.

spec/models/post_spec.rb
```diff
@@ -952,7 +952,7 @@ def post_with_body(body, user=nil)
 
   it "will unhide the post but will keep the topic invisible/unlisted" do
     hidden_topic = Fabricate(:topic, visible: false)
-    first_post = create_post(topic: hidden_topic)
+    _first_post = create_post(topic: hidden_topic)
```
eviltrout: Yeah I'm not sure why I didn't, it's removed now.

config/routes.rb
tgxworld: I might be missing something here but I was able to get the loop method working 

https://github.com/tgxworld/discourse/commit/bcefc080ac50e35b0e3ebe9460b31f5cd6667df9

I'm not quite sure what the problem `put` or `post` routes would cause. However, if we have routes with the `as: :some_path` options, the loop wouldn't work but based on the routes that we're changing we won't have that problem

config/routes.rb
tgxworld: Ah ok :+1:  I guess we'll have to deprecate `users/` routes and make it known to client libraries somehow?

config/routes.rb
tgxworld: Ahh ok I should have ran the specs first :P I'll have a look

tgxworld: Maybe something like that? https://github.com/tgxworld/discourse/commit/d04f6c2ff7a4d11378ddcd2e29803aea259badcd

eviltrout: Oh I see, that's very nice. Will take it!

eviltrout: Oh actually I spoke too fast. Removing the `as: whatever` is a breaking change because those path helpers are no longer present in the app. Even if our tests pass, we can't count on plugins not having them. I'm going to merge without this for now, but if you want to figure out a solution for the `as` thing in a follow up commit that's cool. (maybe only add `as` if the path is `/u/`?)

PR #4783, opened 2017-03-29
Allow to order search results by the topic creation date

bekicot:
The feature is based on
https://meta.discourse.org/t/allow-to-order-search-results-by-the-topic-creation-date/38544

image

### DIFF
--- a/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
+++ b/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
@@ -11,6 +11,8 @@ const SortOrders = [
   {name: I18n.t('search.latest_post'), id: 1, term: 'order:latest'},
   {name: I18n.t('search.most_liked'), id: 2, term: 'order:likes'},
   {name: I18n.t('search.most_viewed'), id: 3, term: 'order:views'},
+  {name: I18n.t('search.latest_topic'), id: 4, term: 'order:latest_topic'},
+
 ];
 
 export default Ember.Controller.extend({
@@ -90,7 +92,9 @@ export default Ember.Controller.extend({
     this._searchOnSortChange = false;
     if (term) {
       SortOrders.forEach(order => {
-        if (term.indexOf(order.term) > -1){
+        let term_start      = term.indexOf(order.term)
+        let char_after_term = term[term_start + (order.term || '').length]
+        if (term_start > -1 && (!char_after_term || char_after_term.match(/\s/))){
           this.set('sortOrder', order.id);
           term = term.replace(order.term, "");
           term = term.trim();
--- a/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
+++ b/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
@@ -75,14 +75,7 @@ export default Ember.Controller.extend({
 
   @computed('q')
   noSortQ(q) {
-    if (q) {
-      SortOrders.forEach((order) => {
-        if (q.indexOf(order.term) > -1){
-          q = q.replace(order.term, "");
-          q = q.trim();
-        }
-      });
-    }
+    q = this.cleanTerm(q);
     return escapeExpression(q);
   },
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -9,8 +9,9 @@ import { iconHTML } from 'discourse-common/helpers/fa-icon';
 const SortOrders = [
   {name: I18n.t('search.relevance'), id: 0},
   {name: I18n.t('search.latest_post'), id: 1, term: 'order:latest'},
-  {name: I18n.t('search.most_liked'), id: 2, term: 'order:likes'},
-  {name: I18n.t('search.most_viewed'), id: 3, term: 'order:views'},
+  {name: I18n.t('search.latest_topic'), id: 2, term: 'order:latest_topic'},
```
eviltrout: You shouldn't change the `id`s here for previous values for backwards compatibility.

app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -90,7 +91,9 @@ export default Ember.Controller.extend({
     this._searchOnSortChange = false;
     if (term) {
       SortOrders.forEach(order => {
-        if (term.indexOf(order.term) > -1){
+        let term_start      = term.indexOf(order.term)
+        let char_after_term = term[term_start + (order.term || '').length]
+        if (term_start > -1 && (!char_after_term || char_after_term.match(/\s/))){
```
eviltrout: This is confusing to me -- why exactly does this need to change? If there's a good reason for it I can't understand why so I'd suggest a comment explaining why.

bekicot: @eviltrout `term.indexOf(order.term)`, if the `order.term` is `order:latest` and we have another `order:latest_topic`. In such case, if we have `Search query order:latst_topic` it will be replaced to `Search Query _topic`

bekicot: The idea if checking the next char, if the char is whitespace or eol, then, it is fine to replace and set the term, if not, we may have another term that has the same prefix

app/assets/javascripts/discourse/controllers/full-page-search.js.es6
bekicot: @eviltrout Fixed :)

app/assets/javascripts/discourse/controllers/full-page-search.js.es6
eviltrout: Couldn't we just match on the full string by concatenating `order:` and `term`?

bekicot: @eviltrout not sure what does it mean, still don't understand the suggestion you suggested. 

From what I understand from your suggestion, the `term` you meant is the search query?
For `order.term` it already concatenated with the sorting method (`latest`, `likes`, `views, `latest_topic`)

What I discover, the problem is if at least 2 `order.term` share the same prefix `in this case order:latest and order:latest_topic`.

Can you explain more about the concatenation of `order:` and `term`? I may understand it incorrectly.

eviltrout: Ah, oops, I realize the problem now, previously I misunderstood. Sorry about that.

In that case, I think it would be a lot clearer to use a regular expression rather than checking the characters manually like that. `/b` should match a boundary or end of string.

```
const matches = term.match(new RegExp(`${order.term}//b`);
if (matches) {
  // ... 
}
```

bekicot: Cool, thanks for the pro Tips :).  Fixed and pushed.

app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -73,32 +75,31 @@ export default Ember.Controller.extend({
...
+  },
+
+  cleanTerm(term) {
     if (term) {
       SortOrders.forEach(order => {
-        if (term.indexOf(order.term) > -1){
+        let matches = term.match(new RegExp(`${order.term}\\b`));
+        if (matches) {
           this.set('sortOrder', order.id);
           term = term.replace(order.term, "");
```
eviltrout: I think you'll need to replace using the regexp too. Otherwise it could replace the first occurrence which is smaller even if it found a later one.

bekicot: Fixed and pushed :)

eviltrout:
Thanks, looks good now 👍

bekicot:
@eviltrout Thanks for the review, learned a lot.

PR #4785, opened 2017-03-30
Trigger DiscourseEvent when topic category is changed

bekicot:
Required in order to fix
https://meta.discourse.org/t/when-moving-a-solved-topic-to-a-category-that-disallows-solutions-remove-solved/58385

I've been checked the DiscourseEvent to trigger such cases, can't found one.
There is also no trigger for `:topic_changed`.

Added `:topic_category_changed` instead of `:topic_changed` is to limit the scope.


tgxworld:
I believe this should be possible by adding a ActiveRecord callback in the topic model which checks if the `category_id` has changed.

bekicot:
@tgxworld Yes it is, working on that now. ;). Closing the issue as it is not a `real` issue.
:) Thanks.

PR #4786, opened 2017-03-30
Move watching-first-post scss to from user to notification option

jomaxro:
Also remove extra classes from scss, made targeting watching-first-post different from all other notification states


tgxworld:
Thanks @jomaxro

jomaxro:
My pleasure @tgxworld.  This was bothering me when working on some CSS...

jomaxro:
Ooh - @tgxworld it looks like the two file changes were separate commits...is there a way to avoid that in the future? Git isn't my specialty.

tgxworld:
@jomaxro O I didn't notice that too. In the future, you can squash your commits :) 

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

jomaxro:
That looks to be command line based. I always just edit in the browser. Will lool into working on files locally going forward.

PR #4787, opened 2017-03-31
Mobile CSS for deleted topics is inconsistent with desktop CSS

jomaxro:
mobile targets a less specific class (`.deleted`) while desktop targets `.deleted>.topic-body`


ZogStriP:
Thanks 👍

PR #4788, opened 2017-03-31
Add a user_seen DiscourseEvent for plugins to hook into

davidtaylorhq:
This would allow plugins to run logic whenever user activity occurs. I'd like to use this to make my whos_online plugin more responsive to users coming 'online'.

An example of how I plan to use this hook can be found [here](https://github.com/davidtaylorhq/discourse-whos-online/blob/beta/plugin.rb#L53). Clearly it is important that plugins don't run any time-consuming code here, as it will be hit every minute for all active users. I use it to update some redis keys, and then queue a sidekiq job.

ActiveRecord callbacks are not an alternative solution here, since `update_column` is being used for performance reasons.


tgxworld:
Thank you @davidtaylorhq

PR #4790, opened 2017-04-03
FEATURE: Allow admins to schedule a topic to be published in the future.

tgxworld:
https://meta.discourse.org/t/create-topic-in-the-future/11997

### DIFF
--- a/app/assets/javascripts/discourse/controllers/edit-topic-status-update.js.es6
+++ b/app/assets/javascripts/discourse/controllers/edit-topic-status-update.js.es6
@@ -5,18 +5,36 @@ import { popupAjaxError } from 'discourse/lib/ajax-error';
 
 const CLOSE_STATUS_TYPE = 'close';
 const OPEN_STATUS_TYPE = 'open';
+const PUBLISH_TO_CATEGORY_STATUS_TYPE = 'publish_to_category';
 
 export default Ember.Controller.extend(ModalFunctionality, {
+  closeStatusType: CLOSE_STATUS_TYPE,
+  openStatusType: OPEN_STATUS_TYPE,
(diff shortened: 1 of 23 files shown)


app/assets/javascripts/discourse/controllers/edit-topic-status-update.js.es6
```diff
@@ -3,20 +3,46 @@ import ModalFunctionality from 'discourse/mixins/modal-functionality';
...
   loading: false,
   updateTime: null,
   topicStatusUpdate: Ember.computed.alias("model.topic_status_update"),
   selection: Ember.computed.alias('model.topic_status_update.status_type'),
-  autoReopen: Ember.computed.equal('selection', OPEN_STATUS_TYPE),
-  autoClose: Ember.computed.equal('selection', CLOSE_STATUS_TYPE),
-  disableAutoReopen: Ember.computed.and('autoClose', 'updateTime'),
-  disableAutoClose: Ember.computed.and('autoReopen', 'updateTime'),
+
+  @computed('selection', 'openStatusType')
```
eviltrout: This is a bit odd to me -- `openStatusType` used to be a local `const` but now it's a computed property, but aren't we 100% sure the value will never change? 

I think it makes a lot more sense for the CP to be based on `selection` alone and just `=== 
 OPEN_STATUS_TYPE`

eviltrout:
Really nice to see the `Discourse.Category` go away :)

app/assets/javascripts/discourse/controllers/edit-topic-status-update.js.es6
tgxworld: Good point :+1: I actually made `openStatusType` a property so that I can use it in the handlebars template.

PR #4792, opened 2017-04-05
FIX: Poll builder validation was incorrect

xfalcox:
Fixes https://meta.discourse.org/t/cannot-create-single-choice-poll-using-the-gui-editor/59654?u=falco

I went ahead and added a type for the default type, so code is more readable, than having a ghost regular type.


tgxworld:
@xfalcox  Feel free to merge once you've fixed the eslint error. Looks good to me :+1:

PR #4793, opened 2017-04-05
Added an API to ask if an incoming email should be dropped at the SMTP level.

rcgordon:
This lets an SMTP server optionally decide if it should reject a mail without
passing it on to Discourse at all, possibly before even reading the
email's payload, to prevent spam-induced backscatter and save resources.

This just does the bare minimum sanity checking that could prevent obvious
backscatter. For legit errors from legit users, Discourse will still send a
much more pleasant reply email.

### DIFF
--- a/app/controllers/admin/email_controller.rb
+++ b/app/controllers/admin/email_controller.rb
@@ -69,6 +69,19 @@ class Admin::EmailController < Admin::AdminController
     end
   end
 
+  def smtp_should_reject
+    params.require(:from)
+    params.require(:to)
+    # These strings aren't localized; they are sent to an anonymous SMTP user.
+    if User.find_by_email(params[:from]).nil? && !SiteSetting.enable_staged_users
+        render json: { reject: true, reason: "Mail from your address is not accepted. Do you have an account here?" }
+    elsif Email::Receiver.new(params[:from]).check_address(params[:to]).nil?
+        render json: { reject: true, reason: "Mail to this address is not accepted. Check the address and try to send again?" }
+    else
+        render json: { reject: false }
+    end
+  end
+
   def handle_mail
     params.require(:email)
     Email::Processor.process!(params[:email])
--- a/app/controllers/admin/email_controller.rb
+++ b/app/controllers/admin/email_controller.rb
@@ -73,7 +73,7 @@ class Admin::EmailController < Admin::AdminController
     params.require(:from)
     params.require(:to)
     # These strings aren't localized; they are sent to an anonymous SMTP user.
-    if User.find_by_email(params[:from]).nil? && !SiteSetting.enable_staged_users
+    if !User.exists?(email: Email.downcase(params[:from])) && !SiteSetting.enable_staged_users
         render json: { reject: true, reason: "Mail from your address is not accepted. Do you have an account here?" }
     elsif Email::Receiver.check_address(Email.downcase(params[:to])).nil?
         render json: { reject: true, reason: "Mail to this address is not accepted. Check the address and try to send again?" }
--- a/app/controllers/admin/email_controller.rb
+++ b/app/controllers/admin/email_controller.rb
@@ -75,7 +75,7 @@ class Admin::EmailController < Admin::AdminController
     # These strings aren't localized; they are sent to an anonymous SMTP user.
     if User.find_by_email(params[:from]).nil? && !SiteSetting.enable_staged_users
         render json: { reject: true, reason: "Mail from your address is not accepted. Do you have an account here?" }
-    elsif Email::Receiver.new(params[:from]).check_address(params[:to]).nil?
+    elsif Email::Receiver.check_address(Email.downcase(params[:to])).nil?
         render json: { reject: true, reason: "Mail to this address is not accepted. Check the address and try to send again?" }
     else
         render json: { reject: false }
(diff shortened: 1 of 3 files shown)


rcgordon:
This PR is in reference to the discussion here: https://meta.discourse.org/t/reducing-backscatter-in-email-interface/59974

Please note that I've literally never seen Ruby code before today, so if this is disastrously bad, please tell me and I won't be offended.

Note that this needs an SMTP server to use this API to be useful; the mail-receiver Docker container uses Postfix, which needs this to happen in a "before queue" filter, which means plugging in something new that uses either the Milter protocol, or something that implements basic SMTP and then reinjects into a second Postfix SMTP process (yikes). This was beyond the scope of this work, at least for tonight, but testing the new API with Curl, it appears to operate as intended.

app/controllers/admin/email_controller.rb
```diff
@@ -69,6 +69,19 @@ def send_digest
     end
   end
 
+  def smtp_should_reject
+    params.require(:from)
+    params.require(:to)
+    # These strings aren't localized; they are sent to an anonymous SMTP user.
+    if User.find_by_email(params[:from]).nil? && !SiteSetting.enable_staged_users
```
ZogStriP: I usually prefer `User.exists?(email: Email.downcase(params[:from]))` to test if an user exists with that email.

app/controllers/admin/email_controller.rb
```diff
@@ -69,6 +69,19 @@ def send_digest
...
+        render json: { reject: true, reason: "Mail from your address is not accepted. Do you have an account here?" }
+    elsif Email::Receiver.new(params[:from]).check_address(params[:to]).nil?
```
ZogStriP: I think it would be better to promote the `check_address` method to a class method.

`Email::Receiver.check_address(Email.downcase(params[:to])).nil?` is shorter

rcgordon:
Thanks, @ZogStriP, I've incorporated your feedback into the patch!

mpalmer:
It looks like Travis CI had a little sad on an issue unrelated to your PR; if you update your branch against the current upstream master and re-push, I think that'll take care of it.

rcgordon:
Ok, this is rebased against the latest and greatest, to hopefully fix the Travis CI issue. Hopefully I didn't totally break this, git is scary.  :)

app/controllers/admin/email_controller.rb
```diff
@@ -69,6 +69,19 @@ def send_digest
     end
   end
 
+  def smtp_should_reject
+    params.require(:from)
+    params.require(:to)
+    # These strings aren't localized; they are sent to an anonymous SMTP user.
+    if !User.exists?(email: Email.downcase(params[:from])) && !SiteSetting.enable_staged_users
+        render json: { reject: true, reason: "Mail from your address is not accepted. Do you have an account here?" }
```
tgxworld: Minor style nitpick here, we follow the convention of 2 whitespaces when indenting our code.

rcgordon: oh, whoops, yes, sorry about that!

rcgordon:
Hmm, reply-to addresses are being rejected by this patch now, I think when I made Email::Receiver.check_address() into a class method.

I suspect it's this here:
https://github.com/discourse/discourse/pull/4793/commits/a51c191a665c592711232c9f2010522f0af2f55d#diff-0327bfc5fb2dd2a7fd635a096e7fb0daL446

Can someone that knows more Ruby tell me if that regex code broke (does that '@' need to become "@@" or something...?)

mpalmer:
(BTW, TravisCI is happy now, so \o/)

I'm not personally seeing any problems with the newly moved `reply_by_email_address_regex`; instance variables are OK in classes (since classes are regular objects, too, they can have instance variables) and what's being stored in there should be invariant across requests, modulo changes to the relevant site settings.  What's the exact behaviour you're seeing?

rcgordon:
> What's the exact behaviour you're seeing?

I get a reject with a reason of "Mail to this address is not accepted. Check the address and try to send again?" ... which is to say that this call is failing:

https://github.com/discourse/discourse/pull/4793/commits/c51af1333806b9d7e2de165b00b0dca5d8cd5dce#diff-303b662a564c972b4fea99cdaf710825L78

This call does not fail for category addresses, just reply-to addresses.

I'll stick some logging in there tomorrow and see what happened.

rcgordon:
> This call does not fail for category addresses, just reply-to addresses.

Disregard this, user error; the '+' in the reply-to address needs to be URL-encoded for the API call, of course. So "to=replies+kdsjfkdsjfksd@example.com" had to be "to=replies%2bkdsjfkdsjfksd@example.com" in my Curl command line.

So this PR is correct as-is, and ready to merge, if everyone is happy with it. There will be a separate PR incoming shortly for the mail-receiver container to hook up the SMTP server's side of things.

rcgordon:
Ok, there are PRs for the Docker container changes needed to make use of this new API:

- mail-receiver container: https://github.com/discourse/mail-receiver/pull/2
- main Discourse container: https://github.com/discourse/discourse_docker/pull/344

These aren't strictly necessary, beyond stopping backscatter when using the mail-receiver container; people running their own SMTP servers just need this PR, of course.

ZogStriP:
Thanks 👍

PR #4795, opened 2017-04-06
FIX: locale typo watched_tags_instructions

thorbenegberts:
Fixed Typo "imit" => "mit"


ZogStriP:
Thanks but all translation changes must be done via [transifex](transifex.com/discourse).

thorbenegberts:
Ah, okay. I'll look into it in the next few days.

PR #4803, opened 2017-04-12
Wrap directory names in quotes

ckeboss:
When there are spaces in any of the directories referenced, bash will error out. This wraps those items in quotes to allow bash to parse the path names properly.


tgxworld:
LGTM :+1:

PR #4804, opened 2017-04-12
Convert non-latin characters to latin in URL

voogarix:
Hi. First of all thanks for greater project. I have installed discourde, everything is ok. There is 1 problem - when i open ne theme in my language (azerbaijani) special character ə Ə doesnt convert to latin alternative ( e or E). İ hope it will be fixed on next release )).


gschlager:
@voogarix You will need to add transliteration rules like these: https://github.com/discourse/discourse/blob/master/config/locales/transliterate.ru.yml

Or you change the way slugs are generated:
https://meta.discourse.org/t/slug-for-azerbaijan-language/55355
https://meta.discourse.org/t/removing-the-concept-of-slugs-for-some-languages/26643/16

ZogStriP:
@voogarix can you make another PR with just your changes?

PR #4807, opened 2017-04-13
FIX: confirm email token for user created via social login

techAPJ:


coding-horror:
what is this in reference to?

techAPJ:
@coding-horror: https://meta.discourse.org/t/auto-add-to-group/60764?u=techapj

Currently users who login via social login method are not getting added to group automatically based on their email domain because their email token is not getting confirmed.

coding-horror:
ok I was just uncertain what this was :+1:

ZogStriP:
There probably shoud be a test here since it's a nasty one ;)

PR #4810, opened 2017-04-14
Fix bench script 

yjukaku:
I tried to clone Discourse so that I could work on some perf patches, but I couldn't get the bench script to work without a few tiny fixes. 

- 'facter' gem requires 'CFPropertyList` gem, but it's not installed automatically
- the `run` method called `Kernel::exit`, which silently exits instead of telling you why. Changed to `Kernel::abort`
- Email addresses created in profile_db_generator were invalid according to the User email validation regex. (Didn't allow `admin@localhost1` as a valid email)

### DIFF
--- a/script/bench.rb
+++ b/script/bench.rb
@@ -49,13 +49,15 @@ def run(command, opt = nil)
       system(command, out: $stdout, err: :out)
     end
 
-  exit unless exit_status
+  abort("Command '#{command}' failed with exit status #{$?}") unless exit_status
 end
 
 begin
   require 'facter'
 rescue LoadError
   run "gem install facter"
+  # Facter requires CFPropertyList, but doesn't install it.
+  run "gem install CFPropertyList"
   puts "please rerun script"
   exit
 end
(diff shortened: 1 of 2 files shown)


script/bench.rb
```diff
@@ -49,13 +49,15 @@ def run(command, opt = nil)
...
 
-  exit unless exit_status
+  abort("Command '#{command}' failed with exit status #{$?}") unless exit_status
 end
 
 begin
   require 'facter'
 rescue LoadError
   run "gem install facter"
+  # Facter requires CFPropertyList, but doesn't install it.
```
tgxworld: Could you point me to the source of this dependency? I don't have this Gem installed on my machine and I'm still able to install the `facter` gem

script/profile_db_generator.rb
```diff
@@ -44,7 +44,7 @@ def sentence
 
 def create_admin(seq)
   User.new.tap { |admin|
-    admin.email = "admin@localhost#{seq}"
+    admin.email = "admin@localhost#{seq}.fake"
```
tgxworld: :+1:

tgxworld:
Thanks for your contribution @yjukaku :+1:

script/bench.rb
yjukaku: The facter gem in 2.x fails when calling Facter.to_hash, because it `require`s 'cfpropertylist'. 

```
irb(main):001:0> require 'facter'
=> true
irb(main):002:0> Facter.to_hash
Error loading fact /Users/me/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/facter-2.4.6/lib/facter/ec2.rb: cannot load such file -- cfpropertylist
Error loading fact /Users/me/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/facter-2.4.6/lib/facter/macosx.rb: cannot load such file -- cfpropertylist
Error loading fact /Users/me/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/facter-2.4.6/lib/facter/processor.rb: cannot load such file -- cfpropertylist
LoadError: cannot load such file -- cfpropertylist
```

It looks like master on `facter` is a rewrite using C++ with ruby bindings, but on the 2.x branch, you can see the require here:

https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/macosx.rb#L9

tgxworld:
@yjukaku  Thanks for fixing :+1:

PR #4816, opened 2017-04-15
FIX: increase host_must_be_valid regex for .technology TLD

jloh:
The `.technology` is 10 chars long and wasn't being matched previously. This you couldn't add it  as an embeddable host for comments.


eviltrout:
thanks!

PR #4818, opened 2017-04-17
The "Groups" section's heading on user admin page was visible to mods

OsamaSayegh:
Has been bothering me for a while so I thought I'd submit a PR :)

Here is a before and after:
Before:

before

After:

after


tgxworld:
Thanks @OsamaSayegh :+1:

PR #4819, opened 2017-04-18
FIX: User website allows new TLDs

xfalcox:


techAPJ:
Looks good to me! Merge at will.

PR #4825, opened 2017-04-20
Let's try to bisect which commit is causing Travis to fail

tgxworld:


tgxworld:
@SamSaffron This commit that I reverted here broke JS tests on Travis for some reason. Maybe you can have a look since I'm not very familiar with what is being changed here.

PR #4826, opened 2017-04-20
FIX: all basic integer settings should have min & max value validation

techAPJ:
https://github.com/discourse/discourse/commit/74a46dce68e828876431a3629f9337ab74f46b0a#commitcomment-20638987

### DIFF
--- a/spec/components/validators/integer_setting_validator_spec.rb
+++ b/spec/components/validators/integer_setting_validator_spec.rb
@@ -21,9 +21,21 @@ describe IntegerSettingValidator do
 
       it "returns true if value is a valid integer" do
         expect(validator.valid_value?(1)).to eq(true)
-        expect(validator.valid_value?(-1)).to eq(true)
         expect(validator.valid_value?('1')).to eq(true)
-        expect(validator.valid_value?('-1')).to eq(true)
+      end
+
+      it "defaults min to 0" do
+        expect(validator.valid_value?(-1)).to eq(false)
+        expect(validator.valid_value?('-1')).to eq(false)
+        expect(validator.valid_value?(0)).to eq(true)
+        expect(validator.valid_value?('0')).to eq(true)
+      end
+
+      it "defaults max to 20000" do
+        expect(validator.valid_value?(20001)).to eq(false)
+        expect(validator.valid_value?('20001')).to eq(false)
+        expect(validator.valid_value?(20000)).to eq(true)
+        expect(validator.valid_value?('20000')).to eq(true)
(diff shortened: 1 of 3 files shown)


ZogStriP:
I would have changed the `IntegerSettingValidator` and set default values there. The yml file should only contain non-default values.

techAPJ:
@ZogStriP Good idea! I will make the changes accordingly.

Just to confirm: if the `max` value of integer setting isn't specified we should default it to `20000`, right?

ZogStriP:
> Just to confirm: if the max value of integer setting isn't specified we should default it to 20000, right?

Whatever you feel is a right "default" value for most "max" values. We can change it later if needs be.

techAPJ:
@ZogStriP Okay, I made the required changes. Can you review this PR again? (not urgent)

spec/components/validators/integer_setting_validator_spec.rb
```diff
@@ -21,9 +21,21 @@
 
       it "returns true if value is a valid integer" do
         expect(validator.valid_value?(1)).to eq(true)
-        expect(validator.valid_value?(-1)).to eq(true)
         expect(validator.valid_value?('1')).to eq(true)
-        expect(validator.valid_value?('-1')).to eq(true)
+      end
+
+      it "defaults min to 0" do
```
tgxworld: This description is alittle odd for me because the test cases aren't really testing if the min is "0". The default min could be 5 and the test cases would still pass.

techAPJ: If the default min would be `5`, I would have replaced:

```
expect(validator.valid_value?(-1)).to eq(false)
```

with

```
expect(validator.valid_value?(4)).to eq(false)
```

> This description is alittle odd for me

What should be the description here?

techAPJ: I just replaced:

```
expect(validator.valid_value?(100)).to eq(true)
```

with

```
expect(validator.valid_value?(0)).to eq(true)
```

So perhaps that makes it a bit clear? Also, changing min to `5` will result in failing tests now. 😄

tgxworld: > Also, changing min to 5 will result in failing tests now.

:+1: This was what I was going for. The test wasn't actually testing that the default is `0` so the description felt odd to me.

techAPJ: Got it. Thanks for review. 👍

PR #4828, opened 2017-04-21
FEATURE: Add normal as a preference for topic subscription state when replying to a topic

cpradio:
Relevant Discussion
https://github.com/discourse/discourse/commit/4dc4c5bebcbc37c3ad38aad12ad5f00b95377471

Notes:
1. Tracking is still the default when the user preference is not set to anything
2. This still abides by the requested subscription must be higher than what is currently set and what is currently set must be higher than normal/regular. (see https://meta.discourse.org/t/change-category-tracking-subscription-to-watching-if-i-post-a-reply/60958/25?u=cpradio)
3. Admin Setting can be default to Normal as well now

Other Notes:
1. If the user's setting of `Automatically track topics I enter` is set to anything other than Never, and that time allotment passes, having the setting `When I post in a topic, set that topic to` Normal will do nothing, as the topic state was already set to Tracking.


coding-horror:
👍

PR #4829, opened 2017-04-23
FEATURE: login by a link from email

fantasticfears:
https://meta.discourse.org/t/passwordless-signin-via-email-link/57569

### DIFF
--- a/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
+++ b/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
@@ -37,14 +37,14 @@ export default Ember.Controller.extend(ModalFunctionality, {
         if (data.user_found === true) {
           key += '_found';
           this.set('accountEmailOrUsername', '');
-          this.set('offerHelp', I18n.t(key, {email: escaped, username: escaped}));
+          this.set('offerHelp', I18n.t(key, { email: escaped, username: escaped }));
         } else {
           if (data.user_found === false) {
             key += '_not_found';
             extraClass = 'error';
           }
 
-          this.flash(I18n.t(key, {email: escaped, username: escaped}), extraClass);
+          this.flash(I18n.t(key, { email: escaped, username: escaped }), extraClass);
         }
       }).catch(e => {
         this.flash(extractError(e), 'error');
@@ -88,7 +88,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
             extraClass = 'error';
           }
 
-          this.flash(I18n.t(key, {email: escaped, username: escaped}), extraClass);
+          this.flash(I18n.t(key, { email: escaped, username: escaped }), extraClass);
         }
       }).catch(e => {
         this.flash(extractError(e), 'error');
--- a/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
+++ b/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
@@ -80,7 +80,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
         if (data.user_found === true) {
           key += '_found';
           this.set('accountEmailOrUsername', '');
-          bootbox.alert(I18n.t(key, {email: escaped, username: escaped}));
+          bootbox.alert(I18n.t(key, { email: escaped, username: escaped }));
           this.send("closeModal");
         } else {
           if (data.user_found === false) {
--- a/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
+++ b/app/assets/javascripts/discourse/controllers/forgot-password.js.es6
@@ -61,6 +61,42 @@ export default Ember.Controller.extend(ModalFunctionality, {
 
     help() {
       this.setProperties({ offerHelp: I18n.t('forgot_password.help'), helpSeen: true });
+    },
+
+    emailLogin() {
+      if (this.get('submitDisabled')) return false;
+
+      this.set('disabled', true);
+
+      ajax('/u/email-login', {
+        data: { login: this.get('accountEmailOrUsername').trim() },
+        type: 'POST'
+      }).then(data => {
+        const escaped = escapeExpression(this.get('accountEmailOrUsername'));
+        const isEmail = this.get('accountEmailOrUsername').match(/@/);
+        let key = 'email_login.complete_' + (isEmail ? 'email' : 'username');
+        let extraClass;
+
+        if (data.user_found === true) {
+          key += '_found';
+          this.set('accountEmailOrUsername', '');
+          bootbox.alert(I18n.t(key, {email: escaped, username: escaped}));
+          this.send("closeModal");
+        } else {
+          if (data.user_found === false) {
+            key += '_not_found';
+            extraClass = 'error';
+          }
+
+          this.flash(I18n.t(key, {email: escaped, username: escaped}), extraClass);
+        }
+      }).catch(e => {
+        this.flash(extractError(e), 'error');
+      }).finally(() => {
+        setTimeout(() => this.set('disabled', false), 1000);
+      });
+
+      return false;
     }
   }
 
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -9,8 +9,8 @@ class SessionController < ApplicationController
 
   before_action :check_local_login_allowed, only: %i(create forgot_password)
   skip_before_action :redirect_to_login_if_required
-  skip_before_action :preload_json, :check_xhr, only: %i(sso sso_login become sso_provider destroy email_login_page email_login)
-  skip_before_action :check_xhr, only: %i(email_login_page)
+  skip_before_action :preload_json, only: %i(sso sso_login sso_provider become destroy email_login)
+  skip_before_action :check_xhr, only: %i(sso sso_login sso_provider become destroy email_login_page)
 
   ACTIVATE_USER_KEY = "activate_user"
 
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -7,7 +7,7 @@ class SessionController < ApplicationController
     render body: nil, status: 500
   end
 
-  before_action :check_local_login_allowed, only: %i(create forgot_password)
+  before_action :check_local_login_allowed, only: %i(create forgot_password email_login email_login_page)
   skip_before_action :redirect_to_login_if_required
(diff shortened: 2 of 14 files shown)


app/controllers/session_controller.rb
```diff
@@ -4,7 +4,8 @@
 class SessionController < ApplicationController
 
   skip_before_filter :redirect_to_login_if_required
-  skip_before_filter :preload_json, :check_xhr, only: ['sso', 'sso_login', 'become', 'sso_provider', 'destroy']
+  skip_before_filter :preload_json, :check_xhr, only: ['sso', 'sso_login', 'become', 'sso_provider', 'destroy', 'email_login_page', 'email_login']
+  skip_before_filter :check_xhr, only: ['email_login_page']
```
techAPJ: Two points to note here:

1) `skip_before_filter :check_xhr` is getting defined twice above.
2) we should not skip `check_xhr` for `email_login`

I believe we should replace above with:

```
skip_before_filter :preload_json, only: ['sso', 'sso_login', 'become', 'sso_provider', 'destroy', 'email_login_page', 'email_login']
skip_before_filter :check_xhr, only: ['sso', 'sso_login', 'become', 'sso_provider', 'destroy', 'email_login_page']
```

app/controllers/users_controller.rb
```diff
@@ -8,7 +8,7 @@
 class UsersController < ApplicationController
 
   skip_before_filter :authorize_mini_profiler, only: [:avatar]
-  skip_before_filter :check_xhr, only: [:show, :password_reset, :update, :account_created, :activate_account, :perform_account_activation, :user_preferences_redirect, :avatar, :my_redirect, :toggle_anon, :admin_login, :confirm_admin]
+  skip_before_filter :check_xhr, only: [:show, :password_reset, :email_login, :update, :account_created, :activate_account, :perform_account_activation, :user_preferences_redirect, :avatar, :my_redirect, :toggle_anon, :admin_login, :confirm_admin]
```
techAPJ: We should not skip `check_xhr` for `email_login` here as well since I believe we are accessing this method via AJAX request.

app/controllers/session_controller.rb
techAPJ: Edited my above comment to add one more point.

fantasticfears: Thanks! I've amend the change to the PR.

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
```diff
@@ -52,6 +52,42 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+      }).then(data => {
+        const escaped = escapeExpression(this.get('accountEmailOrUsername'));
+        const isEmail = this.get('accountEmailOrUsername').match(/@/);
+        let key = 'email_login.complete_' + (isEmail ? 'email' : 'username');
+        let extraClass;
+
+        if (data.user_found === true) {
+          key += '_found';
+          this.set('accountEmailOrUsername', '');
+          bootbox.alert(I18n.t(key, {email: escaped, username: escaped}));
```
tgxworld: Style nitpick: Missing ` ` after/before `{` and `}`

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
```diff
@@ -52,6 +52,42 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+          this.send("closeModal");
+        } else {
+          if (data.user_found === false) {
+            key += '_not_found';
+            extraClass = 'error';
+          }
+
+          this.flash(I18n.t(key, {email: escaped, username: escaped}), extraClass);
```
tgxworld: Style nitpick: Missing  after/before { and }

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
```diff
@@ -52,6 +52,42 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+        }
+      }).catch(e => {
+        this.flash(extractError(e), 'error');
```
tgxworld: would `popupAjaxError` work here?

app/controllers/users_controller.rb
```diff
@@ -521,6 +522,36 @@ def admin_login
...
+
+    json = { result: "ok" }
+    unless SiteSetting.forgot_password_strict
+      json[:user_found] = user_presence
+    end
+
+    render json: json
+
+  rescue RateLimiter::LimitExceeded
+    render_json_error(I18n.t("rate_limiter.slow_down"))
```
tgxworld: Since we're using the default rate limiter message here. We can just let [application_controller](https://github.com/discourse/discourse/blob/master/app/controllers/application_controller.rb#L101-L103) rescue the error.

app/mailers/user_notifications.rb
```diff
@@ -53,6 +53,13 @@ def forgot_password(user, opts={})
                 email_token: opts[:email_token])
   end
 
+  def email_login(user, opts={})
+    build_email(user.email,
+                template: "user_notifications.email_login",
+                locale: user_locale(user),
+                email_token: opts[:email_token])
```
tgxworld: Not directly related to the PR but I think it is  a good time to refactor some of the methods in this `Mailer`. The only difference between alot of the method is the `template` option. Perhaps we can just extract the duplication into a private method?

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
```diff
@@ -52,6 +52,42 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
       return false;
+    },
+
+    emailLogin() {
+      if (this.get('submitDisabled')) return false;
+
+      this.set('disabled', true);
+
+      ajax('/u/email-login', {
+        data: { login: this.get('accountEmailOrUsername').trim() },
```
tgxworld: I think we should only accept email here. Allowing the login link to be triggered with just the username may lead to abuse of the feature.

coding-horror:
I worry about security concerns here, so I think we need to vet this one especially thoroughly...

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
fantasticfears: It's a modal so the error is placed on the top of it. `popupAjaxError` takes the error from `extractError` and shows as a dialog.

app/controllers/users_controller.rb
fantasticfears: It's however not default :( perhaps it's better not to expose how much time the user should wait. [LimitExceed](https://github.com/discourse/discourse/blob/master/lib/rate_limiter/limit_exceeded.rb) extracts the error message based on required interval and `type` string.

app/mailers/user_notifications.rb
fantasticfears: I extracted 6 method by leveraging `.build_email_with_email_token`

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
fantasticfears: Resetting password process goes down the same path, it sends the very same token to user.  I am not sure the process. The dialog may need some polish here?

app/assets/javascripts/discourse/controllers/forgot-password.js.es6
```diff
@@ -35,15 +35,51 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
           if (data.user_found === false) {
             key += '_not_found';
             extraClass = 'error';
           }
 
-          this.flash(I18n.t(key, {email: escaped, username: escaped}), extraClass);
+          this.flash(I18n.t(key, { email: escaped, username: escaped }), extraClass);
+        }
+      }).catch(e => {
+        this.flash(extractError(e), 'error');
```
tgxworld: can we use `popupAjaxError` here?

app/mailers/user_notifications.rb
```diff
@@ -10,11 +10,15 @@ class UserNotifications < ActionMailer::Base
...
-                locale: user_locale(user),
-                email_token: opts[:email_token])
+  %i(signup
+     confirm_old_email
+     confirm_new_email
+     email_login
+     admin_login
+     account_created).each do |mail_name|
+    define_method(mail_name) do |user, opts={}|
+      build_email_with_email_token(user, "user_notifications.#{mail_name}", opts[:email_token])
```
tgxworld: I'll prefer if we avoid using meta-programming techniques here. Extracting the common logic into a private method is a good enough refactor.

fantasticfears:
Done. There seems a network failure in Travis.

tgxworld:
Thanks @fantasticfears  I'll review this again after we release 1.8 at the end of this month.

app/controllers/session_controller.rb
```diff
@@ -231,6 +232,49 @@ def create
     (user.active && user.email_confirmed?) ? login(user) : not_activated(user)
   end
 
+  def email_login_page
```
tgxworld: Hi @fantasticfears , I'm refactoring this and reviewing this and was wondering why we need a page for this? Was there a reason why it wasn't handled like the admin login?

fantasticfears: This seems clear as we don't need to scan the code to understand different path.

fantasticfears: I am going to do a rebase. Do you think we should merge this into `email_login` method?

SamSaffron:
@fantasticfears  can you review this and rebase etc... I don't want this to be lost, we are in a reasonable position to merge it in.

fantasticfears:
Will do. I feel the test suite is a sand castle in my hand when build this...

fantasticfears:
rebased

SamSaffron:
image


We can't have 2 primary buttons, only reset password should be primary, the other button should be normal. 

Except for that this looks good to me, @tgxworld  can you merge.

SamSaffron:
This is conflicting it needs a rebase.

fantasticfears:
rebased

coding-horror:
This will need to be deferred to Discourse 2.0 I think

SamSaffron:
@tgxworld I would really like this done, the biggest missing piece here is that we need it to start off behind a feature flag default disabled. then we can feel a bit safer about merging it.

tgxworld:
Closing in favor of https://github.com/discourse/discourse/pull/5579/files

PR #4833, opened 2017-04-25
FIX: make TopicEmbed trashable

techAPJ:
Requested here: https://meta.discourse.org/t/a-posts-topicembed-isnt-destroyed-when-the-post-is-deleted/60904?u=techapj

This PR allows `TopicEmbed` to be trashed when topic is deleted. Essentially four things is added in this PR:

1) If the topic is trashed, trash associated topic_embed record as well. Thus allowing new topic_embed record to be created for same `embed_url` (more info about this in point 3).

2) If the topic is recovered, recover associated topic_embed record as well.

3) If a new topic_embed record is created that has same `embed_url` which was trashed earlier, then that old (trashed) record will be permanently destroyed thus preventing accidental recovery of that old (trashed) record.

4) When the topic is permanently destroyed, destroy associated topic_embed record as well.

@eviltrout can you please review this PR? Thanks!

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -40,12 +40,16 @@ class Topic < ActiveRecord::Base
     update_category_topic_count_by(-1) if deleted_at.nil?
     super(trashed_by)
     update_flagged_posts_count
+    self.topic_embed.trash! if has_topic_embed?
   end
 
   def recover!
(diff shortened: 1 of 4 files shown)


eviltrout:
Great job, merge when ready!

techAPJ:
Thanks @eviltrout ❤️

PR #4836, opened 2017-04-26
FIX: Clean up unused staged users job not completing.

tgxworld:
@ZogStriP  Can you review this? Thank you!

### DIFF
--- a/app/jobs/scheduled/clean_up_unused_staged_users.rb
+++ b/app/jobs/scheduled/clean_up_unused_staged_users.rb
@@ -6,11 +6,21 @@ module Jobs
     def execute(args)
       destroyer = UserDestroyer.new(Discourse.system_user)
 
-      User.joins(:user_stat)
-          .where(staged: true)
-          .where("users.created_at < ?", 1.year.ago)
-          .where("user_stats.post_count = 0")
-          .find_each { |user| destroyer.destroy(user) }
+      User.joins("LEFT JOIN posts ON posts.user_id = users.id")
+        .where("posts.user_id IS NULL")
+        .where(staged: true)
+        .where("users.created_at < ?", 1.year.ago)
+        .find_each do |user|
+
+        begin
+          destroyer.destroy(user)
+        rescue => e
+          Discourse.handle_job_exception(e,
+            message: "Cleaning up unused staged user",
+            extra: { user_id: user.id }
+          )
+        end
+      end
     end
 
   end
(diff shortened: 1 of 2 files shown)


app/jobs/scheduled/clean_up_unused_staged_users.rb
```diff
@@ -6,11 +6,21 @@ class CleanUpUnusedStagedUsers < Jobs::Scheduled
     def execute(args)
       destroyer = UserDestroyer.new(Discourse.system_user)
 
-      User.joins(:user_stat)
-          .where(staged: true)
-          .where("users.created_at < ?", 1.year.ago)
-          .where("user_stats.post_count = 0")
```
tgxworld: I think we need to add a ensure consistency job for `user_stats.post_count` because there are users that have a `user_stats.post_count` of `0` but `user.posts.count` shows otherwise.

app/jobs/scheduled/clean_up_unused_staged_users.rb
```diff
@@ -6,11 +6,21 @@ class CleanUpUnusedStagedUsers < Jobs::Scheduled
...
-          .find_each { |user| destroyer.destroy(user) }
+      User.joins("LEFT JOIN posts ON posts.user_id = users.id")
+        .where("posts.user_id IS NULL")
+        .where(staged: true)
+        .where("users.created_at < ?", 1.year.ago)
+        .find_each do |user|
+
+        begin
+          destroyer.destroy(user)
+        rescue => e
```
tgxworld: `PostDestroyer` would raise an exception if certain conditions are not met. Therefore, we don't want to have the entire scheduled job blow up just because we are not able to destroy a user.

ZogStriP: Good point, probably want to add that comment in the file itself ;)

spec/jobs/clean_up_unused_staged_users.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe Jobs::CleanUpUnusedStagedUsers do
+  let(:user) { Fabricate(:user) }
+  let(:staged_user) { Fabricate(:user, staged: true) }
+
+  context 'when staged user is unused' do
+    it 'should clean up the staged user' do
+      user
```
ZogStriP: What are you calling `user` here? Can't you use `let!(:user) { Fabricate(:user) }`?

PR #4837, opened 2017-04-26
FEATURE: private_email site setting to prevent data going out in email

eviltrout:


config/site_settings.yml
```diff
@@ -703,6 +703,7 @@ email:
     default: "smime.p7s|signature.asc"
   enable_forwarded_emails: false
   always_show_trimmed_content: false
+  secure_email: false
```
nlalonde: I wonder if "Security" is where people would expect to find this setting... I guess it's more about privacy, not security. So "Email" is probably best.

ZogStriP:
LGTM 👍

PR #4838, opened 2017-04-26
FIX: 'read' filter in top menu showing new topics

vinothkannans:
https://meta.discourse.org/t/read-filter-in-top-menu-showing-new-topics/61654

### DIFF
--- a/lib/topic_query.rb
+++ b/lib/topic_query.rb
@@ -165,7 +165,7 @@ class TopicQuery
 
   def list_read
     create_list(:read, unordered: true) do |topics|
-      topics.order('COALESCE(tu.last_visited_at, topics.bumped_at) DESC')
+      topics.where('tu.last_visited_at IS NOT NULL').order('tu.last_visited_at DESC')
     end
   end
 
--- a/spec/components/topic_query_spec.rb
+++ b/spec/components/topic_query_spec.rb
@@ -458,6 +458,8 @@ describe TopicQuery do
 
       context 'list_unread' do
         it 'lists topics correctly' do
+          new_topic = Fabricate(:post, user: creator).topic
+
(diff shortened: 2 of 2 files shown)


lib/topic_query.rb
```diff
@@ -165,7 +165,8 @@ def list_latest
 
   def list_read
     create_list(:read, unordered: true) do |topics|
-      topics.order('COALESCE(tu.last_visited_at, topics.bumped_at) DESC')
+      topics.where('tu.last_visited_at IS NOT NULL')
+        .order('COALESCE(tu.last_visited_at, topics.bumped_at) DESC')
```
eviltrout: If `tu.last_visited_at IS NOT NULL` then the `COALESCE` here isn't doing anything and can be removed.

I would also add a spec to make sure it works properly.

vinothkannans: But I can see the change in result. Normally it includes new topics. After i added the `where` condition it showing correctly.

Anyway adding spec will give best solution for this 👍

vinothkannans: Sorry. I misunderstood. Now only I read about what is `COALESCE`. I will update this PR soon.

eviltrout:
Thanks for removing the `COALESCE` - it still needs a test though.

vinothkannans:
@eviltrout I am trying to add. Not yet found any example. I will add it asap :+1:.

eviltrout:
It looks like this is the existing spec for `list_read`:

https://github.com/discourse/discourse/blob/master/spec/components/topic_query_spec.rb#L459-L464

Perhaps you could just add a new topic in the `it` block and then make sure it's not included?

vinothkannans:
@eviltrout it's done :+1:

eviltrout:
Thanks, looks great!

PR #4839, opened 2017-04-26
FEATURE: provide more details when performing a bulk add to group

techAPJ:
https://meta.discourse.org/t/provide-more-details-when-performing-a-bulk-add-to-group/60665?u=techapj

This PR adds more information on success page.

screen shot 2017-04-27 at 01 25 44

### DIFF
--- a/app/controllers/admin/groups_controller.rb
+++ b/app/controllers/admin/groups_controller.rb
@@ -24,13 +24,22 @@ class Admin::GroupsController < Admin::AdminController
 
   def bulk_perform
     group = Group.find(params[:group_id].to_i)
+    users_added = 0
     if group.present?
       users = (params[:users] || []).map {|u| u.downcase}
-      user_ids = User.where("username_lower in (:users) OR email IN (:users)", users: users).pluck(:id)
-      group.bulk_add(user_ids) if user_ids.present?
+      valid_emails = valid_usernames = {}
+      valid_users = User.where("username_lower IN (:users) OR email IN (:users)", users: users).pluck(:id, :username_lower, :email)
+      valid_users.each do |vu|
+        valid_emails[vu[1]] = valid_usernames[vu[2]] = vu[0]
+        vu.slice!(1..2)
+      end
+      invalid_users = users.reject! { |u| valid_emails[u] || valid_usernames[u] }
+      valid_users.flatten!
+      group.bulk_add(valid_users) if valid_users.present?
+      users_added = valid_users.count
     end
 
-    render json: success_json
+    render json: { success: true, message: I18n.t('groups.success.bulk_add', users_added: users_added), users_not_added: invalid_users }
   end
 
   def create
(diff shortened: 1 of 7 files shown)


app/controllers/admin/groups_controller.rb
```diff
@@ -24,13 +24,25 @@ def bulk
...
+      end
+      invalid_users = users.reject! { |u| valid_emails[u] || valid_usernames[u] }
+      valid_users.flatten!
+      group.bulk_add(valid_users) if valid_users.present?
+      users_added = valid_users.count
     end
 
-    render json: success_json
+    message = I18n.t('groups.success.bulk_add', users_added: users_added)
+    message = "#{message} <br><br> #{I18n.t('groups.success.bulk_add_invalid_users', invalid_users: invalid_users.join(', '))}" unless invalid_users.blank?
```
eviltrout: I am not a huge fan of building HTML on the server side like this.

Why not return the invalid users as an array that you could display in the hbs template by using an #each?

techAPJ: Good point. Made it so.

app/controllers/admin/groups_controller.rb
```diff
@@ -24,13 +24,22 @@ def bulk
 
   def bulk_perform
     group = Group.find(params[:group_id].to_i)
+    users_added = 0
     if group.present?
       users = (params[:users] || []).map {|u| u.downcase}
-      user_ids = User.where("username_lower in (:users) OR email IN (:users)", users: users).pluck(:id)
-      group.bulk_add(user_ids) if user_ids.present?
+      valid_emails = valid_usernames = {}
```
tgxworld: Both of these variables are actually referring to the same hash so I'm not sure if there is what you intended?

app/controllers/admin/groups_controller.rb
```diff
@@ -24,13 +24,22 @@ def bulk
...
+      valid_users = User.where("username_lower IN (:users) OR email IN (:users)", users: users).pluck(:id, :username_lower, :email)
+      valid_users.each do |vu|
```
techAPJ: I am modifying the array as well via `vu.slice!(1..2)`.. (this is done so as to pass IDs only to `bulk_add` method).

app/controllers/admin/groups_controller.rb
techAPJ: Oops, yeah. Fixed now.

app/controllers/admin/groups_controller.rb
tgxworld: Ah in that case, you can do the following which I feel is much easier to understand and read.

```
valid_users.map! do |id, username_lower, email| 
  valid_emails[email] =  valid_usernames[username_lower] = id
  id
```

techAPJ: Awesome, thanks for the tip! Done via https://github.com/discourse/discourse/commit/e3f82140d8fa7f7e01378ccd5dcad57a96f2909b.

PR #4840, opened 2017-04-26
FEATURE: phase 1 of supporting multiple email addresses

LeoMcA:
https://meta.discourse.org/t/additional-email-support/59847?u=leomca

@SamSaffron I hope this doesn't make you cry too much


SamSaffron:
overall this looks good to me, need to review it carefully, but not noticing anything particularly dangerous about the change, just hoping we caught all the edge cases :)

db/migrate/20170329064703_create_user_emails.rb
```diff
@@ -0,0 +1,26 @@
+class CreateUserEmails < ActiveRecord::Migration
+  def up
+    create_table :user_emails do |t|
+      t.references :user
```
SamSaffron: we never really use this in migrations, so I am not sure I want to start now... t.integer :user_id + explicit index creation should do the trick.

app/controllers/users_controller.rb
```diff
@@ -311,7 +311,7 @@ def create
       return fail_with("login.reserved_username")
     end
 
-    if user = User.where(staged: true).find_by(email: params[:email].strip.downcase)
+    if user = User.where(staged: true).find_by_email(params[:email].strip.downcase)
```
tgxworld: Usually, it is no recommended to make cosmetic changes like these as it pollutes the commit history. Also, `find_by_*` have been [deprecated in Rails 4](http://guides.rubyonrails.org/4_0_release_notes.html) and will be removed in Rails 5.

app/models/user.rb
```diff
@@ -233,7 +234,8 @@ def self.find_by_username_or_email(username_or_email)
   end
 
   def self.find_by_email(email)
-    find_by(email: Email.downcase(email))
+    user_email = UserEmail.find_by(email: Email.downcase(email))
+    user_email.user if user_email
```
tgxworld: In Ruby 2.3, you can do something like this, `UserEmail.find_by(email: Email.downcase(email))&.user`

app/models/user.rb
```diff
@@ -912,6 +914,23 @@ def logged_out
     DiscourseEvent.trigger(:user_logged_out, self)
   end
 
+  def email
+    primary_email.email if primary_email
```
tgxworld: `primary_email&.email` :wink:

app/models/user.rb
```diff
@@ -912,6 +914,23 @@ def logged_out
...
+    if primary_email
+      primary_email.email = email
+    else
+      build_primary_email(email: email)
+    end
+  end
+
+  def associate_primary_email
+    primary_email.user_id = self.id
+    primary_email.save
```
tgxworld: Perhaps `primary_email.update!(user_id: self.id)`?

app/models/user_email.rb
```diff
@@ -0,0 +1,18 @@
...
+  validates :email, format: { with: EmailValidator.email_regex }, if: :email_changed?
+  validates :email, email: true, if: :email_changed?
+
+  def strip_downcase_email
+    if self.email
+      self.email = self.email.strip
+      self.email = self.email.downcase
+    end
+  end
+end
```
tgxworld: Can we [add the annotations](https://github.com/ctran/annotate_models) as well :)

app/controllers/users_controller.rb
LeoMcA: > Usually, it is not recommended to make cosmetic changes like these as it pollutes the commit history.

I agree, but since the email column has moved from the User model to the UserEmail model, this isn't cosmetic, is it?

> Also, find_by_* have been deprecated in Rails 4 and will be removed in Rails 5.

And since we define `find_by_email` in the User model, this won't be deprecated, will it?

tgxworld: Ah sorry my bad. So we're defining a `find_by_email` method in the `User` model. Hmm I would rename the method though cause it is confusing now. `find_by_email` used to mean that we're looking for a`User` based on the email column. Perhaps `find_by_user_email`?

LeoMcA: Sure, although do we want to keep `find_by_email` as a deprecated method so we don't suddenly break every plugin which makes use of it?

> Perhaps find_by_user_email?

The way the method is currently written, that would be the best name. Though, I'm now wondering if actually in all these places we want to use a `find_by_primary_email` method, and save `find_by_user_email` for any functionality we want to change as part of phase 2.

app/models/user_email.rb
LeoMcA: I knew I was forgetting something!

app/controllers/users_controller.rb
SamSaffron: This line does not even make sense, absolutely recommend renaming it. 

find_by_email is not a scope so it will not filter on staged: true, so what we have here is both a bug and bugs waiting to happen. perhaps the scope should be `with_email` ... and then `User.where(staged: true).with_email('bob@bob.com').first`

SamSaffron:
timeline wise, keep in mind we can not integrate this in prior to the 1.8 release, it will be first on the train after that. 

Still unclear at how:

```
if user = User.where(staged: true).find_by_email(params[:email].strip.downcase)
```

works given that this is not a scope. 

```
 def self.find_by_email(email)
 -    find_by(email: Email.downcase(email))
 +    user_email = UserEmail.find_by(email: Email.downcase(email))
 +    user_email.user if user_email

```

db/migrate/20170329064703_create_user_emails.rb
```diff
@@ -1,16 +1,20 @@
 class CreateUserEmails < ActiveRecord::Migration
   def up
     create_table :user_emails do |t|
-      t.references :user
+      t.integer :user_id
```
LeoMcA: @SamSaffron I removed the `t.references` and added a couple of other indexes. I haven't fully wrapped my head around indexes, though, so I'm not completely sure if they make sense.

app/controllers/users_controller.rb
LeoMcA: > find_by_email is not a scope so it will not filter on staged: true

I'm a bit confused, because it does seem to filter. I've added a test in my latest commit which passes with the line as it is, but fails if it's changed to `User.find_by_email(params[:email].strip.downcase)`, meaning the `.where(staged: true)` is doing something.

I can also see it filtering by staged: true in the console:

```
pry(main)> User.where(staged: true).find_by_email('no_email')
  UserEmail Load (0.3ms)  SELECT  "user_emails".* FROM "user_emails" WHERE "user_emails"."email" = 'no_email' LIMIT 1  [["email", "no_email"]]
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."staged" = 't' AND "users"."id" = -1 LIMIT 1  [["staged", true], ["id", -1]]
=> nil
```

LeoMcA:
I think my last couple of comments didn't appear because I accidentally started a new review. Just readded them as normal comments.

> timeline wise, keep in mind we can not integrate this in prior to the 1.8 release, it will be first on the train after that.

No problem, considering the complexity of this, I'll be happy if we integrate it prior to 2.0!

> Still unclear at how: `if user = User.where(staged: true).find_by_email(params[:email].strip.downcase)` works given that this is not a scope.

Isn't this how... it all works?

```
pry(main)> User.find_by_username('system')
  User Load (0.4ms)  SELECT  "users".* FROM "users" WHERE "users"."username_lower" = 'system' LIMIT 1  [["username_lower", "system"]]
=> #<User:0x007f1e1d5a1f58
[...]
pry(main)> User.where(staged: true).find_by_username('system')
  User Load (0.6ms)  SELECT  "users".* FROM "users" WHERE "users"."staged" = 't' AND "users"."username_lower" = 'system' LIMIT 1  [["staged", true], ["username_lower", "system"]]
=> nil
pry(main)> User.find_by_salt("33c5382bf9aee6d930b80208658d892d")
  User Load (0.4ms)  SELECT  "users".* FROM "users" WHERE "users"."salt" = '33c5382bf9aee6d930b80208658d892d' LIMIT 1  [["salt", "33c5382bf9aee6d930b80208658d892d"]]
=> #<User:0x007f1e16cffe28
[...]
pry(main)> User.where(staged: true).find_by_salt("33c5382bf9aee6d930b80208658d892d")
  User Load (0.5ms)  SELECT  "users".* FROM "users" WHERE "users"."staged" = 't' AND "users"."salt" = '33c5382bf9aee6d930b80208658d892d' LIMIT 1  [["staged", true], ["salt", "33c5382bf9aee6d930b80208658d892d"]]
=> nil
```

SamSaffron:
hmmm but find_by_email was implemented as a method on User now in this PR. it is just confusing to have it both on scope and on user, besides it being a second away from being deprecated ... you are going to have to define a proper scope here.

SamSaffron:
@LeoMcA we are ready to merge this in @tgxworld will be in charge of the merge. Can you resolve the conflicts?

LeoMcA:
@SamSaffron @tgxworld done!

db/migrate/20170329064703_create_user_emails.rb
```diff
@@ -0,0 +1,31 @@
...
+    create_table :user_emails do |t|
+      t.integer :user_id
+      t.string :email, limit: 513
+      t.timestamps
+    end
+    add_index :user_emails, :user_id
+    execute "CREATE UNIQUE INDEX index_user_emails_on_email ON user_emails ((lower(email)));"
+    execute "INSERT INTO user_emails (user_id, email, created_at, updated_at) SELECT id, email, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP FROM users WHERE users.id != -1"
+    change_table :users do |t|
+      t.remove :email
```
tgxworld: You'll have to delay dropping this column otherwise everything blows up during deployment on setups where migrations are run while web servers quite still kept running. You can have a look at https://github.com/discourse/discourse/blob/c53eeaaa6806b2e872f6ee21d4db01d5c437c152/lib/column_dropper.rb

Also, I'll prefer if we keep this column around first so that we don't lose everyone's email if something does terribly wrong during the migration. The column can be dropped maybe a month or two after things have stabilized.

db/migrate/20170329064703_create_user_emails.rb
```diff
@@ -0,0 +1,31 @@
...
+      t.integer :primary_email_id
+    end
+    add_index :users, :primary_email_id
+    execute "UPDATE users SET primary_email_id = user_emails.id FROM user_emails WHERE user_emails.user_id = users.id"
+    execute "UPDATE users SET primary_email_id = -1 WHERE id = -1"
```
tgxworld: Wouldn't seed already handle this?

db/migrate/20170329064703_create_user_emails.rb
```diff
@@ -0,0 +1,31 @@
+class CreateUserEmails < ActiveRecord::Migration
```
tgxworld: I would prefer if we can split this migration into the smaller pieces:

1 migration for UserEmails table
1 Migration for transferring records
1 migration for altering the user tables

Smaller migrations are usually recommend so you can easily rollback changes in steps instead of having one giant migration to fight with.

lib/admin_user_index_query.rb
```diff
@@ -106,7 +106,8 @@ def filter_by_search
       if ip = IPAddr.new(params[:filter]) rescue nil
         @query.where('ip_address <<= :ip OR registration_ip_address <<= :ip', ip: ip.to_cidr_s)
       else
-        @query.where('username_lower ILIKE :filter OR email ILIKE :filter', filter: "%#{params[:filter]}%")
+        @query.joins(:primary_email)
+              .where('username_lower ILIKE :filter OR email ILIKE :filter', filter: "%#{params[:filter]}%")
```
tgxworld: This query doesn't look right to me. We're joining on the `primary_email` tables but the query is still referencing the `email` column on the users table.

lib/avatar_lookup.rb
```diff
@@ -12,7 +12,7 @@ def [](user_id)
   private
 
   def self.lookup_columns
-    @lookup_columns ||= %i{id email username uploaded_avatar_id}
```
tgxworld: Hmm is this safe to nuke without replacing it?

lib/import_export/topic_importer.rb
```diff
@@ -18,7 +18,7 @@ def perform
 
     def import_users
       @export_data[:users].each do |u|
-        existing = User.where(email: u[:email]).first
+        existing = User.with_email(email: u[:email]).first
```
tgxworld: This query isn't right since the scope is expecting an email and not a hash.

spec/controllers/users_controller_spec.rb
```diff
@@ -786,6 +786,20 @@ def post_user
       include_examples 'failed signup'
     end
 
+    context 'with an existing email' do
+      email = 'leo@m.ca'
```
tgxworld: This is sort of declaring a global variable here.

lib/import_export/topic_importer.rb
LeoMcA: I'm not sure I understand: the scope is basically an alias for `.includes(:primary_email).where(user_emails: { email: ? }) }` so shouldn't it work the same as the original `.where` query?

lib/avatar_lookup.rb
LeoMcA: Not sure, it didn't make any tests fail, but I added it back in regardless ;)

db/migrate/20170329064703_create_user_emails.rb
LeoMcA: Done - I grouped them I think a little more logically than just "one migration modifies one table" let me know what you think

db/migrate/20170329064703_create_user_emails.rb
LeoMcA: Right, I've got the migration which drops this table saved in another fork: https://github.com/LeoMcA/discourse/commit/17d8286615216e3faba577d0ce58a86484db1a7a

db/migrate/20170329064703_create_user_emails.rb
LeoMcA: Yep, not sure why this was here

db/migrate/20170329064703_create_user_emails.rb
LeoMcA: Also because of this I've had to change code in a few places to explicitly specify the `user_emails.email` column rather than just `email` (which is now ambiguous)

lib/admin_user_index_query.rb
LeoMcA: So it did work when I was dropping `users.email` because there was only one `email` column across the two tables (in the `user_emails` table) - now there's two, I've specified we want the `user_emails.email` column

spec/controllers/users_controller_spec.rb
LeoMcA: I don't completely understand - isn't it local to the context block? - but I've defined it in a `:let` now

LeoMcA:
I wonder if we've scope creeped a little on the aims of phase 1:

@SamSaffron, @tgxworld what are your thoughts on renaming the `with_email` scope to `with_primary_email` (because that's what it really is)

In a similar vein, I think we should rename the `find_by_email` class method to `find_by_primary_email` (and update its functionality to match that name) - or did we want to remove it and use the `with_primary_email` scope instead?

Then in phase 2 we can add a `with_email` scope and `find_by_email` class method which query all emails (primary and additional) and positively decide where we want to use them in the codebase

SamSaffron:
We discussed this internally, @tgxworld will be finishing this off and sorting out conflicts and so on.

tgxworld:
Will wait till Monday to merge this in. Don't want to be playing around with migrations on a Friday. Nice work @LeoMcA  :+1:

tgxworld:
Thanks @LeoMcA. Merged in a56e98bfc4388917aa0dc62cdbf55a787828c67b

PR #4844, opened 2017-04-27
Allow inclusion of elided text from emails when sending to a category…

davidtaylorhq:
… address

The `always show trimmed content` site setting currently only works for group PMs, and replies to topics. This PR should make it also work for new topics created using the `Custom incoming email address` of a category


ZogStriP:
Cool, can you add a test?

davidtaylorhq:
I couldn't find any existing tests for the `always_show_trimmed_content` site setting, so I've added two: one for replies to existing topics (this was already working), and one for new topics created by emailing in to a category (this is what I've fixed).

davidtaylorhq:
@zogstrip Are these tests ok?

spec/components/email/receiver_spec.rb
```diff
@@ -296,6 +296,12 @@ def process(email_name)
       expect(topic.posts.last.raw).to eq("This is a reply :)\n\n<details class='elided'>\n<summary title='Show trimmed content'>&#183;&#183;&#183;</summary>\n---Original Message---\nThis part should not be included\n</details>")
     end
 
+    it "adds the 'elided' part of the original message for public replies when always_show_trimmed_content is enabled" do
+      SiteSetting.always_show_trimmed_content = true
+      expect { process(:original_message) }.to change { topic.posts.count }
```
tgxworld: Let's be a little more explicit here. How many new posts are we expecting?

spec/components/email/receiver_spec.rb
```diff
@@ -462,6 +468,16 @@ def process(email_name)
       expect { process(:new_user) }.to change(Topic, :count)
     end
 
+    it "adds the 'elided' part of the original message when always_show_trimmed_content is enabled" do
+      SiteSetting.always_show_trimmed_content = true
+
+      user = Fabricate(:user, email: "existing@bar.com", trust_level: SiteSetting.email_in_min_trust)
+      expect { process(:forwarded_email_to_category) }.to change(Topic, :count) # Topic created
```
tgxworld: Let's follow existing conventions in our code base and use `.to change { Topic.count }`

spec/components/email/receiver_spec.rb
```diff
@@ -462,6 +468,16 @@ def process(email_name)
...
+
+      new_post, = *Post.last(1)
```
tgxworld: I believe `Post.last` is sufficient here.

davidtaylorhq:
Thanks @tgxworld, I've made those changes

lib/email/receiver.rb
```diff
@@ -370,6 +370,7 @@ def process_destination(destination, user, body, elided)
 
         create_topic(user: user,
                      raw: body,
+                     elided: elided,
```
SamSaffron: one thing that is unclear from the tests here is how this does not add elided unconditionally. 
Can you repeat the first test with SiteSetting.always_show_trimmed_content = false ensuring elided content is not there?

davidtaylorhq: Good point, I've added an explicit test for when `always_show_trimmed_content` is `false`.

davidtaylorhq:
@SamSaffron is this additional test ok?

davidtaylorhq:
Any chance of getting this merged? It would be useful for my community.

@ZogStriP @SamSaffron @tgxworld

spec/components/email/receiver_spec.rb
```diff
@@ -483,6 +495,16 @@ def process(email_name)
       expect { process(:new_user) }.to change(Topic, :count)
     end
 
+    it "adds the 'elided' part of the original message when always_show_trimmed_content is enabled" do
+      SiteSetting.always_show_trimmed_content = true
+
+      user = Fabricate(:user, email: "existing@bar.com", trust_level: SiteSetting.email_in_min_trust)
+      expect { process(:forwarded_email_to_category) }.to change{Topic.count}.by(1) # Topic created
+
+      new_post, = Post.last
```
tgxworld: There is an extra `,` here.

tgxworld:
Lets try this out @davidtaylorhq. Thank you for your work on this :+1:

PR #4847, opened 2017-05-02
Allow users to resend/update email from confirmation page

eviltrout:
Previously, you had to try logging in again before presented with the options. Now they are present on the confirmation screen.


coding-horror:
there's no way for anyone else to end up on "your" confirmation screen is there? It's the equivalent of a custom GUID URL for you?

nlalonde:
After activating my account, and then reloading the page at `/u/account-created`, I still have the options to resend and change my email. When I try to change my email, it gives an error message.

I think `/u/account-created` should detect if you're already logged in and redirect.

eviltrout:
@coding-horror yes, it's based on a session variable. The URL doesn't even indicate which user it is.

@nlalonde good call, doin it

PR #4848, opened 2017-05-03
Perf optimization for mailing list

tgxworld:


coding-horror:
Why does this need to be a PR?

lib/email/styles.rb
```diff
@@ -228,7 +228,7 @@ def replace_relative_urls
 
       @fragment.css('[href]').each do |element|
         href = element['href']
-        if href =~ /^\/\/#{host}/
+        if href.start_with?("\/\/#{host}")
```
ZogStriP: I'm curious, what's the perf improvement of using `start_with?` as opposed to the regex?

app/mailers/user_notifications.rb
```diff
@@ -80,10 +80,21 @@ def mailing_list(user, opts={})
     @since = opts[:since] || 1.day.ago
     @since_formatted = short_date(@since)
 
-    @new_topic_posts      = Post.mailing_list_new_topics(user, @since).group_by(&:topic) || {}
```
ZogStriP: Do we use these scope elsewhere in the app? If not they should probably be removed.

ZogStriP:
LGTM 👍

tgxworld:
> Why does this need to be a PR?

I'm just making sure I don't break the tests

ZogStriP:
> I'm just making sure I don't break the tests

What don't you run the specs on your machine?

lib/email/styles.rb
tgxworld: `start_with?` is faster https://github.com/JuanitoFatas/fast-ruby#stringmatch-vs-stringstart_withstringend_with-code-start-code-end

I was curious as well so I wrote this script

```
require 'memory_profiler'

test_string = 'abcde'
abc = "abc"
MemoryProfiler.report do
  test_string.start_with?(abc)
  # test_string =~ /\/\/#{abc}/
end.pretty_print
```

`start_with?` doesn't allocate any objects while using Regexp allocates 4 strings for this case.

PR #4850, opened 2017-05-03
Add discourseEmbedClass option and apply classname to iframe html

patdavid:
This will allow the embed javascript to include a parameter
`discourseEmbedClass` that will be passed to the iframe page
where it will be parsed and applied to the iframe <html> element
as a class.

This allows targeting of CSS for different embeds.

### DIFF
--- a/app/views/layouts/embed.html.erb
+++ b/app/views/layouts/embed.html.erb
@@ -63,6 +63,19 @@
             }
           }
 
+          // Set html class
+          var params = location.href.split('?');
+          if ( params.length > 1 ) {
+            params = params[1].split('&');
+            for ( p in params ) {
+              var param = params[p].split('=');
+              if( param[0] == 'embed_class' ){
+                document.documentElement.className = param[1];
+              }
+            }
+          }
+
+
         };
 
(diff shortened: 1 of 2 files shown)


patdavid:
This is in reference to: https://meta.discourse.org/t/embedding-in-more-than-one-host/62002/

patdavid:
Could possibly ping @SamSaffron and @ZogStriP as well.

eviltrout:
I don't like the idea of the client passing this through. It seems like it could easily be used for XSS issues.

It should be part of the `embeddable_host` model and then inserted automatically by the rails application, plus an admin interface to add the class name to the particular host.

eviltrout:
(Also I meant to say thanks for the effort! Sorry that came across as rude.)

It just needs to go back to the drawing board.

patdavid:
Even better!  I don't have the depth of understanding required to probably help much in this case the `embeddable_host` model. :(

Is there something I could do to help with that in mind?

eviltrout:
I'll post in the meta topic. Maybe someone can offer to help you out.

PR #4853, opened 2017-05-05
Suggestion for string readability in client.en.yml

aputtu:
I misread the string while translating into another spoken language, and other might benefit from a more simple sentence.


tgxworld:
Thank you @aputtu

PR #4856, opened 2017-05-05
discourse-details: Add German locale

caugner:
Adds the German locale for the standard plugin *discourse-details*.

### DIFF
--- a/plugins/discourse-details/config/locales/client.de.yml
+++ b/plugins/discourse-details/config/locales/client.de.yml
@@ -1,0 +1,7 @@
+en:
+  js:
+    details:
+      title: Details ausblenden
+    composer:
+      details_title: Zusammenfassung
+      details_text: "Dieser Text wird versteckt"
--- a/plugins/discourse-details/config/locales/client.de.yml
+++ b/plugins/discourse-details/config/locales/client.de.yml
@@ -1,4 +1,4 @@
-en:
+de:
   js:
     details:
       title: Details ausblenden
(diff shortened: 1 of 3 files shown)


plugins/discourse-details/config/locales/client.de.yml
```diff
@@ -0,0 +1,7 @@
+en:
```
ZogStriP: This needs to be `de`, otherwise it's overwriting the `en` locale ;)

plugins/discourse-details/config/locales/server.de.yml
```diff
@@ -0,0 +1,3 @@
+en:
```
ZogStriP: Ditto

plugins/discourse-details/config/locales/client.de.yml
caugner: Oops, thank you, just fixed that. 👼

plugins/discourse-details/config/locales/server.de.yml
caugner: Ditto

plugins/discourse-details/config/locales/server.en.yml
```diff
@@ -1,3 +1,3 @@
-en:
+de:
```
tgxworld: This shouldn't be changed

SamSaffron:
OK I merged this by hand.

caugner:
@SamSaffron Thank you so much and sorry about this. I promise next time I'll fix the nit in time.

PR #4857, opened 2017-05-08
Add support for including class_name per embeddable_host

patdavid:
This applies a class_name to the <html> element of the embed for each host, as defined on: `/admin/customize/embedding`.

This is to address a better solution as suggested by @eviltrout here: https://github.com/discourse/discourse/pull/4850#issuecomment-299025680 and on [meta.discourse.org](https://meta.discourse.org/t/embedding-in-more-than-one-host/62002/10) to use the embeddable_host model.


eviltrout:
This looks quite good to me! However, there are no tests. At the very least there should be a controller spec that confirms that if the class_name is present in the record you're viewing then it should assign the `embeddable_css_class` variable for the template to render.

patdavid:
Something more than the test performed here?
https://github.com/discourse/discourse/pull/4857/files#diff-1987a252ba3efe037c7bf36f9c6f16b1R99

eviltrout:
The line you linked to is just code, not a test?

I mean a rspec controller spec to test that the code is functioning properly.

patdavid:
I'm not really sure how to proceed with the test portion here. :(

Is there someone I could ask or should I maybe mention it in the meta thread?

eviltrout:
I can give you a strong hint here. If it's not enough posting on meta might be a good idea!

Basically, in **spec/controllers/embed_controller_spec.rb** you could modify one of the records that are created in the `"with multiple hosts"` `before` block to include the class name attribute.

Then in the `it` block after the `get` has executed, you can add a new line like:

```ruby
expect(assigns(:embeddable_css_class)).to eq('what it should equal')
```

patdavid:
Son of a gun it works (and passes).  @eviltrout this should address the spec concern?

eviltrout:
Thanks, merged!

PR #4859, opened 2017-05-09
FEATURE: admin can set a timer to remind them of a topic

nlalonde:
Inspired by "publish topic to category", I've re-used the TopicStatusUpdate code again. This adds the idea that each user can have their own topic timer. We probably need a new name for this model.

### DIFF
--- a/app/jobs/regular/topic_reminder.rb
+++ b/app/jobs/regular/topic_reminder.rb
@@ -1,0 +1,28 @@
+module Jobs
+  class TopicReminder < Jobs::Base
+
+    def execute(args)
+      topic_status_update = TopicStatusUpdate.find_by(id: args[:topic_status_update_id])
+
+      topic = topic_status_update&.topic
+      user = topic_status_update&.user
+
+      if topic_status_update.blank? || topic.blank? || user.blank? ||
+          topic_status_update.execute_at > Time.zone.now
+        return
+      end
+
+      user.notifications.create(
+        notification_type: Notification.types[:topic_reminder],
+        topic_id: topic.id,
+        post_number: 1,
+        data: { topic_title: topic.title, display_username: user.username }.to_json
+      )
+
+      topic_status_update.trash!(Discourse.system_user)
+
+      true
+    end
+
+  end
+end
--- a/db/migrate/20170509202157_change_uniqueness_index_on_topic_status_updates.rb
+++ b/db/migrate/20170509202157_change_uniqueness_index_on_topic_status_updates.rb
@@ -1,0 +1,21 @@
+class ChangeUniquenessIndexOnTopicStatusUpdates < ActiveRecord::Migration
+  def up
+    execute "DROP INDEX idx_topic_id_status_type_deleted_at"
+
+    execute <<~SQL
+    CREATE UNIQUE INDEX idx_topic_id_status_type_user_id_deleted_at
+    ON topic_status_updates(topic_id, status_type, user_id)
+    WHERE deleted_at IS NULL
+    SQL
+  end
+
+  def down
+    execute "DROP INDEX idx_topic_id_status_type_user_id_deleted_at"
+
+    execute <<~SQL
+    CREATE UNIQUE INDEX idx_topic_id_status_type_deleted_at
+    ON topic_status_updates(topic_id, status_type)
+    WHERE deleted_at IS NULL
+    SQL
+  end
+end
--- a/spec/jobs/topic_reminder_spec.rb
+++ b/spec/jobs/topic_reminder_spec.rb
@@ -1,0 +1,58 @@
+require 'rails_helper'
+
+describe Jobs::TopicReminder do
+  let(:admin) { Fabricate(:admin) }
+  let(:topic) { Fabricate(:topic, topic_status_updates: [
+    Fabricate(:topic_status_update, user: admin, status_type: TopicStatusUpdate.types[:reminder])
+  ]) }
+
+  before do
+    SiteSetting.queue_jobs = true
+  end
+
+  it "should be able to create a reminder" do
+    topic
+    topic_status_update = topic.topic_status_update(admin)
+    Timecop.freeze(1.day.from_now) do
+      expect {
+        described_class.new.execute(topic_status_update_id: topic_status_update.id)
+      }.to change { Notification.count }.by(1)
+      expect( admin.notifications.where(notification_type: Notification.types[:topic_reminder]).first&.topic_id ).to eq(topic.id)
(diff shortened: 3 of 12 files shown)


jomaxro:
Is there a Meta topic about this?  I'd like to discuss the idea - specifically, if this is restricted to admins, or if staff (moderators) can also use this.

db/migrate/20170509202157_change_uniqueness_index_on_topic_status_updates.rb
```diff
@@ -0,0 +1,21 @@
+class ChangeUniquenessIndexOnTopicStatusUpdates < ActiveRecord::Migration
+  def up
+    execute "DROP INDEX idx_topic_id_status_type_deleted_at"
+
+    execute <<~SQL
```
tgxworld: Are we going to be allowing multiple reminders per topic?

nlalonde: Each person can set their own reminder, yes.

nlalonde:
@jomaxro Hmm, there's no topic yet. This is an idea @coding-horror had and a first implementation of it. It's only available to staff, in the same UI as the auto-close setting.

jomaxro:
@nlalonde sounds good. Until i can test the feature myself, my only concern was admin-only or staff-only. Thanks for clearing that up!

db/migrate/20170509202157_change_uniqueness_index_on_topic_status_updates.rb
```diff
@@ -0,0 +1,21 @@
...
+    CREATE UNIQUE INDEX idx_topic_id_status_type_user_id_deleted_at
```
tgxworld: I don't think this is what we want here. This index means that other topic status updates like `close topic` can be set more than once as long as it is created by a different user. Perhaps a json column that can store dynamic data might be better here? In the column we can maintain the list of user ids and the timings that they would want to be reminded at.

app/jobs/regular/topic_reminder.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      topic = topic_status_update&.topic
+      user = topic_status_update&.user
+
+      if topic_status_update.blank? || topic.blank? || user.blank? ||
+          topic_status_update.execute_at > Time.zone.now
+        return
+      end
+
+      user.notifications.create(
```
tgxworld: I usually use `create!`  so that it doesn't silently fail.

app/jobs/regular/topic_reminder.rb
```diff
@@ -0,0 +1,28 @@
...
+        notification_type: Notification.types[:topic_reminder],
+        topic_id: topic.id,
+        post_number: 1,
+        data: { topic_title: topic.title, display_username: user.username }.to_json
+      )
+
+      topic_status_update.trash!(Discourse.system_user)
+
+      true
```
tgxworld: Hmm is this required?

app/jobs/regular/topic_reminder.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+  end
+end
```
tgxworld: Missing newline at the end of the file.

spec/jobs/topic_reminder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  let(:topic) { Fabricate(:topic, topic_status_updates: [
+    Fabricate(:topic_status_update, user: admin, status_type: TopicStatusUpdate.types[:reminder])
+  ]) }
+
+  before do
+    SiteSetting.queue_jobs = true
+  end
+
+  it "should be able to create a reminder" do
+    topic
```
tgxworld: This line is not necessary since we'll be initializing the record in the line below.

spec/jobs/topic_reminder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    topic_status_update = topic.topic_status_update(admin)
+    Timecop.freeze(1.day.from_now) do
+      expect {
+        described_class.new.execute(topic_status_update_id: topic_status_update.id)
+      }.to change { Notification.count }.by(1)
+      expect( admin.notifications.where(notification_type: Notification.types[:topic_reminder]).first&.topic_id ).to eq(topic.id)
+      expect( TopicStatusUpdate.where(id: topic_status_update.id).first ).to be_nil
```
tgxworld: Maybe `expect(topic.reload.topic_status_updates).to eq([])` so that we ensure the topic does not have any topic status updates.

spec/jobs/topic_reminder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    topic_status_update = topic.topic_status_update(admin)
+    topic_status_update.update_attribute(:execute_at, 8.hours.from_now)
+    Timecop.freeze(6.hours.from_now) do
+      expect {
+        described_class.new.execute(topic_status_update_id: topic_status_update.id)
+      }.to_not change { Notification.count }
+    end
+  end
+
+  it "does nothing if topic was deleted" do
```
tgxworld: Hmm this test and `"does nothing if it was trashed before the scheduled time"` is testing the code path.

spec/models/topic_spec.rb
```diff
@@ -1275,6 +1275,72 @@ def build_topic_with_title(title)
...
+    let(:reminder) { Fabricate(:topic_status_update, execute_at: 2.hours.from_now, status_type: TopicStatusUpdate.types[:reminder], user: admin) }
+
+    it "returns nil if there aren't any" do
+      expect(topic.topic_status_update).to eq(nil)
+    end
+
+    it "returns the next one to execute" do
+      later  = Fabricate(:topic_status_update, execute_at: 24.hours.from_now, status_type: TopicStatusUpdate.types[:close])
+      sooner = Fabricate(:topic_status_update, execute_at: 8.hours.from_now, status_type: TopicStatusUpdate.types[:open])
+      topic.topic_status_updates = [later, sooner]
```
tgxworld: I wanted to avoid scenarios like this when I redesigned the feature. Having two topic status updates per topic is too complicated and probably not required at this point in time.

tgxworld: Hmm I spoke too soon. We do need to allow multiple public and private topic updates though... Perhaps we should add a `public` column on the model? One topic is only allowed to have one public status update but multiple private status update.

nlalonde:
Thanks for the review @tgxworld . I'll do the "delete topic" option first, and then redo this with changes.

PR #4860, opened 2017-05-10
FEATURE: bulk relist

OsamaSayegh:
https://meta.discourse.org/t/make-listed-again-in-the-bulk-actions/58286/5

I hope I did it right :)

### DIFF
--- a/app/assets/javascripts/discourse/controllers/topic-bulk-actions.js.es6
+++ b/app/assets/javascripts/discourse/controllers/topic-bulk-actions.js.es6
@@ -26,18 +26,12 @@ export default Ember.Controller.extend(ModalFunctionality, {
   categoryId: Ember.computed.alias('model.category.id'),
 
   onShow() {
-    let showRelistButton = false;
-    const topics = this.get('model.topics');
-    for (let t = 0; t < topics.length; t++) {
-      if (!topics[t].visible) {
-        showRelistButton = true;
-        break;
-      }
-    }
-    if (showRelistButton && !_buttons[9]) {
+    const showRelistButton = this.get('model.topics').some(t => !t.visible);
+    const relistButtonIndex = _buttons.findIndex(b => b.action === 'relistTopics');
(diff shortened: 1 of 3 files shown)


OsamaSayegh:
Do I need to fill the form every time I submit a PR?

coding-horror:
Your cla email must match your github email for it to "know" you

app/assets/javascripts/discourse/controllers/topic-bulk-actions.js.es6
```diff
@@ -26,6 +26,19 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+    const topics = this.get('model.topics');
+    for (let t = 0; t < topics.length; t++) {
+      if (!topics[t].visible) {
+        showRelistButton = true;
+        break;
+      }
+    }
+    if (showRelistButton && !_buttons[9]) {
+      addBulkButton('relistTopics', 'relist_topics');
+    } else if (!showRelistButton && _buttons[9]) {
```
eviltrout: Hard coding an index here seems dangerous. What is the condition actually checking? Is it if the button already exists? In that case it should be checking by some kind of indentifier.

app/assets/javascripts/discourse/controllers/topic-bulk-actions.js.es6
```diff
@@ -26,6 +26,19 @@ export default Ember.Controller.extend(ModalFunctionality, {
   categoryId: Ember.computed.alias('model.category.id'),
 
   onShow() {
+    let showRelistButton = false;
+    const topics = this.get('model.topics');
+    for (let t = 0; t < topics.length; t++) {
```
eviltrout: I think this could be replaced by something like 

`const showRelistButton = this.get('model.topics').some(t => !t.visible)`

OsamaSayegh:
@eviltrout  Thanks for your feedback! would you mind taking another look? I've just made some changes :)

eviltrout:
Seems good now, thanks :)

PR #4861, opened 2017-05-10
Expose limit for admin/users/list API

JaredReisinger:
Expose querystring access to the "limit" parameter so that we're not restricted to only the first 100 users.  (See also the discussion at https://meta.discourse.org/t/using-the-api-for-user-management-the-100-user-limit/37562.)

Added tests to ensure that the parameter can't be used in a SQL injection attack, and that if anything is amiss it falls back to the previously-existing behavior.

_Do note that this PR is offered in advance of full agreement from folks that this is even a reasonable thing to do... but it gives a starting point for the discussion._

### DIFF
--- a/lib/admin_user_index_query.rb
+++ b/lib/admin_user_index_query.rb
@@ -24,7 +24,20 @@ class AdminUserIndexQuery
     'read_time' => 'user_stats.time_read'
   }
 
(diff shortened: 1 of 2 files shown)


JaredReisinger:
Closing this as per @oblakeerickson's comments about the API design.  😄

PR #4863, opened 2017-05-11
Moderator post creation should not fail

deivid-rodriguez:
These are just some initial changes extracted from #4691, so they are easier to review. Once (if) this is merged, I'll rebase the other PR to get it landed as well.

### DIFF
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -107,8 +107,8 @@ class PostCreator
       topic_creator = TopicCreator.new(@user, guardian, @opts)
       return false unless skip_validations? || validate_child(topic_creator)
     else
-      @topic = Topic.find_by(id: @opts[:topic_id])
-      if (@topic.blank? || !guardian.can_create?(Post, @topic))
+      @topic = Topic.with_deleted.find_by(id: @opts[:topic_id])
+      if (@topic.blank? || !guardian.can_create?(Post, @topic)) && !skip_validations?
         errors[:base] << I18n.t(:topic_not_found)
         return false
       end
@@ -362,13 +362,15 @@ class PostCreator
   end
 
   def create_topic
-    return if @topic
-    begin
-      topic_creator = TopicCreator.new(@user, guardian, @opts)
-      @topic = topic_creator.create
-    rescue ActiveRecord::Rollback
-      rollback_from_errors!(topic_creator)
+    unless @topic
+      begin
+        topic_creator = TopicCreator.new(@user, guardian, @opts)
+        @topic = topic_creator.create
+      rescue ActiveRecord::Rollback
+        rollback_from_errors!(topic_creator)
+      end
     end
+
     @post.topic_id = @topic.id
     @post.topic = @topic
     if @topic && @topic.category && @topic.category.all_topics_wiki
--- a/spec/models/post_mover_spec.rb
+++ b/spec/models/post_mover_spec.rb
@@ -268,7 +268,6 @@ describe PostMover do
 
           # Check out the original topic
           topic.reload
-          expect(topic.posts_count).to eq(2)
           expect(topic.highest_post_number).to eq(3)
           expect(topic.featured_user1_id).to be_blank
           expect(topic.like_count).to eq(0)
--- a/spec/models/post_mover_spec.rb
+++ b/spec/models/post_mover_spec.rb
@@ -225,6 +225,12 @@ describe PostMover do
           expect(topic.closed).to eq(true)
         end
 
+        it "moving all posts will create a 'topic is closed' moderator post" do
+          new_topic = topic.move_posts(user, [p1.id, p2.id, p3.id, p4.id], title: "new testing topic name", category_id: category.id)
+
+          expect(topic.posts).to include(have_attributes(post_type: Post.types[:small_action], action_code: "closed.enabled"))
+        end
+
         it 'does not move posts that do not belong to the existing topic' do
           new_topic = topic.move_posts(
             user, [p2.id, p3.id, p5.id], title: 'Logan is a pretty good movie'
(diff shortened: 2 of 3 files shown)


lib/post_creator.rb
```diff
@@ -107,8 +107,8 @@ def valid?
       topic_creator = TopicCreator.new(@user, guardian, @opts)
       return false unless skip_validations? || validate_child(topic_creator)
     else
-      @topic = Topic.find_by(id: @opts[:topic_id])
-      if (@topic.blank? || !guardian.can_create?(Post, @topic))
+      @topic = Topic.with_deleted.find_by(id: @opts[:topic_id])
+      if (@topic.blank? || !guardian.can_create?(Post, @topic)) && !skip_validations?
```
tgxworld: Why is `skip_validations` required here? I don't see any validations in this line or within the following closure

tgxworld: Why is `skip_validations` required here? I don't see any validations in this line or within the following closure

spec/models/post_mover_spec.rb
```diff
@@ -201,6 +201,12 @@
           expect(topic.closed).to eq(true)
         end
 
+        it "moving all posts will create a 'topic is closed' moderator post" do
+          new_topic = topic.move_posts(user, [p1.id, p2.id, p3.id, p4.id], title: "new testing topic name", category_id: category.id)
+
+          expect(topic.posts).to include(have_attributes(post_type: Post.types[:small_action], action_code: "closed.enabled"))
```
tgxworld: Maybe `Post.last` instead of checking all the posts?

spec/models/post_mover_spec.rb
```diff
@@ -314,6 +319,18 @@
...
 
+      context "moving from a deleted topic" do
+        before do
+          topic.trash!
+        end
+
+        it "adds a moderator post at the location of the first moved post" do
+          topic.move_posts(user, [p2.id, p4.id], title: "new testing topic name")
+
+          expect(topic.posts).to include(have_attributes(post_type: Post.types[:small_action], action_code: "split_topic", post_number: 2, sort_order: 2))
```
tgxworld: Same here, I'll prefer if we find the right post instead of checking all the posts.

tgxworld:
@deivid-rodriguez Sorry for the late review as we working towards the 1.8 release. This PR looks OK to me aside from some minor comments.

lib/post_creator.rb
deivid-rodriguez: @tgxworld Both tests added here fail without this check.

For example, in the case of creating a "topic is closed" moderator post, we need to check `skip_validations?` because otherwise the previous condition will fail, since the `topic` is already closed when creating the moderator post, and thus `!guardian.can_create?(Post, @topic)` will be false (https://github.com/discourse/discourse/blob/860a7e1fc006d87ef324740f51c108d973f17e28/lib/guardian/topic_guardian.rb#L29).

In general I think all those guardian checks should be skipped whenever we create moderator posts (since regular validations don't apply for those), so that's why they pass the `skip_validations` parameter. That was the case everywhere except for here.

spec/models/post_mover_spec.rb
deivid-rodriguez: But the topic is not the last post here, the last post is the "4 posts were split to a new topic..." moderator post. Since I don't think we care about the specific order moderator posts are created in, I think it's fine.

spec/models/post_mover_spec.rb
deivid-rodriguez: Makes sense to me and matches better the test description. What about 825089bf5b675d496032e479a4c0a2935c94889b?

deivid-rodriguez:
@tgxworld Thanks for the review. I replied to your comments and rebased against latest master to make sure it still works ok.

deivid-rodriguez:
@tgxworld The build passed! Is this ready now?

spec/models/topic_spec.rb
```diff
@@ -607,17 +607,17 @@ def build_topic_with_title(title)
...
       end
 
-      it "should not increment moderator_posts_count" do
+      it "should still increment moderator_posts_count" do
         expect(topic.moderator_posts_count).to eq(0)
 
         topic.add_moderator_post(user, "winter is never coming")
 
-        expect(topic.moderator_posts_count).to eq(0)
+        expect(topic.moderator_posts_count).to eq(1)
```
tgxworld: Hmm I had a closer review and I don't think this behavior is right here, if validations fail the post shouldn't be created at all. I'm just afraid this might have security implications.

lib/post_creator.rb
tgxworld: I'm abit worried here because of the comment in

https://github.com/discourse/discourse/blob/80e348d226583d7dbbb48a914ced96d9fef60490/lib/guardian/topic_guardian.rb#L22

I would rather we pass in an additional option to allow a post to be created in a deleted topic rather than making the condition dependent on skip_validations

spec/models/post_mover_spec.rb
tgxworld: Hmm this test looks like a duplicate of the test above it.

spec/models/topic_spec.rb
deivid-rodriguez: I really don't want to keep pushing anymore. I'm too tired. Thanks!

erlend-sh:
Sorry that we couldn't fet this merged @deivid-rodriguez . We really do appreciate all the work you put into this. With bigger changes like these that have security implications the review process can get quite long.

Did you make a topic on meta.discourse.org for this issue? If there's a legitimate bug here then we should at least track it.

coding-horror:
I don't think we were able to repro the bug this was supposedly fixing??

https://meta.discourse.org/t/error-when-moving-post-to-existing-topic/38790/5

PR #4865, opened 2017-05-12
Fix the padding on the login and register buttons.

chrisgavin:
Currently the register and login buttons have their left padding set to 0. In combination with the background of the buttons being set to non-transparent on hover this results in a very odd looking button.

Before this patch:

image

After this patch:

image

### DIFF
--- a/app/assets/stylesheets/common/base/login.scss
+++ b/app/assets/stylesheets/common/base/login.scss
@@ -103,7 +103,5 @@ $input-width: 220px;
 button#login-link, button#new-account-link
 {
   background: transparent;
-  padding-left: 0;
-  margin-left: 20px;
   color: dark-light-choose(scale-color($primary, $lightness: 35%), scale-color($secondary, $lightness: 65%));
 }


coding-horror:
did you test this on mobile as well?

chrisgavin:
I don't think it looks any worse on mobile, but maybe there should be some margin between the buttons?

Before
image

After
image

coding-horror:
I believe it is done that way because otherwise it does not fit on narrow
phones like iPhone 5. Check in your emulator.

chrisgavin:
Okay, I've tried it out in the Chrome device emulator and it looks okay. I can also try it out on an actual iPhone 5 when I get home if that would be helpful.

Before (With Button Color)
image

After (With Button Color)
image

After (Without Button Color)
image

SamSaffron:
merging this in and then will tweak mobile a bit, thanks!

SamSaffron:
seems fine on mobile now as well, no need to change anything.

PR #4866, opened 2017-05-12
Add pagination to /admin/users/list API

JaredReisinger:
Prior to this, only the first 100 active/new/etc. users were available via the `/admin/users/list` API.  This change adds support for a `page=#` querystring parameter so that *all* of the users can be retrieved.  Requests for pages past the last user result in an empty-list response; requests for negative pages (or zero) just return the first page.

Added tests to cover pagination.


JaredReisinger:
FWIW, the test failure looks unrelated (and is a timing-based test to boot, which I know can sometimes be tricky to test reliably).

JaredReisinger:
@oblakeerickson : Hey, Blake, is there anything special I should do to get traction on this PR?  I know you likely have other things on your plate, I just wanted to make sure this was seen.  Thanks!

oblakeerickson:
Thanks!! Sorry for the delay.

JaredReisinger:
No worries!  Thanks!

PR #4868, opened 2017-05-15
Bump ruby version

rimian:


tgxworld:
Might want to just bump it to 2.4.1

rimian:
My staging server tells me it's running 2.3.3. But I bumped it for you :)

tgxworld:
Thanks @rimian  :+1: `.ruby-version` is used by rbenv so it only affects development.

PR #4870, opened 2017-05-15
FIX: Ensure we clean up socket servers.

tgxworld:
cc @SamSaffron  Can you review this? Thank you!

### DIFF
--- a/lib/socket_server.rb
+++ b/lib/socket_server.rb
@@ -16,7 +16,8 @@ class SocketServer
   end
 
   def stop
-    @server.close if @server
+    @server&.close rescue nil
+    FileUtils.rm_f(@socket_path)
     @server = nil
     @blk = nil
   end
@@ -26,9 +27,14 @@ class SocketServer
   def new_accept_thread
     server = @server
     Thread.new do
-      done = false
-      while !done
-        done = !accept_connection(server)
+      begin
+        done = false
+        while !done
+          done = !accept_connection(server)
+        end
+      ensure
+        self.stop
+        Rails.logger.info("Cleaned up socket server at #{@socket_path}")
       end
     end
   end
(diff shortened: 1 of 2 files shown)


SamSaffron:
hmmm... we need a test here if we are changing internals. 

I guess this ensures that if the container restarts and pids end up being reused we unconditionally clean up, but I though unicorn already cleaned up on boot? 

need more context about the actual bug being fixed here, also I feel this change is at high risk of leaving orphans around cause there is no cleanup anymore.

tgxworld:
> I guess this ensures that if the container restarts and pids end up being reused we unconditionally clean up, but I though unicorn already cleaned up on boot?

@SamSaffron  This is related to the question I asked you about on Slack yesterday. Unicorn only cleans up the files for the unicorn workers. The files for unicorn master and Sidekiq are not being nuked when the thread dies so I made this change to give us a clean pattern of cleaning up. My believe is that the `SocketServer` should be responsible of cleaning up itself once it stops. Also, we're not calling `@server.close` at all for any of the `SocketServer` that we start so I was wondering if that might leave us in a weird state.

> also I feel this change is at high risk of leaving orphans around cause there is no cleanup anymore.

Hmm which orphans are you referring to? All files in the stats directory are nuked on boot should the cleanup fail.

lib/socket_server.rb
```diff
@@ -19,16 +19,22 @@ def stop
     @server.close if @server
     @server = nil
     @blk = nil
+    FileUtils.rm_f(@socket_path)
```
SamSaffron: need to protect against errors here, cause we have @server.close if @server ... similarly only rm if the socket path is there

SamSaffron:
Sure seems safe enough!

PR #4871, opened 2017-05-15
FEATURE: remove table wrapping posts in notification emails

LeoMcA:
Per: https://meta.discourse.org/t/simplified-html-emails/62600/12?u=leomca

This removes the table wrapping the post in notification emails, but keeps the table wrapping the user details.

I tested on litmus.com against the [10 most popular email clients](https://emailclientmarketshare.com/), with an email generated by this code, and with an email generated by the current code, to compare.

There were a couple of small regressions, I'll take advice on whether it's worth hunting down the reasons and fixing them:

In Outlook on Windows in versions from 2007 to 2016, and in Windows 10 Mail, there was a loss of padding under the user details header:

*With table:*

ol2007-vertical-allowed-1366

*Without table:*

ol2007-vertical-allowed-1366

And on all devices running iOS 10.2, a blue border appeared next to the avatar in the user details header:

*With table:*

ipadpro13in-vertical-allowed-1366

*Without table:*

ipadpro13in-vertical-allowed-1366

### DIFF
--- a/app/views/email/_post.html.erb
+++ b/app/views/email/_post.html.erb
@@ -1,31 +1,23 @@
-<table class='post-wrapper <%= post.whisper? ? "whisper" : "" %> <%= use_excerpt ? "excerpt" : ""%>'>
-  <tbody>
+<div class='post-wrapper <%= post.whisper? ? "whisper" : "" %> <%= use_excerpt ? "excerpt" : ""%>'>
+  <table>
     <tr>
+      <td class='user-avatar'>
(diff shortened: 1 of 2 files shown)


coding-horror:
I think this has promise, can you take a look @eviltrout if you agree we can slip it into 1.8 since the email content is not really changing, just the vast majority is moved outside the table?

eviltrout:
Seems good to me! Let's try it out.

PR #4872, opened 2017-05-16
Update crawler_detection.rb

mcmcclur:
Add HTTrack to the list of detected crawlers so that Discourse will serve vanilla HTML per https://meta.discourse.org/t/a-basic-discourse-archival-tool/62614/25


SamSaffron:
sure, seems fine to me, thanks!

PR #4874, opened 2017-05-17
Use `bundle exec` for docker_dev rake

davidtaylorhq:
See https://meta.discourse.org/t/qunit-tests-wont-pass-in-discourse-dev-docker-image/62797/9


SamSaffron:
thanks!

PR #4875, opened 2017-05-17
Image optimisation

lukescammell:
Small savings of between 1-1000 bytes on approximately 5,000 files.

Notable reduction on `/public/images/welcome/discourse-edit-post-animated.gif` from 164,585 bytes to 42,728 bytes.

Also on `/discourse/spec/fixtures/images/large & unoptimized.png` making it very unsemantic ;) (from 421,730 bytes to 248,080 bytes

Ignoring those two, the total saving is 477,006 bytes. Not a lot, but hey…


ZogStriP:
> Also on `/discourse/spec/fixtures/images/large & unoptimized.png` making it very unsemantic ;) (from 421,730 bytes to 248,080 bytes

Well, the idea is to use that image to test the optimizer. So it's wrong to optimize it.

coding-horror:
We would need to redo this anyway for emoji updates, thanks anyhow

PR #4877, opened 2017-05-19
FEATURE: Require spec helpers for plugins

rimian:
This will allow plugin developers to load fixtures and helpers in specs.  

  * Follows any symlinked plugins

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -31,9 +31,8 @@ Spork.prefork do
 
   # Require plugin helpers at plugin/[plugin]/spec/plugin_helper.rb (includes symlinked plugins).
   if ENV['LOAD_PLUGINS'] == "1"
-    Dir[Rails.root.join("plugins/**")].each do |plugin|
-      helper = "#{plugin}/spec/plugin_helper.rb"
-      require helper if File.exists?(helper)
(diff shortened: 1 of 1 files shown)


spec/rails_helper.rb
```diff
@@ -29,6 +29,14 @@
   Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
   Dir[Rails.root.join("spec/fabricators/*.rb")].each {|f| require f}
 
+  # Require plugin helpers at plugin/[plugin]/spec/plugin_helper.rb (includes symlinked plugins).
+  if ENV['LOAD_PLUGINS'] == "1"
+    Dir[Rails.root.join("plugins/**")].each do |plugin|
```
SamSaffron: this is fine, but I think "plugins/*" would suffice here

SamSaffron: actually ... why not search for them properly and avoid the double lookup

```
Dir[Rails.root.join("plugins/*/spec/plugin_helper.rb")]
```

rimian:
I tried  that before but globs don't work with symlinks. :(

spec/rails_helper.rb
rimian: Hey that worked! Thanks.

SamSaffron:
cool, seems fine to me them 👍

PR #4887, opened 2017-05-26
Update server.ru.yml

Toxu-ru:


ZogStriP:
This needs to be translated on [transifex](https://www.transifex.com/discourse/discourse-org/).

PR #4888, opened 2017-05-29
FEATURE: require name when accepting invite if 'full name required' setting is enabled

techAPJ:
Requested here: https://meta.discourse.org/t/require-password-for-trusted-user-invites/13750/52?u=techapj

@nlalonde can you please review this PR?


techAPJ:
@nlalonde Thanks! ❤️

PR #4889, opened 2017-05-29
Add new (experimental) mbox importer

gschlager:
This adds a refactored version of the [existing mbox importer](https://github.com/discourse/discourse/blob/master/script/import_scripts/mbox.rb). The old one will still be there until this works completely.

What's different to the existing mbox importer?
* Configuration with a `settings.yml` file. There's currently no support for environment variables.
* Reuses [Email::Receiver](https://github.com/discourse/discourse/blob/master/lib/email/receiver.rb) for parsing emails. There's no reason for duplicating that code anymore.
* Marks all imported messages as "arrived via email".
* Hides the elided part of messages in a `[detail]` section.
* The indexing and importing of messages can be interrupted and resumed at any time.
* Better detection of related messages (replies).
* Currently it's not possible to import into the `uncategorized` category.
* Category names can be configured with an optional `metadata.yml` stored inside of directories. Otherwise the directory name is used as category name.
* During the indexing phase it stores all the data needed for the import in the `index.db`. So, there's no need to reparse emails during import unless an email contains attachments.
* The import order of messages is calculated in the database because sqlite has great support for grfor [Recursive Common Table Expressions](https://sqlite.org/lang_with.html). There's no need to do all the heavy work in Ruby.
* Due to performance problems imported emails are currently cooked with the regular cooking method. This will probably change in the future to use [EmailCook](https://github.com/discourse/discourse/blob/master/lib/email_cook.rb) again.
* Oh, and the importer doesn't crash anymore. 🎉 (At least not with the large mailing lists I tried.)

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -28,6 +28,9 @@ module Email
     class InvalidPostAction            < ProcessingError; end
 
     attr_reader :incoming_email
+    attr_reader :raw_email
+    attr_reader :mail
+    attr_reader :message_id
 
     def initialize(mail_string)
       raise EmptyEmailError if mail_string.blank?
@@ -241,7 +244,7 @@ module Email
 
     def try_to_encode(string, encoding)
       encoded = string.encode("UTF-8", encoding)
-      encoded.present? && encoded.valid_encoding? ? encoded : nil
+      !encoded.nil? && encoded.valid_encoding? ? encoded : nil
     rescue Encoding::InvalidByteSequenceError,
            Encoding::UndefinedConversionError,
            Encoding::ConverterNotFoundError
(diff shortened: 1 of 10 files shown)


lib/email/receiver.rb
```diff
@@ -241,7 +244,7 @@ def fix_charset(mail_part)
 
     def try_to_encode(string, encoding)
       encoded = string.encode("UTF-8", encoding)
-      encoded.present? && encoded.valid_encoding? ? encoded : nil
+      !encoded.nil? && encoded.valid_encoding? ? encoded : nil
```
gschlager: `encoded.present?` can crash with "ArgumentError: invalid byte sequence in UTF-8"

ZogStriP:
Awesome. Thanks 👍

PR #4890, opened 2017-05-30
Language fixes for client

comradekingu:


jomaxro:
I'm curious if there is a topic on [Meta](https://meta.discourse.org) about these changes?  Many of the edits seem to change the string from active to passive voice...

PR #4892, opened 2017-05-30
Update code-of-conduct.md

aabowen:
Please add usage rights.


SamSaffron:
Hi Alex 

totally fine to add something here what should we be adding, can you link me to some examples.

aabowen:
Is the CoC under a Creative Commons? Can we use and change the CoC if we use attribution, for example? Can we edit/change it as needed/wanted?

SamSaffron:
Yes to everything, can you include some sample text of you would like to see and some links to other CoCs that contain such text.

SamSaffron:
Let's just keep the discussion here please: 

https://meta.discourse.org/t/usage-rights-of-coc/63615

PR #4893, opened 2017-05-31
Remove columns that are no longer being used.

tgxworld:


tgxworld:
@nlalonde Can you review this? We're dropping columns here so I would like to have an extra pair of :eyes:

tgxworld:
@nlalonde Updated :)

SamSaffron:
This looks fine, but be sure to push business as soon as you migrate, I
have a very strong feeling this will stall

PR #4896, opened 2017-06-02
Advances search logical and by multiple tags 

dmacjam:
**User story:** As a user, I want to search posts which contains all of the tags I entered in order to find posts in more detailed way.

Related discussions: 
https://meta.discourse.org/t/does-multiple-tag-search-support-and-vs-or/56041
https://meta.discourse.org/t/multiple-tags-search-query-takes-too-long-big-forum/56023/8

New checkbox _Contains all tags_ is added to advanced search. If checkbox is checked, separator between tags is '+' (represents logical AND) instead of ',' (represents logical OR, already in use). Backend for search is done in Postgre built-in full text and it has same performance (query plan is below doing only one sequential database scan on tags and topic_tags tables) for any number of tags. 

I added two new rspec tests into tags context and moved old test for logical or search into the same context.

query-plan

### DIFF
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -77,7 +77,8 @@ export default Em.Component.extend({
             likes: false,
             private: false,
             seen: false
-          }
+          },
+          all_tags: false
         },
         status: '',
         min_post_count: '',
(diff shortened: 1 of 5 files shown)


eviltrout:
This seems good to me. Maybe we'll merge next week!

nlalonde:
Looks great!

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -230,13 +231,15 @@ export default Em.Component.extend({
 
     const match = this.filterBlocks(REGEXP_TAGS_PREFIX);
     const tags = this.get('searchedTerms.tags');
+    const contain_all_tags = this.get('searchedTerms.special.all_tags');
 
     if (match.length !== 0) {
-      const existingInput = _.isArray(tags) ? tags.join(',') : tags;
+      const join_char = contain_all_tags ? '+' : ',';
```
tgxworld: We use :camel: case in our Javascript files :)

PR #4897, opened 2017-06-02
FIX: Groups that do not have any owners should not allow membership re…

tgxworld:
…quests.

### DIFF
--- a/db/migrate/20170602132735_fix_group_allow_membership_requests.rb
+++ b/db/migrate/20170602132735_fix_group_allow_membership_requests.rb
@@ -1,0 +1,13 @@
+class FixGroupAllowMembershipRequests < ActiveRecord::Migration
+  def up
+    execute <<~SQL
+    UPDATE groups g
+    SET allow_membership_requests = 'f'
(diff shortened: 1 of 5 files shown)


db/migrate/20170602132735_fix_group_allow_membership_requests.rb
```diff
@@ -0,0 +1,13 @@
+class FixGroupAllowMembershipRequests < ActiveRecord::Migration
+  def up
+    execute <<~SQL
+    UPDATE groups g
+    SET allow_membership_requests = 'f'
+    WHERE COALESCE((SELECT 1 FROM group_users gu WHERE gu.owner = 't' AND gu.group_id = g.id LIMIT 1), 0) != 1
```
eviltrout: Why not use `WHERE NOT EXISTS(SELECT 1 FROM group_users gu WHERE gu.owner = 't' AND gu.group_id = g.id)`? Seems simpler.

tgxworld:
Thank you for reviewing @eviltrout :+1:

PR #4900, opened 2017-06-05
Clarify category tag settings text

barryvan:
https://meta.discourse.org/t/change-to-category-tag-settings-text/63926

This change eliminates the ambiguity in the previous copy, which suggested that the setting prevented the tags from being used in other categories, rather than the actual behaviour of constraining what can be used in the edited category.


nlalonde:
Thanks! :+1:

PR #4902, opened 2017-06-06
Max password length

rimian:
This removes some redundancy when checking user passwords. :)

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -188,7 +188,7 @@ class SessionController < ApplicationController
     params.require(:login)
     params.require(:password)
 
-    return invalid_credentials if params[:password].length > User.max_password_length
+    return invalid_credentials if User.password_too_long?(params[:password])
 
     login = params[:login].strip
     login = login[1..-1] if login[0] == "@"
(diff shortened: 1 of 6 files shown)


ZogStriP:
Thanks @rimian but I'm not sure I see the point of this PR. You are basically testing if `>` works properly in Ruby...

rimian:
Here's a good [example](https://github.com/choiceaustralia/discourse-donations/blob/master/app/controllers/discourse_donations/charges_controller.rb#L18) of why I refactored it. :)

I've removed the duplicated code that needs to be cut and pasted every time someone needs to check if the password is valid.

ZogStriP:
I don't understand how "`User.password_too_long?`" is any better than "`password.size > User.max_password_length`"

rimian:
It isn't any better on its own. But it does occur about 5 times in the code base which is bad code smell.

I refactored these duplicates because it breaks the DRY principle which is: "Every piece of knowledge must have a single, unambiguous, authoritative representation within a system." What we have here is a single calculation being duplicated over and over. Refactor! :)

I found a reference which coincidentally comes from @coding-horror! Funny! [See Point 11](https://blog.codinghorror.com/a-pragmatic-quick-reference/).

I understand this isn't the ideal solution because there is other code that breaks DRY.

SamSaffron:
I am so sorry, we have a policy on this this:

This is not making any functional change to our codebase.

PR #4905, opened 2017-06-07
FIX: Strip zero-width characters from search terms.

tgxworld:
cc @eviltrout Can I get your thoughts on this? I was wondering if we should just handle it client side but there were too many places which touches the search term that it became alot easier to handle it server side.

### DIFF
--- a/app/assets/javascripts/discourse/lib/search.js.es6
+++ b/app/assets/javascripts/discourse/lib/search.js.es6
@@ -94,9 +94,9 @@ export function searchForTerm(term, opts) {
     };
   }
 
-  var promise = ajax('/search/query', { data: data });
+  let promise = ajax('/search/query', { data: data });
 
-  promise.then(function(results){
+  promise.then(results => {
     return translateResults(results, opts);
   });
 
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -161,10 +161,12 @@ class Search
     @limit = Search.per_facet
     @valid = true
 
+    # Removes any zero-width characters from search terms
+    term.to_s.gsub!(/[\u200B-\u200D\uFEFF]/, '')
     term = process_advanced_search!(term)
 
     if term.present?
-      @term = Search.prepare_data(term.to_s)
+      @term = Search.prepare_data(term)
       @original_term = PG::Connection.escape_string(@term)
     end
 
--- a/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
+++ b/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
@@ -186,6 +186,11 @@ export default Ember.Controller.extend({
 
     ajax("/search", { data: args }).then(results => {
       const model = translateResults(results) || {};
+
+      if (results.grouped_search_result) {
+        this.set('q', results.grouped_search_result.term);
+      }
+
       setTransient('lastSearch', { searchKey, model }, 5);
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/discourse/lib/search.js.es6
```diff
@@ -94,9 +94,9 @@ export function searchForTerm(term, opts) {
     };
   }
 
-  var promise = ajax('/search/query', { data: data });
+  let promise = ajax('/search/query', { data: data });
```
jjaffeux: Sorry, I'm still new to es6, is there any reason not to use a `const` here ?

lib/search.rb
```diff
@@ -161,10 +161,12 @@ def initialize(term, opts=nil)
     @limit = Search.per_facet
     @valid = true
 
+    # Removes any zero-width characters from search terms
+    term.to_s.gsub!(/[\u200B-\u200D\uFEFF]/, '')
```
jjaffeux: Wonder if this wouldn’t be better:

```ruby
term.to_s.encode('UTF-8', :invalid => :replace, :undef => :replace)
```

tgxworld: Hmm a zero-width character isn't invalid or undef I think.

```
"test".codepoints.insert(2, 8203).pack("U*").encode("UTF-8", invalid: :replace, undef: :replace).codepoints
```

app/assets/javascripts/discourse/lib/search.js.es6
tgxworld: Good catch :+1:

lib/search.rb
jjaffeux: Indeed you are right

SamSaffron:
A couple of questions... 

1. Why is this a problem?
2. Why not simply do this on the server?

ZogStriP:
> Why not simply do this on the server?

Pretty sure it's done on the server

https://github.com/discourse/discourse/pull/4905/files#diff-d1651f3391b9271699ed9e4d0756712cR165

SamSaffron:
so what problem are we solving here? someone is able to erroneously trigger an ajax req that does nothing if they use magic keyboard skills?

jjaffeux:
I think this is an attempt to fix: https://meta.discourse.org/t/discourse-narrative-bot-beta-feedback/58621/265

app/assets/javascripts/discourse/lib/search.js.es6
eviltrout: BTW there is some disagreement in the JS world about whether to const everything or not. You've probably noticed that I prefer it. The idea is that it avoids bugs where you change things you didn't mean to. Also, in theory it should make JS faster because engines can optimize variables they know don't change.

In practice though, it stops very few bugs and even if we didn't transpile there's no engine where const is faster than let.

I still like it as a way of linting, but if someone prefers `let` I don't mind. As long as it's not `var` I'll accept it happily.

(I still prefer const myself)

eviltrout:
This seems fine to me @tgxworld

app/assets/javascripts/discourse/lib/search.js.es6
Enenche07: 😀😆👍🥰🥰🧗‍♂️🏇🏂

Enenche07 (commented):
Alphanumeric...is so awful

PR #4906, opened 2017-06-07
FIX: add -no-file-changed switch so that backups succeed if files change during backup

pfaffman:
In the unlikely event that a file gets uploaded when tar is running, it fails, as described [here](https://meta.discourse.org/t/backup-fails-if-someone-uploads-during-backup/64129).

Adding `--warning=no-file-changed` to the switches solves this.

I tested by running a backup and restoring it (albeit, to the same instance).


pfaffman:
Ooops. This has lots of other stuff that shouldn't be included.

PR #4907, opened 2017-06-07
FIX: allow tar to finish if files change during backup

pfaffman:
This makes tar not fail if a file is changed during backup.

I tested by making a backup and restoring (albeit to the same instance).


coding-horror:
does this look ok @tgxworld @mpalmer

pfaffman:
See [this post](https://meta.discourse.org/t/backup-fails-if-someone-uploads-during-backup/64129?u=pfaffman) for more info.

PR #4913, opened 2017-06-08
FEATURE: Option to link embed button to the top of the page

rimian:
This allows an embedded post to link back to the topic (the top of the page) instead of the last comment. In my case: this is useful for branding and stuff. 

I'm not sure if I need to create empty translations for the settings.

My branch name is not really relevant.

Fixed: The post limit can be zero.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -3492,7 +3492,7 @@ en:
 
         embed_by_username: "Username for topic creation"
         embed_post_limit: "Maximum number of posts to embed"
-        embed_link_to_top: "Link to the top of the page"
+        embed_link_to_top: "When embedding, the continue button links to the top of the topic instead of the bottom"
         embed_username_key_from_feed: "Key to pull discourse username from feed"
         embed_title_scrubber: "Regular expression used to scrub the title of posts"
         embed_truncate: "Truncate the embedded posts"
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -3492,6 +3492,7 @@ en:
 
         embed_by_username: "Username for topic creation"
         embed_post_limit: "Maximum number of posts to embed"
+        embed_link_to_top: "Link to the top of the page"
         embed_username_key_from_feed: "Key to pull discourse username from feed"
         embed_title_scrubber: "Regular expression used to scrub the title of posts"
         embed_truncate: "Truncate the embedded posts"
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -1071,7 +1071,7 @@ embedding:
     hidden: true
   embed_link_to_top:
     default: false
-    hidden: true
+    hidden: false
   embed_title_scrubber:
     default: ''
     hidden: true
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -1069,6 +1069,9 @@ embedding:
   embed_post_limit:
     default: 100
     hidden: true
+  embed_link_to_top:
+    default: false
+    hidden: true
   embed_title_scrubber:
     default: ''
     hidden: true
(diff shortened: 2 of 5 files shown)


config/site_settings.yml
```diff
@@ -1069,6 +1069,9 @@ embedding:
   embed_post_limit:
     default: 100
     hidden: true
+  embed_link_to_top:
+    default: false
+    hidden: true
```
jomaxro: Curious why this is a hidden site setting?  This seems like something that could be exposed 🤔

rimian: I was just being conservative :) happy to change that

jomaxro: I don't have any sites that embed personally - but this seems like something that would be quite useful to certain types of blogs, so hiding it seemed odd. Not one of the devs though, so just take my comment as an "opinion".

eviltrout: Agreed it needn't be private.

spec/controllers/embed_controller_spec.rb
```diff
@@ -20,11 +20,15 @@
 
     before do
       Fabricate(:embeddable_host)
+      SiteSetting.stubs(:embed_link_to_top).returns(true)
```
eviltrout: You don't need to stub site settings in tests. It's better to do `SiteSetting.embed_link_to_top = true`

spec/controllers/embed_controller_spec.rb
```diff
@@ -20,11 +20,15 @@
...
       controller.request.stubs(:referer).returns('http://eviltrout.com/some-page')
     end
 
     it "allows a topic to be embedded by id" do
-      topic = Fabricate(:topic)
+      topic = Fabricate(:topic) do
```
eviltrout: This could probably be done as `post = Fabricate(:post)` which automatically receives a `topic`. Then if you need the topic do `post.topic`

config/locales/client.en.yml
```diff
@@ -3492,6 +3492,7 @@ en:
 
         embed_by_username: "Username for topic creation"
         embed_post_limit: "Maximum number of posts to embed"
+        embed_link_to_top: "Link to the top of the page"
```
eviltrout: I would suggest a description that indicates what the setting does: "When embedding, the continue button links to the top of the topic instead of the bottom"

spec/controllers/embed_controller_spec.rb
rimian: No worries. I can change that. There's quite a few [examples](https://github.com/discourse/discourse/blob/master/spec/controllers/application_controller_spec.rb#L180) of stubbing out settings already in the code base. Would that be a different use case?

rimian:
Thanks for the feedback @eviltrout

rimian:
I see this is failing because there are missing translations. What should I do here? Should I create place holders so these tests pass?

SamSaffron:
this is a site setting so I think it needs to be localized in server.en as well... @eviltrout do you know why we need these both on client and server localizations?

SamSaffron:
@eviltrout we got to decide what to do here? what is your call?

spec/controllers/embed_controller_spec.rb
```diff
@@ -23,8 +23,20 @@
...
 
+    it "is successful if limit the setting is zero" do
+      SiteSetting.embed_link_to_top = false
+      topic = Fabricate(:topic)
+      SiteSetting.embed_post_limit = 0
+
+      get :comments, topic_id: topic.id
+      expect(response).to be_success
+    end
+
```
gschlager: I think there should be a spec that makes sure it returns the right URL depending the value of `embed_link_to_top`.

SamSaffron:
I want this change, but sadly it has gotten stale and is unmergable now, going to close for now but feel free to open a new tested PR here @rimian  we do want the change.

PR #4917, opened 2017-06-09
Search posts with images

dmacjam:
**User story:** As a user, I want to filter posts in Advanced search results which contains image(s) in order to better filter the results.

Related discussion:
https://meta.discourse.org/t/feature-request-add-includes-image-to-the-advanced-search-options/54103

Searching by images is added as a new filter `in:image`. It uses `image_url` attribute of posts extracted in [`CookedPostProcessor`](https://github.com/discourse/discourse/blob/master/lib/cooked_post_processor.rb#L291-L297) 

It is working for uploaded images, images pasted as links and oneboxed images (tested on Imgur). To support oneboxed images I changed the order of `image_url` extraction to start after oneboxing in `CookedPostProcessor`.


eviltrout:
This seems like a good idea but I don't think `in:images` makes sense as a filter. Shouldn't it be `has:image` or `with:images` or something like that?

lib/cooked_post_processor.rb
```diff
@@ -76,8 +77,6 @@ def post_process_images
       limit_size!(img)
       convert_to_link!(img)
     end
-
-    update_post_image
```
tgxworld: Is there a reason we need this to be in the mutex? I'm assuming there must be a reason it was placed out of the mutex in the first place.

lib/cooked_post_processor.rb
```diff
@@ -97,8 +96,8 @@ def extract_images_for_post
     # all image with a src attribute
     @doc.css("img[src]") -
     # minus, emojis
-    @doc.css("img.emoji") -
-    # minus, image inside oneboxes
+    @doc.css("img.emoji") +
+    # plus, image inside oneboxes
     oneboxed_images -
```
nlalonde: You don't need to add the `oneboxed_images` because they're already found by `@doc.css("img[src]")`.

spec/components/search_spec.rb
```diff
@@ -628,6 +628,17 @@ def search
...
 
+    it 'can find posts with images' do
+      post_uploaded = Fabricate(:post_with_uploaded_image)
+      post_with_image_urls = Fabricate(:post_with_image_urls)
+      Fabricate(:post)
+
+      CookedPostProcessor.new(post_uploaded).update_post_image
+      CookedPostProcessor.new(post_with_image_urls).update_post_image
+
+      expect(Search.execute('in:image').posts.map(&:id).sort).to eq([post_uploaded.id, post_with_image_urls.id].sort)
```
tgxworld: Instead of sorting, you can use the [contain_exactly](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher) matcher

dmacjam: Thank you, it's more easier to read ;)

lib/cooked_post_processor.rb
dmacjam: The idea is that `update_post_image` must be called after `post_process_images` and after oneboxing is applied (`post_process_oneboxes`) to be able to extract `image_url`  for oneboxed images. It doesn't need to be inside the mutex, it can be placed also at the end of the `post_process_oneboxes` method. However, I think this way it is more cleaner how post is being postprocessed.

dmacjam:
I updated the pull request based on your suggestions and changed the filter name to `with:images`.

lib/cooked_post_processor.rb
tgxworld: If there isn't a good reason to put it inside the mutex, I would prefer not to make this change since we don't know why it was taken out of the mutex in the first place.

dmacjam: Ok, I've put it inside the `post_process_oneboxes`

nlalonde:
Looks good. There are conflicts though. Can you rebase and squash commits?

PR #4925, opened 2017-06-14
Makes sure we clear tonable_emojis cache

jjaffeux:


app/models/emoji.rb
```diff
@@ -60,6 +60,7 @@ def self.clear_cache
     Discourse.cache.delete(cache_key("standard_emojis"))
     Discourse.cache.delete(cache_key("aliases_emojis"))
     Discourse.cache.delete(cache_key("all_emojis"))
+    Discourse.cache.delete(cache_key("tonable_emojis"))
```
ZogStriP: At some point we should start having an array of the cache keys and iterate over them and call `Discourse.cache.delete`

jjaffeux: what's your opinion on this ? To avoid duplicating all the key names, what should we do? Something like this ?

```ruby
CACHE_KEYS = {
  standard_emojis: "standard_emojis",
  aliases_emojis: "aliases_emojis",
  all_emojis: "all_emojis",
  tonable_emojis: "tonable_emojis"
}

def self.clear_cache
  CACHE_KEYS.values.each { |value| Discourse.cache.delete(cache_key(value)) }
end

def self.tonable_emojis
  Discourse.cache.fetch(cache_key(CACHE_KEYS[:tonable_emojis])) { db['tonableEmojis'] }
end

###
```

jjaffeux: It certainly would be enough, it was just an opportunity to avoid repeating the "string" at multiple places, but your solution totally works for me.

jjaffeux: I will work on the PR

ZogStriP: No rush ;)

PR #4929, opened 2017-06-14
Improves native emoji conversion

jjaffeux:


app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -7,7 +7,7 @@ let _unicodeRegexp;
 export function setUnicodeReplacements(replacements) {
   _unicodeReplacements = replacements;
   if (replacements) {
-    _unicodeRegexp = new RegExp(Object.keys(replacements).join("|"), "g");
+    _unicodeRegexp = new RegExp(Object.keys(replacements).sort().reverse().join("|"), "g");
```
ZogStriP: Why the `.reverse()`?

jjaffeux: we want the longer emojis sequences to be match first:

👩🏾‍🎓  before 👩🏾‍  before 👩

does it make sense ?

ZogStriP: It does. Might be worth adding a comment in the code

jjaffeux: https://github.com/discourse/discourse/pull/4930

Thanks for feedback

PR #4932, opened 2017-06-15
FIX: Validate interpolation keys used in translation overrides.

tgxworld:
https://meta.discourse.org/t/discobot-translation-missing-error/64429/6?u=tgxworld

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -409,6 +409,11 @@ en:
           attributes:
             execute_at:
               in_the_past: "must be in the future."
+        translation_overrides:
+          attributes:
+            value:
+              invalid_interpolation_keys: 'The following interpolation key(s) are invalid: "%{keys}"'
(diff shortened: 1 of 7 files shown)


config/locales/server.en.yml
```diff
@@ -409,6 +409,11 @@ en:
           attributes:
             execute_at:
               in_the_past: "must be in the future."
+        translation_overrides:
+          attributes:
+            value:
+              invalid_interpolation_keys: 'The following interpolation key(s) are not : %{keys}'
```
ZogStriP: not what? not valid? invalid? not used?

tgxworld: Ah good catch!

tgxworld:
Merged in https://github.com/discourse/discourse/commit/b5ec241716b6fb09598e9e82080a3b5a2f645473

tgxworld:
Thanks for reviewing @ZogStriP

PR #4934, opened 2017-06-16
RTL digest emails

hosnas:
Allows RTL digest emails that fixes [this issue](https://meta.discourse.org/t/emails-for-rtl-languages-have-ltr-direction/41972).
For this to work, I have done the following major additions:
1- adding appropriate direction to body and tables
2- making text-direction compatible with rtl
3- making float, padding, and margin tags compatible with rtl


nlalonde:
Thanks for fixing this @hosnas

PR #4935, opened 2017-06-17
FIX: version check

albaguirre:
The api.discourse.org server apparently does not like GET
requests that have a port number in the headers Host entry.

Additionally fix the actual GET query; it should use query parameters not a body.

### DIFF
--- a/lib/discourse_hub.rb
+++ b/lib/discourse_hub.rb
@@ -42,8 +42,9 @@ module DiscourseHub
   def self.singular_action(action, rel_url, params={})
     JSON.parse(Excon.send(action,
       "#{hub_base_url}#{rel_url}",
-      body: params.to_query,
-      headers: { 'Referer' => referer, 'Accept' => accepts.join(', ') }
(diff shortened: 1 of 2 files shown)


dwhagar:
I'd just like to say that this solves the problem perfectly.

coding-horror:
sure @tgxworld  can review

dwhagar:
This may cause an error that I was initially unaware of.

    Job exception: undefined method `>' for nil:NilClass

albaguirre:
@dwhagar , looks like the query itself was also incorrect. I fixed that issue too now.

dwhagar:
I just checked not only that the little sad face turns to a happy face in the admin panel, but that the sidekiq scheduler, when you trigger it manually, also has Jobs::VersionCheck go from FAILED to OK after both fixes have been applied.

Thank you @albaguirre, this was driving me crazy.

tgxworld:
Thanks for fixing @albaguirre :+1:

PR #4936, opened 2017-06-19
Optimize import script - vanilla_mysql.rb

meldsza:
Use WHERE id > last_id instead of OFFSET because OFFSET causes performance issues


ZogStriP:
Thanks 👍

PR #4938, opened 2017-06-19
FEATURE: setting to only use the key when finding the related post of…

LeoMcA:
… an email reply

this fixes email-in threading problems when using a SMTP server which modifies the message_id
header, like Amazon SES

https://meta.discourse.org/t/email-in-replies-thread-wrongly/59365/17?u=leomca


coding-horror:
What do you think of this @zogstrip ?

ZogStriP:
Looks good. Thanks 👍

PR #4941, opened 2017-06-23
Emoji ordering

jjaffeux:
This commit uses
http://www.unicode.org/emoji/charts/emoji-ordering.html to generate an
automatic ordering of emojis. We also don't have to maintain a full
emoji list for groups but only sub sections.


SamSaffron:
👍

PR #4942, opened 2017-06-27
Added plugin outlet near editor's preview.

nbianca:
This is required for the updates I made to canned replies plugin from this [PR](https://github.com/discourse/discourse-canned-replies/pull/22).

This should help plugins display stuff using the preview pane.

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/d-editor.hbs
+++ b/app/assets/javascripts/discourse/templates/components/d-editor.hbs
@@ -28,7 +28,8 @@
 
   <div class="d-editor-preview-wrapper {{if forcePreview 'force-preview'}}">
     <div class="d-editor-preview">
-      {{{preview}}}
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/templates/components/d-editor.hbs
```diff
@@ -28,7 +28,8 @@
 
   <div class="d-editor-preview-wrapper {{if forcePreview 'force-preview'}}">
     <div class="d-editor-preview">
-      {{{preview}}}
+      <div class="d-editor-cooked">{{{preview}}}</div>
+      <div class="d-editor-plugin">{{plugin-outlet name="editor-preview"}}</div>
```
SamSaffron: do we need the extra `d-editor-plugin` div here? 

I think `plugin-outlet name="editor-preview"` will already wrap it in a div... 

Going to merge this in for now, but perhaps a PR to clean this up and unwrap is in order.

PR #4943, opened 2017-06-28
Include HTTP Host header in nginx logs

mpalmer:
This is crucial in multisite installations, because otherwise the nginx logs
are fairly useless, however it can also be quite handy to know what
hostnames are being sent to your site.  The variable is quoted, because it
is untrusted input (it is taken directly from the HTTP request), but nginx
helpfully escapes the quoting character automagically, so we don't have to
worry about that.

For now, the log analysis plugin *recognises* the new log format
(and continues to recognise the previous format, for backwards
compatibility), but doesn't do anything with the new log entry field.  This
means your multisite performance plugin data is still broken, but it's no
worse than it was before.

### DIFF
--- a/config/nginx.sample.conf
+++ b/config/nginx.sample.conf
@@ -25,7 +25,7 @@ map $http_x_forwarded_proto $thescheme {
   https https;
 }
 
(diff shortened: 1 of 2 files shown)


begrif:
In the PATTERN, why are some quotes backslash escaped and some not? `\"` should be the same as `"`, unless I'm misreading it.

PR #4947, opened 2017-06-30
Mybb permalinks

pfaffman:
Add permalink support (user, category, topic, *not* posts) to mybb importer.

Also add configure database and table prefix via environment variables

### DIFF
--- a/script/import_scripts/mybb.rb
+++ b/script/import_scripts/mybb.rb
@@ -37,6 +37,7 @@ class ImportScripts::MyBB < ImportScripts::Base
   end
 
   def execute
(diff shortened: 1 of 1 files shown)


pfaffman:
This is how most other importers handle permalinks.

Arguably, permalnks should be disabled by default, though. I'll add an environment variable tweak for that if desired.

nlalonde:
I don't think we need an env variable for disabling permalinks. The method call can be removed before running the script, or run `DELETE FROM permalinks` afterwards if you don't want them.

Thanks for adding this!

PR #4950, opened 2017-07-04
FEATURE: emoji generation now uses emoji-db

jjaffeux:
- unicode 10
- emoji 5.0
- support for google_blob and facebook_messenger sets


SamSaffron:
yay 🥇

PR #4951, opened 2017-07-05
FIX: ensure we clear emoji cache before recompilation

jjaffeux:
I'm not 100% sure it's the root cause, but in my tests it seems to fix the issue.


techAPJ:
Seems safe! Merging in.

PR #4952, opened 2017-07-05
Allow optional import_mode param for posts in api

mudasirraza:
This should allow a new post to be created in import mode, which will be useful for creating topics in bulk without notifying users and all such sort of things.

relevant discussion:
https://meta.discourse.org/t/disable-count-pop-up-notifications/65543

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -571,6 +571,9 @@ class PostsController < ApplicationController
       params[:skip_validations] = params[:skip_validations].to_s == "true"
       permitted << :skip_validations
 
+      params[:import_mode] = params[:import_mode].to_s == "true"
+      permitted << :import_mode
+
       # We allow `embed_url` via the API
       permitted << :embed_url
 
--- a/spec/controllers/posts_controller_spec.rb
+++ b/spec/controllers/posts_controller_spec.rb
@@ -631,6 +631,26 @@ describe PostsController do
 
         expect(response.body).to eq(original)
       end
+
+      it 'allows to create posts in import_mode' do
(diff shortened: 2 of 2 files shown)


ZogStriP:
Awesome. Would you mind adding a test for that?

mudasirraza:
@ZogStriP test case has been added

SamSaffron:
We got to finalize this, the change looks fine, but I am not following how the "rest" of the change will work. 

cause as it stands nothing is checking for import mode and checking for it would introduce a security risk cause it would allow end users to "secret in" topics.

ZogStriP:
I agree, this params should **at least** only be permitted for API calls.

mudasirraza:
Sorry if i'm missing something here. This param is in fact only for API calls. I've included it in `is_api?` block. I'm not sure about security risk here, this param is permitted here same like skip_validations.

ZogStriP:
Might bad. I 👀-balled it too quickly

SamSaffron:
Sorry, my point is ... where is this import_mode thing consumed? How is it consumed?

mudasirraza:
These params are passed via [new_post_manager](https://github.com/discourse/discourse/blob/91f0f76fb1a40bbd4dd6dd46dd4e5bb9663d0f82/lib/new_post_manager.rb#L182) into [post_creator](https://github.com/discourse/discourse/blob/master/lib/post_creator.rb) where it is consumed at several places.
It is same as if creating posts in [import scripts](https://github.com/discourse/discourse/blob/master/script/import_scripts/base.rb#L477)

SamSaffron:
no probs, sure looks safe to me ... thanks

PR #4958, opened 2017-07-07
FEATURE: Search posts by filetype

dmacjam:
**User story:** As a user, I want to filter posts in Advanced search which contains links to files or uploads, e.g. `filetype:pdf` lists all posts with pdf files.

This is a power user feature as UI for inputting file extensions will make the advanced search page more complicated. 

_Related discussion:_ https://meta.discourse.org/t/feature-request-add-includes-image-to-the-advanced-search-options/54103
_Related pull request with search for posts with images:_ https://github.com/discourse/discourse/pull/4917

The implementation idea is to extract file extension and persist it into database for both url links and uploads. Therefore, new columns with indices were added to `TopicLinks` and `Uploads`. Then searching by filetypes at the backend is a combination of searching these columns (distinct search)


app/jobs/onceoff/migrate_upload_extensions.rb
```diff
@@ -0,0 +1,11 @@
+module Jobs
+
+  class MigrateUploadExtensions < Jobs::Onceoff
+    def execute_onceoff(args)
+      Upload.all.each do |upload|
```
ZogStriP: This is going to 💥 on Discourses with a large number of uploads. It's better to use `Upload.find_each`

dmacjam: Thanks for pointing this, batch loading is much better. :+1:

lib/search.rb
```diff
@@ -456,6 +456,19 @@ def self.advanced_filters
...
+    file_extensions = match.split(",")
+
+    posts.where("posts.id IN (
+      SELECT post_id FROM topic_links
+      WHERE extension IN (?)
+      UNION
+      SELECT post_uploads.post_id FROM uploads
+      JOIN post_uploads ON post_uploads.upload_id = uploads.id
+      WHERE lower(uploads.extension) IN (?)
+      )", file_extensions, file_extensions)
```
tgxworld: You can use `:file_extensions` in the SQL and just pass a single option `{ file_extensions: file_extensions }`

spec/components/search_spec.rb
```diff
@@ -703,6 +703,25 @@ def search
...
+
+    it "can find posts which contains filetypes" do
+      post1 = Fabricate(:post,
+                        raw: "http://example.com/image.png")
+      post2 = Fabricate(:post,
+                         raw: "Discourse logo\n"\
+                              "http://example.com/logo.png\n"\
+                              "http://example.com/vector_image.svg")
+      post_with_upload = Fabricate(:post)
+      post_with_upload.uploads = [Fabricate(:upload)]
```
tgxworld: This can be simplified to `Fabricate(:post, uploads: [Fabricate(:upload)])`

spec/components/search_spec.rb
```diff
@@ -703,6 +703,25 @@ def search
...
+      Fabricate(:post)
+
+      TopicLink.extract_from(post1)
+      TopicLink.extract_from(post2)
+
+      expect(Search.execute('filetype:svg').posts.map(&:id)).to eq([post2.id])
```
tgxworld: Why do we need to map the `id` here? I believe comparing against the post record should work just fine.

dmacjam: Comparing by id is faster, isn't it?

tgxworld: It might be but I readability is more important here.

SamSaffron:
looks like we have a conflict now that needs resolving.

dmacjam:
There is actually no conflict (only new test added) so it is very easy to resolve.

PR #4959, opened 2017-07-07
Add a nil check on the connection before attempting to exec it

farmdawgnation:
Ran into an issue while trying to [import mbox files](https://meta.discourse.org/t/howto-import-mbox-mailing-list-files/51233) that I tracked back to this line. I'm not an avid rubyist so I'm not entirely sure this is the Right Solution™, but it got my import running after I made this change.

Figured that others would probably run into the same issue so I thought I'd submit a PR to call attention to the issue, if nothing else. =)

Cheers!

### DIFF
--- a/script/import_scripts/base.rb
+++ b/script/import_scripts/base.rb
@@ -203,7 +203,7 @@ class ImportScripts::Base
       return true
     end
   ensure
-    connection.exec('DROP TABLE import_ids')
+    connection.exec('DROP TABLE import_ids') unless connection.nil?
   end
 
   def created_user(user)


coding-horror:
Is this ok @ZogStriP ?

ZogStriP:
It would probably be better to change this to use a SQL transaction instead and add the `ON COMMIT DROP` option to the `TEMP TABLE`.

farmdawgnation:
@ZogStriP I don't think that would fix the problem here, which is that `connection` is `Nil`.

ZogStriP:
Why is `connection` `nil` then?

farmdawgnation:
That I don't know. It would seem that `ActiveRecord::Base.connection.raw_connection` likely returns `nil` but I wasn't able to try down why that might be happening. I spent some time digging through the Rails docs trying to figure out how that could happen and `puts`'ing a few things but I was quite befuddled at what was actually going on in this code.

My best guess is that this method was being executed before the database connection was fully set up? But once again, I had issues disentangling what was actually going on here.

tgxworld:
This looks good to me. `connection` may be `nil` if `import_ids` is empty.

PR #4960, opened 2017-07-07
FIX: Explicit error when category description post is bad

xfalcox:


tgxworld:
Tests? :)

tgxworld:
ping @xfalcox

xfalcox:
Added tests @tgxworld :smile:

spec/components/post_revisor_spec.rb
```diff
@@ -213,6 +213,21 @@
...
       end
 
+      context "invalid description without paragraphs" do
+        before do
+          subject.revise!(post.user, { raw: "# This is a title" })
+          category.reload
+        end
+
+        it "returns a error for the user" do
+          expect(post.errors.present?).to eq(true)
```
tgxworld: We should test for the description of the error otherwise another error may cause this test to incorrectly pass.

xfalcox: Done!

tgxworld:
@xfalcox LGTM :+1:

PR #4961, opened 2017-07-08
SECURITY: when redeeming invite link verify that user does not exist with that email

techAPJ:


techAPJ:
Discussing this issue with team, in the meantime @eviltrout has already pushed out a fix.

attritionorg:
Since this is flagged SECURITY, can you summarize the scenario/risk for this?

coding-horror:
We don't do discussion here, ask at meta.discourse.org

PR #4962, opened 2017-07-08
Added events to show and hide the preview pane.

nbianca:
The plugin-outlet defined in the preview is useless if the preview pane is hidden.

I consider that plugins should may be able to show and hide the preview pane.

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-editor.js.es6
@@ -33,12 +33,10 @@ export default Ember.Component.extend({
 
     this.appEvents.on('composer:show-preview', () => {
       this.set('showPreview', true);
-      this.keyValueStore.set({ key: 'composer.showPreview', value: true });
     });
 
     this.appEvents.on('composer:hide-preview', () => {
       this.set('showPreview', false);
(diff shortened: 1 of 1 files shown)


SamSaffron:
Do we need an `off` for these events when you hide the composer? I don't want to keep the event listening around if the composer is not visible on the page (or even if the composer is minimized)

nbianca:
I made the edits.

SamSaffron:
One last style thing, I think it is better to have a central spot for changing key value store as opposed to having it in 3 places. 

I would just observe when showPreview changes and amend it based of that. It means there will only be one spot that amends the store which makes the code more robust.

SamSaffron:
excellent, thanks!

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -42,6 +50,11 @@ export default Ember.Component.extend({
     return showPreview ? I18n.t('composer.hide_preview') : I18n.t('composer.show_preview');
   },
 
+  @observes('showPreview')
```
techAPJ: This resulted in a failing test. Fixed via: https://github.com/discourse/discourse/commit/3fb1c21dd5fbc17fda9e4c75953d19ed8009798e

:)

PR #4963, opened 2017-07-09
Added event trigger for file uploads.

nbianca:
This will be used for [this plugin](https://github.com/nbianca/discourse-gif-to-mp4).

### DIFF
--- a/lib/upload_creator.rb
+++ b/lib/upload_creator.rb
@@ -27,6 +27,11 @@ class UploadCreator
     @opts = opts
   end
 
+  attr_accessor :upload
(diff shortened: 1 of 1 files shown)


SamSaffron:
Can we use this instead? 

https://github.com/discourse/discourse/blob/master/lib/cooked_post_processor.rb#L38 

I think we want to retain the old upload, and just recompress and replace with the new non-gif version after the fact

nbianca:
That makes sense. I will look into it.

SamSaffron:
ok, closing this for now, let me know how you go

PR #4964, opened 2017-07-10
For rake plugin:spec, only load ruby files ending in _spec.rb

davidtaylorhq:
This matches the default behaviour of rspec, as per

> When you run RSpec without giving it specific file names, it determines which
> files to load by applying a pattern to the provided directory arguments or
> spec (if no directories are provided). By default, RSpec uses the following
> pattern:
> 
> `"**{,/*/**}/*_spec.rb"`

[docs](https://relishapp.com/rspec/rspec-core/docs/command-line/pattern-option)


SamSaffron:
Sure, seems ok to me, thanks!

PR #4967, opened 2017-07-11
Override site settings based on locale

fantasticfears:


spec/fixtures/site_settings/locale_override.yml
```diff
@@ -0,0 +1,14 @@
+basic:
+  default_locale:
+    default: 'en'
+    enum: 'LocaleSiteSetting'
+
+search:
+  min_search_term_length:
+    client: true
+    default:
+      development: 1
```
SamSaffron: I am not a fan of this part of the feature. Let's take it out, default per environment is a recipe for confusion and bugs.

SamSaffron: I prefer the syntax: 

```
default: 'test'
locale_default: 
   zh_CN: 1
```

It is a bit more verbose but I feel it is a lot clearer.

SamSaffron:
Need to do a test for cache invalidation when you switch default locale.

spec/fixtures/site_settings/locale_override.yml
fantasticfears: Then I'll just push the existing overrides under `config/initializers/005-site_settings.rb`. And what about `hidden` property? There are only 4 settings under `developer` which are different than default. I guess it's not a problem to see those in development or test either. I can also do a sweep by the way.

SamSaffron: Yeah I see the problem cause now you can specify per env and per locale defaults, resolution order is confusing... 

I guess, 

per-env is priority 1 default
per-locale is priority 2 default
default is priority 3

codifying this in a config file is somewhat tricky, can you do a dev topic on meta with possible resolutions here, we need to sort out the "right" format here.

fantasticfears: Here it is: https://meta.discourse.org/t/sitesetting-file-format-update-for-locale-override-and-hidden-params-based-on-rails-env/66254

fantasticfears:
@SamSaffron Could you help with reviewing please?
@techAPJ @tgxworld I saw your commits around this area. Could you help with reviewing please if you have some time?

config/site_settings.yml
```diff
@@ -1326,7 +1310,7 @@ user_preferences:
   default_email_mailing_list_mode: false
   default_email_mailing_list_mode_frequency:
     enum: 'MailingListModeSiteSetting'
-    default: 0
+    default: 1
```
SamSaffron: why was this changed?

SamSaffron:
My #1 concern here is performance, can you confirm lookups are still just as fast as they used to be. 

Overall the PR looks good, lots of tests and code looks well organised

config/site_settings.yml
fantasticfears: In `MailingListModeSiteSetting`, only 1 and 2 are valid values. 0 is removed 2 months ago

fantasticfears:
Sure, I'll look into that. What's the best way to do it? Profile env or a benchmark?

SamSaffron:
I would just benchmark, mainly concerned about read, I assume you are still
bolting the methods onto SiteSetting and none of that changed?

fantasticfears:
No, none of them have changed. I extracted the type check and replaced defaults hash for a new class. In theory, it should have no problem. I'll go check with benchmark.

fantasticfears:
Slightly hit on performance ~ -2% / -25% / -12%

Before
Calculating -------------------------------------
current cache    400.339k (± 4.5%) i/s -      2.030M in   5.081421s
change default locale with current cache refreshed
181.503  (± 5.0%) i/s -    918.000  in   5.071343s
change site setting    172.146  (± 4.1%) i/s -    864.000  in   5.028832s


After
Calculating -------------------------------------
current cache    389.542k (± 4.5%) i/s -      1.953M in   5.022523s
change default locale with current cache refreshed
136.545  (± 5.1%) i/s -    686.000  in   5.039420s
change site setting    151.618  (± 4.6%) i/s -    765.000  in   5.056182s

SamSaffron:
build appears to be failing for legit reasons, can you have a look? 

My aim is to merge this in next week

fantasticfears:
Rebased and ready to go

SamSaffron:
why is there a hit on perf?  this does concern me

fantasticfears:
I ran the profiler and double check the bench. The profiler gives me even better result comparing to the old. AR code takes much more time so it's I/O related.

The second bench runs fine:

```
Calculating -------------------------------------
       current cache    391.790k (± 5.5%) i/s -      1.962M in   5.022182s
change default locale with current cache refreshed
                        147.401  (± 4.7%) i/s -    742.000  in   5.045432s
 change site setting    175.085  (± 4.6%) i/s -    884.000  in   5.059692s
```

Read/Write is within a reasonable variations (~5%). Changing default site locale should takes a bit more time since the `defaults` cache is expired and rebuilt as well as the current to yield expected behavior.

fantasticfears:
To be extreme cautious, I ran the benchmark in the production docker.

```Before
Before
Calculating -------------------------------------
current cache    167.518k (±12.1%) i/s -    822.983k in   5.000478s
change default locale with current cache refreshed
174.173  (±16.7%) i/s -    845.000  in   5.015281s
change site setting    132.956  (±16.5%) i/s -    663.000  in   5.124766s

After
Calculating -------------------------------------
current cache    167.170k (±12.2%) i/s -    824.688k in   5.022784s
change default locale with current cache refreshed
79.876  (±16.3%) i/s -    392.000  in   5.067448s
change site setting    129.085  (±13.2%) i/s -    636.000  in   5.032536s
```

I've introduced a change behaviour since I added `email` and `username` type into the grid. It will changes the data type in the db like `notification_email`.

SamSaffron:
Thanks! 

Merging this in now!

davidtaylorhq:
@fantasticfears @SamSaffron since this was merged I'm seeing some test failures in my plugin's specs.

I have a site setting defined like this
```
plugins:
    some_setting_string:
        hidden: true
```
But then when I try and use it, any values assigned don't persist. For example on the rails console:

```
[1] pry(main)> SiteSetting.some_setting_string = "hello world"
  SiteSetting Load (0.5ms)  SELECT  "site_settings".* FROM "site_settings" WHERE "site_settings"."name" = 'some_setting_string' LIMIT 1  [["name", :some_setting_string]]
   (0.1ms)  BEGIN
  SQL (0.9ms)  INSERT INTO "site_settings" ("name", "value", "data_type", "created_at", "updated_at") VALUES ('some_setting_string', 'hello world', 1, '2017-08-03 15:12:00.725952', '2017-08-03 15:12:00.725952') RETURNING "id"  [["name", "some_setting_string"], ["value", "hello world"], ["data_type", 1], ["created_at", Thu, 03 Aug 2017 15:12:00 UTC +00:00], ["updated_at", Thu, 03 Aug 2017 15:12:00 UTC +00:00]]
   (0.4ms)  COMMIT
=> "hello world"
[2] pry(main)> SiteSetting.some_setting_string
=> nil
```

If I change the definition to
```
plugins:
    some_setting_string:
        default: ''
        hidden: true
```
then it works fine. 

So as far as my plugin goes, this is fine, I'll just specify the default. 

I'm assuming defaults are now required, so can we make it throw an exception if a default isn't specified? The silent failure made it hard to track down the source of the issue.

fantasticfears:
Good point, I'll included that later. Thanks for your report. I prefer the explicit value that always give your some type hint sense.

SamSaffron:
Yeah us guessing a default is not really supported you got to specify one

davidtaylorhq:
Yep, requiring a default totally makes sense. It would just be good if it failed loudly

fantasticfears:
#5024 should fix this. @davidtaylorhq

PR #4971, opened 2017-07-11
Load plugins for qunit travis tests

davidtaylorhq:
The tests will fail for this PR, because of all the broken qunit tests that are currently being ignored by travis.

See https://meta.discourse.org/t/plugin-qunit-tests-are-not-running-as-part-of-rake-qunit-test/59577/6?u=david_taylor


tgxworld:
Awesome! This is great!

PR #4972, opened 2017-07-11
Fix for the last broken discourse-details qunit test

davidtaylorhq:
See https://meta.discourse.org/t/plugin-qunit-tests-are-not-running-as-part-of-rake-qunit-test/59577/6?u=david_taylor


tgxworld:
Thank you for taking the time to fix this :+1:

PR #4973, opened 2017-07-12
Allow docker test rake task to be used for single plugins

davidtaylorhq:
This is useful for adding continuous integration to individual plugin repositories. For example, I've just set up discourse-chat-integration to run its tests on travis automatically. (Currently it applies this patch manually).

This change adds support for a `SINGLE_PLUGIN` environment variable. If set, it only runs rspec tests for that specific plugin. If left unset, everything continues to function as it currently does.

(Specific qunit tests can already be run using the `MODULE` or `FILTER`  environment variables)

[example travis file making use of this feature](https://github.com/discourse/discourse-chat-integration/blob/master/.travis.yml)

[working test](https://travis-ci.org/discourse/discourse-chat-integration/builds/252770221#L6970)


SamSaffron:
This is fine, but let's add some documentation for this somewhere first.

davidtaylorhq:
Where would you like the documentation? As a comment at the top of the file? I think all of the environment variables is probably a bit long to fit in the description line
```
desc 'Run all tests (JS and code in a standalone environment)'
```

davidtaylorhq:
@SamSaffron I've added some documentation as comments, but maybe there's a better place for them to go?

Once this is merged I'll also update my [meta topic](https://meta.discourse.org/t/setting-up-plugin-continuous-integration-tests-on-travis-ci/59612) on plugin CI testing with the details.

davidtaylorhq:
@SamSaffron I've updated this so it works with your new SKIP_CORE env variable, and also added the new ENV variables to the documentation at the top

lib/tasks/docker.rake
```diff
@@ -1,3 +1,27 @@
+# rake docker:test is designed to be used inside the discourse/docker_test image
+# running it anywhere else will likely fail
+#
+# Environment Variables (specific to this rake task)
+# => SKIP CORE                 set to 1 to skip core rspec tests
```
tgxworld: Missing underscore here?

tgxworld:
Thank you for all your work so far :+1:  @davidtaylorhq

SamSaffron:
@davidtaylorhq  awesome! a big one missing is the ability to run all qunit tests for ALL plugins (without core) can you swing adding that?

davidtaylorhq:
I'm not sure how to detect which qunit tests are for plugins or for core - as far as I can see qunit offers filtering by "Module" or keyword... nothing for test file path.

Any ideas?

SamSaffron:
yeah ... that has been my conundrum here as well :) 

I have no idea how to swing this but totally need it, maybe core can be hacked somehow to allow this?

davidtaylorhq:
Ok cool, I'll have a look in more detail next week.

davidtaylorhq:
For reference: second PR is here https://github.com/discourse/discourse/pull/4985

PR #4974, opened 2017-07-12
FIX: name can be searched by mention autocomplete

fantasticfears:


spec/models/user_search_spec.rb
```diff
@@ -12,6 +12,7 @@
   let(:user4)     { Fabricate :user, username: "mrpink",   name: "Steve Buscemi",  last_seen_at: 7.days.ago }
   let(:user5)     { Fabricate :user, username: "mrbrown",  name: "Quentin Tarantino", last_seen_at: 6.days.ago }
   let(:user6)     { Fabricate :user, username: "mrwhite",  name: "Harvey Keitel",  last_seen_at: 5.days.ago }
+  let(:user7)     { Fabricate :user, name: "Régis" }
```
ZogStriP: Awesome. Can you also add a test for a Chinese name? (ie. non-latin-based language)

fantasticfears: Yes. From what I've experienced, debouncing for autocomplete seems bumped. But it's another issue.

fantasticfears:
It seems qunit failure is not relevant to the PR...

fantasticfears:
This PR is ready now

SamSaffron:
hmmm ... are username matches higher precedence than name matches? what about all the rules about who shows up first (participants ... etc)

fantasticfears:
I'll look into that

lib/query_limiter.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def initialize(limit)
+    @limit = limit
+  end
+
+  def step(&blk)
+    @limit -= yield(@limit).to_i if @limit > 0 && blk
+    self
+  end
+end
```
SamSaffron: this to me feels like way too small a piece to require its own file in lib

rriemann:
What is missing? Can we merge this after two months?

fantasticfears:
Search part got another polish. Once Rails 5 is merged, we are good to go. It's getting there!

fantasticfears:
This PR needs some love. I didn't get many luck here. PG search function are used. But it's quite hard to merge AR Relation with Arel::SelectManager. At least, I couldn't find a viable way to integrate them.
The problem as of this time is that after merge (:60), AR query is wiped out.
I believe we can build this query with Arel and finally make this work. It would require us to rewrite `scoped_users` which introduces repetitions. For example, `users.not_suspended` has to be rewritten in Arel. On the other hand, hand-written SQL will be quite complicated too since the complexity of this module.
Such changes may be conflicted with our code standard. So I need a call here.

rriemann:
Has @SamSaffron answered your call? :)

SamSaffron:
I am putting this on hold for now, @jjaffeux  is working on adding richer support for usernames so they can potentially include unicode, he will review when he gets to it and look at implementing

PR #4975, opened 2017-07-15
initial pass at color simplification

awesomerobot:
This very much simplifies the dark-light-diff into a sane number of variables based on contrast intensity 

e.g., 

$primary = (darkest grey in the case of the default theme)
$primary-medium = low contrast to $secondary (lightest grey in the case of the default theme)
$primary-low = medium contrast to $secondary  (medium grey in the case of the default theme)

See variables.scss for the full list. 

I tested this against all the test content I could come up with locally, but would be worth testing with an active forum because it stripped out all the existing dark-light-diffs.


coding-horror:
Nice!!

coding-horror:
Can you herd this forward this next week @eviltrout ? We need to test dark and light themes too.

eviltrout:
This is really hard to review , but seems reasonable. I'll merge and we can revert if there are issues.

PR #4976, opened 2017-07-15
FIX: user is nil in populate command

dmacjam:
If user exists, it's nil because of missing assignment in populate command.


ZogStriP:
Thanks

PR #4977, opened 2017-07-17
FEATURE: phase 1 of supporting multiple email addresses

tgxworld:


tgxworld:
Main ruby tests are passing but our plugin tests are broken.

app/models/invite.rb
```diff
@@ -32,7 +32,7 @@ class UserExists < StandardError; end
   def user_doesnt_already_exist
     @email_already_exists = false
     return if email.blank?
-    u = User.find_by("email = ?", Email.downcase(email))
+    u = User.find_by_email(Email.downcase(email))
```
jjaffeux: Why are we doing `Email.downcase` two times ? (multiple instances of this in the commit)

https://github.com/discourse/discourse/blob/master/app/models/user.rb#L235

app/models/invite.rb
```diff
@@ -98,11 +98,9 @@ def self.create_invite_by_email(email, invited_by, opts=nil)
     group_ids = opts[:group_ids]
     send_email = opts[:send_email].nil? ? true : opts[:send_email]
     custom_message = opts[:custom_message]
-
     lower_email = Email.downcase(email)
```
jjaffeux: Shouldn’t be needed? (see other comments)

app/models/invite.rb
tgxworld: good catch! :+1:

app/models/email_token.rb
```diff
@@ -71,7 +71,7 @@ def self.confirm(token)
       end
 
       if user
-        return User.find_by(email: Email.downcase(user.email)) if Invite.redeem_from_email(user.email).present?
+        return User.find_by_email(Email.downcase(user.email)) if Invite.redeem_from_email(user.email).present?
```
jjaffeux: `Email.downcase` probably not needed here

PR #4979, opened 2017-07-18
FEATURE: New emoji-picker

jjaffeux:
This commit introduces a brand new emoji picker for Discourse:

- faster navigation
- filtering of emojis
- dropping modal on large screen to allow for multiple emojis selection
- updated design
- persistent state


app/assets/javascripts/discourse/components/emoji-picker.js.es6
```diff
@@ -12,41 +12,30 @@ export const EMOJI_USAGE = "emojiUsage";
 export const EMOJI_SCROLL_Y = "emojiScrollY";
 export const EMOJI_SELECTED_DIVERSITY = "emojiSelectedDiversity";
 const PER_ROW = 11;
+const customEmojis = _.map(_.keys(extendedEmojiList()), function(code) {
```
ZogStriP: I think you can write it this way in ES6

```
const customEmojis = _.map(_.keys(extendedEmojiList()), code => { code, src: emojiUrlFor(code) });
```

jjaffeux: yes indeed function is not needed here god catch 👍  thx

jjaffeux: concerning the exact syntax of dropping `{}` it won’t be valid

PR #4981, opened 2017-07-21
FIX: limited search results

dmacjam:
If there are more full-page search results available, inform users about that.

- Added flag indicating if there are more search results (by querying for limit + 1).
- Added support for backend search pagination (using offset), which will be needed for infinite loading of more search results.
- Added infinite loading of search results to advanced search frontend - in total 10 pages of 50 results are loaded when scrolling which results in 500 search results (not more because of the offset performance). If there are still more results, it tells users to narrow down their search criteria.

Related discussion: 
https://meta.discourse.org/t/results-missing-from-search/64969
https://meta.discourse.org/t/search-results-limited-to-50-each-time/34270


SamSaffron:
The tests need to be a bit smarter here, they should ensure the offset is correct by checking to see that when you ask for page 2 you get proper page 2 results in expected order.

spec/components/search_spec.rb
```diff
@@ -802,4 +802,32 @@ def search
...
+      Search.stubs(:per_filter).returns(number_of_results)
+    end
+
+    it 'returns more results flag' do
+      results = Search.execute('hello', type_filter: 'topic')
+      results2 = Search.execute('hello', type_filter: 'topic', page: 2)
+
+      expect(results.posts.length).to eq(number_of_results)
+      expect(results.more_full_page_results).to eq(true)
+      expect(results2.posts.length).to eq(number_of_results)
```
nlalonde: Add check here that post ids aren't same as the first page. This test will pass if the page param is ignored I think.

dmacjam:
Added infinite loading of search results and rspec tests for pagination are fixed. Thanks for suggestion.

app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -158,6 +162,11 @@ export default Ember.Controller.extend({
     return iconHTML(expanded ? "caret-down" : "caret-right");
   },
 
+  @computed('page')
+  isLastPage(page) {
+    return page == PAGE_LIMIT;
```
nlalonde: Should be `===` instead of `==`. js tests are failing.

coding-horror:
when can we merge this?

PR #4982, opened 2017-07-21
add DiscourseEvent triggers necessary to update a user's permissions before they're notified

LeoMcA:
This adds three new triggers:

`:notify_mailing_list_subscribers` - called before sending a notification about a post to users with mailing list mode turned on
`:before_create_notifications_for_users` - called before sending a notification about a post to a number of users
`:before_create_notification` - called before sending a notification about a post to a single user

These triggers are necessary to query an API and update a user's group memberships (and therefore permissions) before notifying them about a post. This allows us to sync a user's permissions from our API on demand, without worrying about whether a user will be notified about a post they don't have permission to view.

I've included a veritable smorgasbord of tests which should cover all of the unique instances of the event being triggered.


eviltrout:
This looks good, thanks!

PR #4985, opened 2017-07-22
Qunit plugin rake tasks

davidtaylorhq:
This adds a new rake command

```
rake plugin:qunit['discourse-chat-integration']
```
If you don't supply a plugin name, it will run all plugin qunit tests.

**Issues:**
(fixed, see comment below)
<s>
**Caching**
The logic for loading tests happens in a .js.erb file, which means sprockets caches it. I see two options to fix this
- Nuclear option: put `rm -fr tmp/cache` at the beginning of the qunit rake task
- Nicer? option (will only invalidate the files we need): Write a temporary file, say `.qunit_env` every time the task is run, and then put depends_on in the js.erb file. 

@SamSaffron are there any nicer ways?

**Polls Tests**
- Somehow the poll widget tests are always there, whatever I do. Somehow they're loading themselves whenever qunit does something like `visit("/admin/plugins/chat");`. I need to do more investigation here... very weird
</s>

---
This PR also improves the docker rake command to support running plugin qunits. Note that I've switched the env variables around so that they're now more consistent.

By default it will now run all core tests, and all plugin tests

There's then `SKIP_CORE`, `SKIP_PLUGIN`, `JS_ONLY`, `RUBY_ONLY` and `SINGLE_PLUGIN` variables which are hopefully self explanatory. There is never any need to use `LOAD_PLUGINS` (that is included automatically in the two `rake plugin:***` tasks which are called).


davidtaylorhq:
I've applied the 'nuclear option' of purging the whole cache before and after running qunit - seems to work fine. Qunit is never gonna be run in production so I think this should be an ok solution.

The issue with self-loading widget tests was caused by the 'auto-load-modules' initializer - it was auto-loading anything containing `/widget/`. I've amended it now so that it ignores anything containing `-test`.

So... I *think* this might be in a usable state now 😄

SamSaffron:
Where are we at on this? Did you figure out a hack to run plugin tests only?

davidtaylorhq:
Yep, this PR accomplishes exactly that. Note that there are some changes to env variables for the docker rake task, so it'll need reviewing properly before merging to make sure it doesn't break your CI system

SamSaffron:
😍 ❤️  💟  

I would love to merge it then, we can sort out our CI, can you correct the conflict ?

davidtaylorhq:
Conflicts corrected 😄

SamSaffron:
Can you explain a bit about what you removed and why (I can see the comment about MODULE and a few others was nuked)

davidtaylorhq:
So I guess what I've done is split into 3 sections. The commit history for this branch roughly matches up to these steps:

# Added some ENV variable control in qunit tests
1. Renamed `test_helper.js`to `test_helper.js.erb` so that it can contain some "server-side" logic.
2. At the bottom of the file it loops through every test and runs `require(...)` on each one. I've added logic so that these are filtered based on two environment variables:
**QUNIT_SKIP_CORE** and **QUNIT_SINGLE_PLUGIN**
3. In order for this ERB file to be regenerated every time the tests are run, I've had to add a couple of lines to `qunit.rake` which purge `/tmp/cache` before and after running
4. Weirdly, all widget tests were being loaded automatically, regardless of whether `require()` was run in `test_helper.js`. I traced this to [module-auto-loader.js](https://github.com/discourse/discourse/pull/4985/commits/dd2e2c154ce5326a55733419f45c72767621b920), and so have patched it to ignore test files. This is the most risky change in the PR, as it is code used in production.

# Added a `plugin:qunit` rake task
This is modelled off the existing `plugin:spec` command. It sets the necessary environment variables, and then runs the standard `qunit:test` rake task. So `rake plugin:qunit` will run all plugin qunit tests. `rake plugin:qunit['discourse-details']` will run only the qunits for discourse-details.
Should be super-useful for plugin developers, and can also be called in the docker CI script

# Amended docker.rake
By default, the docker task will run
- Core rspec
- Plugin rspec
- Core qunit
- Plugin qunit

The ENV variables available are:

> SKIP_CORE, SKIP_PLUGIN, JS_ONLY, RUBY_ONLY and SINGLE_PLUGIN

which are hopefully self-explanatory.

There's no need to set `LOAD_PLUGINS`, since that is handled by the `rake plugin:spec` and `rake plugin:qunit` tasks. The `MODULE` and `FILTER` variables will still work, but since there's now a better method of loading tests for plugins I removed them from the comments at the top. 

---

Hope that makes sense 😄

SamSaffron:
All makes sense, but can you do something to more surgically expire the cache ONLY for test_helper.js.erb. 

Every time we nuke tmp/cache its 15 second delay rebuilding all assets at least, if there is nothing you can do I guess we can live with this.

davidtaylorhq:
From what I understand, it's being cached by sprockets. 

I can't find anything in the sprockets documentation that allows for triggering a cache expiry manually.

The folder structure inside `tmp/cache` looks like random gibberish to me, but I'm guessing it's somehow based on a hash of the files. Regardless, I have no idea how to find which randomly named file corresponds to `test_helper.js`...

Would be great if we can find a way to do it though, as 15 seconds every time tests are run would be quite annoying in dev...

SamSaffron:
Does this pattern work here? 

https://github.com/discourse/discourse/blob/master/test/javascripts/plugin_tests.js.erb

note all the depend_on stuff. 

What you can do is autogenerate a "version" file to depend on (which is gitignored) and then bump that version file up a number each time specs run, seems like a tidy way of doing it.

davidtaylorhq:
>Does this pattern work here?

I don't think so, because we have no changing files to depend on, it's just an ENV variable change.

Ideally we want to do something like

`depends_on environment_variable`

but that's not something Sprockets supports.

> What you can do is autogenerate a "version" file to depend on (which is gitignored) and then bump that version file up a number each time specs run, seems like a tidy way of doing it.

I'll try and do this, sounds like a good solution 😄

davidtaylorhq:
I found an infinitely superior solution - use URL parameters, in the same way that MODULE and FILTER are passed in already. Makes everything much simpler, so no weird cache issues.

Also has the added bonus that you can now visit

```
http://localhost:3000/qunit?qunit_skip_core=1&qunit_single_plugin=discourse-chat-integration
```
in development and only show the plugin you're working on :tada:

How does that look @SamSaffron ?

SamSaffron:
So the erb re-evaluates every time you change query params? sounds awesome to me, will merge tomorrow.

davidtaylorhq:
I've moved the logic purely into JavaScript, so there's no erb to re-compile. It just checks the URL query parameters at runtime and loads the tests accordingly.

SamSaffron:
cool, lets try it then !!! thanks so much

PR #4988, opened 2017-07-24
Support for sending PMs to email addresses

nbianca:


SamSaffron:
Let us know when this is ready. 

Some stuff to consider:

- Send PM to email should be gated by trust level, out of the box only available to TL4 and above. 
- If there is no way to "reply" via email then the option should be disabled for staff
- Be sure to add some screenshots of how this works

nbianca:
I fixed the merge conflicts. I have also tested and it works as expected.

SamSaffron:
Looks like there are some rubocop failures ... can you install rubocop locally and run it on the files. 

Which editor are you using? 

@tgxworld we need a post on meta that explains the whole rubocop thing ASAP

lib/topic_creator.rb
```diff
@@ -162,11 +162,20 @@ def process_private_message(topic)
...
 
-    unless @opts[:target_usernames].present? || @opts[:target_group_names].present?
+    unless @opts[:target_usernames].present? || @opts[:target_emails].present? || @opts[:target_group_names].present?
       rollback_with!(topic, :no_user_selected)
     end
 
+    if @opts[:target_emails].present?
+      if !SiteSetting.reply_by_email_enabled? then
+        rollback_with!(topic, :reply_by_email_disabled)
+      elsif @user.trust_level < 3 then
```
SamSaffron: This looks hardcoded to me, it really should be a site setting.

SamSaffron:
we also need quite a few tests here  , especially for the server side / api side of things.

config/site_settings.yml
```diff
@@ -830,6 +830,9 @@ trust:
   min_trust_to_send_messages:
     default: 1
     enum: 'TrustLevelSetting'
+  min_trust_to_send_email_messages:
+    default: 3
```
SamSaffron: I would default this to 4, 3 is too low.

SamSaffron: also, lets add one more setting so we can only enable this for staff. default is only staff are allowed this feature.

lib/topic_creator.rb
```diff
@@ -213,4 +239,29 @@ def add_groups(topic, groups)
   def check_can_send_permission!(topic, obj)
     rollback_with!(topic, :cant_send_pm) unless @opts[:skip_validations] || @guardian.can_send_private_message?(obj)
   end
+
+  def find_or_create_user(email, display_name)
+    user = nil
+
+    User.transaction do
```
SamSaffron: Is there a reason for the transaction here? I should be wrapped in a parent transaction anyway

config/site_settings.yml
nbianca: Ok, I will add that.

lib/topic_creator.rb
nbianca: Makes sense. I missed the parent transaction.

config/locales/server.en.yml
```diff
@@ -1010,6 +1011,7 @@ en:
     summary_max_results: "Maximum posts returned by 'Summary This Topic'"
 
     enable_private_messages: "Allow trust level 1 (configurable via min trust level to send messages) users to create messages and reply to messages. Note that staff can always send messages no matter what."
+    enable_private_email_messages: "Allow trust level 4 (configurable via min trust level to send messages) users to send private email messages. Note that staff can always send messages no matter what."
```
SamSaffron: the text here should be improved, this requires that enable_staged_users is true

SamSaffron:

PR #4990, opened 2017-07-25
Add charset preference when bulk importing vBulletin

quangbuule:
This works for my case with`DB_CHARSET=latin1`.
The charset converting would be disabled if not set `DB_CHARSET`.

### DIFF
--- a/script/bulk_import/vbulletin.rb
+++ b/script/bulk_import/vbulletin.rb
@@ -5,6 +5,44 @@ require "htmlentities"
 class BulkImport::VBulletin < BulkImport::Base
 
   SUSPENDED_TILL ||= Date.new(3000, 1, 1)
+  CHARSET_MAP = {
+    "armscii8" => nil,
+    "ascii"    => Encoding::US_ASCII,
+    "big5"     => Encoding::Big5,
+    "binary"   => Encoding::ASCII_8BIT,
+    "cp1250"   => Encoding::Windows_1250,
+    "cp1251"   => Encoding::Windows_1251,
+    "cp1256"   => Encoding::Windows_1256,
+    "cp1257"   => Encoding::Windows_1257,
+    "cp850"    => Encoding::CP850,
+    "cp852"    => Encoding::CP852,
+    "cp866"    => Encoding::IBM866,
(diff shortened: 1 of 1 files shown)


script/bulk_import/vbulletin.rb
```diff
@@ -5,6 +5,44 @@
 class BulkImport::VBulletin < BulkImport::Base
 
   SUSPENDED_TILL ||= Date.new(3000, 1, 1)
+  CHARSET_MAP = {
+    "armscii8" => nil,
```
ZogStriP: Where did you get that list from? Do we need it? I mean, can't we just pass the string to the `encode` method? It's weird this isn't supported by the standard library...

script/bulk_import/vbulletin.rb
```diff
@@ -13,10 +51,19 @@ def initialize
     username = ENV["DB_USERNAME"] || "root"
     password = ENV["DB_PASSWORD"]
     database = ENV["DB_NAME"] || "vbulletin"
+    charset  = ENV["DB_CHARSET"] || "utf8"
```
ZogStriP: Is there a way to automatically detect the charset used?

script/bulk_import/vbulletin.rb
quangbuule: Old version of mysql2:
http://www.rubydoc.info/gems/mysql2/0.3.11/Mysql2/Client

script/bulk_import/vbulletin.rb
quangbuule: Getting automatically is not correct in some cases. I mean, the database, even the table could be `utf8_general_ci`, but the data that the vBulletin saved still in `latin1`.

script/bulk_import/vbulletin.rb
quangbuule: Charset of mysql2 client and the ruby standard lib are not consistent. For example, mysql2 accept "utf8" while Ruby standard using "utf-8" (will throw an error if we put "utf8").
That's why there is CHARSET_MAP there.

ZogStriP:
Thanks a lot!

PR #4993, opened 2017-07-25
FIX: run sidekiq command documentation

dmacjam:


SamSaffron:
Yeah kind of agree with @tgxworld here, maybe just change it to `bundle exec sidekiq` no need for documentation about how to stop it then.

tgxworld:
Thank you @dmacjam

dmacjam:
Yes, that's better. Updated.

PR #4995, opened 2017-07-25
Improve “server warmup” check for qunit rake task

davidtaylorhq:
Add an exception to the `rescue` line


SamSaffron:
This seems totally fine, thanks!

PR #4996, opened 2017-07-25
safely call html_safe on category description

ryantm:
The `categories.description` column is not modified as "not null", so it is possible for the description to be nil. This changes the code not call html_safe on nil.

### DIFF
--- a/app/views/categories/index.html.erb
+++ b/app/views/categories/index.html.erb
@@ -6,7 +6,7 @@
       <h2><a href='<%= c.url %>' itemprop='item'>
         <span itemprop='name'><%= c.name %></span>
       </a></h2>
-      <span itemprop='description'><%= c.description.html_safe %></span>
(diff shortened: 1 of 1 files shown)


ryantm:
Here is the stack trace this fixes:
```
Started HEAD "/" for 72.14.178.104 at 2017-07-25 18:35:15 +0000
Processing by CategoriesController#index as */*
  Rendered categories/index.html.erb within layouts/application (2.3ms)
Completed 500 Internal Server Error in 93ms (ActiveRecord: 11.5ms)
ActionView::Template::Error (undefined method `html_safe' for nil:NilClass
Did you mean?  html_safe?)
/var/www/discourse/app/views/categories/index.html.erb:9:in `block in _app_views_categories_index_html_erb__928370358248483064_69978898940560'
```

SamSaffron:
thanks, great catch!

PR #4997, opened 2017-07-25
FEATURE: Watched Words: when posts contain words, take action

nlalonde:
Four actions that can be taken:

* **Block** the post at the time it is submitted. User sees a message saying they've used a word that's not allowed.
* **Flag** the post as inappropriate, but allow it to be submitted.
* **Censor** the words, which replaces the `censored_words` site setting being removed by this feature.
* **Require Approval** of the post, hiding it until staff reviews it.

There's a max of 1000 words per action. In practice, "bad word lists" I found rarely contain more than a few hundred. `*` can be used as a wildcard.

https://meta.discourse.org/t/actions-on-watched-words/65148

Suggestions for a better place for this in Admin? It probably belongs with the three "Screened" items, but they're in "Logs" which seems wrong...

### DIFF
--- a/app/assets/javascripts/admin/components/admin-watched-word.js.es6
+++ b/app/assets/javascripts/admin/components/admin-watched-word.js.es6
@@ -1,0 +1,19 @@
+import { iconHTML } from 'discourse-common/helpers/fa-icon';
+import { bufferedRender } from 'discourse-common/lib/buffered-render';
+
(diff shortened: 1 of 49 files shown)


nlalonde:
Merge recreated `app/assets/javascripts/pretty-text/engines/markdown-it/censored.js.es6` and probably some other things that shouldn't be there. Will re-do.

PR #5000, opened 2017-07-26
Allow whitespace around % scaler in markdown image syntax

davidtaylorhq:


SamSaffron:
looks good to me, thanks!

PR #5001, opened 2017-07-26
Fix for `plugin:qunit` in RAILS_ENV=test

davidtaylorhq:
@SamSaffron small fix to the earlier PR 

The `LOAD_PLUGINS` environment variable is read by `application.rb` before the rake task is run, so we need to spawn a new process after changing its value.

This is now the same approach that the existing `plugin:spec` command uses.

### DIFF
--- a/lib/tasks/plugin.rake
+++ b/lib/tasks/plugin.rake
@@ -101,14 +101,20 @@ desc 'run plugin qunit tests'
 task 'plugin:qunit', [:plugin, :timeout] do |t, args|
   args.with_defaults(plugin: "*")
 
-  ENV['LOAD_PLUGINS'] = '1'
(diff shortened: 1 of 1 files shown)


SamSaffron:
this does not feel right, should it not be `bundle exec rake` ?

davidtaylorhq:
I've just copied the way that `plugin:spec` does it:

https://github.com/davidtaylorhq/discourse/blob/5c5f77bda98da5c8da0cb413345f9bb26386ff3e/lib/tasks/plugin.rake#L94

Shall I change both of them to `bundle exec`?

SamSaffron:
OK, lets try it out, maybe bundler already fussed with path at that point.

PR #5003, opened 2017-07-26
[FIX] invalid byte sequence in UTF-8

mtawil:
The name should be scrubbed before to avoid encoding issues when using ```name.blank?```.

### DIFF
--- a/script/bulk_import/base.rb
+++ b/script/bulk_import/base.rb
@@ -518,8 +518,8 @@ class BulkImport::Base
   end
 
   def fix_name(name)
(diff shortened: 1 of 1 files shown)


mtawil:
But you already scrubbed it if it's not a blank!

ZogStriP:
Well, that `scrub!` should not be there then :)

mtawil:
Ok now it's scrubbed if the name has not a valid encoding 👍

SamSaffron:
sure this seems ok to me, thanks!

PR #5006, opened 2017-07-27
Replace rmmseg gem for cppjieba_rb since better dictionary

fantasticfears:


fantasticfears:
New binding gem is here. @SamSaffron https://github.com/fantasticfears/cppjieba_rb

It's extracted from #4974.

app/services/search_indexer.rb
```diff
@@ -31,7 +31,7 @@ def self.update_index(table, id, raw_data)
     end
 
     # for user login and name use "simple" lowercase stemmer
-    stemmer = table == "user" ? "simple" : Search.long_locale
+    stemmer = table == "user" ? "simple" : Search.ts_config
```
SamSaffron: hmm why was this change to use ts_config? I guess long_local may not even be available in pg?

fantasticfears: PG took a text search configuration as the parameters for `TO_TSVECTOR` & `TO_TSQUERY`. It isn't related to locale so it's better for me to name with that intent.

From PG's doc:
```
CREATE TEXT SEARCH DICTIONARY english_ispell (
    TEMPLATE = ispell,
    DictFile = english,
    AffFile = english,
    StopWords = english
);
```

SamSaffron:
OK, let's give this a shot! thanks!

PR #5008, opened 2017-07-28
FIX: Pull last emoji-db revision

jjaffeux:
This commit will add new images in some sets (most notably emoji_one) and fix a bug where
:man_facepalming: was using :person_facepalming: image which is in fact
represented as a woman in most sets.


jjaffeux:
screen shot 2017-07-28 at 17 30 41

PR #5009, opened 2017-07-30
Allow for no positioning on emoji picker

gdpelican:
Right now the emoji picker relies on there being a '#reply-control' element present (amongst several other elements specific to the discourse composer). This allows passing a 'noPositioning' argument to the component which gives the default positioning all the time.

### DIFF
--- a/app/assets/javascripts/discourse/components/emoji-picker.js.es6
+++ b/app/assets/javascripts/discourse/components/emoji-picker.js.es6
@@ -448,7 +448,7 @@ export default Ember.Component.extend({
       $picker.css(_.merge(attributes, options));
     };
 
-    if(Ember.testing) {
+    if(Ember.testing || this.get('noPositioning')) {
       desktopPositioning();
       return;
     }
--- a/app/assets/javascripts/discourse/components/emoji-picker.js.es6
+++ b/app/assets/javascripts/discourse/components/emoji-picker.js.es6
@@ -21,6 +21,8 @@ export function resetCache() {
 let $picker, $filter, $results, $list, scrollPosition, $visibleSections, _checkTimeout;
 
 export default Ember.Component.extend({
+  automaticPositioning: true,
+
   willDestroyElement() {
     this._super();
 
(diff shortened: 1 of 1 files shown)


tgxworld:
@jjaffeux  Does this look OK to you?

jjaffeux:
Looks ok, but not a huge fan of the name.

What do you think of {{emoji-picker desktop=true}}

```if(this.get("desktop")) {}```

or even `desktopPositioning`. In short, I would prefer a name saying what it does, that a name saying what it doesn’t do.

**Comment by gdpelican** [comment]
Hmmm, but 'desktop', isn't actually what it does (since this code is run on mobile as well)?

I'm basically looking for a way to avoid all of the composer-specific selectors ('#reply-control', '.d-editor-container', '.d-header', '.d-editor-preview-wrapper') which are checked later in the method and which cause the picker to error if not opened from the composer; so it's really something like `defaultPosition` or `skipRelativePositioning` or something like this.

**Comment by SamSaffron** [comment]
I would go with `automaticPositioning` and have that default to true. 

So you would use this to override (has to be explicitly set to false) 

```
{{emoji-picker automaticPositioning=false}}
```

To get the desired behavior.

jjaffeux:
I like this one, I will accept the PR with this and if you can also remove ´Ember.testing´ and use this in the test, it would be perfect 👌

jjaffeux:
You now have to use it in emoji-picker acceptance tests otherwise they will fail ;)

SamSaffron:
It has to explicitly check for false with ===

SamSaffron:
@jjaffeux  this is such a tiny PR ... can you make this go away :) if you need to just commit the change yourself go for it

jjaffeux:
Fixed by https://github.com/discourse/discourse/pull/5074

tgxworld:
@jjaffeux I would usualy pull down the branch and do a `git commit --amend` so that the original author gets the attribution. I've seen instances in other open source projects I worked on where people do care about the attribution. Just some thoughts I have :thought_balloon:

jjaffeux:
You are probably right, but it's such a tiny change... I figured it would be ok.

tgxworld:
@jjaffeux No worries about it. I've just seen instances where some people get really upset even over doc changes xD

PR #5012, opened 2017-07-31
"Watched Words" tab should only be visible to admins

OsamaSayegh:
Meta topic: https://meta.discourse.org/t/automatic-actions-based-on-words-found-in-posts/66878/5


nlalonde:
Thank you! I'm currently moving Watched Words so it's a sub-section of Logs, so I'll be sure to include your fix there.

PR #5013, opened 2017-07-31
FIX: add additional email to tests and clean up resulting mess

LeoMcA:
As a first step to [making use of additional emails](https://meta.discourse.org/t/additional-email-support/59847/9?u=leomca) I've added a secondary email to `Fabricate(:user)` to ensure that a user actually having multiple emails doesn't break everything.

Thankfully, it only broke a few things, which I fixed. However, I want to highlight what I've done to `AdminUserIndexQuery` because I'm concerned it's _deeply_ immoral.

With the `.joins` where it was we had the problem of duplicate users in the result. Adding a `.distinct` after solved that problem, but created a new one when the query contained `ORDER BY COALESCE(...)` because `for SELECT DISTINCT, ORDER BY expressions must appear in select list`.

So, my ultimate solution is to do all the filtering and making distinct in a subquery, which is then ordered in the outer query. However I know for sure this breaks `.pluck`, and who knows what else. Is there a better way?

### DIFF
--- a/app/jobs/scheduled/pending_users_reminder.rb
+++ b/app/jobs/scheduled/pending_users_reminder.rb
@@ -12,7 +12,7 @@ module Jobs
           query = query.where('users.created_at < ?', SiteSetting.pending_users_reminder_delay.hours.ago)
         end
 
-        newest_username = query.limit(1).pluck(:username).first
+        newest_username = query.limit(1).select(:username).first&.username
 
         return true if newest_username == previous_newest_username # already notified
 
--- a/app/controllers/admin/groups_controller.rb
+++ b/app/controllers/admin/groups_controller.rb
@@ -23,6 +23,7 @@ class Admin::GroupsController < Admin::AdminController
         valid_emails[email] = valid_usernames[username_lower] = id
         id
       end
+      valid_users.uniq!
       invalid_users = users.reject! { |u| valid_emails[u] || valid_usernames[u] }
       group.bulk_add(valid_users) if valid_users.present?
       users_added = valid_users.count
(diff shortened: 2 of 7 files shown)


coding-horror:
hmm we had a fair number of unanticipated regressions based on the new 1-to-many relationship between users and email, so caution is warranted here

LeoMcA:
> So, my ultimate solution is to do all the filtering and ordering in a subquery, which is then made distinct in the outer query.

Hang on, I've done the opposite: I've done all the filtering and making distinct in the subquery, and then ordered it in the outer query.

I'll edit my OP and then I think I need some :zzz:

SamSaffron:
@tgxworld this is yours for review

app/jobs/scheduled/pending_users_reminder.rb
```diff
@@ -12,7 +12,7 @@ def execute(args)
           query = query.where('users.created_at < ?', SiteSetting.pending_users_reminder_delay.hours.ago)
         end
 
-        newest_username = query.limit(1).pluck(:username).first
+        newest_username = query.limit(1).select(:username).first&.username
```
tgxworld: Why is this change needed?

LeoMcA: Fast pluck doesn't seem to work with subqueries:

     Failure/Error: construct_relation_for_association_calculations.pluck(*cols)
     
     NoMethodError:
       undefined method `empty?' for #<Arel::Nodes::TableAlias:0x007ffaa4b27338>
     # ./lib/freedom_patches/fast_pluck.rb:59:in `pluck'
     # ./app/jobs/scheduled/pending_users_reminder.rb:15:in `execute'
     # ./spec/jobs/pending_users_reminder_spec.rb:13:in `block (3 levels) in <top (required)>'

tgxworld:
Thanks @LeoMcA

PR #5015, opened 2017-08-01
Enhance BulkImport pre_cook

quangbuule:
Input:
```
le_quang_buu
foo
bar
```
Output of previous behavior:

> le<em>quang</em>buu foo bar

Output of new behavior (preventing emphasizing inside words and converting `\n` into `<br>\n`):

> le_quang_buu
> foo
> bar

### DIFF
--- a/script/bulk_import/base.rb
+++ b/script/bulk_import/base.rb
@@ -18,7 +18,8 @@ class BulkImport::Base
     @raw_connection = PG.connect(dbname: db[:database], host: db[:host_names]&.first, port: db[:port])
 
     @markdown = Redcarpet::Markdown.new(
-      Redcarpet::Render::HTML,
+      Redcarpet::Render::HTML.new(hard_wrap: true),
+      no_intra_emphasis: true,
       fenced_code_blocks: true,
       autolink: true
     )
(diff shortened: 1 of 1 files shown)


quangbuule:
https://github.com/quangbuule/discourse/blob/bulk-import-patch-2/script/bulk_import/base.rb#L580

That line should be:
```ruby
```

script/bulk_import/base.rb
```diff
@@ -579,8 +580,11 @@ def pre_cook(raw)
     cooked = @markdown.render(cooked).scrub.strip
 
     cooked.gsub!(/\[QUOTE="?([^,"]+)(?:, post:(\d+), topic:(\d+))?"?\](.+?)\[\/QUOTE\]/im) do
-      username, post_id, topic_id = $1, $2, $3
-      quote = @markdown.render($4.presence || "").scrub.strip
+      username, post_id, topic_id, quote = $1, $2, $3, $4
+
+      quote = quote.scrub.strip
```
ZogStriP: Why did you remove the markdown rendering? Sometimes, a quote has markdown/html and we need to render it.

quangbuule: In my opinion:
When it's html, we don't need to do markdown.
When it's markdown, we already rendered just few lines upper (in L580).

Could you send me some cases we need to render again?

SamSaffron:
@ZogStriP is on holidays at the moment, as soon as he is back in a week or so he will decide what to do here. Sorry about the delay.

ZogStriP:
Ok, let's try it.

PR #5016, opened 2017-08-01
Exclude bundle directory when running rubocop

davidtaylorhq:
If I run `bundle exec rubocop` in the root of discourse, it tries to lint many thousands of files from dependencies inside the `bundle` directory. This PR adds an exclude rule for anything inside the `bundle` directory, so we're just linting Discourse files.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,9 @@ AllCops:
   TargetRubyVersion: 2.4
   DisabledByDefault: true
   Exclude:
-    - bundle/**/*
+    - 'bundle/**/*'
+    - 'vendor/**/*'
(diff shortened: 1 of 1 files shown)


SamSaffron:
oh oh ... @tgxworld looks like rubocop changed and now the rule that was is no longer... this is not boding well

davidtaylorhq:
This broke travis:

screen shot 2017-08-01 at 16 44 44

Looks like setting the `Exclude` setting overrode the default, rather than appending to it. I've now committed the defaults as well, so hopefully it'll work on Travis again

SamSaffron:
thanks!

PR #5018, opened 2017-08-02
FIX: First load of embed comments would always fail on WebKit

xfalcox:
https://meta.discourse.org/t/first-load-on-embedded-comments-fail-on-webkit/67206?u=falco


tgxworld:
Looks OK to me. Let's try this out.

PR #5019, opened 2017-08-02
Fix Redis command errors when trying to start app with a readonly Redis.

tgxworld:


SamSaffron:
it looks ok to me

tgxworld:
Thanks for reviewing @SamSaffron

PR #5021, opened 2017-08-02
reindex broken search data

fantasticfears:
Since `cppjieba` is in place, it changes the search data vector. Basically magically rebuilt the search index.

@SamSaffron concerns for performance hit?

### DIFF
--- a/app/jobs/scheduled/reindex_search.rb
+++ b/app/jobs/scheduled/reindex_search.rb
@@ -4,7 +4,75 @@ module Jobs
     every 1.day
 
     def execute(args)
-      Search.rebuild_problem_posts
+      rebuild_problem_topics
+      rebuild_problem_posts
+      rebuild_problem_categories
+      rebuild_problem_users
+    end
+
+    def rebuild_problem_categories(limit = 10000)
+      categories = load_problem_categories(limit)
+
+      categories.each do |category|
+        SearchIndexer.index(category, force: true)
+      end
+    end
+
+    def rebuild_problem_users(limit = 10000)
+      users = load_problem_users(limit)
+
+      users.each do |user|
+        SearchIndexer.index(user, force: true)
+      end
+    end
+
+    def rebuild_problem_topics(limit = 10000)
+      topics = load_problem_topics(limit)
+
+      topics.each do |topic|
+        SearchIndexer.index(topic, force: true)
+      end
+    end
+
+    def rebuild_problem_posts(limit = 10000)
+      posts = load_problem_posts(limit)
+
+      posts.each do |post|
+        SearchIndexer.index(post, force: true)
+      end
+    end
+
+    private
+
+    def load_problem_posts(limit)
+      Post.joins(:topic)
+        .where('posts.id IN (
+                SELECT p2.id FROM posts p2
+                LEFT JOIN post_search_data pd ON pd.locale = ? AND pd.version = ? AND p2.id = pd.post_id
+                WHERE pd.post_id IS NULL
+                )', SiteSetting.default_locale, Search::INDEX_VERSION)
+        .limit(limit)
+    end
+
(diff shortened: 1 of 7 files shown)


app/services/search_indexer.rb
```diff
@@ -149,4 +225,18 @@ def characters(string)
...
+    if SiteSetting.search_index_core_version < 1
+      if ['zh_TW', 'zh_CN', 'ja', 'ko'].include?(SiteSetting.default_locale) || SiteSetting.search_tokenize_chinese_japanese_korean
+        self.force_reindex_posts_topics
+        self.force_reindex_categories
+        self.force_reindex_users
+      end
+
+      SiteSetting.search_index_core_version = 1
+    end
+  end
```
SamSaffron: this should be in a once off job, I don't particularly like having this in the core class.

version should be stored in the search table and then we can have a regular weekly job that forces are reindex on posts that have an earlier version and cleanly recover from partial re-index

note I am going to be out for a week, so it is going to take a while to merge this in.

fantasticfears: Sure, got it

db/migrate/20170803123704_add_version_to_search_data.rb
```diff
@@ -0,0 +1,8 @@
+class AddVersionToSearchData < ActiveRecord::Migration
+  def change
+    add_column :post_search_data, :core_version, :integer, default: 0
```
SamSaffron: hmmm ... lets use the name `version` here, core version is a bit confusing

SamSaffron:
lets remove the once off and instead default to version 0 and have version 1 be in code. 

I think we are well overdue for a full search re-index due to other bugs and this helps remove a whole bunch of code.

fantasticfears:
Done

app/jobs/scheduled/reindex_search.rb
```diff
@@ -1,10 +1,69 @@
...
+
 module Jobs
   # if locale changes or search algorithm changes we may want to reindex stuff
   class ReindexSearch < Jobs::Scheduled
     every 1.day
 
     def execute(args)
-      Search.rebuild_problem_posts
+      limiter = Search::QueryLimiter.new(args[:limit] || 10_000)
+      limiter.step { |limit| rebuild_problem_categories(limit) }
```
SamSaffron: I don't really think we need all this fancy...  just limit to 10000 topics / 10000 posts / 10000 users. no need to do a limit on 10000 things.  code will be much easier to understand, one less class.

spec/services/search_indexer_spec.rb
```diff
@@ -13,14 +13,16 @@
     expect(raw_data.split(' ').length).to eq(2)
   end
 
+  let(:version_num) { 12345 }
   it 'correctly indexes a post' do
     data = "<a>This</a> is a test"
-
+    stub_constant(Search, 'INDEX_VERSION', version_num)
```
SamSaffron: mocking and stubbing makes me super uneasy. in this context. 

I would ... index something, then update the row to -1 and make sure it gets re-indexed, something like that ... no need for fancy mocking.

fantasticfears:
Fixed

SamSaffron:
now we are talking! looking great. 

I am considering adding an index on version,locale,(post_id|topic_id) so the left join is fast in the job. but since the query only runs daily this is not critical. worth measuring the impact though.

SamSaffron:
Oh ... one thing I noticed... we are loading 10k posts / topics etc into memory... that is not ideal. 

What we should be doing is selecting ids and then using find_each or something along those lines so the in-memory batches are not huge and we dont bloat sidekiq

SamSaffron:
ok committed a fix for memory bloating.

fantasticfears:
Thanks!

PR #5022, opened 2017-08-02
FIX: generate valid markdown from <br></b> in an email

LeoMcA:


coding-horror:
these look fine as long as @ZogStriP signs off on it!

PR #5023, opened 2017-08-03
solution for encoding problem

Alavi1412:
This will solve the problem for not showing topics with non-English slugs


SamSaffron:
sorry I feel like we need to sort this out cleanly in rails, not hack every instance internally. closing ... waiting on rails  5 for a comprehensive fix.

PR #5025, opened 2017-08-04
Update translations.

zarie:


techAPJ:
Thanks for the PR but translations needs to be updated via [Transifex](https://www.transifex.com/projects/p/discourse-org/).

zarie:
Ok, no problem.

PR #5027, opened 2017-08-04
FIX: allow an admin to click on blank errors

LeoMcA:
Oops, in https://github.com/discourse/discourse/pull/5026 I forgot to show a string in the admin interface for blank errors so they can actually be clicked on and investigated:

screenshot from 2017-08-04 17-35-12

This fixes that!

### DIFF
--- a/app/serializers/incoming_email_details_serializer.rb
+++ b/app/serializers/incoming_email_details_serializer.rb
@@ -17,7 +17,7 @@ class IncomingEmailDetailsSerializer < ApplicationSerializer
 
   def error
     if @error_string.blank?
-      "Unrecognized Error"
+      I18n.t("emails.incoming.unrecognized_error")
     else
       @error_string
     end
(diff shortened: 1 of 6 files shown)


app/serializers/incoming_email_details_serializer.rb
```diff
@@ -16,7 +16,11 @@ def initialize(incoming_email, opts)
   EMAIL_RECEIVER_ERROR_PREFIX = "Email::Receiver::".freeze
 
   def error
-    @error_string
+    if @error_string.blank?
+      "Unrecognized Error"
```
ZogStriP: Can this be I18nlized?

LeoMcA: @ZogStriP done!

app/serializers/incoming_email_details_serializer.rb
```diff
@@ -17,7 +17,7 @@ def initialize(incoming_email, opts)
 
   def error
     if @error_string.blank?
-      "Unrecognized Error"
+      I18n.t("emails.incoming.unrecognized_error")
```
ZogStriP: You can shorten this to

```ruby
def error
  @error_string.presence || I18n.t("emails.incoming.unrecognized_error")
end
```

;)

LeoMcA: Neat, I was wanting to do something in that form but couldn't work it out... now I know!

PR #5028, opened 2017-08-05
FIX: allows onebox to force GET hosts returning wrong headers on HEAD

jjaffeux:
See https://meta.discourse.org/t/cant-onebox-overwatch-forums-topic/67066/5 for more context

https://us.battle.net/forums/en/overwatch/topic/20758326009?page=4#post-70 should now be correctly oneboxed


jjaffeux:
@techAPJ thx for checking this

PR #5029, opened 2017-08-06
Edit maxlength categories longer

hackvothuong:
I change maxlength categories from 50 to 60. Maxlength="50" so short for some website. Please accept my pull. Thank you!


ZogStriP:
Thanks but this won't work as the length is enforced on the server-side as well.

PR #5035, opened 2017-08-09
Remove Spork dependency.

nbianca:


spec/rails_helper.rb
```diff
@@ -4,217 +4,4 @@
...
-require 'spork'
 require 'rbtrace'
-#uncomment the following line to use spork with the debugger
-#require 'spork/ext/ruby-debug'
-
-Spork.prefork do
-  # Loading more in this block will cause your tests to run faster. However,
-  # if you change any configuration or code from libraries loaded here, you'll
-  # need to restart spork for it take effect.
-  require 'fabrication'
```
tgxworld: I think you should only remove `Spork.prefork` but whatever that is within the block is still required.

PR #5036, opened 2017-08-10
FIX: Staged users were incorrectly created multiple times with same e…

tgxworld:
…mail.

@ZogStriP Please review again hehe The job is failing in production for some sites with the following error

```
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_user_emails_on_email"
DETAIL:  Key (lower(email::text))=(no-reply@local) already exists.
```

This is due to the fact that for the same email, multiple staged users have been created.


ZogStriP:
Ooopss. Did not catch that. LGTM 👍

tgxworld:
@ZogStriP not possible to catch because we got into a really weird state in production.

PR #5037, opened 2017-08-10
FEATURE: ember native select

jjaffeux:
#### Usage

```
{{d-select-box data=data value=selectedDataId}}
```

#### Design

To make this component highly customisable, multiple choices have been made.

##### Components architecture

The select-box is a combination of multiple components:

* select-box
  * select-box-header
  * select-box-filter
  * select-box-collection
    * select-box-row

Each sub component has a simple API to work with others and can be fully replaced.

##### select-box-wrapper

To allow for all designs the full component is surrounded by a `div.select-box-wrapper`, which can be used to add a drop shadow for example.

### DIFF
--- a/app/assets/javascripts/discourse/components/select-box.js.es6
+++ b/app/assets/javascripts/discourse/components/select-box.js.es6
@@ -1,0 +1,211 @@
+import { observes } from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  classNames: "select-box",
+
+  classNameBindings: ["expanded:is-expanded"],
+
+  attributeBindings: ['componentStyle:style'],
+  componentStyle: function() {
+    return Ember.String.htmlSafe(`width: ${this.get("maxWidth")}px`);
+  }.property("maxWidth"),
+
+  expanded: false,
+  focused: false,
+
+  caretUpIcon: "caret-up",
+  caretDownIcon: "caret-down",
+  icon: null,
+
+  value: null,
+  noDataText: I18n.t("select_box.no_data"),
+  lastHoveredId: null,
+
+  idKey: "id",
+  textKey: "text",
+  iconKey: "icon",
+
+  filterable: false,
+  filter: "",
+  filterPlaceholder: I18n.t("select_box.filter_placeholder"),
+  filterIcon: "search",
+
+  selectBoxRowComponent: "select-box/select-box-row",
+  selectBoxFilterComponent: "select-box/select-box-filter",
+  selectBoxHeaderComponent: "select-box/select-box-header",
+  selectBoxCollectionComponent: "select-box/select-box-collection",
+
+  maxCollectionHeight: 200,
+  maxWidth: 200,
+  verticalOffset: 0,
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/discourse/components/select-box.js.es6
```diff
@@ -0,0 +1,179 @@
...
+    return Ember.String.htmlSafe(`width: ${this.get("maxWidth")}px`);
+  }.property("maxWidth"),
+
+  expanded: false,
+
+  caretUpIcon: "caret-up",
+  caretDownIcon: "caret-down",
+  icon: null,
+
+  data: [],
```
eviltrout: If you assign a default this way, every instance of `select-box` will access the same array. You need to set the default in the `init()` method when it's an object. Same for `selectedData`, `filteredData` and any other array or object as a default.

app/assets/javascripts/discourse/components/select-box.js.es6
```diff
@@ -0,0 +1,179 @@
...
+      return d[this.get("idKey")] === this.get("value");
+    });
+
+    if(!_.isUndefined(selectedData)) {
+      this.set("selectedData", this._normalizeData(selectedData));
+    }
+  },
+
+  _remapData(data) {
+    return _.map(data, (d) => {
```
eviltrout: You could do this in one  line since an arrow method returns by default. `return _.map(data, d => this._normalizeData(d))`

You could also just use regular javascript instead of underscore for the map, although I am less picky about that:

`data.map(d => this._normalizeData(d))`

jjaffeux: done

app/assets/javascripts/discourse/components/select-box.js.es6
jjaffeux: done

app/assets/stylesheets/common/components/select-box.scss
```diff
@@ -0,0 +1,238 @@
...
+    padding: 5px 0;
+  }
+
+  .icon {
+    margin-right: 5px;
+  }
+}
+
+.select-box .select-box-offscreen, .select-box .select-box-offscreen:focus {
+  clip: rect(0 0 0 0) !important;
```
SamSaffron: I really don't like this, any way to get rid of all the !important?

jjaffeux: done

coding-horror:
Nice work! This old select2 has been a thorn in our side for years now.

jjaffeux:
It's still here for sometimes, a huge part of the work is now to replace them everywhere, I will start doing it.

The category dropdown in the composer would be a good candidate I think if @SamSaffron and @eviltrout, it's used a lot more, so we will know faster if there are big issues and it has a more complex row content.

eviltrout:
I say just start replacing them. Perhaps one commit per select at first so we can roll back easily if there is a problem.

coding-horror:
Yes let's start replacing asap

PR #5039, opened 2017-08-11
Add image to whitelisted SVG elements

imagejan:
This allows SVG images containing embedded images to be rendered correctly.

### DIFF
--- a/lib/upload_creator.rb
+++ b/lib/upload_creator.rb
@@ -8,8 +8,8 @@ class UploadCreator
   TYPES_TO_CROP ||= %w{avatar card_background custom_emoji profile_background}.each(&:freeze)
 
   WHITELISTED_SVG_ELEMENTS ||= %w{
-    circle clippath defs ellipse g line linearGradient path polygon polyline
-    radialGradient rect stop svg text textpath tref tspan use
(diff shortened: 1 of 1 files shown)


SamSaffron:
is image safe? I am concerned you could use that as a vector for attack by including a ton of images on a 3rd party server causing a denial of service.

imagejan:
I'm not sure about possible security issues. If there are concerns with generally whitelisting `image`, it would be good if the whitelist of allowed elements could be configurable via admin interface, so that we can easily add it in our own discourse instance.
My intended use case are svg images that contain embedded base64-encoded images via `<image xlink:href="data:image/png;base64, ..."/>`. Maybe this can be special-cased somehow to avoid "external" images but allow embedded ones?!

ZogStriP:
I support adding a site setting for `whitelisted SVG elements` 👍

SamSaffron:
I am with @ZogStriP  we should ship with super strict defaults, cause a single SVG with 10000 images would be a disaster. BUT if you want to opt in for that risk, go for it.

SamSaffron:
closing in favor of the proposed PR that would make this configurable.

PR #5040, opened 2017-08-12
Scroll is tracked on search page.

nbianca:
Hello,

I implemented a new component "scroll-tracker" that is meant to be used to track position on a page. This component has two attributes: the name - which is mandatory and is a unique key; a tag - which is a value that is used to validate previous scrolling value.

In my case, the search term is going to be tag. Let's say I search for 'A' and I navigate until the middle of the page. After that, I search for 'B' and it will set the scroll position to the previous one, something that should not happen.

Many thanks to @jjaffeux for his help.

### DIFF
--- a/app/assets/javascripts/discourse/components/scroll-tracker.js.es6
+++ b/app/assets/javascripts/discourse/components/scroll-tracker.js.es6
@@ -1,0 +1,41 @@
+import Scrolling from 'discourse/mixins/scrolling';
+
+export default Ember.Component.extend(Scrolling, {
+
+  didReceiveAttrs() {
+    this._super();
+
+    this.set('trackerName', `scroll-tracker-${this.get('name')}`);
+  },
+
+  didInsertElement() {
+    this._super();
+
+    this.bindScrolling({ name: this.get('name') });
+  },
+
+  didRender() {
+    this._super();
+
+    const data = this.session.get(this.get('trackerName'));
+    if (data && data.position >= 0 && data.tag === this.get('tag')) {
+      Ember.run.next(() => $(window).scrollTop(data.position + 1));
+    }
+  },
+
+  willDestroyElement() {
+    this._super();
+
+    this.unbindScrolling(this.get('name'));
+  },
+
+  scrolled() {
+    this._super();
+
+    this.session.set(this.get('trackerName'), {
+      position: $(window).scrollTop(),
+      tag: this.get('tag'),
+    });
+  },
+
+});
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/scroll-tracker.js.es6
```diff
@@ -0,0 +1,42 @@
...
+
+    this.bindScrolling({ name: this.get('name') });
+  },
+
+  didRender() {
+    this._super();
+
+    const data = this.session.get(this.get('trackerName'));
+    if (data && data.position && data.position >= 0 && data.tag == this.get('tag')) {
+      $(window).scrollTop(data.position + 1);
```
jjaffeux: Why do we scrollTop two times here please?

app/assets/javascripts/discourse/components/scroll-tracker.js.es6
```diff
@@ -0,0 +1,42 @@
+import Scrolling from 'discourse/mixins/scrolling';
+
+export default Ember.Component.extend(Scrolling, {
+
+  didReceiveAttrs() {
+    this._super();
+
+    this.set('trackerName', 'scroll-tracker-' + this.get('name'))
```
jjaffeux: minor pick, but why not using es6 interpolation? 

```es6
`scroll-tracker-${this.get('name')}`
```

app/assets/javascripts/discourse/components/scroll-tracker.js.es6
```diff
@@ -0,0 +1,42 @@
...
+    this._super();
+
+    this.bindScrolling({ name: this.get('name') });
+  },
+
+  didRender() {
+    this._super();
+
+    const data = this.session.get(this.get('trackerName'));
+    if (data && data.position && data.position >= 0 && data.tag == this.get('tag')) {
```
jjaffeux: I think we should use `===` here or eslint will complain

jjaffeux: is there a possibility we get `data`, but we don’t get a `data.position` ?

app/assets/javascripts/discourse/components/scroll-tracker.js.es6
jjaffeux: Also  we need a ; as eslint is reporting

app/assets/javascripts/discourse/components/scroll-tracker.js.es6
jjaffeux: also can you explain the `+ 1` please ? I just used it when I made it work, but not sure what's the reason behind this?

jjaffeux:
Hi @nbianca, thx for the contribution and glad I can help.

It looks good to me now, I would love to have a test for this but it might be complicated (not sure how to handle back button properly for example?). Have you any idea on how to accomplish this ? If we can get back button in tests, it should be rather easy to test, visit("/search"), fillIn query, scroll, store position, click result, back, assert position is the same.

If you have time to answer the question about the "+ 1" would be cool, otherwise I will just dig later ;)

jjaffeux:
Merged, thx @nbianca

PR #5042, opened 2017-08-13
SECURITY: Do not show latest/top topics on 404 for login_required sites

davidtaylorhq:


coding-horror:
Oh wow this was indeed a major oversight thanks for this can you review
@techapj?

techAPJ:
Thanks @davidtaylorhq 👍

tgxworld:
Is it possible to add a test here so that we can prevent this from regressing? :)

coding-horror:
The exploit is very tiny, you have to leak a download link first, I did not
understand how narrow this is, so probably do not need to bother

tgxworld:
@coding-horror Adding a test isn't dependent on how exploitable a bug is but rather a good practice so we can prevent this from regressing in the future when someone decides to come along and refactor this piece of code.

PR #5044, opened 2017-08-14
Fix: unlinked topic search model

fantasticfears:


SamSaffron:
wait, we should probably also add a dependent destroy here, to make sure it is cleaned up.

fantasticfears:
I've extracted them into a concern so that no one will forget to add the dependent.

SamSaffron:
much better, I like this change

PR #5045, opened 2017-08-14
Small refactor for user search

fantasticfears:
In Rails spec, only `search_for_user` is extracted to helper. Changes are basically a place change.

### DIFF
--- a/spec/integration/user_search_spec.rb
+++ b/spec/integration/user_search_spec.rb
@@ -1,0 +1,116 @@
+require 'rails_helper'
+
+describe "UserSearch orders according to various order" do
+
+  let(:topic)     { Fabricate :topic }
+  let(:topic2)    { Fabricate :topic }
+  let(:topic3)    { Fabricate :topic }
+  let(:topic4)    { Fabricate :topic }
+  let(:user1)     { Fabricate :user, username: "mrb", name: "Michael Madsen", last_seen_at: 10.days.ago }
+  let(:user2)     { Fabricate :user, username: "mrblue",   name: "Eddie Code", last_seen_at: 9.days.ago }
+  let(:user3)     { Fabricate :user, username: "mrorange", name: "Tim Roth", last_seen_at: 8.days.ago }
+  let(:user4)     { Fabricate :user, username: "mrpink",   name: "Steve Buscemi",  last_seen_at: 7.days.ago }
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/lib/user-search.js.es6
```diff
@@ -44,7 +44,7 @@ function organizeResults(r, options) {
   if (r === CANCELLED_STATUS) { return r; }
 
   var exclude = options.exclude || [],
-      limit = options.limit || 5,
+      limit = options.limit || 6,
```
SamSaffron: this is just a bit off, the number of results are limited to 5... not 6. I think the change just creates confusion.

fantasticfears: Well, according to the previous logic, the intention is 6. I'll change that to 5.

fantasticfears:
Spec checked and the test for limit is added.

SamSaffron:
I am still somewhat confused... 

I thought what we do is search for 6, and render 5... The reason we do that is so we can conditionally show the "more results" thing. Not following what bug this behavior change is fixing.

fantasticfears:
Ouch, I misunderstand your previous intention. I didn't alter the behaviour previously...

So my point is that as it iterates over the `result`, the default limit is closed range. However the array index starts at 0, this is so confusing when skim through the code as the limit reads 5 but returns 6 items. That costs me a hour to poke around in the test suite :( This is such a old practice problem since C.

Coming back to this part of code:
> I thought what we do is search for 6, and render 5... The reason we do that is so we can conditionally show the "more results" thing.

I am not sure where the code is explicit about searching for 6, Rails is default to 20. The limit value is used to trim the addition results.
image
I did a double check on Meta, it renders 6 users...

As a conclusion, `userSearch` tries to search for 20 users and shows 6 users by default...

SamSaffron:
I have been struggling with this all day... but I am going to refer to: 

In particular to the 

> Refactoring only NOT welcome

This changes the git history for the test file and does not really fix anything. I guess the test is in the wrong spot, but moving it should be up to a core developer. 

We have been bitten big times by refactoring PRs in the past, they are enormously difficult to deal with and the policy is there for a reason. 

Sorry.

fantasticfears:
I see. I will work on the big fix pr instead.

SamSaffron:
That is fine, just careful not to shuffle code around.

fantasticfears:
Will do

PR #5046, opened 2017-08-14
Some more HTML to Markdown fixes

LeoMcA:
- DRY up the code for parsing emphasis spans
- Support all sorts of strange things email clients might do with spaces
- Support newlines leading emphasis spans


coding-horror:
If this looks good @zogstrip let's merge it.

PR #5047, opened 2017-08-14
FEATURE: add full editing access to queued posts

xrav3nz:
> **Discussion & Screenshots:** https://meta.discourse.org/t/ability-to-change-the-category-of-a-post-before-approving/66754/7?u=xrav3nz

For pending new topics: the body of the post, title, categories and the tags are all editable.

For pending new replies: only the body is applicable and thus editable

Open to suggestions! Cheers 💪 

----

RE: "recording the edit" feature:

I have decided to split that into a separate PR because:

- keeping this PR small is easier for reviews
- it's really a different feature and out of the scope for this PR

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/queued-post.hbs
+++ b/app/assets/javascripts/discourse/templates/components/queued-post.hbs
@@ -20,23 +20,40 @@
     </div>
     <div class='clearfix'></div>
 
-    <span class='post-title'>
-      {{i18n "queue.topic"}}
-      {{#if post.topic}}
-        {{topic-link post.topic}}
-      {{else}}
-        {{post.post_options.title}}
-      {{/if}}
-      {{category-badge post.category}}
-    </span>
+    {{#if editTitleAndCategory}}
+      <span class="edit-title">
+        {{text-field value=buffered.title maxlength=siteSettings.max_topic_title_length}}
+      </span>
+      {{category-chooser value=buffered.category_id}}
+    {{else}}
+      <span class='post-title'>
+        {{i18n "queue.topic"}}
+        {{#if post.topic}}
+          {{topic-link post.topic}}
+        {{else}}
+          {{editables.title}}
+        {{/if}}
+        {{category-badge editables.category}}
+      </span>
+    {{/if}}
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/templates/components/queued-post.hbs
```diff
@@ -20,24 +20,41 @@
...
     </div>
 
+    {{#if showTags}}
+      <div class="list-tags">
+        {{#each tags as |t|}}
+          {{discourse-tag t}}
+        {{/each}}
+      </div>
+    {{else if editTags}}
+      {{tag-chooser tags=buffered.tags categoryId=buffered.category_id width='100%'}}
```
SamSaffron: hmmm should this not be handled by CSS? width 100% makes me nervous

SamSaffron:
overall this PR is super tidy, can't wait to merge it in! thanks

app/assets/javascripts/discourse/templates/components/queued-post.hbs
xrav3nz: I thought this is the intended way to [set the width on the tag chooser `select2` container](https://github.com/discourse/discourse/blob/130ae2cd4b6d60accde8267f51b0442ccd3217a5/app/assets/javascripts/discourse/components/tag-chooser.js.es6#L61)?

Without setting the width manually, the tag chooser is about 10px wider than everything else. `select2` internally sets the width with an inline style, which I cannot override without an `!important`.  /facepalm

p.s. The `100%` is wrapped by the parent `div.cooked`, and this is [set in the CSS.](https://github.com/discourse/discourse/blob/1146772debf861c5c8b6e4e3cebdba31a1138922/app/assets/stylesheets/desktop/queued-posts.scss#L18)

SamSaffron: oh my ... lucky @jjaffeux  is working on nuking select2 completely so this kind of hackery will no longer be needed ... he will sort this out as well when he gets to replacing the tag-chooser

SamSaffron:
thanks heaps for this!

PR #5052, opened 2017-08-16
FEATURE: record edit revisions on queued posts

xrav3nz:
> **Discussion & Screenshots:** https://meta.discourse.org/t/ability-to-change-the-category-of-a-post-before-approving/66754/8?u=xrav3nz

Since we already store complete revisions for regular post edits, why not for queued post edits as well? ;)

This PR utilizes `PostCreator` and `PostRevisor`:

- All edits by staff is recorded to `queued_post.post_options[:changes]`. 
- When approving a pending queued post, `PostCreator` creates a post with the original content, then `PostRevisor` revises the post to reflect the latest changes and creates a `PostRevision`. 
- ~~Post creation and revision are both wrapped in a transaction while relevant processing jobs are only queued upon success.~~ Post creation and revision are take out of the transaction block: 8813b4d.

----

This adds quite a bit of complexity to the current posts approval flow. Open to suggestions on possible ways to simplify the process and/or improve testing. Cheers

### DIFF
--- a/app/models/queued_post.rb
+++ b/app/models/queued_post.rb
@@ -64,37 +64,23 @@ class QueuedPost < ActiveRecord::Base
   end
 
   def approve!(approved_by)
-    created_post = nil
+    created_post = PostCreator.new(user, create_options.merge(skip_validations: true, acting_user: approved_by)).create!
 
-    creator = PostCreator.new(user, create_options.merge(skip_validations: true, skip_jobs: true))
-    revisor = nil
-    QueuedPost.transaction do
+    created_post.revise(
+      User.find(post_options['changes']['editor_id']),
+      post_options['changes'],
+      skip_validations: true,
+      force_new_version: true
+    ) if revised?
+
+    transaction do
+      UserBlocker.unsilence(user, approved_by) if user.silenced?
       change_to!(:approved, approved_by)
-
-      UserSilencer.unsilence(user, approved_by) if user.silenced?
-
-      created_post = creator.create
-
-      unless created_post && creator.errors.blank?
-        raise StandardError.new(creator.errors.full_messages.join(" "))
-      end
-
-      if revised?
-        revisor = PostRevisor.new(created_post)
-        revisor.revise!(
-          User.find(post_options['changes']['editor_id']),
-          post_options['changes'],
-          skip_validations: true,
-          skip_jobs: true,
-          force_new_version: true
-        )
-      end
     end
-
-    # Do sidekiq work outside of the transaction
-    creator.enqueue_jobs
-    revisor.post_process_post if revised?
-
+  rescue
+    PostDestroyer.new(approved_by, created_post).destroy if created_post&.persisted?
+    raise
+  else
     DiscourseEvent.trigger(:approved_post, self, created_post)
     created_post
   end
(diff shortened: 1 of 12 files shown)


xrav3nz:
@SamSaffron do you mind taking a look at this when you're free? 🙈 

Some of the changes in this PR is helpful to the future work on improving the pre-approval queue.

Cheers 💪

app/models/queued_post.rb
```diff
@@ -73,10 +78,22 @@ def approve!(approved_by)
...
         raise StandardError.new(creator.errors.full_messages.join(" "))
       end
+
+      if revised?
+        revisor = PostRevisor.new(created_post)
+        revisor.revise!(
+          User.find(post_options['changes']['editor_id']),
+          post_options['changes'],
+          skip_validations: true,
+          skip_jobs: true,
```
SamSaffron: My concern here is that you are introducing a timing problem 

If post is created, queues a post processor, then is revised ... no post processor runs, you will have a situation where cooked may not be correct. The post process is immediately queued from the post creator. 

In fact I would not do any of the transactioning here, or job omissions can happen, just create and revise outside of transaction, its simplest and if disaster hits you can add a cleanup the just deleted the post / topic.

xrav3nz: Could you please further explain the potential timing problem?

AFAICT, `PostRevisor` does not depend on `cooked`, and the post processor will be queued twice (once by `PostCreator` and another by `PostRevisor`) when the transaction block commits.

Removing the transaction block can be tricky, since it does quite a few things at the same time: change the status of itself (queued post), unblock the user, and create/revise the post. 

Will look into other possible solutions in the meantime.

SamSaffron:
The timing issue (which we have seen before) is that the job can be queued
and executed prior to the transaction in post revisor finishing, that leads
to a nightmare to debug situation

app/models/queued_post.rb
```diff
@@ -60,26 +64,26 @@ def create_options
...
-    # Do sidekiq work outside of the transaction
-    creator.enqueue_jobs
-
-    DiscourseEvent.trigger(:approved_post, self, created_post)
     created_post
+  rescue
+    PostDestroyer.new(approved_by, created_post).destroy if created_post&.persisted?
+    raise
+  else
+    DiscourseEvent.trigger(:approved_post, self, created_post)
```
xrav3nz: This is in an `else` clause so that it will not start the `PostDestroyer`.

app/models/queued_post.rb
```diff
@@ -60,24 +64,23 @@ def create_options
...
+      change_to!(:approved, approved_by)
     end
-
-    # Do sidekiq work outside of the transaction
-    creator.enqueue_jobs
-
+  rescue
+    PostDestroyer.new(approved_by, created_post).destroy if created_post&.persisted?
+    raise
+  else
```
xrav3nz: It's critical to place the transaction block at the end, so that

- should anything raise during post creation/revision: the created post is destroyed.
- should anything raise inside the transaction block: the transaction is rolled back and the created post is destroyed.

xrav3nz:
I'll clean this up and then reopen.

PR #5053, opened 2017-08-16
Spec for local auth check

fantasticfears:


spec/controllers/session_controller_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'rails_helper'
 
 describe SessionController do
+  shared_examples 'failed to continue local local' do
```
tgxworld: Hmm I'm not quite sure what the description means here

spec/controllers/session_controller_spec.rb
```diff
@@ -1,6 +1,12 @@
...
+    it 'renders 500 when tried' do
```
tgxworld: Just `it "should return the right response"` is probably enough

tgxworld:
@fantasticfears Thank you :+1:  Instead of controller tests, can we write the new tests as integration tests? Rails 5 is going to deprecated controller tests so lets reduce the amount of work we need to do

fantasticfears:
What...I googled and didn't found some articles about deprecated controller tests. Can you give me a helpful link for it? If the work is not huge, I can do it for sure. In fact, I am about to split the session controller since it's large to scan through

fantasticfears:
You got me. Those were terrible phrases. Fixed

tgxworld:
Here is the link http://blog.bigbinary.com/2016/04/19/changes-to-test-controllers-in-rails-5.html

fantasticfears:
I see what you mean. And thanks, It's a good read. Though rspec 3.5+ still supports the controller spec which keeps the same helper methods. In a word, this code doesn't age.

See http://rspec.info/blog/2016/07/rspec-3-5-has-been-released/ Rails: Support for Rails 5

In this case, I think as a controller spec are reasonable since no redirection. If I am going to refactor something for SSO, it will be some integration test.

fantasticfears:
Some more research, xhr is in rails4, not in rails5. Only way to use it now would be some magic hack

tgxworld:
Thank you @fantasticfears :+1:

PR #5055, opened 2017-08-16
Annotate table schema

fantasticfears:


SamSaffron:
sure, I went ahead and ran one just now!

PR #5060, opened 2017-08-18
fixes overflowing div width on user activity page

xrav3nz:
My last PR (https://github.com/discourse/discourse/pull/5056) caused a regression:

> Hmm [#5056] regressed on iPad, now it is exceeding the available space and making everything tiny.
> -- @coding-horror, [on meta](https://meta.discourse.org/t/long-words-not-wrapping-in-user-activity-page/68181/7?u=xrav3nz).

Detailed explanation of the fix and why it works: https://meta.discourse.org/t/long-words-not-wrapping-in-user-activity-page/68181/8?u=xrav3nz

### DIFF
--- a/app/assets/stylesheets/desktop/user.scss
+++ b/app/assets/stylesheets/desktop/user.scss
@@ -143,7 +143,6 @@
    }
 
   .user-right {
-    width: 900px;
     display: table-cell;
   }
 


coding-horror:
Hmm I think max-width is more correct here? What about narrow browser widths?

xrav3nz:
The table has a width of `100%`, and this is capped by a parent `.wrap` container (with `max-width: 1110px`).

We already have a fixed `170px` width on the left column (the navigation area), thus setting another fixed `900px `on the right column (content area) will force the table width to be a fixed `~1070px`. This caused the problem you discovered on iPad (i.e. narrow browser width).

By dropping the fixed width on the right column, it will automatically adjust to `100% - left_width`.

coding-horror:
all right let's try it then

xrav3nz:
🙈 🤞

PR #5061, opened 2017-08-18
Extend config/version.rb with more informations

darix:
This gives installations not using git checkouts
to provide all the informations needed for the
internal version checks and version display in
the dashboard.

The build:stamp rake task was extended to also
add the new informations.

### DIFF
--- a/lib/discourse.rb
+++ b/lib/discourse.rb
@@ -311,25 +311,44 @@ module Discourse
   def self.git_version
     return $git_version if $git_version
 
-    # load the version stamped by the "build:stamp" task
-    f = Rails.root.to_s + "/config/version"
-    require f if File.exists?("#{f}.rb")
-
-    begin
-      $git_version ||= `git rev-parse HEAD`.strip
-    rescue
-      $git_version = Discourse::VERSION::STRING
-    end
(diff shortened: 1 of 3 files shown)


SamSaffron:
OK, let's try this out, thanks @darix

lib/discourse.rb
```diff
@@ -311,25 +311,44 @@ def self.request_refresh!(user_ids: nil)
...
+    if File.exists?(version_file)
+      require version_file
+      version_value = yield
     end
+
+    # file does not exist or does not define the expected global variable
+    unless version_value
+      begin
+        version_value = `#{git_cmd}`.strip
+      rescue # sollte noch ausspezifiziert werden…
```
ZogStriP: It would be better to keep the comments in English ;)

darix: the worse part is that it is a TODO and i wanted to specify the common exceptions i want to catch instead of the broad sweep handler :P

Just remove the comment or actually limit the exceptions we want to handle there as well?

ZogStriP: I depends, can we do anything on the exceptions? If not, then it's fine to leave it that way.

PR #5062, opened 2017-08-18
Preparation for using chrome for qunit in docker images

davidtaylorhq:
Move use_chrome option to ENV variable
Rewrite script to work with node 6 (current LTS version used in discourse_docker)
Add node stuff to gitignore


SamSaffron:
cool lets try it!

PR #5063, opened 2017-08-18
Unnest top level selectors from `body`

minusfive:
Top level selectors are unnecessarily nested under the `body` element selector, which causes unnecessary output/css bloat and extra unnecessary nesting + extra stray selectors on theme overrides when using Sass' `@extend`.

[Here's a gist showing `desktop.css` and `mobile.css` output diffs.](https://gist.github.com/minusfive/111efb4ad18fefdec70e2471218f6762/revisions)


eviltrout:
Thanks!

PR #5065, opened 2017-08-18
Split alias levels in mentionable and messageable levels.

nbianca:


SamSaffron:
This looks really nice, but we got to add a few tests here, it has a bunch of security implications we want to make sure we did not miss anything.

nbianca:
There are also tests failing so I'm going to take care of that too.

nbianca:
There are few groups (:moderators, :admins, :trust_level_1, :trust_level_2). By default, these are now messageable, but I believe they were not before?

SamSaffron:

PR #5066, opened 2017-08-18
Docker lint SINGLE_PLUGIN

davidtaylorhq:
@tgxworld this restores `SINGLE_PLUGIN` functionality to the discourse_test docker image linting after https://github.com/discourse/discourse/commit/8ecf383c557d2ff7dbcd22b4faa942c235d44f2a - I use it for testing plugins on travis and don't really want to be linting the entirety of discourse core every time.

Hope it's ok 😃

### DIFF
--- a/lib/tasks/docker.rake
+++ b/lib/tasks/docker.rake
@@ -35,11 +35,18 @@ end
 
 desc 'Run JS and Ruby linters'
 task 'docker:lint' do
-  success = run_or_fail("bundle exec rubocop --parallel")
-  success = run_or_fail("eslint app/assets/javascripts test/javascripts")
-  success = run_or_fail("eslint --ext .es6 app/assets/javascripts test/javascripts plugins")
+  success = true
 
-  exit 1 if !success
+  if ENV["SINGLE_PLUGIN"]
+    success &&= run_or_fail("bundle exec rubocop --parallel plugins/#{ENV["SINGLE_PLUGIN"]}")
+    success &&= run_or_fail("eslint --ext .es6 plugins/#{ENV['SINGLE_PLUGIN']}")
+  else
+    success &&= run_or_fail("bundle exec rubocop --parallel") unless ENV["SKIP_CORE"]
+    success &&= run_or_fail("eslint app/assets/javascripts test/javascripts") unless ENV["SKIP_CORE"]
+    success &&= run_or_fail("eslint --ext .es6 app/assets/javascripts test/javascripts plugins") unless ENV["SKIP_PLUGINS"]
+  end
+
+  exit 1 unless success
 end
 
 desc 'Run all tests (JS and code in a standalone environment)'
--- a/lib/tasks/docker.rake
+++ b/lib/tasks/docker.rake
@@ -2,6 +2,8 @@
 # running it anywhere else will likely fail
 #
 # Environment Variables (specific to this rake task)
+# => SKIP_LINT                 set to 1 to skip linting (eslint and rubocop)
+# => SKIP_TESTS                set to 1 to skip all tests
 # => SKIP_CORE                 set to 1 to skip core tests (rspec and qunit)
 # => SKIP_PLUGINS              set to 1 to skip plugin tests (rspec and qunit)
 # => INSTALL_OFFICIAL_PLUGINS  set to 1 to install all core plugins before running tests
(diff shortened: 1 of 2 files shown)


lib/tasks/docker.rake
```diff
@@ -33,11 +35,18 @@ end
...
+    success &&= run_or_fail("bundle exec rubocop --parallel plugins/#{ENV["SINGLE_PLUGIN"]}")
+    success &&= run_or_fail("eslint --ext .es6 plugins/#{ENV['SINGLE_PLUGIN']}")
+  else
+    success &&= run_or_fail("bundle exec rubocop --parallel") unless ENV["SKIP_CORE"]
+    success &&= run_or_fail("eslint app/assets/javascripts test/javascripts") unless ENV["SKIP_CORE"]
+    success &&= run_or_fail("eslint --ext .es6 app/assets/javascripts test/javascripts plugins") unless ENV["SKIP_PLUGINS"]
+  end
 
-  exit 1 if !success
+  exit 1 unless success
```
SamSaffron: just a tiny one... I find unless SUPER confusing. in this case it is questionably ok, but it makes code require tons of mental gymnastics just to figure out what it does

open_door unless police_officer || clown

open_door if !policy_officer && !clown

davidtaylorhq: Fair enough, will put that back how it was

SamSaffron:
pr seems fine to me! @tgxworld feel free to merge.

lib/tasks/docker.rake
```diff
@@ -33,9 +35,16 @@ end
 
 desc 'Run JS and Ruby linters'
 task 'docker:lint' do
-  success = run_or_fail("bundle exec rubocop --parallel")
-  success = run_or_fail("eslint app/assets/javascripts test/javascripts")
-  success = run_or_fail("eslint --ext .es6 app/assets/javascripts test/javascripts plugins")
+  success = true
+
+  if ENV["SINGLE_PLUGIN"]
```
tgxworld: `PLUGIN_NAME` or `PLUGIN` is clearer to me.

lib/tasks/docker.rake
```diff
@@ -14,6 +14,8 @@
 # Other useful environment variables (not specific to this rake task)
 # => COMMIT_HASH    used by the discourse_test docker image to load a specific commit of discourse
 #                   this can also be set to a branch, e.g. "origin/tests-passed"
+# => SKIP_LINT      if using the discourse_test docker image, this will skip the docker:test task, and only run the docker:lint task
```
tgxworld: Shouldn't this be added to `docker_test.rb` instead?

lib/tasks/docker.rake
davidtaylorhq: It's been SINGLE_PLUGIN for the past month (since https://github.com/discourse/discourse/pull/4973). Can change it if absolutely necessary, but it will mean making the change in `docker:test`, adjusting travis config on a few plugins, and updating howto docs on Meta. Let me know 🙂

lib/tasks/docker.rake
davidtaylorhq: Sure, when I added this documentation I was trying to have a "single place" where all the info for using the docker container could be found. Note that this is under the "not specific to this rake task" heading. That said, it would be useful to have it in docker_test.rb as well - I will copy the relevant stuff over later today.

tgxworld: If you don't mind can you extract `script/docker_test.rb` into a Rake test in this file?

davidtaylorhq: It would be tidier, but I worry it could cause issues. @SamSaffron was there a reason it was refactored out of the rake task in https://github.com/discourse/discourse/commit/7289e0b5fd0aafd0fd8004be373e6487d76fdbbe ?

Maybe something to do with trying to do `git pull` while `rake` is running? I doubt it would go well trying to update files that are currently open... Or maybe that dependencies need to be installed **after** `git pull`, but before running the rake task...

SamSaffron: no really sure what the point is of mucking with docker_test.rb ... it is so thin on logic, it was just intended a tiny script that updates the git repo and launches the rake task. 

Given this structure it allows to update the rake task and pick up the update without needing to build a new base.

I wonder if anything if the "lint/test" stuff should be removed from there. 

Conceptually the "linting" is part of testing and perhaps that should just be another flag to the test rake task.

tgxworld: Something like

```
rake docker:test['lint_only']
```
?

davidtaylorhq: Using environment variables will make it much easier to change when running inside a docker container in CI, so how about `script/docker_test.rb` simply runs `rake docker:test`

Then inside the `rake docker:test` task we have both linting and testing. The linting bit can be surrounded by `unless ENV['SKIP_LINT']`, and the testing bit can be surrounded by `unless ENV['SKIP_TESTS']`

tgxworld: Sure sounds fine to me :

davidtaylorhq:
@tgxworld I have combined them into one rake task, surrounded by `unless ENV['SKIP_LINT']` and `unless ENV['SKIP_TESTS]`. Unfortunately the diff is messy because of the indentation changes.

SamSaffron:
I like this, @tgxworld feel free to merge and add to our CI

tgxworld:
Thank you for the awesome work @davidtaylorhq

PR #5067, opened 2017-08-18
Improved metadata for tags.

nbianca:


nbianca:
I do not believe that the failed test is my fault.

SamSaffron:
let's see what this does, thanks!

PR #5068, opened 2017-08-19
last attempt at fixing user page layout

xrav3nz:
> CONTEXT: https://meta.discourse.org/t/long-words-not-wrapping-in-user-activity-page/68181

drop the width setting on the right column *completely* so it can adjust on it own.

---

_Yet again_, my last PR broke the user preference page layout... 😢 

**What happened?**

The right column is a `div.user-right.user-preferences`:

Turns out it has _two_ different CSS width properties. After dropping `width: 900px` with my last PR, `width: 170px` was secretly applied. This led to the layout problem eviltrout discovered.

---

Many thanks to @davidtaylorhq, who kindly helped me to test the change in various browsers / devices / widths. ❤️

### DIFF
--- a/app/assets/stylesheets/desktop/user.scss
+++ b/app/assets/stylesheets/desktop/user.scss
@@ -19,7 +19,15 @@
 }
 
 .user-preferences {
+  display: table-cell;
+  vertical-align: top;
   padding-top: 10px;
+  padding-left: 30px;
+
+  h3 {
+    color: $primary;
+    margin: 20px 0 10px 0;
+  }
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
No worries, don't feel bad, thanks so much for your efforts, we will have a
look at the PR next week!

coding-horror:
This is super broken on iPad still right now, using meta as a baseline, so I am going to merge this.. overall this change has been kind of a disaster

xrav3nz:
I have to admit that you're right -- this has been a disaster. Thank you for bearing with me; it has been quite a lesson for me.

PR #5069, opened 2017-08-20
Password mail with IP addresses.

begrif:
Save the user's IP address and browser `User-Agent:` when creating password reset tokens. Then when sending password reset email, include those in the message sent.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -248,7 +248,7 @@ class SessionController < ApplicationController
     user = User.find_by_username_or_email(params[:login])
     user_presence = user.present? && user.id > 0 && !user.staged
     if user_presence
-      user_agent = request.user_agent || '(unidentified browser)'
+      user_agent = ERB::Util.html_escape(request.user_agent) || '(unidentified browser)'
       email_token = user.email_tokens.create(email: user.email, remote_ip: request.remote_ip, user_agent: user_agent)
       Jobs.enqueue(:critical_user_email, type: :forgot_password, user_id: user.id, email_token: email_token.token)
     end
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -248,7 +248,7 @@ class SessionController < ApplicationController
     user = User.find_by_username_or_email(params[:login])
     user_presence = user.present? && user.id > 0 && !user.staged
     if user_presence
-      user_agent = ERB::Util.html_escape(request.user_agent) || '(unidentified browser)'
+      user_agent = request.user_agent || '(unidentified browser)'
       email_token = user.email_tokens.create(email: user.email, remote_ip: request.remote_ip, user_agent: user_agent)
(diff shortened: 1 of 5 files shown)


db/migrate/20170816343911_add_password_change_ip_to_email_tokens.rb
```diff
@@ -0,0 +1,11 @@
+class AddPasswordChangeIpToEmailTokens < ActiveRecord::Migration
+  def up
+    add_column :email_tokens, :remote_ip, :inet
```
tgxworld: Can the following columns be `null`? :thought_balloon:

app/controllers/session_controller.rb
```diff
@@ -248,7 +248,8 @@ def forgot_password
     user = User.find_by_username_or_email(params[:login])
     user_presence = user.present? && user.id > 0 && !user.staged
     if user_presence
-      email_token = user.email_tokens.create(email: user.email)
+      user_agent = request.user_agent || '(unidentified browser)'
```
tgxworld: `(unidentified browser)` should probably be translated using `I18n` here

db/migrate/20170816343911_add_password_change_ip_to_email_tokens.rb
begrif: Yes.

app/controllers/session_controller.rb
begrif: Sure, I can do that.

begrif:
I think the ideal situation is that this is logged and admins of more technical forums can enable it in emails at their sites, but it is not shown by default.

SamSaffron:
we could possibly enable this out of the box for moderators and admins. I am not sure. I think you should split this up. 

The first non-controversial change is just to log somewhere. cause then you could look at the user log and see that someone is trying to compromise an account and see where it is.

coding-horror:
Sure I support displaying the extra info for staff (mods/admins) only at
first. That is a fine idea.

pain0x0:
yes long user agent strings can confuse non-tech people but still would be ok to still have it i guess?
and yes a panel in admin section where it logs password reset and show's ip etc would be good and having option where user can report if its not them

coding-horror:
As long as user agent is only shown to staff password resets I support merging this in @ZogStriP

app/mailers/user_notifications.rb
```diff
@@ -48,10 +48,25 @@ def confirm_new_email(user, opts = {})
   end
 
   def forgot_password(user, opts = {})
+    remote_ip = I18n.t('no_remote_ip')
+    user_agent = I18n.t('no_browser_user_agent')
+
+    if opts[:email_token] then
```
ZogStriP: We only use `then` in inline `case` statements.

ZogStriP:
Also, this needs to be only shown to `staff` members.

SamSaffron:
We can't keep this open forever, closing for now, @begrif can you resubmit a PR when it is ready.

PR #5070, opened 2017-08-21
FEATURE: prettify slug generation and specs

fantasticfears:
Regards the null out after changing the site settings. I am afraid it's expensive. 2 options, 1st trigger a regular job. 2nd regular cleaner like reindexer.

### DIFF
--- a/lib/slug.rb
+++ b/lib/slug.rb
@@ -5,7 +5,7 @@ module Slug
   CHAR_FILTER_REGEXP = /[:\/\?#\[\]@!\$&'\(\)\*\+,;=_\.~%\\`^\s|\{\}"<>]+/ # :/?#[]@!$&'()*+,;=_.~%\`^|{}"<>
   MAX_SLUG_LENGTH = 255
 
-  def self.for(string, default = 'topic')
+  def self.for(string, default = 'topic', max_length = MAX_SLUG_LENGTH)
     slug =
       case (SiteSetting.slug_generation_method || :ascii).to_sym
       when :ascii then self.ascii_generator(string)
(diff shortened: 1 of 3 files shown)


fantasticfears:
@SamSaffron I have a disparity here. https://travis-ci.org/discourse/discourse/builds/267106586
In my local chrome, it works with this changeset even it breaks CI. Chrome always send the percent encoded string. Rails encode the parameter wrong. So our persistent strategy in this changeset which stores the percent encoded slug does not work.
So hack slug parameters?

SamSaffron:
@tgxworld will review this, but we want the rails 5 upgrade done first

fantasticfears:
As an aside, Rails 5.1 returns the UTF-8 string. So I'll revisit this after our Rails 5 upgrade.

tgxworld:
Hi @fantasticfears, Rails 5.1 is more or less stable. Let me know if this PR needs to be updated.

fantasticfears:
Sure, I am going to review this PR.

SamSaffron:
not following, do we even need any changes here post 5.1 ... I thought the framework fixed it.

fantasticfears:
If you no longer need to persist it to encoded, we should put on hold on this... But I will extract the test though.

lib/slug.rb
```diff
@@ -13,11 +14,12 @@ def self.for(string, default = 'topic')
       end
     # Reject slugs that only contain numbers, because they would be indistinguishable from id's.
     slug = (slug =~ /[^\d]/ ? slug : '')
+    slug = slug.length >= MAX_SLUG_LENGTH ? '' : slug
```
SamSaffron: are we sure about this, why not truncate?

fantasticfears: I was definitely out of my mind...will fix

fantasticfears:
rebased and spec fixed

SamSaffron:
I am just so confused here, what specifically is this PR fixing now?

fantasticfears:
Sorry, I should change the title and leave comments as well.

Now the encoded generator will try to downcase and restrict slug string which is longer than a very large number 255.
Moreover, previous encoded generator may leave out a trailing dash. This is fixed in this PR too.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/7c3123a2dd814663c86cd6b4ed3619ca3c02aab4

Sorry for the delay @fantasticfears

PR #5073, opened 2017-08-22
FEATURE: import Github profile picture

xrav3nz:
> DISCUSSION: https://meta.discourse.org/t/import-github-image-as-profile-image-when-authenticating-with-github/49132

Bumped into the discussion topic when I was looking into some omniauth-related stuff.

The behaviour is consistent with the existing Facebook/TwitterAuthenticator:

- if the user had uploaded a custom avatar: Nothing happens
- if not: Retrieve and upload the avatar. Select this as the avatar unless the user had selected gravatar.

### DIFF
--- a/lib/auth/github_authenticator.rb
+++ b/lib/auth/github_authenticator.rb
@@ -91,6 +91,8 @@ class Auth::GithubAuthenticator < Auth::Authenticator
       end
     end
 
+    retrieve_avatar(user, data)
+
     result.user = user
     result
   end
@@ -102,6 +104,8 @@ class Auth::GithubAuthenticator < Auth::Authenticator
       screen_name: data[:github_screen_name],
       github_user_id: data[:github_user_id]
     )
+
+    retrieve_avatar(user, data)
   end
 
   def register_middleware(omniauth)
@@ -112,5 +116,16 @@ class Auth::GithubAuthenticator < Auth::Authenticator
               strategy.options[:client_secret] = SiteSetting.github_client_secret
            },
            scope: "user:email"
+  end
+
+  protected
+
+  def retrieve_avatar(user, data)
+    return unless user
+    return if user.user_avatar&.custom_upload_id.present?
+
+    if (avatar_url = data[:image]).present?
+      Jobs.enqueue(:download_avatar_from_url, url: avatar_url, user_id: user.id, override_gravatar: false)
+    end
   end
 end
--- a/lib/auth/github_authenticator.rb
+++ b/lib/auth/github_authenticator.rb
@@ -118,14 +118,11 @@ class Auth::GithubAuthenticator < Auth::Authenticator
            scope: "user:email"
   end
 
-  protected
+  private
 
   def retrieve_avatar(user, data)
-    return unless user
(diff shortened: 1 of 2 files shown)


lib/auth/github_authenticator.rb
```diff
@@ -113,4 +117,15 @@ def register_middleware(omniauth)
...
   end
+
+  protected
+
+  def retrieve_avatar(user, data)
+    return unless user
+    return if user.user_avatar&.custom_upload_id.present?
+
+    if (avatar_url = data[:image]).present?
+      Jobs.enqueue(:download_avatar_from_url, url: avatar_url, user_id: user.id, override_gravatar: false)
```
xrav3nz: `avatar_url` will be something like `https://avatars0.githubusercontent.com/u/6376558?v=4`. 

It has an optional parameter size `s`, with a default and maximum at 460px x 460px.

coding-horror:
Nice work!

spec/components/auth/github_authenticator_spec.rb
```diff
@@ -169,4 +182,30 @@
...
 
   end
+
+  describe 'avatar retrieval' do
+    context 'when user has a custom avatar' do
+      let(:user_avatar) { Fabricate(:user_avatar, custom_upload: Fabricate(:upload)) }
+      let(:user_with_custom_avatar) { Fabricate(:user, user_avatar: user_avatar) }
+
+      it 'does not enqueue a download_avatar_from_url job' do
+        Jobs.expects(:enqueue).with(:download_avatar_from_url, anything).never
```
tgxworld: Instead of mocking, lets configure `SiteSetting.queue_jobs = false` and test for the actual job in the queue. This may be helpful https://github.com/mperham/sidekiq/wiki/Testing

spec/components/auth/github_authenticator_spec.rb
```diff
@@ -169,4 +182,30 @@
...
+
+        result = authenticator.after_authenticate(auth_token_for(user_with_custom_avatar))
```
tgxworld: The `result` variable is probably not needed here?

lib/auth/github_authenticator.rb
```diff
@@ -113,4 +117,15 @@ def register_middleware(omniauth)
            },
            scope: "user:email"
   end
+
+  protected
```
tgxworld: It is unlikely, at least for the moment, that someone would create a subclass from this class. Can we make this a private method instead? I would just add the test cases into one of the existing test case where the authentication is successful.

lib/auth/github_authenticator.rb
```diff
@@ -113,4 +117,15 @@ def register_middleware(omniauth)
...
+
+  def retrieve_avatar(user, data)
+    return unless user
+    return if user.user_avatar&.custom_upload_id.present?
+
+    if (avatar_url = data[:image]).present?
```
tgxworld: I would move this condition to the first line of this method and just `return` if there isn't `avatar_url`

xrav3nz:
Thanks for the pointers and TIL `sidekiq/testing`!

I added an additional commit just to reflect the changes; feel free to squash later.

PR #5075, opened 2017-08-22
fix misuse of TLD

johnmuhl:


config/locales/server.en.yml
```diff
@@ -980,7 +980,7 @@ en:
     fixed_category_positions: "If checked, you will be able to arrange categories into a fixed order. If unchecked, categories are listed in order of activity."
     fixed_category_positions_on_create: "If checked, category ordering will be maintained on topic creation dialog (requires fixed_category_positions)."
     add_rel_nofollow_to_user_content: "Add rel nofollow to all submitted user content, except for internal links (including parent domains). If you change this, you must rebake all posts with: \"rake posts:rebake\""
-    exclude_rel_nofollow_domains: "A list of domains where nofollow should not be added to links. tld.com will automatically allow sub.tld.com as well. As a minimum, you should add the top-level domain of this site to help web crawlers find all content. If other parts of your website are at other domains, add those too."
```
tgxworld: I don't think we need to remove `top-level` here?

johnmuhl: The `top-level domain of this site` is going to be `com`, `net`, `org` etc and I don't think the instructions mean for you to add `com` to this field; using the term `top-level` seems confusing to me. Perhaps `..add the main domain of this site...` would be more clear.

tgxworld: Ah I see what you mean now. Thanks for explaining.

tgxworld:
Thank you for the PR @johnmuhl

PR #5078, opened 2017-08-23
change count>0 to exists

lelelelemon:
count > 0 will issue query select count(*) 
exists? will issue query select 1 limit 1, which is faster.


coding-horror:
How much faster, did you benchmark the change for comparison?

tgxworld:
Looks good to me @lelelelemon :+1: A `limit 1` is more efficient in cases where there are multiple records.

PR #5084, opened 2017-08-24
Removed icons + no-glyph classes from /user sub-navs, fixed indentation

minusfive:


eviltrout:
We "temporarily" removed icons via CSS. Let's make it permanent.

PR #5087, opened 2017-08-24
Added spanish translation for user/activity/no_replies message

minusfive:


eviltrout:
Sorry but we have to do non-english languages through transifex

PR #5089, opened 2017-08-24
Suggest running discourse not script/discourse

pfaffman:
Having `discourse restore` offer copy/paste of backups is awesome, but doesn't work (rails won't load) if you call script/discourse.

### DIFF
--- a/script/discourse
+++ b/script/discourse
@@ -91,7 +91,7 @@ class DiscourseCLI < Thor
       puts "You must provide a filename to restore. Did you mean one of the following?\n\n"
 
       Dir["public/backups/default/*"].each do |f|
-        puts "script/discourse restore #{File.basename(f)}"
+        puts "discourse restore #{File.basename(f)}"
       end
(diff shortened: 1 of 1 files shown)


SamSaffron:
hmmm it runs script/discourse... 

https://github.com/discourse/discourse_docker/blob/master/templates/web.template.yml#L175-L180

but I guess the greater good here is just to write discourse here.

pfaffman:
Well, here's [this discussion](https://meta.discourse.org/t/command-line-restore-for-web-only-container-cant-load-thor/68570). As @tgxworld  points out, the Greater Good is to guess whether the script is running  in a production container or a development environment and act accordingly.

Now that I finally understand when to use `discourse` vs `script/discourse` I think all that's required to tell would be a test for `/usr/local/bin/discourse`; would you like for me to add that and submit another PR?

coding-horror:
Yes please go ahead with that

PR #5090, opened 2017-08-24
Fix + remove unnecessary indentation, dead code, cleanup user.css

minusfive:
This also fixes user-stream padding on mobile for empty notifications/activity views.


eviltrout:
Thank you!

PR #5094, opened 2017-08-25
Simplify user-nav styles to facilitate theming

minusfive:
Icons will now inherit the text colour, which will make the thicker ones look a tad darker, but seems to me a reasonable compromise for removing code which ultimately [unnecessarily] complicates theming.

I'm also using the existing `.user-nav` class rather than the unnecessarily nested `.user-main .nav-pills` used before, which should further facilitate targeting the element for theming.

### DIFF
--- a/app/assets/stylesheets/common/base/user.scss
+++ b/app/assets/stylesheets/common/base/user.scss
@@ -3,23 +3,6 @@
   .d-icon-heart {
     color: $love !important;
   }
-  .nav-pills {
-    a {
-      color: dark-light-choose(scale-color($primary, $lightness: 40%), scale-color($secondary, $lightness: 40%));
-    }
-
-    a:hover i {
-      color: $quaternary;
-    }
-
-    a.active i {
-      color: $secondary;
-    }
-
-    i {
-      color: dark-light-choose(scale-color($primary, $lightness: 55%), scale-color($secondary, $lightness: 55%));
-    }
-  }
 }
 
 .user-field {
@@ -161,11 +144,9 @@
 .user-nav {
   margin: 5px 0px;
   padding-top: 10px;
-  .d-icon {
-    margin-right: 5px;
-  }
-  .d-icon-comment {
-    margin-right: 2px;
+
+  li a {
+    color: dark-light-choose(scale-color($primary, $lightness: 40%), scale-color($secondary, $lightness: 40%));
   }
 }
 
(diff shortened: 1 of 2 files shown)


minusfive:
The icon used for "notifications" seems to be the one which stands-out the most as the darkest. One solution to make the menu look more harmonious could be to [use the outline-only version of that icon](http://fontawesome.io/icon/comment-o/). Thoughts @eviltrout ?

coding-horror:
Er what? Stop the presses, I would like to see screenshots of this proposed
change before it goes in please?

minusfive:
@coding-horror no problem, will post screenshots

minusfive:
@coding-horror, @eviltrout here are the screenshots:

### Current (Before PR)


### After (PR)


### Suggested "Notifications" Icon Change
In case the new styles make the current [filled] icon seem too dark

minusfive:
And here what they look like in dark, just in case:

### Current (Before PR)


### After (PR)


### Suggested "Notifications" Icon Change
In case the new styles make the current [filled] icon seem too dark


It looks like it has the opposite effect in the dark theme (icons look lighter), but the difference is still pretty subtle methinks.

minusfive:
Just forced-pushed a rebase with recent commits merged on master

eviltrout:
Looks fine to me. We can merge with @coding-horror's approval.

coding-horror:
I guess it is fine with the glyph change, but in general the glyphs are supposed to be lighter than the text. Consider the buttons at the bottom of every topic, or even here on github look at the glyphs next to the tabs.

img_0899

minusfive:
@coding-horror so if we really want to retain that look, I'd suggest accomplishing it by modifying the icon's opacity, rather than specifying a different colour, which would still be simpler to theme. Here are some screenshots of a quick mock of that working on both light and dark themes, along with the option of retaining the color difference on selection, or making the icon's color match the text's when selected (like Github does it):

### Light - Icon opacity `.6`


### Light - Selected - Icon opacity `.6`


### Light - Selected - Icon opacity `1`
![])(http://take.ms/oCe8J)

### Dark - Icon opacity `.6`


### Dark - Selected - Icon opacity `.6`


### Dark - Selected - Icon opacity `1`


Let me know which would you prefer and I can add a commit to this PR before merging.

minusfive:
I'd suggest icon opacity `1` on selected, but that's just my opinion ;).

coding-horror:
As long as we achieve the goal, that's fine by me. See also:

https://twitter.com/i/moments/880688233641848832

minusfive:
OK, I've pushed a new version including this tweak — I'm applying `opacity: .65;` to any `.d-icon` used on `.nav-pills > li > a`. So the icon should be 35% lighter than the text (or darker on dark themes), regardless of text colour.

PR #5096, opened 2017-08-25
To enable the linkedin plugin to work

robglarsen:
gem omniauth-oauth2 1.4.0 breaks the linkedin plugin

Downgrading to omniauth-oauth2 to 1.3.1 allows the plugin to work.

The plugin is here 

https://github.com/discourse/discourse-plugin-linkedin-auth


cpradio:
Related discussion:
https://meta.discourse.org/t/linkedin-oauth2-plugin/46818/27?u=cpradio

robglarsen:
CLA done

cpradio:
Cool, since this had to be made in the discourse project, one of the team will have to approve it.

SamSaffron:
I definitely can not accept such a widescale change on our lockfile, that is upgrading TONS of gems each which need careful testing. 

Regarding downgrading omniauth. I would much prefer if the dependency on our side is simply updated cause there is a fix for the gem upstream 

https://github.com/decioferreira/omniauth-linkedin-oauth2/commits/master

I just updated the dependency on the plugin, which should hopefully fix it.

PR #5104, opened 2017-08-26
Add avatar-flair plugin outlet for user card

davidtaylorhq:
This allows plugins to add content next to the avatar in a user card. I'd like to use this in discourse-whos-online to add a 'presence' icon.

screen shot 2017-08-26 at 23 17 28

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
+++ b/app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
@@ -10,6 +10,7 @@
           flairColor=user.primary_group_flair_color
           groupName=user.primary_group_name}}
       {{/if}}
+      {{plugin-outlet name="user-card-avatar-flair" args=(hash user=user) tagName='div'}}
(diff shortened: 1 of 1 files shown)


coding-horror:
This looks good but what about a green ring outline? Too much?

davidtaylorhq:
Yeah I've experimented with a "glow" around the avatar which looks quite nice:

screen shot 2017-08-27 at 23 39 54

One problem is that it conflicts with the "Original poster, most recent poster" glow
screen shot 2017-08-27 at 23 43 17

The other problem is that I think adding classes to each avatar image from a plugin will be a lot harder than adding "flair", so I thought I'd start with the easy option ;)

coding-horror:
Hmm yeah in that case, **if presence is the overriding concern** I would just suppress the "original poster, most recent poster" glow since it is a bit of a special case anyway.

SamSaffron:
Sure, the outlet looks fine to me!

PR #5105, opened 2017-08-28
FIX: make /new-topic redirect survive full screen login

LeoMcA:
This is a reversion, of sorts, to #3555, because the change meant that the redirect didn't survive full screen login, making the /new-topic route rather useless for sites using fsl (like my own).

#3555 came after @SamSaffron's [comment](https://github.com/discourse/discourse/commit/d1632c1dbd256411bbb348f638bb7f5f61f05eae##commitcomment-11694599):
> $.cookie is odd usage, I would not like us to use this. we need a different way, we have sessions for this kind of stuff.

To respect this, I originally tried accessing `shouldRedirectToUrl` from the session in `app/controllers/users/omniauth_callbacks_controller.rb`, but it looks like the session didn't survive authentication, because it wasn't there.

I imagine for this reason the established principle, across the codebase, is to use a `destination_url` cookie for redirect:

[`app/controllers/application_controller.rb`](https://github.com/discourse/discourse/blob/master/app/controllers/application_controller.rb#L568-L570):
```ruby
# save original URL in a cookie (javascript redirects after login in this case)
cookies[:destination_url] = destination_url
redirect_to :login
```

[`app/controllers/users/omniauth_callbacks_controller.rb`](https://github.com/discourse/discourse/blob/master/app/controllers/users/omniauth_callbacks_controller.rb#L44-L47):
```ruby
if cookies[:destination_url].present?
  origin = cookies[:destination_url]
  cookies.delete(:destination_url)
end
```

I humbly propose if we don't want to use cookies for this, we later move to sessions in one fell swoop - rather than having multiple, some partially functional, post-authentication redirection methods.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/login.js.es6
+++ b/app/assets/javascripts/discourse/controllers/login.js.es6
@@ -90,7 +90,6 @@ export default Ember.Controller.extend(ModalFunctionality, {
           // Trigger the browser's password manager using the hidden static login form:
           const $hidden_login_form = $('#hidden-login-form');
           const destinationUrl = $.cookie('destination_url');
-          const shouldRedirectToUrl = self.session.get("shouldRedirectToUrl");
           const ssoDestinationUrl = $.cookie('sso_destination_url');
           $hidden_login_form.find('input[name=username]').val(self.get('loginName'));
           $hidden_login_form.find('input[name=password]').val(self.get('loginPassword'));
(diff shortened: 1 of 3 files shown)


SamSaffron:
I stand by shouldRedirectToUrl as a cookie is a potential problem cause why would we need that cookie AND destination_url

I wish this was a bit more tested... at the moment it is very easy for it to regress.

PR #5108, opened 2017-08-29
FIX: Only show topic timer to staff

jomaxro:
Resolves https://meta.discourse.org/t/how-to-create-a-timed-topic/60731/9


tgxworld:
LGTM :+1:

PR #5113, opened 2017-08-29
FEATURE: Add placeholders to broken and large image files

vinothkannans:
As per [this](https://meta.discourse.org/t/add-setting-to-disable-hotlinking-of-large-image-files/67909/12?u=vinothkannans) pr-welcome.

### DIFF
--- a/app/jobs/regular/pull_hotlinked_images.rb
+++ b/app/jobs/regular/pull_hotlinked_images.rb
@@ -25,6 +25,7 @@ module Jobs
       raw = post.raw.dup
       start_raw = raw.dup
       downloaded_urls = {}
+      broken_images, large_images = [], []
 
       extract_images_from(post.cooked).each do |image|
         src = original_src = image['src']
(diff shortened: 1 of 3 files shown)


SamSaffron:
Can you add some screenshot here?

vinothkannans:
ss-broken-large-image-samples

In above screenshot large image placeholder is linked to original image URL.

coding-horror:
Looks good to me!

app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -98,6 +101,28 @@ def execute(args)
         post.revise(Discourse.system_user, changes, options)
       elsif downloaded_urls.present?
         post.trigger_post_process(true)
+      elsif broken_images.present? || large_images.present?
```
SamSaffron: something does not look right to me... 

What if both downloaded_urls are present AND broken_images / large_images are present , can you add a test that has all 3 types of images (broken, too big, just right)

vinothkannans: It won't be a problem. If `downloaded_urls` are present then we are triggering post process. There `downloaded_urls` will be replaced with local urls in cooked html. Then again post process will trigger `pull_hotlinked_images` job. Now `downloaded_urls` will be empty since it already replaced. So other condition `elsif broken_images.present? || large_images.present?` will happen now.

Also I added spec function with all the combinations
https://github.com/discourse/discourse/pull/5113/files#diff-aec0754ea13a668c66ff2216f7a96c37R95

SamSaffron:
cool, let's try this out!

PR #5114, opened 2017-08-29
uses new select-box for quick topic edit

jjaffeux:


jjaffeux:
Note: I don’t like this `width=350` as it makes it hard to customise it, I will come up with a better solution in the next days.

PR #5120, opened 2017-08-30
FIX: redirect loop for new users visiting /new-topic using full screen login

LeoMcA:
#5105 exposed a new problem where users who were required to complete the sign up modal would get redirected back to the full screen login before being able to do so. This fixes that.

Wasn't sure whether `Discourse.showing_signup` was the best place to store that variable, happy to move it elsewhere.

### DIFF
--- a/app/assets/javascripts/discourse.js.es6
+++ b/app/assets/javascripts/discourse.js.es6
@@ -8,6 +8,7 @@ const Discourse = Ember.Application.extend({
   _docTitle: document.title,
   RAW_TEMPLATES: {},
   __widget_helpers: {},
+  showingSignup: false,
 
   getURL(url) {
     if (!url) return url;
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse.js.es6
```diff
@@ -7,6 +7,7 @@ const Discourse = Ember.Application.extend({
   rootElement: '#main',
   _docTitle: document.title,
   RAW_TEMPLATES: {},
+  showing_signup: false,
```
eviltrout: in javascript we use camelCase like `showingSignup`, so that should be changed.

LeoMcA:
@eviltrout updatedToCamelCase

tgxworld:
@techAPJ Does this look OK to you?

techAPJ:
Looks good to me. @LeoMcA can you resolve the conflicts so that I can merge this PR?

LeoMcA:
@techAPJ 👍 conflicts resolved

techAPJ:
Thanks! 👍

PR #5126, opened 2017-09-01
Match capitalization of Groups to other stats on user > about box

minusfive:


eviltrout:
Thanks, minor but good catch

PR #5133, opened 2017-09-06
bbcode find close tag loop end condition

guoyunhe:
Only break loop when close tag has been found. Otherwise, keep searching until the end of string.


techAPJ:
Thanks! :+1:

PR #5103, opened 2017-08-26
Do not load javascripts for disabled plugins

davidtaylorhq:

PR #5134, opened 2017-09-06
Fix plugin JS for login_required

davidtaylorhq:
My PR #5103 inadvertently broke the javascript loading for the login page on login_required sites. This is a fix for that problem. @ZogStriP


ZogStriP:
Thanks 👍

ZogStriP:
Would be awesome to have a test ;)

PR #5137, opened 2017-09-06
Add discourse-presence as a core plugin

davidtaylorhq:
Following discussion here: https://meta.discourse.org/t/discourse-presence-who-is-writing-a-reply/48420/27?u=david_taylor


davidtaylorhq:
I don't think the travis failure is related - tests are failing on master as well

ZogStriP:
Let's try it!

PR #5139, opened 2017-09-07
Revert plugin js changes

davidtaylorhq:
@ZogStriP looks like my PR #5103 has [broken docker_manager](https://meta.discourse.org/t/admin-upgrade-page-not-showing-up/69535)

When investigating why this didn't get picked up by tests, I realise my PR also broke the QUnit test loader, so zero qunit tests are running. (but since none are running, none are failing, so travis was still happy).

I think the best thing for now is to revert, and then revisit how to implement this in a way that doesn't break everything else.

Sorry!


ZogStriP:
Yeah, let's revert it.

PR #5151, opened 2017-09-08
Nabble multiple lists

pfaffman:
Apparently whoever wrote this wrote it for a Nabble instance with just one mailing list.

This now creates a category for each list and puts the data in them as one might expect.


pfaffman:
Oh. Lordy. This has all my commits. I'll delete this and try again.

PR #5154, opened 2017-09-08
Discourse-presence improvements

davidtaylorhq:
Some improvements to discourse-presence following @eviltrout's review [here](https://meta.discourse.org/t/discourse-presence-who-is-writing-a-reply/48420/38?u=david_taylor)

Changes made:
- Added guardian checks to endpoint
- Added security to messagebus publishing
- Added specs for security measures
- Moved all logic into component

- Enable plugin by default


eviltrout:
Thanks!

PR #5155, opened 2017-09-08
Only publish PM reply messagebus notifications to allowed users

davidtaylorhq:


eviltrout:
Thank you!

tgxworld:
Let us add a test for this? :+1:

PR #5170, opened 2017-09-11
FEATURE: Remove broken onebox thumbnail images completely

vinothkannans:
If onebox's thumbnail image is broken or larger then it will remove it completely from the onebox.

### DIFF
--- a/app/jobs/regular/pull_hotlinked_images.rb
+++ b/app/jobs/regular/pull_hotlinked_images.rb
@@ -107,6 +107,11 @@ module Jobs
         images = doc.css("img[src]") - doc.css("img.avatar")
         images.each do |tag|
           src = tag['src']
+          if tag['class'] == 'thumbnail' && tag.parent['class'] == 'onebox-body' &&
+            (broken_images.include?(src) || large_images.include?(src))
+            tag.remove
+            next
+          end
           if broken_images.include?(src)
             tag.name = 'span'
             tag.set_attribute('class', 'broken-image fa fa-chain-broken')
--- a/spec/jobs/pull_hotlinked_images_spec.rb
+++ b/spec/jobs/pull_hotlinked_images_spec.rb
@@ -94,6 +94,46 @@ describe Jobs::PullHotlinkedImages do
         expect(post.cooked).to match(/<img src=.*\/uploads/)
       end
 
+      it 'removes broken image' do
+        stub_request(:head, image_url).to_return(status: 404)
+        stub_request(:get, large_image_url).to_return(body: png, headers: { "Content-Type" => "image/png" })
+
+        post = Fabricate(:post, raw: "
+<img src='#{large_image_url}'>
+#{url}
+        ")
+
+        Jobs::ProcessPost.new.execute(post_id: post.id)
+        Jobs::PullHotlinkedImages.new.execute(post_id: post.id)
+        Jobs::ProcessPost.new.execute(post_id: post.id)
+        Jobs::PullHotlinkedImages.new.execute(post_id: post.id)
+
+        post.reload
+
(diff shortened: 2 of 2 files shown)


lib/cooked_post_processor.rb
```diff
@@ -361,8 +361,6 @@ def pull_hotlinked_images(bypass_bump = false)
     return if disable_if_low_on_disk_space
     # don't download remote images for posts that are more than n days old
     return unless @post.created_at > (Date.today - SiteSetting.download_remote_images_max_days_old)
-    # we only want to run the job whenever it's changed by a user
-    return if @post.last_editor_id == Discourse.system_user.id
```
ZogStriP: Why was this removed?

app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -107,6 +107,11 @@ def execute(args)
         images = doc.css("img[src]") - doc.css("img.avatar")
         images.each do |tag|
           src = tag['src']
+          if tag['class'] == 'thumbnail' && tag.parent['class'] == 'onebox-body' &&
```
ZogStriP: Wouldn't it be simpler to do this instead?

```
images = doc.css("img[src]") - doc.css("img.avatar") - doc.css(".onebox-body img.thumbnail")
```

lib/cooked_post_processor.rb
vinothkannans: After the system_user edit for `downloaded local copies of images` again we need to call `pull_hotlinked_images` to replace broken & large images with placeholder.

In 1st step downloaded local images will replaced in `raw` content. Here post will be updated automatically by system user. Then `CookedPostProcess` will be called and downloaded onebox thumbnail will replaced with local copy. Now we have to call `pull_hotlinked_images` to check broken and large image (we can't do it in 1st step since there raw post only changed. Now we changing cooked content). Now last edit is made by system user. That's why I removed that.

1. `pull_hotlinked_images`
2. `CookedPostProcess`
3. `pull_hotlinked_images`

app/jobs/regular/pull_hotlinked_images.rb
vinothkannans: I don't want to skip those onebox images. I am finding it and removing from cooked html using Nokogiri.

lib/cooked_post_processor.rb
vinothkannans: @ZogStriP I just removed it from this PR to be it merged. I can add the above code in another PR. But it gives spec error since [this code](https://github.com/discourse/discourse/blob/master/spec/components/cooked_post_processor_spec.rb#L550). Any help?

vinothkannans:
This PR is now mergable :+1:

coding-horror:
Can you review @gschlager?

gschlager:
Looks good to me. Could you please add a test for this? 😉

vinothkannans:
@gschlager RSpec test function added :+1:

gschlager:
Thanks, but I don't think this spec is testing the right thing. It still works even without your changes in `pull_hotlinked_images.rb`. And it tests only for broken images, not for large images.

Could you take another look please?

vinothkannans:
@gschlager I fixed it. And added new spec function for large image also. It looks like travis failed for some other reason.

SamSaffron:
we got to hold off here a bit for @ZogStriP my concern here is the vast majority of the issues we are seeing in the wild is that finaldestination is not following an image redirect correctly.  lets figure that out first

SamSaffron:
Yeah I am closing this for now, we have some serious issues downloading totally valid images that need to be sorted first @gschlager  / @ZogStriP

PR #5172, opened 2017-09-11
FIX: memoization in PostAnalyzer.raw_mentions

balshor:
Due to a missing 's' when assigning the computed value to @raw_mentions, the raw_mentions function in PostAnalyzer would never retrieve a memoized value.

Meta topic: https://meta.discourse.org/t/broken-memoization-of-mentions-in-posts/69671


ZogStriP:
Thanks 👍

PR #5176, opened 2017-09-12
FIX: don't activate un-confirmed email on omniauth authentication

LeoMcA:
Description of the problem here: https://meta.discourse.org/t/email-change-despite-not-confirming-it/69891?u=leomca

/cc @ZogStriP


ZogStriP:
Looks good :+1:

PR #5177, opened 2017-09-12
FIX: Handle incoming emails without email address in From header

gschlager:
Emails can contain more than one "From" address. So, `mail.from` is either a `String` (one address) or an `Array` (zero or more than one address). The parse function now tries to use the first valid address from the array.

It now also handles the case where there is no "From" header at all or if the "From" header contains no email address. It's throwing a `NoSenderDetectedError` in both cases.


ZogStriP:
Looks good 👍

PR #5178, opened 2017-09-13
FIX: add missing Polish translation

heinzek:
This commit add translate to
[pl_PL.post_links.title.many]
[pl_PL.topic.filter_to.many]


gschlager:
Thanks for the PR but we only accept translations updates through [Transifex](https://www.transifex.com/discourse/discourse-org).

PR #5195, opened 2017-09-19
100.years.ago -> Time.at(0)

henrik:
Avoids a confusing bug in 100 years time ;) Perhaps unlikely, but let's be optimistic.


henrik:
Signed the CLA.

PR #5196, opened 2017-09-19
pass model to user-preferences-account plugin outlet

LeoMcA:
as per: https://meta.discourse.org/t/important-changes-to-plugin-outlets-for-ember-2-10/54136


eviltrout:
thank you!

PR #5197, opened 2017-09-20
Add spec to cover "Badge Type has many Badges" relation

lidimayra:
Relation was not being covered in specs. This PR adds this expectation.

Also: fix filename.
`_spec` suffix was missing. By default, RSpec ignores files in this situation.
Ref: https://github.com/rspec/rspec-core/issues/642


lidimayra:
My pleasure! :smile_cat:

PR #5199, opened 2017-09-21
TEST ONLY: PollFeed integration test

xrav3nz:
> CONTEXT: https://meta.discourse.org/t/simple-rss-is-buggy-and-should-be-forked-or-replaced/27529/23

This is _almost_ a spec-only PR that introduces integration tests for PollFeed (except for a tiny bug fix discovered by the new tests). There will be a follow-up PR to actually replace the SimpleRSS gem with the built-in Ruby RSS module.

The new tests focus on parsing and extracting the correct information out of the RSS/XML feed. Open to suggestions, and let me know if you think more cases should be covered (i.e. all logic in `PollFeed::Feed`).

### DIFF
--- a/app/jobs/scheduled/poll_feed.rb
+++ b/app/jobs/scheduled/poll_feed.rb
@@ -101,7 +101,9 @@ module Jobs
       end
 
       def content
-        @article_rss_item.content.try(:force_encoding, "UTF-8").try(:scrub) || @article_rss_item.description.try(:force_encoding, "UTF-8").try(:scrub)
+        @article_rss_item.content_encoded&.force_encoding("UTF-8")&.scrub ||
+          @article_rss_item.content&.force_encoding("UTF-8")&.scrub ||
+          @article_rss_item.description&.force_encoding("UTF-8")&.scrub
       end
 
       def title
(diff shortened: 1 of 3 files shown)


app/jobs/scheduled/poll_feed.rb
```diff
@@ -101,7 +101,9 @@ def url
       end
 
       def content
-        @article_rss_item.content.try(:force_encoding, "UTF-8").try(:scrub) || @article_rss_item.description.try(:force_encoding, "UTF-8").try(:scrub)
+        @article_rss_item.content_encoded&.force_encoding("UTF-8")&.scrub ||
+          @article_rss_item.content&.force_encoding("UTF-8")&.scrub ||
+          @article_rss_item.description&.force_encoding("UTF-8")&.scrub
```
xrav3nz: I haven't found much references for the `content` tag. Most RSS feeds that I've been fiddling with seems to be using the [`content:encoded` tag](http://www.rssboard.org/rss-profile#namespace-elements-content-encoded), namely the [Dota2 Blog Feed](http://blog.dota2.com/feed/), [Discourse Blog Feed](https://blog.discourse.org/feed), etc.

Do you think it'll be safe to remove the parsing for the `content` tag, and let whoever using it to adapt to our change? (Should be a minority)

tgxworld:
LGTM. Thank you for contributing to Discourse :+1:

PR #5200, opened 2017-09-22
FIX: do not escape already escaped chars in URL

gschlager:
Fix for `Job exception: undefined method `author' for nil:NilClass`
https://meta.discourse.org/logs/show/0e87a429023632664db4a78684170ac2

Escaping already escaped chars in URL produced invalid URL


tgxworld:
:+1:

PR #5201, opened 2017-09-22
FEATURE: regex matching for `rake posts:remap` and `rake posts:delete_word`

xrav3nz:
> CONTEXT & DISCUSSION: https://meta.discourse.org/t/replace-a-string-in-all-posts/48729/45?u=xrav3nz

This PR adds regex pattern matching support for `rake posts:remap` and `rake posts:delete_word`.

```ruby
> rake posts:remap['pattern', 'replace', type]
> rake posts:delete_word['pattern', type]
```

where `type` defaults to `string` if omitted.

Cheers 💪 

----

EDIT: The failed test seems to be flaky and is not relevant to this change.

### DIFF
--- a/lib/tasks/posts.rake
+++ b/lib/tasks/posts.rake
@@ -42,23 +42,18 @@ end
 
 desc 'Rebake all posts matching string/regex and optionally delay the loop'
 task 'posts:rebake_match', [:pattern, :type, :delay] => [:environment] do |_, args|
+  args.with_defaults(type: 'string')
   pattern = args[:pattern]
-  type = args[:type]
-  type = type.downcase if type
-  delay = args[:delay].to_i if args[:delay]
+  type = args[:type]&.downcase
+  delay = args[:delay]&.to_i
+
   if !pattern
     puts "ERROR: Expecting rake posts:rebake_match[pattern,type,delay]"
     exit 1
   elsif delay && delay < 1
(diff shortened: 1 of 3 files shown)


lib/tasks/posts.rake
```diff
@@ -42,31 +42,26 @@ end
...
-  else
+  elsif type != 'string' && type != 'regex'
     puts "ERROR: Expecting rake posts:rebake_match[pattern,type] where type is string or regex"
     exit 1
   end
 
   rebaked = 0
   total = search.count
 
-  search.find_each do |post|
```
SamSaffron: Hmmm we stopped using find_each here so every single post gets loaded into memory, this is not really going to work on large databases.

xrav3nz: Sorry I'm a bit lost here; this _is_ using `find_each` so not all records are loaded at once. Are you referring to the other tasks that use `each`?

SamSaffron: see: https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each find_each performs the work in batchs so memory usage is low.

SamSaffron: ahh did not notice you did a find each after the raw_match, my bad

SamSaffron:
It looks safe, it is tested, I say we merge.  thanks!

PR #5202, opened 2017-09-24
Use `PostDestroyer` to delete topics when performing bulk operation

OsamaSayegh:
This PR fixes this [bug](https://meta.discourse.org/t/bulk-deletion-of-topics-not-logged/50407?u=osama). Also it fixes a related problem (quite an edge case) where staff users couldn't undelete topics that were bulk deleted (for some reasons undelete would just fail silently).

Let me know what you think :)

### DIFF
--- a/lib/topics_bulk_action.rb
+++ b/lib/topics_bulk_action.rb
@@ -135,7 +135,9 @@ class TopicsBulkAction
 
     def delete
       topics.each do |t|
-        t.trash! if guardian.can_delete?(t)
+        if guardian.can_delete?(t)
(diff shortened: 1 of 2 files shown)


SamSaffron:
I like this change but we need a test here ...

OsamaSayegh:
Hi @SamSaffron 

There is already a test for bulk delete here:

https://github.com/discourse/discourse/blob/master/spec/components/topics_bulk_action_spec.rb#L68-L78

The test was failing due to my changes so I've updated my PR and now the test passes.

Thoughts?

ZogStriP:
Looks fine. Thanks 👍

PR #5203, opened 2017-09-25
adding a button at a new plugin outlet location in backup#index

kajatiger:
for our plugin **discourse-sync-to-googledrive** we needed to add a button for the downloader


eviltrout:
Thank you!

PR #5204, opened 2017-09-25
FEATURE: Use new WebShare API on supported platforms

xfalcox:
https://meta.discourse.org/t/introducing-the-web-share-api-web-google-developers/70515?u=falco

### DIFF
--- a/app/assets/javascripts/discourse/components/share-popup.js.es6
+++ b/app/assets/javascripts/discourse/components/share-popup.js.es6
@@ -81,6 +81,14 @@ export default Ember.Component.extend({
     Ember.run.scheduleOnce('afterRender', this, this._focusUrl);
   },
 
+  _webShare(url) {
+    // We can pass title and text too, but most share targets do their own oneboxing
+    navigator.share({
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/components/share-popup.js.es6
```diff
@@ -81,6 +81,14 @@ export default Ember.Component.extend({
     Ember.run.scheduleOnce('afterRender', this, this._focusUrl);
   },
 
+  _webShare(url) {
+    // We can pass title and text too, but most share targets do their own oneboxing
+    navigator.share({
+      url: url
+    })
+    .catch((error) => console.warn('Error sharing', error));
```
gschlager: Should we use `_showUrl()` as fallback in case of an error?

xfalcox: I believe the current implementation also throws an error on user dismiss in the native picker, so that would be bad.

If we can somehow differentiate between real errors and user dismiss, this is a good idea!

PR #5207, opened 2017-09-27
Fix randomly failing test.

tgxworld:


spec/components/active_record/connection_adapters/postgresql_fallback_adapter_spec.rb
```diff
@@ -95,13 +95,18 @@
         expect(ActiveRecord::Base.connection_pool.connections.count).to eq(0)
 
         ActiveRecord::Base.clear_all_connections!
+
+        ActiveRecord::Base.connection_pool.connections.each do |c|
+          puts c
```
SamSaffron: I believe I can puts

PR #5208, opened 2017-09-27
FIX: async reload of locales could result in missing translations

gschlager:
Reloading of locales via MessageBus could result in missing translations when `I18n.reload!` was called by `MessageBus` while `I18n.t()` was used.

image

I think this can be removed because
* I couldn't find any other reference to `i18n-flush`
* `I18n.reload!` is executed by the same method that published `i18n-flush` which resulted in duplicate reloads

### DIFF
--- a/app/models/translation_override.rb
+++ b/app/models/translation_override.rb
@@ -17,21 +17,16 @@ class TranslationOverride < ActiveRecord::Base
 
     translation_override = find_or_initialize_by(params)
     params.merge!(data) if translation_override.new_record?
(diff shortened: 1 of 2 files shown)


tgxworld:
> I18n.reload! is executed by the same method that published i18n-flush which resulted in duplicate reloads

I think the subscription is to sync up all the locales among all the unicorn workers.

gschlager:
@tgxworld Do you have any idea how I can test if this is really the case?
Anyway, I'm closing this for now and will try to find another solution.

PR #5209, opened 2017-09-28
FIX: update group.has_messages field weekly

techAPJ:
Fixes this issue: https://meta.discourse.org/t/messaging-group-inboxes-and-archives-show-even-after-all-messages-deleted/48467

@SamSaffron: `has_messages` field was added to `groups` table in [this commit](https://github.com/discourse/discourse/commit/a8b5192efd43bcbf64ffe7ae44e5c5b7f3be0b51#diff-34c506df3814d6196b4857679170c2b4), it is supposed to keep track of if there are any messages in groups. It gets updated to `true` when a topic is created but does not reverts to `false` when the topic is deleted.

This PR adds a weekly check to make sure that `has_messages` field is up-to-date for all the groups.

Review please? :)

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -308,6 +308,7 @@ class Group < ActiveRecord::Base
   def self.ensure_consistency!
     reset_all_counters!
     refresh_automatic_groups!
+    refresh_has_messages!
   end
 
   def self.reset_all_counters!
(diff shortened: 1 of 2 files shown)


app/models/group.rb
```diff
@@ -331,6 +332,17 @@ def self.refresh_automatic_groups!(*args)
     args.each { |group| refresh_automatic_group!(group) }
   end
 
+  def self.refresh_has_messages!
+    exec_sql <<-SQL
+      UPDATE groups g SET has_messages = false
+      WHERE NOT EXISTS (SELECT tag.id
```
ZogStriP: I would also add a `WHERE has_messages = true` so we don't update the rows for no reason ;)

techAPJ: Good idea! Made it so.

PR #5210, opened 2017-09-28
Serialise 'nil' custom fields as empty dictionary, not null

davidtaylorhq:
At the moment, if a user has no custom fields, it gets serialized as a JSON "null". This can cause a lot of weird errors when trying to get/set them in javascript.

A number of plugins have been working around this by overriding the serialiser method with something like this:
```
add_to_serializer(:user, :custom_fields, false) {
    if object.custom_fields == nil then
      {}
    else
      object.custom_fields
    end
  }
```

E.g.:
https://github.com/xfalcox/discourse-signatures/blob/master/plugin.rb#L28-L36
https://github.com/Ebsy/discourse-nationalflags/blob/master/plugin.rb#L31-L38
https://github.com/davidtaylorhq/discourse-telegram-notifications/blob/master/plugin.rb#L184-L190

### DIFF
--- a/app/serializers/current_user_serializer.rb
+++ b/app/serializers/current_user_serializer.rb
@@ -135,7 +135,7 @@ class CurrentUserSerializer < BasicUserSerializer
     end
 
     if fields.present?
-      User.custom_fields_for_ids([object.id], fields)[object.id]
+      User.custom_fields_for_ids([object.id], fields)[object.id] || {}
     else
       {}
     end
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -387,7 +387,7 @@ class UserSerializer < BasicUserSerializer
     end
 
     if fields.present?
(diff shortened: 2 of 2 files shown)


SamSaffron:
Sure this seems fine to me but for this kind of stuff I would expect a spec

SamSaffron:
howdy @davidtaylorhq  I think my comment got lost somewhere, any way you can write a spec for this. I want the behavior to be well specified.

davidtaylorhq:
Yep, will do 👍

SamSaffron:
a test pretty please ;)

davidtaylorhq:
@SamSaffron sorry this has been open for so long. 

I have written some tests, but need to fix the massive performance regression I accidentally introduced in the process (The user serializer spec increased from taking 0.3s to 3s). 

Will open another PR once I've fixed that

SamSaffron:
Don't worry about it, I added to a test that covers this area so we are good!

PR #5213, opened 2017-09-29
FIX: response error when deleting non-existent tags

ecdemis123:
This PR is to address the [issue](https://meta.discourse.org/t/api-inconsistent-error-message-when-deleting-a-tag/70729) that arises when trying to delete a non-existent tag.  Updates the API to return 'not_found' instead of a general Rails error. 

First time contributing to this repo, so please let me know if there are any processes that I may have overlooked when creating this PR. Cheers!


ZogStriP:
Awesome! Thanks 👍

ecdemis123:
:clinking_glasses:

PR #5215, opened 2017-10-01
Add specs for EmailTemplatesController

gschlager:
@tgxworld Here are the tests for https://github.com/discourse/discourse/commit/1a378126259b1461bb935d1385778842d33b088a
(And yes, I found and corrected a bug while writing those tests 😉 )

First time using `shared_examples`... I hope I'm using them correctly.

### DIFF
--- a/app/controllers/admin/email_templates_controller.rb
+++ b/app/controllers/admin/email_templates_controller.rb
@@ -47,11 +47,9 @@ class Admin::EmailTemplatesController < Admin::AdminController
 
     error_messages = []
     if subject_result[:error_messages].present?
-      TranslationOverride.upsert!(I18n.locale, "#{key}.subject_template", subject_result[:old_value])
       error_messages << format_error_message(subject_result, "subject")
     end
     if body_result[:error_messages].present?
-      TranslationOverride.upsert!(I18n.locale, "#{key}.text_body_template", body_result[:old_value])
       error_messages << format_error_message(body_result, "body")
     end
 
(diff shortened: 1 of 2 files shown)


tgxworld:
@gschlager Thank you! Could we write the controller specs as request specs instead? Controller specs are no longer recommended by the Rails/Rspec team in Rails 5 and beyond.

spec/controllers/admin/email_templates_controller_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    expect(Admin::EmailTemplatesController < Admin::AdminController).to eq(true)
+  end
+
+  context "while logged in as an admin" do
+    before do
+      @user = log_in(:admin)
+    end
+
+    after do
+      UserHistory.delete_all
```
tgxworld: Hmm do we need this here? The transaction should be rollbacked after each test.

spec/controllers/admin/email_templates_controller_spec.rb
```diff
@@ -0,0 +1,215 @@
+require 'rails_helper'
+
+describe Admin::EmailTemplatesController do
+
+  it "is a subclass of AdminController" do
```
tgxworld: I know we use this in a bunch of places in our code base but I'm personally not a fan of it. I would rather we write a simple spec for an anon user to verify that the access is invalid.

spec/controllers/admin/email_templates_controller_spec.rb
gschlager: Not really. It just was the easiest way to solve a particular problem in a test. 😉 I'm going to change it.

gschlager:
@tgxworld I'm now using Request specs. I had to set an accept header for PUT and DELETE since `.json` in the URL doesn't work in this case. I also added tests for (anonymous) users instead of using the `is a subclass of AdminController` spec.

tgxworld:
@gschlager Looks good to me :+1:

PR #5219, opened 2017-10-03
Restrict creation of staged users for incoming emails

gschlager:
* checks the email domain whitelist or blacklist before creating staged users
* doesn't send subscription emails to new users
* doesn't create staged users when incoming email is rejected

(Well, it still creates staged users in some cases. I have a solution for that as well, but the specs aren't ready yet.)

@ZogStriP Sorry for the large PR. I can split it up if you want.


ZogStriP:
Awesome. LGTM 👍

schungx:
Awesome x2

PR #5225, opened 2017-10-05
UX: discourse restore -- sort by date

pfaffman:
Now, if you do `./script/discourse restore` with no filename to restore, it sorts the suggested files in reverse chron, so the top file will most likely be the one you want.


tgxworld:
Thanks!

PR #5227, opened 2017-10-06
UX: Improve Delete Tag Confirm Message

majakomel:
Topic: https://meta.discourse.org/t/delete-tag-warning-improvement/60495

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2457,7 +2457,7 @@ en:
       delete_confirm:
         one: "Are you sure you want to delete this tag and remove it from 1 topic it is assigned to?"
         other: "Are you sure you want to delete this tag and remove it from {{count}} topics it is assigned to?"
-        no_topics: "Are you sure you want to delete this tag?"
+      delete_confirm_no_topics: "Are you sure you want to delete this tag?"
       rename_tag: "Rename Tag"
       rename_instructions: "Choose a new name for the tag:"
       sort_by: "Sort by:"
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2454,7 +2454,10 @@ en:
       tags: "Tags"
       choose_for_topic: "choose optional tags for this topic"
       delete_tag: "Delete Tag"
-      delete_confirm: "Are you sure you want to delete that tag?"
+      delete_confirm:
+        one: "Are you sure you want to delete this tag and remove it from 1 topic it is assigned to?"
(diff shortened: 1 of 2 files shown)


coding-horror:
Nice improvement 👍

config/locales/client.en.yml
```diff
@@ -2454,7 +2454,10 @@ en:
       tags: "Tags"
       choose_for_topic: "choose optional tags for this topic"
       delete_tag: "Delete Tag"
-      delete_confirm: "Are you sure you want to delete that tag?"
+      delete_confirm:
+        one: "Are you sure you want to delete this tag and remove it from 1 topic it is assigned to?"
+        other: "Are you sure you want to delete this tag and remove it from {{count}} topics it is assigned to?"
+        zero: "Are you sure you want to delete this tag?"
```
gschlager: It's not possible to use the `zero` key in the English locale. I suggest you use something like `delete_confirm_no_topics` instead.

majakomel:
@gschlager is this better solution?
`zero` key seemed to work fine with me before though, why is it better to avoid it in case of English locale? (Sorry if obvious, I'm more of a newbie here)

gschlager:
Yeah, the i18n gem has a special case for the `zero` key, but English uses only `one` and `other` for pluralizations. See http://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html#en if you want to know more about this. ;-)

We use Transifex for translations and it also supports only `one` and `other` as keys when the source language is English. So, yes, it works with the i18n gem, but our translators won't be able to translate the `zero` key.

config/locales/client.en.yml
```diff
@@ -2454,7 +2454,10 @@ en:
       tags: "Tags"
       choose_for_topic: "choose optional tags for this topic"
       delete_tag: "Delete Tag"
-      delete_confirm: "Are you sure you want to delete that tag?"
+      delete_confirm:
+        one: "Are you sure you want to delete this tag and remove it from 1 topic it is assigned to?"
+        other: "Are you sure you want to delete this tag and remove it from {{count}} topics it is assigned to?"
+        no_topics: "Are you sure you want to delete this tag?"
```
gschlager: Could you please move `no_topics` to `delete_confirm_no_topics`? It shouldn't be a sub-key of `delete_confirm`.
Our [tests are failing on Travis](https://travis-ci.org/discourse/discourse/jobs/284214941) because we check that only `zero` and `other` keys exist. Other than that it looks fine!

PR #5228, opened 2017-10-07
UX: Cap likes under posts

OsamaSayegh:
Meta topic: https://meta.discourse.org/t/the-most-liked-topic-of-any-discourse/59649/7?u=osama

Screenshot:

capped_likes

Not sure what number to limit the likes at. Currently it's 25. Is that too high or too low? Should the number be configurable in site settings?


SamSaffron:
Replied here: 

https://meta.discourse.org/t/the-most-liked-topic-of-any-discourse/59649/10?u=sam

Let's discuss there before implementing anything.

PR #5230, opened 2017-10-09
Avoid auto-unpinning a user's watched topic

chrisbeach:
Discussion: https://meta.discourse.org/t/don-t-auto-unpin-topics-that-a-user-is-watching/71664


SamSaffron:
I am sorry, I just don't agree with this change, I don't see it as related.

PR #5231, opened 2017-10-10
FIX: delete staged users when the incoming email is rejected

gschlager:


gschlager:
@ZogStriP Could you take a quick look if deleting staged users is fine the way I'm doing it?

lib/email/receiver.rb
```diff
@@ -752,6 +753,16 @@ def send_subscription_mail(action, user)
       message = SubscriptionMailer.send(action, user)
       Email::Sender.new(message, :subscription).send
     end
+
+    def delete_staged_users
+      @staged_users.each do |user|
+        begin
+          UserDestroyer.new(Discourse.system_user).destroy(user, quiet: true)
+        rescue UserDestroyer::PostsExistError
```
tgxworld: Instead of rescuing the error which is slow, maybe we can just add a filter in the `@staged_users` query to only return users without any posts?

spec/services/user_destroyer_spec.rb
```diff
@@ -45,6 +45,11 @@
         StaffActionLogger.any_instance.expects(:log_user_deletion).with(@user, anything).once
         destroy
       end
+
+      it "shouldn't log the action if quiet is true" do
+        StaffActionLogger.any_instance.expects(:log_user_deletion).with(@user, anything).never
```
tgxworld: I see the surrounding code using stubs but we're trying to move away from it :grin: 

We can just do something like 

```
expect {}.to_not change { UserHistory.count } 
```

gschlager:
@tgxworld Thanks for the feedback! I'll try to copy the right spec next time 😉

PR #5233, opened 2017-10-11
move mailgun to top--elastic email seems to confuse people

pfaffman:
There has been a spate of people having trouble with Elastic Email. I was wondering why so many people were using it, but see now that it's not only on the list, but it's at the top.

Mailgun seems to be the easiest one to set up, so I thought moving it back to the top of the list might encourage those looking for the easiest solution to try Mailgun first.


pfaffman:
@coding-horror , do you want this?

PR #5234, opened 2017-10-12
Discobot Localisation: Fix missing backtick

flesser:
Discobot's German localisation had a minor formatting error when printing it's capabilities because of a missing backtick.


flesser:
@discoursebot No, thanks. I'm neither reading nor signing such a wall of text just to report such a minor issue! 😞

techAPJ:
@flesser Thanks for the PR. But the translation changes needs to be made in Transifex: https://www.transifex.com/projects/p/discourse-org/ for them to persist,

PR #5235, opened 2017-10-12
removing scale-color, first pass

awesomerobot:
This removes the majority of scale-color instances in our sass files and replaces them with color variables defined in `stylesheets/common/foundation/variables.scss`

This is part of the ongoing process that will allow us to eliminate the overlapping systems for color functions/definitions (dark-light-diff, scale-color, dark-light-choose, etc)

See: https://meta.discourse.org/t/cleaning-up-our-color-palette/61263

I've tested this locally with a few different color schemes and on mobile, but it also needs a closer look on an active site with more data

### DIFF
--- a/app/assets/stylesheets/common/admin/admin_base.scss
+++ b/app/assets/stylesheets/common/admin/admin_base.scss
@@ -512,7 +512,7 @@ $mobile-breakpoint: 700px;
     }
 
     .desc {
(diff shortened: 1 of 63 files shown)


coding-horror:
@eviltrout you should review this

eviltrout:
This is super hard to eyeball for me, so maybe we should merge after a new release is cut and deal with any regressions in colours.

coding-horror:
OK, then let's make a new version early tomorrow?

PR #5237, opened 2017-10-14
UX: cap likes 2

OsamaSayegh:
Meta: https://meta.discourse.org/t/the-most-liked-topic-of-any-discourse/59649/10?u=osama


SamSaffron:
OK, we can try this out, I don't think it adds any cost. 

Thanks!

PR #5239, opened 2017-10-16
add importer for modx forum

pfaffman:
Importer for Modx forum.

Tested only with a sample database. I never got a live database from the client, but it works fine the the sample database I had for testing.


pfaffman:
@ZogStriP , I think this is in your wheelhouse and git's not recommending anyone.

SamSaffron:
brand new file 🎊  I say we just merge it in for now! thanks!

PR #5240, opened 2017-10-17
postprocessing ignored cook method

gschlager:
Postprocessing of posts ignored the `cook_method`.
When a post had a `cook_method` of `email` it looked fine at first, but it didn't after Sidekiq ran the `ProcessPost` job.

image

This PR moves the logic into `PostAnalyzer`. The example from above looks now correct.

image

There are two more changes in this PR:
* `EmailCook` doesn't use the oneboxer anymore. It was slowing down mbox imports and is now handled by the `ProcessPost` job.
* I replaced the `def cook(*args)` with `def cook(raw, opts = {})` because I find the splat extremely confusing and hard to read.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -5,7 +5,6 @@ require_dependency 'enum'
 require_dependency 'post_analyzer'
 require_dependency 'validators/post_validator'
 require_dependency 'plugin/filter'
-require_dependency 'email_cook'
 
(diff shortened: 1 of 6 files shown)


gschlager:
Hey @tgxworld! Could you please take a look at this? The specs should be ok this time 😉

PR #5258, opened 2017-10-23
Use the CategoryHashtag::SEPARATOR

pmatsinopoulos:
This is a small enhancement on an existing test. Instead of hard coding the `:` symbol, use the `CategoryHashtag::SEPARATOR` constant, like we do in the previous examples in the same file.


ZogStriP:
Thanks 👍

PR #5259, opened 2017-10-23
Add groups to the SSO provider payload

xfalcox:


tgxworld:
tests? :grin:

xfalcox:
@tgxworld can you take a look at this today? This change is needed so I can land this PR:

https://github.com/discourse/discourse-auth-proxy/pull/6

app/controllers/session_controller.rb
```diff
@@ -51,6 +51,7 @@ def sso_provider(payload = nil)
         sso.external_id = current_user.id.to_s
         sso.admin = current_user.admin?
         sso.moderator = current_user.moderator?
+        sso.groups = current_user.groups.map(&:name)
```
tgxworld: I think it'll be better to just `pluck` the names?

PR #5268, opened 2017-10-26
FEATURE: Allow the user to select a custom home page

notriddle:
https://meta.discourse.org/t/remember-the-selected-view-tab-on-homepage/16982/19

### DIFF
--- a/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
@@ -14,6 +14,8 @@ export default Ember.Controller.extend(PreferencesTabController, {
       'enable_quoting',
       'disable_jump_reply',
       'automatically_unpin_topics',
+      'allow_private_messages',
+      'user_home',
     ];
 
     if (makeDefault) {
@@ -50,6 +52,14 @@ export default Ember.Controller.extend(PreferencesTabController, {
     let key = this.get("themeKey");
     previewTheme(key);
   },
+
+  userSelectableHome: [
+    { name: I18n.t('filters.latest.title'), value: 1 },
+    { name: I18n.t('filters.categories.title'), value: 2 },
+    { name: I18n.t('filters.unread.title'), value: 3 },
+    { name: I18n.t('filters.new.title'), value: 4 },
+    { name: I18n.t('filters.top.title'), value: 5 },
+  ],
 
   actions: {
     save() {
--- a/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
@@ -4,6 +4,8 @@ import { default as computed, observes } from "ember-addons/ember-computed-decor
 import { currentThemeKey, listThemes, previewTheme, setLocalTheme } from 'discourse/lib/theme-selector';
 import { popupAjaxError } from 'discourse/lib/ajax-error';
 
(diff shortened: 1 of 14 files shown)


notriddle:
I have signed the CLA.

notriddle:
Rebased.

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
```diff
@@ -50,6 +53,26 @@ export default Ember.Controller.extend(PreferencesTabController, {
     let key = this.get("themeKey");
     previewTheme(key);
   },
+  
+  homeChanged() {
+    let home = Discourse.SiteSettings.top_menu.split("|")[0].split(",")[0];
+    switch (Number(this.get('model.user_option.user_home'))) {
```
ZogStriP: Why not use a hash instead of a giant switch/case?

```
const USER_HOMES = { 1: "latest", 2: "categories", ..., 5: "top" };
homeChanged() {
  const home = USER_HOMES[this.get('model.user_option.user_home')] || Discourse.SiteSettings.top_menu.split("|")[0].split(",")[0];
  setDefaultHomepage(home);
}
`

Comment (notriddle) [reply app/assets/javascripts/discourse/controllers/preferences/interface.js.es6]: Good call.

File: app/controllers/application_controller.rb [hunk 2]
```diff
@@ -308,7 +308,7 @@ def guardian
   end
 
   def current_homepage
-    current_user ? SiteSetting.homepage : SiteSetting.anonymous_homepage
+    current_user&.user_option&.homepage || SiteSetting.anonymous_homepage
```
SamSaffron: wait, what if homepage is set to nil in user option, it should be homepage then not anonymous_hompage

test with "unread" as the first in the list

app/models/user_option.rb
```diff
@@ -165,6 +176,7 @@ def update_tracked_topics
 #  theme_key                        :string
 #  theme_key_seq                    :integer          default(0), not null
 #  allow_private_messages           :boolean          default(TRUE), not null
+#  user_home                        :integer          default(0), not null
```
SamSaffron: this should not default to 0 it should default to nil, unset by user

SamSaffron:
getting close ❤️

app/controllers/application_controller.rb
notriddle: `homepage` is not the setting. `user_home` is the setting.

`homepage` is a method with [this definition](https://github.com/discourse/discourse/pull/5268/files#diff-162f0f01d64443df28e0ffe2a57be378R131).

app/models/user_option.rb
notriddle: Done.

app/controllers/application_controller.rb
SamSaffron: I see, makes sense

SamSaffron:
OK only last bit is that I am unsure about the name

```
User.first.user_option.user_home
```

Its already a user option so why does it need the prefix? 

I think `homepage` or `homepage_id` work better and are more consistent

I would go with `homepage_id` and then you can keep your hompage helper that returns text.

notriddle:
user_home ➡️ homepage_id

notriddle:
This should fix all of the complaints.

SamSaffron:
Thanks heaps for all your work here @notriddle !!! Merging this in! 🍧

PR #5273, opened 2017-10-28
add alt 'archive.org_bot' user-agent

gojomo:
Add `archive.org_bot' – another user-agent used by Internet Archive when crawling for Wayback Machine.


ZogStriP:
Thanks

PR #5276, opened 2017-10-30
Include information about iframe state in message

c-holtermann:
As discussed in this post: https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963/233?u=christoph_holtermann it would be useful to have information about the loading state of the embedded discourse.


SamSaffron:
I see , but what triggers the loaded event here? seems like you would either always get loaded or never get loaded cc @eviltrout

c-holtermann:
The loaded event is triggered by the span with class replies that contains the number of replies. It  is only present after load has finished. It works for sure but for a good coding style it would be better to let the signal come from somewhere "deeper" in the code. My solution may break if someone decides to rename the class of replies or to not use a span. In addition this doesn't take care of an error state which may also be possible. I didn't check out all possible states. Loading and loaded where those that interested me and those work. But it's not a beautiful solution at the moment, rather a hack. I don't speak ruby so I had to stick to the javascript layer for the moment and the existence of the replies span is at the moment a sure marker at that level.

app/views/layouts/embed.html.erb
```diff
@@ -38,8 +38,12 @@
         }
 
         window.onload = function() {
-          // Send a post message with our loaded height
-          postUp({type: 'discourse-resize', height: document['body'].offsetHeight});
+          var state = "loading";
+          if (document.querySelectorAll("span.replies").length > 0) {
```
eviltrout: I don't love this approach - I think it would be better if the rails app rendered the state into the DOM in data-* attributes somewhere and the javascript read those for the state. Otherwise it's tied to the specific elements in the DOM.

c-holtermann: Is there a part of the DOM that can be read inside and outside of the iframe ? If there is then I'm fine with reading that info. Still sending messages seems nice because otherwise the script from the outside needs to regularly check the DOM. In case of a message it can just sit and wait until a message arrives. But I'm not really deep into messaging in javascript. Maybe there's a possibility to trigger a message on change of a data attribute ?

c-holtermann: @eviltrout what I just wrote was about the communication between javascript inside the iframe and javascript outside of the iframe. Rereading your comment I now assume that I got you wrong. 
There is
I) Rails
II) javascript inside the iframe
--- iframe border ---
III) javascript outside the iframe

I think it would be very nice if I) rails modified the DOM (inside the iframe) setting a clearly defined state flag.
Then the II) javascript inside the iframe could send a message through the iframes border including the state flag.
That message could then be read outside the iframe III).

This pull request is a hack because it doesn't respect the rails level. I very much appreciate a proper approach on the rails level.

eviltrout: What I was suggesting is mostly the same as your PR. Instead of counting elements, just look for a DOM element. Rails generates the contents of the iframe, so it can drop a data-* element in there when it renders the contents. The javascript can look for that element, and `postUp`.

c-holtermann:
I changed the way the information is being transmitted by adding a data-attribute to the header element in the ruby templates. Javascript checks this attribute. @eviltrout you said you wanted a data element. Is this ok or do you rather want an empty element like a div or a <data> element ?

tgxworld:
@eviltrout Does the PR look OK to you?

app/views/layouts/embed.html.erb
```diff
@@ -37,9 +37,16 @@
...
 
-        window.onload = function() {
-          // Send a post message with our loaded height
-          postUp({type: 'discourse-resize', height: document['body'].offsetHeight});
+	window.onload = function() {
+	  // get state info from data attribute
+	  var header = document.querySelector('header');
+	  var state = 'unknown';
+	  if (header) {
+	    var state = header.getAttribute('data-embed-state');
```
eviltrout: You are redeclaring `state` here. No `var` keyword should be present.

eviltrout:
There is still one minor issue that needs to be fixed @tgxworld - I made a comment.

app/views/layouts/embed.html.erb
c-holtermann: ok, I'll fix this.

SamSaffron:
I want this change but too risky to merge in now, @eviltrout  can you consider merging this post 1.9 ?

app/views/layouts/embed.html.erb
```diff
@@ -37,9 +37,16 @@
           }
         }
 
-        window.onload = function() {
-          // Send a post message with our loaded height
-          postUp({type: 'discourse-resize', height: document['body'].offsetHeight});
+	window.onload = function() {
```
tgxworld: The indentation seems off to me here.

Shawndrea:
I will be willing to hell but everytime i do i gt crossed so im just watchin yall work so well together

Shawndrea:
I will be willing to help but everytime i do i gt crossed so im just watchin yall work so well together

app/views/layouts/embed.html.erb
c-holtermann: Yes. I fixed it.

c-holtermann:
Travis fails one build with something about Cakeday. I don't understand. Another build from another pull request failed with the same message so I guess it's got nothing to do with these changes ?

SamSaffron:
Sorry for the mega delay here, merging it in now 🎊

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

c-holtermann:
Thank you for merging !

c-holtermann:
Found an unfortunate typo. See #6043.

PR #5277, opened 2017-10-31
FIX: typo in russian localization

juwain:


gschlager:
Thanks for the PR, but we only accept translations updates through [Transifex](https://www.transifex.com/discourse/discourse-org).

PR #5282, opened 2017-11-01
FIX: Topic links onebox differently if end in /

vinothkannans:
Links to the internal topics are oneboxed differently if there’s a / in the end.


coding-horror:
Does this look ok @SamSaffron

vinothkannans:
It looks like Travis is failing for some other reason `Failure/Error: raise ConnectionTimeoutError, msg`. So it is mergable I guess.

vinothkannans:
I updated the code recently. Have a look again.

techAPJ:
Looks good to me. 👍

PR #5283, opened 2017-11-02
Fix "duplicate method" issue

viktorbenei:
Fixing http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/DuplicateMethods

Readers are defined (https://github.com/discourse/discourse/blob/master/lib/single_sign_on.rb#L61), so only writers have to be generated.


viktorbenei:
@discoursebot CLI signed

tgxworld:
LGTM :+1:  Thank you @viktorbenei

PR #5285, opened 2017-11-02
Adds primary user group as a class to quote

ckeboss:
This feature addition will add the class `group-PRIMARY_USER_GROUP` to
the quote `aside`. `PRIMARY_USER_GROUP` will be the primary user group
of the user being quoted. This is similar to the class that is added to
a `topic-post`.


test/javascripts/lib/pretty-text-test.js.es6
```diff
@@ -270,6 +269,16 @@ QUnit.test("Quotes", assert => {
 </aside>`,
          "handles nested quotes properly");
 
+  assert.cookedOptions(`[quote="bob, post:1, topic:1"]\ntest quote\n[/quote]`, { lookupPrimaryUserGroupByPostNumber: function() { return "aUserGroup" } },
```
eviltrout: I would put your args on a different line, and use ES6 function syntax:

`{ lookupPrimaryUserGroupByPostNumber: () => "aUserGroup" }`

lib/pretty_text/helpers.rb
```diff
@@ -25,6 +25,14 @@ def avatar_template(username)
       UrlHelper.schemaless(UrlHelper.absolute(user.avatar_template))
     end
 
+    def lookup_primary_user_group(username)
+      return "" unless username
+      user = User.find_by(username_lower: username.downcase)
+      return "" unless user.present?
+
+      return user.primary_group.try(:name) || ""
```
eviltrout: You should omit `return` if it's the last line of your function

app/assets/javascripts/pretty-text/engines/discourse-markdown/quotes.js.es6
```diff
@@ -132,4 +145,14 @@ export function setup(helper) {
   });
 
   helper.whiteList(['img[class=avatar]']);
+  helper.whiteList({
+    custom(tag, name, value) {
+      if (tag === 'aside' && name === 'class') {
+        const m = /^quote group\-(.+)$/.exec(value);
```
eviltrout: Could probably just `return /^quote group\-(.+)$/.exec(value);` right? Or perhaps to force a boolean:

`return !!/^quote group\-(.+)$/.exec(value);`

app/assets/javascripts/pretty-text/engines/discourse-markdown/quotes.js.es6
```diff
@@ -34,9 +34,30 @@ const rule = {
...
+      primaryGroupName = options.lookupPrimaryUserGroupByPostNumber(postNumber, topicId);
+    } else if (options.lookupPrimaryUserGroup) {
+      // server-side, we need to lookup the primary user group from the username
+      primaryGroupName = options.lookupPrimaryUserGroup(username);
+    }
+    
+    let token   = state.push('bbcode_open', 'aside', 1);
+    token.attrs = [];
+    
+    if (primaryGroupName) {
```
eviltrout: You are returning empty strings for this function - so that would pass the empty check and allow this to end up as `group-` sometimes. You should check for length.

PR #5291, opened 2017-11-06
UX: Large image placeholder style changed in cooked html

vinothkannans:
Large image placeholder UI changed as inline in cooked html. **And totally removed the large images inside onebox**.

### DIFF
--- a/app/jobs/regular/pull_hotlinked_images.rb
+++ b/app/jobs/regular/pull_hotlinked_images.rb
@@ -8,6 +8,8 @@ module Jobs
 
     sidekiq_options queue: 'low'
 
+    LARGE_IMAGES = "large_images".freeze
+
     def initialize
       @max_size = SiteSetting.max_image_size_kb.kilobytes
     end
@@ -46,7 +48,8 @@ module Jobs
       raw = post.raw.dup
       start_raw = raw.dup
       downloaded_urls = {}
-      broken_images, large_images = [], []
+      large_images = post.custom_fields[LARGE_IMAGES].presence || []
+      broken_images, new_large_images = [], []
 
       extract_images_from(post.cooked).each do |image|
         src = original_src = image['src']
@@ -57,7 +60,7 @@ module Jobs
         if is_valid_image_url(src)
           begin
             # have we already downloaded that file?
-            unless downloaded_urls.include?(src)
+            unless downloaded_urls.include?(src) || large_images.include?(src) || broken_images.include?(src)
               if hotlinked = download(src)
                 if File.size(hotlinked.path) <= @max_size
                   filename = File.basename(URI.parse(src).path)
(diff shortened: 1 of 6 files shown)


app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -46,7 +48,8 @@ def execute(args)
       raw = post.raw.dup
       start_raw = raw.dup
       downloaded_urls = {}
-      broken_images, large_images = [], []
+      large_images = eval(post.custom_fields[LARGE_IMAGES].presence || "[]")
```
ZogStriP: Why is `eval` needed here?

app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -104,15 +108,18 @@ def execute(args)
 
       end
 
+      post.custom_fields[LARGE_IMAGES] = large_images.to_s
```
ZogStriP: Why are you doing a `.to_s`? CustomField should support arrays. If they don't, it's best to serialize using JSON.

app/jobs/regular/pull_hotlinked_images.rb
SamSaffron: yes please ... absolutely no eval here.

lib/cooked_post_processor.rb
```diff
@@ -81,11 +81,70 @@ def post_process_images
...
 
     images.each do |img|
+      next if large_images.include?(img["src"]) && add_large_image_placeholder!(img)
+
       limit_size!(img)
       convert_to_link!(img)
     end
   end
 
+  def add_large_image_placeholder!(img)
```
ZogStriP: Nokogiri API is way too verbose. Is there a way we could inject HTML instead?

lib/cooked_post_processor.rb
```diff
@@ -81,11 +81,70 @@ def post_process_images
...
+        lspan.add_previous_sibling(i)
+      end
+    end
+
+    img.remove
+    true
+  end
+
+  def large_images
+    @large_images ||= eval(@post.custom_fields[Jobs::PullHotlinkedImages::LARGE_IMAGES].presence || "[]")
```
ZogStriP: Same here. It's best to avoid `eval`.

lib/cooked_post_processor.rb
vinothkannans: Yes. But we already using it in same class to create lightbox and I thought it's more clean comparing to raw HTML. I guess you feel like those repeated lines of Nokogiri are not necessary here. So I refactored everything by common methods. Also I changed some existing code using those methods. Please have a look now.

SamSaffron:
@tgxworld is this good to merge?

SamSaffron:
@ZogStriP is this good to merge now?

PR #5293, opened 2017-11-07
FEATURE: Yahoo Groups forum importer

jonbartlett:
This importer pulls raw data exported into a simple MongoDB collection. The exporter can be found at 
https://github.com/jonbartlett/yahoo-groups-export

Supports:
- Basic user creation from Yahoo "Profile Name" and "From" / email address
- Topics and posts based on Yahoo Group structure. Posts are ordered by Yahoo Group "msgid".

Does not support:
- File attachments to posts

### DIFF
--- a/script/import_scripts/yahoogroup.rb
+++ b/script/import_scripts/yahoogroup.rb
@@ -1,0 +1,159 @@
+require File.expand_path(File.dirname(__FILE__) + "/base.rb")
+require 'mongo'
+
+# Import YahooGroups data as exported into MongoDB by:
+#   https://github.com/jonbartlett/yahoo-groups-export
+#
+#   Optionally paste these lines into your shell before running this:
+#
+#   =begin
+#   export CATEGORY_ID=<CATEGORY_ID>
+#   =end
+
+class ImportScripts::YahooGroup < ImportScripts::Base
+
+  MONGODB_HOST = '192.168.10.1:27017'
+  MONGODB_DB   = 'syncro'
+
+  def initialize
+    super
+
+    client = Mongo::Client.new([ MONGODB_HOST ], database: MONGODB_DB)
+    db = client.database
+    Mongo::Logger.logger.level = Logger::FATAL
+    puts "connected to db...."
+
+    @collection = client[:posts]
+
+    @user_profile_map = {}
+
+  end
+
+  def execute
+    puts "", "Importing from Mongodb...."
+
+    import_users
+    import_discussions
+
+    puts "", "Done"
+  end
+
+  def import_users
+
+    puts '', "Importing users"
+
+    # fetch distinct list of Yahoo "profile" names
+    profiles = @collection.aggregate(
+                 [
+                  { "$group": { "_id": { profile: "$ygData.profile"  } } }
+                 ]
+            )
+
+    user_id = 0
+
+    create_users(profiles.to_a) do |u|
+
+      user_id = user_id + 1
+
+      # fetch last message for profile to pickup latest user info as this may have changed
+      user_info = @collection.find("ygData.profile": u["_id"]["profile"]).sort("ygData.msgId": -1).limit(1).to_a[0]
+
+      # Store user_id to profile lookup
+      @user_profile_map.store(user_info["ygData"]["profile"], user_id)
+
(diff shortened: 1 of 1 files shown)


gschlager:
Thanks for the new importer!

Could you please run RoboCop and fix all the issues it reports?
https://meta.discourse.org/t/rubocop-has-landed-on-discourse-policewoman-policeman/67211

Also, it would be great if you could squash your commits when you are done 😉

script/import_scripts/yahoogroup.rb
```diff
@@ -0,0 +1,150 @@
...
+
+# Import YahooGroups data as exported into MongoDB by:
+#   https://github.com/jonbartlett/yahoo-groups-export
+
+class ImportScripts::YahooGroup < ImportScripts::Base
+
+  def initialize
+    super
+
+    client = Mongo::Client.new([ '192.168.10.1:27017'], :database => 'syncro')
```
gschlager: These should be constants and/or environment variables.
I suggest doing something like this: https://github.com/discourse/discourse/blob/master/script/import_scripts/mbox.rb#L22-L25

script/import_scripts/yahoogroup.rb
```diff
@@ -0,0 +1,150 @@
...
+      end
+
+      topic = {
+        id: tidx+1,
+        user_id: @user_profile_map[topic_post["ygData"]["profile"]] || -1,
+        raw: topic_post["ygData"]["messageBody"],
+        created_at: Time.at(topic_post["ygData"]["postDate"].to_i),
+        cook_method: Post.cook_methods[:raw_html],
+        title: topic_title,
+        category: ENV['CATEGORY_ID'],
```
gschlager: The usage of this environment variable is quite hidden. I'd suggest documenting it at the beginning of the import script (see https://github.com/discourse/discourse/blob/master/script/import_scripts/mbox.rb#L6-L12).

ZogStriP:
I'm curious, why do you need the MongoDB database? Seems to me that the importer could directly talk to Yahoo Groups API.

script/import_scripts/yahoogroup.rb
jonbartlett: ok, documented at the start of the import script.

jonbartlett: commits squashed as well!

jonbartlett:
@ZogStriP The importer could talk direct to the Yahoo Group API but you would lose the ability to perform any aggregation and cleansing prior to the import. The intermediary db is there for just that.

gschlager:
@jonbartlett I think you could have used an SQLite database instead of MongoDB, like we do with the mbox importer, so there wouldn't be the need for installing another database. And I guess it would be great if the import script did all the talking to the Yahoo API as well. I don't see a reason for having two different scripts.

Anyway, I'm merging this in. Thanks for all your work! And feel free to submit another PR if you want to make some improvements to the import script. 😉

jonbartlett:
Great thanks @gschlager. MongoDB was a bit of a folly in that I just wanted to explore it so take your point about SQLite.

WRT to the import script talking to Yahoo direct, that is an option however the advantages of the intermediary DB are great in that data can be easily manipulated before it goes back into Discourse. I am looking to do this now with the Yahoo Group forum I have just migrated. Turns out the Yahoo Group data isn't that clean and some changes to message structure have occurred over the (long) lifetime of the product.

mcwumbly:
FWIW, for the two importers I wrote, I chose to separate the two scripts:

1. export old forum to some database
2. import to Discourse

(In my case, I just used a directory of .json text files for the database.)

In my case, it was possible that we might lose access to the old forum before we were ready to import it to our production Discourse instance.  It also gave us time to test step 2 independently in a test instance of Discourse several times to review how we wanted to map groups/categories etc, without having to keep hitting the API of the source forum.

Just chiming in to support that pattern of separating export and import in general...

ylluminate:
@jonbartlett can this also be used import and create users that can then activate their accounts if they so desire to move with the migration?

Dreadaxe:
> can this also be used import and create users that can then activate their accounts if they so desire to move with the migration?

@jonbartlett write:
> Basic user creation from Yahoo "Profile Name" and "From" / email address

PR #5295, opened 2017-11-07
some improvements for importers

discoursehosting:
- better handling of code blocks for bbpress and phpbb3
- import only users that actually participate in bbPress part of the wordpress installation
- import password hashes for bbpress
- create permalinks for topics in bbpress


ZogStriP:
Thanks

PR #5297, opened 2017-11-09
[DEV] fixes sublime text 2 project loading issue

typeoneerror:
I'm probably the only person who still uses ST2, but here we are.


typeoneerror:
@discoursebot done

PR #5303, opened 2017-11-11
FEATURE: Option to export multiple categories using export_category method

vinothkannans:


SamSaffron:
Looks fine to me @nlalonde  can you review and merge?

nlalonde:
I missed the discussion where all the work was moved into one class, the "base" exporter. Why was that done?

vinothkannans:
@nlalonde In another task where I created `CategoryStructureExporter` it needed many duplicated methods which all are from `CategoryExporter` class. Also I found so many code duplication in `ImportExport` module. So I decided to have base class.

Now we can create new exporter easily in few steps. And all the exporters following same structure of output. So any exported json file can be merged in single import method. `bundle exec rake import:file["file_name"]`

To say simply I just refactored everything in ImportExport

PR #5307, opened 2017-11-11
FIX: do not move small post actions

gschlager:
Fixes `ActiveRecord::RecordNotUnique (PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_posts_on_topic_id_and_post_number"`

I'll add specs if we want this to be changed.


SamSaffron:
This is probably fine, but really it should be tested somehow ideally.

PR #5308, opened 2017-11-13
fix error "extension "hstore" already exists"

bill2004158:
Error executing 'postInstallation': ERROR:  extension "hstore" already exists

I just saw that error from console, hope this can fix that error.


SamSaffron:
This is probably fine, but unfortunately we can not check this PR till we have a CLA signed 😢

tgxworld:
ping @bill2004158

bill2004158:
already signed?

tgxworld:
Thanks! @bill2004158

PR #5310, opened 2017-11-13
FIX: Changing misattributed Mahatma Gandhi's quote

quimgil:
As reported at https://meta.discourse.org/t/gandhi-misattributed/74022


quimgil:
CLA signed.

PR #5311, opened 2017-11-14
FEATURE: Replace SimpleRSS with Ruby RSS module

xrav3nz:
> DISCUSSION: https://meta.discourse.org/t/simple-rss-is-buggy-and-should-be-forked-or-replaced/27529/23?u=xrav3nz

This PR got a bit chubby, so here's a brief breakdown:

1. More tests: `PollFeedJob` should correctly parse ATOM feeds
  
    #5199 adds the integration tests for parsing and creating topics from RSS feed, this adds the tests for ATOM feed

2. `FeedItemAccessor` 

    <details>
    <summary>  
     TL;DR this provides a consistent interface for accessing the content of a feed item element.
    </summary>

    <br/>

    Say we have a `rss_item` and an `atom_entry`, both parsed by the built-in Ruby RSS module. The interface for accessing the `title` are quite different: `rss_item.title` and `atom_entry.title.content` respectively.

    `FeedItemAccessor` acts as a general wrapper that provides a consistent interface: `accessor.element_content(:title)` will extract the content of the `title` element, and thus saves the pain from dealing with two different type of feed elements.
    </details>

3. `FeedElementInstaller`

    It 'installs' new elements so that `RSS::Parser` will accept and parses non-standard RSS/ATOM elements. EDIT: See https://github.com/discourse/discourse/pull/5311#issuecomment-345152707.

4. replace `SimpleRSS` with Ruby RSS module

    Having the aforementioned points already in-place, this is actually the easier part. 🕶 

---

Open to suggestions 💪 Cheers

### DIFF
--- a/app/jobs/scheduled/poll_feed.rb
+++ b/app/jobs/scheduled/poll_feed.rb
@@ -2,9 +2,12 @@
 # Creates and Updates Topics based on an RSS or ATOM feed.
 #
 require 'digest/sha1'
+require 'open-uri'
+require 'rss'
+require_dependency 'feed_item_accessor'
+require_dependency 'feed_element_installer'
 require_dependency 'post_creator'
 require_dependency 'post_revisor'
-require 'open-uri'
 
 module Jobs
   class PollFeed < Jobs::Scheduled
@@ -46,13 +49,11 @@ module Jobs
 
     def import_topic(topic)
       if topic.user
-        TopicEmbed.import(topic.user, topic.url, topic.title, CGI.unescapeHTML(topic.content.scrub))
+        TopicEmbed.import(topic.user, topic.url, topic.title, CGI.unescapeHTML(topic.content))
       end
     end
 
     class Feed
-      require 'simple-rss'
-
       def initialize
         @feed_url = SiteSetting.feed_polling_url
         @feed_url = "http://#{@feed_url}" if @feed_url !~ /^https?\:\/\//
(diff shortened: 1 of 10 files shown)


app/jobs/scheduled/poll_feed.rb
```diff
@@ -79,35 +76,43 @@ def topics
...
 
       def fetch_rss
-        SimpleRSS.parse open(@feed_url, allow_redirections: :all)
-      rescue OpenURI::HTTPError, SimpleRSSError
+        if SiteSetting.embed_username_key_from_feed.present?
+          FeedElementInstaller.install_rss_element(SiteSetting.embed_username_key_from_feed)
+          FeedElementInstaller.install_atom_element(SiteSetting.embed_username_key_from_feed)
+        end
+
+        RSS::Parser.parse(open(@feed_url, allow_redirections: :all), false)
```
SamSaffron: This feels a bit unsafe due to infinite redirection potential, should probably use final destination here to be consistent and probably excon to download (we do want a handover method on final destination so maybe add that as well)

xrav3nz: > should probably use final destination here to be consistent and probably excon to download

Didn't know about the latter, will do!

> a handover method on final destination

Could you explain this a bit more? Specifically what is a handover method and what is it expected to do?

SamSaffron: specifically final destination is a bit wasteful cause it does a HEAD only to do a GET afterwards to get the actual data, in this case a single GET would do, so it would be nice if you could use it to handover the "open" GET request and avoid the HEAD req.

xrav3nz: Thanks for taking the time to explain that, but I am still a bit confused 😰 on the following point  

> in this case a single GET would do

The proposed flow to resolve the potential infinite redirection is to: (1) get the final destination of the feed url (with HEAD, internally), and then (2) `excon` GET's the resolved url.

The HEAD request is unneeded when there are no (infinite) redirection, of which we cannot be sure without FinalDestination (that sends HEADs). What am I missing here?

SamSaffron: Thing is FinalDestination could just use GET and hand over the get something like this could work

```
txt = FinalDestination.new(url).get
```

Then it can internally be smart enough to do GETs instead of HEAD

This optimises for the general case where a single GET is good enough to find the final destination, redirects are pretty much the same cost be they HEAD or GET so overall cost is unconditionally cheaper.

SamSaffron: I am scope creeping here, so for starters just implement final destination as it is implemented elsewhere, I can merge it in and then we can do another pr for any final destination optimisations.

xrav3nz:
Will look at the FinalDestination optimizations and open another PR when ready!

---

Raising my question about non-standard elements again:

> The current SimpleRSS-backed `PollFeedJob` supports parsing a non-standard RSS/ATOM element for the SiteSetting `embed_username_key_from_feed`. Was this intentional? Should we continue to support this?

The RSS module is not friendly towards non-standard elememts, at all. I put together [`FeedElementInstaller`][1] as a solution after digging into the RSS module's source. The caveat is, namespaced elements like `custom:element` doesn't work :/

If we do decide to continue support non-standard tags, I think I should try to improve `FeedElementInstaller` to accept the namespaced elements (as per [RSS 2.0 Specifications#Extending RSS][2]), otherwise I will drop `FeedElementInstaller` altogether.

[1]: https://github.com/discourse/discourse/pull/5311/commits/4f173d34a3bf931fb1c272f80822a327760d70b7
[2]: http://cyber.harvard.edu/rss/rss.html#extendingRss

xrav3nz:
As per [RSS 2.0 Specifications#Extending RSS][1]:

> A RSS feed may contain [non-standard elements], only if those elements are **defined in a namespace**

`FeedElementInstaller` only installs valid (i.e. namespaced) elements. Let me know if this may not be acceptable.

[1]: http://cyber.harvard.edu/rss/rss.html#extendingRss

SamSaffron:
I think this is ok, @eviltrout what is your call here?

coding-horror:
I vote we merge this @samsaffron if you are Ok with it. It's now or never (post 2.0)

SamSaffron:
OK, merging this, thanks @xrav3nz

PR #5313, opened 2017-11-14
FEATURE: search logs page

techAPJ:
Discussion here: https://meta.discourse.org/t/search-logs-page/73281


SamSaffron:
This looks good to me, we should do another change to make the search logs keep loading beyond 100

PR #5318, opened 2017-11-15
FIX: Hotlinked image not replaced with placeholder

vinothkannans:
Hotlinked image links are not replaced with placeholder while clicking "Rebuild HTML" button.


ZogStriP:
The fix is actually more involved. I fixed it locally. I just need to update the specs. I'll push it tomorrow.

PR #5322, opened 2017-11-16
Cast the numeric keys of `USER_HOMES` to numbers

notriddle:
When converting from hard-coded functions to a map,
the key `1` got converted to `"1"`.
This broke the user-customizable homepage dropdown menu.


SamSaffron:
Thanks!

PR #5323, opened 2017-11-17
If there is a specific poll error message it should be shown to the user

angusmcleod:
e.g. when voting error messages such as poll.post_is_deleted and poll.topic_must_be_open_to_vote are thrown from the server ([see here](https://github.com/discourse/discourse/blob/master/plugins/poll/plugin.rb#L42)), however the client only ever displays the generic error message.


ZogStriP:
Thanks 👍

PR #5329, opened 2017-11-20
FIX: Presence bug

davidtaylorhq:
**Repro steps:**
- User 1 starts writing reply, then closes browser window
- Some time later, User 2 starts writing a reply. It looks like User 1 is still writing. 20 seconds later, User 1 drops off the "replying" indicator

**Expected behavior:**
User 1 drops off the "replying" indicator as soon as User 2 opens the composer

**Cause:**
I mis-used Ruby's `||=` operator, which meant that sometimes the logic on the right hand side wouldn't get triggered if the left hand side was already true.

**Fix:**
I added a test case for this situation, and have fixed the logic

### DIFF
--- a/plugins/discourse-presence/plugin.rb
+++ b/plugins/discourse-presence/plugin.rb
@@ -122,9 +122,9 @@ after_initialize do
         if topic
           guardian.ensure_can_see!(topic)
 
-          any_changes = false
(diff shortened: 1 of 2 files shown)


ZogStriP:
Awesome, I had this happen a few times 👍

PR #5332, opened 2017-11-21
FIX: updating topic stats failed silently for invalid topics

gschlager:
Updating topic stats failed silently when the topic was invalid (e.g. because the topic title is too short).


lib/post_creator.rb
```diff
@@ -412,7 +412,7 @@ def update_topic_stats
       attrs[:word_count] = (@topic.word_count || 0) + @post.word_count
       attrs[:excerpt] = @post.excerpt(220, strip_links: true) if new_topic?
       attrs[:bumped_at] = @post.created_at unless @post.no_bump
-      @topic.update_attributes(attrs)
+      @topic.update_columns(attrs)
```
gschlager: I wonder if I should update the `updated_at` column as well, since `update_columns()` doesn't change that value.

PR #5357, opened 2017-11-23
FIX: invited users were not granted trust level based on their group

techAPJ:
https://meta.discourse.org/t/bulk-invites-and-trust-level/73535

If the user enters a password when accepting invite they were not granted trust level based on their group privileges. It was because `password_required` was set to `true` when creating user record and when the user was updated again when granting trust level the password validation was raising an error saying that the password is empty, this was because `password_required` was still set to `true`.

This commit fetches fresh user record after the user is created so that the user record can be updated successfully.


coding-horror:
Are there any security risks with this change?

techAPJ:
@coding-horror No security risks here, just fetching fresh user record so that it can be updated subsequently/

PR #5364, opened 2017-11-24
FEATURE: New site setting for additional allowed filetypes for staff

OsamaSayegh:
https://meta.discourse.org/t/setting-to-only-let-staff-upload-media-attachments/26301/8?u=osama


config/site_settings.yml
```diff
@@ -758,6 +758,11 @@ files:
     default: 'jpg|jpeg|png|gif'
     refresh: true
     type: list
+  additional_authorized_extensions_for_staff:
```
tgxworld: Maybe we should just name it `authorized_extensions_for_staff`? `additional_authorized_extensions_for_staff` seems abit verbose to me. 

cc @coding-horror

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -274,7 +283,7 @@ export function authorizedImagesExtensions() {
 }
 
 export function authorizesAllExtensions() {
-  return Discourse.SiteSettings.authorized_extensions.indexOf("*") >= 0;
+  return rawExtensions().indexOf("*") >= 0;
```
tgxworld: I don't think we need to concat the settings, I'll prefer if we just add another condition that checks against `siteSettings.authorized_staff_extensions` if the user is a staff.

OsamaSayegh:
Hi @tgxworld, thanks for the feedback

I've updated thid PR, can you take a look now? thanks!

SamSaffron:
note we need to wait one more week on this, we plan to release 1.9 in the next week and this is a risky merge.

SamSaffron:
@ZogStriP can you be in charge of getting this merged in?

lib/validators/upload_validator.rb
```diff
@@ -50,38 +50,50 @@ def maximum_attachment_file_size(upload)
...
+    authorized_extensions(upload) - FileHelper.images
   end
 
   def authorizes_all_extensions?(upload)
+    if upload.user&.staff?
+      return true if SiteSetting.authorized_extensions_for_staff.include?("*")
+    end
     extensions = upload.for_theme ? SiteSetting.theme_authorized_extensions : SiteSetting.authorized_extensions
-    extensions.include?("*")
+    return extensions.include?("*")
```
ZogStriP: You don't need that `return` ;) Ruby always return the value of the last expression

spec/controllers/uploads_controller_spec.rb
```diff
@@ -147,6 +147,40 @@
         expect(id).to be
       end
 
+      it 'respects `authorized_extensions_for_staff` setting when staff upload file' do
+        SiteSetting.authorized_extensions = ""
+        SiteSetting.authorized_extensions_for_staff = "*"
+        @user.update_columns(moderator: true)
+
+        message = MessageBus.track_publish('/uploads/composer') do
```
ZogStriP: We don't need the messagebus now that @SamSaffron added the hijack middleware.

ZogStriP:
@OsamaSayegh this looks good 👍 Can you update the tests to remove the MessageBus tracking part which isn't needed anymore?

OsamaSayegh:
Hi @ZogStriP thanks for the review

FYI I just pushed a couple of commits to address the things you pointed out :)

SamSaffron:
@ZogStriP  can we aim to get this merged this week?

ZogStriP:
Thanks 👍

PR #5365, opened 2017-11-25
Update client.ru.yml

Toxu-ru:
Adding value (few)


gschlager:

Toxu-ru:
Is it possible to add via Transifex values ​​that are not in the English version of the file: `few`
`[ru.notifications.group_message_summary.few]` 
etc.

gschlager:
Yes, you can do that.

PR #5366, opened 2017-11-26
Move queuedForTyping assignment

angusmcleod:
See further: https://meta.discourse.org/t/composer-messages-exception/74829


SamSaffron:
thanks!

PR #5369, opened 2017-11-27
FIX: Error if queued post not found while updating

vinothkannans:


app/controllers/queued_posts_controller.rb
```diff
@@ -20,6 +20,8 @@ def index
   def update
     qp = QueuedPost.where(id: params[:id]).first
 
+    return render_json_error "Post not found or already updated." if qp.blank?
```
tgxworld: Tip here, any strings that are client facing needs to be translated. Also anytime you notice that you're duplicating a string in your application code and tests, it'll be much better to add the string to i18n so that we avoid duplicated strings.

spec/controllers/queued_posts_controller_spec.rb
```diff
@@ -31,6 +31,21 @@
...
+    context 'not found' do
+      it 'returns json error' do
+        id = qp.id
+        qp.destroy
+
+        put :update, params: {
+          id: id, queued_post: { state: 'approved' }
+        }, format: :json
+
+        expect(response).not_to be_success
```
tgxworld: I'm personally not a fan of using `not_to be_success` as it just means `response.status != 200`. Instead I would test directly for the right response status. Is it a 404, 403 or 500?

spec/controllers/queued_posts_controller_spec.rb
```diff
@@ -31,6 +31,21 @@
     let!(:user) { log_in(:moderator) }
     let(:qp) { Fabricate(:queued_post) }
 
+    context 'not found' do
+      it 'returns json error' do
+        id = qp.id
+        qp.destroy
```
tgxworld: When we don't have to check the return value of `destroy` or `save`, we should use `destroy!` or `save!` so that an error is raised instead of having the action fail silently.

spec/controllers/queued_posts_controller_spec.rb
```diff
@@ -31,6 +31,21 @@
     let!(:user) { log_in(:moderator) }
     let(:qp) { Fabricate(:queued_post) }
 
+    context 'not found' do
+      it 'returns json error' do
+        id = qp.id
```
tgxworld: This variable doesn't seemed to be reused anywhere.

vinothkannans:
@tgxworld changes are done 👍

tgxworld:
@vinothkannans  Nice :+1:

PR #5370, opened 2017-11-27
FEATURE: replace GUID in image name uploaded from iOS

gschlager:
https://meta.discourse.org/t/image-hash-code-in-description-of-forum-topic-shared-on-facebook-twitter/74382

Currently it replaces the GUID only in images uploaded from iOS. I didn't want to change the alt text for images uploaded from other systems since uploading files with a GUID as name could be a valid usecase.

<del>Unfortunately, importing `isAppleDevice` from `safari-hacks` adds a lot of complexity to a simple feature and I'm not sure that's warranted. What's your your opinion, @SamSaffron ?</del>

I moved `isAppleDevice()` into `discourse/lib/utilities`. Makes much more sense to me.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,6 +27,7 @@
 //= require ./discourse/lib/formatter
 //= require ./discourse/lib/eyeline
 //= require ./discourse/lib/show-modal
+//= require ./discourse/lib/safari-hacks
 //= require ./discourse/mixins/scrolling
 //= require ./discourse/lib/ajax-error
 //= require ./discourse/models/model
@@ -49,7 +50,6 @@
 //= require ./discourse/lib/autocomplete
 //= require ./discourse/lib/after-transition
 //= require ./discourse/lib/debounce
-//= require ./discourse/lib/safari-hacks
 //= require_tree ./discourse/adapters
 //= require ./discourse/models/post-action-type
 //= require ./discourse/models/post
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,7 +27,6 @@
 //= require ./discourse/lib/formatter
 //= require ./discourse/lib/eyeline
 //= require ./discourse/lib/show-modal
-//= require ./discourse/lib/safari-hacks
 //= require ./discourse/mixins/scrolling
 //= require ./discourse/lib/ajax-error
 //= require ./discourse/models/model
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -285,6 +286,21 @@ function uploadTypeFromFileName(fileName) {
   return isAnImage(fileName) ? 'image' : 'attachment';
 }
 
+function isGUID(value) {
+  return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-fA-F]{12}$/i.test(value);
```
ZogStriP: In the last `[0-9a-fA-F]{12}`, the uppercase `A-F` isn't need since the regex is globally case insensitive ;)

app/assets/javascripts/application.js
```diff
@@ -27,6 +27,7 @@
 //= require ./discourse/lib/formatter
 //= require ./discourse/lib/eyeline
 //= require ./discourse/lib/show-modal
+//= require ./discourse/lib/safari-hacks
```
ZogStriP: I would extract `isAppleDevice` from `safari-hacks`

gschlager: Yeah, I was thinking about moving it into `discourse/lib/utilities`. Not sure if there's even a better place.

ZogStriP: seems fine to me

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -285,6 +285,21 @@ function uploadTypeFromFileName(fileName) {
...
 
+function isGUID(value) {
+  return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(value);
+}
+
+function imageNameFromFileName(fileName) {
+  const split = fileName.split('.');
+  const name = split[split.length-2];
+
+  if (exports.isAppleDevice() && isGUID(name)) {
```
gschlager: Using `exports.isAppleDevice()` feels like a hack to me. Is there a better way to make this testable? It looks like I can't override `navigator.userAgent`

SamSaffron:
Wait, what happens on android?

gschlager:
Chrome on Android works just fine according to my tests. As far as I can see only Safari on Android replaces the original filename with a GUID during upload.

coding-horror:
Must be new to iOS 11 I guess

gschlager:
> Must be new to iOS 11 I guess

Maybe. I updated my iPad to iOS 10, so I can't test that anymore.

PR #5373, opened 2017-11-27
Feature: Add service worker registration method to plugin API

featheredtoast:
Adds a plugin hook to register service workers.

Adds registered service worker scripts to the /service-worker.js template.


tgxworld:
Merged in https://github.com/discourse/discourse/commit/b094894c9404b2e2806796999d9e84c66c17203b.

@featheredtoast Instead of generating a new copy of the js file each time, we precompile and serve it via  the proxy route you added.

tgxworld:
Thank you for the work @featheredtoast

PR #5378, opened 2017-11-28
FIX: Skip plain text pasting if clipboard have file data

vinothkannans:
If you paste into the composer and the clipboard has data available in both an image and text format, they both appear:

### DIFF
--- a/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
+++ b/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
@@ -1,10 +1,4 @@
-export function isAppleDevice() {
-  // IE has no DOMNodeInserted so can not get this hack despite saying it is like iPhone
-  // This will apply hack on all iDevices
-  return navigator.userAgent.match(/(iPad|iPhone|iPod)/g) &&
-         navigator.userAgent.match(/Safari/g) &&
-         !navigator.userAgent.match(/Trident/g);
-}
+import { isAppleDevice } from 'discourse/lib/utilities';
 
 // we can't tell what the actual visible window height is
 // because we cannot account for the height of the mobile keyboard
--- a/app/assets/javascripts/discourse/lib/show-modal.js.es6
+++ b/app/assets/javascripts/discourse/lib/show-modal.js.es6
@@ -1,4 +1,4 @@
-import { isAppleDevice } from 'discourse/lib/safari-hacks';
+import { isAppleDevice } from 'discourse/lib/utilities';
 
 export default function(name, opts) {
   opts = opts || {};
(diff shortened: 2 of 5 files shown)


vinothkannans:
Travis error not related to this PR

coding-horror:
I think it should be the other way around ; skip image data if you have
plain text.

vinothkannans:
Users can use "Paste as plain text" context menu or <kbd>Ctrl + Shift + V</kbd> to paste plain text. That's why I skipped plain text by default. If we did it in other way then the user can't do "Paste as image" action at all.

coding-horror:
Only when both paste types are detected, then we should skip one of them,
the image one. We should not prioritize image pastes unless there is no
other way to do this.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -616,6 +616,16 @@ export default Ember.Component.extend({
     Ember.run.scheduleOnce("afterRender", () => $textarea.focus());
   },
 
+  paste: function(e) {
+    var items = (e.clipboardData || e.originalEvent.clipboardData).items;
```
tgxworld: Note that we prefer `let` instead of `var` because 

> let allows you to declare variables that are limited in scope to the block, statement, or expression on which it is used. This is unlike the var keyword, which defines a variable globally, or locally to an entire function regardless of block scope.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -616,6 +616,16 @@ export default Ember.Component.extend({
...
+
+    $.each(items, function(i, d) {
```
tgxworld: In ES6, you can use [forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) instead of `$.each`

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -616,6 +616,16 @@ export default Ember.Component.extend({
     Ember.run.scheduleOnce("afterRender", () => $textarea.focus());
   },
 
+  paste: function(e) {
```
tgxworld: Method definitions are nicer starting from ES2015, something like 

```
paste(e) {
  ... 
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

app/assets/javascripts/discourse/components/d-editor.js.es6
tgxworld: O you should probably have a look at [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) as well.

PR #5380, opened 2017-11-29
Fix infinite loop when poll step is zero

cavneb:
Fixes https://meta.discourse.org/t/number-rating-poll-allows-step-width-0/74542

### DIFF
--- a/plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
+++ b/plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
@@ -87,7 +87,10 @@ export default Ember.Controller.extend({
     if (isMultiple) {
       return this._comboboxOptions(pollMinInt + 1, count + 1);
     } else if (isNumber) {
-      const pollStepInt = parseInt(pollStep) || 1;
+      let pollStepInt = parseInt(pollStep, 10);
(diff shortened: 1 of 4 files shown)


ZogStriP:
This is great 👍 Would you mind adding a test ensuring we don't regress?

cavneb:
Definitely

cavneb:
@ZogStriP I've been learning about the the test / lint setup as we go. It's still not passing Travis. I will pick up on it tomorrow 👍

spec/models/optimized_image_spec.rb
```diff
@@ -118,7 +118,7 @@ def local(url)
         end
 
         it "downloads a copy of the original image" do
-          Tempfile.any_instance.expects(:close!)
+          Tempfile.any_instance.expects(:close!).twice
```
tgxworld: Just curious, is this related to the problem in this PR? :smile_cat:

ZogStriP: Yeah that's weird.

cavneb: Tests don't pass locally without it. I'll remove it and run it through again. It must be my environment

ZogStriP:
Thanks 👍

PR #5383, opened 2017-11-30
FIX: start sidekiq process for all queues in non-dev

rossdakin:
Related to discourse/discourse#4735

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,7 +1,7 @@
 ---
 development:
   :concurrency: 5
-  :queues:
-    - [critical,4]
-    - [default, 2]
-    - [low]
+:queues:
+  - [critical,4]
+  - [default, 2]
+  - [low]


SamSaffron:
is there a reason this is required? Our production docker image takes care of it.

rossdakin:
Maybe I'm doing it wrong.

We're using a [Bitnami release](https://github.com/bitnami/bitnami-docker-discourse) that was only processing the default queue. This change fixed it, though I'm not familiar enough with container architecture to know if it's the right thing to do.

coding-horror:
yes bitnami is an unsupported install, so while I appreciate the effort, we have a policy of only supporting our official docker install, to keep our open source support / human power requirements under control. Definitely document it on meta.discourse.org for others though.

rossdakin:
Got it, thanks @coding-horror!

Out of curiosity, is there any down side to this particular code change?

SamSaffron:
Probably no downside except that we need to test it and confirm it has no
weird side effects

PR #5385, opened 2017-11-30
FEATURE: support search click through tracking for user, category and tags

techAPJ:
https://meta.discourse.org/t/search-logs-page/73281/11?u=techapj

This commit adds following features:

- support for tracking click through to user, tag and category
- new filter for search result type (topic, user, tag and category)
- new filter for search type (header, full page)

This commit also removes "most viewed topic" column from search logs page because we are now tracking multiple click through entities, so topic is not a special entity anymore. This also improves query perf. The query now takes `20.5ms` to run, as opposed to `655.9ms` previously.

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-logs-search-logs.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-logs-search-logs.js.es6
@@ -1,4 +1,11 @@
 export default Ember.Controller.extend({
   loading: false,
-  period: "all"
+  period: "all",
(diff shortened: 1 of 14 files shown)


SamSaffron:
I don't feel the filtering of "Tag/User" etc is really of value there, we can add that kind of breakdown when you drill down to a term.

techAPJ:
@SamSaffron Okay, I removed the "search result type" filter.

PR #5386, opened 2017-11-30
FIX: Unable to choose user title as <none>

vinothkannans:


app/assets/javascripts/discourse/controllers/preferences/badge-title.js.es6
```diff
@@ -1,28 +1,9 @@
...
 
 export default Ember.Controller.extend(BadgeSelectController, {
 
   filteredList: function() {
     return this.get('model').filterBy('badge.allow_title', true);
-  }.property('model'),
+  }.property('model')
 
-  actions: {
-    save() {
```
ZogStriP: Does that mean that there is another way of saving the value and that this one was redundant?

vinothkannans: Yes. Currently we do exactly same thing in two places **[1](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/badge-title.js.es6#L11-L26), [2](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/controllers/preferences/badge-title.js.es6#L11-L26)**. I don'w know why. And only the component save function is worked for me while checking.

PR #5391, opened 2017-12-04
FEATURE: introduces minimum trust level for polls

jjaffeux:
This commit makes `poll_enabled` less misleading and introduces `poll_minimum_trust_level_to_create`. If poll are enabled they will always be cooked, and if you have the required trust level you can create polls. As a side effect, it also fixes a bug where rebaking a post created by staff member when `poll_enabled=false` would end up not cooking it.

It also adds more tests to ensure settings are respected.


ZogStriP:
LGTM 👍

PR #5395, opened 2017-12-04
FEATURE: Convert HTML to Markdown while pasting in composer

vinothkannans:


coding-horror:
Can you review @jjaffeux

vinothkannans:
@schungx You should check the code first. Anyway I already aware of that. Thank you.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -636,23 +638,44 @@ export default Ember.Component.extend({
...
+      const html = clipboard.getData("text/html");
+      const self = this;
+      ajax('/composer/parse_html', {
+        type: 'POST',
+        data: { html }
+      }).then(response => {
+        self.appEvents.trigger('composer:replace-text', placeholder, response.markdown);
+      }).catch(error => {
+        if (!plainText) {
+          popupAjaxError(error);
```
SamSaffron: If something fails, why not simply paste in the text in non html, not sure if reporting the error to the user actually helps

vinothkannans: I am reporting only if the `plainText` is not available. If it is available it will be already pasted.

config/routes.rb
```diff
@@ -301,6 +301,7 @@
   get "session/current" => "session#current"
   get "session/csrf" => "session#csrf"
   get "composer_messages" => "composer_messages#index"
+  post "composer/parse_html" => "composer#parse_html"
```
SamSaffron: A controller test or 2 please! 

What if the html content is 3 megabytes? can this take out the parser?

vinothkannans: we already having spec for this [here](https://github.com/discourse/discourse/blob/master/spec/components/html_to_markdown_spec.rb). Should I check this against large input data?

schungx:
@vinothkannans if you look closely, when somebody copy a table, say from Word or from a webpage, there will be both `text/html` and `text/plain`.  In your code, you'll detect that the `plainText` is a table, and handle it first _before_ checking for `text/html`.  That is sub-optimal, because `text/html`, if present, should supersede `text/plain`.  There may be formatting inside the table that can be preserved if you handle `text/html` first.

This edge case is when the copied content is a full table and only a table.

vinothkannans:
@schungx I aware of that too. Both are very new features for Discourse. Only in this order live data testing is possible with both plain text to tables and html to markdown conversions. If I switch it now then almost all the times `text/plain` will be bypassed whenever copy paste from excel tables.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -636,23 +638,44 @@ export default Ember.Component.extend({
...
-    let preventDefault = false;
+    const plainText = clipboard.getData("text/plain");
+    const placeholder = plainText || this.get('pastePlaceholder');
+    let handled = false;
 
-    if (types.some(t => t === "text/plain")) {
-      const text = clipboard.getData("text/plain");
-      const table = this._detectTable(text);
+    if (plainText) {
+      const table = this._detectTable(plainText);
```
jjaffeux: minor complaint but actually here we are not detecting a table, but trying to extract one, so the name could be better (you are actually detecting a table in the middle of detectTable logic).

```
cons table = this._extractTable(plainText);
```

would make more sense I think

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -636,23 +638,44 @@ export default Ember.Component.extend({
...
         this._addText(this._getSelected(), table);
-        preventDefault = true;
+        handled = true;
       }
-    } else if (types.some(t => t === "Files")) {
-      preventDefault = true;
     }
 
-    if (preventDefault) {
+    if (!handled && types.some(t => t === "text/html")) {
```
jjaffeux: Wouldn’t it be better to use includes here ?

```
types.includes("text/html")
```

It looks to me that `some` allows you to provide a custom function if you have a more complicated logic than a `===`

note there's at least another occurrence of some later

app/assets/javascripts/discourse/components/d-editor.js.es6
jjaffeux: btw, are we sure `clipboard.types` will always return an array ?

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -636,23 +638,44 @@ export default Ember.Component.extend({
     return null;
   },
 
+  @computed
```
jjaffeux: not sure why this needs to be a property here, are we using this elsewhere than in `paste(e)` ?

if not I would just replace 

```
const placeholder = plainText || this.get('pastePlaceholder');
```

by

```
const placeholder = plainText || I18n.t('pasting');
```

and remove this

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -636,23 +638,44 @@ export default Ember.Component.extend({
...
+  pastePlaceholder() {
+    return I18n.t('pasting');
+  },
+
   paste(e) {
```
jjaffeux: Generally speaking I find this method hard to follow, the meaning of `handled` is not always clear to me, and part of the code could probably be more explicit.

For example, could we replace this by a name, fully or partially ?
```
(handled || (types.some(t => t === "Files") && !plainText))
```

config/routes.rb
jjaffeux: is there any js test about this somewhere? couldn’t see it?
At least one simple test would be cool, fillIn("xxx", textearea), check conversion in md and html in preview, should be easy enough to do.

app/assets/javascripts/discourse/components/d-editor.js.es6
vinothkannans: yes.

config/routes.rb
vinothkannans: Ruby spec added. @jjaffeux for Qunit test I have to trigger paste event manually and also have to wait for the server response to render the response in editor. Is there any possibility for that?

jjaffeux: Do you ? I don’t think pasting triggers a request.

vinothkannans: no in this particular functionality I am calling web request whenever paste event triggered with html data.

jjaffeux:
Merge 👍 great job @vinothkannans

PR #5398, opened 2017-12-05
FIX: topic timer offset applied two times

jjaffeux:
timezone offset was calculated and sent from browser to server, it would get applied on UTC time generated from '2013-11-22 5:00' format for example and then sent back to browser which would display it thinking it's UTC time using `moment(utcTime)` when it's in fact an UTC time with offset of the initial user timezone.

This is impossible to automatically test in the current app state. Easiest reproduction is in live browser after setting your timezone to `America/New_York`, when setting a topic timer to later_today, after save, the time under the topic should be off to something roughly equal +1/-1  hour to your timezone offset.


jjaffeux:
sam might not be able to check this soon, can you review this @tgxworld please ?

PR #5399, opened 2017-12-05
FIX: makes sur we have polls before validating

jjaffeux:


plugins/poll/plugin.rb
```diff
@@ -312,12 +312,14 @@ def voters
     # only care when raw has changed!
     return unless self.raw_changed? || force
 
-    validator = DiscoursePoll::PostValidator.new(self)
-    return unless validator.validate_post
-
     validator = DiscoursePoll::PollsValidator.new(self)
     return unless (polls = validator.validate_polls)
 
+    if !polls.empty?
```
ZogStriP: You can also say `if polls.present?`

PR #5400, opened 2017-12-05
FEATURE: convert plain text emails to markdown

gschlager:
* Detects quote levels (as long as quotes start with `>`) and modifies quotes, so that they are recognized by our Markdown engine.
* Detects and allows Markdown code blocks with three backticks within the plain text.
* Everything else that looks like a special character in HTML or Markdown gets escaped in order to maintain the actual text from the email.
* Converts tabs into two spaces and each leading space gets replaced by a non-breaking whitespace. This makes code snippets somehow readable.
* Supports RFC 3676 (`format=flowed` and `DelSp=yes`). Otherwise maintains the line breaks from the original email (may change in a future version).
* Removes duplicate links like `http://example.com/foo.html <http://example.com/foo.html>` which are mostly the result of email clients converting HTML to plain text.

Other changes:
* The plain text to markdown conversion is currently only used by the experimental mbox import script and by the email receiver when the incoming email is sent to a mailing list mirror category.
* The experimental mbox import script now uses the regular cook method for imported emails. The email cook method is probably going to be deprecated.

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -2,6 +2,7 @@ require "digest"
 require_dependency "new_post_manager"
 require_dependency "post_action_creator"
 require_dependency "html_to_markdown"
+require_dependency "plain_text_to_markdown"
 require_dependency "upload_creator"
 
 module Email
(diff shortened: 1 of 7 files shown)


techAPJ:
👍

PR #5405, opened 2017-12-07
move /tags/ topic list out of footer

majakomel:
makes html more consistent with other topic lists 
https://meta.discourse.org/t/is-the-html-for-tags-topic-list-supposed-have-a-different-structure/74828


ZogStriP:
Thanks

PR #5406, opened 2017-12-07
[FIX] Only include added fields if object responds to them

gdpelican:
Per the discussion here:
https://meta.discourse.org/t/babble-a-chat-plugin/31753/760

The data-explorer plugin (and perhaps others), can leave a serializer in a state where it's unable to correctly serialize custom fields added by a plugin.

This fixes that by ensuring that the object can receive the message before sending it.


gdpelican:
Hm, on second thought, this isn't quite right; it should be perfectly valid to pass a block to add_to_serializer and not need the value to care if the object responds to it or now. Going to rethink.

PR #5411, opened 2017-12-08
install imagemagick@6 for OSX native development

orlando:
This PR fixes the documentation to install ImageMagick 6 instead of 7 (default version when running `brew install imagemagick`)

## Notes
The only place I found this information was in https://meta.discourse.org/t/beginners-guide-to-install-discourse-on-mac-os-x-macos-for-development/15772/112. Should we add this somewhere else, like we did with PostgreSQL 10?


SamSaffron:
Looks fine to me, really want to get 7 support at some point.

PR #5412, opened 2017-12-08
A typo fix

rizka10:
Mailinglist --> Mailing list


gschlager:
Thanks, but you shouldn't change the key, otherwise this won't work anymore.

rizka10:
Ah, true. I should have updated the key in the code, too of course. Sorry, that was pretty stupid. But I guess the key's name doesn't even matter.

gschlager:
Yeah, I suggest you update only the text and leave everything else alone.

SamSaffron:
ok fixed this by editing the text thanks for the PR

PR #5413, opened 2017-12-08
adding Persian translation

lalehmb:


gschlager:

PR #5414, opened 2017-12-08
Updates api_test_seeds for multiple auth_token support

typeoneerror:
See https://meta.discourse.org/t/discourse-ruby-api-testing-unknown-attribute-auth-token-for-user/75704 for reasoning. 

I'm not super familiar with Discourse, but this seems right-ish. Happy to make adjustments. In summary, this seed hadn't been updated since auth token's were moved to a separate model and not a field on `User` (this was my understanding in any case).


SamSaffron:
looks good to me! thanks

PR #5415, opened 2017-12-09
update server fr.yml

jeremielondon:
Little french typo


ZogStriP:
Thanks but all our translations must be done on https://www.transifex.com/discourse/discourse-org/.

PR #5416, opened 2017-12-10
grey out disabled badges in admin overview

majakomel:
from discussion: https://meta.discourse.org/t/indication-of-disabled-badges-in-admin-overview/75609

before:
screen shot 2017-12-10 at 16 35 50
after:
screen shot 2017-12-10 at 16 35 22


Throne3d:
It looks like the failed Travis build is just the single job failing to clone the git repo – [here](https://travis-ci.org/discourse/discourse/jobs/315025200) should probably be restarted.

SamSaffron:
looks good to me! thanks!

PR #5417, opened 2017-12-11
FIX: save whisper drafts

majakomel:
Meta thread: https://meta.discourse.org/t/drafted-whispers-revert-to-non-whispers-on-page-reload/72202


SamSaffron:
ahh I forgot our server side just does blobs for storage so it is super flexible. looks good to me! thanks

PR #5419, opened 2017-12-12
FIX: save registration_ip_address for staged users logging in via social auth

techAPJ:
Sets `registration_ip_address` for users overtaking their staged account via social login (e.g. Google login).


ZogStriP:
LGTM 👍

techAPJ:
Thanks!

PR #5421, opened 2017-12-12
FIX : proper error for creating topic with tag withour tag creation p…

aki1221:
…ermission.

Creating a topic with a new tag without tag-creation permission fails silently.This
commit raises tag creation not allowed exception for the same.

### DIFF
--- a/lib/topic_creator.rb
+++ b/lib/topic_creator.rb
@@ -29,6 +29,10 @@ class TopicCreator
       topic.errors[:base] << I18n.t("tags.staff_tag_disallowed", tag: staff_only.join(" "))
     end
 
+    unless @guardian.can_create_tag?
+      topic.errors[:base] << I18n.t("tags.creation_not_allowed")
+    end
+
     DiscourseEvent.trigger(:after_validate_topic, topic, self)
     valid &&= topic.errors.empty?
 
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -3499,6 +3499,7 @@ en:
     title: "Tags"
     staff_tag_disallowed: "The tag \"%{tag}\" may only be applied by staff."
     staff_tag_remove_disallowed: "The tag \"%{tag}\" may only be removed by staff."
+    creation_not_allowed: "Tag creation not allowed."
   rss_by_tag: "Topics tagged %{tag}"
 
   finish_installation:


lib/topic_creator.rb
```diff
@@ -27,6 +27,8 @@ def valid?
     # not sure where this should go
     if !@guardian.is_staff? && staff_only = DiscourseTagging.staff_only_tags(@opts[:tags])
       topic.errors[:base] << I18n.t("tags.staff_tag_disallowed", tag: staff_only.join(" "))
+    elsif !@guardian.can_create_tag? || !SiteSetting.tagging_enabled?
+      topic.errors[:base] << "tag creation not allowed"
```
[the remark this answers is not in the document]
aki1221: i have to do it for english or for all languages in locales.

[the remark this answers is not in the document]
gschlager: You need to add it only for English.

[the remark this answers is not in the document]
aki1221: hi 
changes requested are done please review them.
thanks

lib/topic_creator.rb
```diff
@@ -27,6 +27,8 @@ def valid?
     # not sure where this should go
     if !@guardian.is_staff? && staff_only = DiscourseTagging.staff_only_tags(@opts[:tags])
       topic.errors[:base] << I18n.t("tags.staff_tag_disallowed", tag: staff_only.join(" "))
+    elsif !@guardian.can_create_tag? || !SiteSetting.tagging_enabled?
+      topic.errors[:base] << I18n.t("errors.messages.not_allowed", operation: "Tag Creation")
```
SamSaffron: We can't have this error cause "operation" is not localized, so you end up having a half localized string, also ideally we need a test here. 

A bit uneasy to have the error handling now in 2 spots due to this change.

aki1221: so should i remove operation and write complete error in locale?

SamSaffron: yes, but I am uneasy about having this handled in 2 spots, what error do you get now when you are missing a tag?

aki1221: hi i have done some changes please check.
i am getting "Tag creation not allowed " error when user can not create tags and site setting tagging is not enabled.

aki1221: hi @SamSaffron 
i have done some changes please review. i am getting "Tag creation not allowed" error when user don't have tag creation permission or site setting -tagging is not enabled.

aki1221:
@gschlager 
i have done some changes as suggest by you. Please review.

SamSaffron:
I am sorry, this simply does not feel like the right spot or scope in the fix. 

We have to check first if there are ANY new tags, prior to checking if you are allowed to create tags. Otherwise we will just block you even if you created no new tags.

aki1221:
@SamSaffron 
sorry for wasting your time i did'nt  consider all cases but i truly want to learn and excited about contributing to open source.it will be helpful if you can suggest me the right path.

PR #5424, opened 2017-12-13
display language name in user locale selector

majakomel:
Make user locale selector consistent with admin. Meta: https://meta.discourse.org/t/is-there-any-way-to-let-language-selector-display-english-instead-of-en/14518


SamSaffron:
This looks good, one thing I would like you to try here is making a test or 2 for it.

PR #5425, opened 2017-12-13
FEATURE: HTML to Markdown conversion using native JavaScript ES6 classes

vinothkannans:


app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -0,0 +1,211 @@
+import parseHTML from 'discourse/helpers/parse-html';
+
+String.prototype.trimLeft = function() {
```
eviltrout: You should not modify the string prototype in this class. It will mean there's a side effect of including this file in the global scope.

If you need a trimLeft function just make it local to this file.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -0,0 +1,211 @@
...
+      return [this.prefix, text, this.suffix].join("");
+    }
+
+    return text;
+  },
+
+  innerMarkdown() {
+    let text = this.element.innerMarkdown();
+
+    if (!["li"].includes(this.element.name)) {
```
eviltrout: Isn't this the same as `this.element.name !== 'li'`?

app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -0,0 +1,211 @@
...
+  heading(name, prefix) {
+    return Tag.extend({
+      name: name,
+      prefix: `\n\n${prefix} `,
+      suffix: "\n\n",
+    });
+  },
+
+  emphasis(name, decorator) {
+    return Tag.extend({name: name, prefix: decorator, suffix: decorator});
```
eviltrout: With ES6, instead of `name: name` you can just do `name`. This should also be done on L54-55 for `name` and `text`.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
vinothkannans: few more tag names may added there in near future. maybe I should remove that for now.

test/javascripts/lib/to-markdown-test.js.es6
```diff
@@ -0,0 +1,35 @@
...
+
+QUnit.test("converts a link", assert => {
+  const html = `<a href="https://discourse.org">Discourse</a>`;
+  const markdown = `[Discourse](https://discourse.org)`;
+  assert.equal(toMarkdown(html), markdown);
+});
+
+QUnit.test("put raw URL instead of converting the link", assert => {
+  let url = "https://discourse.org";
+  const html = function () { return `<a href="${url}">${url}</a>` };
```
eviltrout: Why not use ES6 function arrow here? `const html = () => \`<a href="${url}">${url}</a>\`` is much shorter.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -0,0 +1,211 @@
+import parseHTML from 'discourse/helpers/parse-html';
+
+String.prototype.trimLeft = function() {
+    return this.replace(/^\s+/,"");
+}
+
+const Tag = Ember.Object.extend({
```
eviltrout: I don't think we need to use the Ember object model for this, since you are never observing or using computed properties. Using a regular Javascript class (or ES6 class) will be significantly faster.

I see you use `extend` below to set up the properties, but that should be possible via a constructor instead.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
eviltrout: I would recommend adding the array support later if you do other tag names, yes. It would be a very small change to put it back for more tag names, but we may never do that!

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -659,27 +660,9 @@ export default Ember.Component.extend({
     }
 
     if (this.siteSettings.enable_rich_text_paste && html && !handled) {
-      const placeholder = `${ plainText || I18n.t('pasting') }`;
-      const self = this;
-
-      this.appEvents.trigger('composer:insert-text', placeholder);
+      const markdown = toMarkdown(html);
```
SamSaffron: this is missing stuff imo, the method should always accept raw as well, because it needs to make some decisions based off that. 

For example: 

If the HTML converted stuff is somehow 10 chars long and raw is 1000 ... would you consider using the HTML? This interface offers no way to heuristically decide if we even want to do html conversion or not.

SamSaffron: additionally can you make sure that IMG paste if it exists in binary in the clipboard always take precedence over html?

vinothkannans: > skip one of them, the image one. We **should not prioritize image pastes unless there is no
other way** to do this.

but in other [PR](https://github.com/discourse/discourse/pull/5378#issuecomment-347643515) @coding-horror told me that image paste should have low priority.

SamSaffron: @vinothkannans  see our chat, prioritize single image pastes please not hotlinking a single image.

PR #5427, opened 2017-12-13
FIX: caret_position.js creating horizontal overflow in RTL layouts

scossar:
When @mentions are added to posts in a forum using a right-to-left layout, they are creating 7000px of horizontal overflow that persists until the browser is refreshed.

This is the smallest possible change to fix the problem. It adds an id to the clone div that's created in `caret_position.js` and then overrides the div's inline styles in `rtl.scss`.

screen shot 2017-12-13 at 3 31 02 pm


SamSaffron:
@scossar something is real ill with this pull request, closing, can you open a new PR here

PR #5434, opened 2017-12-14
FEATURE: Don't delete email from server on pop3 polling

erseco:
Added an option for avoiding deletion of emails from server, this allows reusing the reply mail account for other uses

### DIFF
--- a/config/locales/server.es.yml
+++ b/config/locales/server.es.yml
@@ -1236,6 +1236,7 @@ es:
     pop3_polling_host: "El host utilizado para hacer polling de e-mails vía POP3."
     pop3_polling_username: "El nombre de usuario de la cuenta POP3 para hacer polling de e-mails."
     pop3_polling_password: "La contraseña de la cuenta POP3 para hacer polling de correos."
+    pop3_polling_delete_from_server: "Eliminar los e-mails del servidor"
     log_mail_processing_failures: "Registra todos los fallos de procesamiento de email a http://yoursitename.com/logs"
     email_in: "Permitir a los usuarios crear nuevos temas por email (requiere el polling pop3). Configura las direcciones en la pestaña \"Ajustes\" de cada categoría."
     email_in_min_trust: "El mínimo nivel de confianza requerido para poder publicar nuevos temas por email."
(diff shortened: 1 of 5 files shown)


config/locales/server.es.yml
```diff
@@ -1236,6 +1236,7 @@ es:
     pop3_polling_host: "El host utilizado para hacer polling de e-mails vía POP3."
     pop3_polling_username: "El nombre de usuario de la cuenta POP3 para hacer polling de e-mails."
     pop3_polling_password: "La contraseña de la cuenta POP3 para hacer polling de correos."
+    pop3_polling_delete_from_server: "Eliminar los e-mails del servidor"
```
ZogStriP: There is no need to provide translations other than English because they will be overwritten next time we pull translations from Transifex.

ZogStriP:
This is 👍. Could you add a test?

erseco:
I tested it locally in a test environment an it works as expected but please tell me if you need a specific test before approve it

Best regards!

ZogStriP:
> please tell me if you need a specific test before approve it

Please add one :)

erseco:
I checked the source and I can't find any TDD system for testing the pop3_polling, can you be more specific about the kind of test that you need? :)

Thank you in advance!

gschlager:
I suggest you add a spec to [poll_mailbox_spec.rb](https://github.com/discourse/discourse/blob/master/spec/jobs/poll_mailbox_spec.rb) where you test that it deletes / does not delete emails depending on the setting.

erseco:
Added a rspec test, please confirm if it's correct

gschlager:
Thanks for the spec. Unfortunately it didn't work. It needed a lot more stubs to work and it's always a good idea to test both cases (it deletes / it does not delete). I committed changes to the specs since I had to try this out myself.

SamSaffron:
note we can stub all we want here and its not going to get us very far unless we know absolutely that the `.delete` works right. 

I want this change, but I am concerned a bit, what happens when there are 20,000 emails in the pop mailbox? are we just going to crash? 

deferring this to 2.0 (which is in a few weeks)

erseco:
I added a **NOTE** in the option to advertise of mailbox cleanup by hand

SamSaffron:
I am merging this thanks for the change and sorry it took so long to merge!

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

PR #5435, opened 2017-12-14
SMF2: don't crash if there are duplicate category names

pfaffman:


script/import_scripts/smf2.rb
```diff
@@ -152,6 +153,9 @@ def import_categories
       parent_id = category_id_from_imported_category_id(board[:id_parent]) if board[:id_parent] > 0
       groups = (board[:member_groups] || "").split(/,/).map(&:to_i)
       restricted = !groups.include?(GUEST_GROUP) && !groups.include?(MEMBER_GROUP)
+      if Category.find_by_name(board[:name])
```
SamSaffron: bit of feedback ... what you mean is `Category.where(name: board[:name]).exists?` that is faster

SamSaffron:
seems fine to me

script/import_scripts/smf2.rb
pfaffman: So noted. I guess that I assumed that if there was a function doing exactly what you wanted that it was a good way to go. :-)

Thanks very much.

PR #5437, opened 2017-12-15
Unicode usernames

jjaffeux:
Initial implementation to support unicode usernames (eg: Анатолий), ATM this PR won‘t work for unicode usernames without a custom username validation logic.

* route for % encoded:
allows router to understand url encoded usernames

* escape username instead of replacing chars:
make sure the username is escaped to avoid xss but is still readable by the backend


SamSaffron:
we got to wait till after the release on this

jjaffeux:
@SamSaffron I think @eviltrout  used part of it in another commit he did few days ago, we can close this one anyways.

safwanrahman:
@jjaffeux Which commit can you please reference? Is it possible to have this in 2.0?

PR #5438, opened 2017-12-16
Proof-of-concept: Data loss on update of single user_field

phaldan:
Tests to proof a changed behaviour of user_field update.
The current update needs a list of all user_fields. When only a single user_field is submitted, the content of all other user_fields will be set to nil (data loss).
In the past it was possible to change the content of a single user_field without side effects to other user_fields.
The changed behaviour is probably introduced on the following commit (Line 102, 105, 115, 121):

https://github.com/discourse/discourse/blob/77d4c4d8dc00491b99aa15dfd78eb22a47ffe663/app/controllers/users_controller.rb#L99-L123

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1533,6 +1533,34 @@ describe UsersController do
 
               expect(user.user_fields[user_field.id.to_s].size).to eq(UserField.max_length)
             end
+
+            it "override required field" do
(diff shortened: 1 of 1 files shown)


tgxworld:
Merged in https://github.com/discourse/discourse/commit/6a2bce19317b05f1eab56f82ced58a248db90d94.

Thank you @phaldan for the test case.

PR #5439, opened 2017-12-16
FIX: Ignore if the default localization node is not an object

vinothkannans:


ZogStriP:
LGTM 👍 Can you add a test?

vinothkannans:
@ZogStriP qunit test case added 👍

ZogStriP:
LGTM :+1:

SamSaffron:
this looks correct to me and safe

PR #5445, opened 2017-12-19
FEATURE: search log term details page

techAPJ:
This PR adds search log term detail page.


SamSaffron:
Merging this in cause I want to see it in action!

PR #5446, opened 2017-12-19
PERF: add option to limit how many categories are processed in one ca…

nlalonde:
…ll to CategoryFeaturedTopic.feature_topics

With many categories, PeriodicalUpdates can take more than 15 minutes to run. Batch the CategoryFeaturedTopic.featured_topics calls.

### DIFF
--- a/app/models/category_featured_topic.rb
+++ b/app/models/category_featured_topic.rb
@@ -2,13 +2,33 @@ class CategoryFeaturedTopic < ActiveRecord::Base
   belongs_to :category
   belongs_to :topic
 
-  # Populates the category featured topics
-  def self.feature_topics
+  NEXT_CATEGORY_ID_KEY = 'category-featured-topic:next-category-id'.freeze
+  BATCH_SIZE = 100
+
+  # Populates the category featured topics.
+  def self.feature_topics(batched: false)
     current = {}
     CategoryFeaturedTopic.select(:topic_id, :category_id).order(:rank).each do |f|
       (current[f.category_id] ||= []) << f.topic_id
     end
-    Category.select(:id, :topic_id, :num_featured_topics).find_each do |c|
+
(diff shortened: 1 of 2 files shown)


app/models/category_featured_topic.rb
```diff
@@ -2,13 +2,33 @@ class CategoryFeaturedTopic < ActiveRecord::Base
...
       (current[f.category_id] ||= []) << f.topic_id
     end
-    Category.select(:id, :topic_id, :num_featured_topics).find_each do |c|
+
+    next_category_id = batched ? ($redis.get(NEXT_CATEGORY_ID_KEY) || 0) : 0
+
+    categories = Category.select(:id, :topic_id, :num_featured_topics)
+      .where('id >= ?', next_category_id)
+      .order('id ASC')
+      .limit(BATCH_SIZE)
```
SamSaffron: I think you probably want a .to_a here to avoid risk of double queries

app/models/category_featured_topic.rb
```diff
@@ -2,13 +2,33 @@ class CategoryFeaturedTopic < ActiveRecord::Base
...
+    if batched
+      if categories.count == BATCH_SIZE
+        next_id = Category.where('id > ?', categories.last.id).order('id asc').limit(1).pluck(:id)[0]
+        next_id ? $redis.setex(NEXT_CATEGORY_ID_KEY, 1.day, next_id) : $redis.del(NEXT_CATEGORY_ID_KEY)
+      else
+        $redis.del(NEXT_CATEGORY_ID_KEY)
+      end
+    end
+
+    categories.find_each do |c|
```
SamSaffron: also then we can avoid the find_each here and just do an each

SamSaffron:
going to merge this and do the little bit of fiddling so we can get the change in quickly

PR #5453, opened 2017-12-21
FIX: pluralization rules for Serbian language

blokovi:
Updated SR pluralization to use 3 keys: one, few, other (as by Transifex)


gschlager:
Thanks! You might want to fix the plural rules used on the server too...
https://github.com/discourse/discourse/blob/master/config/locales/plurals.rb#L95

blokovi:
Thanks gschlade, I fixed that file too.

PR #5455, opened 2017-12-24
Remove wrapper `<span>` from user nav plugin outlet

davidtaylorhq:
As-per https://meta.discourse.org/t/plugin-outlet-incorrectly-handles-multiple-plugins-at-the-same-outlet/32206/24

I did a search of all-the-plugins, and it looks like the only one currently using the user-main-nav outlet is `dl-license-keys`

```
$ find . -name user-main-nav
./plugins/dl-license-keys/assets/javascripts/discourse/templates/connectors/user-main-nav
```


SamSaffron:
This is correct, any other spots where we are messing up here?

PR #5459, opened 2017-12-28
FIX: caret clone div position for RTL locales

scossar:
Currently adding a mention or a category link to the composer when the locale is RTL creates 7000px of horizontal overflow that persists until the page is refreshed.  This PR adds `id='caret-clone'` to the clone div and then overrides the clone div's inline styles in `rtl.scss`.

### DIFF
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -56,3 +56,9 @@ float: left !important;
 .rtl .dashboard-right {
   float: left !important;
 }
+
+// Right and left will be flipped by the r2 gem
+.rtl #caret-clone[style] {
+  right: auto !important;
+  left: -7000px !important;
+}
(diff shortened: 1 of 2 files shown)


vinothkannans:
@scossar your fix is working perfectly on my dev environment. But still I think directly editing the js is better than overriding via css.

https://github.com/discourse/discourse/blob/ec65ae76b946a6fffb27f976503385ea9de06ce1/vendor/assets/javascripts/caret_position.js#L57-L66

I suggest to add below code by replacing above lines

```js
  const isRTL = $("html").css("direction") === "rtl";
  clone.css({
    border: "1px solid black",
    padding: important("padding"),
    resize: important("resize"),
    "max-height": textarea.height() + "px",
    "overflow-y": "auto",
    "word-wrap": "break-word",
    position: "absolute",
    left: isRTL ? "auto" : "-7000px",
    right: isRTL ? "-7000px" : "auto"
  });
```

Yes. It is still similar to your fix.

scossar:
Yes, that's better. Thanks!

scossar:
I've changed it to check for rtl with `$('html').hasClass('rtl')`. In the future, we might want to change the way we are setting the direction on the `html` element. Right now it's being done with the CSS `direction` property. It would be more correct to set it through the `dir` attribute on the html element.

PR #5467, opened 2018-01-03
Add search not operator

kasiabulat:
I added the simple version of search operator "not" (~) for tags. Currently it has the highest precedence.
Link to the issue: https://meta.discourse.org/t/new-search-operator-not/52087

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -458,20 +458,13 @@ class Search
     end
   end
 
-  advanced_filter(/tags?:([a-zA-Z0-9,\-_+!]+)/) do |posts, match|
-    if match.include?('!')
-      search_parts = match.split('!')
-      positive_part = search_parts[0]
-      negative_part = search_parts[1]
+  advanced_filter(/^tags?:(.*)/) do |posts, match|
+    search_tags(posts, match)
+  end
 
-      positive_tags = search_tags(posts, positive_part)
-      negative_tags = search_tags(posts, negative_part)
-
-      positive_tags.where(["posts.id NOT IN (?)",
-                           negative_tags.pluck("id")]).order("id")
-    else
-      search_tags(posts, match)
-    end
+  advanced_filter(/\-tags?:(.*)/) do |posts, match|
+    negative_tags = search_tags(posts, match)
+    posts.where(["posts.id NOT IN (?)", negative_tags.pluck("id")]).order("id")
   end
 
   advanced_filter(/filetypes?:([a-zA-Z0-9,\-_]+)/) do |posts, match|
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -458,9 +458,9 @@ class Search
     end
   end
 
-  advanced_filter(/tags?:([a-zA-Z0-9,\-_+~]+)/) do |posts, match|
-    if match.include?('~')
-      search_parts = match.split('~')
+  advanced_filter(/tags?:([a-zA-Z0-9,\-_+!]+)/) do |posts, match|
+    if match.include?('!')
+      search_parts = match.split('!')
       positive_part = search_parts[0]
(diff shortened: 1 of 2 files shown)


SamSaffron:
from a performance perspective I am a bit concerned here, this could be shuffling around arrays with 100s of thousands of post ids and cause memory blowout. 

also we really need some tests here to demonstrate how it works.

kasiabulat:
I added some tests to illustrate how the operator works, I'll try to optimize the solution.

kasiabulat:
I've checked how to improve it and found the website: https://explainextended.com/2009/09/16/not-in-vs-not-exists-vs-left-join-is-null-postgresql/, where "NOT EXISTS" is said to be better than "NOT IN" as postgresql is converting it to anti-join. However, both solutions are checked to work in about 1sec time for tables with about 100 000 rows. For now, I don't know how to rewrite the query to be using "NOT EXISTS" instead of "NOT IN" as I 'm operating on Active Records, not SQL queries. Should I continue trying, choose a different way somehow or can "NOT IN" be left?

kasiabulat:
I changed the operator symbol as proposed in the discussion on meta.discourse.org.

lib/search.rb
```diff
@@ -458,27 +458,13 @@ def self.advanced_filters
...
-      posts.where("topics.id IN (
-      SELECT tt.topic_id
-      FROM topic_tags tt, tags
-      WHERE tt.tag_id = tags.id
-      GROUP BY tt.topic_id
-      HAVING to_tsvector(#{default_ts_config}, array_to_string(array_agg(tags.name), ' ')) @@ to_tsquery(#{default_ts_config}, ?)
-      )", tags.join('&'))
-    else
-      tags = match.split(",")
+  advanced_filter(/^tags?:(.*)/) do |posts, match|
```
tgxworld: Is there a reason that `([a-zA-Z0-9,\-_+]+)/)` has been changed to `(.*)`?

lib/search.rb
```diff
@@ -496,6 +482,30 @@ def self.advanced_filters
 
   private
 
+    def search_tags(posts, match)
+      return if match.nil?
+
+      if match.include?('+')
+        tags = match.split('+')
+        posts.where("topics.id IN (
+              SELECT tt.topic_id
```
tgxworld: Minor nitpick but lets indent the code like so:
```
posts.where("topics.id IN (
  SELECT DISTINC...
  FROM
  WHERE
")
```

lib/search.rb
```diff
@@ -458,27 +458,13 @@ def self.advanced_filters
...
-      posts.where("topics.id IN (
-      SELECT DISTINCT(tt.topic_id)
-      FROM topic_tags tt, tags
-      WHERE tt.tag_id = tags.id
-      AND tags.name in (?)
-      )", tags)
-    end
+  advanced_filter(/\-tags?:(.*)/) do |posts, match|
+    negative_tags = search_tags(posts, match)
+    posts.where(["posts.id NOT IN (?)", negative_tags.pluck("id")]).order("id")
```
tgxworld: Hmm I'm quite uncomfortable with `negative_tags.pluck("id")` because it would trigger an extra SQL query. I think in `search_tags` we can just decide if it is a negative match or not and use `tags.name NOT IN` instead

lib/search.rb
kasiabulat: OK.

lib/search.rb
kasiabulat: I'll change it back.

tgxworld:
Thank you @kasiabulat :+1: Merged in https://github.com/discourse/discourse/commit/b71cf6d422b43a8b10079978e8649ef5e44ac742

PR #5469, opened 2018-01-04
FIX: Add guard to prevent a primary `UserEmail` from being reassigned.

tgxworld:


gschlager:
This broke a few specs.
Does it work when a staged user signs up and the email address gets moved to the new user?

tgxworld:
@gschlager Oops I override the activerecord error message. 

> Does it work when a staged user signs up and the email address gets moved to the new user?

Hmm is there a spot where we move email addresses? I thought the user would just be marked as staged.

gschlager:
Regarding staged users. Scratch that - I misremembered. There's no new user created, the existing user just gets unstaged. So, there shouldn't be a problem in that case.

PR #5473, opened 2018-01-05
FIX: make group "add username" placeholder text available to non-admins

chrisbeach:
Also add some CSS classes to the group editor page to allow CSS custimisation


SamSaffron:
This has grown a bit too much from the scope of the original fix, I am so so sorry, but I am closing this, assigned https://meta.discourse.org/t/text-glitch-on-group-admin-page/77303 to @tgxworld  who will organise and test this. (pretty sure there are a bunch of edge cases left here)

PR #5478, opened 2018-01-09
Support ruby 2.5.0

jsuchal:
Support for ruby 2.5.0

The BOM encoding behaviour somehow changed between ruby releases, so I had to change the logic of opening the csv file for import.


techAPJ:
Awesome, thank you! ❤️

PR #5480, opened 2018-01-10
Add dotenv-rails gem to have .env loaded in rake tasks

Dorian:
Otherwise I couldn't get the `GlobalSetting.developer_emails` to have the value from `.env` loaded when doing a `rake admin:create`.

If there is some sort of `require 'foreman/dotenv'` that exists, I think it could be placed in `Rakefile` to have the same result.


SamSaffron:
Sorry, I am not comfortable adding this dependency, especially not to all environments. 

Can you do a dev topic on meta.discourse.org to discuss?

PR #5487, opened 2018-01-11
FIX: GitHub auth always asking to verify email for new users

vinothkannans:


Supermathie:
I remember back in the day (i.e. YEARS ago) github would send unverified emails as verified, presumably this has been fixed?

vinothkannans:
https://meta.discourse.org/t/should-github-authentication-send-activation-email/17568

As per this topic I hope GitHub fixed their side of problem.

PR #5488, opened 2018-01-11
Add prompt and HD settings to the Google OAuth2 plugin.

gchallen:
Discussed here: https://meta.discourse.org/t/improve-google-login-with-multiple-accounts/75964.

As suggested, this patch adds the prompt and hd options supported by the omniauth-google-oauth2 Gem to the Discourse settings menu.

I'm not great at Ruby, so this patch could still use a bit of love. Specifically:

* It would be great to validate that the prompt setting was an allowed value, specifically none, consent, or select_account
* Translations are needed

If someone could help drag this one across the finish line, I'd appreciate it. Thanks in advance! This is really going to help us with our installation.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -247,6 +247,12 @@ login:
     default: false
   google_oauth2_client_id: ''
   google_oauth2_client_secret: ''
+  google_oauth2_prompt:
+    client: true
+    default: 'none'
+  google_oauth2_hd:
+    client: true
+    default: ''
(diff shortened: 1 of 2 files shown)


gchallen:
CLA signed.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/f74d6bb605f0395f4cba5e69d3c32206ca7c39a8. 

Thank you @gchallen

PR #5490, opened 2018-01-12
FIX: Top offset value of #main-outlet element is always 0

vinothkannans:
I believe behavior of offset method changed in jQuery and it is not respecting padding anymore. So now I am getting offset value for first child element of `#main-outlet`.


SamSaffron:
should we not just merge this? what did we decide?

vinothkannans:
@sam as per others opinion in [dev topic](https://dev.discourse.org/t/3043) I guess we like the current header. @coding-horror closed that topic. It means he don't like to get back the old header. In that case we have to fix the bug in other way and this PR can be closed.

It is not affecting end users and no one complained. So we have time to decide and fix in either way.

coding-horror:
I didn't understand that topic -- I don't see any header behavior
difference.

tgxworld:
>  and this PR can be closed. 

Closing as per ^

PR #5491, opened 2018-01-12
Lock down gem versions so bundle update can be run

OriPekelman:
Currently, if you try to run `bundle update` the resulting Gemset does not allow to deploy discourse.

I believe the Gemfile should always be locked down to a functional set.

Locking down ember-source and pg does the trick.

See: 
https://meta.discourse.org/t/error-on-fresh-install-miniracer-runtimeerror-referenceerror-settimeout-is-not-defined/77663

Otherwise you get 

```MiniRacer::RuntimeError: ReferenceError: setTimeout is not defined
JavaScript at <anonymous>:4927:23
JavaScript at internalRequire (<anonymous>:60:16)
JavaScript at internalRequire (<anonymous>:56:24)```

and 

```rake aborted!
Gem::LoadError: Specified 'postgresql' for database adapter, but the gem is not loaded. Add `gem 'pg'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).```

I did not update the lockfile in this PR so it should result in no changes. Possibly there are higher versions that can be used. I chose the ones from the current Gemfile.lock


SamSaffron:
I am ok with the gemfile changes provided they are commented, but updating everything in the lock file is a massive change we need to work through slowly, can you re-submit just with the Gemfile changes?

OriPekelman:
Hi,

Just the Gemfile changes make Travis sad... but with pleasure. I'll revert the last commit.

O.

PR #5496, opened 2018-01-13
FIX: graphs should go to zero for missing dates

techAPJ:


app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -142,14 +143,13 @@ Report.reopenClass({
         group_id: groupId
       }
     }).then(json => {
-      // Add a percent field to each tuple
```
techAPJ: removed `percentage` because it wasn't being used anywhere.

PR #5497, opened 2018-01-13
Fix: RTL topic-admin-menu position

scossar:
Checks if an RTL locale is being used to set the inline styles for the topic-admin-menu position. Fixes RTL css for `#topic-progress-wrapper .topic-admin-popup-menu.right-side`.

Before:
screen shot 2018-01-13 at 10 49 16 am
After:
screen shot 2018-01-13 at 10 59 08 am

Before (`.topic-admin-popup-menu.right-side`):
screen shot 2018-01-13 at 10 51 08 am
After:
screen shot 2018-01-13 at 10 59 30 am


SamSaffron:
thanks!

PR #5498, opened 2018-01-13
FEATURE: grant badges in post admin wrench

xrav3nz:
> CONTEXT & DISCUSSION: https://meta.discourse.org/t/feature-proposal-improved-badge-granting-workflow/58631/3

This PR adds a "Grant Badge" action in the post admin wrench:

Which then opens a modal for admins to grant a specific badge:

### DIFF
--- a/app/assets/javascripts/discourse/controllers/grant-badge.js.es6
+++ b/app/assets/javascripts/discourse/controllers/grant-badge.js.es6
@@ -1,10 +1,11 @@
 import computed from "ember-addons/ember-computed-decorators";
 import { extractError } from 'discourse/lib/ajax-error';
 import ModalFunctionality from "discourse/mixins/modal-functionality";
+import GrantBadgeController from "discourse/mixins/grant-badge-controller";
 import Badge from 'discourse/models/badge';
 import UserBadge from 'discourse/models/user-badge';
 
-export default Ember.Controller.extend(ModalFunctionality, {
+export default Ember.Controller.extend(ModalFunctionality, GrantBadgeController, {
   topicController: Ember.inject.controller("topic"),
   loading: true,
   saving: false,
@@ -23,27 +24,6 @@ export default Ember.Controller.extend(ModalFunctionality, {
     const protocolAndHost = window.location.protocol + '//' + window.location.host;
 
     return url.indexOf('/') === 0 ? protocolAndHost + url : url;
-  },
-
-  @computed('allBadges.[]', 'userBadges.[]')
-  grantableBadges(allBadges, userBadges) {
-    const granted = userBadges.reduce((map, badge) => {
-      map[badge.get('badge_id')] = true;
-      return map;
-    }, {});
-
-    return allBadges.filter(badge => {
-      return badge.get('enabled')
-        && badge.get('manually_grantable')
-        && (!granted[badge.get('id')] || badge.get('multiple_grant'));
-    });
-  },
-
-  noGrantableBadges: Ember.computed.empty('grantableBadges'),
-
-  @computed('selectedBadgeId', 'grantableBadges')
-  selectedBadgeGrantable(selectedBadgeId, grantableBadges) {
-    return grantableBadges && grantableBadges.find(badge => badge.get('id') === selectedBadgeId);
   },
 
   @computed("saving", "selectedBadgeGrantable")
@@ -66,17 +46,13 @@ export default Ember.Controller.extend(ModalFunctionality, {
 
   actions: {
     grantBadge() {
-      const username = this.get('post.username');
-
       this.set('saving', true);
 
-      UserBadge.grant(this.get('selectedBadgeId'), username, this.get('badgeReason'))
+      this.grantBadge(this.get('selectedBadgeId'), this.get('post.username'), this.get('badgeReason'))
         .then(newBadge => {
-          this.get('userBadges').pushObject(newBadge);
           this.set('selectedBadgeId', null);
           this.flash(I18n.t(
-            'badges.successfully_granted',
-            { username: username, badge: newBadge.get('badge.name') }
+            'badges.successfully_granted', { username: this.get('post.username'), badge: newBadge.get('badge.name') }
           ), 'success');
         }, error => {
           this.flash(extractError(error), 'error');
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/discourse/routes/topic.js.es6
```diff
@@ -85,6 +85,10 @@ const TopicRoute = Discourse.Route.extend({
       this.controllerFor('modal').set('modalClass', 'history-modal');
     },
 
+    showGrantBadgeModal() {
+      showModal('grant-badge', { model: this.modelFor('topic'), title: 'admin.badges.grant_badge' });
```
xrav3nz: I recycled and used translations from `admin`, let me know if this is not okay.

config/locales/client.en.yml
```diff
@@ -2483,6 +2483,7 @@ en:
         other: "%{count} granted"
       select_badge_for_title: Select a badge to use as your title
       none: "(none)"
+      successfully_granted: "Successfully granted %{badge} to %{username}"
```
xrav3nz: This is the only new translation I added. Open to copy changes if needed.

app/assets/javascripts/discourse/routes/topic.js.es6
gschlager: I'd say it's fine to reuse them. It's the same functionality.
The only place I'd probably use a new translation is the action in the wrench menu. There are size constraints and I imagine some language could have a need for using a shorter label...

SamSaffron:
hmm... is there a way to reuse more code with the admin implementation?

app/models/badge.rb
```diff
@@ -219,6 +219,10 @@ def slug
     Slug.for(self.display_name, '-')
   end
 
+  def manually_grantable?
+    query.blank? && !system?
```
xrav3nz: This definition will need some 👀

xrav3nz:
> is there a way to reuse more code with the admin implementation?

Definitely not an Ember expert, but I am leaning towards a "no", since _most_ of it are view specific.

jorgemanrubia:
Hey so coincidentally I was working on this one too. I was close to create a PR when I saw this. [Here are my changes](https://github.com/discourse/discourse/compare/master...jorgemanrubia:master)

One thing I did was to extract a static method `calculateGrantableBadges` to the `Badge` model, to reuse it in both admin/topic pages. That part and the test should be easy to add to this PR if we want to.

app/assets/javascripts/discourse/controllers/grant-badge.js.es6
```diff
@@ -0,0 +1,87 @@
...
+  userBadges: [],
+
+  @computed('topicController.selectedPosts')
+  post() {
+    return this.get('topicController.selectedPosts')[0];
+  },
+
+  @computed('post')
+  badgeReason(post) {
+    const url = post.get('url');
```
jorgemanrubia: This can be simplified since there is no need to add protocol and host to the url. 

In the server, the url is only used to extract the post and it will work as it is (it is only interested in the path).

jorgemanrubia: Another change in my PR is that I was closing the dialog automatically right after granting the badge (unless there was an error). Instead of the flash message, I was showing a spinner while in the  "saving" state, and closing it right after, like in the "set topic timer" form. Do we want to keep it open after clicking "Grant"? If we don't, you can get rid of some more code there too.

In any case, good job here @xrav3nz 👏, I was just pointing out some differences in my PR in case we can use some of it before it's lost, but yours looks 👌.

xrav3nz: Thanks ❤️ 

> In the server, the url is only used to extract the post and it will work as it is

Interestingly enough, even if only given a path, `URI.parse` does not raise an exception and produces a `URI::Generic` object. But since the API expects a full URL, I think it's better to send in a full URL in case of backend implementation changes?

>  closing the dialog automatically right after granting the badge

I was a bit torn on how the modal should behave too! 

IMHO closing the modal after setting a timer / changing owner is fine because the post itself will reflect the changes, on other hand, closing the modal after granting badges _may_ lead to confusions.

But I do agree that the modal does not necessarily need to stay open after a successful grant, other than for providing visual feedback.

SamSaffron:
I would like to get this merged in tomorrow, @jorgemanrubia do you feel it is good to go? 

I am fine to merge this in as is and then do another PR or two to refine it if needed, it is admin only functionality so its a lot less risky

jorgemanrubia:
Of my comments above, the only change I'd do is extracting the logic for `grantableBadges` which is duplicated in both controllers and only tested on the admin side.

I extracted that logic to a [static method in the model](https://github.com/discourse/discourse/compare/master...jorgemanrubia:master#diff-6a401d27f67ce2d6c99bf2dba6453f6fR166), [wrote a test for it](https://github.com/discourse/discourse/compare/master...jorgemanrubia:master#diff-582811f2fb793a1db73cf0570ca2ae09R53) and called it in [both](https://github.com/discourse/discourse/compare/master...jorgemanrubia:master#diff-150c961c1d2a67bcfdf4c3e4b0c5aaf1R46) [places](https://github.com/discourse/discourse/compare/master...jorgemanrubia:master#diff-150c961c1d2a67bcfdf4c3e4b0c5aaf1R46).

We could do that, or extract it to a [service](https://guides.emberjs.com/v2.1.0/applications/services/) or a mixin to be included in both controllers, but I'd definitely refactor it. The logic represents a domain concept (a grantable badge) and the rules that define it: being `enabled` and `manually_grantable`, not being in use or being `multiple_grant`, etc. It is good to have that defined in a single place.

I think the PR is great otherwise 👍🏻

xrav3nz:
@jorgemanrubia must have missed your first comment above. I will look at this again later today. Thanks for the tips on service/mixin, they seems more Ember-y than simply a static method.

xrav3nz:
👋 Just a quick update, I am in the process of extracting the common logic into a mixin. I am going away for the weekend but will try to finish this up by Monday.

xrav3nz:
This PR got a bit massive (sorry!), but the commits are self-contained and should hopefully help with the review. 

The latest change extracted the common logic between the `admin` and `post wrench` badge granting flow into a mixin (with a test for the mixin).

SamSaffron:
OK let's try this out, one general concern I have here that extends to a bunch of admin functions we carry is that we are growing application.js with code that only admins need. Will do a topic about this on meta so ti is not lost. 

Thanks @xrav3nz

PR #5500, opened 2018-01-14
Correct Arabic month names

mtawil:
Correct Arabic month names. It's not right to concatenate month names from two different calendars [Assyrian calendar](https://en.wikipedia.org/wiki/Assyrian_calendar) and [Gregorian calendar](https://en.wikipedia.org/wiki/Gregorian_calendar)


gschlager:
LGTM. Current version of that file in moment.js [uses the same names](https://github.com/moment/moment/blob/develop/locale/ar.js#L61-L72).
We should probably update moment.js when we upgrade JS dependencies for the 2.0 release.

PR #5505, opened 2018-01-16
Accessibility: focus management in topics

riking:
This work is a prerequisite to pointing screen readers at the right spot when you enter in the middle of a topic.

 - Applied hover styling to post actions when focused - this way, you can actually see that the button is tabbed to instead of being entirely invisible.
 - Made topic posts, and topic list items, valid tab targets.
    - Removed the `<a class=tabLoc>` at the start of every post, because the post itself is now a valid target.
 - Made the J/K keyboard navigation focus each post / topic list item, and use focus as the first test for what's currently selected.
 - The obvious complement to that is to use the existing `.selected` styling for when the element is focused. *Edge does not support the `:focus-within` selector*, so we need to use JS to track this.
 - A final cleanup commit to set `outline: 0` on our newly focusable elements, since they have custom styling.


tgxworld:
Oops reopening this. The keyboard selector was appearing when clicking on topics in.

SamSaffron:
sorry, closing this, its going to need to be re-done and very carefully tested, this was a bit of a quick hack kind of thing.

PR #5506, opened 2018-01-16
FEATURE: Keyboard shortcut to refocus composer

riking:
https://meta.discourse.org/t/accessibility-focus-management-in-topics/78109/3?u=riking

It was very annoying to get back to the composer after losing it!

Added to the keyboard shortcuts help dialog, and reshuffled it a bit. Translation keys were intentionally kept so our translators don't have to copy paste to update.

The setTimeout(0) is necessary because otherwise, the `C` actually gets inserted!

This commit also fixes an incidental listener leak.

image


SamSaffron:
looks good to me, should be `shift + c` not `shift + C` to match current translations... will fix

PR #5507, opened 2018-01-16
UX: Implementing a z-index system

awesomerobot:
This creates a z-index function that looks for values in a SASS map. 

This gives us a quick at-a-glance reference to where most elements sit in the z-index, and limits our values to some extent.


eviltrout:
This is very cool 👍

PR #5508, opened 2018-01-16
remove superfluous 'the' from translation file

pacharanero:
remove superfluous 'the' from translation file

Original text: 'Use the HTML instead of the text for incoming email.' sounds odd for native English speakers.
 
I propose the slight modification 'Use HTML instead of text for incoming email.'


SamSaffron:
thanks!

PR #5510, opened 2018-01-17
Feature: add topic posts count to webhook post serializer

scossar:
This change would make it possible for the WordPress plugin to get the correct number of Discourse comments on a topic when posts have been moved or deleted from the topic.

### DIFF
--- a/app/serializers/web_hook_post_serializer.rb
+++ b/app/serializers/web_hook_post_serializer.rb
@@ -1,4 +1,7 @@
 class WebHookPostSerializer < PostSerializer
+
+  attributes :topic_posts_count
+
   def include_topic_title?
     true
   end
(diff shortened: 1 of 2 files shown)


app/serializers/web_hook_post_serializer.rb
```diff
@@ -14,4 +17,9 @@ def include_topic_title?
       false
     end
   end
+
+  def topic_posts_count
+    object.topic && object.topic.posts_count
```
tgxworld: I'm thinking that it might be better to return `0` here when `object.topic` is `nil`.

scossar: Sure, that makes sense.

tgxworld:
Looks good to me :+1:

PR #5516, opened 2018-01-22
enable New Topic button when user can post to subcategory

pmusaraj:
See https://meta.discourse.org/t/users-mistake-top-level-categories-for-lack-of-permissions/72227/20


SamSaffron:
This is not going to work cause its introducing another ajax call.

SamSaffron:
sorry, closing this per: 

https://meta.discourse.org/t/users-mistake-top-level-categories-for-lack-of-permissions/72227/25?u=sam

The extra ajax call is a no go

PR #5517, opened 2018-01-22
FEATURE: Allow plugins to register a new locale (and some bugfixes)

gschlager:
Allows plugins to register new locales. It can either use an existing locale as fallback or the plugin must ship all required locale files. Discourse validates the locale during validation, so it doesn't register a locale when required files are missing.

The following example adds
* `es_MX` which uses `es` as fallback. The plugin must only include `assets/locales/es_MX.js.erb` and the locale files in `config/locales`.
* `foo` which is a new locale. The plugin must ship all required files including locale files for `MessageFormat` and `Moment.js`. Also, it must register plural rules.

```ruby
register_locale("es_MX", name: "Spanish (Mexico)", nativeName: "Español (México)", fallbackLocale: "es")

register_locale("foo",
                name: "Foo",
                nativeName: "Foo Bar",
                plural: {
                  keys: [:one, :other],
                  rule: lambda { |n| n == 1 ? :one : :other }
                })
```

Directory structure of the example plugin:
```
custom-locales
├── assets
│   └── locales
│       ├── es_MX.js.erb
│       └── foo.js.erb
├── config
│   └── locales
│       ├── client.es_MX.yml
│       ├── client.foo.yml
│       ├── server.es_MX.yml
│       └── server.foo.yml
├── lib
│   └── javascripts
│       └── locale
│           ├── message_format
│           │   └── foo.js
│           └── moment_js
│               └── foo.js
└── plugin.rb
```

---

This PR also contains some bugfixes:
* The application shouldn't crash when a selected locale doesn't exist anymore.
* The server didn't use `SiteSetting.default_locale` as fallback. Only the client did.
* English pluralization rules were loaded for some languages, but it didn't matter because the affected languages have the same rules.


app/models/locale_site_setting.rb
```diff
@@ -19,43 +22,35 @@ def self.language_names
     return @language_names if @language_names
 
     @lock.synchronize do
-      @language_names ||= YAML.load(File.read(File.join(Rails.root, 'config', 'locales', 'names.yml')))
+      @language_names ||= begin
+        names = YAML.load(File.read(File.join(Rails.root, 'config', 'locales', 'names.yml')))
+
+        DiscoursePluginRegistry.locales.each do |locale, options|
+          if !names.key?(locale) && options[:name] && options[:nativeName]
+            names[locale] = { "name" => options[:name], "nativeName" => options[:nativeName] }
```
eviltrout: Is there a reason you're using string keys rather than symbols here?

gschlager: The `names` are loaded with `YAML.load` which also uses string keys. I didn't want to break anything, so I continued using string keys.

PR #5518, opened 2018-01-22
Fix notifications of staged users after sign-up

gschlager:
* Deletes old notifications when a staged users signs up. Otherwise their can be quite a lot of notifications and when the user change the username during sign-up, none of the notification links will work.

* Makes sure that discobot messages staged users after they sign up.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -235,6 +235,8 @@ class User < ActiveRecord::Base
       user.staged = false
       user.active = false
       user.custom_fields[FROM_STAGED] = true
+
+      DiscourseEvent.trigger(:user_unstaged, user)
     end
     user
   end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -383,7 +386,7 @@ def saw_notification_id(notification_id)
 
   def read_first_notification?
     if (trust_level > TrustLevel[1] ||
-        created_at < TRACK_FIRST_NOTIFICATION_READ_DURATION.seconds.ago)
+        first_seen_at < TRACK_FIRST_NOTIFICATION_READ_DURATION.seconds.ago)
```
gschlager: @tgxworld Do you have any objections to using `first_seen_at` instead of `created_at`? It shouldn't make a difference for regular users, but it's the easiest way to make it work for previously staged users.

PR #5520, opened 2018-01-22
FIX: Only enqueue topic_edited hook when a post has a topic

scossar:
When the first post in a topic has been deleted and is then updated through the API, using an admin API key, it currently returns a 500 error: `undefined method 'id' for nil:NilClass` called from `app/models/web_hook.rb:48`.

This PR adds a check to make sure the post has an associated topic before enqueueing the topic_edited webhook.

### DIFF
--- a/config/initializers/012-web_hook_events.rb
+++ b/config/initializers/012-web_hook_events.rb
@@ -19,7 +19,7 @@ end
 
 DiscourseEvent.on(:post_edited) do |post, topic_changed|
   WebHook.enqueue_post_hooks(:post_edited, post)
-  WebHook.enqueue_topic_hooks(:topic_edited, post.topic) if post.is_first_post? && topic_changed
+  WebHook.enqueue_topic_hooks(:topic_edited, post.topic) if post.is_first_post? && topic_changed && post.topic
 end
 
 %i(


ZogStriP:
This looks great 👍 Would you mind adding a test too?

config/initializers/012-web_hook_events.rb
```diff
@@ -19,7 +19,7 @@
 
 DiscourseEvent.on(:post_edited) do |post, topic_changed|
   WebHook.enqueue_post_hooks(:post_edited, post)
-  WebHook.enqueue_topic_hooks(:topic_edited, post.topic) if post.is_first_post? && topic_changed
+  WebHook.enqueue_topic_hooks(:topic_edited, post.topic) if post.is_first_post? && topic_changed && post.topic
```
SamSaffron: hmm... same goes for the post_edited hook

it should also check for post.topic, I would not trigger any events if the topic is not there.

scossar:
I'm not sure how to add a test for this in `models/web_hook_spec.rb`. To cause the error the topic needs to be deleted and `topic_changed` needs to be set to `true`.  This happens when a `:title` parameter is passed to `PostsController#update` for a deleted topic.

> it should also check for post.topic, I would not trigger any events if the topic is not there.

`WebHook#enqueue_post_hooks` tries to access the topic with `&.topic`, so no error is being thrown if the topic isn't there.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/9d0807224bc8df425241d33a1ce26171d7c887a8 with tests. Thanks for fixing this @scossar

PR #5521, opened 2018-01-23
FEATURE: Implement 2factor login TOTP

featheredtoast:
https://meta.discourse.org/t/two-factor-local-login-proposal/76689/1

TOTP support only. Prompts for a second factor ("2-step authentication") in the following areas:

- Local Login
- Admin Login
- Password Reset
- Email Change


app/controllers/users_controller.rb
```diff
@@ -146,6 +146,29 @@ def username
     end
   end
 
+  def toggle_second_factor
```
SamSaffron: I am not a fan of "toggle" actions that rely on state, I also feel user controller is getting quite big, perhaps a second_factor controller with an update method and then caller tells it if it wants it enable or disabled.

app/controllers/users_controller.rb
```diff
@@ -146,6 +146,29 @@ def username
...
+        :critical_user_email,
+        type: :account_second_factor_disabled,
+        user_id: user.id
+      )
+    else
+      user.create_totp
+      user.user_second_factor.enabled = true
+      user.user_second_factor.save!
+    end
+    render json: {}
```
SamSaffron: this either worked or not ... it should probably send through json_success like other spots do

app/controllers/session_controller.rb
```diff
@@ -233,6 +238,14 @@ def create
       return admin_not_allowed_from_ip_address(user)
     end
 
+    if user.totp_enabled?
+      unless user.authenticate_otp(params[:second_factor_token])
+        secure_session["second-factor-#{request.remote_ip}"] = "true"
```
SamSaffron: this feels risky cause ip can change between requests especially in corporates behind a big NAT

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -304,6 +304,23 @@ const User = RestModel.extend({
     });
   },
 
+  loadSecondFactorCodes(password) {
+    return ajax("/session/second-factor", {
```
SamSaffron: maybe this also goes into the second factor controller

SamSaffron:
hmm the name user_second_factor is a little bit strange to me, not sure ... maybe we should go with `user_totp` if this is intended just for totp? when we implement say fido is it going to go into user_second_factors?

SamSaffron:
also can you make sure that there is a test covering every new controller action you added?

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -41,6 +41,16 @@ export default Ember.Controller.extend(CanCheckEmails, {
     return userPath(`${username}/preferences`);
   },
 
+  @computed('model.second_factor_enabled')
+  show_second_factor(second_factor_enabled) {
```
eviltrout: In Javascript we use camelCase style for object properties. This should be `showSecondFactor` and `displaySecondFactorEnabled`.

Having said that, `showSecondFactor` is basically an alias to a model property.  It wouldn't be bad to just have `{{#if model.second_factor_enabled}}` in your template.

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -41,6 +41,16 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+    return second_factor_enabled != null;
+  },
+
+  @computed('model.second_factor_enabled')
+  display_second_factor_enabled(second_factor_enabled) {
```
eviltrout: I feel like this might be better put in the template, that way you don't have to disable escaping.

```hbs
{{#if model.second_factor_enabled}}
  {{i18n "yes_value"}}
{{else}}
  {{i18n "no_value"}}
{{/if}}
```

app/assets/javascripts/admin/models/admin-user.js.es6
```diff
@@ -168,6 +168,17 @@ const AdminUser = Discourse.User.extend({
     }).catch(popupAjaxError);
   },
 
+  disableSecondFactor() {
+    const self = this;
```
eviltrout: We are migrating away from `self = this` as a pattern. Instead use ES6 arrow functions so that `this` points at the correct thing.

app/assets/javascripts/admin/models/admin-user.js.es6
```diff
@@ -168,6 +168,17 @@ const AdminUser = Discourse.User.extend({
...
+    return ajax("/admin/users/" + this.get('id') + "/disable_second_factor", {
+      type: 'PUT'
+    }).then(function() {
+      self.setProperties({
```
eviltrout: If you're only updating one property it's preferable to use `this.set('second_factor_enabled', true)` rather than `setProperties`

app/assets/javascripts/discourse/controllers/login.js.es6
```diff
@@ -23,14 +24,23 @@ export default Ember.Controller.extend(ModalFunctionality, {
   authenticate: null,
   loggingIn: false,
   loggedIn: false,
+  secondFactorRequired: false,
 
   canLoginLocal: setting('enable_local_logins'),
   loginRequired: Em.computed.alias('application.loginRequired'),
 
+  @observes('secondFactorRequired')
```
eviltrout: We try to avoid observers wherever possible.

In this case, you know exactly when you are updating `secondFactorRequired` - you could just call this function when you do! Or even have a function `updateSecondFactorRequired(required)` that would set it too.

app/assets/javascripts/discourse/controllers/password-reset.js.es6
```diff
@@ -45,7 +47,17 @@ export default Ember.Controller.extend(PasswordValidation, {
             DiscourseURL.redirectTo(result.redirect_to || '/');
           }
         } else {
-          if (result.errors && result.errors.password && result.errors.password.length > 0) {
+          if (result.errors && result.errors.second_factor) {
+            this.set('secondFactorRequired',true);
+            this.set('password', null);
+            this.set('errorMessage', 'Invalid token');
```
eviltrout: This needs to be translated. You should also use `setProperties` here as you are setting 3 in a row.

app/assets/javascripts/discourse/controllers/password-reset.js.es6
```diff
@@ -45,7 +47,17 @@ export default Ember.Controller.extend(PasswordValidation, {
...
+          }
+          else if (this.get('secondFactorRequired') === true) {
```
eviltrout: Minor but `this.get('secondFactorRequired')` should be sufficient here without `=== true`

app/assets/javascripts/discourse/controllers/preferences/second-factor.js.es6
```diff
@@ -0,0 +1,66 @@
+import { default as computed } from 'ember-addons/ember-computed-decorators';
+import DiscourseURL from 'discourse/lib/url';
+import { userPath } from 'discourse/lib/url';
+import { popupAjaxError } from 'discourse/lib/ajax-error';
+
+export default Ember.Controller.extend({
+
+  loading: false,
+  password: null,
+  second_factor_image: null,
```
eviltrout: These properties should be camelCase too.

app/assets/javascripts/discourse/controllers/preferences/second-factor.js.es6
```diff
@@ -0,0 +1,66 @@
...
+  newUsername: null,
+
+  @computed('second_factor_image','second_factor_key')
+  loaded(second_factor_image, second_factor_key) {
+    return second_factor_image && second_factor_key;
+  },
+
+  @computed('loading')
+  submitButtonText(loading) {
+    if (loading) return "Loading";
```
eviltrout: This text needs to be translated via I18n

app/assets/javascripts/discourse/controllers/preferences/second-factor.js.es6
```diff
@@ -0,0 +1,66 @@
...
+      this.get('content').loadSecondFactorCodes(this.get('password')).then((resp) => {
+        if(resp.error) {
+          this.set('errorMessage',resp.error);
+          return;
+        }
+        this.set('errorMessage',null);
+        this.set('second_factor_key', resp.key);
+        this.set('second_factor_image', resp.qr);
+      })
+        .catch(popupAjaxError)
```
eviltrout: This indentation is odd here. I'd prefer if `.catch` was on the previous line following the paren.

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -304,6 +304,23 @@ const User = RestModel.extend({
...
+      data: { login: this.get('username'),
+              password: password},
+      type: 'POST'
+    });
+  },
+
+  toggleSecondFactor(token) {
+    return ajax(userPath(`${this.get('username_lower')}/preferences/second-factor`), {
+      dataType: 'json',
+      data: { token: token },
```
eviltrout: Minor but in ES6 this can just be ` { token } `

app/assets/javascripts/discourse/routes/preferences-second-factor.js.es6
```diff
@@ -0,0 +1,21 @@
+import RestrictedUserRoute from "discourse/routes/restricted-user";
+
+export default RestrictedUserRoute.extend({
+  model: function() {
```
eviltrout: instead of `model: function()` you should use the new syntax of `model()`, ditto for the other functions in this class.

app/assets/javascripts/discourse/templates/modal/login.hbs
```diff
@@ -32,6 +32,21 @@
             </tr>
           </table>
         </div>
+        <div id="second-factor" style="display: none;">
```
eviltrout: This part of the template seems to be duplicated. You should use an ember component to avoid repeating yourself.

app/assets/javascripts/discourse/templates/password-reset.hbs
```diff
@@ -16,20 +16,28 @@
...
-        </div>
-
-        <button class='btn btn-primary' {{action "submit"}}>{{i18n 'user.change_password.set_password'}}</button>
+        {{#if secondFactorRequired}}
+          <h2>{{i18n 'login.second_factor_title'}}</h2>
+          <p>{{i18n 'login.second_factor_description'}}</p>
+          <div class="input">
+            {{input value=secondFactor id="second-factor" autofocus="autofocus"}}
+          </div>
+          <button class='btn btn-primary' {{action "submit"}}>Submit</button>
```
eviltrout: This text is not translated. Also this should probably be a `{{d-button}}`

app/controllers/session_controller.rb
```diff
@@ -271,6 +284,22 @@ def current
     end
   end
 
+  def second_factor
+    RateLimiter.new(nil, "login-hr-#{request.remote_ip}", SiteSetting.max_logins_per_ip_per_hour, 1.hour).performed!
+    RateLimiter.new(nil, "login-min-#{request.remote_ip}", SiteSetting.max_logins_per_ip_per_minute, 1.minute).performed!
+    if user = User.find_by_username_or_email(params[:login])
+      unless user.confirm_password?(params[:password])
+        invalid_credentials
```
eviltrout: Minor but why not `return invalid_credentials`? One line shorter and reads better

app/controllers/session_controller.rb
```diff
@@ -271,6 +284,22 @@ def current
...
+        return
+      end
+      qrcode = RQRCode::QRCode.new(user.provisioning_uri)
+      qrcode_svg = qrcode.as_svg(offset: 0, color: '000',
```
eviltrout: Formatting is a bit weird here. This line has two keys and the others have one. One style I enjoy is like this:

```ruby
some_method(
  offset: 0,
  color: '000'
)
```

That way if the method name changes or you assign it to a variable you don't need to change the indentation of all following lines which pollutes the git history.

app/models/user.rb
```diff
@@ -1002,6 +1007,33 @@ def from_staged?
     custom_fields[User::FROM_STAGED]
   end
 
+  def totp
```
eviltrout: Our `User` class is already quite big. At the very least I think we might want to put this in a mixin and include it.

We could also consider creating a helper class like `UserTotp.new(user)` then call `create`, `provision_url`, etc on it.

app/models/user_second_factor.rb
```diff
@@ -0,0 +1,3 @@
+class UserSecondFactor < ActiveRecord::Base
+  belongs_to :user, class_name: 'User'
```
eviltrout: `class_name` should not be necesary here.

featheredtoast:
Thank you both for the reviews! I'll start working on the recommendations for reactors and add in the missing tests. Wanted to get the working prototype and architecture approved before building out test expectations too much.

app/controllers/session_controller.rb
featheredtoast: The only part where this is used is to identify which rate limiter to use -- the "login" one or the "second factor token" one.

An alternative to using this method is to detect whether the second_factor_token parameter is in the request. It could trivially give attackers a few more attempts against a different rate limiter, but it would still not allow a crazy amout of brute forcing.

SamSaffron:
Cool @featheredtoast ... where are we now, how close to having this mergable?

featheredtoast:
Hi Sam, sorry if it wasn't clear from last week's update - I consider this
mergable, pending any other feedback the team might have.

I rebased last week, but can do one more to get things up to date if you
are ready for it now.

featheredtoast:
Just rebased/squashed and added 2-factor protections against the login-over-email page. My own sanity checking also confirms everything still in working order - should be good to go now!

tgxworld:
Thank you @featheredtoast for the work! Merged in https://github.com/discourse/discourse/pull/5612

PR #5522, opened 2018-01-23
FIX: mods should be able to grant badges in post wrench

xrav3nz:
> CONTEXT: https://meta.discourse.org/t/feature-proposal-improved-badge-granting-workflow/58631/32?u=xrav3nz

A follow-up to https://github.com/discourse/discourse/pull/5498; moderators should be able to grant badges in post wrench too:

###### and plz do hate me for not having a test, considered it but didn't...

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-admin-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-admin-menu.js.es6
@@ -64,7 +64,9 @@ export function buildManageButtons(attrs, currentUser) {
       action: 'changePostOwner',
       className: 'change-owner'
     });
+  }
 
+  if (currentUser.staff) {
     contents.push({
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/widgets/post-admin-menu.js.es6
```diff
@@ -64,7 +64,9 @@ export function buildManageButtons(attrs, currentUser) {
       action: 'changePostOwner',
       className: 'change-owner'
     });
+  }
 
+  if (currentUser.staff) {
```
xrav3nz: Didn't add this to the above `if currentUser.staff` block because I do not want the position of this button to change.

SamSaffron:
thanks!

PR #5531, opened 2018-01-25
Lock down gem versions so bundle update can be run

OriPekelman:
New clean PR without Gemfile.lock update see #5491 

Note that Travis will complain and build will stop with:
```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.
The dependencies in your gemfile changed
You have added to the Gemfile:
* ember-source (= 2.13.3)
* pg (~> 0.21.0)
You have deleted from the Gemfile:
* ember-source
* pg```


SamSaffron:
thanks

PR #5537, opened 2018-01-26
FIX: allowed href scheme link can start with a +

majakomel:
It makes it possible to add links for international phone numbers that start with a plus sign
meta: https://meta.discourse.org/t/tel-links-not-working-with-plus-sign/76483


coding-horror:
can we scope this to only mailto: ? I am worried about security implications of allowing extra characters in _all_ hrefs

majakomel:
I have updated the code so it allows + only for tel:

SamSaffron:
can we add a test for this, either in pretty_text_spec.rb or the js one?

majakomel:
yep, added

PR #5538, opened 2018-01-27
FEATURE: Adds option to search only in titles

jorgemanrubia:
It adds a new search option `in:title` that will make it search only in titles.

It also adds a new checkbox at the top of the advanced search UI:

screen recording 2018-01-27 at 02 01 pm

See: 

- https://meta.discourse.org/t/add-in-title-modifier-to-advanced-search/38546?source_topic_id=77463
- https://meta.discourse.org/t/search-only-within-topic-titles/27600?source_topic_id=77463

## Overview of changes

It changes the way posts and topics are indexed to use [PostgreSQL full-text search weights](https://www.postgresql.org/docs/9.1/static/textsearch-controls.html). Then, it uses the weight associated with `title` to filter by title when searching posts.

Before, it was concatenating `title + body + category` and storing the indexed result. Now, it will store the concatenation of weighted indexed parts.

This is my first experience with PostgreSQL full-text search. I think it is the proper way to do it after checking the docs. The weights system is a little bit bizarre with 4 values (A, B, C, D) but seems to work pretty well.

As a side effect, results will take weight into consideration when sorting. So `title` > `body` > `category`. From [12.3.3. Ranking Search Results](https://www.postgresql.org/docs/9.1/static/textsearch-controls.html):

> If no weights are provided, then these defaults are used:
> {0.1, 0.2, 0.4, 1.0}

So for posts, weights will be 1.0 for title `title`, 0.4 for `body` and 0.2 for `category`. I think this is ok, but it's definitely something good to check before merging the PR. I tested this locally with a database with 19k posts from a forum I used to run, and the sorting made sense to me.

I will add some comments in the code to explain the changes next.

## Reindexing required

`in:title` will only work after re-indexing all the posts. Not sure how to do it automatically when shipping this:

- Should it trigger a reindexing job as part of the deploy? Is there support in place for such tasks? I guess we could launch a job from a migration if not
- Should it just increase [the `INDEX_VERSION` ](https://github.com/jorgemanrubia/discourse/blob/258051809096598a7390cfbdbac3191d4cf865e3/lib/search.rb#L4) and rely on the [indexing job](https://github.com/jorgemanrubia/discourse/blob/2c56f8df7c1eba7ff81bc66eaafd1e27a8c4ff19/app/jobs/scheduled/reindex_search.rb#L3) that runs every day?

In development I have used `rake search:reindex` for testing this.

### DIFF
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -444,19 +444,7 @@ export default Em.Component.extend({
 
   @observes('searchedTerms.special.in.likes')
   updateSearchTermForSpecialInLikes() {
-    const match = this.filterBlocks(REGEXP_SPECIAL_IN_LIKES_MATCH);
-    const inFilter = this.get('searchedTerms.special.in.likes');
-    let searchTerm = this.get('searchTerm') || '';
-
-    if (inFilter) {
-      if (match.length === 0) {
-        searchTerm += ` in:likes`;
-        this.set('searchTerm', searchTerm.trim());
-      }
-    } else if (match.length !== 0) {
-      searchTerm = searchTerm.replace(match, '');
-      this.set('searchTerm', searchTerm.trim());
-    }
+    this.updateSearchTermForSpecialIn('likes', REGEXP_SPECIAL_IN_LIKES_MATCH);
   },
 
   @observes('searchedTerms.special.in.title')
@@ -466,36 +454,12 @@ export default Em.Component.extend({
 
   @observes('searchedTerms.special.in.private')
   updateSearchTermForSpecialInPrivate() {
-    const match = this.filterBlocks(REGEXP_SPECIAL_IN_PRIVATE_MATCH);
-    const inFilter = this.get('searchedTerms.special.in.private');
-    let searchTerm = this.get('searchTerm') || '';
-
-    if (inFilter) {
-      if (match.length === 0) {
-        searchTerm += ` in:private`;
-        this.set('searchTerm', searchTerm.trim());
-      }
-    } else if (match.length !== 0) {
-      searchTerm = searchTerm.replace(match, '');
-      this.set('searchTerm', searchTerm.trim());
-    }
+    this.updateSearchTermForSpecialIn('private', REGEXP_SPECIAL_IN_PRIVATE_MATCH);
   },
 
   @observes('searchedTerms.special.in.seen')
   updateSearchTermForSpecialInSeen() {
-    const match = this.filterBlocks(REGEXP_SPECIAL_IN_SEEN_MATCH);
-    const inFilter = this.get('searchedTerms.special.in.seen');
-    let searchTerm = this.get('searchTerm') || '';
-
-    if (inFilter) {
-      if (match.length === 0) {
-        searchTerm += ` in:seen`;
-        this.set('searchTerm', searchTerm.trim());
-      }
-    } else if (match.length !== 0) {
-      searchTerm = searchTerm.replace(match, '');
-      this.set('searchTerm', searchTerm.trim());
-    }
+    this.updateSearchTermForSpecialIn('seen', REGEXP_SPECIAL_IN_SEEN_MATCH);
   },
 
   @observes('searchedTerms.status')
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -18,6 +18,7 @@ const REGEXP_TAGS_REPLACE          = /(^(tags?:|#(?=[a-z0-9\-]+::tag))|::tag\s?$
 
 const REGEXP_IN_MATCH                 = /^(in|with):(posted|watching|tracking|bookmarks|first|pinned|unpinned|wiki|unseen|image)/ig;
 const REGEXP_SPECIAL_IN_LIKES_MATCH   = /^in:likes/ig;
+const REGEXP_SPECIAL_IN_TITLE_MATCH   = /^in:title/ig;
 const REGEXP_SPECIAL_IN_PRIVATE_MATCH = /^in:private/ig;
 const REGEXP_SPECIAL_IN_SEEN_MATCH    = /^in:seen/ig;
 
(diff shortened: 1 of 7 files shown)


app/services/search_indexer.rb
```diff
@@ -14,63 +14,109 @@ def self.scrub_html_for_search(html)
     HtmlScrubber.scrub(html)
   end
 
-  def self.update_index(table, id, raw_data)
-    raw_data = Search.prepare_data(raw_data, :index)
+  # @param raw_entries The list of entries to index. When more than one entry is provided, they will get a weight. The
+  # first one will have a priority A; he second one a priority B, etc.
+  # When only 1 entry is provided, no weights will be used.
+  def self.update_index(table, id, *raw_entries)
+    raw_data = Search.prepare_data(raw_entries.join(' '), :index)
```
jorgemanrubia: This methods now supports providing multiple entries to index and, when more than 1 entry is provided, it will assign them a weight.

I didn't add support for defining the specific weights and, instead, used the convention of deducing it from the order. So the first entry gets an `A`. The second a `B`. And so. I think this works well for what we need and keeps the code simpler.

app/services/search_indexer.rb
```diff
@@ -14,63 +14,109 @@ def self.scrub_html_for_search(html)
...
+    else
+      weight_letters = ('A'..'D').to_a
+      ts_vectors.collect.with_index do |ts_vector, index|
+        weight_letter = weight_letters[index]
+        "setweight(#{ts_vector}, '#{weight_letter}')"
+      end.join(" || ")
+    end
+  end
+
   def self.update_topics_index(topic_id, title, cooked)
```
jorgemanrubia: While weights were only needed for posts, I added them for topics too for consistency

app/services/search_indexer.rb
```diff
@@ -14,63 +14,109 @@ def self.scrub_html_for_search(html)
...
+                    VALUES (:id, #{ts_vector}, :locale, :raw_data, :version)",
+                                      statement_params)
     end
-  rescue
+
+    rescue
     # don't allow concurrency to mess up saving a post
   end
 
+  # for user login and name use "simple" lowercase stemmer
```
jorgemanrubia: I extracted a few methods from `SearchIndexer#update_index`

app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -424,15 +426,14 @@ export default Em.Component.extend({
     }
   },
 
-  @observes('searchedTerms.special.in.likes')
-  updateSearchTermForSpecialInLikes() {
-    const match = this.filterBlocks(REGEXP_SPECIAL_IN_LIKES_MATCH);
-    const inFilter = this.get('searchedTerms.special.in.likes');
+  updateSearchTermForSpecialIn(key, regexp){
```
jorgemanrubia: Extracted method to remove some duplicated code for the "searching in" logic

app/services/search_indexer.rb
```diff
@@ -14,63 +14,109 @@ def self.scrub_html_for_search(html)
...
+    end
+
+    statement_params
+  end
+
+  def self.update_index_param_name_for(index)
+    "indexable_fragment_#{index}".to_sym
+  end
+
+  def self.build_ts_vector(stemmer, indexable_entries)
```
jorgemanrubia: This is the place where we add weights to entries when more than 1 entry is provided. When only 1 entry is provided (users, categories) it works like it used to.

jorgemanrubia:
When searching exact terms with `"..."` it won't work with titles. This is already happening in current version: titles are ignored when performing exact searches. 

The reason is that [these queries check `posts.raw`](https://github.com/jorgemanrubia/discourse/blob/5093fcd31c4330b403f43de0a9aa7a1824ef63d3/lib/search.rb#L690), which does not include the title. I can tweak that and use `post_search_data.raw_data` which is the concatenation of `title+cooked+category name` but wasn't really sure if we want to tweak that.

SamSaffron:
That 1.freeze I may have added there sure is odd, I would just bump version to 2 and rely on it the indexing job. 

Can you review the existing code that attempts to give a bump to title. Perhaps there is a bunch of code that can be removed now? eg: https://github.com/discourse/discourse/blob/master/lib/search.rb#L746 

I think that checkbox in the UI has a bit too much prominence I would bump it to the bottom of the advanced area. 

Overall the PR looks great, nice feature.

jorgemanrubia:
@SamSaffron I added some commits addressing your comments, thanks!

Regarding the ordering, [this commit](https://github.com/discourse/discourse/pull/5538/commits/74fc6d86dc2cbb781346a9843a91447d2bc4c424) will use the new weights to bump titles. With the default weights of `{0.1, 0.2, 0.4, 1.0}` it wasn't bumping titles as in the current implementation, where title matches _always_ appear first. I experimented with a few values and making the relation `1` to `0.1` keeps the same behaviour.

A problem with this change is that, until the database is reindexed, title bumping in results won't work. Not sure if that's ok. If it is not, we can rollback that change and deploy it a few days after getting this shipped to be safe.

jorgemanrubia:
@SamSaffron I was having a look at [`Job::ReindexSearch`](https://github.com/jorgemanrubia/discourse/blob/2c56f8df7c1eba7ff81bc66eaafd1e27a8c4ff19/app/jobs/scheduled/reindex_search.rb#L3) and I am seeing that it has a limit of `10000` for posts. I wonder if that's limit is too short for very active communities (such as Meta). I understand it's for performance reasons, but what about getting rid of the limit, adding a `select` for only loading `ids` into the records, and then using `find_each` to process them in batches and prevent memory issues. 

Something like:

```ruby
      posts = Post.joins(:topic)
        .select(:id)
        .where(...)
...
      posts.find_each{|post| indexing_stuff }
```

SamSaffron:
OK, I took some time today to re-do this work based on your awesome changes.  per: https://github.com/discourse/discourse/commit/86d12bd44bb23767958364743a2afdf064331c81 

I opted to hack this in myself cause I was very concerned about maintenance and needed to know exactly how everything works together. 

I opted out of UI changes for now for a couple more betas so the index has some time to rebuild. 

Overall I was very happy with this change and I think it provides discourse with a significantly better search. 

ranking wise I went with `title` `category` `tags` `body` this seems to be doing the trick.

coding-horror:
yes this was an excellent nudge to improve search on titles which we did get a lot of complaints about .. glad we were actually doing something wrong, that makes it easier to fix.. thank you @jorgemanrubia

PR #5540, opened 2018-01-29
FEATURE: Mixed text direction support

scossar:
Adds a mixed_text_direction_support setting. When enabled it:

- adds a toggle-direction button to the composer
- sets the text direction on all cooked posts
- sets the text direction on the fancy_title
- sets the text direction on category descriptions
- sets the text direction on category names
- sets the text direction on topic excerpts
- adds a dir attribute to text fields, the dir is set based on the text field's content

For titles and category descriptions it's setting the dir in an inner span. Setting it on the outer element, for example a `h1` tag, will mess up the page formatting.

There are no tests yet, but these changes shouldn't make any existing tests fail.

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -8,6 +8,7 @@ import { emojiSearch, isSkinTonableEmoji } from 'pretty-text/emoji';
 import { emojiUrlFor } from 'discourse/lib/text';
 import { getRegister } from 'discourse-common/lib/get-owner';
 import { findRawTemplate } from 'discourse/lib/raw-templates';
+import { siteDir } from 'discourse/lib/text-direction';
 import { determinePostReplaceSelection, clipboardData } from 'discourse/lib/utilities';
 import toMarkdown from 'discourse/lib/to-markdown';
 import deprecated from 'discourse-common/lib/deprecated';
@@ -106,6 +107,17 @@ class Toolbar {
       title: 'composer.olist_title',
       perform: e => e.applyList(i => !i ? "1. " : `${parseInt(i) + 1}. `, 'list_item')
     });
+
+    if (Discourse.SiteSettings.support_mixed_text_direction) {
+      this.addButton({
+        id: 'toggle-direction',
+        group: 'extras',
+        icon: 'arrows-h',
+        shortcut: 'Shift+6',
+        title: 'composer.toggle_direction',
+        perform: e => e.toggleDirection(),
+      });
+    }
 
     if (site.mobileView) {
       this.groups.push({group: 'mobileExtras', buttons: []});
@@ -647,6 +659,14 @@ export default Ember.Component.extend({
     return null;
   },
 
+  _toggleDirection() {
+    const $textArea = $(".d-editor-input");
+    let currentDir = $textArea.attr('dir') ? $textArea.attr('dir') : siteDir(),
+        newDir = currentDir === 'ltr' ? 'rtl' : 'ltr';
+
+    $textArea.attr('dir', newDir).focus();
+  },
+
   paste(e) {
     if (!$(".d-editor-input").is(":focus")) {
       return;
@@ -724,6 +744,7 @@ export default Ember.Component.extend({
         addText: text => this._addText(selected, text),
         replaceText: text => this._addText({pre: '', post: ''}, text),
         getText: () => this.get('value'),
+        toggleDirection: () => this._toggleDirection(),
       };
 
       if (button.sendAction) {
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -112,7 +112,7 @@ class Toolbar {
       this.addButton({
         id: 'toggle-direction',
         group: 'extras',
-        icon: 'arrows-h',
+        icon: 'exchange',
         shortcut: 'Shift+6',
         title: 'composer.toggle_direction',
         perform: e => e.toggleDirection(),
--- a/app/assets/javascripts/discourse/components/text-field.js.es6
+++ b/app/assets/javascripts/discourse/components/text-field.js.es6
@@ -1,7 +1,35 @@
 import computed from "ember-addons/ember-computed-decorators";
+import { siteDir } from "discourse/lib/text-direction";
+import { isRTL } from "discourse/lib/text-direction";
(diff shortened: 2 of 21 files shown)


app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -107,6 +108,17 @@ class Toolbar {
       perform: e => e.applyList(i => !i ? "1. " : `${parseInt(i) + 1}. `, 'list_item')
     });
 
+    if (Discourse.SiteSettings.support_mixed_text_direction) {
+      this.addButton({
+        id: 'toggle-direction',
+        group: 'extras',
+        icon: 'arrows-h',
+        shortcut: 'Shift+6',
```
SamSaffron: hmmm is this not a bit too common?

SamSaffron:
@eviltrout can you review? 

@scossar noticing there are no tests here can you add some?

app/assets/javascripts/discourse/components/text-field.js.es6
```diff
@@ -1,7 +1,35 @@
 import computed from "ember-addons/ember-computed-decorators";
+import { siteDir } from "discourse/lib/text-direction";
```
eviltrout: You can import all three on one line:

`import { siteDir, isRTL, isLTR } from ...`

app/assets/javascripts/discourse/components/text-field.js.es6
```diff
@@ -1,7 +1,35 @@
...
+import { isRTL } from "discourse/lib/text-direction";
+import { isLTR } from "discourse/lib/text-direction";
 
 export default Ember.TextField.extend({
-  attributeBindings: ['autocorrect', 'autocapitalize', 'autofocus', 'maxLength'],
+  attributeBindings: ['autocorrect', 'autocapitalize', 'autofocus', 'maxLength', 'dir'],
+
+  @computed
+  dir() {
+    if (Discourse.SiteSettings.support_mixed_text_direction) {
```
eviltrout: In a component you should use `this.siteSettings` instead of `Discourse.SiteSettings`

app/assets/javascripts/discourse/components/text-field.js.es6
```diff
@@ -1,7 +1,35 @@
...
+      if (val) {
+        return isRTL(val) ? 'rtl' : 'ltr';
+      } else {
+        return siteDir();
+      }
+    }
+  },
+
+  keyUp() {
+    if (Discourse.SiteSettings.support_mixed_text_direction) {
```
eviltrout: Again should be `this.siteSettings`

app/assets/javascripts/discourse/lib/text-direction.js.es6
```diff
@@ -0,0 +1,28 @@
+const ltrChars = 'A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF';
+const rtlChars = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC';
+
+export function isRTL(text) {
+  const rtlDirCheck = new RegExp('^[^'+ltrChars+']*['+rtlChars+']');
```
eviltrout: Rather than parsing this every time it's called, perhaps you should move `rtlDirCheck` outside this function as a constant? Then it would only be parsed once.  (Same with `ltrDirCheck` below)

The only gotcha is you have to reset `lastIndex` before you use it., but it should be more efficient.

app/assets/javascripts/discourse/lib/text-direction.js.es6
```diff
@@ -0,0 +1,28 @@
...
+    let $e = $(e),
+      textContent = $e.text();
+    if (textContent) {
+      isRTL(textContent) ? $e.attr('dir', 'rtl') : $e.attr('dir', 'ltr');
+    }
+  });
+}
+
+export function siteDir() {
+  return $('html').hasClass('rtl') ? 'rtl' : 'ltr';
```
eviltrout: You should cache this result. If it's in the html of the document I don't expect it to change. You can use a variable outside the function like `let _siteDir;` then `if (!siteDir) { ... calculate it }`

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -58,7 +59,13 @@ const Topic = RestModel.extend({
 
   @computed('fancy_title')
   fancyTitle(title) {
-    return censor(emojiUnescape(title || ""), Discourse.Site.currentProp('censored_words'));
+    let fancyTitle = censor(emojiUnescape(title || ""), Discourse.Site.currentProp('censored_words'));
```
eviltrout: It wasn't in here before, but can you try `this.site` instead of `Discourse.Site` here? If it works it would be a win.

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -58,7 +59,13 @@ const Topic = RestModel.extend({
...
+
+    if (Discourse.SiteSettings.support_mixed_text_direction) {
```
eviltrout: This should be `this.siteSettings` - although the `Discourse.Site` above makes me wonder if it's not wired up properly. If it doesn't work falling back to `Discourse.SiteSettings` is fine.

app/assets/javascripts/discourse/models/topic.js.es6
scossar: It doesn't work. this.site isn't defined for the categories route.

scossar:
I've added a some tests. I'd like to improve the tests for `lib/text-direction`. My IDE goes a little nuts when I try to add mixed ltr/rtl text to a string. I'd also like to add some tests for making sure that the cooked post content is decorated correctly.

I've incorporated most of the changes that were suggested. I don't think the regular expression's `lastIndex` needs to be reset, since it's not using the global flag.

It's hard to find a good icon for the direction-toggle button, I've changed it to `exchange`. That seems a little better.

app/assets/javascripts/discourse/models/topic.js.es6
eviltrout: Ugh, I should fix that at some point.

eviltrout:
Looks good now!

SamSaffron:
@eviltrout feel to merge in we really want this deployed to macudacity

PR #5542, opened 2018-01-29
add a custom user onebox

majakomel:
meta: https://meta.discourse.org/t/improve-the-user-onebox-in-discourse/76087/8

screen shot 2018-01-29 at 22 17 32


SamSaffron:
this is great, but we really need at least 1 test here!

ZogStriP:
Awesome 👍

PR #5543, opened 2018-01-30
Feature: Invision Power Board Importer

pfaffman:
Here's an importer for Invision Power Board. I've used it on two sites that were very different (different table names, dramatically different format of raw posts).

I don't know it it'll work for anyone, but several people are asking for it.

### DIFF
--- a/script/import_scripts/ipboard.rb
+++ b/script/import_scripts/ipboard.rb
@@ -1,0 +1,1019 @@
+# coding: utf-8
+require "mysql2"
+require File.expand_path(File.dirname(__FILE__) + "/base.rb")
+require 'htmlentities'
+begin
+  require 'reverse_markdown' # https://github.com/jqr/php-serialize
+rescue LoadError
+  puts
+  puts 'reverse_markdown not found.'
+  puts 'Add to Gemfile, like this: '
+  puts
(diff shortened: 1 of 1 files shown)


script/import_scripts/ipboard.rb
```diff
@@ -0,0 +1,1019 @@
...
+
+      # Fix or leave bogus username?
+      username = find_user_by_import_id(quote[1]) || quote[1]
+      puts "username: #{username}, post_id: #{post_num}, topic_id: #{topic_num}" unless QUIET
+      puts "Before fixing a quote: #{raw}\n**************************************** " unless QUIET
+      post_string = post_num ? ", post:#{post_num}" : ""
+      topic_string = topic_num ? ", topic:#{topic_num}" : ""
+      raw.gsub!(quote_regex, "\n[quote=\"#{username}#{post_string}#{topic_string}\"]\n\n")
+      puts "AFTER!!!!!!!!!!!!1: #{raw}" unless QUIET
+      sleep 1
```
nlalonde: Why sleep here?

pfaffman: I think it must have been because I was tired?

This is not code that I'm proud of, but I'm sick of it and people are asking for it, so I'm submitting bad code rather than no code. If this is the most embarrassing thing you find, I'll be elated.

PR #5545, opened 2018-01-31
Rename "private" to "personal" in setting keys

techAPJ:
@ZogStriP can you please help me review this? (not urgent) Thanks!

NOTE: This PR should be merged after @nlalonde releases new beta this week.


ZogStriP:
Looks good 👍

eviltrout:
Won't this break every plugin that are using those settings?

I don't think we can accept that level of breakage without backwards compatibility.

I would suggest adding methods with the old names to `SiteSetting` that log a message saying "X is deprecated and has been renamed to Y"

That way plugins that reference the old settings will continue to work.

techAPJ:
> I would suggest adding methods with the old names to SiteSetting that log a message saying "X is deprecated and has been renamed to Y"

@eviltrout Sure, I will look into it and do so. Just curious, did we do something similar when we deprecated "blocked" in favour of "silence"? If not, should we do this for them as well?

eviltrout:
@techAPJ I didn't - in that case I searched our all-the-plugins repo and say that nobody was using the settings. You will likely find the same with most of yours. You could only deprecate the ones that are being used by plugins if you like.  However I would say one of our big upcoming customers definitely changes those settings.

techAPJ:
@eviltrout Got it. Thanks for the clarification. I will go ahead and deprecate all the settings I changed, just to be on the safe side.

techAPJ:
@eviltrout Done in: https://github.com/discourse/discourse/pull/5545/commits/285ef358cc6531e60dcabe97812ba949d8108630.

Thanks for the suggestion!

PR #5547, opened 2018-01-31
add og metadata for groups

majakomel:
which makes onebox for groups possible: https://meta.discourse.org/t/onebox-for-groups/79155

### DIFF
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -1,0 +1,3 @@
+<% content_for :head do %>
+  <%= raw crawlable_meta_data(title: @title, description: @description_meta) %>
+<% end %>
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -17,6 +17,12 @@ describe GroupsController do
       expect(::JSON.parse(response.body)['basic_group']['id']).to eq(group.id)
     end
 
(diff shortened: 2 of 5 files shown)


SamSaffron:
awesome, can you add a few screenshots on the topic on meta?

spec/controllers/groups_controller_spec.rb
```diff
@@ -17,6 +17,12 @@
       expect(::JSON.parse(response.body)['basic_group']['id']).to eq(group.id)
     end
 
+    it "responds with HTML" do
+      Guardian.any_instance.expects(:can_see?).with(group).returns(true)
```
SamSaffron: hmmm why is this needed? just create a visible group. the mocking here just risks instability long term

app/views/groups/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% content_for :head do %>
+  <%= raw crawlable_meta_data(title: @title, description: @description_meta) %>
+<% end %>
```
SamSaffron: we should add one test or 2 that renders views and makes sure content is there.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/76a2fc3d0782aede35b77bae1604ef297d7ee29c

Thanks @majakomel

PR #5548, opened 2018-01-31
FIX: Remove activation link from account approved email

jomaxro:


coding-horror:
I am not sure this is correct -- in some circumstances the link might be
needed.

jomaxro:
Definitely needs a proper review from @gschlager - I agree that I may be not thinking of a code path here.

PR #5560, opened 2018-02-02
FIX: dir-span when null is passed as argument

scossar:
The `dir-span` helper function is returning the string 'null' for the category description when no description is present. This affects sites that have not enabled the mixed-text-direction support setting.

This fix returns an empty string instead.


scossar:
screen shot 2018-02-02 at 9 04 29 am

PR #5562, opened 2018-02-02
FEATURE: Theme settings

OsamaSayegh:
Meta topic: https://meta.discourse.org/t/add-support-for-theme-settings/78103/5?u=osama

I extracted JS code from the site-setting component and model into mixins instead of rewriting the exact same logic. If there is a better way to do that let me know :)

Screenshots in the linked topic.


OsamaSayegh:
Any ideas why some of the tests I wrote for this are failing on Travis but not when I run them locally?

SamSaffron:
To me this looks like there is a state issue in your specs where something is getting cached and never invalidated, I would try a bundle exec rspec --bisect, first try running the specs with the same seed that is running when failed.

OsamaSayegh:
I updated this so now it allows theme settings to be used in CSS as per this topic: https://meta.discourse.org/t/size-topic-avatar/80112/14?u=osama

As to the specs, I tried running the specs with the same seed and they passed. I also gave `--bisect` a shot but it found no failures so it didn't help much. 😕

SamSaffron:
OK will aim to have this merged in early next week, will help debugging the fails

db/migrate/20180118215249_create_theme_settings.rb
```diff
@@ -0,0 +1,12 @@
+class CreateThemeSettings < ActiveRecord::Migration[5.1]
+  def change
+    create_table :theme_settings do |t|
+      t.string :name, limit: 255, null: false
+      t.integer :data_type, null: false
+      t.text :value
+      t.references :theme, foreign_key: true, null: false
```
SamSaffron: oh ... we usually do not include these in our tables

app/serializers/theme_serializer.rb
```diff
@@ -24,11 +24,7 @@ def filename
   end
 
   def target
-    case object.target_id
-    when 0 then "common"
-    when 1 then "desktop"
-    when 2 then "mobile"
-    end
+    Theme.targets.invert[object.target_id]&.to_s
```
SamSaffron: hmmm much too fancy for a serializer it has to generate this inverted hash each time it serializes. if logic is on theme it makes more sense

Theme.lookup_target(object.target_id)&.to_s

app/models/theme_setting.rb
```diff
@@ -0,0 +1,64 @@
+class ThemeSetting < ActiveRecord::Base
+  belongs_to :theme
+
+  validates_presence_of :name, :theme
+  validates :data_type, numericality: { only_integer: true }
+  validates :name, length: { maximum: 255 }
+
+  after_commit do
```
SamSaffron: I wonder if this should be after_save for extra simplicity, cause if you roll back you may corrupt cache?

app/assets/javascripts/admin/mixins/setting-object.js.es6
```diff
@@ -0,0 +1,29 @@
...
+        vals.addObject({ name: I18n.t(v.name), value: v.value });
+      } else {
+        vals.addObject(v);
+      }
+    });
+    return vals;
+  }.property('valid_values'),
+
+  allowsNone: function() {
+    if ( _.indexOf(this.get('valid_values'), '') >= 0 ) return 'admin.site_settings.none';
```
SamSaffron: at this point it is no longer a site_setting ... probably a "setting" localization should reflect

OsamaSayegh:
Sorry, I messed this up and there seems to be no way back. Here is a new PR https://github.com/discourse/discourse/pull/5611

PR #5567, opened 2018-02-05
Switch ids in References-Header field of mails.

fwolfst:
This change allows email-clients to show threaded views of mails as
expected.  Apparently most algorithms expect the message ids of mails
in the Reference-header-field to be sorted such that they build a
traversal through the thread, so the oldest (original) message being
first, then its child, grandchild and so on until it arrives at the
message id that the "new" mail (that is to be sent) is the reply to.

MSGA [1]
+- Re: MSGA [1-1]
|  +- Re: Re: MSGA [1-2-1]
|  +- Re: Re: MSGA [1-2-2]
+- Re: MSGA [1-1]

If the stuff in brackets would be the message ID, the References-Header
field of a message that is a reply to [1-2-1] should look like:

References: 1, 1-1, 1-2-1

Discussion took place in:
https://meta.discourse.org/t/e-mail-threading-in-ml-mode-does-not-work-in-thunderbird

Main information taken from:
https://www.jwz.org/doc/threading.html

### DIFF
--- a/lib/email/sender.rb
+++ b/lib/email/sender.rb
@@ -109,7 +109,7 @@ module Email
         else
           @message.header['Message-ID']  = post_message_id
           @message.header['In-Reply-To'] = referenced_post_message_ids[0] || topic_message_id
-          @message.header['References']  = [referenced_post_message_ids, topic_message_id].flatten.compact.uniq
+          @message.header['References']  = [topic_message_id, referenced_post_message_ids].flatten.compact.uniq
         end
 
         # https://www.ietf.org/rfc/rfc2919.txt


gschlager:
@ZogStriP You should test it with Mac Mail before merging, because you switched the order within the `References` header in https://github.com/discourse/discourse/commit/82555ca7612e497df31f9f4fc6969049a9887988

fwolfst:
@gschlager Thanks, good pick and good to know that somebody can manually test on proprietary software. I did not check previous commits regarding that topic. And somehow I assumed that there wouldnt be any tests for mails, but I was mistaken. Awesome!

riking:
I think that referenced_post_ids still needs to be reversed, so that the first post being replied to is on the end!

fwolfst:
`I think that referenced_post_ids still needs to be reversed, so that the first post being replied to is on the end!`

Nope. See example in the commit / PR - message and in the threading link. The last entry in the references-list ist the same as the reply-to, which is the "last" (=NEWEST) message that was replied to. So to say, the reference list is ordered by date, or like a "route" from the very first topic-starting mail until the latest.

riking:
But doesn't line 111, literally just above this one, say that the reply-to message is the first entry in the list?

    @message.header['In-Reply-To'] = referenced_post_message_ids[0]

fwolfst:
Yes, it says

    @message.header['In-Reply-To'] = referenced_post_message_ids[0] ||
topic_message_id

Which is correct. To be precise, it would be
referenced_post_message_ids[-1] I think, but afaiu since discourse usually
models only one-level-deep hierarchies and given a "I reply by mail to one
other mail" (the rfc does not support answering multiple different mails
within ONE mail) what I observed in my mailing-list scenario is that there
is only two-references anyway:

  - the "top" message (topic_message)
  - the message that I hit reply-to

. Does that clear things?

Point is: in the Reference-List ("array") **the last entry should equal the
`reply-to`**. And here reply-to is literal: you can reply to a reply of a
reply ...

tgxworld:
@ZogStriP From the discussions, seems like this is ready to be merged?

fwolfst:
@ZogStriP  Awesome, thanks! I noticed though that some tests in the CI failed. Unfortunately I was not able to get the test-setup running quick enough. Afair it should be fairly trivial to adjust the tests. Sorry about that.

PR #5576, opened 2018-02-08
PERF: a faster way to count tags used per category

nlalonde:
A fix for the problem described here:

https://meta.discourse.org/t/slow-tags-query-when-alphabetical-sort-is-disabled/78133

With the category_tag_stats table, we can quickly find most used tags for a given set of categories. The tricky part is keeping these counts correct.

### DIFF
--- a/app/models/category_tag_stat.rb
+++ b/app/models/category_tag_stat.rb
@@ -1,0 +1,61 @@
+class CategoryTagStat < ActiveRecord::Base
+  belongs_to :category
+  belongs_to :tag
+
+  def self.topic_moved(topic, from_category_id, to_category_id)
+    if from_category_id
+      self.where(tag_id: topic.tags.map(&:id), category_id: from_category_id)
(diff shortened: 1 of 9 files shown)


app/models/category_tag_stat.rb
```diff
@@ -0,0 +1,61 @@
...
+         WHERE tag_id in (:tag_ids)
+           AND category_id = :category_id
+     RETURNING tag_id
+      SQL
+
+      tag_ids = topic.tags.map(&:id)
+      updated_tag_ids = self.exec_sql(sql, tag_ids: tag_ids, category_id: to_category_id).map { |row| row['tag_id'] }
+
+      (tag_ids - updated_tag_ids).each do |tag_id|
+        CategoryTagStat.create(tag_id: tag_id, category_id: to_category_id, topic_count: 1)
```
tgxworld: I think it'll be better to use `create!` if we're not checking for the return value of `create`

PR #5598, opened 2018-02-15
FEATURE:  Whitelist div[data-theme-*]

hnb-ku:
Whitelisting data attributes on `<div>` elements that match [data-theme-*] allows theme authors to target elements in posts and not have those attributes sanitized.


SamSaffron:
Thanks heaps!

PR #5603, opened 2018-02-17
Fix active/hover tag select box

naveed-ahmad:
ezgif com-video-to-gif

Yeah minor, but still it was bugging me. 

ezgif com-video-to-gif 1


jjaffeux:
It's made on purpose to allow for multiple tag selection to take less vertical space.

screen shot 2018-02-19 at 10 51 42

I could make it smarted and know how many tags are selected, and might do it (not sure cause I don’t want to use js for styling to allow people to easily change styles using css only), but for now I think it's reasonable to keep it that way.

PR #5605, opened 2018-02-17
Added the required hooks for discourse-favorites plugin.

nbianca:
This PR adds `plugin-outlet`s before category titles so a star symbol can be added by the Favorites plugin.

Also, I had to make a small change to the top period extracting method because otherwise prefixes "top/" and "favorites/" were considered the same.

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
+++ b/app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
@@ -11,6 +11,7 @@
             {{#if c.read_restricted}}
               {{d-icon 'lock'}}
             {{/if}}
+            {{category-title-before category=c}}
             {{c.name}}
           </h3>
         </a>
--- a/app/assets/javascripts/discourse/components/category-title-before.js.es6
+++ b/app/assets/javascripts/discourse/components/category-title-before.js.es6
@@ -1,0 +1,3 @@
+export default Ember.Component.extend({
+  tagName: ''
+});
--- a/app/assets/javascripts/discourse/templates/components/categories-only.hbs
+++ b/app/assets/javascripts/discourse/templates/components/categories-only.hbs
@@ -24,6 +24,7 @@
               <div class='subcategories'>
                 {{#each c.subcategories as |s|}}
                   <span class='subcategory'>
+                    {{category-title-before category=s}}
                     {{category-link s hideParent="true"}}
                     {{category-unread category=s}}
                   </span>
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/discourse/routes/build-topic-route.js.es6
```diff
@@ -91,7 +91,7 @@ export default function(filter, extras) {
       const topicOpts = {
         model,
         category: null,
-        period: model.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : ''),
+        period: model.get('for_period') || (filter.indexOf('top/') >= 0 ? filter.split('/')[1] : ''),
```
SamSaffron: hmmm what is this change for?

nbianca: I had to add a few new topic lists: favorites/latest, favorites/new, favorites/unread. When I added these, the filter would be misparsed and consider 'latest', 'new' or 'unread' as period which was wrong. The timeperiod should only be extracted when the filter is a 'top' filter (it contains 'top/').

nbianca:
Rebased on top of master.

app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
```diff
@@ -11,6 +11,7 @@
             {{#if c.read_restricted}}
               {{d-icon 'lock'}}
             {{/if}}
+            {{plugin-outlet name="category-title-before" noTags=true args=(hash category=c)}}
```
tgxworld: Let's move this into a component so that we don't have to duplicate the same line 3 times. :smile:

app/assets/javascripts/discourse/routes/build-topic-route.js.es6
tgxworld: @nbianca It'll be better to move this fix into another PR or at least into another commit. Otherwise, the commit is actually doing more than what the commit message say it does.

app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
nbianca: Done.

app/assets/javascripts/discourse/routes/build-topic-route.js.es6
nbianca: @tgxworld Done, see #5614.

tgxworld:
Thanks @nbianca!

PR #5610, opened 2018-02-19
FEATURE: Allow staffs to tag PMs

vinothkannans:


app/assets/javascripts/discourse/lib/render-tag.js.es6
```diff
@@ -3,7 +3,12 @@ export default function renderTag(tag, params) {
   tag = Handlebars.Utils.escapeExpression(tag);
   const classes = ['tag-' + tag, 'discourse-tag'];
   const tagName = params.tagName || "a";
-  const href = (tagName === "a" && !params.noHref) ? " href='" + Discourse.getURL("/tags/" + tag) + "' " : "";
+  let path;
+  if (tagName === "a" && !params.noHref) {
+    const current_user = Discourse.User.current();
+    path = params.isPrivateMessage && current_user && current_user.staff ? `/u/${current_user.username}/messages/tag/${tag}` : `/tags/${tag}`;
+  }
+  const href = path ? " href='" + Discourse.getURL(path) + "' " : "";
```
techAPJ: We should use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here as well.

app/assets/javascripts/discourse/lib/render-tags.js.es6
```diff
@@ -43,7 +44,7 @@ export default function(topic, params){
     buffer = "<div class='discourse-tags'>";
     if (tags) {
       for(let i=0; i<tags.length; i++){
-        buffer += renderTag(tags[i]) + ' ';
+        buffer += renderTag(tags[i], { isPrivateMessage: isPrivateMessage }) + ' ';
```
techAPJ: I think `{ isPrivateMessage: isPrivateMessage }` can be written simply as `{ isPrivateMessage }`. ([Reference documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer).)

SamSaffron:
I don't feel like we can merge this until we have the site settings

1. allow_staff_tags_on_messages: default false

^^^ this also means that only staff can see the tags on messages

I do not want to deal with non staff yet, they are a can of giant worms, I also feel this should be default off until we sort out all the kinks

vinothkannans:
@SamSaffron it is already limited only to the staffs. I willl add the site setting soon.

SamSaffron:
ok, yeah just the site setting then, so its default off. 

its a hard feature to explain so I want to live with it for a bit before exposing it to the general public by default.

vinothkannans:
Conflicts with the master branch are resolved and the site setting is added. @SamSaffron it is now mergeable.

config/site_settings.yml
```diff
@@ -1571,6 +1571,8 @@ tags:
     type: list
     client: true
     default: ''
+  allow_staff_to_tag_in_pm:
```
SamSaffron: ok, minor rename, `allow_staff_to_tag_pms`

app/models/topic_tag.rb
```diff
@@ -1,22 +1,28 @@
...
-  belongs_to :tag, counter_cache: "topic_count"
+  belongs_to :tag
 
   after_create do
-    if topic&.category_id
-      if stat = CategoryTagStat.where(tag_id: tag_id, category_id: topic.category_id).first
-        stat.increment!(:topic_count)
-      else
-        CategoryTagStat.create(tag_id: tag_id, category_id: topic.category_id, topic_count: 1)
+    if topic && topic.archetype != Archetype.private_message
```
tgxworld: Perhaps a test case for this to prevent us from incorrectly increasing the count.

lib/topic_query.rb
```diff
@@ -269,6 +269,14 @@ def list_private_messages_group_archive(user)
     create_list(:private_messages, {}, list)
   end
 
+  def list_private_messages_tag(user)
+    list = private_messages_for(user, :all)
+    tag_id = Tag.where('name ilike ?', @options[:tags][0]).pluck(:id).first
```
tgxworld: This is actually inefficient because it fetches all the `id`s into an array and discard the rest. It'll probably be better to just fetch one record from the DB and select the column you want. Maybe use `.select(:id)` if we're concern about forming an entire ActiveRecord object.

config/routes.rb
```diff
@@ -597,20 +598,20 @@
...
-    group_name: RouteFormat.username
-  }
+
+  scope "/topics", username: RouteFormat.username do
+    get "created-by/:username" => "list#topics_by", as: "topics_by"
+    get "private-messages/:username" => "list#private_messages", as: "topics_private_messages"
+    get "private-messages-sent/:username" => "list#private_messages_sent", as: "topics_private_messages_sent"
+    get "private-messages-archive/:username" => "list#private_messages_archive", as: "topics_private_messages_archive"
+    get "private-messages-unread/:username" => "list#private_messages_unread", as: "topics_private_messages_unread"
+    get "private-messages-tag/:username/:tag_id.json" => "list#private_messages_tag", as: "topics_private_messages_tag", constraints: StaffConstraint.new
```
tgxworld: We have support for filtering [multiple tags in public topics](https://meta.discourse.org/t/tag-intersections-page/47952) so we should probably do the same for private messages. Might in the next iteration?

app/serializers/topic_tags_mixin.rb
```diff
@@ -0,0 +1,13 @@
+module TopicTagsMixin
```
tgxworld: I would move this into a `concerns` folder under `serializers` so that it is clear this is a mixin.

tgxworld: Nice idea to use a mixin though :+1: We get to nuke so many lines of code.

app/serializers/topic_view_serializer.rb
```diff
@@ -239,7 +239,7 @@ def include_pending_posts_count?
   end
 
   def include_tags?
-    SiteSetting.tagging_enabled
+    SiteSetting.tagging_enabled && (!object.topic.private_message? || scope.can_tag_pms?)
```
tgxworld: I'm seeing variants of `SiteSetting.tagging_enabled && (!object.topic.private_message? || scope.can_tag_pms?)` being duplicated throughout in this PR. Maybe we can move the logic into a `guardian` method that takes in the topic as a parameter?

lib/discourse_tagging.rb
```diff
@@ -4,10 +4,10 @@ module DiscourseTagging
   TAGS_FILTER_REGEXP = /[\/\?#\[\]@!\$&'\(\)\*\+,;=\.%\\`^\s|\{\}"<>]+/ # /?#[]@!$&'()*+,;=.%\`^|{}"<>
 
   def self.tag_topic_by_names(topic, guardian, tag_names_arg, append: false)
-    if SiteSetting.tagging_enabled
+    if guardian.can_tag?(topic)
       tag_names = DiscourseTagging.tags_for_saving(tag_names_arg, guardian) || []
 
-      old_tag_names = topic.tags.map(&:name) || []
+      old_tag_names = topic.tags.pluck(:name) || []
```
tgxworld: :+1:

lib/guardian.rb
```diff
@@ -129,6 +129,12 @@ def can_moderate?(obj)
   alias :can_see_flags? :can_moderate?
   alias :can_close? :can_moderate?
 
+  def can_tag?(obj)
+    return false unless obj && obj.is_a?(Topic)
```
tgxworld: Hmm why do we have to check whether the object is a `Topic`? The only place using this method is always going to pass it an instance of `Topic.`

lib/topic_query.rb
```diff
@@ -269,6 +269,14 @@ def list_private_messages_group_archive(user)
...
+    list = list.joins("JOIN topic_tags tt ON tt.topic_id = topics.id AND
+                      tt.tag_id = #{tag_id}")
```
tgxworld: I'm wondering if we can combine the query ran in `tag_id` into a sub-query here. It'll allow us to send only one query to the DB.

lib/topic_query.rb
tgxworld: I'm thinking if we are expecting a partial tag name considering that we are using `ilike` here.

spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -64,4 +66,44 @@
       end
     end
   end
+
+  let(:user) { Fabricate(:user) }
```
tgxworld: Dup of https://github.com/discourse/discourse/pull/5610/files#diff-5790c3234f6398446eb0fc1ebf01245aR10

spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -64,4 +66,44 @@
...
+  let(:moderator) { Fabricate(:moderator) }
```
tgxworld: I think we should declare "global" `let`s at the top of the file to avoid duplications. `let` specific to a particular test case should probably be declared within the scope of the `describe` block.

lib/topic_query.rb
```diff
@@ -269,6 +269,14 @@ def list_private_messages_group_archive(user)
     create_list(:private_messages, {}, list)
   end
 
+  def list_private_messages_tag(user)
```
tgxworld: We should add a test case for this method.

spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -64,4 +66,44 @@
...
+    it "should not include the tag for normal users" do
+      json = serialize_topic(pm, user)
+      expect(json[:tags]).to eq(nil)
+    end
+
+    it "should include the tag for staff users" do
+      json = serialize_topic(pm, moderator)
+      expect(json[:tags]).to eq([tag.name])
+
+      json = serialize_topic(pm, Fabricate(:admin))
```
tgxworld: `Fabricate(:admin)` can be moved into `let(:admin)` just like how it is done for `moderator`

spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -64,4 +66,44 @@
...
+      expect(json[:tags]).to eq([tag.name])
+    end
+
+    it "should not include the tag if pm tags disabled" do
+      SiteSetting.allow_staff_to_tag_pms = false
+
+      json = serialize_topic(pm, moderator)
```
tgxworld: To avoid duplication here I would use a `[moderator, admin].each` loop

spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -64,4 +66,44 @@
...
+      SiteSetting.allow_staff_to_tag_pms = true
+    end
+
+    it "should not include the tag for normal users" do
+      json = serialize_topic(pm, user)
+      expect(json[:tags]).to eq(nil)
+    end
+
+    it "should include the tag for staff users" do
+      json = serialize_topic(pm, moderator)
```
tgxworld: Same for https://github.com/discourse/discourse/pull/5610/files#r169893022

app/controllers/list_controller.rb
```diff
@@ -332,6 +333,7 @@ def set_category
   def build_topic_list_options
     options = {}
     params[:page] = params[:page].to_i rescue 1
+    params[:tags] = [params[:tag_id]] if params[:tag_id].present? && guardian.can_tag_pms?
```
tgxworld: We need a test case for this as well :smile:

app/models/tag.rb
```diff
@@ -43,7 +28,7 @@ def self.update_topic_counts
         SELECT COUNT(topics.id) AS topic_count, tags.id AS tag_id
         FROM tags
         LEFT JOIN topic_tags ON tags.id = topic_tags.tag_id
-        LEFT JOIN topics ON topics.id = topic_tags.topic_id AND topics.deleted_at IS NULL
+        LEFT JOIN topics ON topics.id = topic_tags.topic_id AND topics.deleted_at IS NULL AND topics.archetype != "private_message"
```
tgxworld: We need a test case for this too smile

vinothkannans:
@tgxworld all the requested changes are done 👍

tgxworld:
Cool lets try this out.

PR #5611, opened 2018-02-19
FEATURE: Theme settings (2)

OsamaSayegh:
I accidentally messed up [the initial PR](https://github.com/discourse/discourse/pull/5562#pullrequestreview-97441642) so here is a fresh start.


OsamaSayegh:
@SamSaffron I addressed the things you pointed out in the previous PR.

I have a doubt about one thing that I'm not sure I've done correctly - you commented on my db migration that you want avoid having something like this in your tables:
```ruby
t.references :theme, foreign_key: true, null: false
```
So I changed it to this:
```ruby
t.integer :theme_id, null: false
```
based on what I saw in existing db migrations. I'm not sure if that's how you want it to be, so please correct me if I'm mistaken.

SamSaffron:
Thanks heap, I am merging this in now and am going to be experimenting with it today

PR #5612, opened 2018-02-20
Featheredtoast two factor login

tgxworld:
Includes my changes after reviewing https://github.com/discourse/discourse/pull/5521


db/migrate/20180109222722_create_user_second_factors.rb
```diff
@@ -0,0 +1,12 @@
+class CreateUserSecondFactors < ActiveRecord::Migration[5.1]
+  def change
+    create_table :user_second_factors do |t|
+      t.integer :user_id, null: false
+      t.string :method, null: false
```
tgxworld: @featheredtoast Is there a reason we need this `method` column as I don't see us using it anywhere.

featheredtoast:
That field would be used for future expansion on this - currently I'm
inserting "TOTP" for the record as a 1-1 relationship, but this would be
expanded for u2f, and backup codes as suggested by @riking here -
https://meta.discourse.org/t/two-factor-local-login-proposal/76689/14?u=awole20

The table would be able to be evolved into a 1-many relationship with
different kinds of second factors.

app/assets/javascripts/discourse/routes/preferences-second-factor.js.es6
```diff
@@ -9,13 +9,7 @@ export default RestrictedUserRoute.extend({
     return this.render({ into: 'user' });
   },
 
-  // A bit odd, but if we leave to /preferences we need to re-render that outlet
-  deactivate() {
```
tgxworld: @featheredtoast I wasn't sure why this was needed at all so I removed it first.

app/controllers/admin/users_controller.rb
```diff
@@ -342,15 +342,20 @@ def reject_bulk
   end
 
   def disable_second_factor
-    guardian.ensure_can_disable_second_factor! @user
-    if @user.user_second_factor.try(:delete)
```
tgxworld: @featheredtoast Generally we want to use `destroy!` so that if it fails to destroy for whatever reason, it'll at least raise an error

app/controllers/second_factor_controller.rb
```diff
@@ -1,51 +0,0 @@
-class SecondFactorController < ApplicationController
```
tgxworld: @featheredtoast I moved this under `UsersController` as I didn't think we need another controller to handle this. I didn't find any major problem with the methods in this controller. The only thing that I though would be useful to highlight is that this controller didn't have any access control checks. That means any user even anon could execute the methods in this controller.

app/helpers/second_factor_helper.rb
```diff
@@ -1,35 +0,0 @@
-module SecondFactorHelper
```
tgxworld: @featheredtoast Rails helpers are meant to extract logic used in views and isn't really a good place to contain server side logic. What I did here instead is to extract the logic here into a module `SecondFactorManager` and included the module in the `User` class.

db/migrate/20180109222722_create_user_second_factors.rb
```diff
@@ -2,8 +2,8 @@ class CreateUserSecondFactors < ActiveRecord::Migration[5.1]
   def change
     create_table :user_second_factors do |t|
       t.integer :user_id, null: false
-      t.string :method
```
tgxworld: @featheredtoast Instead of storing the same string multiple times in each column, I represented the method as an integer in https://github.com/discourse/discourse/pull/5612/commits/14f3594f9f39dce7f6a5dda44287560b0a6e53e8#diff-62eb46e48ba621e1c2c1c91dc9763c1bR4 which also takes up less space.

tgxworld:
@featheredtoast I thought this PR was really pleasant to review so I wanted to thank you for you work on this. While the PR has been merged, I left some comments in my [commit](https://github.com/discourse/discourse/pull/5612/commits/14f3594f9f39dce7f6a5dda44287560b0a6e53e8) so that you can get some feedback :smile: Thank you once again!

featheredtoast:
Hey, thank you for your detailed review! I'll look it over and see what I can improve upon - I really appreciate the feedback. 🎉

PR #5613, opened 2018-02-21
QR code display fix on dark backgrounds

featheredtoast:
https://meta.discourse.org/t/2fa-qr-code-not-visible-on-dark-theme/81152?u=awole20

This will put a white div with a padding of 5 around the QR code so that it's visible on a dark background.

Styles are currently inline - will gladly refactor all inline styles, but I'm not sure the best place for them in the scss.

### DIFF
--- a/app/assets/javascripts/discourse/templates/preferences-second-factor.hbs
+++ b/app/assets/javascripts/discourse/templates/preferences-second-factor.hbs
@@ -48,7 +48,11 @@
 
         <div class="control-group">
           <div class="controls">
-            {{{secondFactorImage}}}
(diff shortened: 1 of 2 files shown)


xfalcox:
> Styles are currently inline - will gladly refactor all inline styles, but I'm not sure the best place for them in the scss.

Preferences styles are set in user.scss

featheredtoast:
Thanks @xfalcox - just fixed the PR, user.scss now contains all the styling for this.

ZogStriP:
Thanks @featheredtoast 👍

PR #5614, opened 2018-02-22
Try extracting time period only when the filter is 'top'.

nbianca:
When developing [discourse-favorites](https://github.com/nbianca/discourse-favorites) I had to add a few new topic lists: favorites/latest, favorites/new, favorites/unread. When I added these, the filter would be misparsed and consider 'latest', 'new' or 'unread' as period which was wrong. The timeperiod should only be extracted when the filter is a 'top' filter (it contains 'top/').

### DIFF
--- a/app/assets/javascripts/discourse/routes/build-topic-route.js.es6
+++ b/app/assets/javascripts/discourse/routes/build-topic-route.js.es6
@@ -91,7 +91,7 @@ export default function(filter, extras) {
       const topicOpts = {
         model,
         category: null,
-        period: model.get('for_period') || (filter.indexOf('/') > 0 ? filter.split('/')[1] : ''),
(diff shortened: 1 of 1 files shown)


tgxworld:
@nbianca How are the new topic lists being added? Could I get some steps to replicate what you're seeing? Thanks!

nbianca:
Hi @tgxworld,

You can check out the plugin I wrote [discourse-favorites](https://github.com/nbianca/discourse-favorites) that adds a few new topic lists related to a user's favorites: latest favorites, newest favorites and unread favorites. Favorites are topics from a specific category.

Basically, the user selects a few categories he considers 'favorite categories' and then he can easily navigate those, by navigating the topic lists that aggregate topics from all of these categories.

tgxworld:
LGTM :+1:

PR #5615, opened 2018-02-24
Adds :puke: as alias to :face_vomiting:

jjaffeux:


eviltrout:
Well, this is it. The best commit in the history of Discourse.

PR #5616, opened 2018-02-24
FIX: set notification level when changing post owner

majakomel:
this a fix for https://meta.discourse.org/t/when-changing-post-owner-notifications-still-sent-from-to-original-owner/61266


tgxworld:
@majakomel I'm not sure if I'm following correctly but from what I understand in the meta topic, the bug is that the notification will be sent from an incorrect user. Therefore, updating the notification level wouldn't fix the bug describing in the topic.

majakomel:
@tgxworld This is a fix for the behaviour described from the 2nd post on. When ownership of the post is changed, the new owner doesn't get any notification if post gets replies (previous owner does). Also if the ownership of a reply is changed, new owner is not tracking the topic if they don't visit the topic. 

Incorrect user issue appears only in that specific case when initial poster mention themself, I'll check that one again and expand on it on meta.

majakomel:
I added a fix for behaviour from the first post too, so when user mentions themself in a post they're not the owner of, they don't get any notification (rather than from the owner of the post who didn't add the mention)

app/services/post_alerter.rb
```diff
@@ -43,7 +43,7 @@ def notify_about_reply?(post)
   end
 
   def after_save_post(post, new_record = false)
-    notified = [post.user]
+    notified = [post.user, post.last_editor].uniq
```
gschlager: 👍 Could you please add a test for this?

app/services/post_owner_changer.rb
```diff
@@ -23,6 +24,13 @@ def change_owner!
       post.set_owner(@new_owner, @acting_user, @skip_revision)
       PostAction.remove_act(@new_owner, post, PostActionType.types[:like])
 
+      if post.post_number == 1
+        TopicUser.change(old_owner.id, @topic.id, notification_level: 2) if old_owner
+        TopicUser.change(@new_owner.id, @topic.id, notification_level: 3)
```
gschlager: I'd prefer if you would use `NotificationLevels.topic_levels[:watching]` instead of a magic number. Same goes for the other notification levels.

app/services/post_owner_changer.rb
```diff
@@ -23,6 +24,13 @@ def change_owner!
       post.set_owner(@new_owner, @acting_user, @skip_revision)
       PostAction.remove_act(@new_owner, post, PostActionType.types[:like])
 
+      if post.post_number == 1
+        TopicUser.change(old_owner.id, @topic.id, notification_level: 2) if old_owner
```
gschlager: I'm not sure if we should change the `notification_level` of the previous owner. Let's discuss this on meta.
If we change it, should we set it to the user's notification level for that category or group or always to `tracking`?

PR #5617, opened 2018-02-25
FIX: reset_db

pfaffman:
To get the database rebuilt requires a restart of the container. 

I also added a rake admin:create, since it's useful, and it's what `boot_dev --init `does.


SamSaffron:
seems fine, thanks!

PR #5620, opened 2018-02-26
FEATURE: rake task for merging users

gschlager:


app/services/user_merger.rb
```diff
@@ -0,0 +1,339 @@
...
+  def move_posts
+    posts = Post.with_deleted
+      .where(user_id: @source_user.id)
+      .order(:topic_id, :post_number)
+      .pluck(:topic_id, :id)
+
+    topic_id = nil
+    post_ids = []
+
+    posts.each do |post|
```
ZogStriP: Instead of using `post[0]` and `post[1]` you can do

```ruby
posts.each do |(topic_id, post_id)|
```

tgxworld: I think 

```
posts.each do |topic_id, post_id|
```

works too.

gschlager: Yeah, that works too. Thanks for the tip! 🙂

PR #5621, opened 2018-02-26
Update xenforo.rb

neerajmalve:
#Fix for the error: xenforo.rb: 160: in `gsub!': invalid byte sequence in UTF-8 (ArgumentError)


gschlager:
Duplicate. Closing in favor of https://github.com/discourse/discourse/pull/5622

PR #5625, opened 2018-02-26
FEATURE: Add external_id to User Serializer

scossar:
Adds the `external_id` to UserSerializer if SSO is enabled.

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -73,7 +73,9 @@ class UserSerializer < BasicUserSerializer
              :primary_group_flair_bg_color,
              :primary_group_flair_color,
              :staged,
-             :second_factor_enabled
+             :second_factor_enabled,
+             :external_id
+
 
   has_one :invited_by, embed: :object, serializer: BasicUserSerializer
   has_many :groups, embed: :object, serializer: BasicGroupSerializer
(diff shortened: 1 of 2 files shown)


app/serializers/user_serializer.rb
```diff
@@ -287,6 +289,14 @@ def primary_group_flair_color
     object.try(:primary_group).try(:flair_color)
   end
 
+  def external_id
+    object.try(:single_sign_on_record).try(:external_id)
```
tgxworld: `try` is slower than `&.` :grin:

spec/serializers/user_serializer_spec.rb
```diff
@@ -196,4 +196,28 @@
...
+      sso.email = "test@example.com"
+      sso.external_id = "1"
+      user = sso.lookup_or_create_user
+      json = UserSerializer.new(user, scope: Guardian.new, root: false).as_json
+      expect(json[:external_id]).to eq("1")
+    end
+  end
+
+  context "when SSO is not enabled" do
+    SiteSetting.enable_sso = false
```
tgxworld: hmm is there a reason why this is outside of the `it` block?

app/serializers/user_serializer.rb
scossar: Thanks! I've updated it. I was using `try` because that's what is used in the other methods in the class.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/51544047dcf94435d606fefc59b5b4c2513b9fa3

PR #5629, opened 2018-02-27
Remove unused columns from topics table

gschlager:


SamSaffron:
@eviltrout are you ok with this?

eviltrout:
It looks like the columns on topics weren't used. `inappropriate_count` is used on `posts` by one of my plugins though. So this should be OK.

db/migrate/20180227160557_drop_action_count_columns_from_topics.rb
```diff
@@ -0,0 +1,9 @@
+class DropActionCountColumnsFromTopics < ActiveRecord::Migration[5.1]
```
SamSaffron: Oh ... we don't need a dummy migration here, just pick the last one currently, it will work fine. also keep in mind column dropper reruns

tgxworld: I think the last migration would likely have been created 60 minutes ago and that would cause the columns to be dropped immediately.

gschlager: I mimicked what has been done in the past when columns were dropped. I think @tgxworld is right. Picking the last migration might lead to columns being dropped immediately. I'm not sure if we really want that to happen.

SamSaffron:
Hmmm I just worry about this, lets hold off at least another 6 months before another cleanup like this.

PR #5630, opened 2018-02-27
Drop unused tables

gschlager:


SamSaffron:
@nlalonde  I recall you worked on this in the past, do you think it is safe to remove this?

nlalonde:
Looks like we stopped using that table at some point... So yes it looks safe to remove.

SamSaffron:
cool, then begone table 🐩

PR #5635, opened 2018-02-28
FIX: do not allow invite notifications from muted user/topic

techAPJ:
This PR adds a check for muted user/topic.

https://meta.discourse.org/t/making-it-so-other-users-cant-invite-you-to-a-topic/63311/11

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -791,6 +791,8 @@ SQL
       raise UserExists.new(I18n.t("topic_invite.user_exists"))
     end
 
+    return true if target_user && invite_existing_muted?(target_user, invited_by)
+
     if target_user && private_message? && topic_allowed_users.create!(user_id: target_user.id)
       add_small_action(invited_by, "invited_user", target_user.username)
 
@@ -833,6 +835,26 @@ SQL
 
   def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
     Invite.invite_by_email(email, invited_by, self, group_ids, custom_message)
+  end
+
+  def invite_existing_muted?(target_user, invited_by)
+    if invited_by.id &&
+       MutedUser.where(user_id: target_user.id, muted_user_id: invited_by.id)
+           .joins(:muted_user)
+           .where('NOT admin AND NOT moderator')
+           .exists?
+      return true
+    end
+
+    if TopicUser.where(
+         topic: self,
+         user: target_user,
+         notification_level: TopicUser.notification_levels[:muted]
+        ).exists?
+      return true
+    end
+
+    false
   end
 
   def email_already_exists_for?(invite)
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -533,6 +533,17 @@ describe Topic do
           expect(topic.reload.allowed_users).to_not include(another_user)
           expect(Post.last.action_code).to eq("removed_user")
         end
+
+        context "from a muted user" do
+          before { MutedUser.create!(user: another_user, muted_user: user) }
+
+          it 'silently fails' do
+            expect(topic.invite(user, another_user.username)).to eq(true)
+            expect(topic.allowed_users).to_not include(another_user)
+            expect(Post.last).to be_blank
+            expect(Notification.last).to be_blank
+          end
+        end
       end
 
       describe 'by email' do
@@ -597,6 +608,17 @@ describe Topic do
           expect(topic.invite(user, another_user.email)).to eq(true)
           expect(topic.reload.allowed_users.last).to eq(another_user)
           expect_the_right_notification_to_be_created
+        end
+
+        context "for a muted topic" do
+          before { TopicUser.change(another_user.id, topic.id, notification_level: TopicUser.notification_levels[:muted]) }
+
+          it 'silently fails' do
+            expect(topic.invite(user, another_user.username)).to eq(true)
+            expect(topic.allowed_users).to_not include(another_user)
+            expect(Post.last).to be_blank
+            expect(Notification.last).to be_blank
+          end
         end
 
         describe 'when user can invite via email' do


app/models/topic.rb
```diff
@@ -830,6 +832,23 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
     Invite.invite_by_email(email, invited_by, self, group_ids, custom_message)
   end
 
+  def invite_existing_muted?(target_user, invited_by)
+    # check if inviter is muted
+    return true if invited_by.id && MutedUser.where(user_id: target_user.id, muted_user_id: invited_by.id)
```
tgxworld: Small nitpick here regarding the style. Usually we use the style `return true if something` only when all the code fits in a single line.

app/models/topic.rb
```diff
@@ -830,6 +832,23 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
     Invite.invite_by_email(email, invited_by, self, group_ids, custom_message)
   end
 
+  def invite_existing_muted?(target_user, invited_by)
+    # check if inviter is muted
```
tgxworld: In my opinion, I don't think we need the comment here since the db query would explain what we're trying to do.

app/models/topic.rb
```diff
@@ -830,6 +832,23 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
...
+        .exists?
+
+    # skip if muted on the topic
+    return true if TopicUser.where(
+      topic: self,
+      user: target_user,
+      notification_level: TopicUser.notification_levels[:muted]
+    ).exists?
+
+    return false
```
tgxworld: No need for `return` here :wink:

spec/models/topic_spec.rb
```diff
@@ -505,6 +505,17 @@ def build_topic_with_title(title)
           expect(topic.reload.allowed_users).to_not include(another_user)
           expect(Post.last.action_code).to eq("removed_user")
         end
+
+        context "from a muted user" do
+          before { MutedUser.create(user: another_user, muted_user: user) }
```
tgxworld: I always prefer to use `create!` since it'll throw an error and point us at the problem when the creation fails.

app/models/topic.rb
```diff
@@ -830,6 +832,23 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
     Invite.invite_by_email(email, invited_by, self, group_ids, custom_message)
   end
 
+  def invite_existing_muted?(target_user, invited_by)
+    # check if inviter is muted
+    return true if invited_by.id && MutedUser.where(user_id: target_user.id, muted_user_id: invited_by.id)
+        .joins(:muted_user)
```
tgxworld: This looks odd to me cause we are joining on the same table. I think you meant to join on `users`?

app/models/topic.rb
```diff
@@ -791,6 +791,8 @@ def invite(invited_by, username_or_email, group_ids = nil, custom_message = nil)
       raise UserExists.new(I18n.t("topic_invite.user_exists"))
     end
 
+    return true if target_user && invite_existing_muted?(target_user, invited_by)
```
tgxworld: I think this will be strange for the person inviting since the method would return true making the client think that the invite was successful when in fact the invitation never happens in the first place.

techAPJ: This is intentional because we do not want to reveal that the user who they are trying to invite has muted them (or that topic). Throwing ambiguous here is not correct IMO.

This behaviour is consistent with how we handle mentions from muted user or for muted topic.

tgxworld: For a muted user, the mention would still look like a mention but this invite would end up with the user not being invited into the topic at all. I think it is really strange that inviting someone would just do nothing if the invitee has muted the inviter

app/models/topic.rb
```diff
@@ -835,6 +837,26 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
     Invite.invite_by_email(email, invited_by, self, group_ids, custom_message)
   end
 
+  def invite_existing_muted?(target_user, invited_by)
+    if invited_by.id &&
+       MutedUser.where(user_id: target_user.id, muted_user_id: invited_by.id)
+           .joins(:muted_user)
```
tgxworld: hmm 4 spaces are use for indentation here.

app/models/topic.rb
```diff
@@ -835,6 +837,26 @@ def invite_by_email(invited_by, email, group_ids = nil, custom_message = nil)
...
+           .where('NOT admin AND NOT moderator')
+           .exists?
+      return true
+    end
+
+    if TopicUser.where(
+          topic: self,
```
tgxworld: hmm 3 spaces are use for indentation here.

app/models/topic.rb
techAPJ: Hmm.. so what do you recommend should be an ideal behaviour here?

app/models/topic.rb
techAPJ: That is because Rubocop wants 4 lines here as per:

> Use 4 (not 2) spaces for indenting a condition in an if statement spanning multiple lines.

app/models/topic.rb
tgxworld: I think the user should still be invited but with notification level set to `muted`

app/models/topic.rb
tgxworld: o wow that is a weird rule.

app/models/topic.rb
techAPJ: So if a muted user invites another user to a certain topic that topic will be muted for invited user? They will not get anymore notifications from that topic just because a muted user invited them to that topic.

Also in case of PMs, inviting a user will create a small action post which will make other participants believe that invited user is aware of discussion/conversation, however since the invite was from a muted user the invited will not be aware of the conversation, and moreover the future mentions will not create any notifications if we set notification level to `muted`.

I think this discussion deserves a dedicated topic/post on meta?

PR #5640, opened 2018-03-01
Add an option to set users to active on mbox import

yanokwa:


tgxworld:
This looks OK to me but I'll differ to @gschlager to merge this.

gschlager:
@yanokwa Staged users should always be inactive. We had multiple commits in December were he had to fix issues with active, staged users. What's your use case that made you create this PR? Could you change it so that it changes the active flag only for regular users?

yanokwa:
Now I don't remember. I think I had forgotten to set staged users to false and so I had users that I had to manually unstage and then activate.

If you can confirm that non-staged users are active, then I don't think this PR is needed.

gschlager:
> If you can confirm that non-staged users are active, then I don't think this PR is needed.

That wasn't the case, but I changed it in https://github.com/discourse/discourse/commit/dc32ee5cbf5309bc0aad253618fad1e8c1da0891

Thanks for the PR anyway! 👍

PR #5641, opened 2018-03-01
single quote password in restore command

wjordan:
Followup to #3283. Quotes passwords passed to shell for backup restore.

The previous PR only fixed the `backup` process, this fixes the `restore` to match.


tgxworld:
Thanks @wjordan!

PR #5642, opened 2018-03-01
Fixed typo in page title

BadAllOff:
Fixed typo in page title


ZogStriP:
Спасибо!

gschlager:
That file is actually coming from [Transifex](https://www.transifex.com/discourse/discourse-org). So, @BadAllOff please make sure that it's updated there as well. Otherwise your changes will be overwritten the next time we pull translations from Transifex.

PR #5647, opened 2018-03-05
FIX: do not treat :: as a valid emoji

jjaffeux:


techAPJ:
Awesome, thank you! 👍

PR #5650, opened 2018-03-05
UX: Improvements for theme UI

OsamaSayegh:
As per Sam's request here: https://meta.discourse.org/t/add-support-for-theme-settings/78103/6?u=osama


SamSaffron:
Thanks!

PR #5654, opened 2018-03-07
FEATURE: Add user-profile-avatar-flair plugin outlet

tshenry:
Adds a plugin outlet in the user profile for avatar flair.


tshenry:
This outlet would be helpful in simplifying the methodology I use in creating a theme component. The [user-card-contents.hbs](https://github.com/discourse/discourse/blob/56352137caa3011e9a2241e3367a7072f3fb2a0e/app/assets/javascripts/discourse/templates/components/user-card-contents.hbs) has a similar outlet, so I'm hoping this isn't problematic to add.

nlalonde:
Having an outlet there makes sense to me since we have one on user cards (`user-card-avatar-flair`). 👍

PR #5657, opened 2018-03-07
FIX: Login redirect path was broken in subfolder installs

xfalcox:
@tgxworld can you check since you had a fix in this exact line?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -633,7 +633,7 @@ class ApplicationController < ActionController::Base
         else
           # save original URL in a cookie (javascript redirects after login in this case)
           cookies[:destination_url] = destination_url
-          redirect_to "/login"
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -633,7 +633,7 @@ def redirect_to_login_if_required
         else
           # save original URL in a cookie (javascript redirects after login in this case)
           cookies[:destination_url] = destination_url
-          redirect_to "/login"
+          redirect_to "#{Discourse.base_uri}/login"
```
ZogStriP: We have a `path` helper for that ;)

`redirect_to path("/login")`

tgxworld:
Merged in https://github.com/discourse/discourse/commit/2097f5330c7c9c364dbfe107d98461375a7ba740

Thanks @xfalcox and @ZogStriP

PR #5658, opened 2018-03-08
FIX: move external_id from UserSerializer to CurrentUserSerializer

scossar:
Moves the external_id from `UserSerializer` to `CurrentUserSerializer`.


SamSaffron:
Yes looks way better to me, thanks!

PR #5659, opened 2018-03-08
FEATURE: set 'Retry-After' header for 429 responses

xrav3nz:
> CONTEXT & DISCUSSION: https://meta.discourse.org/t/when-rate-limiting-is-hit-return-return-a-retry-after-header/81838?u=xrav3nz

I am assuming: 

`Retry-After` shouldn't be set in actions like the following, where the retry time is (intentionally) hidden from the end user.

https://github.com/discourse/discourse/blob/b57d4c87a416bfc6d77e7d1e5fd4717cd559dbf6/app/controllers/invites_controller.rb#L141-L152

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -107,7 +107,15 @@ class ApplicationController < ActionController::Base
   end
 
   def render_rate_limit_error(e)
-    render_json_error e.description, type: :rate_limit, status: 429, extras: { wait_seconds: e&.available_in }
+    retry_time_in_seconds = e&.available_in
+
+    render_json_error(
+      e.description,
+      type: :rate_limit,
+      status: 429,
+      extras: { wait_seconds: retry_time_in_seconds },
+      headers: { 'Retry-After': retry_time_in_seconds },
+    )
   end
 
   # If they hit the rate limiter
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -523,12 +531,15 @@ def custom_emoji
...
+    # obj       - a translated string, an ActiveRecord model, or an array of translated strings
     # opts:
-    #   type   - a machine-readable description of the error
-    #   status - HTTP status code to return
+    #   type    - a machine-readable description of the error
+    #   status  - HTTP status code to return
+    #   headers - extra headers for the response
     def render_json_error(obj, opts = {})
       opts = { status: opts } if opts.is_a?(Integer)
+      opts.fetch(:headers, []).each { |name, value| headers[name.to_s] = value }
```
tgxworld: I think this should be `opts.fetch(:headers, {})`

spec/integration/rate_limiting_spec.rb
```diff
@@ -7,10 +7,14 @@
   before do
     RateLimiter.enable
     RateLimiter.clear_all!
+
+    @old_value = Rails.application.config.consider_all_requests_local
+    Rails.application.config.consider_all_requests_local = false
```
tgxworld: Why do we need to disable this? The specs are still passing for me after removing this line. :thinking:

tgxworld:
Thank you @xrav3nz it looks good to me aside from some minor comments.

xrav3nz:
`consider_all_requests_local` needs to be false so Rails does not rescue the exception before we do. 

But you're right, it is passing without explicitly disabling it because the value was already set to false for the test environment.

https://github.com/discourse/discourse/blob/9b651adadb8457828dba29645fade516107a3b71/config/environments/test.rb#L13-L14

tgxworld:
Thank you so much @xrav3nz

PR #5663, opened 2018-03-09
FIX: Excerpts could be too long for column

gschlager:
<del>This changes the data type of `topics.excerpt` from `VARCHAR(1000)` to `TEXT`. If I'm [reading this correctly](https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.2#Reduce_ALTER_TABLE_rewrites) that's not a costly operation in Postgres.</del>

<del>Why? Because excerpts can contain an unknown number of HTML (e.g. from the `<details>`).
https://meta.discourse.org/t/excerpt-gets-too-long-under-certain-circumstances-activerecord-valuetoolong/77579</del>

Also, I made improvements to the text length calculation when the excerpt contains `<details>`.

### DIFF
--- a/lib/excerpt_parser.rb
+++ b/lib/excerpt_parser.rb
@@ -49,7 +49,8 @@ class ExcerptParser < Nokogiri::XML::SAX::Document
   end
 
   def include_tag(name, attributes)
-    characters("<#{name} #{attributes.map { |k, v| "#{k}=\"#{escape_attribute(v)}\"" }.join(' ')}>", false, false, false)
+    characters("<#{name} #{attributes.map { |k, v| "#{k}=\"#{escape_attribute(v)}\"" }.join(' ')}>",
+               truncate: false, count_it: false, encode: false)
   end
 
   def start_element(name, attributes = [])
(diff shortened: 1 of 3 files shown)


coding-horror:
Hmm, are we sure we want to change this? I feel like we're making a radical change based on a bizarre set of rare circumstances?

gschlager:
I can remove the migration. I'd say with my changes to how we calculate the excerpt we are pretty safe.
The example posted on meta has a excerpt length of 587 chars in the DB.

lib/excerpt_parser.rb
```diff
@@ -144,11 +144,19 @@ def end_element(name)
...
       @in_details_depth -= 1
       if @in_details_depth == 0
-        full = "<details><summary>#{clean(@summary_contents)}</summary>#{clean(@detail_contents)}</details>"
-        if @current_length + full.length > @length
-          @excerpt << "<details class='disabled'><summary>#{@summary_contents[0..@length]}</summary></details>"
+        @summary_contents = clean(@summary_contents)
+        @detail_contents = clean(@detail_contents)
+
+        if @current_length + @summary_contents.length >= @length
+          characters(@summary_contents, true, true, false,
```
ZogStriP: These `true, true, false` arguments should be made named parameters. Otherwise it gets hard to understand.

PR #5666, opened 2018-03-11
FIX: error when group_in_subject enabled but no group in pm

LeoMcA:


tgxworld:
Looks good to me. THank you @LeoMcA

PR #5667, opened 2018-03-12
FEATURE: replace PM tags dropdown with a dedicated tags page

techAPJ:


techAPJ:
@tgxworld can you review this PR? Thanks. :)

app/assets/javascripts/discourse/routes/user-private-messages-tags.js.es6
```diff
@@ -0,0 +1,21 @@
+import { ajax } from 'discourse/lib/ajax';
+
+export default Discourse.Route.extend({
+  model() {
+    return ajax('/tags/personal_messages').then(result => {
```
tgxworld: We should use a `catch` here to handle error cases.

spec/models/tag_spec.rb
```diff
@@ -114,7 +113,7 @@ def make_some_tags(count: 3, tag_a_topic: false)
 
     it "returns all pm tags if user is a staff and pm tagging is enabled" do
       SiteSetting.allow_staff_to_tag_pms = true
-      expect(described_class.pm_tags(guardian: Guardian.new(Fabricate(:admin)))).to match_array(@private_tags.map(&:name))
+      expect(described_class.pm_tags(guardian: Guardian.new(Fabricate(:admin))).length).to eq(2)
```
tgxworld: hmm testing for the length is not sufficient. We should also test for the objects we're expecting in the array.

tgxworld:
@techAPJ Some minor comments but generally it looks really good to me :+1:

techAPJ:
Thanks @tgxworld ❤️

PR #5669, opened 2018-03-12
Reset widget decorators after each test

gschlager:
Resetting widget decorators after each test should prevent build errors like the one we are seeing on Travis right now with the Cakeday plugin.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/widget.js.es6
+++ b/app/assets/javascripts/discourse/widgets/widget.js.es6
@@ -32,6 +32,10 @@ export function applyDecorators(widget, type, attrs, state) {
   return [];
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/widget.js.es6
```diff
@@ -32,6 +32,14 @@ export function applyDecorators(widget, type, attrs, state) {
   return [];
 }
 
+export function resetDecorators() {
+  for (var key in _decorators) {
```
eviltrout: Why not `Object.keys(_decorators).forEach(k => delete _decorators[key]);`? One line and much simpler.

PR #5677, opened 2018-03-15
FIX: prevents ruby warning with circular reference

jjaffeux:
@tgxworld can you look at this please ? I see no reason for `archive_user_id: archive_user_id` (and it's not valid) but maybe it was on purpose.


tgxworld:
Oops my bad. Thanks for fixing!

PR #5679, opened 2018-03-15
FIX: topic_tracking_state.publish_private_message parameter thinko

Supermathie:
@tgxworld: pretty sure this is what you meant here?


tgxworld:
hehe @jjaffeux fixed it too in https://github.com/discourse/discourse/commit/c0a986c1819c21aa6f4832f205f3f00e940b2650

PR #5684, opened 2018-03-19
Add theme stylesheet(s) to the crawler layout

angusmcleod:


SamSaffron:
I will merge this in post 2.0, but I wondered what @hnb-ku thinks about it? I guess it is not too much high risk and better than nothing.

hnb-ku:
I did a bit of testing @SamSaffron 

Since the html for the current theme is already added to the crawler view, this is definitely an improvement for themes that add **static html** 

user view custom theme: 
custom-html

Crawler view of the same theme before the changes:

custom-html-crawler-before

Crawler view after adding the stylesheet:

custom-html-crawler-after

All template modifications / scripts added in `<script type="text/discourse-plugin">` tags lead in errors in the current crawler view - before changes - because `Discourse` is not defined. 

So, any content added with the pluginAPI and any template modification in themes does not show up in the crawler view anyway and therefore cannot be styled.

There are some considerations for custom themes

Since JS customisations in discourse script tags are not loaded, any customisations that involves changing styles with JS  will remain in their initial styles set in the stylesheet that's loaded in the crawler view

If a theme sets an item to `display: none` in the stylesheet and then uses script to remove the property after some event has occurred, those changes will not happen in the crawler view and the item will continue to have the `display: none` property. 

Print view also seems to be slightly affected by adding the stylesheet to the crawler view.

Before changes

print-custom-theme-before

After changes:

custom-theme-print-after

While the print view strips out most styles, some of them make it in like custom fonts.

Since the html for the user and crawler views differ, some things will look a bit off.

Here's a user view for custom theme: 

custom-theme

And here's the current crawler view for the same theme:

crawler-custom-theme-before

And here's what the crawler view looks like after adding the sheet

custom-theme-crawler-after

I am definitely not in any position to make any comments about SEO (snake oil jokes aside) I could see cases where loading the stylesheet would lead to low text / background contrast. None of this happens intentionally of course, it's just a result of the different html structure for the user and crawler views   

I'm happy to do more testing if needed.

SamSaffron:
Thanks merging this in now!

PR #5688, opened 2018-03-20
FIX: Make sure a post has replies before accessing the reply_id

scossar:
If the reply to a post is moved to a new topic, embedded comments for the post will break, with the error `undefiled property ID for nil class`. This is because moving a post doesn't reset the reply_count for the post's parent. This PR just checks that the post has replies. It doesn't deal with resetting the reply_count.


nlalonde:
Looks like a good check to add.

PR #5690, opened 2018-03-21
FEATURE: Match spaces in search with underscores in text

majakomel:
Meta: https://meta.discourse.org/t/match-spaces-in-search-with-underscores-in-text/82795

### DIFF
--- a/spec/components/search_spec.rb
+++ b/spec/components/search_spec.rb
@@ -695,6 +695,13 @@ describe Search do
       expect(Search.execute('"3.0, eta is"').posts.length).to eq(0)
     end
 
+    it 'can match exact phrases when they are separated by other than spaces' do
+      topic = Fabricate(:topic, created_at: 3.months.ago)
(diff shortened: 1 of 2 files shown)


spec/components/search_spec.rb
```diff
@@ -695,6 +695,13 @@ def search
       expect(Search.execute('"3.0, eta is"').posts.length).to eq(0)
     end
 
+    it 'can match exact phrases when they are separated by other than spaces' do
+      topic = Fabricate(:topic, created_at: 3.months.ago)
+      Fabricate(:post, raw: 'this_is_a_test', topic: topic)
+
+      expect(Search.execute('"this is a test"').posts.length).to eq(1)
```
tgxworld: We should validate for the exact post here. Someone may add a `let!(:post) { Fabricate(:post) }` outside of this scope and may cause the test to incorrectly pass.

SamSaffron:
Sorry Maja, we decided against this change 😢

PR #5700, opened 2018-03-22
FEATURE: control web crawlers access with white/blacklist

nlalonde:
Something we may need to change since the whitelist can block all API requests if they don't include a whitelisted user agent: Currently we include requests without browser-like user agents as web crawlers, including "DiscourseAPI".

### DIFF
--- a/app/assets/javascripts/admin/models/report.js.es6
+++ b/app/assets/javascripts/admin/models/report.js.es6
@@ -137,11 +137,11 @@ const Report = Discourse.Model.extend({
 
   @computed('data')
   sortedData(data) {
-    return this.get('xaxisIsDate') ? data.toArray().reverse() : data.toArray();
+    return this.get('xAxisIsDate') ? data.toArray().reverse() : data.toArray();
   },
 
   @computed('data')
-  xaxisIsDate() {
+  xAxisIsDate() {
     if (!this.data[0]) return false;
     return this.data && moment(this.data[0].x, 'YYYY-MM-DD').isValid();
   }
--- a/app/assets/javascripts/admin/models/report.js.es6
+++ b/app/assets/javascripts/admin/models/report.js.es6
@@ -2,6 +2,7 @@ import { ajax } from 'discourse/lib/ajax';
 import round from "discourse/lib/round";
 import { fmt } from 'discourse/lib/computed';
 import { fillMissingDates } from 'discourse/lib/utilities';
+import computed from 'ember-addons/ember-computed-decorators';
 
 const Report = Discourse.Model.extend({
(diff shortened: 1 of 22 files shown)


lib/middleware/request_tracker.rb
```diff
@@ -155,12 +157,20 @@ def log_request_info(env, result, info)
...
 
-    if rate_limit(env)
+    if rate_limit(request)
       result = [429, {}, ["Slow down, too Many Requests from this IP Address"]]
       return result
     end
 
+    if block_crawler(request)
+      log_request = false
+      result = [403, {}, []]
```
SamSaffron: we probably want some sort of message explaining this here

lib/middleware/request_tracker.rb
```diff
@@ -266,6 +276,14 @@ def rate_limit(env)
     end
   end
 
+  def block_crawler(request)
+    request.get? &&
+      !request.xhr? &&
```
SamSaffron: I wonder if we should be a bit more atomic in our crawler blocking, but we can defer on this for now and look at numbers first

app/models/web_crawler_request.rb
```diff
@@ -0,0 +1,76 @@
...
+      write_cache!(Time.now.utc)
+      write_cache!(Time.now.utc.yesterday)
+      return
+    end
+
+    self.last_flush = Time.now.utc
+
+    date = date.to_date
+
+    $redis.smembers(user_agent_list_key(date)).each do |user_agent, _|
```
SamSaffron: hmmm how, big is this set getting? should we not be removing all committed rows as we go from the set in an single eval?

nlalonde: Yeah we can remove the elements from the set as we write with `spop`. I'll give that a try.

lib/middleware/request_tracker.rb
nlalonde: 👍

app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -114,21 +120,31 @@ const Report = Discourse.Model.extend({
...
+    return this.get('xaxisIsDate') ? data.toArray().reverse() : data.toArray();
+  },
 
-  dataReversed: function() {
-    return this.get("data").toArray().reverse();
-  }.property("data")
+  @computed('data')
+  xaxisIsDate() {
+    if (!this.data[0]) return false;
+    return this.data && moment(this.data[0].x, 'YYYY-MM-DD').isValid();
```
nlalonde: `moment('Wotbox/2.01 (+http://www.wotbox.com/bot/)', 'YYYY-MM-DD').isValid() === true`  :cry:

PR #5701, opened 2018-03-23
FEATURE: add profile_background and card_background support for Discourse SSO

misaka4e21:
Synchronize `UserProfile.profile_background` and `UserProfile.card_background` against a remote site with Discourse SSO.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -1,5 +1,6 @@
 require_dependency 'rate_limiter'
 require_dependency 'single_sign_on'
+require_dependency 'url_helper'
 
 class SessionController < ApplicationController
   class LocalLoginNotAllowed < StandardError; end
@@ -53,6 +54,26 @@ class SessionController < ApplicationController
         sso.admin = current_user.admin?
         sso.moderator = current_user.moderator?
         sso.groups = current_user.groups.pluck(:name).join(",")
+
+        sso.avatar_url = Discourse.store.cdn_url UrlHelper.absolute(
+          "#{Discourse.store.absolute_base_url}/#{Discourse.store.get_path_for_upload(current_user.uploaded_avatar)}"
+        ) unless current_user.uploaded_avatar.nil?
+        sso.profile_background_url = UrlHelper.absolute upload_cdn_path(
+          current_user.user_profile.profile_background
+        ) if current_user.user_profile.profile_background.present?
+        sso.card_background_url = UrlHelper.absolute upload_cdn_path(
+          current_user.user_profile.card_background
+        ) if current_user.user_profile.card_background.present?
+
+        sso.avatar_url = Discourse.store.cdn_url UrlHelper.absolute(
+          "#{Discourse.store.absolute_base_url}/#{Discourse.store.get_path_for_upload(current_user.uploaded_avatar)}"
+        ) unless current_user.uploaded_avatar.nil?
+        sso.profile_background_url = UrlHelper.absolute upload_cdn_path(
+          current_user.user_profile.profile_background
+        ) if current_user.user_profile.profile_background.present?
+        sso.card_background_url = UrlHelper.absolute upload_cdn_path(
+          current_user.user_profile.card_background
+        ) if current_user.user_profile.card_background.present?
 
         if sso.return_sso_url.blank?
           render plain: "return_sso_url is blank, it must be provided", status: 400
--- a/app/jobs/regular/download_profile_background_from_url.rb
+++ b/app/jobs/regular/download_profile_background_from_url.rb
@@ -1,0 +1,28 @@
+module Jobs
+
+  class DownloadProfileBackgroundFromUrl < Jobs::Base
+    sidekiq_options retry: false
+
+    def execute(args)
+      url = args[:url]
+      user_id = args[:user_id]
+
+      raise Discourse::InvalidParameters.new(:url) if url.blank?
+      raise Discourse::InvalidParameters.new(:user_id) if user_id.blank?
+
+      return unless user = User.find_by(id: user_id)
+
+      begin
+        UserProfile.import_url_for_user(
+          url,
+          user,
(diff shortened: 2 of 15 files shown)


riking:
You included rebased committs from the Discourse developers in this PR - try rebasing against master again.

misaka4e21:
Hello there, I did the rebase and fixed the format. Is there anything I could do in order to make this PR merged?

app/jobs/regular/download_profile_background_from_url.rb
```diff
@@ -0,0 +1,28 @@
+module Jobs
+
+  class DownloadProfileBackgroundFromUrl < Jobs::Base
```
techAPJ: We should add a test for this job.

app/models/discourse_single_sign_on.rb
```diff
@@ -203,10 +221,36 @@ def change_external_attributes_and_override(sso_record, user)
       end
     end
 
+    profile_background_missing = !user.user_profile.profile_background.present?
```
techAPJ: No need for this variable. What you are looking for is `user.user_profile.profile_background.blank?`

app/models/discourse_single_sign_on.rb
```diff
@@ -203,10 +221,36 @@ def change_external_attributes_and_override(sso_record, user)
...
+      if profile_background_changed || profile_background_missing
+        Jobs.enqueue(:download_profile_background_from_url,
+            url: profile_background_url,
+            user_id: user.id,
+            is_card_background: false
+        )
+      end
+    end
+
+    card_background_missing = !user.user_profile.card_background.present?
```
techAPJ: Same here. `user.user_profile.card_background.blank?`

app/models/discourse_single_sign_on.rb
```diff
@@ -161,13 +161,31 @@ def match_email_or_create_user(ip_address)
           )
         end
 
+        if profile_background_url.present?
```
techAPJ: There should be test cases for changes made here. For reference, see: https://github.com/discourse/discourse/blob/6f40037ba551e6e495e6f51263940b574721e66a/spec/models/discourse_single_sign_on_spec.rb#L446-L552

app/models/user_profile.rb
```diff
@@ -75,6 +76,36 @@ def rebake!
     update_columns(bio_cooked: cooked, bio_cooked_version: BAKED_VERSION)
   end
 
+  def self.import_url_for_user(background_url, user, options = nil)
```
techAPJ: Test case for this method please? :)

For reference: https://github.com/discourse/discourse/blob/ee69d58a59c824a7a0b1146710930c5861d3ca26/spec/models/user_avatar_spec.rb#L19-L62

app/controllers/session_controller.rb
```diff
@@ -53,6 +53,10 @@ def sso_provider(payload = nil)
         sso.admin = current_user.admin?
         sso.moderator = current_user.moderator?
         sso.groups = current_user.groups.pluck(:name)
+        base_url = Discourse.base_url
+        sso.avatar_url = base_url + current_user.avatar_template.gsub('{size}', '360')
```
techAPJ: Is this change related to this PR? Also, there should be a test to verify this change.

misaka4e21: I think `avatar_url` is not provided in `sso_provider`, causing the existing code for sso client unable to grab avatar from sso provider. Please let me know if I'm wrong or it was added to another place. I also wondered if `avatar_url`, `profile_background_url` and `card_background_url` should be provided by the sso provider.

I added all of these test cases just now.

techAPJ:
The PR looks good, but I would prefer if someone else from the team (maybe @ZogStriP ?) approve these changes as well.

app/controllers/session_controller.rb
ZogStriP: Does this also work when uploads are stored on S3?

app/jobs/regular/download_profile_background_from_url.rb
```diff
@@ -0,0 +1,28 @@
...
+      return unless user = User.find_by(id: user_id)
+
+      begin
+        UserProfile.import_url_for_user(
+          url,
+          user,
+          is_card_background: args[:is_card_background],
+        )
+      rescue Discourse::InvalidParameters => e
+        raise e unless e.message == 'url'
```
ZogStriP: What does this `e.message == 'url'` do?

misaka4e21:
Hello there…
Command `RAILS_ENV=test bundle exec rake spec` passes normally on my computer; however, it fails on travis-ci each time, even with the same ruby version 2.3.4p301 of my developing environment.

I'll try writing a new spec rule against S3 uploading then.

misaka4e21:
I did edit `spec/controllers/session_controller_spec.rb` to let it pass the unit test on travis-ci.
I think that S3 upload storage should work, though I cannot test it on a real installation.

misaka4e21:
@ZogStriP 
Is there any problem to be solved or to be explained?

misaka4e21:
I have cleaned the commit history.
Sorry for the mistake that I once merged (not rebase against) master into the commit.

app/models/discourse_single_sign_on.rb
```diff
@@ -189,14 +189,47 @@ def match_email_or_create_user(ip_address)
...
+
+          if card_background_url.present?
+            Jobs.enqueue(:download_profile_background_from_url,
+              url: card_background_url,
+              user_id: user.id,
+              is_card_background: true
+            )
+          end
+
+          if profile_background_url.present?
```
ZogStriP: Why is this being repeated?

misaka4e21: I made this incorrect when I was rebasing the tree against master to create a clean commit history. The problem is already fixed currently.

ZogStriP:
LGTM. Thanks 👍

PR #5704, opened 2018-03-24
Add 'topicTags' attribute to structure that is passed to callbacks re…

astruyk:
…gistered with the 'api.addPostTransformCallback(...)' function so they are available there.

Making these available allows you to do special formatting of posts based on tags applied to the topic without having to resort to hacks (e.g. checking for `.tag-<tagname>` styles in the page, or including the entire `topic` object through a `api.includePostAttributes('topic')` call).

### DIFF
--- a/app/assets/javascripts/discourse/lib/transform-post.js.es6
+++ b/app/assets/javascripts/discourse/lib/transform-post.js.es6
@@ -121,6 +121,7 @@ export default function transformPost(currentUser, site, post, prevPost, nextPos
   postAtts.userCustomFields = post.user_custom_fields;
   postAtts.topicUrl = topic.get('url');
   postAtts.isSaving = post.isSaving;
+  postAtts.topicTags = topic.tags;
 
   const showPMMap = topic.archetype === 'private_message' && post.post_number === 1;
   if (showPMMap) {


app/assets/javascripts/discourse/lib/transform-post.js.es6
```diff
@@ -121,6 +121,7 @@ export default function transformPost(currentUser, site, post, prevPost, nextPos
   postAtts.userCustomFields = post.user_custom_fields;
   postAtts.topicUrl = topic.get('url');
   postAtts.isSaving = post.isSaving;
+  postAtts.topicTags = topic.tags;
```
tgxworld: This seems fine to me but we should only assign this if tags are enabled on the site.

SamSaffron: wait, why do we need this on every single post? we don't even have category from what I can see.

something does not feel right here if we are passing tags around but not category. @eviltrout thoughts?

SamSaffron:
Sorry, this is not sitting right for me, can you do a topic on meta, we do not pass category along like this so I see no reason to give tags a free ride.

PR #5705, opened 2018-03-26
FEATURE: Webhook for group and category events

vinothkannans:


app/models/group.rb
```diff
@@ -571,6 +574,16 @@ def self.owner_of(groups, user)
     self.member_of(groups, user).where("gu.owner")
   end
 
+  def trigger_group_created_event
+    DiscourseEvent.trigger(:group_created, self)
```
tgxworld: Hmm should we skip the trigger for automatic groups?

app/models/category.rb
```diff
@@ -511,6 +514,16 @@ def self.find_by_slug(category_slug, parent_category_slug = nil)
   def subcategory_list_includes_topics?
     subcategory_list_style.end_with?("with_featured_topics")
   end
+
+  def trigger_category_created_event
+    DiscourseEvent.trigger(:category_created, self)
```
tgxworld: Do we need this for uncategorized or default categories that are seeded?

spec/models/group_spec.rb
```diff
@@ -411,17 +411,37 @@ def real_staff
...
+      event = DiscourseEvent.track_events { subject.save! }.first
+
+      expect(event[:event_name]).to eq(:group_created)
+      expect(event[:params].first).to eq(subject)
+    end
+  end
+
+  describe 'destroy' do
+    before do
+      @group = Fabricate(:group)
```
tgxworld: While instance variables are OK. I think we should just standardize and use `let`.

spec/models/group_spec.rb
```diff
@@ -411,17 +411,37 @@ def real_staff
...
+      @user = Fabricate(:user)
+      @group.add(@user)
+      @group.save!
```
tgxworld: `Group#add` has already saved the record once so I don't think we need to save here :)

spec/models/group_spec.rb
```diff
@@ -411,17 +411,37 @@ def real_staff
...
+    end
 
-    g.destroy
+    it "it deleted correctly" do
+      @group.destroy
```
tgxworld: Generally it is preferred to use `destroy!` over `destroy` anytime you're not checking for the return value

spec/models/group_spec.rb
```diff
@@ -411,17 +411,37 @@ def real_staff
...
+      expect(User.where(id: @user.id).count).to eq 1
+      expect(GroupUser.where(group_id: @group.id).count).to eq 0
+    end
+
+    it 'triggers a extensibility event' do
+      event = DiscourseEvent.track_events { @group.destroy }.first
```
tgxworld: Same as https://github.com/discourse/discourse/pull/5705/files#r176991267

app/controllers/admin/groups_controller.rb
```diff
@@ -55,7 +55,10 @@ def update
 
     # group rename is ignored for automatic groups
     group.name = group_params[:name] if group_params[:name] && !group.automatic
-    save_group(group) { |g| GroupActionLogger.new(current_user, g).log_change_group_settings }
+    save_group(group) do |group|
+      GroupActionLogger.new(current_user, group).log_change_group_settings
+      DiscourseEvent.trigger(:group_updated, group)
```
tgxworld: lets test that the event is triggered in the controllers as well :)

tgxworld:
@vinothkannans Aside from some minor comments, this PR looks good to me. Are we planning to expose the web hooks on the admin interface to allow users to enable/disable the webhooks?

vinothkannans:
@tgxworld yes we should. how I forgot the front-end part :) let me do

app/controllers/admin/groups_controller.rb
vinothkannans: it's added :+1:

app/models/group.rb
vinothkannans: We already have default `system` and `discobot` users along with [`user_created` event](https://github.com/discourse/discourse/blob/35745166b5a4d1f48b206192b21d8f8b44be67ef/app/models/user.rb#L1193-L1196) without any issues. So I think it won't be a problem.

app/models/category.rb
vinothkannans: same as https://github.com/discourse/discourse/pull/5705#discussion_r177211381

vinothkannans:
@tgxworld I made the changes 👍.

PR #5707, opened 2018-03-26
Improve protection against accidental drops and renames

gschlager:
Most of the changes are improvements of and additions to the specs.

* Protects against table renames and show the correct error message.
* Improves the specs. Some of them were always green, no matter if the drops were actually prevented or not.
* Immediately run delayed migrations if the first migration was run < 10 minutes ago.
* Drop tables only if they actually exist. Dropping columns already had that check.


SamSaffron:
Looks good and very well tested, thanks!

PR #5710, opened 2018-03-27
FIX: discobot fails when max_emojis_in_title=0

Supermathie:
* If discobot is enabled but max_emojis_in_title==0,
  try to strip emoji from the title when creating a new post


Supermathie:
Fixes error seen in prod:

Job exception: Failed to create post: ["Title can't have more than 0 emoji"]
in Jobs::NarrativeInit

plugins/discourse-narrative-bot/lib/discourse_narrative_bot/new_user_narrative.rb
```diff
@@ -182,8 +182,13 @@ def say_hello
       #{instance_eval(&@next_instructions)}
       RAW
 
+      title = I18n.t("#{I18N_KEY}.hello.title", title: SiteSetting.title)
+      if SiteSetting.max_emojis_in_title == 0
+        title = title.gsub(/:\w+:/, '').strip
```
ZogStriP: - Are emojis in `title` always in their unescaped representation? (as opposed to their Unicode representation)
- What about `:mermaid:t4:`?

Supermathie: re the 1st point: primary concern is the default titles as we ship (which all use `:robot:`) - if somebody is modifying the title on their own site they are probably aware of their own site setting restrictions.

PR #5711, opened 2018-03-27
FIX: CDN_URL hostname should be in GlobalSetting.hostnames

Supermathie:
This avoids the need to specify the CDN_URL hostname as a DISCOURSE_BACKUP_HOSTNAME in the container configuration.


tgxworld:
Looks good to me @Supermathie :+1:

PR #5712, opened 2018-03-28
Remove use of `rescue nil`.

tgxworld:
* `rescue nil` is a really bad pattern to use in our code base.
  We should rescue errors that we expect the code to throw and
  not rescue everything because we're unsure of what errors the
  code would throw. This would reduce the amount of pain we face
  when debugging why something isn't working as expected. I've
  been bitten countless of times by errors being swallowed as a
  result during debugging sessions.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -390,7 +390,13 @@ class Admin::UsersController < Admin::AdminController
     ip = params[:ip]
 
     # should we cache results in redis?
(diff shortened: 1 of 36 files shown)


coding-horror:
I think a better way to look at this is *almost never* and when you do,
leave a comment explaining why. There are still times when it can be
necessary, but it should be quite rare. Just like “nobody should ever use
goto”, yes that is 99% true and yet.. only a sith deals in absolutes.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/142571bba010eedbdfc1452d42beccc72389c373

PR #5716, opened 2018-03-28
fixed typo in translation

bald:


nlalonde:
To work with us on translations, join this project:
https://www.transifex.com/projects/p/discourse-org/

bald:
Could someone change the translations from this patch in transifex? I don't want to setup an account for this minor things.

PR #5722, opened 2018-03-29
Update instructions to use correct ruby version

mdoyle13:
Bundler will not finish installing gems under Ruby 2.0.0. The instructions should maybe use the minimum ruby version in the travis.yml file.


tgxworld:
Thanks @mdoyle13

PR #5723, opened 2018-03-31
Feature: Add unseen-topic and new-posts classnames to topic list items

featheredtoast:
Adds classname descriptions for new and unread posts on topic-list-items.

Allows for new/unread topic-list-items to be targeted with css directly, making the post list items easier to theme.

Ran into this when experimenting with some interesting theming ideas that requires css selection for new/unread post list rows themselves and thought it might be useful for other themes as well.


tgxworld:
This seems OK to me :+1:

PR #5724, opened 2018-03-31
Log to stdout when configured

pierreozoux:
First contribution here, thanks for your help!

The goal of the PR is to make rails log everything to stdout. This is desired in some environment and a good practice recommended by [12factorapp](https://12factor.net/) for instance.

I checked quickly the code, and there is another possibility, make it a lig provider, and put this logic in a separate initializer.

Please let me know, and I'll update the PR accordingly.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -21,6 +21,11 @@ Discourse::Application.configure do
   config.assets.digest = true
 
   config.log_level = :info
+  
+  # log to stdout when configured
+  if ENV['LOG_STDOUT']
+    config.logger = Logger.new(STDOUT)
+  end
 
   if GlobalSetting.smtp_address
     settings = {


tgxworld:
@SamSaffron  What do you think about this?

config/environments/production.rb
```diff
@@ -21,6 +21,11 @@
   config.assets.digest = true
 
   config.log_level = :info
+  
+  # log to stdout when configured
+  if ENV['LOG_STDOUT']
+    config.logger = Logger.new(STDOUT)
```
tgxworld: You probably want to do this after `100-logster` initializer.

SamSaffron:
I am not sure about this ... if anything this should be a global setting so you would pass it in with DISCOURSE_LOG_TO_STDOUT that is the pattern we use for configuring this, the position of this needs to be fixed as well.

SamSaffron:
closing this for now, but feel free to reopen as a global.

polarathene:
@SamSaffron @pierreozoux Did anything come of this? was there a new PR for supporting sending logs to stdout?

Considering that Docker is the only officially supported install method afaik, it strikes me as a bit odd that we can't manage logs via the standard support for it with containers.

SamSaffron:
It is not a trivial change to do cleanly, the container is streaming multiple logs ... rails logs, nginx logs, sidekiq logs, unicorn stderr and out. Consuming this cleanly from the wrapped JSON consumer provided by docker would require a fair amount of internal shuffling. 

Splitting into NGINX/Ruby containers is a fair piece of work cause some getting all the mounts right is a tricky thing especially since you want to support sendfile. Splitting Ruby into unicorn / sidekiq is wasteful cause memory is reused at the moment between unicorn master and worker.

polarathene:
> Consuming this cleanly from the wrapped JSON consumer provided by docker would require a fair amount of internal shuffling.

Fair enough. 

Though if each log file has some identifier(or a way to identify that it's rails/nginx/sidekiq/etc), you could filter / search through that easily enough. While dockers default logging driver is JSON, it does support other drivers like journald as the sidekiq quote I provided mentions using. But I guess you'd run into the same issue regardless if there is no common identifier to filter the logs with.

It seems the alternatives for handling logs via such containers, is to use an additional container to manage logs via a shared volume. remote_syslog2 can handle this, or as pointed out by [k8s docs](https://kubernetes.io/docs/concepts/cluster-administration/logging/#using-a-sidecar-container-with-the-logging-agent), "sidecar" containers that tail a file and output to stdout is another workaround.

> Splitting Ruby into unicorn / sidekiq is wasteful cause memory is reused at the moment between unicorn master and worker.

As in the separate processes access the same memory regions? Since Ruby 2.0, [any forking should let child processes share memory of the parent via CoW](https://marianposaceanu.com/articles/on-ruby-2-0-memory-usage-unicorn-and-heroku)?

If you meant in regards to [page cache](https://en.wikipedia.org/wiki/Page_cache) where files/libraries are read into memory. Those page caches are shared across containers when the files are from the same shared image layer, but this support is dependent on what storage driver is used.

As I'm not a Ruby guy, I don't know what memory re-use you're referring to with [unicorn forking sidekiq](https://meta.discourse.org/t/why-does-discourse-have-to-use-docker/26041/32). Is it benefiting from the memory used by unicorn in some way?

> Splitting into NGINX/Ruby containers is a fair piece of work cause some getting all the mounts right is a tricky thing especially since you want to support sendfile.

Could you elaborate on that a bit more? What is tricky with getting the mounts right for you? 

I tried searching the internet for whatever it was you were referring to(assuming specific to nginx and sendfile), but didn't have any luck other than an [old post by you on the Discourse forums](https://meta.discourse.org/t/can-discourse-ship-frequent-docker-images-that-do-not-need-to-be-bootstrapped/33205/50). Are you referring to the `try_files` paths in the nginx conf?

---

### Kinda off-topic:

That said I generally don't have much of an issue with a web container that has both nginx and the webapp in the same image. Separation is nice when there is value in it, such as:

- utilizing/sharing the service/image with other projects(which also has the benefit mentioned earlier with file and memory savings), also useful for scaling out instances/resources based on load.
- having more control over the version/config and a clear separation of what part of the project is specifically for it alone.
- the service / concern managed by the container is not tightly coupled(in most cases, you can switch out mysql for mariadb for example, or nginx and apache). The benefit may be less obvious to an individual project, but when you want to deploy multiple projects and they lock in a bias, you can lose the benefit of just working with a single service provider(nginx vs apache rather than juggle both).

Another reason would be to favour using the official images off DockerHub, just changing the configs via flags/options when running the container(env vars, mounts, configs, ports, etc), or when that isn't sufficient, extending an official image as a base, or using multi-stage builds.

Not only does that provide a clearer picture of what a project is specifically doing differently, it keeps consistency elsewhere with any other project using that service, there is more trust in the official images and less time/effort required to inspect what an individual project does when they go more DIY as if it were bare-metal or VM.

Likewise, official images often offer not only different versions(which could be sync'd across projects when appropriate), but also different bases, such as Alpine(sometimes good, sometimes bad), Ubuntu or Debian(with slim variant). A developer can more confidently make any additional changes on either end this way too, so this is more flexible.

---

Regarding Discourse, I understand why the official/primary deployment with Docker is a single image. I've read about the issues with PG upgrades and how Discourse is approaching it among other things(like updating nginx config via web UI integration). This is great as it gets a reliable/working setup out of the box and requires far less experience from the user to get Discourse up and running, along with maintaining it and receiving support.

Splitting it up by service / concern further, especially regarding the database is desirable for those more experienced with using Docker and doing more than just deploying Discourse(and are willing to handle the additional effort to maintain). There's evident interest in it, and while I understand that the Discourse team does not want to maintain two separate docker approaches and thus won't adopt such, it would be great if the community had some unofficial support this that interested users could be directed to. 

Judging by what is going on in the [base Dockerfile](https://github.com/discourse/discourse_docker/blob/c2c7a3d8f3aad26b0b1aea30eb5bf475d910ebc2/image/base/Dockerfile), there isn't a clear separation, so unless you are quite familiar with the project, it's not entirely clear what is there for what purpose. How much of that expects x86_64?

You're using Ubuntu 16.04, which according to git blame hasn't been updated for 3 years, and that was not long after the release came out for it. I assume it hasn't been updated since for some reason, perhaps because it might affect other parts of the Dockerfile requiring more time investment to maintain?  You appear to install binary version of Node 10 from nodesource, but then also install nodejs from the 16.04 package manager after? 

uglify has been pinned to 2.x since 2017, [it's not clear why](https://github.com/discourse/discourse_docker/pull/348), there is no development on that since mid 2017, while there is a uglify-js v3, terser(has some prior uglify devs iirc) seems to have been adopted by the community more now(webpack/parcel/microbundle).

polarathene:
Somehow, since after posting this, [one of the links in my prior comment](https://meta.discourse.org/t/can-discourse-ship-frequent-docker-images-that-do-not-need-to-be-bootstrapped/33205/50) here is no longer available, instead rendering:

> Oops! That page doesn’t exist or is private.

PR #5731, opened 2018-04-02
FEATURE: Theme settings in handlebars templates

OsamaSayegh:
Meta topic: https://meta.discourse.org/t/add-support-for-theme-settings/78103/16?u=osama


SamSaffron:
❤️ 

I like this a lot, but we got to have some sort of test even if it is just checking that we actually add the text we need to the hbs file

OsamaSayegh:
Just added a spec. Also I noticed that this feature wasn't working when overriding a .raw template so I had to register the helper for a second time using `Handlebars.registerHelper` to make it work.

SamSaffron:
Cool @OsamaSayegh  ❤️  lets try this out!

PR #5741, opened 2018-04-05
Add wizard_js to translations that can be overridden from a plugin

angusmcleod:
Aside from making it possible to override ``wizard_js`` translations from a plugin, this change will also make the default locale fallback to English work for ``wizard_js`` translations added in a plugin.

See further: https://meta.discourse.org/t/custom-wizard-plugin/73345/78?u=angus


gschlager:
Thanks! 👍

PR #5742, opened 2018-04-06
Little Mistake

SidVal:
Change `yandax, bing and yahoo` to `Yandax, Bing and Yahoo`


SidVal:
> Discourse CLA
> Thank you!   
> Your response has been recorded.

gschlager:
@SamSaffron Do we even need to mention those companies? I think we can remove "respected by yandax, bing and yahoo"

coding-horror:
One of them is not spelled correctly as well

SamSaffron:
Oddly enough google are the only large search company that ignores this
setting, I guess we could call that out or just remove the company names,
not sure what the best copy change is Jeff can make the call

caugner:
See also #5745 (didn't see this PR before; my PR spells "Yandex" correctly though – and adds a full stop).

@SamSaffron Note that Baidu [seems to ignore the `crawl-delay` directive](https://translate.google.com/translate?sl=auto&tl=en&js=y&prev=_t&hl=en&ie=UTF-8&u=https%3A%2F%2Fbaike.baidu.com%2Fitem%2FCrawl-delay&edit-text=&act=url) as well, so it's not just Google.

SamSaffron:
OK I just stopped calling out the search engines, people will need to do own research, @jomaxro when cleaning up we can link it to a relevant topic on meta that explains who respects this and who does not.

PR #5744, opened 2018-04-06
Differentiate one/other in I18n for select-kit’s hasReached(Max|Min)imum

caugner:
The I18n keys `js.select_kit.min_content_not_reached` and
`js.select_kit.max_content_reached` did not differentiate between
one or more `item(s)`, which can be difficult to translate.

While it works in English: `You can only select {{count}} item(s).`
It doesn't work e.g. in German: `Du kannst nur {{count}} Eintrag/Einträge auswählen.`

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1251,8 +1251,12 @@ en:
       no_content: No matches found
       filter_placeholder: Search...
       create: "Create: '{{content}}'"
-      max_content_reached: "You can only select {{count}} item(s)."
-      min_content_not_reached: "Select at least {{count}} item(s)."
+      max_content_reached:
+        one: "You can only select {{count}} item."
+        other: "You can only select {{count}} item(s)."
+      min_content_not_reached:
+        one: "Select at least {{count}} item."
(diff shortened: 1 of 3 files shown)


caugner:
I have to admit that I didn't test that this actually works without adapting [select-kit.js.es6](https://github.com/discourse/discourse/blob/e33aef729f412ff2a195e80c060d4859ebb75cdf/app/assets/javascripts/select-kit/components/select-kit.js.es6#L228-L239), but I'm assuming it works out of the box like it does for `js.links_lowercase` in [topic-map.js.es6](https://github.com/discourse/discourse/blob/ee9be65b2c8e22d734459877694cc4171e4e5ea1/app/assets/javascripts/discourse/widgets/topic-map.js.es6#L120).

caugner:
Two tests did indeed fail due to the changed I18n value:
https://travis-ci.org/discourse/discourse/jobs/363308224#L7952

I have adapted the expectation of those tests accordingly.

jjaffeux:
I have tests for maximum/maximumLabel in a branch I will merge it later, arpit needed this feature quite urgently. I will merge this commit thanks.

PR #5746, opened 2018-04-08
Changes for theme-creator

davidtaylorhq:
FAO @SamSaffron 

Some changes to client-side theme models and controllers to make them easier to re-use for user created themes.


SamSaffron:
this looks fine to me! thanks!

PR #5758, opened 2018-04-12
FEATURE: theme settings in RAW hbs templates

OsamaSayegh:
https://meta.discourse.org/t/add-support-for-theme-settings/78103/21?u=osama

### DIFF
--- a/app/assets/javascripts/discourse/helpers/theme-setting-injector.es6
+++ b/app/assets/javascripts/discourse/helpers/theme-setting-injector.es6
@@ -9,12 +9,28 @@ function inject(context, key, value) {
     value = value.replace(/\\u0022/g, '"');
   }
 
+  if (!(context instanceof Ember.Object)) {
+    injectPlainObject(context, key, value);
+    return;
+  }
+
   if (!context.get("themeSettings")) {
     context.set("themeSettings", {});
   }
   context.set(`themeSettings.${key}`, value);
 }
 
+function injectPlainObject(context, key, value) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/helpers/theme-setting-injector.es6
```diff
@@ -9,12 +9,28 @@ function inject(context, key, value) {
...
+
   if (!context.get("themeSettings")) {
     context.set("themeSettings", {});
   }
   context.set(`themeSettings.${key}`, value);
 }
 
+function injectPlainObject(context, key, value) {
+  if (!context.themeSettings) {
+    Object.assign(context, { themeSettings: {} });
```
jjaffeux: IE11 has not support of Object.assign AFAIK

app/assets/javascripts/discourse/helpers/theme-setting-injector.es6
```diff
@@ -9,12 +9,28 @@ function inject(context, key, value) {
...
+  }
+  Object.assign(context.themeSettings, { [key]: value });
```
jjaffeux: IE11 has not support of Object.assign AFAIK

OsamaSayegh:
@jjaffeux thanks for pointing that out. I've switched to using underscore `assign` which I've verified works with IE 11.

SamSaffron:
OK, lets try this out @vinothkannans  can you confirm it is all working for you now! 

Thanks Osama

PR #5766, opened 2018-04-16
 Some space btw text

SidVal:
I think it lacks a space between the variables


tgxworld:
Thanks for the @SidVal  but the `span`s have margins that ensures the word and count do not stick together.

PR #5771, opened 2018-04-19
FIX: TopicEmbed.import should update title and author

xrav3nz:
We ran into a problem where `PollFeed` job didn't update the post author with the one found in the RSS feed.

Then I found out `TopicEmbed.import` only updates the post body and neglects new title or author, when a post with the same `embed_url` already exists.

---

Consider the scenario:

1. A blog page with Discourse embedded comment hits `embed_controller#comments`
2. The topic doesn't exist yet, so it tries to fetch the content from RSS feed, but it didn't because feed polling is rate limited
4. It then crawls the webpage and creates a topic with [the default `embed_by_username`](https://github.com/discourse/discourse/blob/ab80986184a498f68d99541c63558ea2089f49d3/lib/topic_retriever.rb#L48)
5. `PollFeed` is schedule to run at a later time, and it fetches the title, content, and author from the RSS feed with [`embed_username_key_from_feed`](https://github.com/discourse/discourse/blob/ab80986184a498f68d99541c63558ea2089f49d3/app/jobs/scheduled/poll_feed.rb#L156).
6. The post should be updated with the title, content and author from the feed.

---

Let me know what you think, cheers! 💪

### DIFF
--- a/spec/models/topic_embed_spec.rb
+++ b/spec/models/topic_embed_spec.rb
@@ -41,8 +41,12 @@ describe TopicEmbed do
       end
 
       it "Supports updating the post" do
-        post = TopicEmbed.import(user, url, title, "muhahaha new contents!")
+        new_user = Fabricate(:user)
+
+        post = TopicEmbed.import(new_user, url, title, "muhahaha new contents!")
+
         expect(post.cooked).to match(/new contents/)
+        expect(post.user).to eq(new_user)
       end
 
       it "Should leave uppercase Feed Entry URL untouched in content" do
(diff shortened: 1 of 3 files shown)


spec/models/topic_embed_spec.rb
```diff
@@ -41,8 +41,12 @@
       end
 
       it "Supports updating the post" do
-        post = TopicEmbed.import(user, url, title, "muhahaha new contents!")
+        new_user = Fabricate(:user)
+
+        post = TopicEmbed.import(new_user, url, title, "muhahaha new contents!")
```
SamSaffron: don't we want to test the title is updated as well? will merge this in and add that.

SamSaffron:
Thanks this looks good to me @eviltrout fyi

PR #5776, opened 2018-04-20
FEATURE: Add option to delete all replies of flagged post

gschlager:
Shows a dialog when the flagged posts gets deleted, it isn't the first post in a topic and it has replies.

image

### DIFF
--- a/app/assets/javascripts/admin/models/flagged-post.js.es6
+++ b/app/assets/javascripts/admin/models/flagged-post.js.es6
@@ -62,11 +62,74 @@ export default Post.extend({
   },
 
   deferFlags(deletePost) {
-    return ajax('/admin/flags/defer/' + this.id, { type: 'POST', cache: false, data: { delete_post: deletePost } }).catch(popupAjaxError);
+    const action = () => {
+      return ajax('/admin/flags/defer/' + this.id, {
+        type: 'POST', cache: false, data: { delete_post: deletePost }
+      });
+    };
+
+    if (deletePost && this._hasDeletableReplies()) {
+      return this._actOnFlagAndDeleteReplies(action);
+    } else {
+      return action().catch(popupAjaxError);
+    }
   },
 
   agreeFlags(actionOnPost) {
-    return ajax('/admin/flags/agree/' + this.id, { type: 'POST', cache: false, data: { action_on_post: actionOnPost } }).catch(popupAjaxError);
+    const action = () => {
+      return ajax('/admin/flags/agree/' + this.id, {
+        type: 'POST', cache: false, data: { action_on_post: actionOnPost }
+      });
+    };
+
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/admin/models/flagged-post.js.es6
```diff
@@ -62,11 +62,74 @@ export default Post.extend({
...
+      return action().catch(popupAjaxError);
+    }
+  },
+
+  _hasDeletableReplies() {
+    return this.get('post_number') > 1 && this.get('reply_count') > 0;
+  },
+
+  _actOnFlagAndDeleteReplies(action) {
+    return new Ember.RSVP.Promise((resolve, reject) => {
```
tgxworld: Not sure if I'm missing something but why do we need to return a promise here? We don't seem to be using the promise anywhere.

app/assets/javascripts/admin/models/flagged-post.js.es6
```diff
@@ -62,11 +62,74 @@ export default Post.extend({
...
+      return ajax(`/posts/${this.id}/reply-ids/all.json`).then(replies => {
```
tgxworld: I think we should avoid this extra request and do it all on the server side. We can use`reply_count` to replace `replies.length`.

app/assets/javascripts/admin/models/flagged-post.js.es6
```diff
@@ -62,11 +62,74 @@ export default Post.extend({
...
 
   agreeFlags(actionOnPost) {
-    return ajax('/admin/flags/agree/' + this.id, { type: 'POST', cache: false, data: { action_on_post: actionOnPost } }).catch(popupAjaxError);
+    const action = () => {
+      return ajax('/admin/flags/agree/' + this.id, {
+        type: 'POST', cache: false, data: { action_on_post: actionOnPost }
+      });
+    };
+
+    if (actionOnPost === 'delete' && this._hasDeletableReplies()) {
```
tgxworld: I think this should really just be a param that we pass to `/admin/flags/agree` to make it delete all the replies as well.

app/assets/javascripts/admin/models/flagged-post.js.es6
gschlager: It's waiting on the promise in the `removeAfter` calls: https://github.com/discourse/discourse/blob/ce97759418255af409693b95ca23f1274df8cd51/app/assets/javascripts/select-kit/components/admin-delete-flag-dropdown.js.es6#L68-L76

app/assets/javascripts/admin/models/flagged-post.js.es6
gschlager: `reply_count` is only the number of direct replies and `replies.length` is the number of all replies, so replacing won't work. I guess I could replace the second call by sending the reply ids to the server.

PR #5778, opened 2018-04-22
FEATURE: per-category approval settings

xrav3nz:
> CONTEXT & DISCUSSION: https://meta.discourse.org/t/all-topics-in-a-category-should-be-confirmed-before-publication/66800?u=xrav3nz

I _was_ going to add new [`CategoryGroup.permission_types`](https://github.com/discourse/discourse/blob/4997ede16c24d2bc5d0e723fd68a745fdb26b27d/app/models/category_group.rb#L7-L10), for easy extensibility in case we want to support requiring approval for specific user groups only. However:

1. We assume everyone has full access when there are no specific permissions set ([serializer](https://github.com/discourse/discourse/blob/4997ede16c24d2bc5d0e723fd68a745fdb26b27d/app/controllers/categories_controller.rb#L110-L112), [SQL](https://github.com/discourse/discourse/blob/4997ede16c24d2bc5d0e723fd68a745fdb26b27d/app/models/category.rb#L124)).
2. The logic for checking posting permissions are pretty uncoupled from checking if a post needs approval (`Guardian` vs `NewPostManager`) anyway.

I wasn't sure if this warrants two new columns, so I ended up creating [two custom fields on `Category`](https://github.com/xrav3nz/discourse/blob/087e2d3af244c63bc2087e23216e1d180a9c4d5a/app/models/category.rb#L10-L14).

Let me know what you think 💪

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2270,6 +2270,8 @@ en:
       allow_badges_label: "Allow badges to be awarded in this category"
       edit_permissions: "Edit Permissions"
       add_permission: "Add Permission"
+      require_topic_approval: "Require moderator approval of all new topics"
+      require_reply_approval: "Require moderator approval of all new replies"
       this_year: "this year"
       position: "position"
       default_position: "Default Position"
--- a/lib/new_post_manager.rb
+++ b/lib/new_post_manager.rb
@@ -82,7 +82,18 @@ class NewPostManager
     is_fast_typer?(manager) ||
     matches_auto_silence_regex?(manager) ||
     WordWatcher.new("#{manager.args[:title]} #{manager.args[:raw]}").requires_approval? ||
-    (SiteSetting.approve_unless_staged && user.staged)
+    (SiteSetting.approve_unless_staged && user.staged) ||
+    post_needs_approval_in_its_category?(manager)
+  end
+
+  def self.post_needs_approval_in_its_category?(manager)
+    if manager.args[:topic_id].present?
+      Category.joins(:topics).find_by(topics: { id: manager.args[:topic_id] }).require_reply_approval?
+    elsif manager.args[:category].present?
+      Category.find(manager.args[:category]).require_topic_approval?
+    else
+      false
+    end
   end
 
   def self.default_handler(manager)
--- a/lib/new_post_manager.rb
+++ b/lib/new_post_manager.rb
@@ -88,7 +88,9 @@ class NewPostManager
 
   def self.post_needs_approval_in_its_category?(manager)
     if manager.args[:topic_id].present?
-      Category.joins(:topics).find_by(topics: { id: manager.args[:topic_id] }).require_reply_approval?
+      cat = Category.joins(:topics).find_by(topics: { id: manager.args[:topic_id] })
+      return false unless cat
+      cat.require_reply_approval?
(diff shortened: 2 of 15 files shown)


config/locales/client.en.yml
```diff
@@ -2262,6 +2262,8 @@ en:
       allow_badges_label: "Allow badges to be awarded in this category"
       edit_permissions: "Edit Permissions"
       add_permission: "Add Permission"
+      require_topic_approval: "Require moderator approval of all new topics"
+      require_post_approval: "Require moderator approval of all new replies"
```
xrav3nz: Updated the copy ("posts" vs "replies") to be more precise! /cc @SamSaffron

lib/new_post_manager.rb
```diff
@@ -82,7 +82,18 @@ def self.post_needs_approval?(manager)
...
     matches_auto_silence_regex?(manager) ||
     WordWatcher.new("#{manager.args[:title]} #{manager.args[:raw]}").requires_approval? ||
-    (SiteSetting.approve_unless_staged && user.staged)
+    (SiteSetting.approve_unless_staged && user.staged) ||
+    post_needs_approval_in_its_category?(manager)
+  end
+
+  def self.post_needs_approval_in_its_category?(manager)
+    if manager.args[:topic_id].present?
+      Category.joins(:topics).find_by(topics: { id: manager.args[:topic_id] }).require_reply_approval?
```
SamSaffron: Oh ... what if the topic_id is invalid here or category_id on topic is null (like messages) `.require_reply_approval?` will explode...

lib/new_post_manager.rb
```diff
@@ -82,7 +82,18 @@ def self.post_needs_approval?(manager)
...
+    elsif manager.args[:category].present?
+      Category.find(manager.args[:category]).require_topic_approval?
```
SamSaffron: this is weird ... is category misnamed here and should be category_id ?

SamSaffron:
A couple more comment added :) 

let me know what you think

lib/new_post_manager.rb
xrav3nz: I was also super confused.

https://github.com/discourse/discourse/blob/22c4000375ce23d9ad1be8fb444ea008b156884f/app/assets/javascripts/discourse/models/composer.js.es6#L30-L35

https://github.com/discourse/discourse/blob/ee1eb1a5bd834ed49e7efe2ee58bb567556a8c7c/app/controllers/posts_controller.rb#L597-L602

I'll _try_ to rename this in this PR, unless it causes much havoc 😅 and I will open another PR.

xrav3nz:
I was on vacay last week and just moved back to school, will make the changes asap!

Giacomino1:

Simone Bartolini

SamSaffron:
@OsamaSayegh I still don't understand how these settings are passed to the UI when editing categories? Where is this information injected into the serializer?

OsamaSayegh:
@SamSaffron Here we're using category custom_fields to store these settings, and our category serializer ships them all to client.

https://github.com/discourse/discourse/blob/bd760e98af86f541080d49eab1633094e945dec9/app/serializers/category_serializer.rb#L18

SamSaffron:
Awesome ... OK let's try this thing out!

PR #5781, opened 2018-04-23
introduces wrap and changesDOM decorators

jjaffeux:
Wrap lets you prepend a function:

```
@wrap(doSomething)
callSomeone() {
}
```

changesDOM is a utility decorator relying on wrap which will check there's actually a DOM element and it's not getting destroyed:

```
@changesDOM
needsDOM() {
  this.$().html("xxx");
}
```

### DIFF
--- a/app/assets/javascripts/discourse-common/lib/changes-dom.js.es6
+++ b/app/assets/javascripts/discourse-common/lib/changes-dom.js.es6
@@ -1,0 +1,14 @@
+import wrap from "discourse-common/lib/wrap";
+
+export function DOMNotReadyError() { }
+DOMNotReadyError.prototype = new Error();
+
+export default function changesDOM(target, key, descriptor) {
+  const checkForDOM = function() {
+    if (!this.element || this.isDestroying || this.isDestroyed) {
+      throw new DOMNotReadyError();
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/discourse-common/lib/changes-dom.js.es6
```diff
@@ -0,0 +1,14 @@
...
+DOMNotReadyError.prototype = new Error();
+
+export default function changesDOM(target, key, descriptor) {
+  const checkForDOM = function() {
+    if (!this.element || this.isDestroying || this.isDestroyed) {
+      throw new DOMNotReadyError();
+    }
+  };
+
+  return wrap(checkForDOM, DOMNotReadyError).call(this, target, key, descriptor);
```
eviltrout: I spoke to @jjaffeux offline, but I we should replace the exception handling here.

Personally I think we could leave out `wrap` for now, but if you'd like to keep it we should return objects for the program flow rather than throw/catch exceptions.

SamSaffron:
I kind of like this, but we can not keep this in our queue now, got to get it into a proper shape and then we can look at it again.

PR #5782, opened 2018-04-23
UX: Tag each autocomplete dialog so they can be hidden

riking:
https://meta.discourse.org/t/default-emoji-plugin-how-to-disable/3415/50?u=riking

image


SamSaffron:
Nice! thanks!

PR #5785, opened 2018-04-25
REFACTOR: move support for user card badge images to a plugin discour…

nlalonde:
…se-user-card-badges

A column from the user_profiles table is being dropped in this commit. Someone please review?


app/models/user_profile.rb
```diff
@@ -126,7 +125,6 @@ def website_domain_validator
 #  bio_cooked_version   :integer
 #  badge_granted_title  :boolean          default(FALSE)
 #  card_background      :string(255)
-#  card_image_badge_id  :integer
```
gschlager: I think it would be a good idea to add `self.ignored_columns = %w(card_image_badge_id)` to this model.

nlalonde:
Don't merge.

app/models/user_profile.rb
nlalonde: Yes that's what I missed. Thanks!

PR #5793, opened 2018-04-27
WIP: fix catogory text_color for box style category badge

scossar:
`category_badge.rb` is used to create category badges for the the Discourse topic onebox, the not-found page, and the email digest. It isn't setting the text color correctly for the box_style category badge.

screen shot 2018-04-27 at 11 09 19 am

This PR fixes that problem both for when the `inline_style` option is used and for when it isn't used.

There is a further issue that when there is an apostrophe in the category description the markup that is returned for the `.badge-category` span is broken due to the way the `title` attrubute is being set:

```
title="This is a category for learning about Discourse development. I" ll="" be="" using="" it="" wp-discourse="" sub-category="" for="" testing="" the="" wordpress="" discourse="" plugin.=""
```
I can look at this some more and either include a fix for it in this PR, or deal with it separately.

### DIFF
--- a/lib/category_badge.rb
+++ b/lib/category_badge.rb
@@ -88,12 +88,14 @@ module CategoryBadge
         when :bar
           'color: #222222; padding: 3px; vertical-align: text-top; margin-top: -3px; display: inline-block;'
         when :box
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am merging this cause it feels better than what we have now. @scossar do we need another PR here as well?

PR #5794, opened 2018-05-01
FEATURE: adds support for emoji search aliases

jjaffeux:


coding-horror:
these aliases are only retrieved on demand, yes? not pre-loaded into the app?

PR #5797, opened 2018-05-01
Feature: Include participants at the bottom of PM emails

featheredtoast:
https://meta.discourse.org/t/email-notification-recipients-unclear-when-pm-is-sent-to-multiple-users/26934/13?u=featheredtoast

Includes a comma separated list in email footers: "This is a PM between..." under the response footer

Lists users first, followed by groups.

Fix: added missing translation for PM mentions

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -447,15 +447,15 @@ class UserNotifications < ActionMailer::Base
 
       participants = "#{I18n.t("user_notifications.pm_participants")} "
       participant_list = []
+      post.topic.allowed_groups.each do |group|
+        participant_list.push "[#{group.name} (#{I18n.t("groups.member_count", count: group.users.count)})](#{Discourse.base_url}/groups/#{group.name})"
+      end
       post.topic.allowed_users.each do |user|
         if SiteSetting.prioritize_username_in_ux?
           participant_list.push "[#{user.username}](#{Discourse.base_url}/u/#{user.username_lower})"
         else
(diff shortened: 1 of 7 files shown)


coding-horror:
can you review this @SamSaffron  we need this to be expedited for 2.0

app/mailers/user_notifications.rb
```diff
@@ -444,6 +444,16 @@ def send_notification_email(opts)
         else
           I18n.t('subject_pm')
         end
+
+      participants = "#{I18n.t("user_notifications.pm_participants")} "
+      participant_list = []
+      post.topic.allowed_users.each do |user|
+        participant_list.push user.username
```
SamSaffron: I am not sure this is correct, if `prioritize username in ux` should we not be showing "name" here, also should we make this hyperlinks to the actual user?

SamSaffron:
Can you add a screenshot here @featheredtoast  ?

featheredtoast:
Completely forgot that 'prioritize username in ux' was optional, I'll switch it to depend on that, and add the links.

Meanwhile, here's a screenshot of what I've got so far:

screen shot 2018-05-02 at 9 07 24 am

featheredtoast:
Full name support added. Added username and group links.

screen shot 2018-05-02 at 10 48 39 am

SamSaffron:
we got to de-emphasize those user links, I want it clickable but the unsubscribe is far more important and this is getting in the way. Just make them links but style them like standard text maybe? 

"The is PM between" ... text should probably be a simpler "Participants: bob, bill, jane"

SamSaffron:
This is fine, I just think you need to remove (2 members) because it is a bit overly verbose. People will most likely know what the group is anyway. 

Once that is done feel free to merge!

PR #5802, opened 2018-05-04
Do not ignore core plugin discourse-local-dates

jjaffeux:


SamSaffron:
oh yeah doing this would have saved me much pain

PR #5803, opened 2018-05-05
Allow plugins to modify class methods/properties

shoshber:
`plugin-api.modifyClass` allows plugin authors to overwrite instance methods of existing classes, but there is no analogous way to overwrite class methods.... until now.

context here: https://meta.discourse.org/t/allow-modifying-class-methods-properties-in-plugin-api/84270

### DIFF
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -45,11 +45,19 @@ class PluginApi {
   }
 
   /**
-   * Allows you to overwrite or extend methods in a class.
+   * Allows you to overwrite or extend methods and properties in a class.
+   *
+   * To modify class methods and properties, add `classMethods` to `opts` with a truthy value.
+   *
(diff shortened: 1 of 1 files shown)


SamSaffron:
something about this api is not sitting right for me... I feel like we should deprecate `modifyClass` which is somewhat ambiguous and create

```
api.reopen
api.reopenClass
```

@eviltrout  is on a break at the moment so we can discuss with him once he is back

eviltrout:
I purposely didn't call it reopen because ES6 classes don't call them class methods, they call them static methods, and eventually the same API will be used for non-ember classes.

I would prefer if we added a new method, `api.modifyClassStatic(...)` rather than a parameter like in this PR.

That way the terms align with ES6 classes and we're better off in the future.

SamSaffron:
modifyClassStatic works better for me 👍🏼

SamSaffron:
I added modifyClassStatic, so we should be a-ok here! thanks heaps for the suggestion @shoshber

PR #5804, opened 2018-05-06
Improved email validation in user search module.

nbianca:
Fixes the bug described [here](https://meta.discourse.org/t/username-preceded-by-detected-as-email/71355/5).

Basically, the bug was caused by poorly validation of email addresses. Everything containing `@` was considered an email, which is not true.


ZogStriP:
Thanks 👍

PR #5809, opened 2018-05-08
FIX: better handling of invite links after they are redeemed

techAPJ:
FIX: deprecate invite_passthrough_hours setting

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -408,7 +408,6 @@ users:
     default: true
   invite_expiry_days:
     default: 30
-  invite_passthrough_hours: 0
   invites_per_page:
     client: true
     default: 40
--- a/db/migrate/20180508142711_remove_invite_passthrough_hours.rb
+++ b/db/migrate/20180508142711_remove_invite_passthrough_hours.rb
@@ -1,0 +1,5 @@
+class RemoveInvitePassthroughHours < ActiveRecord::Migration[5.1]
+  def change
+    execute "DELETE FROM site_settings WHERE name = 'invite_passthrough_hours'"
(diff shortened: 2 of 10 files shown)


config/site_settings.yml
```diff
@@ -408,7 +408,6 @@ users:
     default: true
   invite_expiry_days:
     default: 30
-  invite_passthrough_hours: 0
```
tgxworld: We need to write a migration to drop site settings in the db here

tgxworld:
@techAPJ I think we should update the commit message to state that we're removing that site setting.

techAPJ:
@tgxworld I updated the commit message already... see https://github.com/discourse/discourse/pull/5809/commits/807200dc054d7a46edc6fe9395f0ff14497fceb1

Just the PR description was not updated, will do so now.

config/site_settings.yml
techAPJ: Done.

db/migrate/20180508142711_remove_invite_passthrough_hours.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveInvitePassthroughHours < ActiveRecord::Migration[5.1]
+  def change
```
tgxworld: @techAPJ This should be `up` only as you can't rollback the SQL statement.

techAPJ: Good catch. Fixed in: https://github.com/discourse/discourse/commit/27efa4580038617c1a043aca17df8d2ca09ba746.

PR #5814, opened 2018-05-11
Fix issue for custom image badges urls

duranmla:
# Description

Whenever the user tries to add a custom image URL in order to change the image of a badge that value won't work as the method only takes into count the `icon` property, this code refactor the method for `icon-or-image` in order to accept the whole object and verify both properties (image and icon) and retrieving the image if it is set or the icon if the image is not set and the icon is a valid font-awesome icon type format.


ZogStriP:
Thanks 👍

ZogStriP:
Some tests here would be awesome ;)

duranmla:
Sorry for the late response @ZogStriP I just realize I had some issues with my notifications preferences at Github. I will do some tests for further cases whenever it feels necessary.

PR #5815, opened 2018-05-11
Allow parameters for group and username filters on directory

joebuhlig:


ZogStriP:
Great 👍 Any chances you could add some tests?

joebuhlig:
Sure. I simply didn't find any tests on the DirectoryItemsController or the UsersRoute to append. Do you need me to create the initial files for these?

ZogStriP:
The tests for the `DirectoryItemsController` are in
https://github.com/discourse/discourse/blob/master/spec/requests/directory_items_controller_spec.rb

The tests for the `users` controller are inn https://github.com/discourse/discourse/blob/master/test/javascripts/acceptance/users-test.js.es6

which should also be testing the routes ;)

joebuhlig:
🤦‍♂️ Things I know but failed to remember.

joebuhlig:
Ok. I think I have those in place. Have a look and let me know if I missed anything.

spec/requests/directory_items_controller_spec.rb
```diff
@@ -28,9 +28,10 @@
 
   context "with data" do
     before do
-      Fabricate(:evil_trout)
-      Fabricate(:walter_white)
-      Fabricate(:staged, username: 'stage_user')
+      @evil_trout = Fabricate(:evil_trout)
```
ZogStriP: These can be replaced with `let!(:evil_trout) { Fabricate(:evil_trout) }`

ZogStriP:
Looks good 👌 Just one nitpick and I'll merge it ;)

spec/requests/directory_items_controller_spec.rb
joebuhlig: That doesn't seem to work in this case:

> `let!` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).

joebuhlig:
Just made an edit that might fit the bill. Using the `let!` at that point doesn't work: 

> `let!` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).

PR #5816, opened 2018-05-12
Add emails:test rake task

pfaffman:
Handy to be able to do an email test without creating an account and logging in!


ZogStriP:
Thanks 👍

PR #5818, opened 2018-05-14
FIX: changing tag group privacy setting was not consistent

techAPJ:
https://meta.discourse.org/t/tags-can-be-used-by-everyone-causes-tag-to-be-usable-only-by-staff/86953

### DIFF
--- a/app/assets/javascripts/discourse/components/radio-button.js.es6
+++ b/app/assets/javascripts/discourse/components/radio-button.js.es6
@@ -5,8 +5,12 @@ export default Ember.Component.extend({
   type : "radio",
   attributeBindings : ["name", "type", "value", "checked:checked", "disabled:disabled"],
 
-  click: function() {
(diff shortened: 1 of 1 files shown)


techAPJ:
Build failure is not related.

coding-horror:
Why can't this be directly merged?

techAPJ:
@coding-horror This is an odd fix which I am not so confident about. It does fixes the underlying issue.

I would appreciate if someone from the team review this PR. Maybe @ZogStriP?

ZogStriP:
This is indeed a weird fix... Having a look 🔍

ZogStriP:
This does fix the issue but a proper fix would be to refactor all our radio buttons and use something similar to the [wizard radio buttons](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/wizard/components/radio-button.js.es6).

PR #5825, opened 2018-05-14
FIX: update cache times for service workers

featheredtoast:
Add a last modified time.

Register newer service workers and claim clients more quickly.

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -160,8 +160,11 @@ class StaticController < ApplicationController
       format.js do
         # https://github.com/w3c/ServiceWorker/blob/master/explainer.md#updating-a-service-worker
         # Maximum cache that the service worker will respect is 24 hours.
-        immutable_for 24.hours
+        # However, ensure that these may be cached and served for longer on servers.
+        immutable_for 1.year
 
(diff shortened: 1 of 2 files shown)


app/controllers/static_controller.rb
```diff
@@ -160,8 +160,11 @@ def service_worker_asset
       format.js do
         # https://github.com/w3c/ServiceWorker/blob/master/explainer.md#updating-a-service-worker
         # Maximum cache that the service worker will respect is 24 hours.
-        immutable_for 24.hours
+        # However, ensure that these may be cached and served for longer on servers.
+        immutable_for 1.year
```
tgxworld: I don't quite agree with this here. Wouldn't we bloat clients/nginx in this case since they end up caching all the service workers file for a year?

featheredtoast: wait, other static files follow this pattern too (cache up to 1 year) ...what's different about service workers?

tgxworld: Hmm I'm actually not sure about why we cache stuff for a year. Either way let's try this out for 2.0 but I'll fix my mistake of changing service-worker file names in 2.1.

PR #5826, opened 2018-05-14
IncomingLink: do not store IP of logged-in users

riking:
Code is pretty simple - just don't store the IP if we have a current user.

Console output on development instance showing this working:

```
 #<IncomingLink:0x000000000b16d1d0
  id: 3,
  created_at: Mon, 14 May 2018 21:46:56 UTC +00:00,
  user_id: 1,
  ip_address: #<IPAddr: IPv4:127.0.0.1/255.255.255.255>,
  current_user_id: nil,  ## <==
  post_id: 18,
  incoming_referer_id: nil>,
 #<IncomingLink:0x000000000b16ceb0
  id: 4,
  created_at: Mon, 14 May 2018 21:55:45 UTC +00:00,
  user_id: 1,
  ip_address: nil,       ## <== ip_address is nil for logged-in users
  current_user_id: 2,    ## <==
  post_id: 19,
  incoming_referer_id: nil>]
```

### DIFF
--- a/app/models/incoming_link.rb
+++ b/app/models/incoming_link.rb
@@ -20,6 +20,7 @@ class IncomingLink < ActiveRecord::Base
       u = User.select(:id).find_by(username_lower: username.downcase)
       user_id = u.id if u
     end
+    ip_address = opts[:ip_address]
 
     if opts[:referer].present?
       begin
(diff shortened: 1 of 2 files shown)


SamSaffron:
thanks this looks good, a longer term change after easy cleanups is to "fold" up the table weekly into a summary table that counts "anon/logged in" per post and then only retain high fidelity for a week. It is a bit tricky cause looking at incoming links would require looking at 2 tables, but if we keep the summary "live" then we can just move to using the summary all the time, then the incoming link table will be used to de-duplicate for up to N days.

PR #5828, opened 2018-05-15
Fix report specs on Ruby 2.3

gschlager:
Our Travis build is broken on Ruby 2.3 because `ceil` doesn't support `precision` as argument. This adds a patch for old Ruby versions.

@SamSaffron Does that look OK to you?


SamSaffron:
no longer supporting ruby 2.3 ... so closing this

PR #5829, opened 2018-05-15
Change condition when displaying URLs for S3 a/v uploads in the editor

majakomel:
This is a minor replacement that doesn't change any behaviour. 

Instead of checking whether S3 uploads are enabled directly with the site settings, it checks it via the helper which I also used for overriding the settings for Azure Blob Storage plugin (it fixes the bug that appeared when uploading audio/video files onto Azure Blob https://meta.discourse.org/t/azure-blob-storage-plugin/79392/13).

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js.es6
+++ b/app/assets/javascripts/discourse/lib/utilities.js.es6
@@ -348,7 +348,7 @@ export function uploadLocation(url) {
   if (Discourse.CDN) {
     url = Discourse.getURLWithCDN(url);
     return /^\/\//.test(url) ? 'http:' + url : url;
-  } else if (Discourse.SiteSettings.enable_s3_uploads) {
+  } else if (Discourse.S3BaseUrl) {
     return 'https:' + url;
   } else {
     var protocol = window.location.protocol + '//',


app/models/site_setting.rb
```diff
@@ -133,6 +133,7 @@ def self.enable_s3_uploads
     end
 
     def self.absolute_base_url
+      return nil if !SiteSetting.Upload.enable_s3_uploads
```
tgxworld: Can we add a test case here because it isn't clear to me what we're trying to fix here.

majakomel: Actually I figured out it's not really needed so I'll just remove it.

SamSaffron:
looks like we have a conflict :( can you try rebasing?

majakomel:
This is rebased now

SamSaffron:
I follow, but I worry that we are relying a bit too much on s3 terminology if we are using this for azure. we should rename these vars to something a bit more flexible and introduce a cleaner way of patching this in.

majakomel:
@SamSaffron should I rename `Discourse.S3CDN` to sth like `Discourse.UploadCDN` and `Discourse.S3BaseUrl` to `Discourse.UploadBaseUrl` then?

SamSaffron:
> should I rename Discourse.S3CDN to sth like Discourse.UploadCDN and Discourse.S3BaseUrl to Discourse.UploadBaseUrl then? 

I say yes! going to close this PR can you open a new one with that change, get @jjaffeux  to quickly review and go ahead and merge 👍

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

PR #5832, opened 2018-05-15
Fix: Safari failing to download backups from email links

featheredtoast:
disable service workers on safari

bypass serviceworker network if this is a URL with a token param

### DIFF
--- a/app/assets/javascripts/discourse/initializers/register-service-worker.js.es6
+++ b/app/assets/javascripts/discourse/initializers/register-service-worker.js.es6
@@ -6,9 +6,10 @@ export default {
           (location.hostname === "localhost");
 
     const isSupported= isSecured && ('serviceWorker' in navigator);
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
 
     if (isSupported) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/initializers/register-service-worker.js.es6
```diff
@@ -6,9 +6,10 @@ export default {
           (location.hostname === "localhost");
 
     const isSupported= isSecured && ('serviceWorker' in navigator);
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
ZogStriP: We might already have that User Agent sniffing code somewhere in a lib. Might be worth checking out.

featheredtoast: I found one for sniffing out Apple, but no dice checking for Safari specifically.

For reference, "Safari" also appears in macOS' Chrome, which is why the regex checks for the strings "chrome" or "android" not appearing as well.

tgxworld:
@featheredtoast Looks good to me :+1:

PR #5835, opened 2018-05-17
Destroy Pikaday properly when date-component is destroyed

angusmcleod:


SamSaffron:
Thanks, just pushed a fix

PR #5836, opened 2018-05-17
Fix super-wide tables in Firefox

barryvan:
In Firefox (tested in Fx61), the "activity metrics" and related tables can become stupendously wide. Adding a `max-width` resolves this issue.


SamSaffron:
wow ... so odd... happy with this patch

PR #5837, opened 2018-05-17
Add plugin outlet in admin theme list

davidtaylorhq:
This will be used by discourse-theme-creator to add avatars next to each theme.

Also adds `belongs_to :user` to the theme model, and adds the theme owner to the serializer


ZogStriP:
Thanks 👍

PR #5838, opened 2018-05-17
FIX for bucket name in GlobalSetting with folder

andrewschleifer:
When `s3_bucket="bucket/folder` in discourse.conf, absolute_base_url
was bucket/folder.s3-region.amazonaws.com

These names are bad, but this mirrors the s3_bucket/s3_bucket_name in
S3Store

N.B. that nearby s3_upload_bucket _should_ include the folder

### DIFF
--- a/app/models/global_setting.rb
+++ b/app/models/global_setting.rb
@@ -85,6 +85,10 @@ class GlobalSetting
       end) == :true
   end
 
+  def self.s3_bucket_name
+    @s3_bucket_name ||= s3_bucket.downcase.split("/")[0]
+  end
+
   # for testing
   def self.reset_s3_cache!
     @use_s3 = nil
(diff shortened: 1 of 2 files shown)


app/models/global_setting.rb
```diff
@@ -85,6 +85,10 @@ def self.use_s3?
       end) == :true
   end
 
+  def self.s3_bucket_name
+    @s3_bucket_name ||= s3_bucket.downcase.split("/")[0]
```
tgxworld: If the `s3_bucket` contains more than just the bucket name, do we consider that a misconfiguration?

andrewschleifer: Commit aa5de3c40acebe3f3577a977cf57ed7630a4b740 added support for subfolders in the bucket name as "bucket/folder". And that works fine if you put that in the admin section.

But if you pass it in via environment variable (which is how it gets stuffed into discourse.conf) then the absolute_base_url is wrong. This PR fixes that.

Having a separate config var for the subfolder is probably a good idea? But that's a bigger change.

PR #5843, opened 2018-05-18
ignore emails that are from the reply by email addresses

ryantm:
discussion here: https://meta.discourse.org/t/spam-with-from-address-set-to-reply-by-email-address/87786/2

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -21,6 +21,7 @@ module Email
     class BouncedEmailError            < ProcessingError; end
     class NoBodyDetectedError          < ProcessingError; end
     class NoSenderDetectedError        < ProcessingError; end
+    class FromReplyByAddressError      < ProcessingError; end
     class InactiveUserError            < ProcessingError; end
     class SilencedUserError            < ProcessingError; end
     class BadDestinationAddress        < ProcessingError; end
@@ -107,6 +108,7 @@ module Email
     def process_internal
       raise BouncedEmailError  if is_bounce?
       raise NoSenderDetectedError if @from_email.blank?
+      raise FromReplyByAddressError if is_from_reply_by_email_address?
       raise ScreenedEmailError if ScreenedEmail.should_block?(@from_email)
 
       user = find_user(@from_email)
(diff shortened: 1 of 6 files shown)


lib/email/receiver.rb
```diff
@@ -107,6 +108,7 @@ def create_incoming_email
     def process_internal
       raise BouncedEmailError  if is_bounce?
       raise NoSenderDetectedError if @from_email.blank?
+      raise FromReplyByAddressError if is_from_reply_by_email_address?
```
ZogStriP: What happens when that exception is raised?

lib/email/receiver.rb
```diff
@@ -203,6 +205,20 @@ def is_bounce?
       true
     end
 
+    def is_from_reply_by_email_address?
+      reply_addresses = [SiteSetting.reply_by_email_address]
```
ZogStriP: Why not use the `reply_by_email_address_regex` function?

lib/email/receiver.rb
ryantm: I think it gets caught by https://github.com/discourse/discourse/blob/master/lib/email/processor.rb#L59 then handled by https://github.com/discourse/discourse/blob/master/lib/email/processor.rb#L71

lib/email/receiver.rb
ryantm: reply_by_email_address_regex is too specific. I want to handle the case where the spammer has removed the + and the reply_key

lib/email/receiver.rb
ryantm: Oh, I looked at that code more and it looks like it is probably still going to send a rejection mail in that case. I'll work on making it better.

lib/email/receiver.rb
ZogStriP: Any chances you could merge both methods and add an argument?

ryantm:
Okay, I think I made it actually work now! And I think I addressed @ZogStriP's concerns about copying the regexp function: I changed the regexp to not match everything if it was empty, and I match, even if the "+code" part is missing.

ryantm:
ping @ZogStriP

ZogStriP:
Thanks 👍

PR #5844, opened 2018-05-18
FIX: don't allow duplicate watched words

OsamaSayegh:
We already have logic in place for server side, this'll just display a little message that says the word already exists

Meta https://meta.discourse.org/t/adding-words-to-watched-words-doesnt-check-for-duplicates/84393?u=osama


ZogStriP:
Thanks 👍

PR #5846, opened 2018-05-18
allow adding tags as a custom subject format for emails

majakomel:
Meta: https://meta.discourse.org/t/allow-adding-tags-as-a-custom-subject-format-for-emails/84424

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -446,7 +446,7 @@ class UserNotifications < ActionMailer::Base
     # tag names
     if opts[:show_tags_in_subject] && post.topic_id
       tags = Topic.find_by(id: post.topic_id)&.tags.first(3)
-      show_tags_in_subject = tags.any? ? tags.map{ |t| "[#{t.name}]" }.join : nil
+      show_tags_in_subject = tags.any? ? tags.map { |t| "[#{t.name}]" }.join : nil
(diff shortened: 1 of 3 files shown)


app/mailers/user_notifications.rb
```diff
@@ -433,6 +443,14 @@ def send_notification_email(opts)
       show_category_in_subject = nil
     end
 
+    # tag names
+    tags = Topic.find_by(id: post.topic_id)&.tags
+    if opts[:show_tags_in_subject] && post.topic_id && tags
```
ZogStriP: I would first check tags are allowed in subject and only then do the SQL query.

ZogStriP: Also, there should probably be a limit of the number of tags retrieved.

majakomel: What would be the best limit? 3 tags? Should there be any specific order as well then?

ZogStriP: 3 seems fine. I would keep the same order as in the post.

SamSaffron:
This looks good to me, should we be adding spaces between tags or having them this close together? 

Note we need to wait 2 weeks to merge this in cause we are stabalizing the 2.0 release at the moment.

SamSaffron:
cool, merging now

PR #5848, opened 2018-05-21
FIX: apply automatic group rules when using social login providers

OsamaSayegh:
Meta topic: https://meta.discourse.org/t/user-set-automatic-groups-is-not-called-for-all-social-logins/81579?u=osama

### DIFF
--- a/app/services/user_authenticator.rb
+++ b/app/services/user_authenticator.rb
@@ -21,7 +21,10 @@ class UserAuthenticator
   end
 
   def finish
-    authenticator.after_create_account(@user, @session) if authenticator
+    if authenticator
+      authenticator.after_create_account(@user, @session)
+      confirm_email
+    end
     @session = nil
   end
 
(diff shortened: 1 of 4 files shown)


app/services/user_authenticator.rb
```diff
@@ -20,8 +20,18 @@ def has_authenticator?
     !!authenticator
   end
 
+  def confirm_email
+    if @session[:email_valid].to_s == 'true' && @session[:email]&.downcase == @user.email
```
tgxworld: There is an `#authenticated?` method that we can use here.

app/services/user_authenticator.rb
```diff
@@ -20,8 +20,18 @@ def has_authenticator?
     !!authenticator
   end
 
+  def confirm_email
```
tgxworld: I would prefer to keep this method private and test `UserAuthenticator#finish` instead.

OsamaSayegh:
@tgxworld thanks for the feedback. I've made the changes you requested. let me know what you think :)

tgxworld:
Looks good to me :+1: Nice work @OsamaSayegh

PR #5850, opened 2018-05-21
TopicViewItem: Do not store IP of logged in users

riking:
Second part of IP logging removals


riking:
Only remaining test failure is a flake.

SamSaffron:
looks good, but I think you are missing a test here, also can you send a separate PR for the flaky spec, so @jjaffeux  can merge in and fix.

SamSaffron:
I think this is reasonably safe but waiting till post 2.1 on it cause index work can be expensive and potentially block stuff

SamSaffron:
I am fine with this, lets try it out

riking:
Oops, I let that sit for a while didn't I

PR #5851, opened 2018-05-21
SearchLog: Do not store IP of logged in users, add retention policy

riking:
Also set a time-based retention policy on search logs, in addition to the size-based retention policy.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1067,6 +1067,7 @@ en:
     search_recent_posts_size: "How many recent posts to keep in the index"
     log_search_queries: "Log search queries performed by users"
     search_query_log_max_size: "Maximum amount of search queries to keep"
+    search_query_log_max_retention_days: "Maximum amount of time to keep search queries, in days. Default is 3 years (1095 days)."
     allow_uncategorized_topics: "Allow topics to be created without a category. WARNING: If there are any uncategorized topics, you must recategorize them before turning this off."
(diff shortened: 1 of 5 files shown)


config/locales/server.en.yml
```diff
@@ -1067,6 +1067,7 @@ en:
     search_recent_posts_size: "How many recent posts to keep in the index"
     log_search_queries: "Log search queries performed by users"
     search_query_log_max_size: "Maximum amount of search queries to keep"
+    search_query_log_max_retention_days: "Maximum amount of time to keep search queries, in days. Default is 3 years (1095 days)."
```
SamSaffron: default should be 6 months, I do not see why we need to hoard 3 years here.

SamSaffron:
retention should be tested with freeze_time

config/locales/server.en.yml
riking: The admin dashboard is trying to do year-over-year analytics on search queries; should search logs be limited to month/quarter scope?

SamSaffron:
sure we can limit that, yearly trends of search are not really that
interesting imo

config/locales/server.en.yml
SamSaffron: largest interval there is a year, so just keep 1 year max I guess.

SamSaffron:
Pushing this out 2 weeks so it is post 2.0 release, can you finish by then? not super urgent

SamSaffron:
merging but will fiddle with the dates a bit

PR #5852, opened 2018-05-21
TopicLinkClick: do not log IP of logged in users

riking:
Just like the other PRs.


riking:
Test failure is a flake.

SamSaffron:
we should add a test here

SamSaffron:
looks good to me! thanks

PR #5853, opened 2018-05-21
UserProfileView: Do not log IP of logged-in users

riking:
Another re-index, and added a test for this one.

### DIFF
--- a/spec/models/user_profile_view_spec.rb
+++ b/spec/models/user_profile_view_spec.rb
@@ -34,6 +34,8 @@ RSpec.describe UserProfileView do
     ['1.1.1.1', '2.2.2.2'].each do |ip|
       add(user_profile_id, ip, other_user.id, time)
       expect(described_class.count).to eq(1)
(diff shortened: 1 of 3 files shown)


spec/models/user_profile_view_spec.rb
```diff
@@ -34,6 +34,8 @@ def add(user_profile_id, ip, user_id = nil, at = nil)
     ['1.1.1.1', '2.2.2.2'].each do |ip|
       add(user_profile_id, ip, other_user.id, time)
       expect(described_class.count).to eq(1)
+      # should not actually log IPs
+      expect(UserProfileView.where(user_id: other_user.id).count(:ip_address)).to eq(0)
```
SamSaffron: this is a bit ... strange, maybe this? 

```
expect(UserProfileView.find_by(user_id: other_user.id).ip_address.to eq(nil)
```

SamSaffron:
Seems pretty safe ... merging ... thanks!

PR #5854, opened 2018-05-22
TEST: Disable flaky composer acceptance tests 

riking:
Split from #5850 

@jjaffeux


SamSaffron:
cool, I just skipped this both today @jjaffeux  will be looking into it.

PR #5866, opened 2018-05-23
Add required ports to cloud install

dnsmichi:
Users sometimes do have a webserver already running on the same host.

This addition marks this as a requirement for the setup wizard.

Reference: https://meta.discourse.org/t/running-other-websites-on-the-same-machine-as-discourse/17247/244


coding-horror:
Thanks, but the setup script checks for this and warns if the ports are occupied, and that's the better place to do that, on demand.

dnsmichi:
Ok - I'm just a friend of reading the requirements before executing any script :-)

PR #5876, opened 2018-05-27
UX: The "enable 2FA" string was hard to translate

gschlager:
The "Enable two factor authentication for enhanced account security" was split into 3 parts which made it really hard to translate.

I wanted to do something like `{{{i18n 'user.second_factor.enable' link=(action "showTwoFactorModal")}}}`, but that didn't work. So, this is the next best solution I could think of.

### DIFF
--- a/app/assets/javascripts/discourse/templates/preferences/account.hbs
+++ b/app/assets/javascripts/discourse/templates/preferences/account.hbs
@@ -71,9 +71,7 @@
     {{#if model.second_factor_enabled}}
       {{i18n 'user.second_factor.disable'}}
     {{else}}
-      {{i18n 'enable'}}
-      <a href {{action "showTwoFactorModal"}}>{{showTwoFactorModalText}}</a>
-      {{i18n 'user.second_factor.enable'}}
+      {{discourse-linked-text action="showTwoFactorModal" text="user.second_factor.enable"}}
     {{/if}}
 
     {{#if isCurrentUser}}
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/templates/preferences/account.hbs
```diff
@@ -71,9 +71,7 @@
     {{#if model.second_factor_enabled}}
       {{i18n 'user.second_factor.disable'}}
     {{else}}
-      {{i18n 'enable'}}
-      <a href {{action "showTwoFactorModal"}}>{{showTwoFactorModalText}}</a>
-      {{i18n 'user.second_factor.enable'}}
+      <span onclick={{action 'showTwoFactorModal'}}>{{{i18n 'user.second_factor.enable'}}}</span>
```
SamSaffron: something does not feel right to me using `span onclick` why doing we need hyperlink inception here?

gschlager: Because I wanted to replace 3 concatenated strings (`enable`, `user.second_factor.title`, `user.second_factor.enable`) with just one. Otherwise it's really hard to translate. The 2nd one is currently even converted into lower case.

Unfortunately the string should contain a link that calls an action and that's probably why @tgxworld used 3 strings instead of just 1.

image

I know my solution is kind of a hack. If there is a better solution I'm all ears! Unfortunately `{{{i18n 'user.second_factor.enable' link=(action "showTwoFactorModal")}}}` didn't work.

SamSaffron:
just updating that we discussed this internally and will be introducing a less hacky pattern here.

gschlager:
I changed it, so that it's less hacky. 😉 Thanks for pointing me in the right direction. 👍 
It could easily be enhanced to support multiple links, but for now 1 link/action was enough.

PR #5877, opened 2018-05-28
Add plugin outlet create-account-after-password

shoshber:
There's a plugin outlet called `create-account-before-password`, so I added `create-account-after-password` (which I needed for a plugin I'm developing). I also added `noTags=true` to both outlets. Since both outlets are inside a `table` element, wrapping them in the default `span` isn't valid html.

(Also https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/plugin-outlet.js.es6#L39)

### DIFF
--- a/app/assets/javascripts/discourse/templates/modal/create-account.hbs
+++ b/app/assets/javascripts/discourse/templates/modal/create-account.hbs
@@ -53,6 +53,7 @@
           {{/if}}
 
           {{plugin-outlet name="create-account-before-password"
+                          noTags=true
                           args=(hash accountName=accountName
                                      accountUsername=accountUsername
                                      accountPassword=accountPassword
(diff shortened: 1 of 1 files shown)


SamSaffron:
Nice I really like that you added noTags

We are stabalizing core atm for 2.0 release (very very soon) so holding off on taking changes ... this looks very safe and technically is a bug fix, but I will leave it to @eviltrout to decide if we merge it in today vs next week.

eviltrout:
This is technically a breaking change; people's styles might have depended on the previously existing tags.

I do not recommend we merge this until after the next release unless we drop the `noTags`.

SamSaffron:
I would argue it’s a fixing change cause a span as a direct descendant of a
table is not going to result in anything consistent since it is invalid

That said, I agree we should wait a week here till we are post release

SamSaffron:
Thanks for the PR @shoshber

PR #5879, opened 2018-05-28
DEV: introduces prettier into the codebase for scss files

jjaffeux:
yarn global add prettier
prettier --write 'app/assets/stylesheets/**/*.scss' '!app/assets/stylesheets/vendor/**'


tgxworld:
We should add `prettier` as a development dependency in [package.json](https://github.com/discourse/discourse/blob/4ad924bcf5acb6c99868887b2a402b7b5f289771/package.json)

tgxworld:
@jjaffeux :+1: LGTM

SamSaffron:
looks like this is done, but I really want some sort of enforcement here as well

PR #5886, opened 2018-05-29
REFACTOR: topics controller specs to requests

OsamaSayegh:


spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+        post_ids: [1, 2, 3]
+      }
+      expect(response).to be_forbidden
+    end
 
-          context 'when there are no changes' do
-            it 'does not call the PostRevisor' do
-              PostRevisor.any_instance.expects(:revise!).never
+    describe 'forbidden to moderators' do
+      let!(:moderator) { sign_in(Fabricate(:moderator)) }
```
tgxworld: This should probably be placed in a `before` block

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+        }
+        expect(response).to be_forbidden
+      end
+    end
 
-              put "/t/#{topic.slug}/#{topic.id}.json", params: {
-                category_id: topic.category_id
-              }
+    describe 'forbidden to trust_level_4s' do
+      let!(:trust_level_4) { sign_in(Fabricate(:trust_level_4)) }
```
tgxworld: `before` block is more accurate here since we don't actually need the variable.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+    describe 'changing ownership' do
+      let!(:editor) { sign_in(Fabricate(:admin)) }
+      let(:topic) { Fabricate(:topic) }
+      let(:user_a) { Fabricate(:user) }
+      let(:p1) { Fabricate(:post, topic_id: topic.id) }
+      let(:p2) { Fabricate(:post, topic_id: topic.id) }
+
+      it "raises an error with a parameter missing" do
+        post "/t/111/change-owner.json", params: { post_ids: [1, 2, 3] }
```
tgxworld: let us use a loop here to DRY up the code. 
```
[
  { post_ids: [1, 2, 3] },
  { username: 'user_a' }
].each do |params|
  ...
end
```

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+        expect(response.status).to eq(400)
+
+        post "/t/111/change-owner.json", params: { username: 'user_a' }
+        expect(response.status).to eq(400)
+      end
+
+      it "calls PostOwnerChanger" do
+        PostOwnerChanger.any_instance.expects(:change_owner!).returns(true)
```
tgxworld: We should test for the actual behavior here instead of stubbing

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+        expect(p1.user).to_not eq(nil)
+        expect(p1.reload.user).to eq(p2.reload.user)
+      end
+
+      it "works with deleted users" do
+        deleted_user = Fabricate(:user)
+        t2 = Fabricate(:topic, user: deleted_user)
+        p3 = Fabricate(:post, topic_id: t2.id, user: deleted_user)
+        deleted_user.save
```
tgxworld: All the `save`s here are unnecessary so let's remove them

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+      end
+
+      it "fails when the user can't see the topic" do
+        put "/t/#{pm.id}/clear-pin.json"
+        expect(response).to be_forbidden
+      end
+
+      describe 'when the user can see the topic' do
+        it "calls clear_pin_for if the user can see the topic" do
+          Topic.any_instance.expects(:clear_pin_for).with(user).once
```
tgxworld: Let us remove the stub here as well

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+  describe '#id_for_slug' do
+    let(:topic) { Fabricate(:post).topic }
+    let(:pm) { Fabricate(:private_message_topic) }
+
+    it "returns JSON for the slug" do
+      get "/t/id_for/#{topic.slug}.json"
+      expect(response).to be_success
+      json = ::JSON.parse(response.body)
+      expect(json).to be_present
```
tgxworld: This line can be removed as well.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+            topic.update!(
+              archetype: Archetype.private_message,
+              category: nil,
+              allowed_users: [topic.user]
+            )
+          end
+
+          context 'when there are no changes' do
+            it 'does not call the PostRevisor' do
+              PostRevisor.any_instance.expects(:revise!).never
```
tgxworld: Let us test for the exact behavior here too

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+    it 'returns a 404 for an ID that is larger than postgres limits' do
+      get "/t/made-up-topic-slug/5014217323220164041.json"
+
+      expect(response.status).to eq(404)
+    end
+
+    context 'a topic with nil slug exists' do
+      before do
+        @nil_slug_topic = Fabricate(:topic)
```
tgxworld: This variable doesn't need to be an instance variable.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+    context 'permission errors' do
+      let(:allowed_user) { Fabricate(:user) }
+      let(:allowed_group) { Fabricate(:group) }
+      let(:secure_category) {
+        c = Fabricate(:category)
+        c.permissions = [[allowed_group, :full]]
+        c.save
+        allowed_user.groups = [allowed_group]
+        allowed_user.save
+        c }
```
tgxworld: Let us use a `do..end` block to be more consistent.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+    end
+
+    it 'records incoming links' do
+      user = Fabricate(:user)
+
+      get "/t/#{topic.slug}/#{topic.id}", params: {
+        u: user.username
+      }
+
+      expect(IncomingLink.count).to eq(1)
```
tgxworld: We should use `expect {}.to_change {}.by(1)`. Otherwise if an incoming link already exists in the test db, this test becomes a false positive.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+      link = IncomingLink.first
+      expect(link.referer).to eq('http://twitter.com')
+    end
+
+    it 'tracks a visit for all html requests' do
+      sign_in(user)
+      get "/t/#{topic.slug}/#{topic.id}"
+      topic_user = TopicUser.where(user: user, topic: topic).first
+      expect(topic_user).to be_present
```
tgxworld: This assertion can be removed too.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+      it "reviews the user for a promotion if they're new" do
+        sign_in(user)
+        user.update_column(:trust_level, TrustLevel[0])
+        get "/t/#{topic.slug}/#{topic.id}.json"
+        user.reload
+        expect(user.trust_level).to eq(1)
+      end
+    end
+
+    context 'filters' do
```
tgxworld: Let's try to remove the stubs in this group of tests.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1464 @@
...
+
+          expect(response).to redirect_to login_path
+        end
+
+        it 'shows the topic if valid api key is provided' do
+          get "/t/#{topic.slug}/#{topic.id}.json", params: { api_key: api_key.key }
+
+          expect(response).to be_successful
+          topic.reload
+          # free test, only costs a reload
```
tgxworld: This comment can be removed since it doesn't add any value.

tgxworld:
Couple of comments but nice work :+1:

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1474 @@
...
+    context 'filters' do
+      def extract_post_stream
+        json = JSON.parse(response.body)
+        json["post_stream"]["posts"].map { |post| post["id"] }
+      end
+
+      before do
+        @post_ids = topic.posts.pluck(:id)
+        22.times do
+          @post_ids << Fabricate(:post, topic: topic).id
```
OsamaSayegh: Thanks for the feedback @tgxworld. I've updated the PR according to your feedback. Could you tell me if it's OK to do a loop like this here?

tgxworld: Hmm can you try benchmarking how long it takes to run this spec? I'm wondering if we can change the chunk size somehow

SamSaffron: yeah forcing a chunk of 22 seems so overkill we should run it on a chunk of 5 ... also 1 spec is enough here, no need to do this very expensive creation destruction 5 times.

spec/requests/topics_controller_spec.rb
```diff
@@ -4,158 +4,1457 @@
...
+      def extract_post_stream
+        json = JSON.parse(response.body)
+        json["post_stream"]["posts"].map { |post| post["id"] }
+      end
+
+      before do
+        TopicView.stubs(:chunk_size).returns(4)
+        @post_ids = topic.posts.pluck(:id)
+        5.times do
+          @post_ids << Fabricate(:post, topic: topic).id
```
OsamaSayegh: @SamSaffron @tgxworld Ok I changed the chunk size to 4 so now we only need to fabricate 5 posts. I think I could even reduce that to 3 if we need to.

tgxworld: I'll get with at little as possible as long as the behavior that we're testing remains the same :)

OsamaSayegh: That's done!

spec/requests/topics_controller_spec.rb
```diff
@@ -1088,35 +1088,18 @@ def extract_post_stream
...
-        expect(extract_post_stream).to eq(@post_ids[0..19])
-      end
-
-      it 'grabs first page when first page is provided' do
-        get "/t/#{topic.slug}/#{topic.id}.json", params: { page: 1 }
-        expect(response).to be_success
-        expect(extract_post_stream).to eq(@post_ids[0..19])
-      end
-
-      it 'grabs correct range when a page number is provided' do
```
tgxworld: @OsamaSayegh I think you misunderstood what @SamSaffron meant. We still need these tests but instead of running them separately in 5 different `it` blocks. We can just group them together under a single `it` block.

OsamaSayegh: Oh, clearly I needed some sleep. That's done ✅

tgxworld:
Thank you for making the changes @OsamaSayegh

PR #5893, opened 2018-05-29
FIX: Try to convert RSS feed from UTF-16 to UTF-8

gschlager:


SamSaffron:
I really like this in theory, but I want to see at least a couple of tests here with problem payloads.

darix:
And you should also check if it not just works with local files. our RSS feed is sent with the proper content type headers. but still the old code that forcefully re-encoded the content did not see it as UTF-8 and then broke the posts.

RSS link for testing https://pixls.us/feed.xml

If there is any problem on our end, I would be happy to hear about it and fix it. But the current code properly imports our RSS feed without breaking things.

fantasticfears:
UTF16 has two variants which is little and big end. Also, there are existing libs for this by guessing real encoding. Google has one lib and ICU has one.

SamSaffron:
@gschlager be sure to follow up once you are back

tgxworld:
@gschlager Ping :P

darix:
JFYI: gitlab is using charlock_holmes

fantasticfears:
@gschlager You made a great point. Just FYI, while trying to convert UTF-8 is a disaster for many language beyond ASCII, it would confuse the future users. They tend to ask why they see a messed up symbols. But they are happy to find an encoding error message and parse it manually (if they use the desktop website, they are happy to choose encoding if they know they can). I'd say it's all because the mess in encoding standards.

gschlager:
OK, I rewrote this to use the `charlock_holmes` gem. I had to [add `libicu-dev` to our Docker container](https://github.com/discourse/discourse_docker/commit/2ad56e6b097b80833695ef644574fd2e511338ed).
⚠️ We need to force a container update before this can be merged!

Discourse will try to respect the `charset` in the `Content-Type` header if it is present and valid. Otherwise it tries to detect the source encoding and convert it to utf-8. When that fails, it forcefully converts to utf-8. I think that should address the concerns raised by @darix and @fantasticfears.

I didn't touch the code of `Email::Receiver` for now. I guess it could use the encoding detection as well.

SamSaffron:
What about https://github.com/jmhodges/rchardet will it do the trick? its a thinner dependency and seems to solve the same problem.

gschlager:
Yeah, let's try the `rchardet` gem. I updated the PR.

darix:
initial testing looks good. but we will test again with the next post on our site.

PR #5894, opened 2018-05-29
REFACTOR: session controller specs to requests

OsamaSayegh:


OsamaSayegh:
@tgxworld the build failed due to a rubocop offence and I've fixed it, but now the build failure seems unrelated to my changes.

PR #5895, opened 2018-05-30
FEATURE: Retry web hook when it is failed

vinothkannans:


tgxworld:
@vinothkannans I feel like this should be behind a site setting.

vinothkannans:
@tgxworld new site setting `retry_web_hook_events` added with default false.

SamSaffron:
this is not feeling right to me. 

1. It retries on any error now (deep in code) unconditionally 
2. There is no control of when it retries ... eg I would have it retry after 1,5,15,60 and then give up. 

This change feels a bit risky to me cause it has potential impact even when site setting is off. rescueing everyting in that exec also does not feel right

Giacomino1:

Simone Bartolini

SamSaffron:
@tgxworld @vinothkannans what is the plan here?

vinothkannans:
> It retries on any error now (deep in code) unconditionally

@SamSaffron I will make it to retry only on web request failures.

> I would have it retry after 1,5,15,60 and then give up.

Yes, it makes sense to have increased time delay between retries. I will follow that.

tgxworld:
@vinothkannans It'll be better not to depend on Sidekiq's retry option. We should just check if the response code is successful or not and just enqueue another job with the same payload that is executed X seconds later.

vinothkannans:
@tgxworld yes, it's a simple and effective solution :+1:

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -110,6 +110,15 @@ def web_hook_request(args, web_hook)
           web_hook_event_id: web_hook_event.id,
           event_type: args[:event_type]
         }, user_ids: User.human_users.staff.pluck(:id))
+
+        if SiteSetting.retry_web_hook_events? && response.status != 200
+          args[:retry_count] ||= 0
+          return if args[:retry_count] > 4
```
tgxworld: Let's try and avoid in-lining magic numbers. If the number is fixed, we should assign it to a constant which would make it clearer in the future what the number stands for.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -110,6 +110,15 @@ def web_hook_request(args, web_hook)
...
+
+          delays = [0, 1, 5, 15, 60]
```
tgxworld: @SamSaffron  Are you OK if we go with an exponential backoff instead? `2 ** retries`? I'll prefer that instead of having to assign a constant for the retries minutes.

tgxworld: @vinothkannans Hmm should we retry immediately the first time it fails? I feel like the first retry should include the delay.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -25,7 +25,7 @@ def execute(args)
           !web_hook.category_ids.include?(args[:category_id]))
 
         raise Discourse::InvalidParameters.new(:payload) unless args[:payload].present?
-        args[:payload] = JSON.parse(args[:payload])
+        args[:payload] = JSON.parse(args[:payload]) unless args[:retry_count].present?
```
tgxworld: Instead of doing this, we should just duplicate a copy of the args at the start and use it when retrying

config/locales/server.en.yml
```diff
@@ -1686,6 +1686,7 @@ en:
     default_categories_watching_first_post: "List of categories in which first post in each new topic will be watched by default."
 
     retain_web_hook_events_period_days: "Number of days to retain web hook event records."
+    retry_web_hook_events: "Retry web hook events if failed."
```
tgxworld: I think we need to improve the description here to include the number of retries and how long it is between each retry.

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -24,6 +24,47 @@
...
+        .to_return(body: 'Invalid Access', status: 403)
+    end
+
+    it 'retry if site setting is enabled' do
+      expect(Jobs::EmitWebHookEvent.jobs.size).to eq(0)
+      subject.execute(
+        web_hook_id: post_hook.id,
+        event_type: described_class::PING_EVENT
+      )
+      expect(Jobs::EmitWebHookEvent.jobs.size).to eq(1)
```
tgxworld: Rspec has a built in helper here to help us DRY up the code

```
expect { }.to change { Jobs::EmitWebHookEvent.jobs.size }.by(1)
```

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -24,6 +24,47 @@
...
+      job = Jobs::EmitWebHookEvent.jobs.first
+      args = job["args"].first
+      expect(args["retry_count"]).to eq(1)
+    end
+
+    it 'does not retry for more than 4 times' do
```
tgxworld: This is related to my comment about magic numbers because if we change the number in the sidekiq job, this description will no longer be accurate.

spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -24,6 +24,47 @@
...
+      expect(Jobs::EmitWebHookEvent.jobs.size).to eq(0)
```
tgxworld: This can be replace with `expect { }.to_not change { Jobs::EmitWebHookEvent.jobs.size }`

vinothkannans:
@tgxworld changes are done.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/f8e91906179afa387cfc7bd8d432e058e46a169a

Thank you @vinothkannans  :+1:

PR #5901, opened 2018-05-31
Correction of pin_notes in Polish translation

kuba-orlik:
The previous version said that every user can pin a topic for themselves (instead of "unpin"), which was quite misleading.


SamSaffron:
hmmm can't this just be entered in transifex?

ZogStriP:
Thanks @kuba-orlik but translations have to be done on transifex otherwise they'd get overwritten the next time we pull them from transifex.

kuba-orlik:
Oh, I get it - thank you for the info, I'll enter it there

PR #5903, opened 2018-05-31
REFACTOR: categories controller specs to requests

OsamaSayegh:


spec/requests/categories_controller_spec.rb
```diff
@@ -39,4 +37,315 @@
       expect(event[:params].first).to eq(category)
     end
   end
+
+  context '#create' do
+    let(:admin) { Fabricate(:admin) }
```
tgxworld: This doesn't seem to be necessary since it has already been declared in line 4

spec/requests/categories_controller_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe CategoriesController do
+  let!(:admin) { Fabricate(:admin) }
```
tgxworld: Do we need to create an admin for every single test? Using the bang version of `let` means we no longer lazy evaluate the block.

OsamaSayegh:
Indeed we don't need to use `let!` here -- I have removed it. thanks for the review @tgxworld <3

PR #5904, opened 2018-06-01
DEV: improve design of site setting default provider

SamSaffron:
This refactors it so "Defaults provider" is only responsible for "defaults"

Locale handling and management of locale settings is moved back into
SiteSettingExtension

This eliminates complex state management using DistributedCache and makes
it way easier to test SiteSettingExtension


ZogStriP:
Yeah @SamSaffron , can you please sign the CLA?! 🤣

PR #5905, opened 2018-06-01
FIX: don't show dashboard warning when s3 settings are globally configured

OsamaSayegh:
https://meta.discourse.org/t/dashboard-warning-when-s3-backups-are-globally-configured/88872?u=osama

### DIFF
--- a/config/discourse_defaults.conf
+++ b/config/discourse_defaults.conf
@@ -175,6 +175,10 @@ s3_access_key_id =
 s3_secret_access_key =
 s3_use_iam_profile = false
 s3_cdn_url =
+enable_s3_uploads =
+enable_s3_backups =
+s3_upload_bucket =
+s3_backup_bucket =
 
 
 ### rate limits apply to all sites
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -896,20 +896,29 @@ files:
   enable_s3_uploads:
     default: false
     client: true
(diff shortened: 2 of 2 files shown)


tgxworld:
@OsamaSayegh This looks good to me. Can you please rebase this? Thanks!

OsamaSayegh:
@tgxworld I've rebased it, but it's not quite ready, I've noticed three failing test cases failed because of this change https://travis-ci.org/discourse/discourse/jobs/386601444 -- particularly because I shadowed the `s3_use_iam_profile` site setting, which made it always evaluate to `false` because the setting is globally configured so by default:

https://github.com/discourse/discourse/blob/c0b0c66a9e043ff905be0b202127b385616f8e30/config/discourse_defaults.conf#L176

Would it cause any issues if we leave the global setting unset instead of giving it default value? Or should I simply unshadow the setting?

SamSaffron:
Yeah we don't need to do this, we should simply check `GlobalSetting` in s3_config_check. No need to add all that shadowed by global stuff it can cause issues if there is overlap.

SamSaffron:
(closing this for attempt #2 ) :)

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

PR #5909, opened 2018-06-04
REFACTOR: user api keys contoller specs to request

OsamaSayegh:


spec/requests/user_api_keys_controller_spec.rb
```diff
@@ -210,10 +201,9 @@
 
       # should overwrite if needed
       args["access"] = "pr"
-      post :create, params: args, format: :json
+      post "/user-api-key.json", params: args
 
-      expect(response.code).to eq("302")
+      expect(response.status).to eq(302)
```
SamSaffron: I like that status is a number ;p so much nicer that using a string

SamSaffron:
thanks!

PR #5911, opened 2018-06-04
REFACTOR: topic controller (2) specs to requests

OsamaSayegh:


spec/requests/posts_controller_spec.rb
```diff
@@ -423,6 +423,62 @@
...
+
+      # choosing an arbitrarily easy to mock trusted activity
+      it 'allows users with api key to bookmark posts' do
+        put "/posts/#{post.id}/bookmark.json", params: {
+          bookmarked: "true",
+          api_key: api_key.key
+        }
+
+        expect(response).to be_success
+        expect(PostAction.where(post: post, user: user).count).to eq(1)
```
tgxworld: We need to test for the type of `PostAction` here. Otherwise, a flag post action being incorrectly created would also allow this test to pass.

spec/requests/posts_controller_spec.rb
```diff
@@ -423,6 +423,62 @@
...
+      end
+
+      it 'allows users with a master api key to bookmark posts' do
+        put "/posts/#{post.id}/bookmark.json", params: {
+          bookmarked: "true",
+          api_key: master_key.key,
+          api_username: user.username
+        }
+
+        expect(response).to be_success
```
tgxworld: We need to assert for the post action created here.

spec/requests/topics_controller_spec.rb
```diff
@@ -1168,6 +1181,219 @@ def extract_post_stream
...
+
+      before do
+        sign_in(user)
+      end
+
+      it "selects the theme the user has selected" do
+        user.user_option.update_columns(theme_key: theme.key)
+
+        get "/t/666"
+        expect(controller.theme_key).to eq(theme.key)
```
tgxworld: Hmm is `controller` made available somewhere in the test?

spec/requests/topics_controller_spec.rb
```diff
@@ -1168,6 +1181,219 @@ def extract_post_stream
...
+
+    describe "themes" do
+      let(:theme) { Theme.create!(user_id: -1, name: 'bob', user_selectable: true) }
+      let(:theme2) { Theme.create!(user_id: -1, name: 'bobbob', user_selectable: true) }
+
+      before do
+        sign_in(user)
+      end
+
+      it "selects the theme the user has selected" do
```
tgxworld: A bunch of the tests are missing assertion for the response code.

spec/requests/topics_controller_spec.rb
OsamaSayegh: From my observation, it seems the `controller` (along with `session`, `request` etc) objects become available after you make the request. So in this case, `controller` is an instance of TopicController that handled the request above.

PR #5914, opened 2018-06-04
REFACTOR: onebox controller specs to requests

OsamaSayegh:


spec/requests/onebox_controller_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+  let(:url) { "http://google.com" }
+
+  it "requires the user to be logged in" do
+    get "/onebox.json", params: { url: url }
+    expect(response.status).to eq(403)
+  end
+
+  describe "logged in" do
+    let(:user) { Fabricate(:user) }
+    let(:html) {
```
tgxworld: We prefer to use `do..end` for multi-line blocks

OsamaSayegh: Done ✅

tgxworld:
:+1:

PR #5915, opened 2018-06-04
FIX: sharing popup not showing on macos/chrome

jjaffeux:
Despite `navigator.share` being defined the call was failing with this error:

```
sharing DOMException: Internal error: could not connect to Web Share interface.
```


xfalcox:
Where exactly do you get this error? MacOS Safari?

PR #5922, opened 2018-06-05
A few fixes to the VBulletin importer

discoursehosting:


script/import_scripts/vbulletin.rb
```diff
@@ -140,20 +149,26 @@ def import_users
...
             import_profile_picture(user, u)
             import_profile_background(user, u)
           end
         }
       end
     end
+
+    @usernames = {}
+    UserCustomField.where(name: 'import_username').pluck(:user_id, :value).each do |user_id, import_username|
+      @usernames[import_username] = user_id
```
gschlager: I guess you moved the username mapping from `post_create_action` to this location in order to make the importer resumable or was there another reason? But why store the `user_id` instead of the username itself? The following should work...

```ruby
@usernames = UserCustomField.joins(:user).where(name: 'import_username').pluck('user_custom_fields.value', 'users.username').to_h
```

discoursehosting: Lazyness in two senses:
- as in: lazy programmer
- but also as in lazy evaluation: delay user lookup until it is actually needed (and yes one join may be faster than 100 separate lookups, and that is where reason #1 comes into play)

gschlager: Would you mind not being a lazy programmer in this case? 😜 
I'd appreciate it if you could update your PR so that it uses a join for loading the usernames.

discoursehosting: We already finished this import and I think the performance impact is negligible so I'm not really eager to revisit and test this all over again, especially since this is already working correctly.

discourse-plugins: Seems we have to do some fixes anyway so yeah I will change this while I'm at it, coming up in a few days, please keep this open in the meanwhile

gschlager: Sure and thanks for fixing the importer! 👍

gschlager:
LGTM! And thanks for "make gschlager happy :)" 😃 
Is it ready to be merged?

discoursehosting:
Yep, ready to be merged!

gschlager:
Thanks! 👍

I'd appreciate it, if you could make Rubocp happy as well the next time you submit a pull request. 🙂 
https://meta.discourse.org/t/rubocop-has-landed-on-discourse-policewoman-policeman/67211

PR #5923, opened 2018-06-05
FIX: do not verify group visibility when checking for mentionable/messageable

techAPJ:
Fixes https://meta.discourse.org/t/urls-for-group-message-error-for-private-groups/70631

@tgxworld Can you review this PR please? Thanks!

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -282,7 +282,7 @@ class GroupsController < ApplicationController
   end
 
   def mentionable
-    group = find_group(:name)
+    group = find_group(:name, ensure_can_see: false)
 
     if group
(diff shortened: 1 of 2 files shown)


app/controllers/groups_controller.rb
```diff
@@ -468,12 +468,11 @@ def group_params(automatic: false)
     params.require(:group).permit(*permitted_params)
   end
 
-  def find_group(param_name)
+  def find_group(param_name, ensure_can_see = true)
```
tgxworld: I prefer to use keyword arguments now a days because `find_group(:name, ensure_can_see: true)` is clearer than `find_group(:name, true)`

techAPJ: Done.

PR #5925, opened 2018-06-05
fix get from url for external uploads without CDN

majakomel:
This PR fixes the problem that appears when using S3 for storing uploads (or Azure Blob plugin) without CDN - get_from_url misses video uploads with absolute_base_url, causing them to get orphaned and thombstoned (post_upload record doesn't get created)

Related to this: https://meta.discourse.org/t/azure-blob-storage-plugin/79392/26?u=maja


SamSaffron:
Lets try this out, thanks @majakomel

PR #5928, opened 2018-06-05
Prefer throwing a new Error object instead of just a string expression

kleinfreund:
In Discourse’s JavaScript code, a lot of error reporting is done in the form `throw 'message'`. In some cases, this hides the error message completely and instead generates a console output that looks like this.

When [discourse/app/assets/javascripts/discourse/models/rest.js.es6: **createProperties**](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/rest.js.es6#L62) throws:

```
undefined              ember:19818:7
defaultDispatch        ember:19818:7
dispatchError          ember:19797:7
run                    ember:2560:9
run                    ember:24451:12
handler                ember:13785:7
setupHandler/<         ember:44352:15
dispatch               jquery:5183:16
add/elemData.handle    jquery:4992:6
```

Note that the top-level error message will be displayed as `undefined`. Instead, the error should be reported as `throw new Error('message')`, generating a way more useful console output:

```
You must overwrite `createProperties()` before saving a record
createProperties@discourse/models/rest:79:13
_saveNew@discourse/models/rest:50:24
save@discourse/models/rest:82:34
createRoute@javascripts/discourse/controllers/admin-plugins-name:23:9
send@ember:34220:28
handler/</<@ember:13806:13
flaggedInstrument@ember:20589:14
handler/<@ember:13805:11
run@ember:2558:16
run@ember:24451:12
handler@ember:13785:7
setupHandler/<@ember:44352:15
dispatch@jquery:5183:16
add/elemData.handle@jquery:4992:6
```

I looked through the source code and changed most occurrences that didn’t follow the pattern `throw new Error(...)`. Note that I didn’t attempt to be more specific about the type of error (e.g. ReferenceError, TypeError, etc.) to throw as I’m not very familiar with the code base. Such a decision should be done by someone else.

Furthermore, I don’t know if these kinds of changes need tests and if so, how the would look like. If this is required, let me know.

*(I’m unsure whether this requires the prefix `FIX` or another prefix, so I left it out.)*

### DIFF
--- a/app/assets/javascripts/admin/services/admin-tools.js.es6
+++ b/app/assets/javascripts/admin/services/admin-tools.js.es6
@@ -127,7 +127,7 @@ export default Ember.Service.extend({
                 if (result.deleted) {
                   resolve();
                 } else {
-                  throw 'failed to delete';
+                  throw new Error('failed to delete');
                 }
               }).catch(() => {
                 bootbox.alert(I18n.t("admin.user.delete_failed"));
--- a/app/assets/javascripts/discourse/adapters/rest.js.es6
+++ b/app/assets/javascripts/discourse/adapters/rest.js.es6
@@ -20,7 +20,7 @@ export function Result(payload, responseJson) {
 // We use this to make sure 404s are caught
 function rethrow(error) {
   if (error.status === 404) {
-    throw "404: " + error.responseText;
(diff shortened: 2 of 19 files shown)


SamSaffron:
thanks @kleinfreund !!! merging it in

PR #5945, opened 2018-06-06
FIX: 302 for all unknown service worker paths

featheredtoast:
Not sure if this is a good idea, but older service workers should find *something*, right?

This is a little more heavy handed that previous attempts - in this case, any older service worker files will be 302'd to the newer one.


tgxworld:
I don't think we should do this. I think the mistake I made is to have a service worker route that keeps changing which is incorrect as you pointed out. Is there context of why we can't just let an old service worker 404?

PR #5947, opened 2018-06-07
BUGFIX: string in server_plugin_outlet needs to be mutable

angusmcleod:
Strings in application_helper are now implicitly frozen since the addition of the ``# frozen_string_literal: true`` flag. The string in question needs to be mutable however otherwise an error is thrown when you attempt to use a server plugin outlet.


gschlager:
Thanks 👍

PR #5958, opened 2018-06-08
REFACTOR: admin site texts controller specs to requests

OsamaSayegh:
While working on this, I noticed that if we make requests like `get "/admin/customize/site_texts/title.json"` we would get a 404 error because `.json` was interpreted as a part of the translation key rather than the format of the request. So I added those new routes to fix this issue.

Let me know if this is OK -- if it's not, we can specify `HTTP_ACCEPT` header for the request as a workaround.


spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -21,18 +25,49 @@
...
+    describe '#show' do
+      it 'returns a site text for a key that exists' do
+        get "/admin/customize/site_texts/js.topic.list.json"
+        expect(response.status).to eq(200)
+
+        json = ::JSON.parse(response.body)
+        expect(json).to be_present
+
+        site_text = json['site_text']
+        expect(site_text).to be_present
```
tgxworld: This line is not required

spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -21,18 +25,49 @@
...
-    context "when logged in as admin" do
-      before do
-        sign_in(admin)
+    describe '#show' do
+      it 'returns a site text for a key that exists' do
+        get "/admin/customize/site_texts/js.topic.list.json"
+        expect(response.status).to eq(200)
+
+        json = ::JSON.parse(response.body)
+        expect(json).to be_present
```
tgxworld: This line is not required

spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -68,6 +103,108 @@
...
+      it 'updates and reverts the key' do
+        orig_title = I18n.t(:title)
+
+        put "/admin/customize/site_texts/title.json", params: { site_text: { value: 'hello' } }
+        expect(response.status).to eq(200)
+
+        json = ::JSON.parse(response.body)
+
+        site_text = json['site_text']
+        expect(site_text).to be_present
```
tgxworld: This line is not required

spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -68,6 +103,108 @@
...
+        expect(site_text['value']).to eq('hello')
+
+        # Revert
+        delete "/admin/customize/site_texts/title.json"
+        expect(response.status).to eq(200)
+
+        json = ::JSON.parse(response.body)
+
+        site_text = json['site_text']
+        expect(site_text).to be_present
```
tgxworld: This line is not required

OsamaSayegh:
@tgxworld done 👍

PR #5959, opened 2018-06-08
add user-preferences-emails-pref-email-settings outlet

LeoMcA:
Necessary to add preferences to the bottom of the Email section, like so:

screenshot from 2018-06-08 15-02-29

I'll be using it in the plugin I'm making described in this topic: https://meta.discourse.org/t/per-user-preference-mark-post-as-read-when-emailed/86256


SamSaffron:
This seems fine to me! thanks!

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

PR #5974, opened 2018-06-11
REFACTOR: admin emojis controller specs to requests

OsamaSayegh:


spec/requests/admin/emojis_controller_spec.rb
```diff
@@ -8,6 +8,19 @@
...
+  describe '#index' do
+    it "returns a list of custom emojis" do
+      CustomEmoji.create!(name: 'osama-test-emoji', upload: upload)
+      Emoji.clear_cache
+
+      get "/admin/customize/emojis.json"
+      expect(response.status).to eq(200)
+
+      json = ::JSON.parse(response.body)
+      expect(json[0]["name"]).to eq("osama-test-emoji")
```
tgxworld: It looks like we're no longer testing for the emoji's URL

OsamaSayegh: Good catch -- fixed 👍

PR #5982, opened 2018-06-11
FIX: use the correct notification levels descriptions for PMs

OsamaSayegh:
https://meta.discourse.org/t/notification-levels-watching-etc-all-descriptions-incorrect-for-private-messages/87826?u=osama


SamSaffron:
Nice such a small change, do you mind adding a small test for this, its worth practicing tests on JS

OsamaSayegh:
Yea, I've not written JS tests before.. this is my first attempt, let me know if it looks good 🐳

SamSaffron:
excellent!

PR #5988, opened 2018-06-12
UX: keyboard shortcuts (j/k) should work with /categories

OsamaSayegh:
Meta: https://meta.discourse.org/t/keyboard-shortcut-k-j-doesnt-work-on-categories-page/15948?u=osama

### DIFF
--- a/app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6
+++ b/app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6
@@ -1,72 +1,73 @@
-import DiscourseURL from 'discourse/lib/url';
-import Composer from 'discourse/models/composer';
-import { minimumOffset } from 'discourse/lib/offset-calculator';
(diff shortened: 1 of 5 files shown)


SamSaffron:
I worry that we are being a bit too fancy here and carrying too much code. 

Perhaps start with a very very simple PR with none of the tab hijacking that just allows j/k on the topic list. I am not sure we really need fancy j/k navigation in the categories section and can just get away with standard tab if users want to do that.

OsamaSayegh:
Yeah, I was doubtful that a minor improvement like this is worth adding 100+ lines, but I didn't want to make that judgment. Will simplify this early next week.

OsamaSayegh:
@SamSaffron I've simplified this to only support topics lists on /categories -- let me know what you think :comet:

SamSaffron:
much better! merging in 👍

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

PR #5989, opened 2018-06-12
Feature - allow user and group specific cards styling

featheredtoast:
An idea from working on the user cards today is to expose the user or group name in the classnames so they can be styled individually.

### DIFF
--- a/app/assets/javascripts/discourse/components/group-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/group-card-contents.js.es6
@@ -9,7 +9,7 @@ export default Ember.Component.extend(CardContentsBase, CleansUp, {
   elementId: 'group-card',
   triggeringLinkClass: 'mention-group',
   classNames: ['no-bg'],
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/group-card-contents.js.es6
```diff
@@ -23,6 +23,9 @@ export default Ember.Component.extend(CardContentsBase, CleansUp, {
   @computed('group.user_count', 'group.members.length')
   moreMembersCount: (memberCount, maxMemberDisplay) => memberCount - maxMemberDisplay,
 
+  @computed('group.name')
+  groupClass: (name) => name ? `group-card-${name}` : '',
```
tgxworld: Is there a case where `group.name` or `username` would be `null`?

featheredtoast: From my testing, username gets set to null and group.name is undefined when the card is hidden

tgxworld: :+1:

PR #5992, opened 2018-06-13
FIX: theme CSS should recompile when theme uploads change

OsamaSayegh:
https://meta.discourse.org/t/how-to-replace-scss-vars-in-themes-uploads/84699?u=osama


SamSaffron:
great fix!

PR #5994, opened 2018-06-14
Feature: add jump to post modal on mobile

featheredtoast:
...rather than using browser prompt(). Specify input as a number (easier for mobile soft keyboards), and auto-focus (one less click).

Rearranges "fullscreen" under the modal, and supports click-out to dismiss/cancel.

Gif is worth 1000 words:

mobile-jump-to-post


featheredtoast:
@awesomerobot one thing that I'd like to just be sure about is if there was a specific reason why the "fullscreen" z-index was over the modal overlays, and if there might be an issue if this were rearranged?

featheredtoast:
great, thanks! 🎉

PR #5997, opened 2018-06-14
Feature: only install service workers for android.

featheredtoast:
canonical path - service-worker2.js
browsers might have a 301 cached on service-worker.js.


SamSaffron:
I don't think this will cut it and @tgxworld is going to need to work through this. We have to still install the service worker if a plugin opts for it for desktop. 

Out of the box we should not install it though.

Giacomino1:

Simone Bartolini

tgxworld:
Closing this in favor of https://github.com/discourse/discourse/commit/af5cd7e548447b6aa86b34b069cc4e88c0fd1256 first.

PR #6010, opened 2018-06-19
FEATURE: Create hidden posts for received spam emails

paroga:
Spamchecker usually have 3 results: HAM, SPAM and PROBABLY_SPAM
SPAM gets usually directly rejected and needs no further handling.
HAM is good message and usually gets passed unmodified.
PROBABLY_SPAM gets an additional header to allow further processing.
This change addes processing capabilities for such headers and marks
new posts created as hidden when received via email.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -113,7 +113,8 @@ class Post < ActiveRecord::Base
     @hidden_reasons ||= Enum.new(flag_threshold_reached: 1,
                                  flag_threshold_reached_again: 2,
                                  new_user_spam_threshold_reached: 3,
-                                 flagged_by_tl3_user: 4)
+                                 flagged_by_tl3_user: 4,
+                                 email_spam_header_found: 5)
   end
 
   def self.types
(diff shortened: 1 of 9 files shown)


config/locales/server.de.yml
```diff
@@ -1363,6 +1363,7 @@ de:
     log_mail_processing_failures: "Protokolliere Fehler bei der E-Mail-Verarbeitung in http://yoursitename.com/logs"
     email_in: "Erlaube Benutzern, neue Themen per E-Mail einzutragen (erfordert manuelles oder POP3-Polling). Konfiguriere die Adressen im „Einstellungen“-Reiter in jeder Kategorie."
     email_in_min_trust: "Minimale Vertrauensstufe um neue Themen per E-Mail erstellen zu können."
+    email_in_spam_header: "Der E-Mail-Header für die Spam-Erkennung."
```
ZogStriP: Quick note: this will be overwritten once we retrieve the translations from transifex. It's best to only provide English strings and then translate them on transifex.

ZogStriP:
Sorry for the delay @paroga. This looks good. What application/service generates these headers?

paroga:
I removed the german translation from the pull request.

> What application/service generates these headers?

e.g. SpamAssassin and Rspamd

ZogStriP:
Thanks 👍

PR #6015, opened 2018-06-20
FEATURE: add website field to SSO

realdlee:
Allows option of including a user website field with sso payload.

https://meta.discourse.org/t/passing-user-website-data-in-sso-payload/87470/2


SamSaffron:
Thanks, yes this looks good to me!

PR #6016, opened 2018-06-20
FIX: Don't count the current user in "Others with this badge."

misaka4e21:
https://meta.discourse.org/t/count-at-others-with-this-badge-includes-current-user/90089

### DIFF
--- a/app/assets/javascripts/discourse/controllers/badges/show.js.es6
+++ b/app/assets/javascripts/discourse/controllers/badges/show.js.es6
@@ -28,6 +28,11 @@ export default Ember.Controller.extend(BadgeSelectController, {
     return username ? userCount : modelCount;
   },
 
+  @computed("model.grant_count", "userBadges.grant_count")
+  othersCount(modelCount, userCount) {
+    return modelCount - userCount;
+  },
+
   @computed("model.allow_title", "model.has_badge", "model")
   canSelectTitle(hasTitleBadges, hasBadge) {
     return this.siteSettings.enable_badges && hasTitleBadges && hasBadge;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/controllers/badges/show.js.es6
```diff
@@ -25,7 +25,7 @@ export default Ember.Controller.extend(BadgeSelectController, {
 
   @computed("username", "model.grant_count", "userBadges.grant_count")
   grantCount(username, modelCount, userCount) {
-    return username ? userCount : modelCount;
+    return username ? userCount : modelCount - 1;
```
tgxworld: hmm how do we know if the user hass been granted that badge or not?

misaka4e21: I added a new computed variable `othersCount` to exclude the current user. It will be shown only if `canShowOthers` is true.

app/assets/javascripts/discourse/controllers/badges/show.js.es6
```diff
@@ -28,6 +28,11 @@ export default Ember.Controller.extend(BadgeSelectController, {
     return username ? userCount : modelCount;
   },
 
+  @computed("model.grant_count")
+  othersCount(grant_count) {
+    return grant_count - 1;
```
SamSaffron: Oh, this is not going to be a simple -1, you could have multiple badges of the type, so you need to subtract the number you have.

misaka4e21: Changed.
Should I consider that other people can also have multiple badges per person?

PR #6023, opened 2018-06-20
FEATURE: Support disabling tag creation for non-staff users.

misaka4e21:


nlalonde:
I was thinking we could add a new value 'staff' to the existing settings, like here:

https://meta.discourse.org/t/limit-tag-creation-to-staff/89919

PR #6028, opened 2018-06-22
UX: Preview multiple color schemes in wizard.

misaka4e21:
It was a dropdown to provide choices of color schemes,
and only one scheme could be shown.
With this commit, multiple color scheme previews can be displayed on
one page at the same time, making admins choose color schemes more
easily.


SamSaffron:
@misaka4e21  I think this PR is a tiny bit too early, we need to provide you with the new themes first prior to sending it through, also make sure you integrate the UX changes proposed by Kris! 

Thanks heaps, closing for now.

PR #6029, opened 2018-06-22
FEATURE: send message when a user reaches tl1

featheredtoast:
Sends a welcome message when a user hits tl1

Can be disabled via the `send_tl1_welcome_message` site setting.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -313,6 +313,11 @@ class User < ActiveRecord::Base
     Jobs.enqueue(:send_system_message, user_id: id, message_type: message_type)
   end
 
+  def enqueue_member_welcome_message
+    return unless SiteSetting.send_tl1_welcome_message?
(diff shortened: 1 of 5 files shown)


coding-horror:
I'd say go ahead and merge this unless you feel it is especially risky and
needing review.

featheredtoast:
Not especially - merged! 👍 

The copy needs changing, and the site setting may need to get moved, but the general functionality is good to go.

SamSaffron:
If we are sending this message we should probably make sure there is no
badge notification for tl1 (if message is enabled)

Otherwise you reach tl1 and get one blue and one green notification which
is odd.

coding-horror:
I don't think it's that weird, no different than a like notification
really.. those are trivial

... and if you had your way, every user would be notified by default on
EVERY SINGLE like, just sayin' ;)

PR #6032, opened 2018-06-25
FIX: display tables in posts history diff

OsamaSayegh:
I'm changing Régis code that he added for this [security fix](https://github.com/discourse/discourse/commit/4fb41663b3c7071dc1ef7d92eb3e5a6516dfe3b5), so I'd like him to review this when he comes back :)

### DIFF
--- a/app/assets/javascripts/discourse/controllers/history.js.es6
+++ b/app/assets/javascripts/discourse/controllers/history.js.es6
@@ -255,7 +255,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
       const tableTags = ["table", "thead", "tbody", "th", "tr", "td"];
       const whiteListObj = { features: { editHistory: true } };
 
-      tableTags.forEach(tag => whiteListObj.features[tag] = true);
+      tableTags.forEach(tag => (whiteListObj.features[tag] = true));
       const whiteLister = new WhiteLister(whiteListObj);
 
       whiteLister.whiteListFeature("editHistory", { custom: () => true });
--- a/app/assets/javascripts/discourse/controllers/history.js.es6
+++ b/app/assets/javascripts/discourse/controllers/history.js.es6
@@ -2,9 +2,9 @@ import ModalFunctionality from "discourse/mixins/modal-functionality";
 import { categoryBadgeHTML } from "discourse/helpers/category-link";
 import computed from "ember-addons/ember-computed-decorators";
 import { propertyGreaterThan, propertyLessThan } from "discourse/lib/computed";
-import { on } from "ember-addons/ember-computed-decorators";
+import { on, observes } from "ember-addons/ember-computed-decorators";
 import { default as WhiteLister } from "pretty-text/white-lister";
-import { sanitize } from "pretty-text/sanitizer";
+import { sanitizeAsync } from "discourse/lib/text";
 
 function customTagArray(fieldName) {
   return function() {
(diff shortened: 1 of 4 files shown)


Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

SamSaffron:
This is fine, but we need 1 test for it 👍

OsamaSayegh:
Added a test ✔️

tgxworld:
@OsamaSayegh Could we also look at adding a test case for the XSS fix? It is unclear to me what the original security fix was trying to prevent at the moment. Also it seems like `td` is vulnerable to XSS as per https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#TABLE

SamSaffron:
> Also it seems like td is vulnerable to XSS as per  

But the code is all sanitized ... so I am not following. The vul is if somehow you could convince the code that generates the diff to add in a new tag. 

A very simple 100% fix here would be to pass the HTML we get out of the diff through a whitelister. This may be a bit of overkill though.

app/assets/javascripts/discourse/controllers/history.js.es6
```diff
@@ -252,8 +252,15 @@ export default Ember.Controller.extend(ModalFunctionality, {
     if (viewMode === "side_by_side_markdown") {
       return html;
     } else {
-      const whiteLister = new WhiteLister({ features: { editHistory: true } });
+      const tableTags = ["table", "thead", "tbody", "th", "tr", "td"];
```
SamSaffron: yeah this is not sitting right ... we have our default markdown whitelister ... why can we not use it.

SamSaffron: https://github.com/discourse/discourse/blob/b54ba4c952d79d56742e76081d692f1281e88512/app/assets/javascripts/pretty-text/engines/discourse-markdown/table.js.es6#L31-L39

SamSaffron: so can't we just ask pretty text for the whitelister it uses ? and simply reuse that?

OsamaSayegh: I've tried using the same whitelister that pretty-text uses and it did fix this particular issue, however, it introduced a whole new bunch of issues. Mainly it messed up the layout and diff coloring (green and red) because all `diff-ins` and `dif-del` classes were stripped from diff HTML.

All this pain is caused by this bit here:

https://github.com/discourse/discourse/blob/9f558d7e74118d8dc0cb1db04f2b7200f5a3618c/lib/discourse_diff.rb#L248-L253

Specifically, Nokogiri Parser. It seems to be decoding HTML entities when parsing our HTML and returning HTML that's not absolutely safe and needs to be sanitized again.

Looks like there is a way to disable HTML decoding, but it doesn't work :/ (see [this issue](https://github.com/sparklemotion/nokogiri/issues/1284)).

SamSaffron: re: `diff-ins and dif-del` that is fine, I would then take the whitelister, clone it and add the 2 new classes you need. I follow about nokogiri making trouble.

OsamaSayegh: Just to be sure before I start working on this, I'll need to whitelist those two classes with every HTML tag we allow in posts (I don't think I'll have to list all the tags manually in an array, I think I can get them from existing code). Is that OK?

coding-horror:
This seems like quite a lot of pain for a rather trivial thing to "fix".. are we sure this is all worth it?

OsamaSayegh:
Ok, never mind my last response. I took a break from this and came back to get it accidentally to work. I simply added back Regis' code and used our pretty-text whitelister and now it works. ✨ 

Let me know if it looks OK, and I'll fix tests and add a couple more for the security issue and this bug.

SamSaffron:
let me know when this is ready to merge

OsamaSayegh:
@SamSaffron Ok, I think this is now in good shape.

SamSaffron:
yay, yes this is exactly what I wanted! great work

PR #6033, opened 2018-06-25
FIX: don't show dashboard warning when s3 settings are globally confi…

OsamaSayegh:
…gured (2)

### DIFF
--- a/config/discourse_defaults.conf
+++ b/config/discourse_defaults.conf
@@ -174,6 +174,9 @@ s3_access_key_id =
 s3_secret_access_key =
 s3_use_iam_profile = false
 s3_cdn_url =
+s3_backup_bucket =
+enable_s3_uploads =
+enable_s3_backups =
 
 
 ### rate limits apply to all sites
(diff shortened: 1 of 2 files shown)


Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

config/discourse_defaults.conf
```diff
@@ -174,6 +174,9 @@ s3_access_key_id =
 s3_secret_access_key =
 s3_use_iam_profile = false
 s3_cdn_url =
+s3_backup_bucket =
```
SamSaffron: we don't need to add these though... enable_s3_uploads is implied if it is specified via globals

OsamaSayegh: Yes, but it looks like if a site doesn't specify those settings in globals then `s3_config_check` is going blow up with method undefined errors. Also tests are failing with the same error if I remove those.

SamSaffron:
OK, I went with this simpler fix for now, yeah this is hairy code. 

https://github.com/discourse/discourse/commit/330e848d4a0f8ff867ef8c52901f33b10a4ece63

PR #6035, opened 2018-06-25
FEATURE: Second factor backup

majakomel:
I'd like to ask for some help to push this forward. Main functionality is there but some things could be done better. It would be great to get some input to move into right direction, there are 2 things I'd particularly like to do better:
1. Currently there are lots of magic numbers going on (secondFactorMethod) in the frontend, where would the best place to store them as constants? 
2. Initially I stored the codes in an array but after a chat with @featheredtoast each code is stored in it's own record with json data - {sha: "", backup_code: ""}, which is somehow duplicated since each needs the same salt to check against and it gets repeated in each entry. What would be the better way to do it?

### DIFF
--- a/app/assets/javascripts/discourse/components/backup-codes.js.es6
+++ b/app/assets/javascripts/discourse/components/backup-codes.js.es6
@@ -1,0 +1,58 @@
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  classNames: ["backup-codes"],
+  backupCodes: null,
+
+  click(event) {
+    if (event.target.id === "backupCodes") {
+      this._selectAllBackupCodes();
+    }
+  },
+
+  didRender() {
+    this._super();
+
+    const $backupCodes = this.$("#backupCodes");
+    if ($backupCodes.length) {
+      $backupCodes.height($backupCodes[0].scrollHeight);
+    }
+  },
+
+  @computed("formattedBackupCodes")
+  base64BackupCode(formattedBackupCodes) {
+    return this._b64EncodeUnicode(formattedBackupCodes);
+  },
+
+  @computed("backupCodes")
+  formattedBackupCodes(backupCodes) {
+    if (!backupCodes) return null;
+
+    return backupCodes.join("\n").trim();
+  },
+
+  actions: {
+    copyToClipboard() {
+      this._selectAllBackupCodes();
+      this.get("copyBackupCode")(document.execCommand("copy"));
+    }
+  },
+
+  // https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
+  _b64EncodeUnicode(str) {
+    return btoa(
+      encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, function toSolidBytes(
+        match,
+        p1
+      ) {
+        return String.fromCharCode("0x" + p1);
+      })
+    );
+  },
+
+  _selectAllBackupCodes() {
+    const $textArea = this.$("#backupCodes");
+    $textArea[0].focus();
+    $textArea[0].setSelectionRange(0, this.get("formattedBackupCodes").length);
+  }
+});
--- a/app/assets/javascripts/discourse/components/backup-codes.js.es6
+++ b/app/assets/javascripts/discourse/components/backup-codes.js.es6
@@ -1,5 +1,17 @@
 import computed from "ember-addons/ember-computed-decorators";
 
+// https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
+function b64EncodeUnicode(str) {
+  return btoa(
+    encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, function toSolidBytes(
+      match,
+      p1
+    ) {
+      return String.fromCharCode("0x" + p1);
+    })
+  );
+}
+
 export default Ember.Component.extend({
   classNames: ["backup-codes"],
   backupCodes: null,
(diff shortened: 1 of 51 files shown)


jjaffeux:
@majakomel concerning the magic numbers, I think that given it's user related, you could put it in User model and import it wherever it's needed.

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

Giacomino1:

Simone Bartolini

app/assets/javascripts/discourse/components/backup-codes.js.es6
```diff
@@ -0,0 +1,58 @@
...
+
+  actions: {
+    copyToClipboard() {
+      this._selectAllBackupCodes();
+      this.get("copyBackupCode")(document.execCommand("copy"));
+    }
+  },
+
+  // https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
+  _b64EncodeUnicode(str) {
```
eviltrout: This is a small thing, but I noticed `_b64EncodeUnicode` doesn't use `this` in any way, so it's a pure function. For such functions, I often prefer to put them outside of the class as a regular function. Funnily enough, it is more "private" because it is not exported from your module, and can only be called by the class (whereas any code could call `_b64EncodeUnicode` if they wanted).

Once you have a pure function, you can also shorten the `base64BackupCode` computed property by a couple of lines up on L23:

```
@computed("formattedBackupCodes")
base64BackupCode: b64EncodeUnicode
```

app/assets/javascripts/discourse/components/second-factor-form.js.es6
```diff
@@ -0,0 +1,36 @@
+import computed from "ember-addons/ember-computed-decorators";
+
```
eviltrout: By looking at this code, I don't know that `1` or `2` means.

You should use constants for this:

```
export const SECOND_FACTOR_CODE = 1;
export const SECOND_FACTOR_BACKUP = 2;
```

Then use those constants everywhere that you use `1` and `2` in the files below.

app/assets/javascripts/discourse/controllers/preferences/second-factor-backup.js.es6
```diff
@@ -0,0 +1,104 @@
...
+          I18n.t("user.second_factor_backup.copied_to_clipboard")
+        );
+      } else {
+        this.set(
+          "errorMessage",
+          I18n.t("user.second_factor_backup.copy_to_clipboard_error")
+        );
+      }
+
+      setTimeout(() => {
```
eviltrout: I think you should add a comment here explaining why there's a timeout as I was a little confused at first. It seems to be to hide the message 2 seconds after it's shown?

Also the code should use `Ember.run.later` instead of`setTimeout` is not run within Ember's runloop.

app/models/concerns/second_factor_manager.rb
```diff
@@ -23,18 +24,88 @@ def authenticate_totp(token)
...
+
+  def generate_backup_codes
+    codes = []
+    10.times do
+      codes << SecureRandom.hex(8)
+    end
+
+    salt = SecureRandom.hex(16)
+    codes_json = codes.map do |code|
+      { :salt => salt,
```
eviltrout: Trivial but we prefer the `salt: salt, code_hash: etc` syntax. I'm not sure why rubocop didn't mention this!

app/views/users/admin_login.html.erb
```diff
@@ -8,11 +8,27 @@
...
+
+          <div id="backup-second-factor-form" style="display: none">
+            <%= form_tag({}, method: :put) do%>
+              <%= label_tag(:second_factor_token, t("login.second_factor_backup_description")) %>
+              <%= render 'common/second_factor_backup_input' %><br><br>
+              <%= submit_tag(t("submit")) %>
+            <%end%>
+
+          </div>
+          <a href="#" id="toggle-form"><%=t "login.second_factor_backup" %></a>
```
eviltrout: In general it's better to do `<a href>` instead of `<a href='#'>` - the first one styles the same way, but doesn't add a hash to the URL if the user clicks it.

jjaffeux:
@featheredtoast can you review the server side of this PR please ?

app/models/concerns/second_factor_manager.rb
```diff
@@ -23,18 +24,88 @@ def authenticate_totp(token)
...
+    end
+  end
+
+  def generate_backup_codes
+    codes = []
+    10.times do
+      codes << SecureRandom.hex(8)
+    end
+
+    salt = SecureRandom.hex(16)
```
featheredtoast: Now that the schema allows it, we should generate and store a different salt per code, so that we are not re-using salts.

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -19,6 +19,8 @@ import PreloadStore from "preload-store";
 import { defaultHomepage } from "discourse/lib/utilities";
 import { userPath } from "discourse/lib/url";
 
+export const SECOND_FACTOR_METHODS = { totp: 1, backup_code: 2 };
```
eviltrout: Constants should always be upper case. I see the `SECOND_FACTOR_METHODS` is, but to be clear the properties should be too. `SECOND_FACTOR_METHODS.TOTP` for example.

app/assets/javascripts/discourse/templates/components/backup-codes.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="wrapper">
+  <textarea id="backupCodes" class="backup-codes-area" readonly>{{formattedBackupCodes}}</textarea>
+
+  {{d-button
+    action="copyToClipboard"
+    class="backup-codes-copy-btn"
+    icon="copy"}}
+
+  <a download="backup_codes.txt"
```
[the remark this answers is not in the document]
eviltrout: I don't think we do!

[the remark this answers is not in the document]
featheredtoast: That's OK, it's super minor - we can enhance later, the rest lgtm!

majakomel:
Thanks for reviews! I'm filling in the gaps in specs and then it'll be good to go

jjaffeux:
Can’t wait to see all this hard work live!

PR #6039, opened 2018-06-26
FIX: scroll to top when accessing admin dashboard

OsamaSayegh:
https://meta.discourse.org/t/scrollbar-doesnt-go-to-the-top-when-opening-the-admin-panel/90119?u=osama


jjaffeux:
Thanks

PR #6042, opened 2018-06-26
Fix 'asscoiated' typo

hellekin:
I know that **Naming is CRITICAL** and that **Refactoring only NOT welcome**.

But since I spotted this (consistent) typo and the change does not affect any
functionality -- I checked the presence of "asscoiated" in the code base, I
guess the first rule trumps the second one.

It also gave me a false pretext to bypass my reluctance to use Google forms and
sign de CLA. Typos hurt the eye.


SamSaffron:
nice catch, thanks!

PR #6043, opened 2018-06-26
Include information about iframe state in message - fix typo

c-holtermann:
In approved pull request #5276 there was a typo. The html template header contained and attribute named data-embed-status. The javascript reads data-embed-state. This does not exist so return is always null. This fix sticks to calling this "state" instead of "status" and renames the attribute to "data-embed-state".


c-holtermann:
works nicely now. Thanks for the merge !

PR #6045, opened 2018-06-27
FEATURE: Add custom S3 Endpoint and DigitalOcean Spaces/Minio support for Backups

rishabhnambiar:
Adds two fields in site settings called `s3_endpoint` and `s3_force_path_style` which can be modified to backup to any S3 compatible service like DigitalOcean Spaces and Minio.

**Examples:**
`endpoint:` 'https://nyc3.digitaloceanspaces.com' for DigitalOcean Spaces
`endpoint:` 'https://s3.amazonaws.com' for AWS S3 (default)

- [x] Tested upload/delete backups and uploads with AWS S3
- [x] Tested upload/delete backups and uploads with DigitalOcean Spaces
- [x] Tested upload/delete backups and uploads with Minio
- [x] Tested on a new Discourse install
- [ ] Tested with Google Cloud Storage
- [ ] Write guides on Meta for setting up DigitalOcean Spaces/Minio


config/site_settings.yml
```diff
@@ -911,6 +911,9 @@ files:
   s3_upload_bucket:
     default: ''
     regex: '^[a-z0-9\-\/]+$' # can't use '.' when using HTTPS
+  s3_endpoint:
+    default: 'https://s3.amazonaws.com'
+    regex: 'http.:\/\/+\w+\.\w+\.com' # http(s)://xyz.abc.com
```
riking: regex is wrong - copy the one from `s3_cdn_url` below

rishabhnambiar: Fixed, sorry about that! :+1:

hosnas:
Is this only for backups? or also it would support [uploads](https://meta.discourse.org/t/extend-s3-configuration-for-other-s3-api-compatible-cloud-storage-solutions/20593?source_topic_id=22235)?

rishabhnambiar:
@hosnas Yes, this is only for backups, I'm still working on upload support, will add to this PR when complete.

app/models/site_setting.rb
```diff
@@ -133,15 +133,20 @@ def self.enable_s3_uploads
...
+        if SiteSetting.Upload.s3_region == "us-east-1"
+          "//#{bucket}.s3.amazonaws.com"
+        elsif SiteSetting.Upload.s3_region == 'cn-north-1'
+          "//#{bucket}.s3.cn-north-1.amazonaws.com.cn"
+        else
+          "//#{bucket}.s3-#{SiteSetting.Upload.s3_region}.amazonaws.com"
+        end
       else
-        "//#{bucket}.s3-#{SiteSetting.Upload.s3_region}.amazonaws.com"
+        "//#{bucket}.#{url_basename}"
```
arrowcircle: I think this else branch can be removed with 
```
return "//#{bucket}.#{url_basename}" if SiteSetting.s3_endpoint != "https://s3.amazonaws.com"
```

Also, conditions with endpoint can be extracted as methods. And check for cn can be extracted and generalized.

rishabhnambiar: Thanks a lot for the review, it's my first ever Ruby app so thanks for the help :grin: 
1) This is what you mean by extracting endpoint as a method?
```
    def self.absolute_base_url
      url_basename = s3_endpoint.split('/')[-1]
    ....
    ....

    def self.s3_endpoint
      SiteSetting.s3_endpoint
    end
```
2) I'm not sure if we have to keep the cn check or not so I'll leave it as it is for now.
3) Is there any advantage of removing the else block other than possibly readability?

arrowcircle: 1. No, I thought about extracting conditions with cn and other regions.
2. If it is here, I think it should be here. Maybe, it should be extended, but its another task.
3. Its decrease of complexity, not readability. Just google refactor if-else statement with return.

Also, according to link http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region,
there is no need for default URL for `US-EAST1`. It can be removed. And CN endpoints are generated with the same rules with the addition of `.cn` in the end.

So it can be refactored with a simple linear flow like
```
endpoint = "s3.#{SiteSetting.Upload.s3_region}.amazonaws.com"
endpoint << '.cn' if SiteSetting.Upload.s3_region =~ /^cn-*/i
```

rishabhnambiar: 1,3) I will do that, thanks! :+1: 

Yes, I wanted to change the if-else conditions myself because the conditions seemed unnecessary according to the docs but I didn't refactor it because the code was already written like this earlier. 
I'll change this.

arrowcircle: Thing is to have no if-else branches. Just linear ifs.
Also, (1) is not needed with the linear flow and remove of default us-east-1.

SamSaffron:
are we good to merge this? It seems done to me. force path style is a bit confusing but I guess there is not much we can do about that. 

Also we allow setting s3 via global settings as well and the 2 new options are missing there.

rishabhnambiar:
Sorry for the delay, by global settings do you mean I should simply add the new fields to ```discourse_defaults.conf```? I've added them now, please have a look.

Or do you mean adding definitions to ```models/global_setting.rb```?

Yes, we're good to merge. I agree that it's confusing, but keeping it enables admins to use endpoints with both, virtual-style and path-style addressing!

SamSaffron:
Close, have a look at how SiteSetting.Uploads works, it first tries to get stuff from Global followed by Site Settings. you need to follow the same pattern. Adding to discourse_defaults is a good first move.

rishabhnambiar:
I've now added the two fields to check for GlobalSettings before SiteSettings but I'm still not sure if something else needs to be added in `global_setting.rb` because other optional parameters like `s3_cdn_url` have also not been defined there.

Also, how do I test GlobalSettings? Changing values in `discourse_defaults.conf` doesn't seem to have any effect.

xfalcox:
Hey @rishabhnambiar did you got to test this using Google Cloud?

rishabhnambiar:
@xfalcox I did try, but it doesn't work with [GCS](https://cloud.google.com/storage/docs/migrating#migration-simple) yet.
I'm having some trouble getting the requests to work because GCS might need slightly different headers. Will let you once it works :+1:

SamSaffron:
This looks pretty safe to me, thanks @rishabhnambiar ! merging it in... cc @xfalcox

rishabhnambiar:
Thanks @SamSaffron, I'll send you a copy of that DigitalOcean Spaces Guide I was writing (similar to [this](https://meta.discourse.org/t/setting-up-file-and-image-uploads-to-s3/7229/)) before I post it on Meta!

PR #6047, opened 2018-06-27
DEV: lower highline requirement

jjaffeux:


SamSaffron:
ideally when we stick to old versions we need a comment in the gemfile about why

PR #6048, opened 2018-06-28
FEATURE: Add Question2Answer import script

realdlee:
This adds an import script for the Question2Answer php forum (https://www.question2answer.org/).

It imports users, categories, topics, posts, likes, and permalinks. It was roughly based on `discourse/script/import_scripts/vbulletin.rb`.


SamSaffron:
thanks heaps!

PR #6050, opened 2018-06-28
FIX: use `logo_url` settign when present for mobile layout instead of site name

OsamaSayegh:
https://meta.discourse.org/t/header-css-not-applied-to-user-api-access-key-confirm-page/90906?u=osama


SamSaffron:
thanks!

PR #6053, opened 2018-06-29
add swipe events for panning topic scroller on mobile out of the way

featheredtoast:
re-implemented with pointer events, and adds a mixin that allows a hammerjs-like interface (position, directions, deltas, and directional velocities).

Implemented mixin on mobile topic scroller, swipe to scroll away.

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-navigation.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-navigation.js.es6
@@ -1,9 +1,11 @@
 import { observes } from "ember-addons/ember-computed-decorators";
 import showModal from "discourse/lib/show-modal";
+import PanEvents from "discourse/mixins/pan-events";
 
-export default Ember.Component.extend({
+export default Ember.Component.extend(PanEvents, {
   composerOpen: null,
   info: null,
+  isPanning: false,
 
   init() {
     this._super();
@@ -91,7 +93,7 @@ export default Ember.Component.extend({
   _collapseFullscreen() {
     if (this.get("info.topicProgressExpanded")) {
       $(".timeline-fullscreen").removeClass("show");
-      setTimeout(() => {
+      Ember.run.later(() => {
         this.set("info.topicProgressExpanded", false);
         this._checkSize();
       }, 500);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/topic-navigation.js.es6
```diff
@@ -109,6 +111,55 @@ export default Ember.Component.extend({
...
+
+    $(".timeline-container").css("bottom", -offset);
+    if(offset > maxOffset) {
+      this._collapseFullscreen();
+    }
+    else if(offset <= 0) {
+      $(".timeline-container").css("bottom", "");
+    }
+    else {
+      setTimeout(() => this._panOpenClose(offset, velocity, direction), 20);
```
jjaffeux: Prefer `Ember.run.later` instead of setTimeout

app/assets/javascripts/discourse/components/topic-navigation.js.es6
```diff
@@ -109,6 +111,55 @@ export default Ember.Component.extend({
     }
   },
 
+  _panOpenClose(offset, velocity, direction) {
+
+    const maxOffset = parseInt($(".timeline-container").css("height"));
```
jjaffeux: given you fetch `$(".timeline-container")` multiple times here, I would store it in a variable.

```
$timelineContainer = $(".timeline-container");
```

app/assets/javascripts/discourse/components/topic-navigation.js.es6
```diff
@@ -109,6 +111,55 @@ export default Ember.Component.extend({
...
+      this.set("isPanning", false);
+    }
+  },
+
+  panEnd(e) {
+    if(!this.get("isPanning")) {
+      return;
+    }
+    this.set("isPanning", false);
+    //calculate whether we should keep open or close...
```
jjaffeux: if possible I would remove the comment by a private function:

```
_isPanOpened(e) {
  (e.deltaY > 200 && e.velocityY > -0.15) || e.velocityY > 0.15
}
```

and use this, it removes the need for a comment while staying explicit.

app/assets/javascripts/discourse/mixins/pan-events.js.es6
```diff
@@ -0,0 +1,89 @@
...
+
+    //calculate velocity from previous event center location
+    const eventDeltaX = e.clientX - oldState.center.x;
+    const eventDeltaY = e.clientY - oldState.center.y;
+    const velocityX = eventDeltaX / timeDiffSeconds;
+    const velocityY = eventDeltaY / timeDiffSeconds;
+    const deltaDistance = Math.sqrt(Math.pow(eventDeltaX, 2) + Math.pow(eventDeltaY, 2));
+    const velocity = deltaDistance / timeDiffSeconds;
+
+    return {
```
jjaffeux: In es6 you don’t need to repeat the value in objects if key has the text than value variable:

```
return {
   startLocation: oldState.startLocation,
   center: {x: Math.round(e.clientX), y: Math.round(e.clientY)},
   velocity,
   velocityX,
   velocityY,
   deltaX,
   deltaY,
   distance,
   start: false,
   timestamp: newTimestamp
}
```

app/assets/javascripts/discourse/components/topic-navigation.js.es6
```diff
@@ -109,6 +111,55 @@ export default Ember.Component.extend({
     }
   },
 
+  _panOpenClose(offset, velocity, direction) {
```
jjaffeux: another nitpick here, but not a huge fan of `_panOpenClose` name here. Could we have:

```
_panOpen(offset, velocity, direction) {
  this._panToggle(offset, velocity, "open");
},

_panClose(offset, velocity, direction) {
  this._panToggle(offset, velocity, "close");
},

_panToggle(offset, velocity, direction) {
}
```

PR #6054, opened 2018-06-29
WIP: Moderator activity reports

scossar:
This adds reports for moderator activity, recent flags, and staff notes.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -607,27 +607,35 @@ class Report
   def self.report_post_edits(report)
     report.data = []
 
-    revisions = PostRevision.joins("INNER JOIN posts ON posts.id = post_revisions.post_id")
-                  .where("post_revisions.user_id > 0")
-                  .where("post_revisions.created_at >= :start_date AND post_revisions.created_at <= :end_date",
-                         start_date: report.start_date, end_date: report.end_date)
-                  .select('post_revisions.user_id AS editor_id,
-                           posts.user_id AS author_id,
-                           post_revisions.number AS revision_version,
-                           posts.version AS post_version,
-                           posts.id AS post_id,
-                           posts.topic_id,
-                           posts.post_number,
-                           posts.edit_reason,
-                           post_revisions.created_at')
-                  .order('post_revisions.created_at DESC')
+    sql = <<~SQL
+    SELECT
+    pr.user_id AS editor_id,
+    p.user_id AS author_id,
+    pr.number AS revision_version,
+    p.version AS post_version,
+    pr.post_id,
+    p.topic_id,
+    p.post_number,
+    p.edit_reason,
+    u.username AS editor_username,
+    pr.created_at,
+    (SELECT u.username FROM users u WHERE u.id = p.user_id) AS author_username
+    FROM post_revisions pr
+    JOIN posts p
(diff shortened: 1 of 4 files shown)


app/models/report.rb
```diff
@@ -482,4 +482,128 @@ def self.report_trending_search(report)
       }
     end
   end
+
+  def self.report_moderator_activity(report)
+    report.data = []
+    mod_data = {}
+    mod_ids = []
+
+    User.real.where(moderator: true).select(:id, :username).as_json.each do |mod|
```
ZogStriP: Slightly more efficient 

```ruby
mod_data = {}

User.real.where(moderator: true).pluck(:id, :username).each do |u| 
  mod_data[u[0]] = { user_id: u[0], username: u[1] }
end

mod_ids = mod_data.keys
```

app/models/report.rb
```diff
@@ -482,4 +482,128 @@ def self.report_trending_search(report)
...
+      mod_data[mod['id']] = {user_id: mod['id'], username: mod['username']}
+      mod_ids << mod['id']
+    end
+
+    return if mod_ids.empty?
+
+    time_read_query = <<~SQL
+    SELECT SUM(uv.time_read) AS time_read,
+    uv.user_id
+    FROM user_visits uv
```
ZogStriP: There's no need to use an alias when you're selecting from only 1 table.

app/models/report.rb
```diff
@@ -482,4 +482,128 @@ def self.report_trending_search(report)
...
+
+    DB.query(topic_count_query).each do |row|
+      mod_data[row.user_id][:topic_count] = row.topic_count
+    end
+
+    DB.query(post_count_query).each do |row|
+      mod_data[row.user_id][:post_count] = row.post_count
+    end
+
+    mod_data.each do |k, v|
```
ZogStriP: Can also be written as

```ruby
report.data = mod_data.values
```

app/models/report.rb
```diff
@@ -482,4 +482,128 @@ def self.report_trending_search(report)
...
+        flag[:response_time] = action.agreed_at - action.created_at
+      elsif action.disagreed_at
+        flag[:response_time] = action.disagreed_at - action.created_at
+      elsif action.deferred_at
+        flag[:response_time] = action.deferred_at - action.created_at
+      else
+        flag[:response_time] = nil
+      end
+      flag[:poster_id] = action.poster_id
+      flag[:poster_username] = User.find(action.poster_id).username
```
ZogStriP: This will generate N+1 requests. Which will become slower as the number of flag actions grows. To prevent this, you need to move that query in your first query ;)

scossar: Thanks! I assumed that was wrong. Will fix.

app/models/report.rb
marmeladze: I currently do not know the whole context and it's just a quick lookup - but wouldn't it be more efficient to replace `each` with `find_each`?

SamSaffron:
@jjaffeux should you be reviewing this in the context of your current changes?

jjaffeux:
@SamSaffron yes I was integrating Simon changes as he was working on this, I told him to create its own pull request so he wouldn’t be bothered with merges conflicts and I would handle this

PR #6055, opened 2018-06-29
PERF: Avoid front loading all gaps.

tgxworld:
On a mega topic with 70K posts, the following query to load all the gaps upfront is very expensive.

screenshot from 2018-06-29 16-36-59

Instead of plucking all the ids, I replaced it with a solution that uses 4 queries and only load the gaps as the client request for each chunk

screenshot from 2018-06-29 16-38-39


tgxworld:
@eviltrout Thank you for reviewing! Will add more comments to explain what the method does. I noticed a client side bug so I'll need to fix that next week before merging.

tgxworld:
Closing this for now. We decided that this complicates the code too much for the benefit that it brings.

PR #6057, opened 2018-06-29
FEATURE: Event on topic merge

jose-hms:
Added a DiscourseEvent and test case for when a topic is merged.

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -911,6 +911,7 @@ class Topic < ActiveRecord::Base
 
     if opts[:destination_topic_id]
       post_mover.to_topic opts[:destination_topic_id]
+      DiscourseEvent.trigger(:topic_merged,  post_mover.original_topic, post_mover.destination_topic)
     elsif opts[:title]
       post_mover.to_new_topic(opts[:title], opts[:category_id])
     end
--- a/spec/requests/topics_controller_spec.rb
+++ b/spec/requests/topics_controller_spec.rb
@@ -198,6 +198,13 @@ RSpec.describe TopicsController do
           expect(result['success']).to eq(true)
           expect(result['url']).to be_present
         end
+        
+        it "triggers an event on merge" do
+          DiscourseEvent.on(:topic_merged) do |original_topic, destination_topic|
+            expect(original_topic.id).to eq(topic.id)
+            expect(destination_topic.id).to eq(dest_topic.id)
+          end
+        end
       end
 
       context 'failure' do


spec/requests/topics_controller_spec.rb
```diff
@@ -257,6 +257,24 @@
           expect(result['url']).to be_present
         end
       end
+
+      context "events" do
+        before do 
+          @topics = 0
+          @update_topic_id = -> (topic, destination) {@topics = topic.id, @topics = destination.id}
+          DiscourseEvent.on(:topic_merged, &@update_topic_id)
```
tgxworld: You can just do the following here

```
DiscourseEvent.on(:topic_merged) do |topic, dest_topic|
  expect(topic)...
  epxect(dest_topic)...
end
```

app/controllers/topics_controller.rb
```diff
@@ -543,6 +543,7 @@ def merge_topic
     guardian.ensure_can_move_posts!(topic)
 
     dest_topic = topic.move_posts(current_user, topic.posts.pluck(:id), destination_topic_id: params[:destination_topic_id].to_i)
+    DiscourseEvent.trigger(:topic_merged, topic, dest_topic)
```
SamSaffron: I support this new event but it should not be triggered in the controller, it should be triggered in topic.move_posts

Also no need to sign in and test the controller then cause this will always happen. 

Closing this for now, feel free to re-submit

SamSaffron:
OK post internal discussion we can keep this open for a bit longer while you have a chance to re-write it so context is not lost

jose-hms:
@SamSaffron 

Thanks, I should be able to re-write later today.

jose-hms:
@SamSaffron or @tgxworld 

Should be good. Re-wrote per your requests.

app/models/topic.rb
```diff
@@ -910,6 +910,7 @@ def move_posts(moved_by, post_ids, opts)
     post_mover = PostMover.new(self, moved_by, post_ids)
 
     if opts[:destination_topic_id]
+      DiscourseEvent.trigger(:topic_merged,  post_mover.original_topic, Topic.find_by(id: opts[:destination_topic_id]))
```
tgxworld: > Topic.find_by(id: opts[:destination_topic_id])

This adds an extra database query. You can trigger this event only after `post_mover.to_topic opts[:destination_topic_id]` and then call `post_mover.destination_topic`

tgxworld:
Awesome Thank you :+1:

PR #6060, opened 2018-07-02
show how many backup codes user has left

majakomel:


app/models/concerns/second_factor_manager.rb
```diff
@@ -45,6 +45,10 @@ def backup_codes_enabled?
       SiteSetting.enable_local_logins
   end
 
+  def remaining_backup_codes
+    self&.user_second_factors&.backup_codes&.length
```
tgxworld: We usually want to do a `.count` instead of `.length` here. `.length` loads all the associations into an array and counts the length of the array which is memory inefficient. `.count` on the other hands converts the SQL query into a count query and returns an integer.

jjaffeux:
@majakomel when you can fix what tgx said and improve the commit title I will merge this 👍 

As title I suggest something like:

FEATURE: shows remaining backup codes in user preferences

app/assets/stylesheets/common/base/user.scss
```diff
@@ -557,16 +557,6 @@
   }
 
   &.second-factor-backup-preferences {
-    padding-left: 0;
```
jjaffeux: I'm unclear if it's part of this PR or something you fixed while doing it? if it's not exacty part of the PR, you can leave it or make another commit if you have time, it would be a little bit more cleaner.

app/assets/javascripts/discourse/controllers/preferences/second-factor-backup.js.es6
```diff
@@ -88,7 +88,9 @@ export default Ember.Controller.extend({
             errorMessage: null,
             backupCodes: response.backup_codes
           });
-          model.set("second_factor_backup_enabled", true);
+
+          this.set("backupEnabled", true);
+          this.set("remainingCodes", 10);
```
tgxworld: We should probably do a length count on `response.backup_codes` instead of hard coding the value here. Also, you can do the following when setting multiple properties at once 

```
this.setProperties({
  backupEnabled: true,
  remainingCodes: response.backup_codes.length
})
```

majakomel: Ah true, thank you!

package.json
```diff
@@ -11,7 +11,7 @@
     "chrome-launcher": "^0.10.2",
     "chrome-remote-interface": "^0.25.6",
     "eslint": "^4.19.1",
-    "prettier": "^1.13.0",
```
jjaffeux: This PR shouldn’t be changing package.json can you remove this file plz ?

app/serializers/user_serializer.rb
```diff
@@ -160,6 +161,14 @@ def second_factor_backup_enabled
     object.backup_codes_enabled?
   end
 
+  def include_second_factor_remaining_codes?
+    (object&.id == scope.user&.id) && object.backup_codes_enabled?
+  end
+
+  def second_factor_remaining_codes
```
jjaffeux: Wondering if this shouldn’t be `second_factor_remaining_backup_codes` what do you think?

majakomel: Agree, for the sake of consistency. Will add a change.

PR #6063, opened 2018-07-03
Add option to skip push for certain notifications

gdpelican:
I'm using the post alerter to notify of events from within a plugin, but want to avoid sending push notifications in this way (I want to send a different payload, and can happily do so manually).

In a similar manner to the 'skip email' setting, I'd like to be able to control whether or not a particular notification triggers a push or not.

### DIFF
--- a/app/services/post_alerter.rb
+++ b/app/services/post_alerter.rb
@@ -393,7 +393,7 @@ class PostAlerter
       skip_send_email: skip_send_email
     )
 
-    if created.id && !existing_notification && NOTIFIABLE_TYPES.include?(type) && !user.suspended?
+    if created.id && !existing_notification && NOTIFIABLE_TYPES.include?(type) && !user.suspended? && !opts[:skip_push]
       post_url = original_post.url
       if post_url
         payload = {
--- a/spec/services/post_alerter_spec.rb
+++ b/spec/services/post_alerter_spec.rb
@@ -28,9 +28,9 @@ describe PostAlerter do
   let!(:evil_trout) { Fabricate(:evil_trout) }
   let(:user) { Fabricate(:user) }
 
-  def create_post_with_alerts(args = {})
+  def create_post_with_alerts(args = {}, opts = {})
     post = Fabricate(:post, args)
-    PostAlerter.post_created(post)
+    PostAlerter.post_created(post, opts)
   end
 
   context "private message" do
@@ -635,12 +635,11 @@ describe PostAlerter do
 
   describe "push_notification" do
     let(:mention_post) { create_post_with_alerts(user: user, raw: 'Hello @eviltrout :heart:') }
+    let(:mention_post_skip_push) { create_post_with_alerts({user: user, raw: 'Hello @eviltrout :heart'}, {skip_push: true}) }
     let(:topic) { mention_post.topic }
 
-    it "pushes nothing to suspended users" do
+    before do
       SiteSetting.allowed_user_api_push_urls = "https://site.com/push|https://site2.com/push"
-
-      evil_trout.update_columns(suspended_till: 1.year.from_now)
 
       2.times do |i|
         UserApiKey.create!(user_id: evil_trout.id,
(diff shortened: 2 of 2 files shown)


ZogStriP:
Thanks but can you add a test as well to ensure we don't regress in the future?

gdpelican:
Ok, done

app/services/post_alerter.rb
```diff
@@ -393,7 +393,7 @@ def create_notification(user, type, post, opts = {})
       skip_send_email: skip_send_email
     )
 
-    if created.id && !existing_notification && NOTIFIABLE_TYPES.include?(type) && !user.suspended?
+    if created.id && !existing_notification && NOTIFIABLE_TYPES.include?(type) && !user.suspended? && !opts[:skip_push]
```
tgxworld: If we only want to `skip_push`, it should only be applied to `push_notification(user, payload)`. This is currently disabling live updates to the app and the event trigger that we fire.

gdpelican: That is the intended behaviour; I don't want the notifications to be serialized to the client or trigger any other side effects like sending push notifications or any other hooks baked into `post_notification_alert`. Maybe there's a better name?

tgxworld: Can you provide us with a sample of how your plugin is using the post alerter? I'm more inclined in letter plugins customize the payload than having an option to disable "live" notification.

gdpelican: Sure.

I want to create notifications in the database for people mentioned in a chat post. So, when someone types `Hello @tgxworld!`, a notification record is stored in the database, with your user ID, the person who mentioned you, and the post / topic that it occurred in. This is the set of info that I want, and makes it really easy to query for, for example, the number of unread mentions you have in your chats.

```
Notification.where(topic_id: <your chat topic ids>, read: false).count
```

I'd like to do this in the same way, that Discourse does for regular posts, which means calling `PostAlerter.after_save_post` on the chat post. This allows me to re-use the logic of pulling out username mentions, expanding group mentions, and all the other good stuff that happens when someone makes a post on Discourse.

However, I don't want that notification to appear in your notifications tray, because the notifications tray already has enough different kinds of notifications (normal, PM, and flags), the default behavior (linking to the post directly) is sensible but not what I want here and is something that's difficult to override both on the client and on the server.

So, essentially I'm looking for a way to call this code (post_alerter#386):
```
    # Create the notification
    created = user.notifications.create!(
      notification_type: type,
      topic_id: post.topic_id,
      post_number: post.post_number,
      post_action_id: opts[:post_action_id],
      data: notification_data.to_json,
      skip_send_email: skip_send_email
    )
```

without triggering side effects. However, it's in the heart of the long `create_notification` method, and there are ~70 lines of code which go into producing that `notification_data.to_json` bit that I'd like to avoid duplicating in my plugin.

gdpelican: How can I help to resolve this? @tgxworld

gdpelican: ?? @tgxworld @SamSaffron

tgxworld: I think what you need here is a custom notification type and not an option to skip creating a notification. From what you're describing, a "mention" in your chat plugin is not the same as a "mention" in the Discourse code base. As long as the type is not part of `PostAlerter::NOTIFIABLE_TYPES`, the server side wouldn't publish anything to the client.

spec/services/post_alerter_spec.rb
```diff
@@ -650,22 +649,20 @@ def set_topic_notification_level(user, topic, level_name)
...
+
+    it "pushes nothing to suspended users" do
+      evil_trout.update_columns(suspended_till: 1.year.from_now)
 
       expect { mention_post }.to_not change { Jobs::PushNotification.jobs.count }
     end
 
     it "correctly pushes notifications if configured correctly" do
       SiteSetting.queue_jobs = false
-      SiteSetting.allowed_user_api_push_urls = "https://site.com/push|https://site2.com/push"
```
tgxworld: Not sure if this is intended or a side effect.

SamSaffron:
I am just so mixed on this, until I see a need for this in core I don't want to add this knob. 

My call here is that you should possibly just have your own custom post alerter here for this behavior. I am not against factoring some of post alerter so you can consume bits of it, but don't want to add this to the public api for now.

PR #6066, opened 2018-07-04
Secondary emails next phase

LeoMcA:
Hope I'm not stomping all over proper PR etiquette by including a number of commits in this.

These should bring functionality in line with my post: https://meta.discourse.org/t/additional-email-address-per-user-account-support/59847/13?u=leomca

### DIFF
--- a/spec/components/email/receiver_spec.rb
+++ b/spec/components/email/receiver_spec.rb
@@ -169,6 +169,13 @@ describe Email::Receiver do
       expect { process(:from_reply_by_email_address) }.to raise_error(Email::Receiver::FromReplyByAddressError)
     end
 
+    it "accepts reply from secondary email address" do
+      Fabricate(:secondary_email, email: "someone_else@bar.com", user: user)
+      expect { process(:reply_user_not_matching) }.to change { topic.posts.count }
+      expect(Post.last.raw).to eq "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
+      expect(Post.last.user).to eq user
+    end
+
     it "raises a TopicNotFoundError when the topic was deleted" do
       topic.update_columns(deleted_at: 1.day.ago)
       expect { process(:reply_user_matching) }.to raise_error(Email::Receiver::TopicNotFoundError)
@@ -490,6 +497,18 @@ describe Email::Receiver do
       expect(topic.topic_users.count).to eq(3)
     end
 
+    it "invites users with a secondary email in the chain" do
+      user1 = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "discourse@bar.com", user: user1)
+      Fabricate(:secondary_email, email: "someone@else.com", user: user1)
+      user2 = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "team@bar.com", user: user2)
+      Fabricate(:secondary_email, email: "wat@bar.com", user: user2)
+
+      expect { process(:cc) }.to change(Topic, :count)
+      expect(Topic.last.allowed_users).to contain_exactly(user1, user2)
+    end
+
     it "cap the number of staged users created per email" do
       SiteSetting.maximum_staged_users_per_email = 1
       expect { process(:cc) }.to change(Topic, :count)
@@ -650,6 +669,15 @@ describe Email::Receiver do
     it "ignores by case-insensitive title" do
       SiteSetting.ignore_by_title = "foo"
       expect { process(:ignored) }.to_not change(Topic, :count)
+    end
+
+    it "associates email from a secondary address with user" do
+      user = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "existing@bar.com", user: user)
+
+      expect { process(:existing_user) }.to change(Topic, :count)
+      expect(Topic.last.posts.last.raw).to eq "Hey, this is a topic from an existing user ;)"
+      expect(Topic.last.user).to eq user
     end
   end
 
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1758,4 +1758,18 @@ describe User do
       filter_by(:filter_by_username_or_email)
     end
   end
+
+  describe "#secondary_emails" do
+    let!(:user) { Fabricate(:user_single_email) }
+
+    it "is empty when user has one email" do
+      expect(user.secondary_emails).to be_empty
+    end
+
+    it "only contains secondary emails" do
+      secondary_email = Fabricate(:secondary_email, user: user)
+      expect(user.secondary_emails).to contain_exactly(secondary_email.email)
+    end
+  end
+
 end
--- a/spec/serializers/admin_user_list_serializer_spec.rb
+++ b/spec/serializers/admin_user_list_serializer_spec.rb
@@ -1,0 +1,76 @@
+require 'rails_helper'
+require_dependency 'user'
+
+describe AdminUserListSerializer do
+
+  context "emails" do
+    let(:admin) { Fabricate(:user_single_email, admin: true, email: "admin@email.com") }
+    let(:user) { Fabricate(:user_single_email, email: "user@email.com") }
+
+    def fabricate_secondary_emails_for(u)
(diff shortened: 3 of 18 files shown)


LeoMcA:
Oops, forgot to run the linter, and fixed up temperamental specs

spec/components/email/receiver_spec.rb
```diff
@@ -490,6 +496,19 @@ def process(email_name)
...
+      user1 = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "discourse@bar.com", user: user1)
+      Fabricate(:secondary_email, email: "someone@else.com", user: user1)
+      user2 = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "team@bar.com", user: user2)
+      Fabricate(:secondary_email, email: "wat@bar.com", user: user2)
+
+      expect { process(:cc) }.to change(Topic, :count)
+      user_ids = Topic.last.allowed_users.pluck(:id)
+      expect(user_ids).to contain_exactly(user1.id, user2.id)
```
tgxworld: No need to pluck for the ids here since you can just do `expect(Topic.last.allowed_users).to contain_exactly(user1, user2)`

spec/components/email/receiver_spec.rb
```diff
@@ -169,6 +169,12 @@ def process(email_name)
       expect { process(:from_reply_by_email_address) }.to raise_error(Email::Receiver::FromReplyByAddressError)
     end
 
+    it "accepts reply from secondary email address" do
+      Fabricate(:secondary_email, email: "someone_else@bar.com", user: user)
+      expect { process(:reply_user_not_matching) }.to change { topic.posts.count }
+      expect(Post.last.user.id).to eq user.id
```
tgxworld: `expect(Post.last.user).to eq(user)` should be sufficient here.

tgxworld: We should check for the content of the post here as well.

spec/components/email/receiver_spec.rb
```diff
@@ -651,6 +670,14 @@ def process(email_name)
       SiteSetting.ignore_by_title = "foo"
       expect { process(:ignored) }.to_not change(Topic, :count)
     end
+
+    it "associates email from a secondary address with user" do
+      user = Fabricate(:user, trust_level: SiteSetting.email_in_min_trust)
+      Fabricate(:secondary_email, email: "existing@bar.com", user: user)
+
+      expect { process(:existing_user) }.to change(Topic, :count)
+      expect(Topic.last.user.id).to eq user.id
```
tgxworld: We need to test for the content of the topic here as well. Let's rewrite the assertion to `expect(Topic.last.user).to eq(user)`.

spec/models/user_spec.rb
```diff
@@ -1758,4 +1758,18 @@ def filter_by(method)
...
   end
+
+  describe "#secondary_emails" do
+    it "is empty when user has one email" do
+      user = Fabricate(:user_single_email)
+      expect(user.secondary_emails).to be_empty
+    end
+
+    it "only contains secondary emails" do
+      user = Fabricate(:user_single_email)
```
tgxworld: We can dry this code by using `let!(:user) { Fabricate(:user_single_email) }`

spec/serializers/admin_user_list_serializer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+require_dependency 'user'
+
+describe AdminUserListSerializer do
+
+  context "emails" do
+    let(:admin) { Fabricate(:user_single_email, admin: true) }
+    let(:user) { Fabricate(:user_single_email) }
+
+    def fabricate_secondary_emails_for(u)
+      Fabricate(:secondary_email, user: u, email: "first@email.com")
```
tgxworld: Let's use a loop here to avoid duplication

spec/serializers/admin_user_list_serializer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      Fabricate(:secondary_email, user: u, email: "second@email.com")
+    end
+
+    shared_examples "shown" do
+      it "contains emails" do
+        expect(json[:email]).to_not be_nil
```
tgxworld: We should test for the exact email here.

spec/serializers/admin_user_list_serializer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+          "first@email.com",
+          "second@email.com"
+        )
+      end
+    end
+
+    shared_examples "not shown" do
+      it "doesn't contain emails" do
+        expect(json[:email]).to be_nil
+        secondary_emails = json[:secondary_emails]
```
tgxworld: Hmm we can probably remove the variable here.

test/javascripts/acceptance/admin-user-emails-test.js.es6
```diff
@@ -0,0 +1,129 @@
+import { acceptance } from "helpers/qunit-helpers";
+
+acceptance("Admin - User Emails", { loggedIn: true });
+
+const response_with_secondary = secondary_emails => {
```
tgxworld: For javascript, we prefer to use camelCase.

test/javascripts/acceptance/admin-user-emails-test.js.es6
```diff
@@ -0,0 +1,129 @@
...
+      "displays secondary email"
+    );
+  });
+});
+
+const assert_multiple_secondary = assert => {
+  assert.equal(
+    find(".display-row.email .value a").text(),
+    "eviltrout@example.com",
+    "displays primary email"
```
tgxworld: I don't think it is necessary to exert for the primary email here.

test/javascripts/acceptance/admin-user-emails-test.js.es6
```diff
@@ -0,0 +1,129 @@
...
+
+QUnit.test("has no secondary emails - clicking primary button", assert => {
+  no_secondary(assert, "email");
+});
+
+QUnit.test("has no secondary emails - clicking secondary button", assert => {
+  no_secondary(assert, "secondary-emails");
+});
+
+const multiple_secondary = (assert, button) => {
```
tgxworld: Is this test required since we already have `assert_multiple_secondary` ?

LeoMcA: Yes, without it we'd only be testing that secondary emails are rendered correctly when an admin is viewing their own profile. We wouldn't be testing to see that secondary emails are rendered correctly when an admin is viewing another profile.

test/javascripts/acceptance/admin-user-emails-test.js.es6
LeoMcA: That's fair, the number of secondary emails shouldn't affect the rendering of the primary email - which we've already tested for.

LeoMcA:
@tgxworld updated based on your review

tgxworld:
Merged in https://github.com/discourse/discourse/commit/21ebb1cd54dd19767a2d821d4c92f24f0c72a12c. 

Thank you @LeoMcA. Note that I removed the plugin outlet because I think that the code to display secondary emails in the user's preferences should remain in core.

LeoMcA:
Wooohoo, and thanks for the comments on the commit @tgxworld, I learned a couple of new things!

> Note that I removed the plugin outlet because I think that the code to display secondary emails in the user's preferences should remain in core.

Absolutely, I agree that there should be code in core displaying a user's secondary emails, and allowing them to edit them. But I *also* think that plugins should be able to override it, as this is an area where it's possible (and in Mozilla's case, certain) an authentication plugin may want to completely take control over how a user adds/removes these addition email addresses:

Mozilla-specific email preferences UI

Because of this, would you consider merging https://github.com/discourse/discourse/pull/6066/commits/d9681cad51fecdb78b0294fe6e73135c16d06962 too? The serializer portion can be taken out (and done in plugin) but I thought might as well be included, because it'll be needed anyway to display the secondary emails in the user's preferences in core.

I was working under the assumption that it's alright that I've not built the user preferences side of things, because I'll just be overriding it for Mozilla. But if that's not the case, and you  want me to build the core UI as well, just let me know :)

jjaarrvviiss:
How is this coming along?

PR #6069, opened 2018-07-05
FIX: Rename `User#usernames` that clashes with `Group#name`.

tgxworld:


app/jobs/scheduled/fix_user_usernames_and_groups_names_clash.rb
```diff
@@ -0,0 +1,29 @@
+module Jobs
+  class FixUserUsernamesAndGroupNamesClash < Jobs::Scheduled
```
ZogStriP: Why is this a recurring job and not a one-time (deferred) migration?

tgxworld: We can't add any database index to ensure uniqueness at the database level. Therefore, there is still a chance a clash to happen.

PR #6071, opened 2018-07-05
SECURITY: Do not allow authentication with disabled plugin-supplied a…

davidtaylorhq:
…uth providers

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -99,6 +99,7 @@ class Users::OmniauthCallbacksController < ApplicationController
     end
 
     Discourse.auth_providers.each do |provider|
+      raise Discourse::InvalidAccess.new("provider is not enabled") unless provider.enabled_setting.nil? || SiteSetting.send(provider.enabled_setting)
       return provider.authenticator if provider.name == name
     end
 
--- a/spec/requests/omniauth_callbacks_controller_spec.rb
+++ b/spec/requests/omniauth_callbacks_controller_spec.rb
@@ -33,6 +33,39 @@ RSpec.describe Users::OmniauthCallbacksController do
       expect(Users::OmniauthCallbacksController.find_authenticator("twitter"))
         .not_to eq(nil)
     end
+
+    context "with a plugin-contributed auth provider" do
+      before do
+        @provider = Plugin::AuthProvider.new
+
+        @provider.authenticator = Auth::OpenIdAuthenticator.new('ubuntu', 'https://login.ubuntu.com', trusted: true)
+        @provider.enabled_setting = "ubuntu_login_enabled"
+
+        Discourse.stubs(:auth_providers).returns [ @provider ]
+      end
+
+      it "finds an authenticator when enabled" do
+        SiteSetting.stubs(:ubuntu_login_enabled).returns(true)
+
+        expect(Users::OmniauthCallbacksController.find_authenticator("ubuntu"))
+          .not_to eq(nil)
+      end
+
+      it "fails if an authenticator is disabled" do
+        SiteSetting.stubs(:ubuntu_login_enabled).returns(false)
+
+        expect { Users::OmniauthCallbacksController.find_authenticator("ubuntu") }
+          .to raise_error(Discourse::InvalidAccess)
+      end
+
+      it "succeeds if an authenticator does not have a site setting" do
+        @provider.enabled_setting = nil
+        SiteSetting.stubs(:ubuntu_login_enabled).returns(false)
+
+        expect(Users::OmniauthCallbacksController.find_authenticator("ubuntu"))
+          .not_to eq(nil)
+      end
+    end
   end
 
   context 'Google Oauth2' do


spec/requests/omniauth_callbacks_controller_spec.rb
```diff
@@ -33,6 +33,39 @@
       expect(Users::OmniauthCallbacksController.find_authenticator("twitter"))
         .not_to eq(nil)
     end
+
+    context "with a plugin-contributed auth provider" do
+      before do
+        @provider = Plugin::AuthProvider.new
```
eviltrout: we prefer the `let` syntax to using instance variables in tests.

spec/requests/omniauth_callbacks_controller_spec.rb
```diff
@@ -33,6 +33,39 @@
...
+
+        @provider.authenticator = Auth::OpenIdAuthenticator.new('ubuntu', 'https://login.ubuntu.com', trusted: true)
+        @provider.enabled_setting = "ubuntu_login_enabled"
+
+        Discourse.stubs(:auth_providers).returns [ @provider ]
+      end
+
+      it "finds an authenticator when enabled" do
+        SiteSetting.stubs(:ubuntu_login_enabled).returns(true)
```
eviltrout: Instead of stubbing SiteSetting, you can say `SiteSetting.ubuntu_login_enabled = true`

spec/requests/omniauth_callbacks_controller_spec.rb
```diff
@@ -33,6 +33,39 @@
...
     end
+
+    context "with a plugin-contributed auth provider" do
+      before do
+        @provider = Plugin::AuthProvider.new
+
+        @provider.authenticator = Auth::OpenIdAuthenticator.new('ubuntu', 'https://login.ubuntu.com', trusted: true)
+        @provider.enabled_setting = "ubuntu_login_enabled"
+
+        Discourse.stubs(:auth_providers).returns [ @provider ]
```
eviltrout: I'm not sure if we can test this without stubbing, but we always prefer to do that if possible. It would be worth spending a few minutes looking at how hard that would be for your test.

davidtaylorhq: I think it is quite difficult without stubbing, but I may have missed something. 

`Discourse.auth_providers` populates from `Discourse.plugins`. I think the only way to populate `Discourse.plugins` is to run `Discourse.activate_plugins!`, which only pulls in plugins from the filesystem at `/plugins`.

Even the spec for `/lib/plugins/instance.rb` doesn't seem to populate `Discourse.plugins` so I think it would need some architectural changes to make this easier to test without stubbing.

spec/requests/omniauth_callbacks_controller_spec.rb
davidtaylorhq: `ubuntu_login_enabled` isn't a real Site Setting (hypothetically it is one that would be added by a plugin)

I can programmatically define a new SiteSetting at the beginning of the test by doing
```
SiteSetting.setting('ubuntu_login_enabled', true)
```
but I'm unsure how to properly clean that up at the end of the test.

spec/requests/omniauth_callbacks_controller_spec.rb
davidtaylorhq: Done 👍

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -99,6 +99,7 @@ def self.find_authenticator(name)
     end
 
     Discourse.auth_providers.each do |provider|
+      raise Discourse::InvalidAccess.new("provider is not enabled") unless provider.enabled_setting.nil? || SiteSetting.send(provider.enabled_setting)
```
tgxworld: We should probably translate the string here.

tgxworld: A nitpick here is that if the line is more than 80 characters long, we prefer to write the `if/unless` statement on multiple lines.

SamSaffron:
as a general rule all these mocks and stubs are not great, I feel it just makes the test suite far more fragile long term. I would just test with concrete core auth providers and amend site settings. 

Once that is done feel free to merge and be sure to also cherry pick into stable and beta.

davidtaylorhq:
Thanks @tgxworld I've made those style changes.

> I would just test with concrete core auth providers and amend site settings

The code path for core auth providers and plugin auth providers is different, so I think we do need to test plugin-contributed things here. Core auth providers are already tested, and are not affected by this issue.

For the site settings I can "cheat" and just piggyback on an existing site setting, but I feel like that's not really the right answer. As for actually "installing" a plugin's auth provider, I can't find any way to do it programmatically without leaving mess behind which could affect the rest of the test suite.

Essentially all I want to do is add something the list returned by `Discourse.auth_providers`:

https://github.com/discourse/discourse/blob/1119afa18a0276e9b78de4d5b57a0f2593c4ad73/lib/discourse.rb#L211-L220

Does anyone have any suggestions on how I can do this without stubs?

PR #6072, opened 2018-07-05
FEATURE: allow selecting a tag when moving posts to a new topic if tags are enabled

majakomel:


spec/models/post_mover_spec.rb
```diff
@@ -19,7 +19,8 @@
   end
 
   context 'move_posts' do
-    let(:user) { Fabricate(:user) }
+    SiteSetting.tagging_enabled = true
```
ZogStriP: So, this line is the reason why the `web_hook_topic_view_serializer` spec fails. It's leaking state and enables tagging for the whole suite which adds the `tags` property ot the topic (thus breaking the former spec).

To solve the problem, that line should be move inside a `before` block **and** you should reset your test database (`RAILS_ENV=test rake db:drop db:create db:migrate`).

Why?

When the specs are launched, that line is executed, thus enabling tagging.
Then, when we're [reseting the default site settings](https://github.com/discourse/discourse/blob/211981ef237e797085e7c04975c614c55dcda9b0/spec/rails_helper.rb#L103-L106) we are using the current value stored in the database as the default value.
Thus enabling tagging by default.
Now, when the `web_hook_topic_view_serializer` spec is run, the `TopicViewSerializer` (via the `TopicTagsMixin`) will add the `tags` property to the topic, thus adding 1 more property and breaking that spec.

ZogStriP:
👍

PR #6073, opened 2018-07-05
Phase 0 for user-selectable theme components

OsamaSayegh:


db/migrate/20180704201756_drop_key_from_themes.rb
```diff
@@ -0,0 +1,15 @@
+class DropKeyFromThemes < ActiveRecord::Migration[5.2]
```
SamSaffron: Oh this is not how we do it ... see: https://github.com/discourse/discourse/blob/master/db/fixtures/000_delayed_drops.rb

lib/middleware/anonymous_cache.rb
```diff
@@ -65,13 +65,14 @@ def is_crawler?
...
 
-      def theme_key
-        key, _ = @request.cookies['theme_key']&.split(',')
-        if key && Guardian.new.allow_theme?(key)
-          key
+      def theme_id
+        ids, _ = @request.cookies['theme_ids']&.split('|')
+        id = ids&.split(",")&.map(&:to_i)&.first
+        if id && Guardian.new.allow_theme?(id)
+          id
```
SamSaffron: we should prep guardian here to call allow theme on every id, not only the first.

db/migrate/20180704201722_add_theme_id_to_user_option.rb
```diff
@@ -0,0 +1,45 @@
...
+         FROM themes
+         INNER JOIN user_options
+         ON themes.key = user_options.theme_key
+         WHERE user_options.user_id = uo.user_id
+       ) WHERE uo.theme_key IN (SELECT key FROM themes)"
+    )
+
+    begin
+      Migration::SafeMigrate.disable!
+      remove_column :user_options, :theme_key
```
SamSaffron: another one, this needs to be delay dropped

db/migrate/20180704201722_add_theme_id_to_user_option.rb
```diff
@@ -0,0 +1,45 @@
...
+    ensure
+      Migration::SafeMigrate.enable!
+    end
+  end
+
+  def down
+    add_column :user_options, :theme_key, :string
```
SamSaffron: less concerned about down migrations, you can just throw an irreversible migration exception here

db/migrate/20180704201742_drop_default_theme_key_from_site_setting.rb
```diff
@@ -0,0 +1,20 @@
+class DropDefaultThemeKeyFromSiteSetting < ActiveRecord::Migration[5.2]
+  def up
+    # preserve the current default theme by copying its ID to the new site setting...
```
SamSaffron: lets get all the work we need done, done in one migration no need for a ton here, very small amounts of data is changing.

app/models/theme.rb
```diff
@@ -99,28 +94,28 @@ def self.expire_site_cache!
   end
 
   def self.clear_default!
-    SiteSetting.default_theme_key = ""
+    SiteSetting.default_theme_id = -1
```
SamSaffron: hmmm is the default 0? cause later on you are checking for blank?

app/services/user_updater.rb
```diff
@@ -85,9 +85,15 @@ def update(attributes = {})
 
     save_options = false
 
-    # special handling for theme_key cause we need to bump a sequence number
-    if attributes.key?(:theme_key) && user.user_option.theme_key != attributes[:theme_key]
-      user.user_option.theme_key_seq += 1
+    # special handling for theme_id cause we need to bump a sequence number
+    if attributes.key?(:theme_ids)
+      user_guardian = Guardian.new(user)
```
SamSaffron: wait why do we need another guardian here if the code above already defines one?

db/migrate/20180706054922_drop_key_column_from_themes.rb
```diff
@@ -0,0 +1,32 @@
+class DropKeyColumnFromThemes < ActiveRecord::Migration[5.2]
+  def up
```
SamSaffron: migration looks good.

lib/guardian.rb
```diff
@@ -358,12 +358,10 @@ def can_export_entity?(entity)
...
-  def allow_theme?(theme_key)
-    if is_staff?
-      Theme.theme_keys.include?(theme_key)
-    else
-      Theme.user_theme_keys.include?(theme_key)
-    end
+  def allow_themes?(theme_ids)
+    theme_ids = [theme_ids] unless theme_ids.is_a?(Array)
+    allowed_ids = is_staff? ? Theme.theme_ids : Theme.user_theme_ids
+    theme_ids.all? { |id| allowed_ids.include?(id) }
```
SamSaffron: This is simpler/faster `(theme_ids - allowed_ids).empty?`

app/services/user_updater.rb
OsamaSayegh: The guardian above is for the acting user (e.g. when an admin changes a user's preferences), where as this new one is for the target user.

app/models/theme.rb
OsamaSayegh: I made the default for this setting (i.e. when there is no default theme) -1.

I made sure `theme_key` becomes `nil` when the setting value is -1 so that we don't end up calling `blank?` on a number. See:
https://github.com/discourse/discourse/pull/6073/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR353

lib/guardian.rb
OsamaSayegh: Done ✔️

app/models/theme.rb
OsamaSayegh: I'm happy to change it to 0 if it makes more sense.

SamSaffron:
getting real close, we are missing an `self.ignored_columns = ['key']` on `Theme` ... have a look at hos this is done on Topic at the top of the file.

OsamaSayegh:
@SamSaffron Done ✅

PR #6074, opened 2018-07-06
FIX: remove delete option from admin wrench for non-admins/moderators when all posts are selected

majakomel:
https://meta.discourse.org/t/tl4-users-can-see-delete-selected-option-when-selecting-all-posts-from-admin-wrench/59982


ZogStriP:
👍

PR #6076, opened 2018-07-07
Remove execjs

gdpelican:
Color me crazy, but it doesn't appear that execjs is used for anything within Discourse anymore?


SamSaffron:
both barber and uglifier need it ... I pulled it out in the past to the front to force the version cause I needed one with mini racer support. 

We could remove the line from Gemfile but the dependency would stay. my vote is just to leave this for now

PR #6078, opened 2018-07-08
Add S3 region eu-west-3 (Paris)

discoursehosting:


SamSaffron:
looks good to me! thanks

PR #6079, opened 2018-07-08
Prevent user custom fields new lines going under the user avatar on the profile page

Canapin:
Prevent user custom fields new lines going under the user avatar on the profile page when the user has no bio.

See https://meta.discourse.org/t/custom-fields-alignment-on-user-profile/91595


tgxworld:
Thank you @Canapin

PR #6080, opened 2018-07-09
UX: show custom footer on user activity and preferences pages

majakomel:
https://meta.discourse.org/t/custom-footer-disappears-on-some-profile-pages/87900


tgxworld:
The use preferences routes all have a common [parent route](https://github.com/discourse/discourse/blob/6cc860be095f92f0da280bf1ff97500c43014770/app/assets/javascripts/discourse/routes/preferences.js.es6). Would it work if we add `showFooter` to only that route?

majakomel:
@tgxworld adding `showFooter` to the parent route works only when initially loading the page, but not when switching between preferences

PR #6081, opened 2018-07-10
UX: Focus on new topic link that appears on the page via keyboard.

misaka4e21:
It was unable to focus on new topic link via Tab key.

With this PR, we can use Tab to focus on the link, then press Enter or Space to activate it.

### DIFF
--- a/app/assets/javascripts/admin/templates/web-hooks-show-events.hbs
+++ b/app/assets/javascripts/admin/templates/web-hooks-show-events.hbs
@@ -21,7 +21,7 @@
         <div class='clearfix'></div>
       </div>
       {{#if hasIncoming}}
-        <div class='alert alert-info clickable' {{action "showInserted"}}>
+        <div class='alert alert-info clickable' tabindex="0" {{action "showInserted"}}>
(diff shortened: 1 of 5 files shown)


coding-horror:
Why is this code needed?

```
+      if (event && event.type === 'keydown') {
+        if (event.key !== 'Enter' && event.key !== ' ') {
+          return;
+        }
+      }
```

If the element has a hyperlink in it, using regular default browser keyboard shortcuts should work?

misaka4e21:
@coding-horror 
The element is a `<div>`, and its only child is a `<span>`. Both of them requires an `onkeydown` attribution to trigger the action that will be activated by other keys such as Tab as well, which is to be limited by these code.

coding-horror:
I see, there is no <a href="">, well that's unfortunate. Carry on!

SamSaffron:
I worry about this keydown stuff being a bit brittle, can you introduce an `<a>` inside the div and then just give it the tabindex?

misaka4e21:
@SamSaffron Yes, it does.

SamSaffron:
excellent, nice fix!

PR #6082, opened 2018-07-10
FIX: ignore self-quotes from the same post when saving

OsamaSayegh:
https://meta.discourse.org/t/topic-quoting-itself-displays-reply-indicator/76085/22?u=osama

### DIFF
--- a/spec/models/quoted_post_spec.rb
+++ b/spec/models/quoted_post_spec.rb
@@ -33,6 +33,28 @@ describe QuotedPost do
     expect(QuotedPost.where(post_id: post4.id).pluck(:quoted_post_id)).to contain_exactly(post1.id, post2.id, post5.id)
   end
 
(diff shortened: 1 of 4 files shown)


spec/models/quoted_post_spec.rb
```diff
@@ -33,6 +33,28 @@
...
+          <div class="quote-controls"></div>
+          <img width="20" height="20" src="/user_avatar/meta.discourse.org/techapj/20/3281.png" class="avatar">techAPJ:
+        </div>
+        <blockquote><p>When the user will v</p></blockquote>
+      </aside>
+    HTML
+    post.save!
+
+    QuotedPost.extract_from(post)
+    expect(QuotedPost.where(post_id: post.id).count).to eq(0)
```
tgxworld: I usually prefer `QuotedPost.exists?(post_id: post.id)` instead of having to do a count.

PR #6083, opened 2018-07-11
PERF: do not carry post ids in memory when rebaking all posts

techAPJ:
@tgxworld can you review this commit? Thanks!


techAPJ:
Note that I added `Post.update_all('baked_version = NULL')` to make this task resumable.

To resume the task, all you'll need to do is run `posts:rebake_uncooked_posts`.

PR #6085, opened 2018-07-12
FEATURE: Add users:disable_2factor rake task

pfaffman:
https://meta.discourse.org/t/admin-locked-out-of-2fa/92156/2?u=pfaffman

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -142,7 +142,7 @@ task "users:disable_2fa", [:username] => [:environment] do |_, args|
   user = find_user(username)
   id = user.id
   UserSecondFactor.where(user_id: id).each(&:destroy!)
-  
+
   puts "", "2 Factor disabled for #{username}", ""
 end
 
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -135,6 +135,17 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
+desc "Disable 2 Factor for user."
+task "users:disable_2factor", [:username] => [:environment] do |_, args|
+  username = args[:username]
+
+  user = find_user(username)
+  id = user.id
+  UserSecondFactor.where(user_id: id).destroy_all
+
+  puts "", "2 Factor disabled for #{username}", ""
+end
+
 def find_user(username)
   user = User.find_by_username(username)
 
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -135,14 +135,14 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
-desc "Disable 2 Factor for user."
-task "users:disable_2factor", [:username] => [:environment] do |_, args|
+desc "Disable 2fa for username."
+task "users:disable_2fa", [:username] => [:environment] do |_, args|
   username = args[:username]
 
   user = find_user(username)
   id = user.id
-  UserSecondFactor.where(user_id: id).destroy_all
-
+  UserSecondFactor.where(user_id: id).each(&:destroy!)
+  
   puts "", "2 Factor disabled for #{username}", ""
 end
 


lib/tasks/users.rake
```diff
@@ -135,6 +135,17 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
+desc "Disable 2 Factor for user."
+task "users:disable_2factor", [:username] => [:environment] do |_, args|
+  username = args[:username]
+
+  user = find_user(username)
+  id = user.id
+  UserSecondFactor.where(user_id: id).destroy_all
```
tgxworld: You'll want to do `UserSecondFactor.where(user_id: id).each(&:destroy!)` instead. If something fails to destroy, it'll raise an error and exit.

lib/tasks/users.rake
```diff
@@ -135,6 +135,17 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
+desc "Disable 2 Factor for user."
+task "users:disable_2factor", [:username] => [:environment] do |_, args|
```
tgxworld: Perhaps a better name would be `disable_2fa`?

pfaffman:
Hey, @tgxworld . I pushed the changes. Do I need to do another PR? (It doesn't seem like it, but I can't quite tell).

lib/tasks/users.rake
```diff
@@ -135,6 +135,17 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
+desc "Disable 2fa for username."
+task "users:disable_2fa", [:username] => [:environment] do |_, args|
+  username = args[:username]
+
+  user = find_user(username)
+  id = user.id
+  UserSecondFactor.where(user_id: id).each(&:destroy!)
```
tgxworld: Oops it looks this shouldn't destroy everything. We have two different methods now so we should only disabled totp?

lib/tasks/users.rake
```diff
@@ -135,6 +135,17 @@ GROUP BY p.user_id
   puts 'Done!', ''
 end
 
+desc "Disable 2fa for username."
```
tgxworld: Small nitpick here but `2fa` should be capitalized.

lib/tasks/users.rake
```diff
@@ -135,6 +135,17 @@ GROUP BY p.user_id
...
+task "users:disable_2fa", [:username] => [:environment] do |_, args|
+  username = args[:username]
+
+  user = find_user(username)
+  id = user.id
```
tgxworld: we don't need this variable here since we can fit `user.id` in the next line.

lib/tasks/users.rake
tgxworld: The description of the rake task might have to change here.

pfaffman: I don't know. That's way beyond my pay grade. I based my PR on what @SamSaffron posted in the link below and assumed that it was a good way to turn of 2fa.

https://meta.discourse.org/t/admin-locked-out-of-2fa/92156/2?u=pfaffman

I think that a rake task to disable two factor authentication would be helpful to people who get locked out. 

It seems like you have already spent more time responding to my PR than it would have taken you to write it from scratch, and it's not clear that I can get it right on one more try, so I'll bow out.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/0ed2834c2deac6e48ec09e7dd6fd9b71bc88d175

lib/tasks/users.rake
pfaffman: Hey, @tgxworld. NOW I see what you mean about totp! :-)

Sadly, the backup codes didn't exist before and I just left my phone in a Paris taxi. I just submitted a PR for how I think we need to deal with the totp issue. Sorry I didn't understand before.

https://github.com/discourse/discourse/pull/6259

PR #6086, opened 2018-07-12
FEATURE: show modal when user without permissions tries to delete their own topic

majakomel:
https://meta.discourse.org/t/why-i-cannot-delete-my-topic/69752/2

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2098,6 +2098,8 @@ en:
         lock_post_description: "prevent the poster from editing this post"
         unlock_post: "Unlock Post"
         unlock_post_description: "allow the poster to edit this post"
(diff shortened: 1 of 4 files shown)


tgxworld:
@majakomel We should add an acceptance test here if possible.

config/locales/client.en.yml
```diff
@@ -318,6 +318,7 @@ en:
     undo: "Undo"
     revert: "Revert"
     failed: "Failed"
+    ok: "OK"
```
tgxworld: Let's use `Yes/No` instead of `OK` since all the other translations are doing so.

coding-horror:
"unpermitted" isn't .. English as far as I can tell? "disallowed"? We shouldn't commit an unword like this ;)

PR #6087, opened 2018-07-12
Edit: add link to official plugins list

pfaffman:
It took me a while to find that https://github.com/discourse/discourse/blob/master/lib/plugin/metadata.rb is where the green-checked plugins get determined.

### DIFF
--- a/docs/PLUGINS.md
+++ b/docs/PLUGINS.md
@@ -3,6 +3,8 @@
 
 If you just want to get some plugins for your Discourse instance, check out [the plugin category](https://meta.discourse.org/c/plugin) at meta. This is the most up to date place for plugin discussion and listing.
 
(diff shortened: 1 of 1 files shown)


docs/PLUGINS.md
```diff
@@ -3,6 +3,8 @@
 
 If you just want to get some plugins for your Discourse instance, check out [the plugin category](https://meta.discourse.org/c/plugin) at meta. This is the most up to date place for plugin discussion and listing.
 
+IF you want to be safe, use only plugins on this list of [Official Plugins](https://github.com/discourse/discourse/blob/master/lib/plugin/metadata.rb):
```
tgxworld: `IF` shouldn't be all caps.

tgxworld:
Thank you @pfaffman

PR #6089, opened 2018-07-14
FIX: grammar/spelling error in server.en.yml

kdwink:


SamSaffron:
Thanks, nice catch! (checked btw and CLA was signed , thanks for that!)

PR #6094, opened 2018-07-16
Set result.target for update (matching _saveNew)

kleinfreund:
Saving a new record (with `save()`) or updating a record (with `update()`) currently behave differently in that only `save()` will return a result object with a reference to the record object (i.e. `result.target` is set). Using `update()` will always miss that reference.

This pull request adds the reference to the result object for `update()` to match how `_saveNew()` works.

**Further notes**:

One reason why this is problematic is the ability to rely solely on *updating* records even for creating them in the first place. This happens when one provides the `id` property for the `attrs` object when calling `this.store.createRecord()`. The application logic treats records with an `id` property as existing and will always lead to a PUT request (whether this logic is desirable is an entirely different matter). With that, one never gets a response containing a reference to the record object from the server. Although `createRecord` already returns that object, one can’t be sure that the server-side representation is exactly identical. I wish to rely on the server-side representation.

### DIFF
--- a/app/assets/javascripts/discourse/models/rest.js.es6
+++ b/app/assets/javascripts/discourse/models/rest.js.es6
@@ -30,6 +30,7 @@ const RestModel = Ember.Object.extend({
 
         self.setProperties(payload);
         self.afterUpdate(res);
+        res.target = self;
         return res;
       })
       .finally(() => this.set("isSaving", false));
--- a/test/javascripts/models/rest-model-test.js.es6
+++ b/test/javascripts/models/rest-model-test.js.es6
@@ -21,13 +21,17 @@ QUnit.test("update", assert => {
   const store = createStore();
   return store.find("widget", 123).then(function(widget) {
     assert.equal(widget.get("name"), "Trout Lure");
+    assert.ok(!widget.get("isSaving"), "it is not saving");
 
-    assert.ok(!widget.get("isSaving"));
     const promise = widget.update({ name: "new name" });
-    assert.ok(widget.get("isSaving"));
-    promise.then(function() {
-      assert.ok(!widget.get("isSaving"));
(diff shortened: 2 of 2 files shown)


kleinfreund:
Slightly related: In `_saveNew`, the record will be eventually send out to the server via an AJAX call. Therefor, shouldn’t the `isNew` property be set to `false` in the `finally` call?

```
.createRecord(store, type, props)
.then(/* ... */)
.finally(() => {
  this.set("isSaving", false);
  this.set("isNew", false);
})
```

Otherwise, `this.store.destroyRecord` won’t delete the record from the server when it was just created and already commited.

In other words, the property `isNew` represents records that are not yet stored on the server side, correct?

eviltrout:
Yes `isNew` is meant to represent a record created in the Javascript application but not persisted yet.

app/assets/javascripts/discourse/models/rest.js.es6
```diff
@@ -30,6 +30,7 @@ const RestModel = Ember.Object.extend({
 
         self.setProperties(payload);
         self.afterUpdate(res);
+        res.target = self;
```
eviltrout: This seems reasonable to me, but we need to add a test for this functionality.

kleinfreund: Adding tests for that right now. Currently, they look like this:

```js
promise.then(function(result) {
  assert.ok(result.target, "it has a reference to the record");
  // ...
});
```

While I’m at it, is there something more I should test for here?

kleinfreund:
@eviltrout I added tests for the `update` and `save` methods including a `target` property in their response.

eviltrout:
Thanks, merged!

PR #6095, opened 2018-07-16
FIX: set default for theme.key, if it has not yet been dropped

davidtaylorhq:
<s>Need to use `Migration::SafeMigrate.disable!`, because the SQL query that gets generated contains the word "DROP"

```
ALTER TABLE "themes" ALTER "key" DROP NOT NULL
```

`column_exists?` is used to make sure this doesn't run if the `key` column has already been dropped.</s>

Set a default value, so that we don't hit issues with the NOT_NULL constraint while waiting for the column to be dropped


SamSaffron:
looks good to me!

PR #6097, opened 2018-07-16
FIX: mixed-text-direction list padding

scossar:
When the support_mixed_text_direction setting is enabled, the text direction of cooked posts may be set in the opposite direction of the loaded stylesheet. The only way I can see to guarantee that the correct padding will be added to a list is to add the padding to both the left and the right of the list.

We don't have the same issue with ordered lists. In that case, the padding is being added by the user-agent stylesheet. On Chrome, that is `-webkit-padding-start: 40px;`. This works correctly for both ltr and rtl lists.

### DIFF
--- a/app/assets/stylesheets/common/base/rtl.scss
+++ b/app/assets/stylesheets/common/base/rtl.scss
@@ -56,3 +56,10 @@
 .rtl .dashboard-right {
   float: left !important;
 }
+
+// This is for the support_mixed_text_direction setting
+.cooked ul[dir="ltr"], .d-editor-preview ul[dir="ltr"],
+.cooked ul[dir="rtl"], .d-editor-preview ul[dir="rtl"] {
+  padding-right: 40px;
+  padding-left: 40px;
+}


scossar:
The `r2` gem that is used for flipping the rtl stylesheet will skip rules that are preceded by a `/* no-r2 */` comment. Unfortunately, we can't use this comment, because the stylesheets are compiled from SASS before they are flipped, so the comment is stripped out.

Ideally, what we want is this:
```
.cooked ul[dir="ltr"], .d-editor-preview ul[dir="ltr"] {
  padding-right: 0;
  padding-left: 40px;
}

.cooked ul[dir="rtl"], .d-editor-preview ul[dir="rtl"] {
  padding-right: 40px;
  padding-left: 0;
}
```
The problem is that we don't know if the user is getting an rtl stylesheet or not. If they are getting the rtl styles, then the above css will be flipped to
```
.cooked ul[dir="ltr"], .d-editor-preview ul[dir="ltr"] {
  padding-left: 0;
  padding-right: 40px;
}

.cooked ul[dir="rtl"], .d-editor-preview ul[dir="rtl"] {
  padding-left: 40px;
  padding-right: 0;
}
```
The easiest solution is to just add the `40px` padding to both the left and the right when the `dir` attribute is set on a list.

PR #6098, opened 2018-07-17
FIX: don't include theme HTML in static pages

OsamaSayegh:
https://meta.discourse.org/t/javascript-not-loading-on-certain-pages/92297?u=osama


SamSaffron:
I think I am so torn here ... on one hand I totally agree to remove this, on the other hand... it is going to cause lots of confusion 

I think the simplest fix here is just to define this at the top: 

```
<script>
Discourse = { 
  _registerPluginCode: function() {};
}
<script>
```

Then we correctly protect our core and don't amend the behavior.

SamSaffron:
I like this change ... it is great for safety!

PR #6099, opened 2018-07-17
FEATURE: Social auth improvements

davidtaylorhq:
This is the start of work to add revoke and reconnect for social logins.

## Visible/functional changes:
- Additional section under `/my/preferences/account` which lists all connected accounts. 
- For providers with support, it is possible to revoke and reconnect accounts. 
- Currently this only actually works for Facebook, but more providers will be added soon.

## Screenshots:
screen shot 2018-07-17 at 16 23 48

screen shot 2018-07-17 at 16 24 05

## Under-the-hood changes:
- Moved client-side social login code from the login controller to `login_method.js.es6`, so that it can be used for reconnecting accounts
- `user.associated_accounts` is now an array of objects, rather than a single string
- New endpoint for revoking an auth provider
- Plugin authenticators are now stored in the `PluginRegistry`, to match up with other similar plugin-contributed functionality
- Plugin and core authentication providers are now ‘equal citizens` on the server side - no more copy/pasting for facebook/twitter/instagram etc.
- Each authenticator can now implement:
    - `enabled?`
    - `description_for_user`
    - `can_revoke?`
    - `can_connect_existing_user?`
    - `revoke`
- Revoke and reconnect ability added to facebook_authenticator

The plugin API remains identical. For now, plugins that have not defined `enabled?` will log a warning, but will still function fine.


lib/auth/facebook_authenticator.rb
```diff
@@ -6,7 +6,50 @@ def name
...
+
+    uri = URI.parse("https://graph.facebook.com/#{info.facebook_user_id}/permissions?access_token=#{SiteSetting.facebook_app_id}|#{SiteSetting.facebook_app_secret}")
+
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request = Net::HTTP::Delete.new(uri.request_uri)
+
+    response = http.request(request)
+
+    if response.kind_of? Net::HTTPSuccess
```
jjaffeux: To take advantage of implicit return of ruby

```
if response.kind_of? Net::HTTPSuccess
  info.destroy
  return true
end

false
```

lib/auth/facebook_authenticator.rb
```diff
@@ -6,7 +6,50 @@ def name
...
+    info.email || info.username || ""
+  end
+
+  def can_revoke?
+    true
+  end
+
+  def revoke(user, skip_remote: false)
+    info = FacebookUserInfo.find_by(user_id: user.id)
+    raise Discourse::NotFound if info.nil?
```
jjaffeux: Should be enough

```ruby
raise Discourse::NotFound unless info
```

lib/auth/facebook_authenticator.rb
```diff
@@ -6,7 +6,50 @@ def name
...
   end
 
-  def after_authenticate(auth_token)
+  def enabled?
+    SiteSetting.enable_facebook_logins
+  end
+
+  def description_for_user(user)
+    info = FacebookUserInfo.find_by(user_id: user.id)
+    return nil if info.nil?
```
jjaffeux:
```ruby
return nil unless info
```

app/assets/javascripts/discourse/models/login-method.js.es6
```diff
@@ -17,12 +17,63 @@ const LoginMethod = Ember.Object.extend({
     );
   },
 
+  @computed
+  prettyName() {
+    return I18n.t("login." + this.get("name") + ".name");
```
jjaffeux:
```js
return I18n.t(`login.${this.get("name")}.name`);
```

app/assets/javascripts/discourse/controllers/preferences/account.js.es6
```diff
@@ -135,6 +170,26 @@ export default Ember.Controller.extend(
...
+        this.set("revoking", true);
+        model
+          .revokeAssociatedAccount(account.name)
+          .then(result => {
+            if (result.success) {
+              model.get("associated_accounts").removeObject(account);
+            } else {
+              bootbox.alert(result.message);
+            }
+            this.set("revoking", false);
```
jjaffeux: Not sure about all the logic here, but should this be in a finally block instead?

app/assets/javascripts/discourse/controllers/preferences/account.js.es6
```diff
@@ -54,6 +55,40 @@ export default Ember.Controller.extend(
...
+    authProviders(accounts) {
+      const allMethods = findAll(
+        this.siteSettings,
+        this.capabilities,
+        this.site.isMobileDevice
+      );
+
+      return allMethods.map(method => {
+        return {
+          method: method,
```
jjaffeux: s/method: method/method You don’t need to repeat if key/value has same name

app/assets/javascripts/discourse/controllers/preferences/account.js.es6
```diff
@@ -54,6 +55,40 @@ export default Ember.Controller.extend(
       );
     },
 
+    @computed("model.associated_accounts")
+    associatedAccountsLoaded(associatedAccounts) {
+      return associatedAccounts !== undefined;
```
jjaffeux: I think it's better to check against typeof here (https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript)

```
return type associatedAccounts !== "undefined";
```

app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -40,6 +40,18 @@ export default Ember.Controller.extend(CanCheckEmails, {
       .join(", ");
   },
 
+  @computed("model.associated_accounts")
+  associatedAccountsLoaded(associatedAccounts) {
+    return associatedAccounts !== undefined;
```
jjaffeux: check other comment on using typeof here instead

SamSaffron:
how backwards compatible is this, will our auth provider plugins continue working ?

davidtaylorhq:
> how backwards compatible is this, will our auth provider plugins continue working ?

Changes are 100% backwards compatible. The new features will not work for plugins until they are updated. 

For example, I installed and set up the patreon plugin locally. I can log in fine, but it looks like this in my user preferences.

screen shot 2018-07-18 at 09 40 01

I will go through and update plugins as soon as this is merged.

---
Thanks for the review @jjaffeux, I will make those changes and then get this merged :)

SamSaffron:
what does (not connected) mean can you not connect Twitter?

davidtaylorhq:
Correct, I haven't implemented revoke or connect for Twitter yet. Only facebook has that functionality so far. However, all the scaffolding is there, so it should be pretty quick for me to add all the others.

lib/auth/facebook_authenticator.rb
```diff
@@ -6,7 +6,50 @@ def name
...
+  def revoke(user, skip_remote: false)
+    info = FacebookUserInfo.find_by(user_id: user.id)
+    raise Discourse::NotFound if info.nil?
+
+    if skip_remote
+      info.destroy!
+      return true
+    end
+
+    uri = URI.parse("https://graph.facebook.com/#{info.facebook_user_id}/permissions?access_token=#{SiteSetting.facebook_app_id}|#{SiteSetting.facebook_app_secret}")
```
tgxworld: If the URL endpoint is fixed, we should store it in a constant and append the params to it.

lib/auth/facebook_authenticator.rb
```diff
@@ -6,7 +6,50 @@ def name
...
+
+    http = Net::HTTP.new(uri.host, uri.port)
```
tgxworld: Let's standardize and use `Excon` instead.

lib/auth/facebook_authenticator.rb
```diff
@@ -20,7 +63,14 @@ def after_authenticate(auth_token)
 
     user_info = FacebookUserInfo.find_by(facebook_user_id: facebook_hash[:facebook_user_id])
 
-    result.user = user_info.try(:user)
+    if existing_account && (user_info.nil? || existing_account.id != user_info.user_id)
+      user_info.destroy! if user_info
+      result.user = existing_account
+      user_info = FacebookUserInfo.create({ user_id: result.user.id }.merge(facebook_hash))
+    else
+      result.user = user_info.try(:user)
```
tgxworld: `user_info&.user` is faster than using `.try`. No reason to use `try` anymore :grinning:

lib/auth/facebook_authenticator.rb
```diff
@@ -20,7 +63,14 @@ def after_authenticate(auth_token)
 
     user_info = FacebookUserInfo.find_by(facebook_user_id: facebook_hash[:facebook_user_id])
 
-    result.user = user_info.try(:user)
+    if existing_account && (user_info.nil? || existing_account.id != user_info.user_id)
+      user_info.destroy! if user_info
+      result.user = existing_account
+      user_info = FacebookUserInfo.create({ user_id: result.user.id }.merge(facebook_hash))
```
tgxworld: it doesn't look like we're checking for the return value of `#create`. Therefore, we'll want to use `create!` here.

lib/auth/github_authenticator.rb
```diff
@@ -6,6 +6,16 @@ def name
     "github"
   end
 
+  def enabled?
+    SiteSetting.enable_github_logins
+  end
+
+  def description_for_user(user)
+    info = GithubUserInfo.find_by(user_id: user.id)
+    return nil if info.nil?
```
tgxworld: Are we able to return a blank string here instead? `info.blank? ? "" : (info.screen_name || "")`

lib/plugin/instance.rb
```diff
@@ -488,6 +489,19 @@ def auth_provider(opts)
     Plugin::AuthProvider.auth_attributes.each do |sym|
       provider.send "#{sym}=", opts.delete(sym)
     end
+
+    begin
+      provider.authenticator.enabled?
+    rescue NotImplementedError
+      # TODO: Remove once plugins are all updated to define enabled? on authenticators
```
tgxworld: Hmm this is very hard to enforce since someone in the future looking at the comment wouldn't know which plugins are affected. I think we can just patch it and print the warning.

spec/components/auth/facebook_authenticator_spec.rb
```diff
@@ -38,6 +38,36 @@
...
+            "Website" => "https://awesome.com"
+          }
+        },
+        "uid" => "100"
+      }
+
+      result = authenticator.after_authenticate(hash, existing_account: user2)
+
+      expect(result.user.id).to eq(user2.id)
+      expect(FacebookUserInfo.find_by(user_id: user1.id)).to be(nil)
```
tgxworld: `.exists?` would be more efficient here.

lib/auth/facebook_authenticator.rb
tgxworld: We should extract the forming of the URL into a method so that it can be reused in our tests. Otherwise, changing the URL would require us to change it multiple spots.

spec/components/discourse_spec.rb
```diff
@@ -59,6 +59,53 @@
...
+          Users::OmniauthCallbacksController::BUILTIN_AUTH + [plugin_auth_provider.authenticator])
+      end
+
+    end
+  end
+
+  context 'enabled_authenticators' do
+    it 'only returns enabled authenticators' do
+      expect(Discourse.enabled_authenticators.length).to be(0)
+      SiteSetting.enable_twitter_logins = true
```
tgxworld: You can do this instead `expect { SiteSetting.enable_twitter_logins = true }.to change { Discourse.enabled_authenticators.length }.by(1)`

SamSaffron:
let's change it so if the interface to "connect" is missing (and it is not connected) we just omit it from the list for the user, this will reduce confusion. 

except for that let's try getting it merged in once @tgxworld's comments are addressed.

Toxu-ru:
Good change. 

Is the translation not working correctly?

Alt Text

davidtaylorhq:
The translation should be working - I just tried changing it on Meta and it worked straight away. Make sure to reload the page fully. Try viewing in incognito mode to be sure.

Toxu-ru:
Is it possible that it does not work if the site version is different from the English version?

PR #6101, opened 2018-07-17
FEATURE: part 2 of dashboard improvements

jjaffeux:
- moderation tab
- sorting/pagination
- improved third parti reports support
- trending charts
- better perf
- many fixes
- refactoring
- new reports


jjaffeux:
admin - discourse meta - general tab

admin - discourse meta - moderation

app/models/incoming_links_report.rb
```diff
@@ -27,6 +28,7 @@ def self.find(type, _opts = {})
     report = IncomingLinksReport.new(type)
 
     report.start_date = _opts[:start_date] || 30.days.ago
+    report.end_date = _opts[:end_date] || Time.now.utc.end_of_day
```
tgxworld: Is there a reason we use `Time.now.utc` instead of following the server's timezone?

app/models/report.rb
```diff
@@ -39,6 +46,28 @@ def self.clear_cache
     end
   end
 
+  def self.wrap_slow_query(timeout = 20000, &block)
+    begin
+      ActiveRecord::Base.connection.transaction do
+        # Set a statement timeout so we can't tie up the server
+        DB.exec "SET LOCAL statement_timeout = #{timeout}"
```
tgxworld: I'm not 100% sure but I think `DB.exec` might not be run in the context of this transaction because it bypasses AR.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+      { properties: [:topic_count], title: I18n.t("reports.moderators_activity.labels.topic_count") },
+      { properties: [:post_count], title: I18n.t("reports.moderators_activity.labels.post_count") }
+    ]
+
+    report.modes = [:table]
+
+    report.data = []
+    mod_data = {}
+
+    User.real.where(moderator: true).pluck(:id, :username).each do |u|
```
tgxworld: Just in case, you might want to use `find_each` instead of `each` here.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+      mod_data[u[0]] = {
+        user_id: u[0],
+        username: u[1],
+        user_url: "/admin/users/#{u[0]}/#{u[1]}"
```
tgxworld: Can we form this URL on the client side instead since it shares the same pattern. This way we just have to send the user_id and the username down which makes the payload smaller too. We'll also use less memory on the server side.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+
+    mod_ids = mod_data.keys
+
+    return if mod_ids.empty?
+
+    time_read_query = <<~SQL
+    SELECT SUM(time_read) AS time_read,
+    user_id
+    FROM user_visits
+    WHERE user_id = ANY(ARRAY#{mod_ids})
```
tgxworld: Instead of matching this again the `user_visits` table, you can filter this against the `users` table and then join on the `user_visits` table. It should be faster since we will be filtering against less records.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+    AND visited_at >= '#{report.start_date}'
+    AND visited_at <= '#{report.end_date}'
+    GROUP BY user_id
+    SQL
+
+    flag_count_query = <<~SQL
+    SELECT agreed_by_id AS user_id,
+    COUNT(*) AS flag_count
+    FROM post_actions
+    WHERE (agreed_by_id = ANY(ARRAY#{mod_ids}) OR disagreed_by_id = ANY(ARRAY#{mod_ids}))
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/6101/files#r203289896

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+    COUNT(*) AS post_count
+    FROM posts
+    WHERE user_id = ANY(ARRAY#{mod_ids})
+    AND created_at >= '#{report.start_date}'
+    AND created_at <= '#{report.end_date}'
+    GROUP BY user_id
+    SQL
+
+    DB.query(time_read_query).each do |row|
+      mod_data[row.user_id][:time_read] = row.time_read if row.user_id
```
tgxworld: `row.user_id` will never be blank since there is a `NOT NULL` constraint on the column.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+      { type: :link, properties: [:username, :user_url], title: I18n.t("reports.moderators_activity.labels.moderator") },
+      { properties: [:flag_count], title: I18n.t("reports.moderators_activity.labels.flag_count") },
+      { type: :seconds, properties: [:time_read], title: I18n.t("reports.moderators_activity.labels.time_read") },
+      { properties: [:topic_count], title: I18n.t("reports.moderators_activity.labels.topic_count") },
+      { properties: [:post_count], title: I18n.t("reports.moderators_activity.labels.post_count") }
+    ]
+
+    report.modes = [:table]
+
+    report.data = []
```
tgxworld: I don't think this is required since `mod_data.values` will return an empty array if the hash is empty.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+    pa.post_action_type_id,
+    pa.created_at,
+    pa.agreed_at,
+    pa.disagreed_at,
+    pa.deferred_at,
+    pa.agreed_by_id,
+    pa.disagreed_by_id,
+    pa.deferred_by_id,
+    pa.user_id AS flagger_id,
+    (select u.username FROM users u WHERE u.id = pa.user_id) AS flagger_username,
```
tgxworld: I think we should do a join on the users table instead since this will cause PostgreSQL to run a sub-queury for each row returned.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+      data[:staff_id] = row.staff_id
+      if row.staff_username && row.staff_id
+        data[:staff_url] = "/admin/users/#{row.staff_id}/#{row.staff_username}"
+      end
+      data[:poster_username] = row.poster_username
+      data[:poster_id] = row.poster_id
+      data[:poster_url] = "/admin/users/#{row.poster_id}/#{row.poster_username}"
+      data[:flagger_id] = row.flagger_id
+      data[:flagger_username] = row.flagger_username
+      data[:flagger_url] = "/admin/users/#{row.flagger_id}/#{row.flagger_username}"
```
tgxworld: I think all these URLs should be generated on the client side.

app/models/report.rb
```diff
@@ -478,8 +613,217 @@ def self.report_trending_search(report)
...
+    p.user_id AS author_id,
+    pr.number AS revision_version,
+    p.version AS post_version,
+    pr.post_id,
+    p.topic_id,
+    p.post_number,
+    p.edit_reason,
+    u.username AS editor_username,
+    pr.created_at,
+    (SELECT u.username FROM users u WHERE u.id = p.user_id) AS author_username
```
tgxworld: Ditto N+1 query problem.

app/models/report.rb
jjaffeux: I can ’t actually do this, as I want reports to be "standalone" and it would force us to carry more logic. Maybe I can reconsider this in the future.

app/models/report.rb
jjaffeux: It seems to work and this is also what we do in data explorer plugin: https://github.com/discourse/discourse-data-explorer/blob/master/plugin.rb#L90

app/models/report.rb
jjaffeux: I can ’t actually do this, as I want reports to be "standalone" and it would force us to carry more logic. Maybe I can reconsider this in the future.

app/models/report.rb
SamSaffron: yes it will always run within the transaction AR makes.

PR #6103, opened 2018-07-18
FIX: TopicLink URLs should be case-insensitive

techAPJ:
@SamSaffron Can you verify this commit? Will `validates_uniqueness_of :url, case_sensitive: false` create issues when the duplicate record (`google.com` vs `Google.com`) is updated? If yes, we might need to run a migration merging existing duplicated records.

### DIFF
--- a/app/models/topic_link.rb
+++ b/app/models/topic_link.rb
@@ -22,11 +22,13 @@ class TopicLink < ActiveRecord::Base
 
   validates_length_of :url, maximum: 500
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
I would just skip the uniqueness check and just do the downcasing. BUT 

The actual fix is not correct. Domain is not case sensitive, but the actual path is...

techAPJ:
Good point. Closing this PR for now, will push a separate commit for Domain only downcasing.

PR #6104, opened 2018-07-18
PERF: Move `EmailLog#reply_key` into new `post_reply_keys` table.

tgxworld:


spec/components/email/receiver_spec.rb
```diff
@@ -813,13 +820,16 @@ def process(email_name)
     end
 
     context "with a valid reply" do
-      it "returns the destination when the key is valid" do
-        Fabricate(:email_log, reply_key: '4f97315cc828096c9cb34c6f1a0d6fe8')
+      it "returns the destyination when the key is valid" do
```
gschlager: typo in `destyination` 😉

lib/email/sender.rb
```diff
@@ -189,7 +195,7 @@ def send
         return skip(e.message)
       end
 
-      # Save and return the email log
+      # Save and return the email logtox
```
gschlager: What's `logtox`?

tgxworld: oops typo

spec/components/email/receiver_spec.rb
tgxworld: hehehe

PR #6106, opened 2018-07-18
FIX: topic owner should watch the new topic when moving posts to a new topic

OsamaSayegh:
https://meta.discourse.org/t/when-a-post-is-moved-to-a-new-topic-by-a-moderator-the-moved-posts-author-is-not-watching/51779/3?u=osama

### DIFF
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -34,6 +34,8 @@ class PostMover
       )
       DiscourseTagging.tag_topic_by_names(new_topic, Guardian.new(user), tags)
       move_posts_to new_topic
+      force_user_to_watch_new_topic
+      new_topic
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/models/post_mover.rb
```diff
@@ -34,6 +34,8 @@ def to_new_topic(title, category_id = nil, tags = nil)
       )
       DiscourseTagging.tag_topic_by_names(new_topic, Guardian.new(user), tags)
       move_posts_to new_topic
+      force_user_to_watch_new_topic
```
tgxworld: This can just be `watch_new_topic` IMO.

spec/requests/topics_controller_spec.rb
```diff
@@ -116,6 +119,30 @@
           expect(Tag.all.pluck(:name)).to contain_exactly("tag1", "tag2")
         end
 
+        it "forces resulting topic owner to watch the new topic" do
```
tgxworld: I don't think this should be tested in a request spec. It should instead be tested in `post_mover_spec`.

OsamaSayegh:
@tgxworld done ✅ 

Thanks for your feedback <3

tgxworld:
@OsamaSayegh Looks good :+1:

PR #6107, opened 2018-07-18
Add User Api Key headers to CORS

pdavide:
When trying to use the [User API key](https://meta.discourse.org/t/user-api-keys-specification/48536)  feature I get a `Request header field User-Api-Key is not allowed by Access-Control-Allow-Headers` error in the browser console (Chrome 67).
This PR updates the `Access-Control-Allow-Headers` header to allow the use of User API key.


pdavide:
@SamSaffron maybe the same headers need to be added to https://github.com/discourse/discourse/blob/7d9c97d66141d35d00258fe544211d9fd7f79a76/config/initializers/008-rack-cors.rb#L42 ?

pdavide:
As explained in [this discussion](https://meta.discourse.org/t/request-header-field-user-api-key-is-not-allowed-by-access-control-allow-headers/92662/5) the file we need to update is `discourse/config/initializers/008-rack-cors.rb`.

PR #6108, opened 2018-07-18
Fix notifications for topics moved between categories

davidtaylorhq:
Fix for https://meta.discourse.org/t/not-receiving-notifications-for-announcements/87275/15

It turns out all of the cases in that topic have regressed since 71f66cd (@ZogStriP). Moving the notification stuff into a Sidekiq job introduced a race condition. If the Sidekiq job runs before the database transaction has been committed, then it fails.

This was not picked up by tests because scheduled jobs are run immediately and in the current thread (and therefore within the current database transaction).

This particular case sometimes occurs inside multiple nested transactions, so simply moving the offending line outside of the transaction is not enough.

I've implemented `TransactionHelper`, which allows us to use `TransactionHelper.after_commit`
to define code to be run after the current transaction has been committed. 
(Slightly adapted version of https://dev.to/evilmartians/rails-aftercommit-everywhere--4j9g)

If anyone has an easier/better way, please let me know!


ZogStriP:
This looks good 👍 

> If anyone has an easier/better way, please let me know!

Maybe @SamSaffron has an idea?

app/helpers/transaction_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+      @callback.call
+    end
+
+    def before_committed!(*); end
+    def rolledback!(*); end
+  end
+
+  def self.after_commit(&blk)
+    ActiveRecord::Base.connection.add_transaction_record(
+        AfterCommitWrapper.new(&blk)
```
tgxworld: Be sure to have rubocop running in your IDE. grin

spec/helpers/transaction_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe TransactionHelper do
+
+  it "runs callbacks after outermost transaction is committed" do
+    outputString = "1"
+
+      # Main transaction
```
tgxworld: Indentation looks off here :stuck_out_tongue_closed_eyes:

tgxworld:
@davidtaylorhq I like the pattern :+1: I think the change is quite safe to merge

SamSaffron:
Maybe `DB.after_commit` and `DB.transaction { ... }` 

Actual implementation can remain as is and support AR transactions but I feel it is a bit cleaner to have a single class for this API vs adding a TransactionHelper that is exposed.

spec/helpers/transaction_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe TransactionHelper do
+
+  it "runs callbacks after outermost transaction is committed" do
```
riking: Add a test verifying that the callback is not run when the transaction is rolled back.

SamSaffron:
@davidtaylorhq definitely does not belong in the gem cause it integrates into AR. 

I would simply have the helper methods live in `MiniSqlMultisiteConnection` and just add the tests against that. 

helps us cause there there is one less class to think about.

davidtaylorhq:
Moved into `MiniSQLMultisiteConnection`, and have added a spec for rollback per @riking's suggestion

PR #6109, opened 2018-07-18
FEATURE: allows to jump to a date in a topic

jjaffeux:


app/assets/javascripts/discourse/models/post-stream.js.es6
```diff
@@ -565,6 +565,15 @@ export default RestModel.extend({
     });
   },
 
+  loadNearestPostToDate(date) {
+    const url = `/posts/by_date/${this.get("topic.id")}/${date}`;
```
tgxworld: I made this mistake recently but apparently we should avoid underscore in URLs.

app/controllers/posts_controller.rb
```diff
@@ -707,6 +713,16 @@ def find_post_from_params_by_number
     find_post_using(by_number_finder)
   end
 
+  def find_post_from_params_by_date
+    by_date_finder = Post
```
tgxworld: You'll need to initialize an instance of a `TopicView` here and pass it the current filters for the topic. Otherwise, this query may return a post that is not within the stream.

app/assets/javascripts/discourse/models/post-stream.js.es6
jjaffeux: Yes but we have : "by_number" So it seemed weird to have one by-date and one by_number

app/controllers/posts_controller.rb
jjaffeux: So what we do in:

find_post_from_params_by_number
find_post_from_params

is also wrong?

PR #6110, opened 2018-07-19
FIX: store records with ID 0 not working reliably

kleinfreund:
*(This was brought up on meta.discourse.org: [Store: What’s the correct type of a record ID?](https://meta.discourse.org/t/store-what-s-the-correct-type-of-a-record-id/92605))*

---

This pull request addresses the inconsistent behavior of creating records in the store that have an ID property with the value `0`. Due to various checks like the following, records with ID 0 were treated like records without an ID.

```js
return !!attrs.id ? this._hydrate(type, attrs) : this._build(type, attrs);
```
I also renamed some instances of `obj` to `attrs`. If a method is called with an `attrs` parameter which the method doesn’t transform and then passes on to another method call, it should still be called `attrs`.

### DIFF
--- a/app/assets/javascripts/discourse/adapters/rest.js.es6
+++ b/app/assets/javascripts/discourse/adapters/rest.js.es6
@@ -42,7 +42,9 @@ export default Ember.Object.extend({
   },
 
   appendQueryParams(path, findArgs, extension) {
-    if (findArgs) {
+    // In the condition, checking for `findArgs` only is not sufficient because
+    // it might be `0`.
+    if (findArgs !== undefined) {
       if (typeof findArgs === "object") {
         const queryString = Object.keys(findArgs)
           .reject(k => !findArgs[k])
--- a/app/assets/javascripts/discourse/models/store.js.es6
+++ b/app/assets/javascripts/discourse/models/store.js.es6
@@ -10,13 +10,11 @@ function flushMap() {
   _identityMap = {};
 }
 
-function storeMap(type, id, obj) {
-  if (!id) {
-    return;
+function storeMap(type, id, record) {
+  if (id !== undefined) {
+    _identityMap[type] = _identityMap[type] || {};
+    _identityMap[type][id] = record;
   }
-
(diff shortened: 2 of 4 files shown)


kleinfreund:
When working on this, I ran the qunit test suite with the filters `rest-model` and `service:store`. With this, I seem to have missed some tests that no longer pass. I’m running the full test suite locally again.

kleinfreund:
When working on this, I ran the qunit test suite with the filters rest-model and service:store. With this, I seem to have missed some tests that no longer pass. I’m running the full test suite locally again.

It would be helpful to know what types of IDs are expected/allowed. From what I understood so far, the ID should be a number, but it might be passed around as a string (i.e. `"123"` or `123`)

kleinfreund:
Now, `createRecord` checks for `attrs.id` containing digits. This ensures that both strings and number IDs are possible.

Also, I needed to push that last commit again because a semicolon was accidentally removed; thus, leading to a failed travis build. I squashed that fix with the last commit so there is nothing to see here.

eviltrout:
This seems like a lot of code - what is the use case of having an id of 0? There's nowhere in discourse where I believe that applies.

kleinfreund:
@eviltrout Having an ID of 0 is not important. But since generally integer numbers are used for record IDs, 0 seems like a strange exclusion. Also, the way the previous code checked for existance of an ID has other cases that might cause unexpected results (e.g. empty string: `id === ""`).

The substantial changes are if conditions where the condition changed from `id` to `id !== undefined` with the semantics of *an ID value exists*. That’s 4 lines of code. Also, the `createRecord` method now throws an error that would otherwise occur later (when sending a request with a non-integer id value to the server).

Most of the lines of codes being changed are due to the rename ob `obj` to `attrs` as explained in the original post.

eviltrout:
I still don't see the advantage of making it more robust here, as supporting the cases of `0` or `""` don't seem like something I ever want it to do?

I'm okay with the rename of `obj` to `attrs` but I'm not on board with the original intent of the PR.

PR #6124, opened 2018-07-20
FEATURE: Webhook for user destroyed event

vinothkannans:


app/models/user.rb
```diff
@@ -1265,9 +1266,14 @@ def update_previous_visit(timestamp)
...
-    DiscourseEvent.trigger(:user_created, self)
-    true
+  %i{
+    user_created
+    user_destroyed
+  }.each do |event|
+    define_method("trigger_#{event}_event") do
+      DiscourseEvent.trigger(event, self)
+      true
+    end
```
SamSaffron: You know sometimes fancy is not really worth is, in this case we are just defining 2 methods, might as well just code them, also makes searching etc easier.

SamSaffron:
I am fine with this except for my minor change request, feel free to merge in once done.

PR #6128, opened 2018-07-20
FIX: notify staff about whispers in watched categories

majakomel:
https://meta.discourse.org/t/new-post-notifications-not-received-for-whispers-in-watched-categories/89747


ZogStriP:
👍

PR #6135, opened 2018-07-21
Fix dead link present in admin account registration instructions.

udan11:
The previous link that instructed users how to create an administrator account is dead now. I replaced the old link with a similar one and also mentioned the possibility of adding a developer email in the configuration file.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -3790,7 +3790,7 @@ en:
       button: "Register"
       title: "Register Admin Account"
       help: "register a new account to get started"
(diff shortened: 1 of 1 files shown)


SamSaffron:
@coding-horror your call on copy here

udan11:
Hey @coding-horror,

Can you check this again? @caugner has a valid point about this being a little confusing, especially when translated. What do you think?

coding-horror:
This whole thing is inside baseball, very detailed, only applicable to installers who are quite technical. So I don't view it as an urgent stop-the-world kind of thing that needs my attention 💃

PR #6136, opened 2018-07-21
Display "Replace Tags" and "Append Tags" buttons only if tagging is enabled.

udan11:
Fixes [this issue](https://meta.discourse.org/t/bulk-tag-buttons-appear-when-tagging-is-disabled/70092).


SamSaffron:
thanks! seems good to me.

PR #6137, opened 2018-07-21
Make the login error message more informative if user tries to log in with cookies disabled.

udan11:
Fixes [this issue](https://meta.discourse.org/t/unknown-error-when-trying-to-log-in-with-incorrect-credentials-with-cookies-disabled/79750).

The function which checks if cookies are enabled is a part of [Modernizr](https://github.com/Modernizr/Modernizr/blob/400db4043c22af98d46e1d2b9cbc5cb062791192/feature-detects/cookies.js).


SamSaffron:
Nice fix, thanks!

PR #6138, opened 2018-07-21
Clear old text and URL when showing up the "Insert Hyperink" modal.

udan11:
Fixes [this issue](https://meta.discourse.org/t/hyperlink-modal-retains-text-on-cancel/75477).

Instead of clearing the text before showing the modal, it was cleared after a successful insert which caused the modal to retain data if the Cancel button was used.

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -870,6 +870,9 @@ export default Ember.Component.extend({
         return;
       }
 
+      this.set("linkUrl", "");
+      this.set("linkText", "");
+
       this._lastSel = this._getSelected();
 
       if (this._lastSel) {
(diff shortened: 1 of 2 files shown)


SamSaffron:
I think it is worth adding a little test for this, can you do that?

udan11:
@SamSaffron Yes, of course. :)

SamSaffron:
Yes this looks great, I am confused a bit about the test failure though, any ideas? Maybe you need to rebase master. also can you use the async pattern there ... see: https://github.com/discourse/discourse/commit/7f7944bc3a7ff8051ac6b1448c0aa1f8a5edc8b8 for example.

udan11:
Hey @SamSaffron,

I am not sure about the test failure. I just supposed that it was something related to the instability of the test suite. It does not look to have anything to do with what I touched. 

By the way, can I refactor the others tests to use the async pattern? I can make another PR for that.

majakomel:
@udan11 I'm already working on moving the tests to async/await pattern :)

udan11:
@majakomel Oh, ok. Good! :)

SamSaffron:
awesome, thanks!

PR #6140, opened 2018-07-21
Add different trigger for the emoji popup of French users.

udan11:
Fixes [this issue](https://meta.discourse.org/t/default-emoji-plugin-how-to-disable/3415/71).

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -434,6 +434,10 @@ export default Ember.Component.extend({
           const full = `:${term}`;
           term = term.toLowerCase();
 
+          if (term.length < self.siteSettings.emoji_autocomplete_min_chars) {
+            return resolve([]);
+          }
+
           if (term === "") {
             return resolve(["slight_smile", "smile", "wink", "sunny", "blush"]);
           }
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1706,6 +1706,7 @@ en:
     enable_emoji_shortcuts: "Common smiley text such as :) :p :( will be converted to emojis"
     emoji_set: "How would you like your emoji?"
     enforce_square_emoji: "Force a square aspect ratio to all emojis."
+    emoji_autocomplete_min_chars: "Minimum number of characters required to trigger autocomplete emoji popup"
 
     approve_post_count: "The amount of posts from a new or basic user that must be approved"
     approve_unless_trust_level: "Posts for users below this trust level must be approved"
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -676,6 +676,11 @@ posting:
     enum: 'EmojiSetSiteSetting'
   enforce_square_emoji:
     default: true
+  emoji_autocomplete_min_chars:
+    client: true
+    default: 0
+    locale_default:
+      fr: 1
   approve_post_count:
     default: 0
   approve_unless_trust_level:


SamSaffron:
OK, looks like the right place, but we got to add a site setting here.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -434,6 +434,12 @@ export default Ember.Component.extend({
           const full = `:${term}`;
           term = term.toLowerCase();
 
+          if (term.length < 1 && I18n.locale == "fr") {
```
ZogStriP: Instead of testing for the French locale I would add a new site setting that controls the minimum number of characters before showing up the emoji modal.

Also we use `===` to test for equality ;)

config/site_settings.yml
```diff
@@ -676,6 +676,9 @@ posting:
     enum: 'EmojiSetSiteSetting'
   enforce_square_emoji:
     default: true
+  emoji_trigger:
+    default: 0
```
gschlager: You could add a `locale_default` for `fr`

config/locales/server.en.yml
```diff
@@ -1702,6 +1702,7 @@ en:
     enable_emoji_shortcuts: "Common smiley text such as :) :p :( will be converted to emojis"
     emoji_set: "How would you like your emoji?"
     enforce_square_emoji: "Force a square aspect ratio to all emojis."
+    emoji_trigger: "Minimum number of characters required to trigger autocomplete emoji popup"
```
SamSaffron: I think this name can be improved a bit `emoji_autocomplete_min_chars`  except for that, I know we don't have many tests to go by here, but I think it would be interesting to add one. Not going to block the PR on it, but if you can figure this out it would be very cool.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -434,6 +434,10 @@ export default Ember.Component.extend({
           const full = `:${term}`;
           term = term.toLowerCase();
 
+          if (term.length < Discourse.SiteSettings.emoji_trigger) {
```
ZogStriP: We try not to rely on globals anymore so we inject the site settings in most Ember object. You can access them directly with `this.siteSettings.emoji_trigger` ;)

udan11: Thanks!

config/locales/server.en.yml
udan11: I have been trying to figure out how to do the test.

```javascript
componentTest("minimum characters for emoji autocomplete == 0", {
  template: "{{d-editor value=value}}",

  beforeEach() {
    this.siteSettings.emoji_autocomplete_min_chars = 0;
  },

  async test(assert) {
    await this.set("value", ":");
    assert.equal(this.$(".autocomplete.ac-emoji").length, 1);
  }
});

componentTest("minimum characters for emoji autocomplete == 1", {
  template: "{{d-editor value=value}}",

  beforeEach() {
    this.siteSettings.emoji_autocomplete_min_chars = 1;
  },

  async test(assert) {
    await this.set("value", ":");
    assert.equal(this.$(".autocomplete.ac-emoji").length, 0);

    await this.set("value", ":s");
    assert.equal(this.$(".autocomplete.ac-emoji").length, 1);
  }
});
```

This is what I got, but for some reason it does not work. Do you have idea?

SamSaffron:
cool, thanks for this merging now.

SamSaffron:
Thanks heaps, merging this in!

PR #6141, opened 2018-07-23
FIX: Don't rely on setting data type read from database

gschlager:
Editing a site setting could fail when the datatype of a setting was changed and that setting was saved to the DB before the the datatype was changed. E.g. this did prevent changing the `default_locale` when the DB contained 7 (`enum`) and the app expected 1 (`string`).

I changed it so that it uses the datatype from the DB when reading the value, but otherwise always lets the Ruby calculate the right datatype.

@tgxworld Could you please take a quick look since you wrote most of the code around site settings?

### DIFF
--- a/lib/site_settings/type_supervisor.rb
+++ b/lib/site_settings/type_supervisor.rb
@@ -100,7 +100,8 @@ class SiteSettings::TypeSupervisor
 
   def to_rb_value(name, value, override_type = nil)
     name = name.to_sym
-    type = @types[name] = (override_type || @types[name] || get_data_type(name, value))
+    @types[name] = (@types[name] || get_data_type(name, value))
+    type = (override_type || @types[name])
 
     case type
     when self.class.types[:float]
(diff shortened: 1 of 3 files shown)


tgxworld:
> you wrote most of the code around site settings?

I actually reverted the commit so it all points to me :P

Anyway can we add a test for this?

gschlager:
I didn't look that closely at the history. 😉 
Sure, I'll add a test 🙂

tgxworld:
haha thank you!

SamSaffron:
I am getting very confused about the flow here, can you do a full on integration test the demos the problem and how this fixes it. Just insert a row in to site setting table with the bad data, and then show how this explodes and how the fix resolves it. 

I am a bit confused about where override is coming from.

gschlager:
I added another test.

The override comes from [`SiteSettingExtension.refresh!`](https://github.com/discourse/discourse/blob/master/lib/site_setting_extension.rb#L246-L248) where Discourse loads the settings stored in the DB and also uses the datatype stored in the DB as override. That's fine as long as the override is only used to correctly convert the DB value, but when it overrides the type within the supervisor it can cause problems like the one I've seen with `default_locale`.

PR #6143, opened 2018-07-23
FEATURE: Drafts view in user profile

pmusaraj:
Initial discussion: https://meta.discourse.org/t/drafts-view-in-profile/13720 

## Visible/functional changes:

- New "Drafts" tab available under Activity in the User Profile. 
- User can resume a draft, remove a draft.
- All of the user's drafts (topics, replies, PMs) are included in the list (and identified).

## Screenshots

drafts-screenshot

## Notes/changes

- I couldn't use the `expand-post` component in the draft list to expand/collapse the excerpts, since `expand-post` does an AJAX request to the post to get the full content, and drafts aren't yet posts. I don't think the drafts list warrants its own expand/collapse component, so the drafts can't be expanded/collapsed. 
- admins can only see a user's draft list. 
- previously, if a user had a draft PM, the composer would open as soon as she visited a `/u/username` route. In this PR, that behaviour is restricted to users visiting `/u/username/messages`
- I am not sure whether this feature requires qunit tests, please advise.


eviltrout:
Thanks for this! It's quite a big PR so it will take a little time for us to review.

pmusaraj:
Ok, I’ll go over it tonight again, I see some Travis failures, will try to fix them

SamSaffron:
Some initial feedback, the change looks good and well organised, I highly recommend adding some client side tests though afaik there are only server side tests here

SamSaffron:
also looks like linting is failing, can you set up prettier and eslint locally (for an added bonus can you confirm there is a topic on meta about this?)

app/assets/javascripts/discourse/models/user-draft.js.es6
```diff
@@ -0,0 +1,62 @@
...
+  NEW_TOPIC_KEY,
+  NEW_PRIVATE_MESSAGE_KEY
+} from "discourse/models/composer";
+
+const UserDraft = RestModel.extend({
+  @on("init")
+  _attachCategory() {
+    const categoryId = this.get("category_id");
+    if (categoryId) {
+      this.set("category", Discourse.Category.findById(categoryId));
```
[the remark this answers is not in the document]
pmusaraj: I tried to use `this.site.categories.findBy` here and where the instance is created (in `user-drafts-stream.js.es6`, but it didn't work. I couldn't find another way to do this...

app/assets/javascripts/discourse/routes/user-activity-drafts.js.es6
```diff
@@ -0,0 +1,26 @@
+export default Discourse.Route.extend({
+  model() {
+    return this.modelFor("user").get("userDraftsStream");
```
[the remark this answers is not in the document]
pmusaraj: Unfortunately, I don't understand how this would work as a promise. I tried a few things, but so far, no luck.

pmusaraj:
I added some client tests in the last few commits. 

I also added an action triggered during the composer's `destroyDraft` method: if user is in the drafts route, and if she then hits Cancel in the composer and "Yes, Abandon", the drafts list will be refreshed. I used 

````
this.get("application.currentRouteName") === 'userActivity.drafts'
````
to check whether the user is in the drafts route, but I didn't see this pattern used elsewhere, so I thought I'd highlight it, in case it's a no-no.

File: app/assets/javascripts/discourse/models/user-draft.js.es6 [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    }
+  },
+
+  @computed("draft_username")
+  editableDraft(draft_username) {
+    return draft_username === Discourse.User.currentProp("username");
```
tgxworld: `import User from "discourse/models/user";`

You can use the above to gain access to the `User` class instead of accessing it via the global.

app/assets/javascripts/discourse/models/user-draft.js.es6
```diff
@@ -0,0 +1,60 @@
...
+    return username.toLowerCase();
+  },
+
+  @computed("usernameLower")
+  userUrl(usernameLower) {
+    return userPath(usernameLower);
+  },
+
+  @computed("topic_id")
+  postUrl(topic_id) {
```
tgxworld: We prefer to use camelCase for Javascript variables.

app/assets/javascripts/discourse/components/user-stream.js.es6
```diff
@@ -57,6 +59,31 @@ export default Ember.Component.extend(LoadMore, {
...
 
+    resumeDraft(item) {
+      const composer = this.container.lookup("controller:composer");
+      if (composer.get("model.viewOpen")) {
+        composer.close();
+      }
+      if (item.get("postUrl")) {
+        DiscourseURL.routeTo(item.get("postUrl"));
+      } else {
+        Draft.get(item.draft_key).then(d => {
```
tgxworld: We need to handle errors here if the promise is rejected.

app/assets/javascripts/discourse/components/user-stream.js.es6
```diff
@@ -57,6 +59,31 @@ export default Ember.Component.extend(LoadMore, {
...
+            draftKey: item.draft_key,
+            draftSequence: d.draft_sequence
+          });
+        });
+      }
+    },
+
+    removeDraft(draft) {
+      const stream = this.get("stream");
+      Draft.clear(draft.draft_key, draft.sequence).then(() => {
```
tgxworld: We should handle errors here if the promise is rejected.

app/assets/javascripts/discourse/models/user-draft.js.es6
```diff
@@ -0,0 +1,60 @@
...
+  @on("init")
+  _attachCategory() {
+    const categoryId = this.get("category_id");
+    if (categoryId) {
+      this.set("category", Discourse.Category.findById(categoryId));
+    }
+  },
+
+  @computed("draft_username")
+  editableDraft(draft_username) {
```
tgxworld: camelCase variable name :grin:

app/assets/javascripts/discourse/models/user-draft.js.es6
```diff
@@ -0,0 +1,60 @@
...
+    return draft_username === Discourse.User.currentProp("username");
+  },
+
+  @computed("username")
+  usernameLower(username) {
+    return username.toLowerCase();
```
tgxworld: We have a `username_lower` column on the `users` table, can that be used somehow?

app/assets/javascripts/discourse/models/user-drafts-stream.js.es6
```diff
@@ -0,0 +1,99 @@
...
+      return Ember.RSVP.resolve();
+    }
+
+    if (this.get("loading")) {
+      return Ember.RSVP.resolve();
+    }
+
+    this.set("loading", true);
+
+    return ajax(findUrl, { cache: "false" })
```
tgxworld: We need to handle errors here as well.

app/assets/javascripts/discourse/templates/components/user-stream.hbs
```diff
@@ -1,3 +1,7 @@
 {{#each stream.content as |item|}}
-  {{user-stream-item item=item removeBookmark=(action "removeBookmark")}}
+  {{user-stream-item  item=item
```
tgxworld: Small nitpick but there are 2 spaces between `user-stream-item` and `item=item`

app/controllers/drafts_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+             user: user,
+             offset: params[:offset].to_i,
+             limit: per_chunk }
+
+    guardian.ensure_can_see_drafts!(user)
+
+    stream = Draft.stream(opts)
+
+    stream.each do |d|
+      parsedData = JSON.parse(d.data)
```
tgxworld: On the ruby side, we use snake_case instead of camelCase

app/controllers/drafts_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      d.raw = parsedData['reply']
+      if parsedData['categoryId'].present? && !d.category_id.present?
+        d.category_id = parsedData['categoryId']
+      end
+    end
+
+    stream = stream.to_a
+
+    if stream.length == 0
```
tgxworld: I don't feel like this logic should be present on the server side. The controller should only be responsibly in generating a standard set of payload and let the client side decides how it should display the drafts based on the different conditions.

app/models/draft.rb
```diff
@@ -43,6 +43,43 @@ def self.find_draft(user, key)
     end
   end
 
+  def self.stream(opts = nil)
+    opts ||= {}
+
+    user_id = opts[:user_id]
+    offset = opts[:offset] || 0
+    limit = opts[:limit] || 30
```
tgxworld: It looks like the default chunk size is duplicated here and in the controller. I think we should only have one location for this logic.

app/serializers/draft_serializer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def include_slug?
+    object.title.present?
+  end
+
+  def closed
+    object.topic_closed
+  end
+
+  def include_slug?
```
tgxworld: This method is duplicated twice.

app/controllers/drafts_controller.rb
```diff
@@ -0,0 +1,49 @@
+class DraftsController < ApplicationController
+  requires_login
+
+  skip_before_action :check_xhr, :preload_json
+
+  def index
```
tgxworld: We should add a request spec for this :)

lib/guardian/user_guardian.rb
```diff
@@ -30,6 +30,10 @@ def can_see_notifications?(user)
     is_me?(user) || is_admin?
   end
 
+  def can_see_drafts?(user)
+    is_me?(user) || is_admin?
```
tgxworld: I think we need a spec for this as well :) Never know when we may regress.

test/javascripts/models/user-drafts-test.js.es6
```diff
@@ -0,0 +1,31 @@
+import UserDraft from "discourse/models/user-draft";
+import { NEW_TOPIC_KEY } from "discourse/models/composer";
+
+QUnit.module("model:user-drafts");
+
+QUnit.test("stream", assert => {
+  var user = Discourse.User.create({ id: 1, username: "eviltrout" });
```
tgxworld: We generally prefer to use `let` over `var` as the variable will then be limited to the scope that it was declared in.

app/assets/javascripts/discourse/models/user-draft.js.es6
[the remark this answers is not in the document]
eviltrout: I think the only solution is to pass in `site: this.site` when creating your model. It looks like `UserDraft` is created in `UserDraftStream` so you probably have to pass `this.site` into `UserDraftStream` when it is created.

app/assets/javascripts/discourse/routes/user-activity-drafts.js.es6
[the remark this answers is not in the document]
eviltrout: You could try something like:

```
let userDraftsStream = this.modelFor('user').get('userDraftsStream')
return userDraftsStream.load().then(() => userDraftsStream);
```

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -874,7 +874,14 @@ export default Ember.Controller.extend({
       if (key === "new_topic") {
         this.send("clearTopicDraft");
       }
-      Draft.clear(key, this.get("model.draftSequence"));
+
+      Draft.clear(key, this.get("model.draftSequence")).then(() => {
+        if (
+          this.get("application.currentRouteName") === "userActivity.drafts"
```
eviltrout: @pmusaraj I do believe that checking the route name is the wrong approach there. My main issue with it is it is putting code that belongs in another route in this controller.

What I would recommend is to use an appEvent. You can trigger something like `this.appEvents.trigger('draft:destroyed', key)`

Then in the userActivity.drafts route, listen for the app event and call refresh.

app/assets/javascripts/discourse/routes/user-activity-drafts.js.es6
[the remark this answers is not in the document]
pmusaraj: Thanks, that works well.

test/javascripts/models/user-drafts-test.js.es6
jjaffeux: in this case it should even be a `const` doesn’t seem mutated later.

lib/guardian/user_guardian.rb
pmusaraj: @tgxworld do you have a example of what this spec should look like?

app/controllers/drafts_controller.rb
pmusaraj: I added something for this in `draft_controller_spec.rb`, let me if it's insufficient @tgxworld.

lib/guardian/user_guardian.rb
tgxworld: Something like https://github.com/discourse/discourse/blob/master/spec/components/guardian_spec.rb#L38-L42

spec/requests/draft_controller_spec.rb
```diff
@@ -20,4 +20,21 @@
     expect(response.status).to eq(200)
     expect(Draft.get(user, 'xxx', 0)).to eq(nil)
   end
+
+end
+
+describe DraftsController do
```
tgxworld: This has been declared above in the file. New assertions should go under a new `describe` block

```
describe "#drafts" do
  it "..." do
    assertions
  end

  it "..." do
    assertions
  end
end
```

tgxworld: This has been declared above in the file. New assertions should go under a new `describe` block

```
describe "#drafts" do
  it "..." do
    assertions
  end

  it "..." do
    assertions
  end
end
```

app/controllers/drafts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    opts = {
+      user: user,
+      offset: params[:offset],
+      limit: params[:limit]
+    }
+
+    guardian.ensure_can_see_drafts!(user)
+    stream = Draft.stream(opts)
+    stream.each do |d|
+      parsed_data = JSON.parse(d.data) rescue nil
```
tgxworld: `rescue nil` is an anti-pattern that we avoid since any and every error is rescued and thrown away. We prefer to rescue the specific error instead.

config/routes.rb
```diff
@@ -727,6 +727,7 @@
 
   get "message-bus/poll" => "message_bus#poll"
 
+  resources :drafts
```
tgxworld: This should be `resources :drafts, only: [:index]` so that we don't generate the other routes that we don't require.

spec/requests/draft_controller_spec.rb
pmusaraj: What's described above is `DraftController` singular, this is the plural. But fair enough, it should be renamed to `#drafts` to avoid exactly this kind of confusion.

app/assets/javascripts/discourse/models/user-drafts-stream.js.es6
pmusaraj: I tried adding an error handler here, but I couldn't figure out how to test it, how to trigger it. Plus, similar models ([user-stream](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/user-stream.js.es6), [user-posts-stream](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/user-posts-stream.js.es6)) don't have error handling for similar calls.

app/controllers/drafts_controller.rb
eviltrout: @tgxworld  thoughts?

spec/requests/draft_controller_spec.rb
tgxworld: O wait is that a different controller or is it just a different action? If it is a different controller it should go into a separate spec file.

pmusaraj: It's a different controller, I'll split it out.

pmusaraj:
@tgxworld if you're looking at this, I don't understand why Travis tests fail... Is it because of something in the PR or because of something else?

tgxworld:
@pmusaraj Can you rebase your commits against master? It looks like something might have broken the Travis test suite. Also, one of your qunit test is failing.

pmusaraj:
Done @tgxworld but now Travis fails on `bundle exec danger`.

gschlager:
danger had problems editing the old comment, because it was created by a different bot. I deleted the comment and restarted the Travis build. Now everything's green. 💚

SamSaffron:
🎊 🍸 👍  merging this in right now!!!

PR #6146, opened 2018-07-23
UX: Display only top categories in hamburger menu

vinothkannans:


app/serializers/current_user_serializer.rb
```diff
@@ -153,9 +157,23 @@ def custom_fields
...
+  def watched_category_ids
+    CategoryUser.lookup(object, :watching).pluck(:category_id)
+  end
+
+  def watched_first_post_category_ids
+    CategoryUser.lookup(object, :watching_first_post).pluck(:category_id)
+  end
+
+  def top_category_ids
+    UserSummary.new(object, Guardian.new).top_categories.pluck(:id)
```
SamSaffron: how expensive is this query? Just curious....

SamSaffron: also... instead of 4 queries to look up category user data here, just do one

SamSaffron: might as well add this to the serializer test as well

vinothkannans: @SamSaffron it may bit costly. `top_categories` method calling db 4 times with select queries _(I hope it can be reduced to 3)_. Since we are using this in two places I like to add `topic_count` and `post_count` columns to `category_users` table if possible.

And that method called by every user summary page visit too.

SamSaffron: why cant this just be done in 1 query?

SamSaffron:
My vote here is just to hold off here till we have a proper cache table for user_id,category_id,count that we keep up to date and synchronize in case of drift daily/weekly 

Its way too much work to be doing on first hit to the site and can get extremely slow for super active users. 

For me this would add 150ms to my initial payload which is way too much.

vinothkannans:
@SamSaffron the changes are done.

test/javascripts/widgets/hamburger-menu-test.js.es6
```diff
@@ -212,7 +215,10 @@ widgetTest("top categories", {
 
   test(assert) {
     assert.equal(this.$(".category-link").length, maxCategoriesToDisplay);
-    assert.equal(this.$(".category-link .category-name").text(), "parent 4parent 5child 2parent 3childparent");
+    assert.equal(
+      this.$(".category-link .category-name").text(),
```
jjaffeux: I think we prefer to use `find` instead of `this.$`

app/assets/stylesheets/common/base/menu-panel.scss
```diff
@@ -108,6 +108,15 @@
     }
   }
 
+  ul.category-links {
```
jjaffeux: do we need `ul` and `li` here ? it's adding specificity which might make styling harder at some point

vinothkannans:
@jjaffeux the requested changes are done.

app/serializers/current_user_serializer.rb
```diff
@@ -153,9 +156,26 @@ def custom_fields
...
-      .pluck(:category_id)
+    CategoryUser.lookup(object, :muted).pluck(:category_id)
+  end
+
+  def top_category_ids
+    categories = CategoryUser.where(user_id: object.id)
+      .where.not(notification_level: CategoryUser.notification_levels[:muted])
+      .select(:category_id, :notification_level)
+
+    categories.sort_by do |c|
```
SamSaffron: this is 1 query

app/serializers/current_user_serializer.rb
```diff
@@ -153,9 +156,26 @@ def custom_fields
...
+      when CategoryUser.notification_levels[:watching]
+        1
+      when CategoryUser.notification_levels[:tracking]
+        2
+      when CategoryUser.notification_levels[:watching_first_post]
+        3
+      end
+    end
+
+    categories.pluck(:category_id).slice(0, MAX_TOP_CATEGORIES_COUNT)
```
SamSaffron: this is a second query... 

instead what you do is order by a case statement in SQL no need to pull the data down

app/assets/javascripts/discourse/widgets/hamburger-categories.js.es6
```diff
@@ -66,8 +66,19 @@ export default createWidget("hamburger-categories", {
     if (categories.length === 0) {
       return;
     }
-    return result.concat(
+    result = result.concat(
       categories.map(c => this.attach("hamburger-category", c))
     );
+
+    return result.concat(
```
SamSaffron: technically you only want the more link if the site has more than 6 categories.

app/assets/javascripts/discourse/widgets/hamburger-menu.js.es6
```diff
@@ -176,18 +176,26 @@ export default createWidget("hamburger-menu", {
   },
 
   listCategories() {
-    const hideUncategorized = !this.siteSettings.allow_uncategorized_topics;
-    const isStaff = Discourse.User.currentProp("staff");
+    const maxCategoriesToDisplay = 6;
+    const categoriesList = this.site.get("categoriesList");
```
SamSaffron: I don't think this works quite right ... you first need to get all the categories ordered by count (in some conditions) and then do the filtering down. 

This change makes the list of categories we show arbitrary afaik, especially since we suppress subcategories.

vinothkannans:
@SamSaffron i did the changes requested.

test/javascripts/widgets/hamburger-menu-test.js.es6
```diff
@@ -125,42 +128,40 @@ widgetTest("general links", {
...
 
-  beforeEach() {
-    const cat = this.site.get("categoriesList")[0];
-
-    const parent = Discourse.Category.create({
-      id: 1,
-      topic_count: 5,
-      name: "parent",
-      url: "https://test.com/parent",
-      show_subcategory_list: true,
```
SamSaffron: this option is confusing, can you debug and figure out when we currently show subcategories in the hamburger and when we do not?

vinothkannans: Currently we hiding subcategories [if it's parent category setting `show_subcategory_list` is enabled](https://github.com/discourse/discourse/blob/03a7d532cf8f09b12573b21ef013c21100d52728/app/assets/javascripts/discourse/widgets/hamburger-menu.js.es6#L183-L185). Even though it's not related to the hamburger menu [we using it](https://meta.discourse.org/t/how-to-hide-sub-categories-in-hamburger-menu/58835/5?u=vinothkannans) to reduce the noise in hamburger. Instead we showcasing it in the parent category's landing page (in boxes/rows).

PR #6166, opened 2018-07-24
FIX: Remove `plugin.enabled?` checks at initialisation time

davidtaylorhq:
Based on an issue pointed out in [this topic](https://meta.discourse.org/t/the-staff-notes-plugin-no-longer-displays-the-little-note-next-to-a-username-after-upgrading/92566/5?u=david_taylor).

Checking `plugin.enabled?` while initialising plugins causes issues in two ways:
  - An application restart is required for changes to take effect. A load-balanced multi-server environment could behave very weirdly if containers restart at different times.
  - In a multisite environment, it takes the `enabled?` setting from the default site. Changes on that site affect all other sites in the cluster.

Instead, `plugin.enabled?` should be checked at runtime, in the context of a request. This commit removes `plugin.enabled?` from many `instance.rb` methods.

I have added a working `plugin.enabled?` implementation for methods that actually affect security/functionality:
  - `post_custom_fields_whitelist`
  - `whitelist_staff_user_custom_field`
  - `add_permitted_post_create_param`

There are no changes to the plugin API interface.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -615,8 +615,8 @@ class PostsController < ApplicationController
       :visible
     ]
 
-    if Post.permitted_create_params.present?
-      permitted.concat(Post.permitted_create_params.to_a)
+    Post.plugin_permitted_create_params.each do |key, plugin|
+      permitted << key if plugin.enabled?
     end
 
     # param munging for WordPress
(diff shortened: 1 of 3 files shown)


SamSaffron:
@eviltrout leaving this one to you to decide on, I think we have no choice here but to do this

eviltrout:
I think this is the correct solution, thank you.

PR #6188, opened 2018-07-26
FIX: forces VPatch.node with svg shadow-root in virtual-dom

jjaffeux:


eviltrout:
Should we add a test to our app to make sure this doesn't regress? I doubt we'll update virtual-dom, but if someone didn't know you patched it in our vendor area it could break again.

jjaffeux:
I can try to write a test for it, not sure yet how to do it though.

SamSaffron:
what is the call on this patch? when can we merge it?

jjaffeux:
waiting for subseting

PR #6197, opened 2018-07-27
Create transliterate.ar.yml

mtawil:
Add a transliteration rules for Arabic


SamSaffron:
this looks right to me, merging it in!  @OsamaSayegh  can you also review (and tweak if you see fit)

PR #6198, opened 2018-07-27
Add a display name to user profile

mtawil:


SamSaffron:
I guess if there is no name we got to use username... thanks!

PR #6199, opened 2018-07-30
FEATURE: backend support for user-selectable components

OsamaSayegh:


app/assets/javascripts/admin/controllers/admin-customize-themes-show.js.es6
```diff
@@ -9,6 +9,12 @@ const THEME_UPLOAD_VAR = 2;
 export default Ember.Controller.extend({
   editRouteName: "adminCustomizeThemes.edit",
 
+  @computed("selectableChildThemes")
+  selectedChildThemeId() {
+    const available = this.get("selectableChildThemes");
+    return available && available[0] ? available[0].get("id") : null;
```
SamSaffron: this is confusing code, why is it picking first?

app/assets/javascripts/admin/controllers/admin-customize-themes-show.js.es6
```diff
@@ -64,16 +70,20 @@ export default Ember.Controller.extend({
 
     let themes = [];
     available.forEach(t => {
-      if (!childThemes || childThemes.indexOf(t) === -1) {
+      if (
+        (!childThemes || childThemes.indexOf(t) === -1) &&
+        _.isEmpty(t.get("childThemes")) &&
```
SamSaffron: we are looking to get rid of underscore soonish, perhaps

https://emberjs.com/api/api/ember/release/functions/@ember%2Futils/isEmpty

app/controllers/application_controller.rb
```diff
@@ -331,28 +331,33 @@ def handle_theme
     resolve_safe_mode
     return if request.env[NO_CUSTOM]
 
-    theme_id = request[:preview_theme_id]&.to_i
+    theme_ids = []
+
+    if preview_key = request[:preview_theme_id]&.to_i
```
SamSaffron: why is this called preview_key? preview_theme_id ? also ... how are you going to preview user selectable theme components? something to think about.

app/models/concerns/preserve_order.rb
```diff
@@ -0,0 +1,13 @@
+module PreserveOrder
```
SamSaffron: I get the whole concern thing and code reuse and all, but find the the indirection here just ends up confusing things, I would just duplicate this twice.

Also the way it takes in a Hash and only looks at the first element feels risky to me.

SamSaffron:
hmmm what happens with components post this change when you try to add a child, does the UI still allow you to do that and error out? We probably want to improve the UI here.

I think its getting quite urgent to do the proper component/theme split

overall the PR looks good, but its a bit big for me to know for sure if something is a problem vs not a problem, we will need careful testing

app/assets/javascripts/admin/controllers/admin-customize-themes-show.js.es6
OsamaSayegh: This is a fix for this bug: https://meta.discourse.org/t/86399

The repro for this bug is to turn on "Theme includes other child themes" on a theme, you'll see the dropdown menu for components has the first item selected. Now if you don't open the dropdown menu and select a component from it, but instead you click Add right away, you'll see this bug.

`selectedChildThemeId` is controlled by the dropdown menu, and it's undefined until you select something from the dropdown menu which is causing this bug. I added this to set an initial value to the property.

I think there is a clearer way to fix this bug, let me try it.

app/controllers/application_controller.rb
OsamaSayegh: Weird, not sure where that came from. I'll fix it.

I can think of 2 ways for previewing user-selectable components 1) easiest is to preview CSS/HTML only with default color scheme or 2) is to allow previewing in context of a parent theme so that you can see how it looks with color schemes of different parents.

OsamaSayegh:
I've made sure the UI filters out all themes that would cause errors when added as children. So it shouldn't cause problems.

I think the main thing we need to be extra mega careful with is the migration code. I've tested it with 17 themes with ~100 child-parent relations and it worked well. Let me know if there is a scenario that I need to verify it doesn't break migration or messes up data.

app/models/theme_field.rb
```diff
@@ -4,6 +4,13 @@ class ThemeField < ActiveRecord::Base
 
   belongs_to :upload
 
+  scope :where_ordered, ->(hash) {
+    return none unless hash.present?
+
+    column, values = hash.first # order by the first element
+    where(hash).order("position(#{column}::text in '#{values.join(',')}')")
```
SamSaffron: this feels SQL injectiony to me, since we always order by theme_ids, why make this so generic? 

Also ... hash is always present in the callers

lib/stylesheet/manager.rb
```diff
@@ -19,38 +20,61 @@ def self.clear_theme_cache!
     cache.hash.keys.select { |k| k =~ /theme/ }.each { |k|cache.delete(k) }
   end
 
-  def self.stylesheet_href(target = :desktop, theme_id = :missing)
-    href = stylesheet_link_tag(target, 'all', theme_id)
-    if href
-      href.split(/["']/)[1]
+  def self.stylesheet_data(target = :desktop, theme_ids = :missing)
+    stylesheets = stylesheet_details(target, "all", theme_ids)
```
SamSaffron: cant this whole array be frozen and cached?

OsamaSayegh:
@SamSaffron I've updated this PR, the array is now frozen and cached and there should no SQL injection risks anymore.

Note: it seems `order("position(theme_id::text in '5,9,7,4')")` doesn't always work as expected :( so I've changed it to something like this:
https://github.com/discourse/discourse/blob/0b3d51a8bc43391b3f64719c6606ff3298128824/app/models/theme.rb#L236-L244

SamSaffron:
tricky, sure let's aim to get this merged on Monday ... will also give @tgxworld  a bit of time to review as well.

app/assets/javascripts/admin/models/theme.js.es6
```diff
@@ -150,7 +151,9 @@ const Theme = RestModel.extend({
 
   saveChanges() {
     const hash = this.getProperties.apply(this, arguments);
-    return this.save(hash).then(() => this.set("changed", false));
+    return this.save(hash)
+      .finally(() => this.set("changed", false))
+      .catch(popupAjaxError);
```
tgxworld: :+1:

app/assets/javascripts/discourse/initializers/live-development.js.es6
```diff
@@ -58,12 +58,16 @@ export default {
...
           $("link").each(function() {
             if (me.hasOwnProperty("theme_id") && me.new_href) {
-              let target = $(this).data("target");
-              if (me.theme_id === themeId && target === me.target) {
+              const target = $(this).data("target");
+              const themeId = $(this).data("theme-id");
+              if (
+                themeIds.indexOf(me.theme_id) !== -1 &&
+                target === me.target &&
+                (!themeId || (themeId && themeId === me.theme_id))
```
tgxworld: We should be able to shorten this to just `(!themeId || (themeId === me.theme_id))`

app/assets/javascripts/discourse/lib/theme-selector.js.es6
```diff
@@ -12,21 +12,25 @@ export function currentThemeKey() {
...
-    if (_.isEmpty(themeId)) {
-      themeId = null;
-    } else {
-      themeId = parseInt(themeId);
-    }
+    elem.content
+      .split(",")
+      .map(num => parseInt(num, 10))
+      .filter(num => !isNaN(num))
+      .forEach(num => themeIds.push(num));
```
tgxworld: I feel like we have too many loops here. Perhaps something like the following would work?

```
elem.content.split(",").forEach(num => {
  num = parseInt(num, 10);
  if (!isNaN(num)) { themeIds.push(num) };
})
```

app/controllers/admin/themes_controller.rb
```diff
@@ -182,11 +182,13 @@ def update
         log_theme_change(original_json, @theme)
         format.json { render json: @theme, status: :ok }
       else
-        format.json {
+        format.json do
+          error = @theme.errors.full_messages.join(", ").presence
+          error = I18n.t("themes.bad_color_scheme") if @theme.errors[:color_scheme].present?
+          error ||= I18n.t("themes.other_error")
```
tgxworld: Looks like we're missing a spec for the changes made here.

app/controllers/application_controller.rb
```diff
@@ -62,8 +62,8 @@ def handle_unverified_request
     after_action :remember_theme_id
 
     def remember_theme_id
-      if @theme_id
-        Stylesheet::Watcher.theme_id = @theme_id if defined? Stylesheet::Watcher
+      if @theme_ids.present?
+        Stylesheet::Watcher.theme_id = @theme_ids&.first if defined? Stylesheet::Watcher
```
tgxworld: If `@theme_ids.present?` is true, we don't have to use `&.` when calling `.first`

app/controllers/application_controller.rb
```diff
@@ -331,28 +331,33 @@ def handle_theme
...
+        theme_ids = ids if guardian.allow_themes?(ids)
       end
     end
 
-    theme_id ||= user_option&.theme_ids&.first
+    theme_ids = user_option&.theme_ids || [] if theme_ids.blank?
 
-    if theme_id && !guardian.allow_themes?(theme_id)
-      theme_id = nil
+    unless guardian.allow_themes?(theme_ids)
```
tgxworld: It looks like we're checking this twice , once on line 346 and another time here.

app/controllers/themes_controller.rb
```diff
@@ -1,27 +1,26 @@
 class ThemesController < ::ApplicationController
   def assets
-    theme_id = params[:id].to_i
+    theme_ids = params[:ids].to_s.split("-").map(&:to_i)
 
-    if params[:id] == "default"
-      theme_id = nil
+    if params[:ids] == "default"
+      theme_ids = [nil]
```
tgxworld: I was wondering if we can just use a blank array here. That feels like the better representation of the data here.

app/controllers/themes_controller.rb
```diff
@@ -1,27 +1,26 @@
...
+      Stylesheet::Manager.stylesheet_data(target, theme_ids).map do |hash|
+        return hash unless Rails.env.development?
+
+        dup_hash = hash.dup
+        dup_hash[:new_href] << (dup_hash[:new_href].include?("?") ? "&" : "?")
+        dup_hash[:new_href] << SecureRandom.hex
+        dup_hash
       end
-    end.compact
+    end.flatten.freeze
```
tgxworld: Is there a reason we need to freeze the object here?

app/models/child_theme.rb
```diff
@@ -1,6 +1,24 @@
 class ChildTheme < ActiveRecord::Base
   belongs_to :parent_theme, class_name: 'Theme'
   belongs_to :child_theme, class_name: 'Theme'
+
+  validate :child_validations
+
+  private
+
+  def child_validations
+    if ChildTheme.exists?(["parent_theme_id = ? OR child_theme_id = ?", child_theme_id, parent_theme_id])
```
tgxworld: I may be wrong here but I think this should be a uniqueness validation on `parent_theme_id` and `child_theme_id`?

app/models/child_theme.rb
```diff
@@ -1,6 +1,24 @@
...
+      errors.add(:base, I18n.t("themes.errors.no_multilevels_components"))
+    end
+
+    if Theme.exists?(id: child_theme_id, user_selectable: true)
```
tgxworld: Same as https://github.com/discourse/discourse/pull/6199/files#r207779563

app/models/theme.rb
```diff
@@ -20,6 +20,16 @@ class Theme < ActiveRecord::Base
   has_many :color_schemes
   belongs_to :remote_theme
 
+  validate :user_selectable_validation
+
+  scope :parent_themes, ->() {
+    where("id NOT IN (SELECT child_theme_id FROM child_themes)")
```
tgxworld: `NOT IN` or any variations of `IN` query will be bad for performance once the result set is large. What you can do here is use a `LEFT JOIN child_themes ON themes.id = child_themes_id WHERE themes.id IS NULL`

app/models/theme.rb
```diff
@@ -70,25 +79,32 @@ def notify_color_change(color)
   end
 
   after_commit ->(theme) do
-    theme.notify_theme_change
-  end, on: :update
+    theme.notify_theme_change(with_scheme: theme.saved_change_to_color_scheme_id?)
+  end, on: [:create, :update]
+
+  def self.get_set_cache(key, &blk)
```
tgxworld: Does the key for the cache need to include the current database name? If we're using the same key, the cache might bleed across multisite.

app/models/theme.rb
```diff
@@ -141,33 +185,28 @@ def self.lookup_target(target_id)
...
 
-    message = themes.map do |theme|
-      refresh_message_for_targets([:mobile_theme, :desktop_theme], theme)
-    end.compact.flatten
     MessageBus.publish('/file-change', message)
   end
 
-  def refresh_message_for_targets(targets, theme)
+  def notify_theme_change(with_scheme: false)
+    theme_ids = [self.id] + (dependant_themes&.pluck(:id) || [])
```
tgxworld: This is creating 3 arrays here. `(dependant_themes&.pluck(:id) || []).unshift(self.id)` will allow us to only create one.

app/models/theme.rb
```diff
@@ -193,58 +229,37 @@ def resolve_dependant_themes(direction)
...
-    uniq << id
-
-    iterations = 0
-    added = [id]
-
-    while added.length > 0 && iterations < 5
-
-      iterations += 1
+    Theme.where(
+      "id IN (SELECT #{select_field}
```
tgxworld: If possible a left join here would be better.

app/models/theme.rb
```diff
@@ -288,21 +303,27 @@ def set_field(target:, name:, value: nil, type: nil, type_id: nil, upload_id: ni
...
   def all_theme_variables
     fields = {}
-    ([self] + (included_themes || [])).each do |theme|
-      theme&.theme_fields.each do |field|
-        next unless ThemeField.theme_var_type_ids.include?(field.type_id)
-        next if fields.key?(field.name)
-        fields[field.name] = field
-      end
+    ids = [self.id] + (included_themes&.map(&:id) || [])
+    _fields = ThemeField
```
tgxworld: Hmm an underscore in front of a variable name usually means the variable is not being used. However, it seems like this is an actual variable which we want to use here.

app/models/theme.rb
```diff
@@ -288,21 +303,27 @@ def set_field(target:, name:, value: nil, type: nil, type_id: nil, upload_id: ni
...
+      .find_by_theme_ids(ids)
+      .where(type_id: ThemeField.theme_var_type_ids)
+
+    _fields.each do |field|
```
tgxworld: It looks like we don't need the variable `_fields` here.

lib/guardian.rb
```diff
@@ -359,9 +359,17 @@ def can_export_entity?(entity)
...
+      return true
+    end
+
+    parent = theme_ids.first
+    components = theme_ids[1..-1] || []
+
+    parent_allowed = Theme.user_theme_ids.include?(parent)
+    children_allowed = (components - Theme.components_for(parent)).empty?
+
+    parent_allowed && children_allowed
```
tgxworld:
```
Theme.user_theme_ids.include?(parent) &&
  (components - Theme.components_for(parent)).empty?
```
I think we should avoid the variables so that we can take advantage of short circuit evaluation.

lib/stylesheet/manager.rb
```diff
@@ -271,7 +299,24 @@ def plugins_digest
   end
 
   def settings_digest
-    Digest::SHA1.hexdigest((theme&.included_settings || {}).to_json)
+    timestamps = ""
+    proc = Proc.new do |&blck|
```
tgxworld: I don't feel like the code here justifies the use of a `Proc`

```
timestamps = ThemeField.where("(name = ? AND type_id = ? AND theme_id = ?) OR (theme_id = ?) ORDER BY updated_at ASC").pluck(:updated_at).map!(&:to_f).join(",")
```

Perhaps something like the above?

spec/models/theme_spec.rb
```diff
@@ -68,14 +68,23 @@
...
+
+  it "doesn't allow multi-level theme components" do
+    grandchild = Theme.create!(name: 'grandchild', user_id: user.id)
+    child = Theme.create!(name: 'child', user_id: user.id)
+    theme = Theme.create!(name: 'theme', user_id: user.id)
+
+    theme.add_child_theme!(child)
+    expect do
+      child.add_child_theme!(grandchild)
+    end.to raise_error(Discourse::InvalidParameters)
```
tgxworld: We should validate for the error message here as well. Note that `raise_error` takes the expected error message as the second parameter.

db/migrate/20180710172959_disallow_multi_levels_theme_components.rb
```diff
@@ -0,0 +1,66 @@
+def already_exists?(parent, child)
+  DB.query("SELECT 1 AS one FROM child_themes WHERE child_theme_id = :child AND parent_theme_id = :parent", child: child, parent: parent).present?
+end
+
+class DisallowMultiLevelsThemeComponents < ActiveRecord::Migration[5.2]
+  def up
+    handeled = []
+    proc = Proc.new do |parent, top_parent|
```
tgxworld: Is there a reason we use a proc here instead of defining a method? Seems abit odd to me.

db/migrate/20180710172959_disallow_multi_levels_theme_components.rb
```diff
@@ -0,0 +1,66 @@
...
+      SELECT parent_theme_id, child_theme_id
+      FROM child_themes
+      WHERE parent_theme_id NOT IN (SELECT child_theme_id FROM child_themes)
+    ")
+
+    top_parents.each do |top_parent|
+      proc.call(top_parent, top_parent)
+    end
+
+    if handeled.size > 0
```
tgxworld: I think there is a typo with `handled` here.

tgxworld:
@OsamaSayegh I personally feel like we're very thin on the tests in this PR. Alot of code has change/has been added but only a couple of specs were updated which makes it harder to follow what is going on here. Perhaps you can take another pass though the code and update them with the relevant test cases?

spec/models/theme_spec.rb
```diff
@@ -265,24 +274,24 @@ def transpile(html)
 
     theme = Theme.create!(name: "bob", user_id: -1)
 
-    expect(Theme.theme_ids).to eq(Set.new([theme.id]))
-    expect(Theme.user_theme_ids).to eq(Set.new([]))
+    expect(Theme.theme_ids).to eq([theme.id])
```
tgxworld: This test is not ensuring that it works for multiple `theme_ids`

app/controllers/application_controller.rb
OsamaSayegh: Yeah we need to do that twice. The first one to check the user can use the themes specified in the `theme_ids` cookie, and this is to check the user can use themes in the `user_options` table.

app/models/child_theme.rb
OsamaSayegh: Hmmm what I want to do here is to make sure the new record's `child_theme_id` value doesn't exist in `parent_theme_id` column and vice versa. In other words, we don't want to allow children to become parents, or parents become children. It's OK if a theme is parent of multiple themes and a theme is a child for multiple parents.

app/models/child_theme.rb
OsamaSayegh: I'm not sure I understand here. What I want to do here is make sure the child theme is not user selectable. How do I do that with uniqueness validation?

OsamaSayegh:
@tgxworld I've added a bunch of tests and made the changes you requested. Let me know what you think ⚡️

app/models/child_theme.rb
tgxworld: Oops ignore me here

app/models/child_theme.rb
tgxworld: Oops ignore me here

spec/models/child_theme_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+describe ChildTheme do
+  it "doesn't allow a user selectable theme to be a child" do
+    parent = Theme.create!(name: "parent", user_id: -1)
+    selectable_theme = Theme.create!(name: "selectable", user_id: -1, user_selectable: true)
+
+    relation = ChildTheme.new(parent_theme: parent, child_theme: selectable_theme)
```
tgxworld: I'm not sure if `relation` is the right variable name here as the object is an ActiveRecord object. I would just name it `child_theme`

app/models/theme.rb
```diff
@@ -20,6 +20,16 @@ class Theme < ActiveRecord::Base
   has_many :color_schemes
   belongs_to :remote_theme
 
+  validate :user_selectable_validation
+
+  scope :parent_themes, ->() {
```
tgxworld: This scope doesn't seem be used anywhere?

spec/requests/application_controller_spec.rb
```diff
@@ -33,4 +33,67 @@
...
+
+    before do
+      sign_in(user)
+    end
+
+    it "selects the theme the user has selected" do
+      user.user_option.update_columns(theme_ids: [theme.id])
+
+      get "/"
+      expect(controller.theme_ids).to eq([theme.id])
```
tgxworld: We need to test for the `response.code` whenever we make a request as it makes it easier to debug should the test fails

spec/requests/application_controller_spec.rb
```diff
@@ -33,4 +33,67 @@
...
+
+      theme.update_attribute(:user_selectable, false)
+
+      get "/"
+      expect(controller.theme_ids).not_to eq([theme.id])
```
tgxworld: We almost always want to avoid using `not_to eq()` because the result value can be anything and the test will still pass. Instead, we want the assertion to be tested against the actual value that we expect.

spec/requests/application_controller_spec.rb
```diff
@@ -33,4 +33,67 @@
...
+
+      get "/"
+      expect(controller.theme_ids).to eq([theme2.id])
+    end
+
+    it "can be overridden with preview_theme_id param" do
+      sign_in(admin)
+      cookies['theme_ids'] = "#{theme.id},#{theme2.id}|#{admin.user_option.theme_key_seq}"
+
+      get "/?preview_theme_id=#{theme2.id}"
```
tgxworld: This is OK but it'll be better to just pass the parameters via `params`

spec/models/child_theme_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+describe ChildTheme do
+  it "doesn't allow a user selectable theme to be a child" do
```
tgxworld: We usually wrap validations test in a `describe "validations"` block

spec/models/child_theme_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    parent = Theme.create!(name: "parent", user_id: -1)
```
tgxworld: It seems like we added a fabricator for `theme`, would be better to use that instead.

spec/models/child_theme_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+describe ChildTheme do
```
tgxworld: It looks like we're missing a test case for the `I18n.t("themes.errors.no_multilevels_components")` error

spec/models/theme_spec.rb
```diff
@@ -68,14 +68,50 @@
...
+  end
+
+  it "doesn't allow a child to be user selectable" do
+    child = Theme.create!(name: 'child', user: user)
+    theme = Theme.create!(name: 'theme', user: user)
+
+    theme.add_child_theme!(child)
+    begin
+      child.update!(user_selectable: true)
+    rescue ActiveRecord::RecordInvalid => e
```
tgxworld: We don't need to `rescue` the error here as we can use `update` instead of `update!`

spec/models/theme_spec.rb
```diff
@@ -116,6 +152,33 @@
...
+    theme2.set_field(target: :common, name: :body_tag, value: "<b>theme2test</b>")
+    theme.save!
+    theme2.save!
+
+    field = Theme.lookup_field([theme.id, theme2.id], :desktop, :body_tag)
+    expect(field).to match(/<b>testtheme1<\/b>/)
+    expect(field).to match(/<b>theme2test<\/b>/)
+  end
+
+  describe ".transform_ids" do
```
tgxworld: We're missing the test case here when `extend: false`

app/models/theme_field.rb
```diff
@@ -4,6 +4,17 @@ class ThemeField < ActiveRecord::Base
 
   belongs_to :upload
 
+  scope :find_by_theme_ids, ->(theme_ids) {
```
tgxworld: scopes have to be tested as well otherwise we wouldn't know if the query here is working or not.

app/models/theme_field.rb
```diff
@@ -4,6 +4,17 @@ class ThemeField < ActiveRecord::Base
...
+    return none unless theme_ids.present?
```
tgxworld: Hmm `none` seems to be undefined.

db/migrate/20180710172959_disallow_multi_levels_theme_components.rb
```diff
@@ -0,0 +1,68 @@
+class DisallowMultiLevelsThemeComponents < ActiveRecord::Migration[5.2]
```
tgxworld: The migration is a little too complicated for me to understand without being able to see the actual data moving. @OsamaSayegh What is the worst possible outcome if this migration goes wrong?

SamSaffron: worst possible outcome is a mess in theme dependencies, or a migration that refused to run

tgxworld: > worst possible outcome is a mess in theme dependencies

Would we be able to migrate the data in such a way that we can still recover if the dependencies become a mess? If not, are we able to break the migration into smaller steps? I'm not sure what is possible but my sense is that this migration is tricky one.

app/models/theme_field.rb
OsamaSayegh: `none` is an ActiveRecord method like `all` etc. It returns an empty ActiveRecord relation (it doesn't query the DB).

https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/none

db/migrate/20180710172959_disallow_multi_levels_theme_components.rb
OsamaSayegh: It's a tricky one indeed. I can provide you with before and after data for a few themes so that you can understand what this migration does and see if the data looks correct.

OsamaSayegh:
btw @tgxworld I've made the new changes you requested.

app/models/theme_field.rb
tgxworld: Ahh icic. :+1:

SamSaffron:
OK time to 🚢  this ! 

Thanks @OsamaSayegh !

PR #6200, opened 2018-07-30
UX: Changes in top categories of hamburger menu

vinothkannans:
Changes as per Jeff's comments [here](https://meta.discourse.org/t/improving-the-category-list-in-the-hamburger-menu/92739/28?u=vinothkannans) & [here](https://meta.discourse.org/t/improving-the-category-list-in-the-hamburger-menu/92739/31?u=vinothkannans). New site setting `hamburger_menu_categories_count` introduced to control the number of categories displaying in the hamburger menu.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/hamburger-categories.js.es6
+++ b/app/assets/javascripts/discourse/widgets/hamburger-categories.js.es6
@@ -51,13 +51,18 @@ export default createWidget("hamburger-categories", {
 
   html(attrs) {
     const href = Discourse.getURL("/categories");
+    let title = I18n.t("filters.categories.title");
+    if (attrs.moreCount > 0) {
+      title += I18n.t("categories.more", { count: attrs.moreCount });
+    }
+
     let result = [
       h(
         "li.heading",
         h(
           "a.d-link.categories-link",
           { attributes: { href } },
-          I18n.t("filters.categories.title")
+          title
         )
       )
     ];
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/widgets/hamburger-categories.js.es6
```diff
@@ -51,14 +51,15 @@ export default createWidget("hamburger-categories", {
 
   html(attrs) {
     const href = Discourse.getURL("/categories");
+    let title = I18n.t("filters.categories.title");
+    if (attrs.moreCount > 0) {
+      title += I18n.t("categories.more", { count: attrs.moreCount });
```
gschlager: @vinothkannans I must have missed this. Concatenating translations in code is never a good idea. It makes it hard to translate in some languages.

config/locales/client.en.yml
```diff
@@ -571,7 +571,7 @@ en:
       topic_stat_sentence:
         one: "%{count} new topic in the past %{unit}."
         other: "%{count} new topics in the past %{unit}."
-      more: "more"
+      more: " (%{count} more) ..."
```
gschlager: @vinothkannans You should always use pluralization (`one` and `other`) when you add a translation with `%{count}` in it. The need for an empty space at the beginning of a text is also a bad sign. See my above comment regarding concatenating strings.

vinothkannans: @gschlager sure, I will do the necessary changes here.

PR #6210, opened 2018-07-31
Order moderators_activity report by moderator username

scossar:
To get this to work, I've had to change `User.real.where(moderator: true).find_each` to `User.real.where(moderator: true).order(:username_lower).each`. 

If this is a problem, the query that sets up the `mod_data` hash could be left as it is and then ordered with  `mod_data = mod_data.sort_by{ |k, v| v[:username].downcase }.to_h`

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -635,7 +635,7 @@ class Report
     report.data = []
     mod_data = {}
 
-    User.real.where(moderator: true).find_each do |u|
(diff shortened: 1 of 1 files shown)


tgxworld:
@scossar Hmm I'm not sure about depending on the order of the keys in a hash. I think we should fix the ordering in the query of the report that we are running.

scossar:
I thought the order of a hash could be depended on since ruby 1.9. I don't like the structure of this query very much though, so I'll rework it.

PR #6213, opened 2018-07-31
REFACTOR: Serve auth provider information in the site serializer

davidtaylorhq:
At the moment core providers are hard-coded in Javascript, and plugin providers get added to the JS payload at compile time. 

Advantages of this refactor:
- We only ship information about enabled auth plugins to the client
- Core auth providers are treated exactly the same as plugin auth providers on the client
- No need to ship 'enabled' settings to the client
- Fewer lines of code 🎉
- Existing tests all pass

There are no changes to the plugin api interface.


SamSaffron:
I support this, its a much cleaner pattern!

eviltrout:
Overall I think it's a great change :+1: 

However, I am wondering about the change from camelCase to snake_case. I understand why it's done (our server APIs are snake_case). But are we 100% sure no plugins were relying on this? What if any templates were overridden or code was extended in a plugin that depends on the old format?

At the very least I want to make sure we've searched through our all-the-plugins repo for the variables that have changed.

davidtaylorhq:
> At the very least I want to make sure we've searched through our all-the-plugins repo for the variables that have changed.

👍 will do

SamSaffron:
Feel free to merge when you feel it is safe (Monday is probably a good day)

davidtaylorhq:
> At the very least I want to make sure we've searched through our all-the-plugins repo for the variables that have changed.

I have searched all-the-plugins, and customer plugins, and have found no occurrences of the camel-case variables I have renamed. So I will merge this on my Monday morning :)

PR #6226, opened 2018-08-01
Improvements to user drafts

pmusaraj:
Improvements and fixes to drafts in user profile, see https://meta.discourse.org/t/drafts-view-in-profile/13720/9 for details.

### DIFF
--- a/app/models/draft.rb
+++ b/app/models/draft.rb
@@ -58,13 +58,15 @@ class Draft < ActiveRecord::Base
         pu.username, pu.name, pu.id user_id, pu.uploaded_avatar_id, pu.username_lower,
         du.username draft_username, NULL as raw, NULL as cooked, NULL as post_number
       FROM drafts d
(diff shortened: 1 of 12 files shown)


app/models/draft.rb
```diff
@@ -58,13 +58,15 @@ def self.stream(opts = nil)
         pu.username, pu.name, pu.id user_id, pu.uploaded_avatar_id, pu.username_lower,
         du.username draft_username, NULL as raw, NULL as cooked, NULL as post_number
       FROM drafts d
+      LEFT JOIN LATERAL json_extract_path_text (d.data::json, 'postId') postId ON TRUE
```
SamSaffron: oh my ... this is getting more complex, longer term we are going to have to just add some columns to the drafts table

pmusaraj: Yeah, only way to get to the post's user right now...

SamSaffron:
cool, let's try this out!

PR #6230, opened 2018-08-02
Rework moderators activity query

scossar:
Any suggestions for improving this would be great. 

I can't find a way of returning the `avatar_template` other than getting it in the `DB.query(query).each` block.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -722,20 +722,10 @@ class Report
     ]
 
     report.modes = [:table]
-
     report.data = []
-    mod_data = {}
-
-    User.real.where(moderator: true).find_each do |u|
-      mod_data[u.id] = {
-        user_id: u.id,
-        username: u.username_lower,
-        user_avatar_template: u.avatar_template,
-      }
-    end
 
     query = <<~SQL
-    WITH m AS (
+    WITH mods AS (
     SELECT
     id AS user_id,
     username_lower AS username,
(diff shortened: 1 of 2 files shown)


jjaffeux:
@scossar have you looked how I do it in `report_post_edits` ?

scossar:
Yes, I think I'm doing it the same way here:

` mod[:user_avatar_template] = User.avatar_template(row.username, row.uploaded_avatar_id)`

I was trying to find a way to get the avatar template from the DB query to avoid having to look it up in the `DB.query(query).each` loop, but that doesn't seem possible.

SamSaffron:
hmmm when is this ready to merge?

app/models/report.rb
```diff
@@ -722,152 +722,144 @@ def self.report_moderators_activity(report)
...
+    ORDER BY m.username
+    SQL
 
-    DB.query(pm_count_query).each do |row|
-      mod_data[row.user_id][:pm_count] = row.pm_count
+    DB.query(query).each do |row|
+      mod = {}
+      mod[:username] = row.username
+      mod[:user_id] = row.user_id
+      mod[:user_avatar_template] = User.avatar_template(row.username, row.uploaded_avatar_id)
```
SamSaffron: Yeah this is how you get templates, note no query runs here ...

scossar:
> hmmm when is this ready to merge?

I'm happy with it.

SamSaffron:
OK, let's try this out then

PR #6233, opened 2018-08-03
FEATURE: add link to see new commits when updates are available for themes

OsamaSayegh:


app/models/remote_theme.rb
```diff
@@ -6,6 +6,9 @@ class RemoteTheme < ActiveRecord::Base
 
   ALLOWED_FIELDS = %w{scss embedded_scss head_tag header after_header body_tag footer}
 
+  GITHUB_REGEXP = /^(http|https):\/\/github\.com\//
```
ZogStriP: This Regexp can be a tiny bit shorter: `/^https?:\/\/github\.com\//`  ;)

OsamaSayegh: Thanks! :D

SamSaffron:
let's try this! thanks

PR #6235, opened 2018-08-04
FEATURE: Admin wrench action to change topic bump date and disable bumping

gschlager:
Adds a new action to the admin wrench of topics.

image

The dialog allows disabling bumping of the selected topic which prevents bumps by replying to the topic, editing the topic's last post or autobumping within the category.
In addition to that it allows changing the topic's bump date to an arbitrary value <= now. The date picker defaults to the current bump date.

image

* Adds a new column (`skip_bump`) to the `topics` table. It doesn't set a default value in order to prevent long locks on the table during the migration. `NULL` is considered to be `false`.
* Adds `bumped_at` and `skip_bump` to the topic serializer if the current user is an admin.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/bump-topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/bump-topic.js.es6
@@ -1,0 +1,79 @@
+import ModalFunctionality from "discourse/mixins/modal-functionality";
+import computed from "ember-addons/ember-computed-decorators";
+import DiscourseURL from "discourse/lib/url";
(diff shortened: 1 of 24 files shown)


SamSaffron:
let's hold tight on this, per dev discussion.

SamSaffron:
closing for now till we redo this, sorry

gschlager:
Replaced by PR https://github.com/discourse/discourse/pull/6246 and https://github.com/discourse/discourse/pull/6247

PR #6236, opened 2018-08-06
FIX: `UserAvatar#update_gravatar!` does not update `User#uploaded_ava…

tgxworld:
…tar`.

https://meta.discourse.org/t/missing-user-profile-pictures/93844/4


tgxworld:
Merged in https://github.com/discourse/discourse/commit/4e11811321cea64709a26f6ff9ba7570f8e45fa6

PR #6238, opened 2018-08-06
FIX: don't require auth providers to set full_screen_login_setting

LeoMcA:
@davidtaylorhq not sure if this was your intention, but https://github.com/discourse/discourse/commit/812add18bdb071d757a6faff4358f0252122f946 forced non-fsl on authentication providers which set `full_screen_login` but don't set `full_screen_login_setting`. This fixes that.


davidtaylorhq:
Good spot - thanks!

LeoMcA:
I was wondering why my authentication provider started popping up in a new window, so did a little digging. Thanks for the quick merge!

PR #6243, opened 2018-08-07
FIX: wrongful redirections when topic starts with BaseUri

bram-rongen:
Fixes https://meta.discourse.org/t/discourse-rewriting-url-path-behaviour-failing/94120

### DIFF
--- a/app/assets/javascripts/discourse/lib/discourse-location.js.es6
+++ b/app/assets/javascripts/discourse/lib/discourse-location.js.es6
@@ -66,7 +66,10 @@ const DiscourseLocation = Ember.Object.extend({
   getURL() {
     const location = get(this, "location");
     let url = location.pathname;
-    url = url.replace(Discourse.BaseUri, "");
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/lib/discourse-location.js.es6
```diff
@@ -66,7 +66,10 @@ const DiscourseLocation = Ember.Object.extend({
   getURL() {
     const location = get(this, "location");
     let url = location.pathname;
-    url = url.replace(Discourse.BaseUri, "");
+
+    if (Discourse.BaseUri.length() > 0 && url.startsWith(Discourse.BaseUri)) {
+      url = url.substring(Discourse.BaseUri.length);
+    }
```
SamSaffron: something about this fix is not sitting right with me, hold tight while we investigate

bram-rongen:
Cool, it's fixed I see! A oneliner is nicer indeed :)

PR #6246, opened 2018-08-08
FEATURE: Add "Reset Bump Date" action to topic admin wrench

gschlager:


lib/guardian/topic_guardian.rb
```diff
@@ -146,4 +146,8 @@ def can_edit_featured_link?(category_id)
     return false unless SiteSetting.topic_featured_link_enabled
     Category.where(id: category_id || SiteSetting.uncategorized_category_id, topic_featured_link_allowed: true).exists?
   end
+
+  def can_reset_bump_date?
+    is_admin?
```
SamSaffron: yeah I would like this to be staff, not just restricted to admin

app/models/topic.rb
```diff
@@ -1370,6 +1370,16 @@ def is_category_topic?
...
   end
 
+  def reset_bumped_at
+    post = ordered_posts.where(
+      user_deleted: false,
+      hidden: false,
+      post_type: Topic.visible_post_types
+    ).last
+
+    update_attribute(:bumped_at, post.created_at)
```
tgxworld: Since we're not checking for the return value, we would want to use `update!` here so it doesn't fail silently.

SamSaffron:
looks good to me, awesome!

PR #6247, opened 2018-08-08
FEATURE: Allow admins to reply without topic bump

gschlager:


app/controllers/posts_controller.rb
```diff
@@ -653,6 +653,11 @@ def create_params
       result[:is_warning] = false
     end
 
+    if params[:no_bump] == "true"
+      raise Discourse::InvalidParameters.new(:no_bump) unless current_user.admin?
```
SamSaffron: oh ... this is a staff feature, not only an admin feature, also please move to guardian

app/assets/javascripts/select-kit/components/composer-actions.js.es6
```diff
@@ -192,6 +192,17 @@ export default DropdownSelectBoxComponent.extend({
       });
     }
 
+    const currentUser = Discourse.User.current();
+
+    if (action === REPLY && currentUser && currentUser.get("admin")) {
```
SamSaffron: yeah we want this for all staff, not only admin

app/assets/javascripts/select-kit/components/composer-actions.js.es6
```diff
@@ -192,6 +192,17 @@ export default DropdownSelectBoxComponent.extend({
       });
     }
 
+    const currentUser = Discourse.User.current();
```
tgxworld: @eviltrout I can't recall where the discussion happened in the past but do you remember why we wanted to avoid accessing models via the `Discourse` global?

tgxworld:
LGTM :+1:

SamSaffron:
excellent, merging this in!

PR #6248, opened 2018-08-08
FIX: don't allow inviting more than `max_allowed_message_recipients` setting allows

OsamaSayegh:


OsamaSayegh:
@eviltrout thanks for the feedback! There is a different guardian method `can_invite_to?` that checks for other things such as sufficient trust level etc. I see how it's confusing now, let me try to clear it up.

OsamaSayegh:
@eviltrout let me know if it's clearer now, thanks.

lib/guardian.rb
```diff
@@ -272,6 +272,7 @@ def can_invite_to?(object, groups = nil)
 
     if object.is_a?(Topic) && object.private_message?
       return false unless SiteSetting.enable_personal_messages?
+      return false if object.reached_recipients_limit? && !is_staff?
```
tgxworld: I think we should update the specs to reflect this change.

app/models/topic_user.rb
```diff
@@ -40,7 +40,7 @@ def notification_reasons
     def auto_notification(user_id, topic_id, reason, notification_level)
       should_change = TopicUser
         .where(user_id: user_id, topic_id: topic_id)
-        .where("notifications_reason_id IS NULL OR (notification_level < :min AND notification_level > :max)", min: notification_level, max: notification_levels[:regular])
+        .where("notifications_reason_id IS NULL OR (notification_level < :max AND notification_level > :min)", max: notification_level, min: notification_levels[:regular])
```
tgxworld: Is this change intentional because the code path that has been updated doesn't seem to be affected by this change here.

app/controllers/topics_controller.rb
```diff
@@ -538,6 +538,12 @@ def invite
       group_names: params[:group_names]
     )
 
+    if !guardian.is_staff? && topic.reached_recipients_limit?
```
tgxworld: I feel like the responsibility of checking if there are too many recipients should be part of `Topic#invite` instead.

OsamaSayegh: I did initially put the code in the model method, but then I wanted to show a message to users when they try to go above the recipients limit and here is where I got a problem. We have `guardian.ensure_can_invite_to!(..)` called in the controller `invite` method, and this guardian method does check if there are too many recipients, so when it fails it's going to raise a generic 403 error. Is it OK if I repeat the code in the model method? I can't think of a cleaner way to achieve that.

app/models/topic_user.rb
OsamaSayegh: Yeah it's not related but I was reading the code and found the placeholders names very confusing. It's just mathematically impossible to find a value larger than `max` and smaller than `min` at the same time.

OsamaSayegh:
I've made a couple changes here:

1. "Automatically track topics I enter..." and "When I post in a topic, set that topic to..." user preferences will no longer apply to PMs

2. groups will now take up 1 slot in PM

app/controllers/topics_controller.rb
```diff
@@ -538,6 +544,12 @@ def invite
       group_names: params[:group_names]
     )
 
+    unless pm_has_slots?(topic)
```
SamSaffron: This does not feel right cause you are allowed it appears to invite multiple groups, so you can cheat and invite 30 groups ? no?

OsamaSayegh: Yes that will allow you to cheat and invite 30 groups. Maybe we should have another site setting for max groups in a PM? Because making every group member take up a slot doesn't feel quite right to me.

SamSaffron: I think this is fine... just add `pm_has_slots?(topic, groups.length)` or something like that.

OsamaSayegh: the `groups` variable here represents groups that you want invited user to be a member of, when you invite by email. These groups won't be invited to the message.

Inviting a group to a message is handled via a separate controller method `invite_group`, and you can't invite groups in bulk, only one by one.

We do a check like this
`topic_allowed_users.count + topic_allowed_groups.count >= SiteSetting.max_allowed_message_recipients`
when you invite a group or user.

lib/post_creator.rb
```diff
@@ -531,7 +531,8 @@ def track_topic
     if @user.staged
       TopicUser.auto_notification_for_staging(@user.id, @topic.id, TopicUser.notification_reasons[:auto_watch])
     else
-      notification_level = @user.user_option.notification_level_when_replying || NotificationLevels.topic_levels[:tracking]
+      notification_level = TopicUser.where(user: @user, topic: @topic).pluck(:notification_level).first if @topic.private_message?
```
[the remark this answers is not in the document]
OsamaSayegh: Agreed, we can just `return` if topic is PM.

app/models/topic_user.rb
tgxworld: @OsamaSayegh I wonder if this might be a on-demand fix for a bug that was introduced in the past. Maybe have a look through the git commit history?

SamSaffron:
cool lets try this out!

PR #6249, opened 2018-08-08
FEATURE: Allow users with group_locked_trust_level to be promoted to tl3

scossar:
I'm not sure if this is a fix, or a feature. This PR allows users with a `group_locked_trust_level` to be automatically promoted to TL3. It prevents users with a `group_locked_trust_level` of 3 from being automatically demoted to TL2.


SamSaffron:
You know, the real problem here is a naming problem. 

group does not intend to "lock" the trust level, it intends to set min trust level. we should rename the column to `group_min_trust_level` it would be far less confusing.

PR #6253, opened 2018-08-09
Added GPLv2 license to JS files

MrWils:
This should fix a GNU LibreJS issue.

GNU LibreJS blocked discourse in the past.


SamSaffron:
Hmmm, I don't see how this would help, we bundle the JS for production, presumably you only need this on the bundled JS, no?

MrWils:
Ha yeah, that might be smarter.  
I didn't knew that.  
I will change this commit when I find some time!

eviltrout:
Additionally, our code is GPL but this applies the header to vendored files
too which may or may not be, as they are created by others. We can only
guarantee Discourse’s code is GPL.

eviltrout:
I understand that, but marking other people's work with the wrong license doesn't help.

MrWils:
You can use the most licenses by using a header so that shouldn't be a problem.   
It might take some time to change all the JS though.

This would might bring some more users since some people use the GNU LibreJS add-on which currently blocks Discourse.  

It might not be a big deal for everyone but it's just a suggestion.

eviltrout:
Thanks, I appreciate it. I think we just need a different approach. We need to take into account minified/concatenated assets plus the vendored assets.

PR #6257, opened 2018-08-10
FEATURE: Do encodeURI on share links if needed.

misaka4e21:
Some Discourse-supported sharing platforms such as Twitter are unable to determine the end of a unicode URL. If you post a URL "https://example.org/t/测试/1" on Twitter, it will be a link of `href="https://example.org/t/"`

If the URL contains any unicode character (usually in the slug part) , it must be urlencoded with `encodeURI(url)` before being passed to `source.generateUrl(link, title)`, or it will be a malformed URL in the sharing tweet.

### DIFF
--- a/app/assets/javascripts/discourse/components/share-popup.js.es6
+++ b/app/assets/javascripts/discourse/components/share-popup.js.es6
@@ -81,7 +81,7 @@ export default Ember.Component.extend({
     if (!this.site.mobileView) {
       $this.css({ left: "" + x + "px" });
     }
-    this.set("link", url);
+    this.set("link", encodeURI(url));
     this.set("visible", true);
 
     Ember.run.scheduleOnce("afterRender", this, this._focusUrl);


config/site_settings.yml
```diff
@@ -170,6 +170,16 @@ basic:
      - facebook
      - google+
      - email
+  share_links_urlencode:
```
gschlager: You need to add a translation in `server.en.yml` when you add a new site setting.

gschlager:
An alternative to adding yet another site setting would be to always encode the URL for every share target. Except for emails you don't even see the link you share. The URL always gets shortened and/or replaced by something like our onebox. Thoughts?

coding-horror:
Why do we even need a site setting? Why can't this be automatic if unicode
is detected in the URL? I vote we make this automatic.

gschlager:
Yeah, it's probably best to always use `encodeURI` no matter what. That makes it a one-line change instead of adding lots of complexity.

SamSaffron:
OK this looks pretty safe, lets try it out! thanks!

PR #6259, opened 2018-08-10
FIX: disable_2fa fix method selection

pfaffman:
The previous code resulted in

    NameError: undefined local variable or method `totp' for main:Object

I now understand what @tgxworld meant about we should only disable totp when I submitted this before.

This is the kind of Ruby stuff that I still don't understand well,(perhaps this isn't the most Ruby way to do this?) but this does what I think is supposed to happen. And it worked just now.


tgxworld:
Thank you @pfaffman and apologies for the delay.

PR #6261, opened 2018-08-12
FIX: update TopicEmbed's title and user correctly

xrav3nz:
#5771 tried to add support for updating title and author of imported `TopicEmbed`s, but I made two mistakes:

1. Title and author won't be updated unless the content also changed
2. Changing the (topic's) user should be done with `PostOwnerChanger` instead of `PostRevisor`.

This PR resolves ⬆️.

### DIFF
--- a/app/models/topic_embed.rb
+++ b/app/models/topic_embed.rb
@@ -71,16 +71,28 @@ class TopicEmbed < ActiveRecord::Base
     else
       absolutize_urls(url, contents)
       post = embed.post
-      # Update the topic if it changed
-      if post && post.topic && content_sha1 != embed.content_sha1
-        post_revision_args = {
-          raw: absolutize_urls(url, contents),
-          user_id: user.id,
-          title: title,
-        }
 
-        post.revise(user, post_revision_args, skip_validations: true, bypass_rate_limiter: true)
-        embed.update_column(:content_sha1, content_sha1)
+      # Update the topic if it changed
+      if post&.topic
+        revision = {}
+
+        if post.user != user
+          PostOwnerChanger.new(
+            post_ids: [post.id],
+            topic_id: post.topic_id,
+            new_owner: user,
+            acting_user: Discourse.system_user
+          ).change_owner!
+
+          # make sure the post returned has the right author
+          post.reload
+        end
+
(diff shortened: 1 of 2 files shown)


app/models/topic_embed.rb
```diff
@@ -71,16 +71,28 @@ def self.import(user, url, title, contents)
...
+
+          # make sure the post returned has the right author
+          post.reload
+        end
+
+        revision[:raw] = absolutize_urls(url, contents) if content_sha1 != embed.content_sha1
+        revision[:title] = title if title != post.topic.title
+
+        post.revise(user, revision, skip_validations: true, bypass_rate_limiter: true) unless revision.empty?
+        embed.update_column(:content_sha1, content_sha1) if revision[:raw]
```
xrav3nz: I was worried if wrapping `PostRevisor` in a transaction block would potentially cause deadlocks, because of its complexity and internal nested transactions. Lmk what you think.

xrav3nz: (also, open suggestions to style refactors; my changes feel a bit awkward to me in terms of flow and readability, but couldn't think of better ways 😢 )

spec/models/topic_embed_spec.rb
```diff
@@ -40,13 +40,22 @@
...
 
         expect(post.cooked).to match(/new contents/)
-        expect(post.topic.title).to eq("I am a new title")
+      end
+
+      it "Supports updating the post author" do
+        new_user = Fabricate(:user)
+        post = TopicEmbed.import(new_user, url, title, contents)
+
         expect(post.user).to eq(new_user)
```
tgxworld: @xrav3nz Hmm how was this assertion passing previously? I feel that something is not right because if this assertion was passing previously, this assertion is either wrong or we're missing another assertion.

tgxworld:
Apologies for the delay in our review @xrav3nz :grin: thank you for contributing!

spec/models/topic_embed_spec.rb
xrav3nz: This passed because the post body changed in the setup before. (Previously, changes to author and title would only be recorded if the post body also changed)

However, you actually reminded me that I need another assertion here: `PostOwnerChanger` also changes the `post.topic.user` (if post is the first one), whereas `PostRevisor` doesn't. Added this and rebased!

PR #6262, opened 2018-08-12
Remove unnecessary code in d-editor CSS

tshenry:
Just a quick one-line edit of the d-editor CSS. There was a `border-bottom: none;` that gets overridden a few lines below by `border-bottom: 1px solid $primary-low;` making it unnecessary.


SamSaffron:
yeah that code looks ... odd , thanks

PR #6264, opened 2018-08-13
FEATURE: Automatically expire keys if not used for a configurable amount of time.

nbianca:
Fixes [this](https://meta.discourse.org/t/apps-collect-display-last-usage-information/87842/5?u=sam).

### DIFF
--- a/app/jobs/scheduled/auto_expire_user_api_keys.rb
+++ b/app/jobs/scheduled/auto_expire_user_api_keys.rb
@@ -1,0 +1,15 @@
+module Jobs
+
+  class AutoExpireUserApiKeys < Jobs::Scheduled
+    every 1.day
+
+    def execute(args)
+      if SiteSetting.expire_user_api_keys_days > 0
+        expire_user_api_keys_days = SiteSetting.expire_user_api_keys_days.days.ago
+
+        UserApiKey.where("last_used_at < ?", expire_user_api_keys_days).update_all(revoked_at: Time.zone.now)
+      end
+    end
+  end
+
+end
--- a/lib/auth/default_current_user_provider.rb
+++ b/lib/auth/default_current_user_provider.rb
@@ -258,6 +258,7 @@ class Auth::DefaultCurrentUserProvider
         raise Discourse::InvalidAccess
       end
 
+      api_key.update_columns(last_used_at: Time.zone.now)
       if client_id.present? && client_id != api_key.client_id
         api_key.update_columns(client_id: client_id)
       end
--- a/spec/integration/user_api_keys_spec.rb
+++ b/spec/integration/user_api_keys_spec.rb
@@ -1,0 +1,16 @@
+require 'rails_helper'
+
+describe 'user api keys integration' do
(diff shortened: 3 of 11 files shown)


app/jobs/scheduled/autoexpire_user_api_keys.rb
```diff
@@ -0,0 +1,15 @@
+module Jobs
+
+  class AutoexpireUserApiKeys < Jobs::Scheduled
```
SamSaffron: naming wise I would recommend AutoExpireUserApiKeys here, and the setting should be expire_user_api_keys_days

lib/auth/default_current_user_provider.rb
```diff
@@ -258,6 +258,7 @@ def lookup_user_api_user_and_update_key(user_api_key, client_id)
         raise Discourse::InvalidAccess
       end
 
+      api_key.update_columns(last_used_at: Time.now)
```
SamSaffron: We should test this as well in an integration test

tgxworld: @nbianca This is a good read about dealing with Time zones in Rails http://danilenko.org/2012/7/6/rails_timezones/. The TLDR is that you almost always want to use `Time.zone.now` over `Time.now`

app/jobs/scheduled/auto_expire_user_api_keys.rb
```diff
@@ -0,0 +1,15 @@
+module Jobs
+
+  class AutoExpireUserApiKeys < Jobs::Scheduled
+    every 1.day
+
+    def execute(args)
+      if SiteSetting.expire_user_api_keys_days > 0
+        expire_user_api_keys_days = SiteSetting.expire_user_api_keys_days.days.ago
+
+        UserApiKey.where("last_used_at < ?", expire_user_api_keys_days).update_all(revoked_at: Time.now)
```
tgxworld: I think we need to think about clean up of revoked keys as well so we don't bloat the table over time.

spec/integration/user_api_keys_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe 'user api keys integration' do
+
+  let(:key) { Fabricate(:readonly_user_api_key) }
+
+  it 'updates last used time on use' do
+    key.update_columns(last_used_at: 7.days.ago)
+
+    freeze_time
+    get '/session/current.json', headers: {
+      HTTP_USER_API_KEY: key.key,
```
tgxworld: `key.key` reads a little odd to me :grin: Perhaps `user_api_key.key` is better?

app/jobs/scheduled/auto_expire_user_api_keys.rb
nbianca: Do you think it would be fine to clean up all revoked keys older than `SiteSetting.delete_user_api_keys_days` (new site setting) or maybe something like `2 * SiteSetting.expire_user_api_keys_days`?

PR #6265, opened 2018-08-13
FIX: Do not show an empty modal when an IP address is allowed or blocked.

nbianca:
Fixes [this](https://meta.discourse.org/t/blacklisting-ip-has-no-feedback/94513).

### DIFF
--- a/app/assets/javascripts/admin/components/screened-ip-address-form.js.es6
+++ b/app/assets/javascripts/admin/components/screened-ip-address-form.js.es6
@@ -59,18 +59,14 @@ export default Ember.Component.extend({
         screenedIpAddress
           .save()
           .then(result => {
-            if (result.success) {
-              this.setProperties({ ip_address: "", formSubmitted: false });
-              this.sendAction(
-                "action",
-                ScreenedIpAddress.create(result.screened_ip_address)
-              );
-              Ember.run.schedule("afterRender", () =>
-                this.$(".ip-address-input").focus()
-              );
-            } else {
-              bootbox.alert(result.errors);
-            }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/components/screened-ip-address-form.js.es6
```diff
@@ -59,7 +59,7 @@ export default Ember.Component.extend({
         screenedIpAddress
           .save()
           .then(result => {
-            if (result.success) {
+            if (!result.errors) {
```
SamSaffron: This feels a bit odd to me at the API level. 

@nlalonde #create returns the address serialized and updater returns json success, we got to pick a side here. technically the promise will fail here if there is an error so you don't need to check for either, render_json_error defaults to 422 status

nlalonde: Right, so it should be render_serialized since it calls `ScreenedIpAddress.create(result.screened_ip_address)`. But validation errors are being returned so I don't think the promise will fail, which is why we're trying to show a modal with the error.

nbianca:
Hey. I have also noticed that editing was not working properly (was showing a generic error message when was failing), so I fixed that too.

I have also uniformized the `#create` and `#update` API. (Y)

nbianca:
Hey. I have also noticed that editing was not working properly (was showing a generic error message when was failing), so I fixed that too.

I have also uniformized the `#create` and `#update` API.

SamSaffron:
Nice! both less code and more consistent 👍  from me, just need to fix the little linting error and go ahead and merge

PR #6267, opened 2018-08-14
UX: Replace Google search with Discourse search on not found page.

misaka4e21:


SamSaffron:
cool thanks! I support this change

PR #6271, opened 2018-08-15
FEATURE: implement s3_canned_acl site setting

abelmokadem:
Implement S3 Canned ACL as a site setting

### DIFF
--- a/spec/components/file_store/s3_store_spec.rb
+++ b/spec/components/file_store/s3_store_spec.rb
@@ -17,6 +17,7 @@ describe FileStore::S3Store do
     SiteSetting.s3_access_key_id = "s3-access-key-id"
     SiteSetting.s3_secret_access_key = "s3-secret-access-key"
     SiteSetting.enable_s3_uploads = true
+    SiteSetting.s3_canned_acl = 'public-read'
   end
 
   shared_context 's3 helpers' do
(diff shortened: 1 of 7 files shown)


spec/components/file_store/s3_store_spec.rb
```diff
@@ -250,6 +251,13 @@
 
   end
 
+  it "s3_canned_acl" do
```
ZogStriP: What's the point of this test? If it's testing the validation, then it should at least test for a bad case.

Also, the validation is already tested in `s3_canned_acl_site_setting_spec.rb`, so not sure what this is testing.

spec/models/s3_canned_acl_site_setting_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+describe S3CannedACLSiteSetting do
+
+  describe 'valid_value?' do
+    it 'returns true for a valid S3 region' do
```
ZogStriP: I don't think that's the right description ;)

SamSaffron:
on hold per: 

https://meta.discourse.org/t/migrating-uploads-to-s3-does-not-work-with-private-bucket/94591/6?u=sam

SamSaffron:
So sorry closing this, lets figure out another way of doing this from a plugin without adding a site setting.

PR #6273, opened 2018-08-15
UX: Show static page title.

nbianca:
This is work in progress as I still have to find a nicer way for the title and also add it to meta description too.

Both "tos" and "terms_of_service" have been used as i18n keys and I could:
(a) add a new `tos` key that would be exactly the same with `terms_of_service`;
(b) use this ugly `if`.

### DIFF
--- a/app/assets/javascripts/discourse/lib/static-route-builder.js.es6
+++ b/app/assets/javascripts/discourse/lib/static-route-builder.js.es6
@@ -34,6 +34,10 @@ export default function(page) {
       this.controllerFor("static").set("model", model);
     },
 
+    titleToken() {
+      return I18n.t(page);
+    },
+
     actions: {
       didTransition() {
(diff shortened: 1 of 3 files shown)


eviltrout:
I could go either way. Sometimes duplicating a key isn't bad because it's gzipped, but I agree there is a certain ugliness to both approaches. It's up to you.

gschlager:
You could use a YML alias if you want to go with (a). That way translators don't need to translate it more than once:

```
    &tos terms_of_service: "Terms of Service"
    tos: *tos
```

Or you could rename the `terms_of_service` key to `tos`. I think it's only used once.

SamSaffron:
generally when we rename keys we do not touch the various locales. The problem with renaming in the various yaml files is that next time we pull translations it will break anyway, its better to break people early so they fix it early.

PR #6274, opened 2018-08-16
Review for https://github.com/discourse/discourse/commit/796164b58c98…

tgxworld:
…9bbfe5a16e2dff120347a7527950.

### DIFF
--- a/app/controllers/user_avatars_controller.rb
+++ b/app/controllers/user_avatars_controller.rb
@@ -203,14 +203,16 @@ class UserAvatarsController < ApplicationController
   def fix_extension(upload)
     # this is relatively cheap
     original_path = Discourse.store.path_for(upload)
+
     if original_path.blank?
-      external_copy = Discourse.store.download(upload) rescue nil
-      original_path = external_copy.try(:path)
+      external_copy = Discourse.store.download(upload)
+      original_path = external_copy&.path
     end
 
-    image_info = FastImage.new(original_path) rescue nil
-    if image_info && image_info.type.to_s != upload.extension
-      upload.update_columns(extension: image_info.type.to_s)
+    image_type = FastImage.type(original_path).to_s
+
+    if image_type.to_s != upload.extension
+      upload.update!(extension: image_type)
       true
     end
   end
--- a/spec/requests/user_avatars_controller_spec.rb
+++ b/spec/requests/user_avatars_controller_spec.rb
@@ -21,6 +21,7 @@ describe UserAvatarsController do
       after do
         FileUtils.rm(Discourse.store.path_for(upload))
       end
+
       # travis is not good here, no image magick
       if !ENV["TRAVIS"]
         let :upload do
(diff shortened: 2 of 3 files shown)


app/controllers/user_avatars_controller.rb
```diff
@@ -203,14 +203,16 @@ def get_optimized_image(upload, size)
   def fix_extension(upload)
     # this is relatively cheap
     original_path = Discourse.store.path_for(upload)
+
     if original_path.blank?
-      external_copy = Discourse.store.download(upload) rescue nil
```
tgxworld: I've been bitten by `rescue nil` many times. Any typos or undefined method are buried so I don't think we should use the pattern.

SamSaffron: happy to have this fixed AS LONG as its also fixed in UploadCreator

spec/requests/user_avatars_controller_spec.rb
```diff
@@ -57,6 +58,12 @@
 
           expect(OptimizedImage.where(upload_id: upload.id).count).to eq(1)
           expect(response.status).to eq(200)
+
+          upload.reload
+
+          expect(upload.extension).to eq('png')
+          expect(File.extname(upload.url)).to eq('.png')
+          expect(File.extname(upload.original_filename)).to eq('.png')
```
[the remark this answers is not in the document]
SamSaffron: those assertions are not correct, all we should be asserting is that the extension is right, original_filename does not and should not matter

app/controllers/user_avatars_controller.rb
SamSaffron: oh was commenting on fast image ... this particular code is from OptimizedImage, so it needs to also be fixed

spec/requests/user_avatars_controller_spec.rb
```diff
@@ -45,7 +46,7 @@
           upload.update_columns(
             original_filename: 'bob.jpg',
             extension: 'jpg',
-            url: upload.url + '.jpg'
+            url: "#{upload.url.sub('.png', '')}.jpg"
```
SamSaffron: this is sort of ok, but really does not matter that much

app/controllers/user_avatars_controller.rb
```diff
@@ -203,14 +203,16 @@ def get_optimized_image(upload, size)
...
-      original_path = external_copy.try(:path)
+      external_copy = Discourse.store.download(upload)
+      original_path = external_copy&.path
     end
 
-    image_info = FastImage.new(original_path) rescue nil
-    if image_info && image_info.type.to_s != upload.extension
-      upload.update_columns(extension: image_info.type.to_s)
+    image_type = FastImage.type(original_path).to_s
```
SamSaffron: UploadCreator has this usage, we can not have one pattern in one spot and another in another spot

spec/requests/user_avatars_controller_spec.rb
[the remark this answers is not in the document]
tgxworld: Why are the assertions incorrect though? It doesn't make sense to me to only have the extension be the only correct one. It'll be confusing to have the url and original filename be different from the extension.

PR #6278, opened 2018-08-16
Allow plugins to intercept uploads in composer

davidtaylorhq:
Allow plugins to handle uploads in the composer for specific file types. 

For example, if you wanted to integrate with a video hosting provider, and send all video files directly to them instead of Discourse.

```js
api.addComposerUploadHandler(["mp4", "mov"], (file) => {
    console.log("Handling upload for", file.name);
    // Do uploading
})
```

### DIFF
--- a/app/assets/javascripts/discourse/lib/plugin-api.js.es6
+++ b/app/assets/javascripts/discourse/lib/plugin-api.js.es6
@@ -35,9 +35,10 @@ import { registerCustomAvatarHelper } from "discourse/helpers/user-avatar";
 import { disableNameSuppression } from "discourse/widgets/poster-name";
 import { registerCustomPostMessageCallback as registerCustomPostMessageCallback1 } from "discourse/controllers/topic";
 import Sharing from "discourse/lib/sharing";
+import { addComposerUploadHandler } from "discourse/components/composer-editor";
 
 // If you add any methods to the API ensure you bump up this number
-const PLUGIN_API_VERSION = "0.8.23";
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -753,6 +754,21 @@ class PluginApi {
...
+   * The normal uploading functionality will be bypassed
+   *
+   * Example:
+   *
+   * addComposerUploadHandler(["mp4", "mov"], (file) => {
+   *    console.log("Handling upload for", file.name);
+   * })
+   */
+  addComposerUploadHandler(extensions, method) {
+    addComposerUploadHandler1(extensions, method);
```
SamSaffron: The `1` here is not needed, I think it just makes stuff a bit more confusing cause no other method follows that pattern

SamSaffron:
This looks fine to me, I assume "about me" and so on don't need this fancy. 

@eviltrout  you ok with this?

davidtaylorhq:
> The 1 here is not needed

Fixed

> "about me" and so on don't need this fancy

Doesn't look like they have any upload functionality at the moment, so I don't think it's needed.

ZogStriP:
> > "about me" and so on don't need this fancy
> 
> Doesn't look like they have any upload functionality at the moment, so I don't think it's needed.

I think @SamSaffron meant the user profile background and user card background ;)

ZogStriP:
LGTM 👍

PR #6279, opened 2018-08-16
First phase of renaming User.group_locked_trust_level.

nbianca:


db/migrate/20180812142030_add_group_min_trust_level.rb
```diff
@@ -0,0 +1,6 @@
+class AddGroupMinTrustLevel < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :group_min_trust_level, :integer
+    execute "UPDATE users SET group_min_trust_level = group_locked_trust_level"
```
tgxworld: This can still result in data loss because this migration will run first before the [guards](https://github.com/discourse/discourse/pull/6279/files#diff-4676c008b11a5480d73d4a6de01e45b9R310) are deployed. Note that we don't bring down all of our app servers at the same time during our deployment process. Instead it is done on a rotation basis where we bring down each app server and deploy it again with the latest changes.

tgxworld: The solution we have now is the mark the old column as read-only using https://github.com/discourse/discourse/blob/master/lib/migration/column_dropper.rb#L14 but this is problematic for backups and restores. I'm trying to get it fixed by this week so maybe we can delay this till then?

tgxworld: The deployment of the app servers can potentially fail after the migration has been run which may lead to a larger data loss window.

app/models/user.rb
```diff
@@ -307,6 +307,24 @@ def self.find_by_username(username)
     find_by(username_lower: username.downcase)
   end
 
+  # TODO: To be removed after column renaming has succesfully completed.
```
SamSaffron: Sorry this is so hairy, there is a lot to learn here

I am a bit confused about this. when this code runs always ` has_attribute?(:group_min_trust_level)` will be around. All we really want here is something that makes sure we ignore the old column.

SamSaffron:
After reading through this its more than that in the name that is wrong. 

I am confused about why we even need this column on the user table at all? The information is already there on the groups the user is a member of, why copy it around and merge it. Seems like a lot of confusing work. 

cc @tgxworld

nbianca:
I updated both PRs. Can you please take another look?

app/models/user.rb
```diff
@@ -308,6 +308,13 @@ def self.find_by_username(username)
     find_by(username_lower: username.downcase)
   end
 
+  def group_min_trust_level
+    GroupUser
+      .where(user_id: id)
+      .includes(:group)
+      .maximum("groups.grant_trust_level")
```
SamSaffron: The name is a bit confusing here ... maybe `group_granted_trust_level` ?

SamSaffron:
given this change is far safer I think all you need to do is defer drop the column we no longer need and then add an ignore_column on the model. 

We can merge this in once 2.1 is out.

lib/promotion.rb
```diff
@@ -121,9 +121,9 @@ def self.recalculate(user, performed_by = nil)
     end
 
     # Then consider the group locked level
-    if user.group_locked_trust_level
+    if user.group_min_trust_level
```
SamSaffron: Oh my this is indeed locked today, yet the UI talks about it being minimal trust level. @nbianca we are going to have to change this whole workflow to allow for this, trust level still needs to be reviewed but after the review if it is lower than min trust level for group it should be raised.

SamSaffron:
also we need an ignore column and column dropper here in the same PR cause I don't want to be left with throwaway data we need to deal with later on.

tgxworld:
@nbianca Post deployment migrations have landed in master so we can switch this PR to use that :grin:

spec/jobs/tl3_promotions_spec.rb
```diff
@@ -24,8 +24,11 @@ def create_qualifying_stats(user)
     run_job
   end
 
-  it "promotes a qualifying tl2 user who has a group_locked_trust_level" do
-    _group_locked_user = Fabricate(:user, trust_level: TrustLevel[2], group_locked_trust_level: TrustLevel[1])
+  it "promotes a qualifying tl2 user who has a group_granted_trust_level" do
+    group = Fabricate(:group, grant_trust_level: 1)
+    _group_locked_user = Fabricate(:user, trust_level: TrustLevel[2])
+    group.add(_group_locked_user)
```
tgxworld: I think we don't need to use an undescore before the variable name here since we're using the variable.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/2070edf8890ce11b1ccad4ed47c85bff49ffcde6. Thank you @nbianca

PR #6280, opened 2018-08-16
Second phase of renaming User.group_locked_trust_level.

nbianca:
To be merged after #6279.

### DIFF
--- a/db/migrate/20180817175248_remove_group_locked_trust_level.rb
+++ b/db/migrate/20180817175248_remove_group_locked_trust_level.rb
@@ -1,0 +1,5 @@
+class RemoveGroupLockedTrustLevel < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :users, :group_locked_trust_level
+  end
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20180817175248_remove_group_locked_trust_level.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveGroupLockedTrustLevel < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :users, :group_locked_trust_level
```
tgxworld: We can't drop columns here since migrations are ran before changes to the code base is deployed. During the window before the latest code is deployed, the existing app will error out because the `group_locked_trust_level` is no longer present. We have a [helper](https://github.com/discourse/discourse/blob/7cbf2c182f8683abdf29fcfc10b8cb68f7c298ac/lib/migration/column_dropper.rb) to help with dropping of columns which you can implement here.

SamSaffron:
going to close this in favor of a single PR in https://github.com/discourse/discourse/pull/6279

PR #6281, opened 2018-08-17
FIX: Converting PNG to JPEG does not set the correct extension.

tgxworld:
@SamSaffron Can you help me to review this? Thank you!

### DIFF
--- a/spec/lib/upload_creator_spec.rb
+++ b/spec/lib/upload_creator_spec.rb
@@ -1,0 +1,69 @@
+require 'rails_helper'
+
+RSpec.describe UploadCreator do
+  let(:user) { Fabricate(:user) }
+
+  describe '#create_for' do
+    describe 'when upload is not an image' do
+      before do
+        SiteSetting.authorized_extensions = 'txt'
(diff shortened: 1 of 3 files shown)


spec/lib/upload_creator_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      it 'should store the upload with the right extension' do
+        expect do
+          UploadCreator.new(file, filename).create_for(user.id)
+        end.to change { Upload.count }.by(1)
+
+        upload = Upload.last
+
+        expect(upload.extension).to eq('png')
+        expect(File.extname(upload.url)).to eq('.png')
+        expect(upload.original_filename).to eq('png_as.jpg')
```
tgxworld: @SamSaffron I think we should keep the original filename as it is for debugging purposes. I found out about this issue by noticing that the original filename and extension was different.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/a9e502936fb87bf480eb2f38f76de408b3556408

PR #6282, opened 2018-08-17
FEATURE: add Content-Security-Policy header

LeoMcA:
As per https://meta.discourse.org/t/add-csp-nonce-source-support/93300/8?u=leomca, this is my cleaned up proof of concept (where all the tests pass, at least locally, fingers crossed).

@xrav3nz I think this is yours to review, once you've started on the team :)

I'll leave some comments below to explain a few of the things I've done, please ask if there's anything I don't comment on that leaves you asking *"dear god, **why**?"*

### DIFF
--- a/app/assets/javascripts/discourse/lib/load-script.js.es6
+++ b/app/assets/javascripts/discourse/lib/load-script.js.es6
@@ -8,6 +8,7 @@ function loadWithTag(path, cb) {
   let finished = false;
   let s = document.createElement("script");
   s.src = path;
+  s.setAttribute("nonce", Discourse.CSPNonce);
   if (Ember.Test) {
     Ember.Test.registerWaiter(() => finished);
   }
@@ -100,7 +101,16 @@ export default function loadScript(url, opts) {
       ajax({
         url: cdnUrl,
         dataType: opts.css ? "text" : "script",
-        cache: true
+        cache: true,
+        converters: {
+          "text script": text => {
+            var script = document.createElement("script");
+            script.text = text;
+            script.setAttribute("nonce", Discourse.CSPNonce);
+            document.head.appendChild(script).parentNode.removeChild(script);
+            return text;
+          }
+        }
       }).then(cb);
     }
   });
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,8 @@ require_dependency 'topic_query'
 require_dependency 'hijack'
 
 class ApplicationController < ActionController::Base
+  before_action :set_csp_header
+
   include CurrentUser
   include CanonicalURL::ControllerExtensions
   include JsonError
@@ -475,6 +477,28 @@ class ApplicationController < ActionController::Base
 
   def secure_session
     SecureSession.new(session["secure_session_id"] ||= SecureRandom.hex)
+  end
+
+  def set_csp_header
+    nonce = SecureRandom.base64
+    request.env["nonce"] = nonce
+    response.headers["X-Discourse-CSP-Nonce"] = nonce
+
+    if Rails.env.development?
+      # don't do this in prod, vulnerable to Host/X-Forwarded-Host modification
+      base_url = request.protocol + request.host_with_port
+    else
+      base_url = Discourse.base_url
+    end
+    protocol = SiteSetting.force_https? ? "https://" : request.protocol
+    host = base_url.sub("http://", protocol)
+    cdn = GlobalSetting.cdn_url.to_s.sub(/^\/\//, protocol).presence
+
+    response.headers["Content-Security-Policy"] =
+      SiteSetting.content_security_policy
+      .gsub("%{nonce}", nonce)
+      .gsub("%{host}", host)
+      .gsub("%{cdn}", cdn || host)
   end
 
   private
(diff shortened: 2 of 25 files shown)


app/assets/javascripts/discourse/lib/load-script.js.es6
```diff
@@ -100,7 +101,16 @@ export default function loadScript(url, opts) {
       ajax({
         url: cdnUrl,
         dataType: opts.css ? "text" : "script",
-        cache: true
+        cache: true,
+        converters: {
```
LeoMcA: This is necessary for async js loading to work with a relatively restrictive CSP. The code is borrowed almost verbatim from:

https://github.com/discourse/discourse/blob/420739e12c7fbf492f914d2c18ce73d501712c98/vendor/assets/javascripts/jquery.js#L97-L112

(with the inclusion of a nonce attribute)

app/controllers/application_controller.rb
```diff
@@ -477,6 +479,28 @@ def secure_session
...
   end
 
+  def set_csp_header
+    nonce = SecureRandom.base64
+    request.env["nonce"] = nonce
+    response.headers["X-Discourse-CSP-Nonce"] = nonce
+
+    if Rails.env.development?
+      # don't do this in prod, vulnerable to Host/X-Forwarded-Host modification
+      base_url = request.protocol + request.host_with_port
```
LeoMcA: So that the CSP will work in local development, and produce a host of `localhost:3000` rather than just `localhost` (which will cause scripts to fail loading)

app/controllers/qunit_controller.rb
```diff
@@ -6,4 +6,10 @@ class QunitController < ApplicationController
   def index
     raise Discourse::InvalidAccess.new if Rails.env.production?
   end
+
+  # make nonce static and set restrictive csp for testing
+  def set_csp_header
+    request.env["nonce"] = "1234"
+    response.headers["Content-Security-Policy"] = "script-src 'nonce-1234' 'unsafe-eval';"
```
LeoMcA: Use a restrictive CSP in qunit to ensure that nonce-ing is working properly for all (tested) `<script>`s

app/views/common/_discourse_javascript.html.erb
```diff
@@ -46,6 +46,7 @@
     Discourse.LetterAvatarVersion = '<%= LetterAvatar.version %>';
     Discourse.MarkdownItURL = '<%= asset_url('markdown-it-bundle.js') %>';
     Discourse.ServiceWorkerURL = Discourse.Environment != "development" ? '<%= Rails.application.assets_manifest.assets['service-worker.js'] %>' : 'service-worker.js';
+    Discourse.CSPNonce = '<%= nonce %>';
```
LeoMcA: Expose the nonce to already-executing JS, to enable loading of scripts loaded by another script

config/locales/server.en.yml
```diff
@@ -1233,6 +1233,7 @@ en:
     blacklisted_crawler_user_agents: 'Unique case insensitive word in the user agent string identifying web crawlers that should not be allowed to access the site. Does not apply if whitelist is defined.'
     slow_down_crawler_user_agents: 'User agents of web crawlers that should be rate limited in robots.txt using the Crawl-delay directive'
     slow_down_crawler_rate: 'If slow_down_crawler_user_agents is specified this rate will apply to all the crawlers (number of seconds delay between requests)'
+    content_security_policy: "Set Content-Security-Policy header"
```
LeoMcA: Should be updated to point to extra documentation once I've written it

config/site_settings.yml
```diff
@@ -1137,6 +1137,9 @@ security:
     default: 'bingbot'
     list_type: compact
   slow_down_crawler_rate: 60
+  content_security_policy:
+    default: "script-src 'unsafe-eval' 'nonce-%{nonce}' %{host}/mini-profiler-resources/ %{host}/service-worker.js %{cdn}/javascripts/;"
```
LeoMcA: We may want to change this (and include other directives like `image-src`) once I've had a chat with the resident CSP expert within Mozilla.

It currently stands like this because:
* `'unsafe-eval'` - needed to execute Ember
* `'nonce-%{nonce}'` - needed to the vast majority of scripts on the page
* `%{host}/mini-profiler-resources/` - needed to execute the mini profiler, it exists outside of Discourse so we can't nonce it
* `%{host}/service-worker.js` - again something we can't nonce
* `%{cdn}/javscripts/` - needed to load some scripts ace.js loads itself, which is behaviour other libraries might copy

app/controllers/application_controller.rb
```diff
@@ -477,6 +479,28 @@ def secure_session
     SecureSession.new(session["secure_session_id"] ||= SecureRandom.hex)
   end
 
+  def set_csp_header
+    nonce = SecureRandom.base64
+    request.env["nonce"] = nonce
+    response.headers["X-Discourse-CSP-Nonce"] = nonce
```
LeoMcA: Expose the nonce in a special header to allow site admins to set their CSP entirely in nginx (among other things I'm sure this makes possible).

xrav3nz:
Awesome, thanks! I will take a closer look coming September. There is a lot I've yet to learn about CSP. Thanks again 💪

LeoMcA:
No worries, if you've got any questions give me a shout!

PR #6284, opened 2018-08-18
FIX: Minimum username length must be less than any group name.

nbianca:
Fixes [this](https://meta.discourse.org/t/setting-min-username-length-greater-than-5-breaks-granting-and-revoking-admin-moderator-permissions/57079/9).

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -162,9 +162,9 @@ en:
       invalid_choice:
         one: 'You specified the invalid choice %{name}'
         other: 'You specified the invalid choices %{name}'
-      min_username_length_exists: "You cannot set the minimum username length above the shortest username."
+      min_username_length_exists: "You cannot set the minimum username length above the shortest username or group name."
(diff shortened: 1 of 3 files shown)


SamSaffron:
This is a bit trickier than this. 

The idea is that when you attempt to change the site setting we should not allow you to change IF you have users with usernames that are shorter than the site setting. Cause weird stuff happens. 

I am on the fence about weather this site setting (min length) should apply to group names at all, especially since we ship with group names called `team` which are very short.

ZogStriP:
> I am on the fence about weather this site setting (min length) should apply to group names at all,

I think it should apply only to groups

SamSaffron:
I see, since we sort of allow shorter groups here per: https://github.com/discourse/discourse/blob/5a6d1ee25788f308985b8f5801d32fce0e4505bc/app/models/group.rb#L601-L604

I am going to close this for now. I am just too mixed on enforcing it. Change is done properly here, just unsure if we need it.

PR #6287, opened 2018-08-20
FEATURE: Pop revise modal on post edited notification

gdpelican:
Ref: https://meta.discourse.org/t/reason-for-edit-when-edited-a-persons-post/22053/6


gdpelican:
Big caveat here is that the modal pop doesn't work yet when selecting from within a topic, I believe because navigating with `DiscourseURL.routeTo(path)` isn't triggering the model to be re-evaluated for some reason.

Fiddling with `revision: { refreshModel: true }` hasn't yielded the right thing yet there. Input welcome!

gdpelican:
Okay, that issue's been resolved and this should now be ready for review.

Not sure what to make of the test failure, something related to the danger gem?
```
bundler: failed to load command: danger (/home/travis/build/discourse/discourse/vendor/bundle/ruby/2.5.0/bin/danger)
```

gschlager:
The error from danger happens because we do a shallow clone on Travis with a depth of 50 commits. Your branch is 407 commits behind master. Could you please do a rebase so that our build works? Thanks.

gdpelican:
Ok, done.

SamSaffron:
@eviltrout can you review, it is a bit of a hairy change

eviltrout:
This is a bit hairier than I'd like. 

I do think the feature is good, but I wonder about how much of the awkwardness in the PR is due to having it update the URL?

Usually I don't recommend making URLs less useful, but perhaps this should be done without updating the URL? Have the notification open the pop up on click and remove the rest of the code?

gdpelican:
Alright, what do you think of this one @eviltrout ? I've done away with the `?revision` query param and opted for adding an optional callback to the notification-item clicked event.

eviltrout:
Looks a lot better @gdpelican, thanks!

There is one last small change I'd like to request - the key `callback` is not really clear about what it does. I would suggest changing it to something like `afterRouteTransition` or `afterLoaded` or any term you prefer :)

gdpelican:
Ok done.

eviltrout:
Thank you very much @gdpelican

PR #6288, opened 2018-08-20
FEATURE: themes and components split

OsamaSayegh:


app/controllers/admin/themes_controller.rb
```diff
@@ -280,4 +282,10 @@ def log_theme_change(old_record, new_record)
     StaffActionLogger.new(current_user).log_theme_change(old_record, new_record)
   end
 
+  def handle_switch
+    param = theme_params[:component]
+    if ["true", "false"].include?(param.to_s) && param.to_s != @theme.component.to_s
+      @theme.switch_type!
```
SamSaffron: I don't like this API, toggle is not great, it should explicitly either set something to a component or a theme. 

@theme.component = params.to_s == "true"
@theme.save! 

something like that ...

SamSaffron:
overall I am very happy with this change, but we are going to get a ton of feedback here.

SamSaffron:
also can you define something as a component in github in you theme config file.

OsamaSayegh:
@SamSaffron I've replaced the `switch_type!` method with two separate methods `switch_to_theme!` and `switch_to_component!`.

SamSaffron:
OK 🎊  I want to see this in action!

PR #6289, opened 2018-08-21
FEATURE: Clean up `PostReplyKey` records.

tgxworld:
* Default retention of 90 days.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1565,6 +1565,7 @@ en:
     unsubscribe_via_email_footer: "Attach an unsubscribe via email mailto: link to the footer of sent emails"
 
     delete_email_logs_after_days: "Delete email logs after (N) days. 0 to keep indefinitely"
+    disallow_reply_by_email_after_days: "Disallow reply by email after (N) days. 0 to keep indefinitely"
     max_emails_per_day_per_user: "Maximum number of emails to send users per day. 0 to disable the limit"
     enable_staged_users: "Automatically create staged users when processing incoming emails."
     maximum_staged_users_per_email: "Maximum number of staged users created when processing an incoming email."
@@ -2554,7 +2555,7 @@ en:
       text_body_template: |
         We're sorry, but your email message to %{destination} (titled %{former_title}) didn't work.
 
-        For security reasons we only accept replies to original notifications for 90 days. Please [visit the topic](%{short_url}) to continue the conversation.
+        For security reasons we only accept replies to original notifications for %{number_of_days} days. Please [visit the topic](%{short_url}) to continue the conversation.
 
     email_reject_topic_not_found:
       title: "Email Reject Topic Not Found"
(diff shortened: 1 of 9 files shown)


lib/email/receiver.rb
```diff
@@ -171,8 +171,20 @@ def process_internal
 
         raise first_exception if first_exception
 
-        if post = find_related_post(force: true)
-          if Guardian.new(user).can_see_post?(post) && post.created_at < 90.days.ago
```
tgxworld: @ZogStriP I wanted to get your thoughts on this. If I'm understanding our requirements correctly, we only want to raise a `OldDestinationError` when the PostReplyKey is invalid. The assumption here is would mean that an invalid `PostReplyKey` means old though.

tgxworld: Hmm I realized that the assumption where if we can't find a `PostReplyKey` means that the destination is old. Reverted in https://github.com/discourse/discourse/pull/6289/files#diff-0327bfc5fb2dd2a7fd635a096e7fb0daR187

lib/email/receiver.rb
```diff
@@ -171,8 +171,20 @@ def process_internal
...
+            all_destinations.each do |destination|
+              self.class.check_address(destination,
+                throw_on_invalid_post_reply_key: true
+              )
+            end
+
+            false
+          end
+
+          if is_invalid && post.created_at < SiteSetting.delete_post_reply_keys_after_days.days.ago
```
ZogStriP: What if `SiteSetting.delete_post_reply_keys_after_days == 0`?

config/locales/server.en.yml
```diff
@@ -1565,6 +1565,7 @@ en:
     unsubscribe_via_email_footer: "Attach an unsubscribe via email mailto: link to the footer of sent emails"
 
     delete_email_logs_after_days: "Delete email logs after (N) days. 0 to keep indefinitely"
+    delete_post_reply_keys_after_days: "Delete post_reply_keys after (N) days. 0 to keep indefinitely"
```
ZogStriP: I would actually name that setting something like `allow_email_reply_for_days`. 

That is more user friendly than the `reply keys` technicality.

lib/email/receiver.rb
ZogStriP: I don't think we need to specifically check for an invalid post reply key. If the user takes too much time to reply via email, just gently tell them to visit the forum.

config/locales/server.en.yml
tgxworld: > That is more user friendly than the reply keys technicality. 
Good idea :+1:  Let me update the name for the site setting.

lib/email/receiver.rb
tgxworld: ah good catch :+1: Let me fix :stuck_out_tongue_closed_eyes:

lib/email/receiver.rb
tgxworld: @ZogStriP Yea I think we can try that too. I think I'm a little overzealous in getting it 100% accurate for post replies via email.

PR #6290, opened 2018-08-21
UX: show full name on /latest page

techAPJ:
https://meta.discourse.org/t/inconsistent-profile-picture-hover-text/94958

### DIFF
--- a/app/assets/javascripts/discourse/templates/list/posters-column.raw.hbs
+++ b/app/assets/javascripts/discourse/templates/list/posters-column.raw.hbs
@@ -1,5 +1,5 @@
 <td class='posters'>
 {{#each posters as |poster|}}
-<a href="{{poster.user.path}}" data-user-card="{{poster.user.username}}" class="{{poster.extraClasses}}">{{avatar poster avatarTemplatePath="user.avatar_template" usernamePath="user.username" imageSize="small"}}</a>
+<a href="{{poster.user.path}}" data-user-card="{{poster.user.username}}" class="{{poster.extraClasses}}">{{avatar poster avatarTemplatePath="user.avatar_template" usernamePath="user.username" namePath="user.name" imageSize="small"}}</a>
 {{/each}}
 </td>
--- a/app/serializers/basic_user_serializer.rb
+++ b/app/serializers/basic_user_serializer.rb
@@ -1,5 +1,9 @@
 class BasicUserSerializer < ApplicationSerializer
-  attributes :id, :username, :avatar_template
+  attributes :id, :username, :name, :avatar_template
+
+  def name
+    Hash === user ? user[:name] : user.try(:name)
+  end
 
   def include_name?
     SiteSetting.enable_names?
@@ -16,5 +20,4 @@ class BasicUserSerializer < ApplicationSerializer
   def user
     object[:user] || object
   end
-
 end
--- a/app/assets/javascripts/discourse/helpers/user-avatar.js.es6
+++ b/app/assets/javascripts/discourse/helpers/user-avatar.js.es6
@@ -30,6 +30,7 @@ function renderAvatar(user, options) {
   options = options || {};
 
   if (user) {
+    const name = Em.get(user, options.namePath || "name");
     const username = Em.get(user, options.usernamePath || "username");
     const avatarTemplate = Em.get(
       user,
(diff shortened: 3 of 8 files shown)


app/serializers/basic_user_serializer.rb
```diff
@@ -1,5 +1,9 @@
 class BasicUserSerializer < ApplicationSerializer
-  attributes :id, :username, :avatar_template
+  attributes :id, :username, :name, :avatar_template
+
+  def name
+    user[:name] rescue nil
```
SamSaffron: why rescue nil here? this seems not right, also if you look at avatar_template you can see this is a dragon of a nightmare object that is sometimes a Hash and sometimes a User

techAPJ: `rescue nil` was not needed per se, I was just trying to be extra cautious here. Thanks for bringing in my notice about the complex case of this object, I amended the code to consider both a Hash and a User object.

app/serializers/basic_user_serializer.rb
```diff
@@ -1,5 +1,13 @@
 class BasicUserSerializer < ApplicationSerializer
-  attributes :id, :username, :avatar_template
+  attributes :id, :username, :name, :avatar_template
+
+  def name
+    if Hash === object
```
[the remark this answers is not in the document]
techAPJ: Good idea! I made the change but instead of `user&.name` I did `user.try(:name)` because there is a case where user is not nil but the object is of `PostAction` which does not contain `name`.

app/serializers/basic_user_serializer.rb
tgxworld: > rescue nil was not needed per se, I was just trying to be extra cautious here.

Note the `rescue nil` should almost never be used because it rescues any type and all type of errors. I've been bitten multiple times by this where we had typos or the methods that is being called is undefined.

app/serializers/basic_user_serializer.rb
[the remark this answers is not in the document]
tgxworld: > I did user.try(:name) because there is a case where user is not nil but the object is of PostAction which does not contain name.

Hmm it sounds like a bug that the `object` is a `PostAction` since the `UserSerializer` is meant to only serialize `User` objects.

app/serializers/basic_user_serializer.rb
```diff
@@ -1,5 +1,9 @@
 class BasicUserSerializer < ApplicationSerializer
-  attributes :id, :username, :avatar_template
+  attributes :id, :username, :name, :avatar_template
+
+  def name
+    Hash === user ? user[:name] : user.try(:name)
```
tgxworld: I noted that we have no existing specs for the serializer but it'll be good to have this tested so that we can document this  behavior in some way and prevent it from regressing.

app/serializers/basic_user_serializer.rb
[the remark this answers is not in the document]
techAPJ: Looks like it.. here is the failing spec that made me realize this:

```
PostActionUsersController with render always allows you to see your own actions
 Failure/Error: expect(response.status).to eq(200)

   expected: 200
        got: 500

   (compared using ==)
 # ./spec/requests/post_action_users_controller_spec.rb:14:in `block (3 levels) in <top (required)>'
 # ------------------
 # --- Caused by: ---
 # NoMethodError:
 #   undefined method `name' for #<PostAction:0x007fd3c553b8a8>
 #   ./app/serializers/basic_user_serializer.rb:8:in `name'
```

Will investigate more about this case.

app/assets/javascripts/discourse/templates/list/posters-column.raw.hbs
```diff
@@ -1,5 +1,5 @@
 <td class='posters'>
 {{#each posters as |poster|}}
-<a href="{{poster.user.path}}" data-user-card="{{poster.user.username}}" class="{{poster.extraClasses}}">{{avatar poster avatarTemplatePath="user.avatar_template" usernamePath="user.username" imageSize="small"}}</a>
+<a href="{{poster.user.path}}" data-user-card="{{poster.user.username}}" class="{{poster.extraClasses}}">{{avatar poster avatarTemplatePath="user.avatar_template" usernamePath="user.username" namePath="user.name" imageSize="small"}}</a>
```
tgxworld: It'll be good if we can add a QUnit acceptance test for this.

app/serializers/basic_user_serializer.rb
techAPJ: Done.

app/assets/javascripts/discourse/templates/list/posters-column.raw.hbs
techAPJ: Done.

PR #6291, opened 2018-08-21
DEV: makes main scss colors accessible from js

jjaffeux:


ZogStriP:
Does it need to be available "everywhere"? I mean, can't this be an admin-only thing?

eviltrout:
I agree with Regis, that this should only be exposed on the admin side of things.

SamSaffron:
what is the specific use case here?

SamSaffron:
Yeah I say we close this for now, I don't really like it that our JS has this knowledge

PR #6292, opened 2018-08-21
FEATURE: Make initial admins TL1

gdpelican:
This mirrors the behaviour in `rake admin:create`:
```
# admin.rake
if admin.trust_level < 1
  admin.change_trust_level!(1)
end
```


app/controllers/finish_installation_controller.rb
```diff
@@ -27,6 +27,7 @@ def register
       @user.password_required!
 
       if @user.save
+        @user.change_trust_level!(1)
```
ZogStriP: What if the `trust_level` was already > 1?

gdpelican: Good catch; I missed the line about loading the existing user if it exists. Fixed now.

SamSaffron:
seems fine to me, thanks

PR #6294, opened 2018-08-21
UX: Keyboard shortcuts will automatically select hovered post.

nbianca:
Fixes [this](https://meta.discourse.org/t/intructions-on-right-and-feedback-on-wrong-keyboard-shortcut-usage/65794/8).


SamSaffron:
Thanks seems fine!

PR #6295, opened 2018-08-21
UX: Show only the expand icon in lightboxes on mobile.

nbianca:
Fixes [this](https://meta.discourse.org/t/black-bar-appearing-on-images/81608).

### DIFF
--- a/app/assets/stylesheets/mobile/lightbox.scss
+++ b/app/assets/stylesheets/mobile/lightbox.scss
@@ -1,0 +1,21 @@
+.lightbox .meta,
+.lightbox:hover .meta {
+  opacity: 0.8;
+}
+
+.meta {
+  height: 25px;
+  width: 25px;
+  bottom: 0;
+  left: 0;
+
+  .filename,
+  .informations {
+    display: none;
+  }
+
+  .expand {
+    bottom: 1px;
+    right: 6px;
+  }
+}
--- a/app/assets/stylesheets/mobile.scss
+++ b/app/assets/stylesheets/mobile.scss
@@ -16,6 +16,7 @@
 @import "mobile/upload";
 @import "mobile/user";
 @import "mobile/history";
+@import "mobile/lightbox";
 @import "mobile/directory";
 @import "mobile/search";
 @import "mobile/emoji";


app/assets/stylesheets/mobile/lightbox.scss
```diff
@@ -0,0 +1,19 @@
+.lightbox:hover .meta {
```
ZogStriP: There's no `:hover` on mobile.

tshenry: Oddly enough it does seem to have an effect. You wouldn't think it would! If you remove the `:hover` pseudo-class the expand icon will show at all times. If you have `:hover` in place, the expand icon will only show when a finger swipe occurs over a lightboxed image. This is at least how it works on Safari on the iPhone.

app/assets/stylesheets/mobile/lightbox.scss
```diff
@@ -0,0 +1,19 @@
...
+  height: 25px;
+  width: 25px;
+  bottom: 0;
+  left: 0;
+
+  .filename, .informations {
+    display: none;
+  }
+
+  .expand {
```
tshenry: I noticed you changed these values. Out of curiosity, do these new values show the expand icon centered within the grey box on the mobile devices/browsers you have checked? When I look at it on my phone it appears a little offset. Check out the screenshots below. I've got what looks centered for me on the left and then the current values in the PR on the right. If it's not consistent across different mobile browsers, it might be best to find a solution that is.

hover

app/assets/stylesheets/mobile/lightbox.scss
ZogStriP: I know it does something, just not what we expect ;) Anyway, on mobile we always want to show the expand arrows.

app/assets/stylesheets/mobile/lightbox.scss
ZogStriP: Yeah, centered looks better 👍

app/assets/stylesheets/mobile/lightbox.scss
nbianca: Done. The new arrows will be visible on mobile at all times with an opacity of 0.8.

I had to add `.lightbox:hover .meta` because otherwise if the user would hold and slide their finger on the image it would the opacity would become 1. I wanted a consistent behavior.

app/assets/stylesheets/mobile/lightbox.scss
nbianca: Ha. I do not have an iPhone and tested using Chrome DevTools. Your coordinates looked a little off, similar to how the new coordinates look to you.

Since you have first-hand experience, I will use your coordinates. Thank you for your feedback, I appreciate it!

tgxworld:
Let's try this out :

PR #6296, opened 2018-08-22
Fix for tag intersection not populating

gdpelican:
Ref https://meta.discourse.org/t/composer-in-intersection-page-tag-field-not-pre-filled/94780


SamSaffron:
we need a test here since this is already some sort of regression fix.

gdpelican:
Okay I've put in a QUnit for this.

PR #6297, opened 2018-08-22
Use S3 dualstack endpoints

tambry:
See [S3 dualstack documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/dual-stack-endpoints.html).

Also support for the `cn-northwest-1` region.


SamSaffron:
Can you expand on why you want to move us to explicitly choose dualstack here? Do you have v6 only containers and non dualstack is v4 only or something?

tambry:
Non-dualstack S3 domains are IPv4-only.

I encountered this problem while browsing a forum, which used Discourse. While the forum itself worked fine over IPv6, I noticed the user uploads were using a non-IPv6 capable S3 endpoint. Instead of notifying the admins of the particular site, I though it'd be better to fix the problem at the root, so it'll be beneficial to everyone.

SamSaffron:
I am so mixed on this, I don't know for certain what will happen if we shift to dualstack and if we will see any side effects. 

If you are using AWS S3 you really ought to be using a S3 CDN 

You are going to have to wait on this while we discuss this internally it may take a while... cc @andrewschleifer

PR #6303, opened 2018-08-22
Add spec for UserSerializer and UserApiKey.

nbianca:


spec/serializers/user_serializer_spec.rb
```diff
@@ -196,4 +196,33 @@
...
+    let(:user_api_key_0) { Fabricate(:readonly_user_api_key, user: user) }
+    let(:user_api_key_1) { Fabricate(:readonly_user_api_key, user: user) }
+    let(:user_api_key_2) { Fabricate(:readonly_user_api_key, user: user) }
+    let(:user_api_key_3) { Fabricate(:readonly_user_api_key, user: user) }
+    let(:user_api_key_4) { Fabricate(:readonly_user_api_key, user: user) }
+
+    before do
+      freeze_time
+
+      user_api_key_0.update_attributes(last_used_at: 2.days.ago, revoked_at: Time.now)
```
tgxworld: I think we can add the attributes as part of the fabrication. 

```
Fabricate(:readonly_user_api_key, user: user, last_used_at: 7.days.ago)
```

spec/serializers/user_serializer_spec.rb
```diff
@@ -196,4 +196,33 @@
...
+      user_api_key_1.update_attributes(last_used_at: 7.days.ago)
+      user_api_key_2.update_attributes(last_used_at: 1.days.ago)
+      user_api_key_3.update_attributes(last_used_at: 4.days.ago, revoked_at: Time.now)
```
tgxworld: We prefer `Time.zone.now` in our code base to avoid any discrepancy in the time zone being used.

spec/serializers/user_serializer_spec.rb
```diff
@@ -196,4 +196,33 @@
...
+      user_api_key_4.update_attributes(last_used_at: 3.days.ago)
+    end
+
+    it "sorts keys by last used at time" do
+      json = UserSerializer.new(user, scope: Guardian.new(user), root: false).as_json
+
+      expect(json[:user_api_keys].size).to eq(3)
+      expect(json[:user_api_keys][0][:last_used_at]).to eq(user_api_key_1.last_used_at)
```
tgxworld: I think matching by the id is more correct here since the column is unique.

nbianca:
Can you please check again? It should be ready to go. :+1:

tgxworld:
@nbianca Thank you :+1:

PR #6306, opened 2018-08-23
FIX: keep files in order when adding multiple uploads

majakomel:
Works as described here: https://meta.discourse.org/t/uploading-images-in-order-of-filename/95147/27

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-editor.js.es6
@@ -53,15 +53,17 @@ export default Ember.Component.extend({
   _xhr: null,
   shouldBuildScrollMap: true,
   scrollMap: null,
-  uploadPosition: null,
+  uploadFilenamePlaceholder: null,
 
-  @computed("uploadPosition")
-  uploadPlaceholder(uploadPosition) {
-    if (uploadPosition) {
-      return `[${I18n.t("uploading")} (${uploadPosition})]() `;
-    } else {
-      return `[${I18n.t("uploading")}]() `;
-    }
+  @computed("uploadFilenamePlaceholder")
+  uploadPlaceholder(uploadFilenamePlaceholder) {
+    const clipboard = I18n.t("clipboard");
+    const filename = uploadFilenamePlaceholder
+      ? uploadFilenamePlaceholder
+      : clipboard;
+    return `[${I18n.t("uploading_filename", {
+      filename: filename
+    })}]() `;
   },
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -53,10 +53,15 @@ export default Ember.Component.extend({
...
   scrollMap: null,
+  uploadPosition: null,
 
-  @computed
-  uploadPlaceholder() {
-    return `[${I18n.t("uploading")}]() `;
+  @computed("uploadPosition")
+  uploadPlaceholder(uploadPosition) {
+    if (uploadPosition) {
+      return `[${I18n.t("uploading")} (${uploadPosition})]() `;
```
ZogStriP: Maybe the `uploadPosition` should be in the translation? That way, other languages can put it where it's best for them.

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -218,6 +223,17 @@ export default Ember.Component.extend({
     }
   },
 
+  _setUploadPosition(data) {
+    if (data.originalFiles.length > 1) {
+      let position = data.originalFiles.indexOf(data.files[0]) + 1;
```
ZogStriP: This works when you drag & drop multiple files (or select multiple files in the browse modal). How does it handle dragging & dropping **large** files one at a time?

SamSaffron: I really really love the idea of using filename, you would still translate it with a placeholder for filename but it would read way better:

`uploading: test.png` is so much clearer... just make sure that you handle the crazy case where people upload 2 files called "test.png" in one go (by selecting from diff folders for example, not 100% sure you can swing it that easily, but we don't want to break here for that)

also in some cases, cut and paste for example you have no proper filename to use, so maybe use `uploading: clipboard` which reads nice.

SamSaffron:
> Filename is risky because of security issues with some of the characters that can appear in filenames etc

Not too risky cause this simply goes in the text area and then gets swapped out, but possibly worth running it through a regex unicode whitelist, so stuff does not get weird with zero width spaces an strange stuff like that. The failure mode though is not that terrible, cause you can enter whatever letter you want into the composer now anyway.

SamSaffron:
:( I have been a bit slow to review this and now we are going to have to push this to 2.2 it introduces some new critical strings.

SamSaffron:
OK I am merging this in now! thanks @majakomel !

PR #6307, opened 2018-08-23
Add a missing word in admin directions

rizka10:
'min trust level to send **email** messages' is a setting of its own.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1185,7 +1185,7 @@ en:
 
     enable_personal_messages: "Allow trust level 1 (configurable via min trust level to send messages) users to create messages and reply to messages. Note that staff can always send messages no matter what."
     enable_system_message_replies: "Allows users to reply to system messages, even if personal messages are disabled"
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am not sure about this change, besides the whole copy there need a bunch of TLC, it does not read nicely. 

@coding-horror any changes you suggest here?

SamSaffron:
OH this is https://github.com/discourse/discourse/pull/4988 

Hence all the confusion

The real name in English is:

"allow personal messaging to NON forum members" 

I am not sure if this really should be configurable, maybe we should just pull the setting and keep it as TL4/Staff ? 

The feature is useful in "support" kind of scenarios where I want to start a discussion with a user but the user is not quite yet a member of the forum.

SamSaffron:
Removed 2 hard to explain translations here. Settings are hidden

https://github.com/discourse/discourse/commit/4205c528d0b2422180e8b6f489f9e7fc6f0aa828

Settings are now hidden will decide if to totally nuke then in a few months, added a todo

PR #6308, opened 2018-08-24
UX: Improve email testing admin tool.

nbianca:
Fixes [this](https://meta.discourse.org/t/email-test-area-should-warn-if-emails-are-disabled/95269).

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-email-index.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-email-index.js.es6
@@ -28,30 +28,25 @@ export default Ember.Controller.extend({
         sentTestEmail: false
       });
 
-      var self = this;
       ajax("/admin/email/test", {
         type: "POST",
         data: { email_address: this.get("testEmailAddress") }
       })
-        .then(
-          function() {
-            self.set("sentTestEmail", true);
-          },
-          function(e) {
-            if (e.responseJSON && e.responseJSON.errors) {
-              bootbox.alert(
-                I18n.t("admin.email.error", {
-                  server_error: e.responseJSON.errors[0]
-                })
-              );
-            } else {
-              bootbox.alert(I18n.t("admin.email.test_error"));
-            }
-          }
+        .then(response =>
+          this.set("sentTestEmailMessage", response.send_test_email_message)
         )
-        .finally(function() {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/admin/controllers/admin-email-index.js.es6
```diff
@@ -34,8 +34,8 @@ export default Ember.Controller.extend({
         data: { email_address: this.get("testEmailAddress") }
       })
         .then(
-          function() {
-            self.set("sentTestEmail", true);
+          function(response) {
+            self.set("sentTestEmailMessage", response.sentTestEmailMessage);
```
eviltrout: While you're in there, you could improve this to use the `(response) =>` form of functions, which would mean you could use `this` instead of `self`. I'd like to remove the `self` pattern from our code.

app/controllers/admin/email_controller.rb
```diff
@@ -11,7 +11,13 @@ def test
     params.require(:email_address)
     begin
       Jobs::TestEmail.new.execute(to_address: params[:email_address])
-      render body: nil
+      if SiteSetting.disable_emails == "yes"
+        render json: { sentTestEmailMessage: 'admin.email.sent_test_disabled' }
```
ZogStriP: It would be better to directly respond with the translated string.

app/controllers/admin/email_controller.rb
```diff
@@ -11,7 +11,13 @@ def test
     params.require(:email_address)
     begin
       Jobs::TestEmail.new.execute(to_address: params[:email_address])
-      render body: nil
+      if SiteSetting.disable_emails == "yes"
+        render json: { sentTestEmailMessage: I18n.t("admin.email.sent_test_disabled") }
+      elsif SiteSetting.disable_emails == "non-staff" && !User.find_by_email(params[:email_address])&.staff?
+        render json: { sentTestEmailMessage: I18n.t("admin.email.sent_test_disabled_for_non_staff") }
+      else
+        render json: { sentTestEmailMessage: I18n.t("admin.email.sent_test") }
```
tgxworld: Our convention is to use snake_case instead of camelCase on the server side even for keys in the payload. We should add a spec for this as well :smile_cat:

PR #6310, opened 2018-08-24
UX: Improve copying from IP lookup component.

nbianca:
- Removed trailing whitespaces caused by textarea.
- Clicking the Copy button will show "copied" for 2 seconds.

### DIFF
--- a/app/assets/javascripts/admin/components/ip-lookup.js.es6
+++ b/app/assets/javascripts/admin/components/ip-lookup.js.es6
@@ -88,10 +88,13 @@ export default Ember.Component.extend({
           text += `: ${location.phone}\n`;
         }
       }
-      const copyRange = $(`<textarea id="copy-range"></textarea>`);
-      copyRange.text(text);
+      const copyRange = $('<p id="copy-range"></p>');
+      copyRange.html(text.trim().replace("\n", "<br>"));
       $(document.body).append(copyRange);
-      copyText(text, copyRange[0]);
+      if (copyText(text, copyRange[0])) {
+        this.set("copied", true);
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/templates/components/ip-lookup.hbs
```diff
@@ -6,7 +6,11 @@
 {{#if show}}
   <div class="location-box">
     <a class="close pull-right" {{action "hide"}}>{{d-icon "times"}}</a>
-    <a class="btn pull-right" {{action "copy"}}>{{d-icon "copy"}}</a>
+    {{#if copied}}
+      <a class="btn btn-hover pull-right" {{action "copy"}}>{{d-icon "copy"}} {{i18n "ip_lookup.copied"}}</a>
```
ZogStriP: I don't think we need the `action` here.

nbianca: `action` is not necessary indeed, but maybe some people are successful in messing their clipboard in less than 2 seconds and are impatient to copy again.

I have no problems in removing it though. :)

app/assets/javascripts/admin/components/ip-lookup.js.es6
```diff
@@ -88,10 +88,13 @@ export default Ember.Component.extend({
           text += `: ${location.phone}\n`;
         }
       }
-      const copyRange = $(`<textarea id="copy-range"></textarea>`);
-      copyRange.text(text);
+      const copyRange = $(`<p id="copy-range"></p>`);
```
jjaffeux: we dont need ``  here there's no interpolated variable.

PR #6312, opened 2018-08-24
FIX: find tags with non-latin names

majakomel:
Meta: https://meta.discourse.org/t/search-by-tag-not-english-characters/92893

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -369,7 +369,7 @@ class Search
     end
   end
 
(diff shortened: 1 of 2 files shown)


Toxu-ru:
This solution was tested locally a year ago, and it worked for the Russian language. But I'm not sure about other languages.

SamSaffron:
very nice! thanks!

PR #6313, opened 2018-08-24
FEATURE: Let users see their user auth tokens.

nbianca:
This is work in progress, but I appreciate any feedback.

### DIFF
--- a/app/assets/javascripts/discourse/templates/preferences/account.hbs
+++ b/app/assets/javascripts/discourse/templates/preferences/account.hbs
@@ -161,6 +161,74 @@
   </div>
 {{/if}}
 
+{{#if canCheckEmails}}
+  <div class="control-group pref-auth-tokens">
+    <label class="control-label">{{i18n 'user.auth_tokens.title'}}</label>
+    <a {{action "revokeAuthToken"}}>{{d-icon "sign-out"}} {{i18n 'user.auth_tokens.logout'}}</a>
+    {{#each model.user_auth_tokens as |token|}}
+      <div class="perf-auth-token">
+        <div class="row auth-token-summary">
+          <div class="auth-token-label">
+            {{d-icon token.icon}} {{token.device_name}}
+            {{#if token.visible}}
+              <a {{action "toggleToken" token}}>{{d-icon "angle-double-up"}}</a>
+            {{else}}
+              <a {{action "toggleToken" token}}>{{d-icon "angle-double-down"}}</a>
+            {{/if}}
+          </div>
+          <div class="auth-token-value">
+            {{token.client_ip}} <span class="muted">&mdash; {{format-date token.seen_at}}</span>
+          </div>
+        </div>
+        {{#if token.visible}}
+          <div class="auth-token-details">
+            <div class="row">
+              <div class="auth-token-label">{{i18n 'user.auth_tokens.ip_address'}}</div>
+              <div class="auth-token-value">{{token.client_ip}}</div>
+            </div>
+            <div class="row">
+              <div class="auth-token-label">{{i18n 'user.auth_tokens.first_seen'}}</div>
+              <div class="auth-token-value">{{format-date token.created_at}}</div>
+            </div>
+            <div class="row">
+              <div class="auth-token-label">{{i18n 'user.auth_tokens.last_seen'}}</div>
+              <div class="auth-token-value">{{format-date token.seen_at}}</div>
+            </div>
+            <div class="row">
+              <div class="auth-token-label">{{i18n 'user.auth_tokens.operating_system'}}</div>
+              <div class="auth-token-value">{{token.os}}</div>
+            </div>
+          </div>
+        {{/if}}
+      </div>
+    {{/each}}
+  </div>
+{{/if}}
+
+{{#if canCheckEmails}}
+  {{#if model.staff}}
+    <div class="control-group pref-auth-tokens">
+      <label class="control-label">{{i18n 'user.auth_tokens.title_logs'}}</label>
+      {{#if model.user_auth_token_logs}}
+        <table class="table">
+          <tr>
+            <th>{{i18n 'user.auth_tokens.ip_address'}}</th>
+            <th>{{i18n 'user.auth_tokens.created'}}</th>
+            <th>{{i18n 'user.auth_tokens.operating_system'}}</th>
+          </tr>
+          {{#each model.user_auth_token_logs as |token|}}
+            <tr>
+              <td>{{token.client_ip}}</td>
+              <td>{{format-date token.created_at}}</td>
+              <td>{{d-icon token.icon}} {{token.os}}</td>
+            </tr>
+          {{/each}}
+        </table>
+      {{/if}}
+    </div>
+  {{/if}}
+{{/if}}
+
 {{plugin-outlet name="user-preferences-account" args=(hash model=model save=(action "save"))}}
 
 <br/>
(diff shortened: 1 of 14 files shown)


SamSaffron:
@nbianca can you include some screenshots here?

nbianca:
This is what it looks like right now. I have logged from US and CH (fake logins) and then when logged again my other login was marked as suspicious login. If the first two were real logins then in the logs you would have seen two "Suspicious login" entries.

127 0 0 1_3000_u_bianca_preferences_account

SamSaffron:
Looks like the right direction, text wise I would use "Recently Used Devices" instead of "Auth token" language cause it is confusing. Also see what Google do cause its a good idea to emulate this.

image

Not sure if we need to expose an "auth log" to users, just list devices that are logged in, show where they are logged in and if we must they can click through to get IP. I really like that "Google" group by actual device it may be a bit tricky for a V1 but its the right direction.

nbianca:
I made some changes to the UI. I hope you like it. The updated commit is coming soon...

127 0 0 1_3000_u_oscar_preferences_account 1
127 0 0 1_3000_u_oscar_preferences_account

coding-horror:
What action can there be other than "Log In"? I'm unlear.

SamSaffron:
Really liking it @nbianca  , not sure if we need the auth log section. 

I am not that happy relying 3rd party here for IP lookups, cause this is a security feature that is leaking out IP addresses to https://ipinfo.io which at least as an admin you can avoid but now we will hit it for every user.  I would much prefer to ship with IP -> location lookup database, discourse instances can pull the database on demand from some link. 

Let's avoid the IP lookups to ipinfo for now and then in 2.2 we can deal with getting a location db locally so we can do lookups without leaking out info.

app/serializers/user_auth_token_serializer.rb
```diff
@@ -0,0 +1,69 @@
...
+    elsif object.user_agent =~ /Windows/i
+      return 'Windows'
+    elsif object.user_agent =~ /macOS/i
+      return 'macOS'
+    end
+
+    I18n.t('staff_action_logs.unknown')
+  end
+
+  def os_icon
```
jjaffeux: Might be slightly better ?

```
  def os_icon
   case os
   when  'Linux'
      'linux'
   when 'Windows'
      'windows'
    when 'macOS'
      'apple'
    else
      'question'
    end
  end
```

nbianca:
@coding-horror There is also log out. :)
@jjaffeux Thanks for review!
@sam I split the PR in two commits: the first is without any location info and the other one includes IpInfo.

ZogStriP:
@nbianca can you make that in 2 PRs?

coding-horror:
do we really need a "log out" action? I feel we should remove that column entirely. 

I checked Google's security pages and they do not show "log out" so I feel this is adding unnecessary complexity, if a company at Google's scale is only showing login events, certainly we can follow suit?

image

coding-horror:
definitely not seeing "log out" here at all.. and I think it's simpler and less confusing to have less data on the screen

image

nbianca:
127 0 0 1_3000_u_oscar_preferences_account 4

Please note that the Authentication Logs are visible only to staff.

app/serializers/user_auth_token_serializer.rb
```diff
@@ -0,0 +1,88 @@
...
+      'iOS'
+    else
+      I18n.t('staff_action_logs.unknown')
+    end
+  end
+
+  def device_name
+    case object.user_agent
+    when /Android/i
+      'Android Device'
```
ZogStriP: These should probably be translatable.

ZogStriP:
Ok, let's try it 👍

app/assets/javascripts/discourse/templates/preferences/account.hbs
```diff
@@ -161,6 +161,74 @@
   </div>
 {{/if}}
 
+{{#if canCheckEmails}}
```
tgxworld: @nbianca I think we need to add a simple QUnit acceptance test here for this feature.

app/controllers/users_controller.rb
```diff
@@ -1097,6 +1098,19 @@ def revoke_account
     end
   end
 
+  def revoke_auth_token
```
tgxworld: @nbianca This controller action needs to be tested.

app/controllers/users_controller.rb
```diff
@@ -1097,6 +1098,19 @@ def revoke_account
...
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
+
+    UserAuthToken.where(user_id: user.id).destroy_all
```
tgxworld: Because we're not checking for return value by `destroy_all` code here. You probably want to do 

```
UserAuthToken.where(user_id: user.id).each(&:destroy!)
```

app/controllers/users_controller.rb
```diff
@@ -1097,6 +1098,19 @@ def revoke_account
...
+
+    MessageBus.publish "/file-change", ["refresh"], user_ids: [user.id]
```
tgxworld: Is this intended to reload the page automatically? If so, we probably want to do a redirect back to the user preferences page.

app/models/user_auth_token.rb
```diff
@@ -11,8 +11,19 @@ class UserAuthToken < ActiveRecord::Base
   # used when token did not arrive at client
   URGENT_ROTATE_TIME = 1.minute
 
+  USER_ACTIONS = ['generate']
+
   attr_accessor :unhashed_auth_token
 
+  before_destroy do
+    UserAuthToken.log(action: 'destroy',
```
tgxworld: We should add a unit test for this callback.

app/serializers/user_auth_token_serializer.rb
```diff
@@ -0,0 +1,88 @@
...
+  def include_action?
+    object.has_attribute?(:action)
+  end
+
+  def client_ip
+    object.client_ip.to_s
+  end
+
+  def include_seen_at?
+    object.has_attribute?(:seen_at)
```
tgxworld: @nbianca Are we passing objects that are not instances of `UserAuthToken` to the serializer? I'll like to clarify why we need to check if the object has the `seen_at` attribute.

app/controllers/users_controller.rb
ZogStriP: Destroying all the `UserAuthToken` will log the user out. Not sure redirecting back to the user preferences page is useful here?

app/serializers/user_auth_token_serializer.rb
ZogStriP: I think @nbianca meant `object.seen_at.present?` ;)

app/controllers/users_controller.rb
nbianca: Noted. Thanks!

app/serializers/user_auth_token_serializer.rb
nbianca: Nope. This is right. I am using the serializer for `UserAuthTokenLogs`, which does not have a `seen_at` column. :)

tgxworld: @nbianca Usually we want each serializer to have a single reponsibility. That means `UserAuthTokenSerializer` should only be expected to receive `UserAuthToken` objects for serialization. The trouble when a serializer is made to support different kind of objects is that we end up having to maintain conditionals which in turn complicates testing. I understand that the purposes of supporting different objects here is meant to avoid repeating code which is great :+1:  However, I think extracting the common logic into a module will have been easier to reason about and still allow us to have `UserAuthTokenSerializer` and `UserAuthTokenLogSerializer`. Do have a look through the changes in https://github.com/discourse/discourse/commit/7256b3bf689aa156086a397ca5264188a3d912ed to see how you can easily include class and instance methods easily via a module next time.

PR #6315, opened 2018-08-26
Add plugin outlet for after site setting filter

gdpelican:
Hello, I'm looking to put something here:

screen_shot_2018-08-25_at_8 23 42_pm

:)

### DIFF
--- a/app/assets/javascripts/admin/templates/site-settings.hbs
+++ b/app/assets/javascripts/admin/templates/site-settings.hbs
@@ -5,6 +5,7 @@
     {{text-field value=filter placeholderKey="type_to_filter" class="no-blur"}}
     {{d-button action="clearFilter" label="admin.site_settings.clear_filter"}}
   </div>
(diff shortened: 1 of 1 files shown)


SamSaffron:
hmmm what in particular are you looking to put there? Can you attach a screenshot of what your plugin is adding?

SamSaffron:
I understand not wanting to bury this, but it definitely does not belong in site settings, this should go in the plugin section.

gdpelican:
Sigh. I really don't want to do that; it makes for a really confusing
experience when I have to point folks to several different places to manage
their settings for a single plugin. Is there another solution here?

SamSaffron:
thanks, that is a far better UI imo.

gdpelican:
@SamSaffron If you were willing to put a plugin outlet here, in `site-settings-category.hbs`:
screen shot 2018-08-27 at 8 35 16 pm

it would save me doing a very, very ugly thing in my plugin to get that to appear sans outlet ;)

PR #6316, opened 2018-08-26
FIX: Reset email preferences controller.

nbianca:
Fixes [this](https://meta.discourse.org/t/cant-change-multiple-users-emails-any-more/95180).


SamSaffron:
Thanks! Probably worth adding a test here

PR #6317, opened 2018-08-26
FIX: Ignore OneBox blacklisted domains.

nbianca:
Fixes [this](https://meta.discourse.org/t/disable-discourse-from-crawling-links/69778/9).

### DIFF
--- a/lib/final_destination.rb
+++ b/lib/final_destination.rb
@@ -26,7 +26,7 @@ class FinalDestination
     "HTTPS_DOMAIN_#{domain}"
   end
 
-  attr_reader :status, :cookie, :status_code
+  attr_reader :status, :cookie, :status_code, :ignored
 
   def initialize(url, opts = nil)
     @url = url
@@ -36,7 +36,15 @@ class FinalDestination
     @force_get_hosts = @opts[:force_get_hosts] || []
     @opts[:max_redirects] ||= 5
     @opts[:lookup_ip] ||= lambda { |host| FinalDestination.lookup_ip(host) }
-    @ignored = [Discourse.base_url_no_prefix] + (@opts[:ignore_redirects] || [])
+
+    @ignored = @opts[:ignore_hostnames] || []
+    [Discourse.base_url_no_prefix].concat(@opts[:ignore_redirects] || []).each do |url|
+      url = uri(url)
+      if url.present? && url.hostname
+        @ignored << url.hostname
+      end
+    end
+
     @limit = @opts[:max_redirects]
     @status = :ready
     @http_verb = @force_get_hosts.any? { |host| hostname_matches?(host) } ? :get : :head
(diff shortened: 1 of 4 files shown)


lib/oneboxer.rb
```diff
@@ -250,9 +250,11 @@ def self.local_user_html(url, route)
 
   def self.external_onebox(url)
     Rails.cache.fetch(onebox_cache_key(url), expires_in: 1.day) do
-      fd = FinalDestination.new(url, ignore_redirects: ignore_redirects, force_get_hosts: force_get_hosts)
+      ignored = SiteSetting.onebox_domains_blacklist.split("|")
+
+      fd = FinalDestination.new(url, ignore_redirects: ignore_redirects, ignore_hostnames: ignored, force_get_hosts: force_get_hosts)
       uri = fd.resolve
-      return blank_onebox if uri.blank? || SiteSetting.onebox_domains_blacklist.include?(uri.hostname)
+      return blank_onebox if uri.blank? || ignored.map { |x| uri.hostname.match?(x) }.any?
```
tgxworld: Let's avoid variable names like `x` because it makes the code harder to grok on first read.

lib/final_destination.rb
```diff
@@ -36,7 +36,15 @@ def initialize(url, opts = nil)
     @force_get_hosts = @opts[:force_get_hosts] || []
     @opts[:max_redirects] ||= 5
     @opts[:lookup_ip] ||= lambda { |host| FinalDestination.lookup_ip(host) }
-    @ignored = [Discourse.base_url_no_prefix] + (@opts[:ignore_redirects] || [])
+
+    @ignored = @opts[:ignore_hostnames] || []
+    ([Discourse.base_url_no_prefix] + (@opts[:ignore_redirects] || [])).each do |url|
```
tgxworld: Let's update this to use `concat` instead of allocating 2 extra arrays .

lib/final_destination.rb
```diff
@@ -36,7 +36,15 @@ def initialize(url, opts = nil)
...
+      url = uri(url)
+      if url.present? && url.hostname
```
tgxworld: Let's add a test case for this so that it becomes clearer why this condition is required.

lib/final_destination.rb
```diff
@@ -131,18 +139,18 @@ def resolve
...
     end
 
+    unless validate_uri
+      log(:warn, "FinalDestination could not resolve URL (invalid URI): #{@uri}") if @verbose
+      return nil
+    end
+
     @ignored.each do |host|
-      if hostname_matches?(host)
+      if @uri && @uri.hostname.match?(host)
```
tgxworld: We can use the lonely operator here `@uri&.hostname&.match?(host)`

lib/oneboxer.rb
tgxworld: Is there a reason we need this check this against `ignored` again when `ignored` is already passed into `FinalDestination` as `ignore_hostnames`?

tgxworld:
@nbianca  Looks good to me :+1:

PR #6325, opened 2018-08-28
FIX: Generate webhook payloads before destroy events

vinothkannans:


SamSaffron:
@tgxworld can you review?

tgxworld:
@vinothkannans I think we should use `around_destroy` here instead - https://stackoverflow.com/questions/4998553/rails-around-callbacks. The documentation around this is quite poor but it basically allows us to do something like this.

```
def around_destroy
  payload = generate_webhook_payload
  yield
  queue_webhook(payload)
end 
```

app/models/concerns/has_web_hooks.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  included do
+    around_destroy :enqueue_web_hook
+  end
+
+  def enqueue_web_hook
+    type = self.class.name.underscore.to_sym
+    payload = WebHook.generate_payload(type, self)
+    yield
+    WebHook.enqueue_hooks(type, "#{self.class.name.underscore}_destroyed".to_sym,
```
tgxworld: This can just be `WebHook.enqueue_hooks(type, "#{type}_destroyed".to_sym,`

app/models/user.rb
```diff
@@ -1067,6 +1068,10 @@ def set_automatic_groups
   end
 
   def email
+    # https://github.com/discourse/discourse/blob/af061efcf5c89a33655c52c0ae36292c2eacf225/spec/jobs/fix_primary_emails_for_staged_users_spec.rb#L21
+    # as per above spec primary email can be nil
+    return primary_email&.email if staged? && !active
```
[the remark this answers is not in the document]
vinothkannans: I think `primary_email` can't be nil. But only in the above commented spec url we making it as nil by removing `user_emails` table. I feel like it's incorrect. So I like to ask @tgxworld whether it's possible to keep using `primary_email.email` by fixing the old spec.

Sure, I will remove those extra lines before merging this :wink:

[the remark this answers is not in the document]
ZogStriP: Have you tried setting up a user webhook and deleting a user? Pretty sure the email records are deleted before the webhook triggers ;)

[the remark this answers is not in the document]
vinothkannans: > Pretty sure the email records are deleted before the webhook triggers

Yes, preventing that is the job for this PR.

app/models/concerns/has_web_hooks.rb
```diff
@@ -0,0 +1,17 @@
+module HasWebHooks
+  extend ActiveSupport::Concern
+
+  included do
+    around_destroy :enqueue_web_hook
+  end
+
+  def enqueue_web_hook
```
tgxworld: I think the name here should be `enqueue_destroy_web_hook`

app/models/concerns/has_web_hooks.rb
```diff
@@ -0,0 +1,17 @@
+module HasWebHooks
```
tgxworld: Maybe `HasDestroyedWebHook`? I feel like `HasWebHooks` is too generic here.

app/models/user.rb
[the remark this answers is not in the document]
tgxworld: > So I need to ask @tgxworld whether it's possible to keep using primary_email.email by fixing the old spec.

Yea I think we should fix the old spec here.

app/models/web_hook.rb
```diff
@@ -41,56 +41,60 @@ def self.active_web_hooks(type)
...
-        payload: WebHookPostSerializer.new(post,
-          scope: self.guardian,
-          root: false
-        ).to_json
+        payload: payload
       )
     end
   end
 
+  def self.generate_payload(type, object, serializer = nil)
```
tgxworld: :+1:

lib/post_destroyer.rb
```diff
@@ -45,6 +45,14 @@ def initialize(user, post, opts = {})
   end
 
   def destroy
+    payload = WebHook.generate_payload(:post, @post)
+    topic = @post.topic
+
+    if @post.is_first_post? && topic
+      topic_view = TopicView.new(topic.id, Discourse.system_user)
+      topic_payload = WebHook.generate_payload(:topic, topic_view, WebHookTopicViewSerializer)
+    end
```
tgxworld: Since `Post` and `Topic` are only soft destroyed, perhaps we can only generate the payload after it has been trashed?

vinothkannans: After a post is soft deleted by author we changing the raw with `(post withdrawn by author...` text. I think it's good to keep it like all other "destroyed" webhooks.

vinothkannans:
@tgxworld done. I hope it's mergable now.

tgxworld:
LGTM :+1:

PR #6331, opened 2018-08-29
Validation of params should restrict to max int

gschlager:


spec/requests/list_controller_spec.rb
```diff
@@ -11,11 +11,6 @@
   end
 
   describe '#index' do
-    it "doesn't throw an error with a negative page" do
-      get "/#{Discourse.anonymous_filters[1]}", params: { page: -1024 }
-      expect(response.status).to eq(200)
-    end
```
gschlager: @tgxworld Was there a reason for allowing negative page numbers?

tgxworld: I think it was just to prevent errors being thrown in our logs. I think we can change it so that negative numbers will just be converted to 0?

SamSaffron: I disagree, our API should be returning HTTP 400 this is an invalid request not a 200, and definitely should not be returning a 500 which signifies to "crappy security scanners" that they are up to something so keep trying till you get a SQL injection jackpot

gschlager:
I removed the special handling for the `page` param. It only allows values between 1 and max int. Negative numbers, arrays and whatnot raise a status 400.

SamSaffron:
Thanks yeah I support this change 👍

PR #6335, opened 2018-08-29
FEATURE: Resolve IP locations.

nbianca:


SamSaffron:
This is going to need a rebase, I am slotting this work for 2.2, no rush at all getting it checked in. 

My pref here is to use: http://lite.ip2location.com/ or https://dev.maxmind.com/geoip/geoip2/downloadable/ . The tricky thing here is "how do we bundle it" ? But the easiest thing here will be to have this be a "value add"... create a rake task to download the db and then bundle it in our docker image. 

Having the DB does unlock some other features so it is strategic.

nbianca:
Hi @SamSaffron,

Those were my concerns too. I was going to discuss these with @ZogStriP on Monday, but your plan sounds really good. I like Maxmind better because we can just use [yhirose/maxminddb](https://github.com/yhirose/maxminddb). What do you think?

config/application.rb
```diff
@@ -201,6 +201,9 @@ def config.database_configuration
     $redis = DiscourseRedis.new
     Logster.store = Logster::RedisStore.new(DiscourseRedis.new)
 
+    require 'discourse_ip_info'
+    $ipinfo = DiscourseIpInfo.new
```
SamSaffron: I really really prefer to avoid globals here ... it can simply be a static hanging off DiscourseIpInfo... 

DiscourseIpInfo.lookup('10.0.0.1') ... then you cache it there. Also we need to make sure this does not explode if a maxmind db is missing, so have it do a file exists prior to loading it up.

lib/discourse_ip_info.rb
```diff
@@ -0,0 +1,27 @@
+require_dependency 'maxminddb'
+
+class DiscourseIpInfo
+
+  def initialize
+    @mmdb_filename = File.join(Rails.root, 'vendor', 'data', 'GeoLite2-City.mmdb')
+
+    begin
+      @mmdb = MaxMindDB.new(@mmdb_filename)
```
SamSaffron: lets use the low memory reader per:  MaxMindDB::LOW_MEMORY_FILE_READER ... and then if we really want to we can put an LRU cache in front, do a quick bench to see how fast the low mem reader is.

SamSaffron: We should ship with a small test mmdb file so we can ensure lookups remain good and so on across gem updates. Plus it will allow us to properly test this class

nbianca:
```
                               user     system      total        real
Low memory                 4.910000   1.980000   6.890000 (  6.890183)
Low memory (10% cached)    4.890000   2.180000   7.070000 (  7.071249)
Low memory (50% cached)    0.020000   0.010000   0.030000 (  0.022946)
Low memory (100% cached)   0.010000   0.010000   0.020000 (  0.023665)
Low memory (4 threads)    25.520000  24.060000  49.580000 ( 29.471233)
Default                    2.010000   0.000000   2.010000 (  2.011072)
Default (10% cached)       2.120000   0.000000   2.120000 (  2.121219)
Default (50% cached)       0.030000   0.000000   0.030000 (  0.032013)
Default (100% cached)      0.030000   0.000000   0.030000 (  0.031720)
Default (4 threads)        2.070000   0.000000   2.070000 (  2.079891)
```

Testing methodology:
* list of 40 IPs
* the test was looking up the 40 ip 50 times

It looks like the low memory reader with a cache does pretty good, however in my case the IPs repeated themselves quite a lot. Otherwise, the results would have been a lot different.

To be honest, I do not think that the default reader is that bad. 64MBs of data is not a lot (considering the recommended machine has enough memory) and the cache would not do much in a real world situation when IPs happen to more unique.

config/application.rb
nbianca: There is already [a check](https://github.com/discourse/discourse/pull/6335/files#diff-9c2f27b3cc391041329980be3a007b8fR8) if MaxMindDb is missing (it actually covers for the case when the file is corrupt too).

SamSaffron:
What is the cost per uncached lookup? (1ms, 10ms, 100ms?) Keep in mind n
the big scheme these lookups will be increadibly rare.

nbianca:
@SamSaffron 3ms (low-memory reader) vs 1ms (default reader)

SamSaffron:
Can you run memory_profiler on this? How many objects does this add in
default mode? I am very much leaning toward the 3ms per lookup department,
this is such a rare page and keeping memory usage down for us is very
critical

lib/discourse_ip_info.rb
```diff
@@ -0,0 +1,39 @@
+require_dependency 'maxminddb'
+
+class DiscourseIpInfo
+  include Singleton
+
+  def initialize
+    begin
+      @mmdb_filename = File.join(Rails.root, 'vendor', 'data', 'GeoLite2-City.mmdb')
+      @mmdb = MaxMindDB.new(@mmdb_filename, MaxMindDB::LOW_MEMORY_FILE_READER)
+    rescue
```
ZogStriP: This is too broad. What specific exceptions are you trying to rescue?

lib/discourse_ip_info.rb
```diff
@@ -0,0 +1,39 @@
+require_dependency 'maxminddb'
+
+class DiscourseIpInfo
+  include Singleton
+
+  def initialize
+    begin
+      @mmdb_filename = File.join(Rails.root, 'vendor', 'data', 'GeoLite2-City.mmdb')
+      @mmdb = MaxMindDB.new(@mmdb_filename, MaxMindDB::LOW_MEMORY_FILE_READER)
```
ZogStriP: Like @SamSaffron suggested, have you run `memory_profiler` on this? 
How many objects does this add in low memory vs default mode?

lib/discourse_ip_info.rb
```diff
@@ -0,0 +1,39 @@
...
+      country_code: result.country.iso_code,
+      region: result.subdivisions.most_specific.name,
+      city: result.city.name,
+    }
+  end
+
+  def get(ip)
+    return {} unless @mmdb
+
+    @cache[ip] || @cache[ip] = lookup(ip)
```
ZogStriP: This can be written as `@cache[ip] ||= lookup(ip)`  ;)

app/controllers/users_controller.rb
```diff
@@ -1108,7 +1108,11 @@ def revoke_auth_token
     user = fetch_user_from_params
     guardian.ensure_can_edit!(user)
 
-    UserAuthToken.where(user_id: user.id).each(&:destroy!)
+    if !SiteSetting.log_out_strict && params[:token_id]
```
ZogStriP: This needs to be tested.

app/controllers/posts_controller.rb
```diff
@@ -123,11 +123,24 @@ def user_posts_feed
...
+    respond_to do |format|
+      format.rss do
+        @posts = posts
+        @title = "#{SiteSetting.title} - #{I18n.t("rss_description.user_posts", username: user.username)}"
+        @link = "#{Discourse.base_url}/u/#{user.username}/activity"
+        @description = I18n.t("rss_description.user_posts", username: user.username)
+        render 'posts/latest', formats: [:rss]
+      end
+
+      format.json do
```
ZogStriP: This needs to be tested.

lib/discourse_ip_info.rb
nbianca: There are plenty of exceptions which should be rescued and are not very intuitive. Some of those are:

- `Errno::ENOENT` if file does not exist;
- `NoMethodError` if the MaxMindDB file is empty (0 bytes, caused by calling `rindex` on `nil`);
- `RuntimeError` if the MaxMindDB is invalid (raised by `maxminddb` gem).

I would not like MaxMindDb to cause any problem if `maxminddb` causes any issues so that's why I adopted a 'rescue-all' attitude.

lib/discourse_ip_info.rb
nbianca: Default reader:
```
# Total allocated: 5178841 bytes (118759 objects)
# Total retained:  121235 bytes (2115 objects)
```

Low memory reader:
```
# Total allocated: 62659823 bytes (118756 objects)
# Total retained:  57733098 bytes (2112 objects)
```

There is a very small difference in the number of objects because there is a big one that holds the whole MaxMind database in memory.

ZogStriP: Yeah, we definitely want the low memory version ;) We can take the 3ms since it's such a low traffic endpoint anyways

lib/discourse_ip_info.rb
nbianca: :cold_sweat:

lib/discourse_ip_info.rb
nbianca: It depends. I believe, on average, this is going to be called for three times for every user that accesses his preferences page (which in my opinion is something more than a low traffic endpoint). Anyway, some of the calls will be cached anyway and after all, 3ms it's not that bad.

SamSaffron:
Overall I feel we are ready to merge this, @ZogStriP  anything blocking from your side? 

One change I would like to see is some sort of nice degradation when maxmind db is missing, @nbianca can you confirm nothing looks or acts horrible if the maxmind db is missing? 

Once we merge this we should quickly merge in the docker base image change and then get this pushed onto meta using the new base image.

PR #6345, opened 2018-08-30
FEATURE: Backup locations

gschlager:
- [x] New `backup_location` site setting which allows choosing between "Local" and "Amazon S3"
- [x] `/admin/backups` lists backups from local storage or S3 according to site setting
- [x] Admins can download backups from local storage and S3 (via presigned URL, valid for 15 seconds)
- [x] Restore works for local storage and S3
- [x] Automatic and manual backups work for local storage and S3
- [x] Uploads work for local storage and S3 (via presigned URL, valid for 6 hours)
- [x] Existing specs are green
- [x] More specs for local storage and S3
- [x] Add migration for site settings and cleanup old, unused settings
- [x] Make sure that script `discourse backup` and `discourse restore` still work
- [x] Error handling when `s3_backup_bucket` is blank
- [x] Configure CORS on `s3_backup_bucket`


config/locales/server.en.yml
```diff
@@ -1345,13 +1349,13 @@ en:
     maximum_backups: "The maximum amount of backups to keep on disk. Older backups are automatically deleted"
     automatic_backups_enabled: "Run automatic backups as defined in backup frequency"
     backup_frequency: "The number of days between backups."
-    enable_s3_backups: "Upload backups to S3 when complete. IMPORTANT: requires valid S3 credentials entered in Files settings."
     s3_backup_bucket: "The remote bucket to hold backups. WARNING: Make sure it is a private bucket."
     s3_endpoint: "The endpoint can be modified to backup to an S3 compatible service like DigitalOcean Spaces or Minio. WARNING: Use default if using AWS S3"
     s3_force_path_style: "Enforce path-style addressing for your custom endpoint. IMPORTANT: Required for using Minio uploads and backups."
     s3_disable_cleanup: "Disable the removal of backups from S3 when removed locally."
```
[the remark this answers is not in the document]
SamSaffron: lets keep this s3 specific for now.

Gemfile.lock
```diff
@@ -44,20 +44,20 @@ GEM
     arel (9.0.0)
     ast (2.4.0)
     aws-eventstream (1.0.1)
-    aws-partitions (1.92.0)
-    aws-sdk-core (3.21.2)
+    aws-partitions (1.104.0)
```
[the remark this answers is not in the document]
gschlager: Done in https://github.com/discourse/discourse/commit/fb9e3e64237e30effb0f845ad6669c2744532936

SamSaffron:
Maybe deploy on a digital ocean so we can have a bit of a play with it?

app/assets/javascripts/discourse/components/backup-uploader.js.es6
```diff
@@ -0,0 +1,55 @@
+import { ajax } from "discourse/lib/ajax";
+import computed from "ember-addons/ember-computed-decorators";
+import UploadMixin from "discourse/mixins/upload";
+
+export default Em.Component.extend(UploadMixin, {
+  tagName: "span",
+
+  @computed("uploading", "uploadProgress")
+  uploadButtonText(uploading) {
```
ZogStriP: What are you not passing `uploadProgress` as an argument of the `uploadButtonText` function?

```javascript
@computed("uploading", "uploadProgress")
uploadButtonText(uploading, progress) {
  return uploading ? I18n.t("admin.backups.uploading_progress", { progress }) : I18n.t("admin.backups.upload.label");
}
```

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -225,6 +225,7 @@ export function validateUploadedFiles(files, opts) {
 }
 
 export function validateUploadedFile(file, opts) {
+  if (opts.skipValidation) return true;
```
ZogStriP: Might be worth adding a test here.

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
+require_dependency "backup_restore/backup_store"
+require_dependency "s3_helper"
+
+module BackupRestore
+  class S3BackupStore < BackupStore
+    DOWNLOAD_URL_EXPIRES_AFTER_SECONDS = 15
```
ZogStriP: Might want to conditionally set (`||=`) these constants in order to prevent warnings when reloading.

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      presigned_url(obj, :put, UPLOAD_URL_EXPIRES_AFTER_SECONDS)
+    end
+
+    private
+
+    def unsorted_files
+      @s3_helper
+        .list
+        .select { |obj| obj.key.match?(/\.t?gz$/i) }
```
ZogStriP: Just curious: would it be possible to do the filtering using the S3 API?

gschlager: Unfortunately that's not possible. The API allows only filtering by prefix.

app/assets/javascripts/discourse/lib/utilities.js.es6
gschlager: Done.

app/assets/javascripts/discourse/mixins/upload.js.es6
```diff
@@ -15,34 +15,46 @@ export default Em.Mixin.create({
     return {};
   },
 
+  calculateUploadUrl() {
+    return (
+      Discourse.getURL(this.getWithDefault("uploadUrl", "/uploads")) +
```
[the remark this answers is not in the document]
gschlager: I didn't change `Discourse.getURL` yet because when I tried it real quick I got a bunch of failing tests. I'll take a closer look tomorrow.

app/assets/javascripts/discourse/components/backup-uploader.js.es6
```diff
@@ -0,0 +1,51 @@
...
+
+  validateUploadedFilesOptions() {
+    return { skipValidation: true };
+  },
+
+  uploadDone() {
+    this.sendAction("done");
+  },
+
+  calculateUploadUrl() {
```
tgxworld: Do we need to define this if `UploadMixin` is always included?

app/controllers/admin/backups_controller.rb
```diff
@@ -46,10 +50,12 @@ def cancel
...
   def email
-    if backup = Backup[params.fetch(:id)]
+    store = BackupRestore::BackupStore.create
+
+    if store.file(params.fetch(:id)).present?
       Jobs.enqueue(:download_backup_email,
-        user_id: current_user.id,
-        backup_file_path: url_for(controller: 'backups', action: 'show')
+                   user_id: current_user.id,
+                   backup_file_path: url_for(controller: 'backups', action: 'show')
```
tgxworld: Indentation here is off :stuck_out_tongue_winking_eye:

app/controllers/admin/backups_controller.rb
```diff
@@ -59,28 +65,35 @@ def email
...
-        render template: 'admin/backups/show.html.erb', layout: 'no_ember', status: 422
+
+      if store.remote?
+        redirect_to backup.source
       else
-        render body: nil, status: 404
+        headers['Content-Length'] = File.size(backup.source).to_s
+        send_file backup.source
       end
+    elsif @error
```
tgxworld: I think we can immediately return this once `@error` is set?

app/controllers/admin/backups_controller.rb
```diff
@@ -173,6 +186,21 @@ def upload_backup_chunk
...
+  def create_upload_url
+    params.require(:filename)
+    filename = params.fetch(:filename)
+
+    return render_error("backup.backup_file_should_be_tar_gz") unless valid_extension?(filename)
+    return render_error("backup.invalid_filename") unless valid_filename?(filename)
+
+    store = BackupRestore::BackupStore.create
+    upload_url = store.generate_upload_url(filename)
+  rescue BackupRestore::BackupStore::BackupFileExists
```
tgxworld: Error handling is really slow in Ruby so we want to avoid using this pattern for control flow. Perhaps we can try using `catch..throw` here?

lib/backup_restore/backup_store.rb
```diff
@@ -0,0 +1,76 @@
+module BackupRestore
+  # @abstract
+  class BackupStore
+    class BackupFileExists < RuntimeError; end
+    class StorageError < RuntimeError; end
+
+    # @return [BackupStore]
+    def self.create(opts = {})
+      case SiteSetting.backup_location
+      when 'local'
```
tgxworld: We should use the constants defined for this.

lib/backup_restore/backup_store.rb
```diff
@@ -0,0 +1,76 @@
...
+        require_dependency "backup_restore/s3_backup_store"
+        BackupRestore::S3BackupStore.new(opts)
+      end
+    end
+
+    # @return [Array<BackupFile>]
+    def files
+      unsorted_files
+        .sort_by { |file| file.last_modified }
+        .reverse
```
tgxworld: Hmm sorting and then reversing seems weird. Will `unsorted_files.sort_by { |file| -file.last_modified }`?

lib/backup_restore/backup_store.rb
```diff
@@ -0,0 +1,76 @@
...
+        BackupRestore::LocalBackupStore.new(opts)
+      when 's3'
+        require_dependency "backup_restore/s3_backup_store"
+        BackupRestore::S3BackupStore.new(opts)
+      end
+    end
+
+    # @return [Array<BackupFile>]
+    def files
+      unsorted_files
```
tgxworld: Perhaps we should memoize this?

lib/backup_restore/local_backup_store.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def file(filename, include_download_source: false)
+      path = path_from_filename(filename)
+      create_file_from_path(path, include_download_source) if File.exists?(path)
+    end
+
+    def delete_file(filename)
+      path = path_from_filename(filename)
+      FileUtils.remove_file(path, force: true) if File.exists?(path)
+      DiskSpace.reset_cached_stats
```
tgxworld: I think we only need to remove this if the path exists

lib/backup_restore/local_backup_store.rb
```diff
@@ -0,0 +1,61 @@
...
+    def download_file(filename, destination, failure_message = "")
+      path = path_from_filename(filename)
+      Discourse::Utils.execute_command('cp', path, destination, failure_message: failure_message)
+    end
+
+    private
+
+    def unsorted_files
+      Dir.glob(File.join(@base_directory, "*.{gz,tgz}"))
+        .map { |filename| create_file_from_path(filename) }
```
tgxworld: I think we can use `map!` here to avoid an extra array allocation.

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
+require_dependency "backup_restore/backup_store"
+require_dependency "s3_helper"
+
+module BackupRestore
+  class S3BackupStore < BackupStore
+    DOWNLOAD_URL_EXPIRES_AFTER_SECONDS ||= 15
+    UPLOAD_URL_EXPIRES_AFTER_SECONDS ||= 21600 # 6 hours
```
tgxworld: Usually for large numbers, writing it as `21_600` is much easier to read than `21600`

lib/backup_restore/s3_backup_store.rb
tgxworld: Let's use `select!` here in case the list is large.

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    private
+
+    def unsorted_files
+      @s3_helper
+        .list
+        .select { |obj| obj.key.match?(/\.t?gz$/i) }
+        .map { |obj| create_file_from_object(obj) }
+    rescue Aws::Errors::ServiceError => e
+      Rails.logger.warn("Failed to list backups from S3: #{e.message.presence || e.class.name}")
```
tgxworld: Do we need the `presence` on `e.message`?

lib/s3_helper.rb
```diff
@@ -63,17 +63,17 @@ def ensure_cors!
     end
 
     unless rule
-      puts "installing CORS rule"
+      rules = [{
+                 allowed_headers: ["Authorization"],
+                 allowed_methods: ["GET", "HEAD"],
+                 allowed_origins: ["*"],
+                 max_age_seconds: 3000
+               }] if rules.nil?
```
tgxworld: The indentation used here is off too.

spec/requests/admin/backups_controller_spec.rb
```diff
@@ -4,17 +4,40 @@
...
 
   before do
     sign_in(admin)
+    SiteSetting.backup_location = BackupLocationSiteSetting::LOCAL
   end
 
   after do
     $redis.flushall
+
+    @paths&.each { |path| File.delete(path) if File.exists?(path) }
```
tgxworld: From the looks of it, `@paths` will never be `nil`

spec/requests/admin/backups_controller_spec.rb
```diff
@@ -88,66 +98,48 @@
     it "uses send_file to transmit the backup" do
       begin
         token = EmailBackupToken.set(admin.id)
-        path = File.join(Backup.base_directory, backup_filename)
-        File.open(path, "w") { |f| f.write("hello") }
-
-        Backup.create_from_filename(backup_filename)
+        path = backup_path(backup_filename)
```
tgxworld: It looks like this line is not used.

spec/requests/admin/backups_controller_spec.rb
```diff
@@ -284,15 +268,13 @@
...
   describe "#email" do
-    let(:backup_filename) { "test.tar.gz" }
-    let(:backup) { Backup.new(backup_filename) }
-
     it "enqueues email job" do
-      Backup.expects(:[]).with(backup_filename).returns(backup)
+      create_backup_files(backup_filename)
 
-      Jobs.expects(:enqueue).with(:download_backup_email,
+      Jobs.expects(:enqueue).with(
```
tgxworld: Not related but since we're at it. Let's test this to right way :wink:

app/controllers/admin/backups_controller.rb
```diff
@@ -1,20 +1,24 @@
...
         store_preloaded("operations_status", MultiJson.dump(BackupRestore.operations_status))
         store_preloaded("logs", MultiJson.dump(BackupRestore.logs))
         render "default/empty"
       end
       format.json do
-        render_serialized(Backup.all, BackupSerializer)
+        begin
+          store = BackupRestore::BackupStore.create
+          render_serialized(store.files, BackupFileSerializer)
+        rescue BackupRestore::BackupStore::StorageError => e
```
tgxworld: `try..catch` instead here as well?

app/models/backup_file.rb
```diff
@@ -0,0 +1,25 @@
+class BackupFile
+  include ActiveModel::SerializerSupport
+
+  attr_accessor :filename,
```
tgxworld: hmm do we need to write access?

lib/backup_restore/local_backup_store.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def delete_file(filename)
+      path = path_from_filename(filename)
+      FileUtils.remove_file(path, force: true) if File.exists?(path)
+      DiskSpace.reset_cached_stats
+    end
+
+    def download_file(filename, destination, failure_message = "")
+      path = path_from_filename(filename)
+      Discourse::Utils.execute_command('cp', path, destination, failure_message: failure_message)
```
tgxworld: :+1:

lib/backup_restore/restorer.rb
```diff
@@ -195,8 +195,15 @@ def sidekiq_has_running_jobs?
...
     def copy_archive_to_tmp_directory
-      log "Copying archive to tmp directory..."
-      Discourse::Utils.execute_command('cp', @source_filename, @archive_filename, failure_message: "Failed to copy archive to tmp directory.")
+      if @store.remote?
+        log "Downloading archive to tmp directory..."
+        failure_message = "Failed to download archive to tmp directory."
+      else
+        log "Copying archive to tmp directory..."
+        failure_message = "Failed to copy archive to tmp directory."
+      end
```
tgxworld:
```
failure_message =
  if @store.remote?
  else
  end
```
We can do something like this too.

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
...
+      presigned_url(obj, :put, UPLOAD_URL_EXPIRES_AFTER_SECONDS)
+    end
+
+    private
+
+    def unsorted_files
+      @s3_helper
+        .list
+        .select { |obj| obj.key.match?(/\.t?gz$/i) }
+        .map { |obj| create_file_from_object(obj) }
```
tgxworld: `map!` too

lib/backup_restore/s3_backup_store.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    def ensure_cors!
+      @s3_helper.ensure_cors!(
+        [{
+           allowed_headers: ["*"],
+           allowed_methods: ["PUT"],
+           allowed_origins: [Discourse.base_url_no_prefix],
+           max_age_seconds: 3000
+         }]
```
tgxworld: Indentation is off here.

lib/site_settings/validations.rb
```diff
@@ -53,4 +53,13 @@ def validate_enable_s3_uploads(new_val)
     validate_error :s3_upload_bucket_is_required if new_val == "t" && SiteSetting.s3_upload_bucket.blank?
   end
 
+  def validate_backup_location(new_val)
+    return unless new_val == BackupLocationSiteSetting::S3
+    validate_error(:s3_backup_requires_s3_settings, setting_name: "s3_backup_bucket") if SiteSetting.s3_backup_bucket.blank?
+
+    unless SiteSetting.s3_use_iam_profile
+      validate_error(:s3_backup_requires_s3_settings, setting_name: "s3_access_key_id") if SiteSetting.s3_access_key_id.blank?
+      validate_error(:s3_backup_requires_s3_settings, setting_name: "s3_secret_access_key") if SiteSetting.s3_secret_access_key.blank?
```
tgxworld: Just curious, are we testing this validation anywhere in the test suite?

tgxworld:
@gschlager The PR looks good to me. I didn't test the S3 backup locally but we can do so once this has been delpoyed to meta.

lib/backup_restore/s3_backup_store.rb
gschlager: `list` returns up to 1000 elements. `Aws::S3::ObjectSummary::Collection` doesn't support `select!`. I could convert it to an array, but I guess that's even more inefficient than using `select`. I'd say in 99% of all cases `select` won't remove any elements, so it should be fine.

lib/backup_restore/backup_store.rb
gschlager: I think it's fine the way it is. Currently `files` and `latest_file` isn't used more than once per `BackupStore` instance. Memoizing would just make it more complex, since we would need to invalidate or update it when files are added or deleted.

lib/backup_restore/s3_backup_store.rb
gschlager: see https://github.com/discourse/discourse/pull/6345#discussion_r224063467

app/assets/javascripts/discourse/mixins/upload.js.es6
```diff
@@ -15,34 +16,46 @@ export default Em.Mixin.create({
...
+  },
+
   _initialize: function() {
     const $upload = this.$(),
-      csrf = Discourse.Session.currentProp("csrfToken"),
-      uploadUrl = Discourse.getURL(
-        this.getWithDefault("uploadUrl", "/uploads")
-      ),
-      reset = () => this.setProperties({ uploading: false, uploadProgress: 0 });
+      reset = () => this.setProperties({ uploading: false, uploadProgress: 0 }),
```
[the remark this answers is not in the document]
gschlager: I'm sorry, can you clarify? Do you mean I should declare a method for `reset` instead of using a `const`?

lib/backup_restore/restorer.rb
gschlager: It might be me, but I find it a lot easier to read with the explicit assignments. I'd like to keep it that way.

lib/backup_restore/s3_backup_store.rb
gschlager: Yes, it's needed. Some AWS exceptions (e.g. `Aws::S3::Errors::BadRequest`) have an empty string as `message`. That's why I'm logging the class name in that case.

app/assets/javascripts/discourse/mixins/upload.js.es6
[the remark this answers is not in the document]
tgxworld: O sometime back we started to prefer 

const a = b;
const b = c;

Over 

const a = b,
  b = c;

lib/site_settings/validations.rb
gschlager: Nope. As far as I can tell we don't have any specs for that file.

lib/backup_restore/restorer.rb
tgxworld: :+1:

tgxworld:
Thank you @gschlager :+1: Merged in https://github.com/discourse/discourse/commit/c29a4dddc1694e2805944f4d5fd0f4fd4514092e

PR #6346, opened 2018-08-30
Add shadowed_by_global for S3 uploads

pfaffman:
I'm still not clear why these need to be added (or aren't there already?), but without these changes these variables seem not to matter if they're in `config/discourse.conf`.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -932,6 +932,7 @@ files:
   enable_s3_uploads:
     default: false
     client: true
+    shadowed_by_global: true
   s3_use_iam_profile: false
   s3_access_key_id: 
     default: ''
@@ -947,6 +948,7 @@ files:
   s3_upload_bucket:
     default: ''
     regex: '^[a-z0-9\-\/]+$' # can't use '.' when using HTTPS
+    shadowed_by_global: true
   s3_endpoint:
     default: 'https://s3.amazonaws.com'
     regex: '^https?:\/\/.+[^\/]$'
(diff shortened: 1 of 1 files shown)


tgxworld:
> but without these changes these variables seem not to matter if they're in config/discourse.conf

Do you mean `config/discourse_defaults.conf`?

pfaffman:
I mean that in production if I create `DISCOURSE_XXX` variables in `app.yml` they end up in `/var/www/discourse/config/discourse.conf`, but they are ignored (which is why I submitted a very similar PR that you accepted recently). `discourse_defaults.conf` says:
```
# DO NOT EDIT THIS FILE
# If you need to make changes create a file called discourse.conf in this directory with your changes
# On import this file will be imported using ERB
```
The variables I've added shadows for here are listed in `discourse_defaults.conf`, but even if they are set, they seem to have no effect. The changes I submitted make them work.

It doesn't make sense that this PR would be needed, as I'm sure that you guys are setting these variables such that they are hidden from the admin interface, but I can't figure out what I'm doing wrong.

My solution must be ham-handed, but it's the only thing I've gotten to work.

tgxworld:
For config in https://github.com/discourse/discourse/blob/master/config/discourse_defaults.conf#L171-L178, `DISCOURSE_S3_BUCKET` will be accessible via `GlobalSetting.s3_bucket`.

SamSaffron:
Can we discuss this on meta, the changes themselves are trivial we will not need a PR here, just need to get to an agreement that the change is desired vs not desired.

PR #6347, opened 2018-08-30
UX: Various improvements to the themes system

OsamaSayegh:


SamSaffron:
This adds a fair number of strings, so holding off till 2.2 to merge

SamSaffron:
OK, this is pretty enormous, but its a change we need and the vast majority is client side.

Merging this in now, thanks heaps @OsamaSayegh

PR #6348, opened 2018-08-30
FEATURE: Enforce two-factor authentication.

nbianca:


nbianca:
I wrote "some users" because it can be "all users" on "staff-only users". I changed that now.

SamSaffron:
This is a fairly complex change, I would like to see some integration tests here for the controller behavior change and a UX test for the client

@featheredtoast  can you also review this please?

ZogStriP:
We definitely need some integration tests for the server and UX test for the client ;)

app/assets/javascripts/discourse/routes/preferences-second-factor.js.es6
```diff
@@ -13,5 +13,46 @@ export default RestrictedUserRoute.extend({
...
+      const currentUser = controller.get('currentUser');
+      const siteSettings = controller.get('siteSettings');
+
+      if (transition.targetName === "preferences.second-factor") {
+        // User wants to set up second factor authentcation.
+        return true;
+      }
+
+      if (!currentUser) {
+        // No user logged in.
```
tgxworld: These comments are actually not that useful since the code pretty much describes what the behavior is. I'll do away with the comments in this case.

SamSaffron:
@coding-horror has pushed this back to 2.3, so any work on this should be reasonably low priority for now

nbianca:
Added banner. Thanks, @featheredtoast!

I think it's ready to be merged. <3

screenshot 2018-11-04 at 21 13 23

featheredtoast:
Looking good!

I'd want to tweak the copy a little bit, that's short and to the point, along the lines of "You are required to enable two factor authentication before accessing this site."

SamSaffron:
I like this feature and would be happy to see it merged (needs a rebase though) but one big sticking point for me is the additional `redirect_to_second_factor_if_required` filter. It makes our code long term more brittle and it is one more filter to remember to bypass in all sorts of places. 

Can this just be a part of `redirect_to_login_if_required` ? This would mean we touch so many less files.

ZogStriP:
This can be merged once https://github.com/discourse/discourse/pull/7010 is merged.

ZogStriP:
@nbianca this needs a rebase ;)

SamSaffron:
@nbianca  can you merge this in first thing when you start your day?

neemiasvf:
I can't wait for this one hehe

SamSaffron:
this is good to merge, go ahead @nbianca once you are online.

neemiasvf:
https://tenor.com/PylC.gif

TonnesM:
> +enforce_second_factor: "Forces users to enable second factor authentication. Select 'all' to enforce it to all users. Select 'staff' to enforce it to staff users only."

Shouldn’t the first line use "Force" as well as "two factor authentication"?

PR #6353, opened 2018-08-31
UX: simplifies UI for listing user auth tokens

jjaffeux:


jjaffeux:
@awesomerobot I'm not so sure about all of this. The existing design is mostly showing nothing more when expanding, so a simple table looks better.

However, with more thought:
1) we will have to add location at some point
2) the existing design works better on mobile

I let you decide what's best here.

SamSaffron:
this is conflicting now that I merged @nbianca's changes

SamSaffron:
closing for now

jjaffeux:
@SamSaffron Yes what was at the beginning a simple lifting ends up being more of a "we should think more about this UI"

PR #6354, opened 2018-08-31
FIX: Miscellaneous issues related to user auth tokens.

nbianca:
* Addresses some concerns raised during [review](https://github.com/discourse/discourse/pull/6313).
* Fixed dark theme.
* Made the whole user token clickable.
* Hopefully fixed for good browser detection from user string.


nbianca:
image

image

PR #6355, opened 2018-08-31
FIX: escape regex chars when searching site texts

majakomel:


SamSaffron:
Looks good to me, awesome!

PR #6358, opened 2018-09-03
DEV: Allow plugins to whitelist specific user custom_fields for editing

davidtaylorhq:
Required due to recent custom_fields changes. This is presenting a few support issues on Meta for various broken plugins. Once this is merged I will make a few small pull requests to the most popular plugins that have been affected.

### DIFF
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -1517,12 +1517,43 @@ describe UsersController do
           end
 
           context "custom_field" do
-            it "does not update the custom field" do
-              put "/u/#{user.username}.json", params: { custom_fields: { test: :it } }
+            before do
+              plugin = Plugin::Instance.new
+              plugin.register_editable_user_custom_field :test2
+            end
+
+            after do
+              User.plugin_editable_user_custom_fields.clear
+            end
+
+            it "only updates allowed user fields" do
+              put "/u/#{user.username}.json", params: { custom_fields: { test1: :hello1, test2: :hello2 } }
 
               expect(response.status).to eq(200)
-              expect(user.custom_fields["test"]).to be_blank
+              expect(user.custom_fields["test1"]).to be_blank
+              expect(user.custom_fields["test2"]).to eq("hello2")
             end
+
+            it "works alongside a user field" do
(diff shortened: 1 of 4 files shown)


spec/requests/users_controller_spec.rb
```diff
@@ -1517,11 +1517,26 @@ def post_user
           end
 
           context "custom_field" do
-            it "does not update the custom field" do
-              put "/u/#{user.username}.json", params: { custom_fields: { test: :it } }
+            before do
+              plugin = Plugin::Instance.new
+              plugin.register_editable_user_custom_field :test2
```
SamSaffron: only minor thing here that is missing is a cleanup here of the static on user

spec/requests/users_controller_spec.rb
```diff
@@ -1517,11 +1517,26 @@ def post_user
...
+            end
+
+            it "only updates allowed user fields" do
+              put "/u/#{user.username}.json", params: { custom_fields: { test1: :hello1, test2: :hello2 } }
+
+              expect(response.status).to eq(200)
```
SamSaffron: hmmm ... so the user send invalid params and we return a 200, I do not really agree with this ... we should return an 400 here.

davidtaylorhq: Fixing this will mean some changes in the client. 

At the moment every custom_field associated with the user is sent to the server when you save preferences. For example, if I hit save on my Meta profile, it sends up `custom_fields[staff_notes_count]: 1`. The server ignores this, so it is fine. If we return an error, it will break things.

The solution is add a 'editable user custom fields whitelist' to the client plugin API as well, but I wonder whether we should save that for after the 2.1 release?

SamSaffron: OK we can keep this "safe" change for 2.1 and do the proper fix for 2.2. We should not be asking the server to save fields it has no permission to save.

SamSaffron:
also I want a test case for when custom user custom fields is a blank array (in fact for extra safety if nothing is set I would just bypass passing this stuff along) just in case we mistakenly are whitelisting an arbitrary array

SamSaffron:
OK this looks safe to me for 2.1 merging! thanks!

PR #6359, opened 2018-09-05
REFACTOR: use tables instead of custom fields for polls

ZogStriP:
This changes the "data layer" of the poll plugin from custom fields to tables.

I also vastly simplified the rules for updating a poll. We can now only update a poll within the edit window. Any votes casted during that period will be deleted if the poll is changed in any way.

I also added the possibility to show the results on `vote` or when the poll is `closed`.

I paid attention not to introduce any **N+1** but I'm not a :robot:.

⚠️ The `MigratePollsData` query is a monster 👻 .


plugins/poll/plugin.rb
```diff
@@ -398,37 +411,72 @@ def voters
...
+  on(:user_destroyed) do |user|
+    PollVote.where(user: user).delete_all
+  end
 
-    polls.each do |_, poll|
-      poll["options"].each do |option|
-        option.delete("voter_ids")
-      end
+  add_to_class(:topic_view, :polls) do
+    @polls ||= begin
```
SamSaffron: what we can do here is piggy back on a custom field. So you would store a single custom field on a post saying "hey I have a poll" and then if that hits go to the polls table. This means that we would only run the complex poll query if one of the posts has a poll. This is much more efficient cause we run the "fetch these custom fields query" for multiple plugins. 99% of posts have no polls.

plugins/poll/db/migrate/20180820080623_migrate_polls_data.rb
```diff
@@ -0,0 +1,82 @@
+class MigratePollsData < ActiveRecord::Migration[5.2]
+  def up
+    execute <<~SQL
```
SamSaffron: I am somewhat worried about this, in particular how does this handle **corrupt** data, we need to make sure that there is no case where we run this migration and just explode. I would kind of prefer if this iterated through all polls and perform the conversion in ruby just cause of the safety aspect.

coding-horror:
Iterating through all polls is fine, there's no way any site will have a
pathological number of polls (say more than 1000) .. not really possible in
my opinion.

plugins/poll/app/models/poll.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+class Poll < ActiveRecord::Base
+  # because we want to use the 'type' column and don't want to use STI
+  self.inheritance_column = nil
+
+  belongs_to :post
+
+  has_many :poll_options, dependent: :destroy
+  has_many :poll_votes, dependent: :delete_all
```
tgxworld: Maybe we don't need this? `poll_options` owns the `poll_votes` and has a `dependent: :delete_all`

plugins/poll/app/models/poll.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  validates :type, inclusion: { in: %w(regular multiple number) }
+  validates :status, inclusion: { in: %w(open closed) }
+  validates :results, inclusion: { in: %w(always vote closed) }
```
tgxworld: Perhaps we can store the strings above in an enum so that we only have to store an integer in the columns?

plugins/poll/app/serializers/poll_serializer.rb
```diff
@@ -0,0 +1,52 @@
...
+    object.max.present? &&
+    (object.number? || object.multiple?)
+  end
+
+  def include_step?
+    object.step.present? && object.number?
+  end
+
+  def options
+    object.poll_options.map { |o| PollOptionSerializer.new(o, root: false).as_json }
```
tgxworld: Perhaps we can declare the `has_many` relationship in the serializer instead?

plugins/poll/db/migrate/20180820073549_create_polls_tables.rb
```diff
@@ -0,0 +1,44 @@
+class CreatePollsTables < ActiveRecord::Migration[5.2]
+  def change
+    create_table :polls do |t|
+      t.integer :post_id, null: false
```
tgxworld: Perhaps we can just do `t.references :post, index: true, foreign_key: true`?

plugins/poll/db/migrate/20180820080623_migrate_polls_data.rb
tgxworld: We'll also need to think about how we're going to handle the deployment process. After this migration has run, the old custom fields are still writable which means that a poll created or vote made before the app servers have been deployed with the new code will be lost. Iterating through the polls and doing the migration in Ruby might be safer too as we can acquire a lock around the poll while the migration is happening.

plugins/poll/spec/controllers/posts_controller_spec.rb
```diff
@@ -191,26 +188,14 @@
...
             put :update, params: {
               id: post_id, post: { raw: new_option }
             }, format: :json
 
             expect(response.status).to eq(200)
             json = ::JSON.parse(response.body)
-            expect(json["post"]["polls"]["poll"]["options"][1]["html"]).to eq("C")
-          end
-
-          it "staff can change the options" do
```
tgxworld: I think we still need this test case to ensure that staff members are able to edit options.

plugins/poll/spec/controllers/posts_controller_spec.rb
```diff
@@ -228,54 +213,19 @@
...
-            }, format: :json
-
-            expect(response.status).to eq(200)
-
-            json = ::JSON.parse(response.body)
-            expect(json["post"]["polls"]["poll"]["options"][1]["html"]).to eq("C")
-            expect(json["post"]["polls"]["poll"]["voters"]).to eq(8)
-            expect(json["post"]["polls"]["poll"]["options"][0]["votes"]).to eq(8)
-            expect(json["post"]["polls"]["poll"]["options"][1]["votes"]).to eq(0)
-          end
```
tgxworld: Are these tests no longer required as well?

plugins/poll/spec/lib/polls_updater_spec.rb
```diff
@@ -1,428 +1,117 @@
 require 'rails_helper'
 
 describe DiscoursePoll::PollsUpdater do
-  let(:user) { Fabricate(:user) }
-
-  let(:post_with_two_polls) do
```
tgxworld: Some test cases have been removed here and the test cases are still relevant to me IMO.

plugins/poll/spec/lib/votes_updater_spec.rb
```diff
@@ -1,94 +0,0 @@
-require 'rails_helper'
-
-describe DiscoursePoll::VotesUpdater do
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/6359/files#r215156653

plugins/poll/spec/controllers/posts_controller_spec.rb
tgxworld: My bad I just read that we simplified the update logic. Is there any previous context of why we do so?

plugins/poll/app/models/poll.rb
ZogStriP: This allows me to count the # of vote(r)s directly on the `Poll` object instead of going through all the `PollOption`.

plugins/poll/db/migrate/20180820080623_migrate_polls_data.rb
ZogStriP: @SamSaffron I was worried about that too and tried my best to handle this in the query. But I guess doing it in Ruby is safer. Will change.

@tgxworld I personally am fine losing votes during the migration and the odds of people making a poll while the migration is running are very very low. I'd much rather lose a bit of data than add a lot more complexity to this migration.

plugins/poll/app/models/poll.rb
ZogStriP: I have grown to hate integer enums. I feel the storage space "gained" is not worth the clarity of

`SELECT * FROM polls WHERE status = 'closed'` 

Which is IMHO 10000x better than 

`SELECT * FROM polls WHERE status = 0`

But I sure can create SQL types to enforce the values on these columns.

tgxworld: A single digit integer takes up 2 bytes while the shortest string here takes up 5 bytes. The longest string in this case takes up 9 bytes. On top of that, we don't know if we'll eventually introduce `type/status/results` that are longer in length. The other benefit is that indexing an integer column is usually faster and smaller in size. The gains are quite significant IMO and is a tradeoff that I think we should take and has already taken in other places of our app. 

On the Ruby side of things, using an enum allows us to avoid specifying a common string in multiple places. I really like the pattern of `Poll.types[:regular]` or `Poll.status[:closed]` so that we only need to declare the string once and avoid any possibility of making a typo in multiple places.

plugins/poll/db/migrate/20180820080623_migrate_polls_data.rb
tgxworld: > during the migration and the odds of people making a poll while the migration is running are very very low

It isn't the duration of the migration that results in the window of the data loss. The window of data loss is open when the migration has run but the app servers have not been redeployed with the latest code. That window can be quite significant depending on the different build processes we have in place.

plugins/poll/app/models/poll.rb
ZogStriP: Sure, feel free to do the changes ;)

plugins/poll/plugin.rb
ZogStriP: I've done that in https://github.com/discourse/discourse/pull/6359/commits/c121b64cee41a33f9c5b5fda1fcf85826968fccf#diff-89c344c997376afa256e678abfef12b0R441

plugins/poll/db/migrate/20180820080623_migrate_polls_data.rb
ZogStriP: New migration is in. It's still a monster. Effectively slower but should be more safe.

plugins/poll/spec/controllers/posts_controller_spec.rb
ZogStriP: Yeah, none are relevant anymore since we don't merge votes anymore (we delete them when the poll has changed)

plugins/poll/spec/lib/votes_updater_spec.rb
ZogStriP: We don't update the votes anymore. We delete them whenever a poll is changed.

SamSaffron:
@ZogStriP  we got to merge this thing in, we are sitting on it for too long, how brave are you feeling?

coding-horror:
Yes we need this moved along.

My only fear is massive breakage on biz or ent tiers, so let's make sure it
is super stable on std and meta/dev first

SamSaffron:
@ZogStriP can you merge first thing when your day starts?

tgxworld:
This PR isn't ready to be merged. There are [some specs](https://github.com/discourse/discourse/pull/6359/files#r215156653) which I don't think should have been removed.

PR #6361, opened 2018-09-05
Added Access-Control-Allow-Methods to CORS configuration

CheshireVillageSoftware:
Browsers are not consistent about which HTTP methods are allowed via CORS requests by default. Particularly DELETE or PUT. See discussion here: https://stackoverflow.com/questions/20478312/default-value-for-access-control-allow-methods

This patch explicitly configures which methods are allowed: POST, PUT, GET, OPTIONS, DELETE


MichaelMalony:
CLA signed

PR #6362, opened 2018-09-05
Added support for CORS origin wildcards.

CheshireVillageSoftware:
Allows CORS http origins to be configured with * wildcards.

Driving use case is mobile hybrid client applications (cordova, Ionic, phonegap). These typically run a server on localhost with a random port. This patch allows http origin to be defined as http://localhost:*

Could also be used to support any subdomain as well: http://*.example.com


MichaelMalony:
CLA signed

SamSaffron:
This is a complex change that is going to need at least a spec or 2 to confirm it does not regress anything or break under certain cases. 

Also can you fix the code indenting there?

CheshireVillageSoftware:
Gave this some more thought and it is probably a niche use case so I went ahead and made plugin instead: https://github.com/CheshireVillageSoftware/discourse-cors-wildcard

I also updated the branch with fixed whitespace and re-worked the logic a bit.

If you don't think this makes sense as a core feature you can go ahead and close this.

Otherwise I can try to put together some tests. It looks like there is currently only one test that touches cors in hijack_spec.rb. I'm guessing that doesn't make sense for general cors specs though.

MichaelMalony:
Gave this some more thought and it is probably a niche use case so I went ahead and made plugin instead: https://github.com/CheshireVillageSoftware/discourse-cors-wildcard

I also updated the branch with fixed whitespace and re-worked the logic a bit.

If you don't think this makes sense as a core feature you can go ahead and close this.

Otherwise I can try to put together some tests. It looks like there is currently only one test that touches cors in hijack_spec.rb. I'm guessing that doesn't make sense for general cors specs though.

SamSaffron:
Yeah lets close this for now, I feel we would need deep testing here and we probably need to do something with message bus as well.

PR #6365, opened 2018-09-05
add secondary smtp settings

mudasirraza:
This will enable secondary smtp settings which can be used for separated smtp settings for multisite tenants.

Relevant discussion:
https://meta.discourse.org/t/multisite-installation-with-seperated-smtp-emails/29999


SamSaffron:
I would much prefer if you open (or hijack) one of the topics on meta here for this. the bottom line is that I am kind of ok with shadowed by global settings for this stuff, but it gets very very very tricky we don't even have a super clean pattern for S3. 

I am pretty strongly against the concept of secondary SMTP, but yeah I guess in some multisites we could allow per site SMTP. 

But I am very very worried about all this increased complexity.

PR #6367, opened 2018-09-06
CHORE: Remove unneeded if/else condition in post#cook

gdpelican:
Just a couple little cleanups I tripped over in the post model 🌷

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -255,7 +255,7 @@ class Post < ActiveRecord::Base
     options[:cook_method] = cook_method
     options[:user_id] = self.user.id if self.user
     # At trust level 3, we don't apply nofollow to links
-    options[:omit_nofollow] = true   if add_nofollow?
+    options[:omit_nofollow] = true   if omit_nofollow?
 
     cooked = post_analyzer.cook(raw, options)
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
This looks fine to me, but going to wait till 2.2 to merge (probably in a week and a half)

app/models/post.rb
```diff
@@ -253,17 +253,11 @@ def cook(raw, opts = {})
 
     options = opts.dup
     options[:cook_method] = cook_method
+    options[:user_id] = self.user.id if self.user
+    # At trust level 3, we don't apply nofollow to links
+    options[:omit_nofollow] = true   if add_nofollow?
```
tgxworld: This should be `unless` instead of `if`

gdpelican: Oops, fixed 😅

gdpelican:
This probably would be a little bit nicer with something like
```
cooked = post_analyzer.cook raw, opts.dup.merge(
  cook_method:   cook_method,
  user_id:       self.user&.id,
  omit_nofollow: omit_nofollow?
).compact
```

but didn't want to do more changes than necessary, it's just a little tidy-up.

SamSaffron:
I am just worried that this enters the "refactoring" type PR which we explicitly do not allow, it already generated so much work reviewing it (I have looked at this multiple times and so has TGX). I am so sorry but I am closing for now.

PR #6368, opened 2018-09-06
FEATURE: Add SiteSetting for s3_configure_tombstone_policy and skip p…

rishabhnambiar:
…olicy generation if turned off

- [x] Add spec

I'm not sure if changing that return statement is enough for disabling policy generation, I haven't managed to test it properly yet. (thanks Docker setup!)

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1347,6 +1347,7 @@ en:
     s3_backup_bucket: "The remote bucket to hold backups. WARNING: Make sure it is a private bucket."
     s3_endpoint: "The endpoint can be modified to backup to an S3 compatible service like DigitalOcean Spaces or Minio. WARNING: Use default if using AWS S3"
     s3_force_path_style: "Enforce path-style addressing for your custom endpoint. IMPORTANT: Required for using Minio uploads and backups."
+    s3_configure_tombstone_policy: "Enable automatic deletion policy for tombstone uploads. IMPORTANT: If disabled, no space will be reclaimed after uploads are deleted."
     s3_disable_cleanup: "Disable the removal of backups from S3 when removed locally."
     backup_time_of_day: "Time of day UTC when the backup should occur."
     backup_with_uploads: "Include uploads in scheduled backups. Disabling this will only backup the database."
--- a/lib/s3_helper.rb
+++ b/lib/s3_helper.rb
@@ -131,6 +131,7 @@ class S3Helper
   end
 
   def update_tombstone_lifecycle(grace_period)
+    return if !SiteSetting.s3_configure_tombstone_policy
     return if @tombstone_prefix.blank?
     update_lifecycle("purge_tombstone", grace_period, prefix: @tombstone_prefix)
   end
--- a/spec/components/s3_helper_spec.rb
+++ b/spec/components/s3_helper_spec.rb
@@ -2,13 +2,12 @@ require "s3_helper"
 require "rails_helper"
 
 describe "S3Helper" do
+  before(:each) do
+    SiteSetting.enable_s3_uploads = true
+    SiteSetting.s3_access_key_id = "abc"
+    SiteSetting.s3_secret_access_key = "def"
 
-  it "can correctly set the purge policy" do
-
-    stub_request(:get, "http://169.254.169.254/latest/meta-data/iam/security-credentials/").
-      to_return(status: 404, body: "", headers: {})
-
-    lifecycle = <<~XML
+    @lifecycle = <<~XML
       <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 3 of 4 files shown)


config/locales/server.en.yml
```diff
@@ -1347,6 +1347,7 @@ en:
     s3_backup_bucket: "The remote bucket to hold backups. WARNING: Make sure it is a private bucket."
     s3_endpoint: "The endpoint can be modified to backup to an S3 compatible service like DigitalOcean Spaces or Minio. WARNING: Use default if using AWS S3"
     s3_force_path_style: "Enforce path-style addressing for your custom endpoint. IMPORTANT: Required for using Minio uploads and backups."
+    s3_configure_tombstone_policy: "Enable automatic policy configuration for Tombstone."
```
SamSaffron: help text can be a bit better "Enable automatic deletion policy configuration for tombstone uploads, without this policy no space will be reclaimed after uploads are deleted" ... something like that

lib/s3_helper.rb
```diff
@@ -131,7 +131,7 @@ def update_lifecycle(id, days, prefix: nil, tag: nil)
   end
 
   def update_tombstone_lifecycle(grace_period)
-    return if @tombstone_prefix.blank?
+    return if not SiteSetting.s3_configure_tombstone_policy or @tombstone_prefix.blank?
```
SamSaffron: `if not` is and `or` are not something we usually use

I would probably do

return if !SiteSetting.s3_configure_tombstone_policy
return if @tombstone_prefix.blank? 

or 


return if !SiteSetting.s3_configure_tombstone_policy || @tombstone_prefix.blank?

rishabhnambiar: my bad, that is much better, will be more careful after this :+1:

rishabhnambiar:
Yup, looking at `s3_store_spec.rb` right now to see what exactly I can add, will be done today.

SamSaffron:
this does test we are setting policy ... so yay ... but we should also simply call `update_tombstone_lifecycle` in a test with `SiteSetting.s3_configure_tombstone_policy = false` and make sure no exceptions are raised. We would raise without cause we would make an HTTP call which would be caught by our test suite. 

@andrewschleifer I am merging this in so you can test the new env var on BBS and so on....

spec/components/s3_helper_spec.rb
```diff
@@ -40,17 +46,28 @@
...
+    SiteSetting.s3_configure_tombstone_policy = false
+
+    stub_request(:get, "https://bob.s3.amazonaws.com/?lifecycle").
+      to_return(status: 200, body: @lifecycle, headers: {})
+
+    hash = Hash.from_xml(@lifecycle)
+    rules = hash["LifecycleConfiguration"]["Rule"]
 
     helper = S3Helper.new('bob', 'tomb')
     helper.update_tombstone_lifecycle(100)
```
rishabhnambiar: I've called `update_tombstone_lifecycle()` when `s3_configure_tombstone_policy = false`.
I believe this is what you wanted @SamSaffron

spec/components/s3_helper_spec.rb
```diff
@@ -40,17 +46,28 @@
...
-    SiteSetting.s3_access_key_id = "abc"
-    SiteSetting.s3_secret_access_key = "def"
+    helper = S3Helper.new('bob', 'tomb')
+    helper.update_tombstone_lifecycle(100)
+  end
+
+  it "can skip policy update when s3_configure_tombstone_policy is disabled" do
+    SiteSetting.s3_configure_tombstone_policy = false
+
+    stub_request(:get, "https://bob.s3.amazonaws.com/?lifecycle").
```
SamSaffron: no need to stub anything here cause nothing should be called.

rishabhnambiar: Done, missed that in a hurry! :+1:

spec/components/s3_helper_spec.rb
```diff
@@ -40,17 +46,25 @@
...
-    SiteSetting.s3_access_key_id = "abc"
-    SiteSetting.s3_secret_access_key = "def"
+    helper = S3Helper.new('bob', 'tomb')
+    helper.update_tombstone_lifecycle(100)
+  end
+
+  it "can skip policy update when s3_configure_tombstone_policy is disabled" do
+    SiteSetting.s3_configure_tombstone_policy = false
+
+    hash = Hash.from_xml(@lifecycle)
```
SamSaffron: I am pretty sure you want to remove this as well and the assertion at the end. Try this without the site setting, it would explode in fire.

SamSaffron:
This looks good to me, merging in @andrewschleifer  can you ensure this is added to the S3 sites we host where you are manually configuring tombstone?

PR #6370, opened 2018-09-06
extract inline JS that's used to store preloaded data

xrav3nz:
I'm extracting inline JS in `application.html.erb` (and its includes) bits by bits to prepare for CSP. 

I'd love a fresh pair of 👀  on the approach; once this initial PR is OK'd, the rest of the inline JS changes should fall in pretty easily.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -55,6 +55,9 @@
     <%= render partial: "common/discourse_stylesheet" %>
     <%= render partial: "common/special_font_face" %>
 
+    <meta id="data-preloaded" data-preloaded="<%= preloaded_json %>">
+    <%= preload_script "preload-application-data" %>
+
     <%= yield :head %>
 
     <%= build_plugin_html 'server:before-head-close' %>
@@ -103,17 +106,6 @@
         <input type="submit" id="signin-button" value="<%= t 'log_in' %>">
       </form>
     <% end %>
-
-    <%- if @preloaded.present? %>
-      <script>
-      (function() {
-        var ps = require('preload-store').default;
-        <%- @preloaded.each do |key, json| %>
-          ps.store("<%= key %>", <%= escape_unicode(json) %>);
-        <% end %>
-      })();
-      </script>
-    <%- end %>
 
     <%= yield :data %>
 
--- a/config/application.rb
+++ b/config/application.rb
@@ -119,6 +119,7 @@ module Discourse
       service-worker.js
       google-tag-manager.js
       google-universal-analytics.js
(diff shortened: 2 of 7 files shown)


config/application.rb
```diff
@@ -116,6 +116,7 @@ def config.database_configuration
       plugin-third-party.js
       markdown-it-bundle.js
       service-worker.js
+      application-footer.js
```
xrav3nz: This is a terrible name (I tried 😢). This is supposed to be the home of all inline JS currently in `application.html.erb`. Suggestions more than welcome.

app/assets/javascripts/application-footer.js
```diff
@@ -0,0 +1,12 @@
+(function() {
+  const preloadedDataElement = document.getElementById("data-preloaded");
+
+  if (preloadedDataElement) {
```
xrav3nz: Check if the element exists, so this works regardless of whether `@preloaded.present?` or not.

app/views/layouts/application.html.erb
```diff
@@ -105,15 +105,9 @@
...
     <%- if @preloaded.present? %>
-      <script>
-      (function() {
-        var ps = require('preload-store').default;
-        <%- @preloaded.each do |key, json| %>
-          ps.store("<%= key %>", <%= escape_unicode(json) %>);
-        <% end %>
-      })();
-      </script>
+      <div id="data-preloaded" data-preloaded="<%= @preloaded.transform_values(&method(:escape_unicode)).to_json %>"></div>
```
SamSaffron: @preloaded.transform_values(&method(:escape_unicode)).to_json is a lot to unwrap in your head 

I would go with a helper here called: `preloaded_json` and make sure there is a test case for it that demos how it works, personally I kind of prefer:

transform_values {|v| escape_unicode(v)} , I am pretty sure it would bench faster as well in a micro bench but the more important thing is that it is a clearer pattern

config/application.rb
SamSaffron: preloaded-application-data.js ?

xrav3nz: Similar ones came across my mind too, only concern is that the file will contain some [non-data-related scripts](https://github.com/discourse/discourse/blob/879067d0009ad4db4669af59215a4921e75afbc1/app/views/common/_discourse_javascript.html.erb#L1-L35) too. I will rename to this for the time being tho.

tgxworld:
This looks OK to me too. @xrav3nz Feel free to merge once we get started on `2.2`.

SamSaffron:
@xrav3nz looks like we have a conflict here

PR #6371, opened 2018-09-07
Updated IPB import script

cebe:
- improve handling of code tags (current version made these unreadable)
- add a converion for lists, these were not converted to markdown before
- improved handling of uploaded attachments:
  - show images directly and add a link to them
  - for non-images just provide a download link


cebe:
No idea what's wrong with travis-ci build now. Does not look related to my changes.

SamSaffron:
change looks safe to me! thanks

hameedacpa:
Are there any update for this script because it will not work with IPB 4x and also it is not compatible with Discourse itself as the structure of tables changed and the script remain as it is

PR #6373, opened 2018-09-07
refactors trending search report to use SearchLog

jjaffeux:


jjaffeux:
This is meant to be merged after release.

app/models/search_log.rb
```diff
@@ -103,19 +103,38 @@ def self.term_details(term, period = :weekly, search_type = :all)
   end
 
   def self.trending(period = :all, search_type = :all)
-    result = SearchLog.select("term,
-       COUNT(*) AS searches,
-       SUM(CASE
+    SearchLog.trending_from(start_of(period), search_type: search_type)
+  end
+
+  def self.trending_from(start_date, options = {})
```
tgxworld: I think this new method needs to be tested.

jjaffeux: I'm mixed on this, as it's already tested by this existing one and is adding more code to add very little confidence. But ok fine, won’t fight for this.

tgxworld: My worry is that the existing tests does not cover all the other options that the method accepts.

PR #6374, opened 2018-09-07
FIX: async reload of locales could result in missing translations

gschlager:
Another try at fixing https://meta.discourse.org/t/translation-missing-in-english-locale/84019. This supersedes https://github.com/discourse/discourse/pull/5208.

Reloading of locales via MessageBus could result in missing translations when `I18n.reload!` is called by `MessageBus` while `I18n.t()` is used.

image

### DIFF
--- a/lib/freedom_patches/translate_accelerator.rb
+++ b/lib/freedom_patches/translate_accelerator.rb
@@ -15,23 +15,21 @@ module I18n
     alias_method :translate_no_cache, :translate
     alias_method :exists_no_cache?, :exists?
     alias_method :reload_no_cache!, :reload!
+    alias_method :locale_no_cache=, :locale=
+
     LRU_CACHE_SIZE = 300
 
     def init_accelerator!
       @overrides_enabled = true
-      reload!
+      execute_reload
     end
 
     def reload!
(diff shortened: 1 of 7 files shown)


SamSaffron:
I prefer to fix #reload! here cause this change is implying we now have two method: 

`reload_functioning` and `reload_broken` and are keeping both around for reasons.

gschlager:
We need `reload_broken` (= "reload right now") only for tests. I could put everything into one `reload!` method and change the behavior depending on Rails env. But before I do that I'll take another look at the specs and see if there's a better way to make them work with a deferred reload of locales...

gschlager:
Calling `I18n.reload!` will now always schedule a reload for a later time instead of doing it right away. The actual reload will happen on calls to `I18n.t`, `I18n.exists?`, `I18n.search` and `I18n.locale=`.

I changed the specs (and shuffled some of them around) so that they will work with this new behavior. The translations used by the affected specs are now stored in `spec/fixtures/i18n/*.yml` files. That's actually less error-prone and mirrors how it works in production. Storing translations directly into the I18n backend doesn't survive a reload and some specs didn't even do what they were supposed to do (anymore).

PR #6375, opened 2018-09-08
FIX: Retry sending email in case of temporary issue.

nbianca:
Fixes [this](https://meta.discourse.org/t/temporary-name-resolution-problem-prevents-emails-from-sending-and-they-arent-queued/70942).

### DIFF
--- a/app/jobs/regular/notify_mailing_list_subscribers.rb
+++ b/app/jobs/regular/notify_mailing_list_subscribers.rb
@@ -5,7 +5,20 @@ module Jobs
   class NotifyMailingListSubscribers < Jobs::Base
     include Skippable
 
+    RETRY_TIMES = [5.minute, 15.minute, 30.minute, 45.minute, 90.minute, 180.minute, 300.minute]
+
     sidekiq_options queue: 'low'
+
+    sidekiq_options retry: RETRY_TIMES.size
+
+    sidekiq_retry_in do |count, exception|
+      case exception.wrapped
+      when SocketError
+        RETRY_TIMES[count]
+      else
+        Jobs::UserEmail.seconds_to_delay(count)
+      end
+    end
 
     def execute(args)
       return if SiteSetting.disable_mailing_list_mode


app/jobs/regular/notify_mailing_list_subscribers.rb
```diff
@@ -7,6 +7,17 @@ class NotifyMailingListSubscribers < Jobs::Base
 
     sidekiq_options queue: 'low'
 
+    sidekiq_options retry: 3
+
+    sidekiq_retry_in do |count, exception|
+      case exception.wrapped
+      when SocketError
```
ZogStriP: Does a `SocketError` always mean it's a temporary issue?

nbianca: In theory, yes. SocketError is quite generic, but I believe most of the times, the socket errors are actually temporary.

I could hard code the exception message from the bug report, but I believe this solution is better.

I also do not think there is no guarantee that even the problem mentioned on Meta is temporary. That could be due to a misconfiguration that will persist.

ZogStriP:
Also, we need to test that job is actually retried when there's this Exception ;)

SamSaffron:
I support this change, but we need a spec here.

app/jobs/regular/notify_mailing_list_subscribers.rb
```diff
@@ -5,8 +5,21 @@ module Jobs
   class NotifyMailingListSubscribers < Jobs::Base
     include Skippable
 
+    RETRY_TIMES = [5.minute, 15.minute, 30.minute, 45.minute, 90.minute, 180.minute, 300.minute]
```
tgxworld: I think we should use the same pattern used [here](https://github.com/discourse/discourse/blob/master/app/jobs/regular/user_email.rb#L166).

nbianca: The pattern there was simple. I do no think it would be the same in this case.

I could use something like `(count + 1) * 15.minutes`? It would have a different meaning, but the first few values match pretty good.

tgxworld: Hmm after reading the topic, why do we need to retry 7 times at those particular intervals? If the job raised an exception, [Sidekiq will add it to the fail queue and retry](https://github.com/mperham/sidekiq/wiki/Error-Handling).

ZogStriP: Since it's a temporary issue, we're just bypassing the "raise an exception" part ;) No need to fill the logs if we can just retry in a bit.

nbianca:
@SamSaffron Writing a spec for this one is not that easy. First of all, I believe that writing a spec for this one would actually test Sidekiq's functionality of retrying a job.

Anyway, it looks like other people have tried doing the same and [did not succeed](https://stackoverflow.com/questions/32874510/how-can-i-test-sidekiq-job-failures).

ZogStriP:
Ok, let's try this.

PR #6376, opened 2018-09-09
FEATURE: add indication if incoming email attachment was rejected and inform sender about it

majakomel:
https://meta.discourse.org/t/e-mail-reject-attachment-template-or-add-note-to-topic-indicating-attachment-rejected-reason-for-rejection/47781/

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -88,6 +88,7 @@ en:
       maximum_staged_user_per_email_reached: "Reached maximum number of staged users created per email."
       no_subject: "(no subject)"
       no_body: "(no body)"
+      missing_attachment: '(Attachment %{filename} is missing)'
       errors:
         empty_email_error: "Happens when the raw mail we received was blank."
         no_message_id_error: "Happens when the mail has no 'Message-Id' header."
@@ -2604,6 +2605,16 @@ en:
         We're sorry, but your email message to %{destination} (titled %{former_title}) didn't work.
 
         There was an unrecognized error while processing your email and it wasn't posted. You should try again, or [contact a staff member](%{base_url}/about).
+
+    email_reject_attachment:
+      title: "Email Attachment Rejected"
+      subject_template: "[%{email_prefix}] Email issue -- Attachment Rejected"
+      text_body_template:  |
+        We're sorry, but the following attachments in your email message to %{destination} (titled %{former_title}) were rejected: %{rejected_attachments}
+
+        This can be caused by the attached file being too large or by an unauthorised file extension.
+
+        If you believe this is an error, [contact a staff member](%{base_url}/about).
 
     email_error_notification:
       title: "Email Error Notification"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -2610,9 +2610,10 @@ en:
       title: "Email Attachment Rejected"
       subject_template: "[%{email_prefix}] Email issue -- Attachment Rejected"
       text_body_template:  |
-        We're sorry, but the following attachments in your email message to %{destination} (titled %{former_title}) were rejected: %{rejected_attachments}
+        Unfortunately some attachments in your email message to %{destination} (titled %{former_title}) were rejected.
 
-        This can be caused by the attached file being too large or by an unauthorised file extension.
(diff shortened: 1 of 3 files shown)


lib/email/receiver.rb
```diff
@@ -900,15 +902,33 @@ def add_attachments(raw, user_id, options = {})
...
+          else
+            is_attachment_rejected = true
+            rejected_attachments << upload.original_filename
+            raw << "\n\n#{I18n.t('emails.incoming.missing_attachment', filename: upload.original_filename)}\n\n"
           end
         ensure
           tmp&.close!
         end
       end
+      notify_about_rejected_attachment(rejected_attachments) if is_attachment_rejected
```
ZogStriP: You don't need the `is_attachment_rejected` since you can test for `rejected_attachments.present?` instead ;)

lib/email/receiver.rb
```diff
@@ -900,15 +902,33 @@ def add_attachments(raw, user_id, options = {})
...
 
       raw
     end
 
+    def notify_about_rejected_attachment(attachments)
+      template_args = {}
+
+      message = Mail::Message.new(@mail)
+      template_args[:former_title] = message.subject
```
ZogStriP: When we don't conditionally add keys to the hash, we usually set all the keys at the same time we declare it.

Like

```ruby
template_args = {
  former_title: message.subject,
  destination: message.to,
  ...
}
```

config/locales/server.en.yml
```diff
@@ -2604,6 +2605,16 @@ en:
 
         There was an unrecognized error while processing your email and it wasn't posted. You should try again, or [contact a staff member](%{base_url}/about).
 
+    email_reject_attachment:
+      title: "Email Attachment Rejected"
+      subject_template: "[%{email_prefix}] Email issue -- Attachment Rejected"
+      text_body_template:  |
+        We're sorry, but the following attachments in your email message to %{destination} (titled %{former_title}) were rejected: %{rejected_attachments}
+
+        This can be caused by the attached file being too large or by an unauthorised file extension.
```
ZogStriP: We should at the very least list the authorized extensions. What would be even better would be to tell the reason why an attachment was rejected.

SamSaffron: there is also a very real  non edge case we should handle here @majakomel which is the default case.

Out of the box we allow no attachments at all except for images. 

I agree with @ZogStriP we should have a bit of a custom message cause this is pretty confusing.

"You may only attach images (png/gif/jpeg) to your post"

"We only allow ".txt, .text" file attachments in emails." 

The very vague message here is a bit confusing.

coding-horror:
Yes this 100%, good clear error messages are absolutely essential, I would
prioritize this.

majakomel:
I have included error messages in the email which shows why exactly the attachment was rejected:

screen shot 2018-09-17 at 10 29 44

lib/email/receiver.rb
```diff
@@ -900,15 +901,42 @@ def add_attachments(raw, user_id, options = {})
...
     end
 
+    def notify_about_rejected_attachment(attachments)
+      errors = []
+
+      attachments.each do |a|
+        error = ''
+        a.errors.messages.each do |attr, msg|
+          error = msg.first
+        end
```
gschlager: This looks weird. Why do a loop over all `messages` when it uses only one error message?

SamSaffron:
One final change here is that I would like a bypass so we do not send this to staged users, I think it can be very confusing in a support scenario.

lib/email/receiver.rb
```diff
@@ -900,15 +901,39 @@ def add_attachments(raw, user_id, options = {})
...
             end
+          else
+            rejected_attachments << upload
+            raw << "\n\n#{I18n.t('emails.incoming.missing_attachment', filename: upload.original_filename)}\n\n"
           end
         ensure
           tmp&.close!
         end
       end
+      notify_about_rejected_attachment(rejected_attachments) if rejected_attachments.present? && !User.find(user_id).staged?
```
ZogStriP: Why not pass the `user` object instead of its `id` and use that in this method? This will prevent from doing a query to the database ;)

PR #6377, opened 2018-09-10
Drop `tif`, `tiff`, `webp` and `bmp` from supported images.

tgxworld:


spec/lib/upload_creator_spec.rb
```diff
@@ -43,6 +43,27 @@
...
+
+      describe 'for webp format' do
+        before do
+          SiteSetting.authorized_extensions = 'webp|png'
+        end
+
+        let(:filename) { "webp_as.png" }
+        let(:file) { file_from_fixtures(filename) }
+
+        it 'should not correct the coerce filename' do
```
tgxworld: @SamSaffron Is that what you meant?

SamSaffron: actually the particular case I mean we should test... take a legit `webp` file, rename it to file.bin. Allow `bin` uploads, upload the `bin` ... it should still be a `.bin`.

This is not the case for png/jpeg cause we would correct the filename.

tgxworld:
Build failure will have to be fixed by the plugin

spec/lib/upload_creator_spec.rb
```diff
@@ -43,6 +43,27 @@
...
+
+      describe 'for webp format' do
+        before do
+          SiteSetting.authorized_extensions = '.webp|.bin'
+        end
+
+        let(:filename) { "webp_as.bin" }
+        let(:file) { file_from_fixtures(filename) }
+
+        it 'should not correct the coerce filename' do
```
SamSaffron: should not correct the coerce filename => should not coerce filename

:)

SamSaffron: 29k is giant for a webp maybe we can make a smaller one for tests we should be able to swing a webp in 1000 bytes

SamSaffron:
this is good to be merged @tgxworld  I would prefer a smaller webp file first though

spec/lib/upload_creator_spec.rb
tgxworld: found a webp file that is 26 bytes.

PR #6378, opened 2018-09-10
extract Google Analytics / Tag Manager inline JS

xrav3nz:
Previously I shoved all the server-rendered data into a `div`:

```html
<div class="hidden" id="id" data-server-rendered="hello world"></div>
```

This works but @SamSaffron brought up a question about HTML semantics. After looking through the docs, I decided to try out the `meta` tag In this PR:

> Metadata contains information about the page. This includes information about styles, scripts and **data to help software use and render the page**.  -- [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#Document_metadata)

```html
<meta id="id" data-server-rendered="hello world" />
```

The only limitation with `meta` is it has to live in `head`, but it's fine 99% of the time (we can always fallback to `div`s for the other 1%).

If we are good with using the `meta` tag, I will amend my previous/future PRs.

### DIFF
--- a/app/assets/javascripts/google-tag-manager.js
+++ b/app/assets/javascripts/google-tag-manager.js
@@ -1,0 +1,17 @@
+(function() {
+  const gtmDataElement = document.getElementById("data-google-tag-manager");
+  const dataLayerJson = JSON.parse(gtmDataElement.dataset.dataLayer);
+
+  // dataLayer declaration needs to precede the container snippet
+  // https://developers.google.com/tag-manager/devguide#adding-data-layer-variables-to-a-page
+  window.dataLayer = [dataLayerJson];
+
+  /* eslint-disable */
+  // prettier-ignore
+  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+  'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+  })(window,document,'script','dataLayer',gtmDataElement.dataset.containerId);
+  /* eslint-enable */
+})();
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,7 +31,7 @@ module ApplicationHelper
     if SiteSetting.ga_universal_auto_link_domains.present?
       result[:allowLinker] = true
     end
-    result.to_json.html_safe
+    result.to_json
   end
 
   def ga_universal_json
--- a/app/assets/javascripts/google-universal-analytics.js
+++ b/app/assets/javascripts/google-universal-analytics.js
@@ -1,0 +1,20 @@
+/* eslint-disable */
+// prettier-ignore
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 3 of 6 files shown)


SamSaffron:
are we technically permitted to have an async attribute on these script tags?

SamSaffron:
hmmm not point rendering anything here if it is not enabled: 

https://github.com/discourse/discourse/blob/103f9b5dc75ae4b2cb0ac985a871489d332db4b0/app/views/layouts/application.html.erb#L47-L48

SamSaffron:
sorry error closing PR ... clicked wrong button

SamSaffron:
I like the meta tags, I think it works nicely... thoughts @tgxworld  ?

xrav3nz:
> are we technically permitted to have an async attribute on these script tags?

Hmm good question. A quick search didn't yield anything useful; will look into this. 

> no point rendering anything here if it is not enabled:

The guards are in the helper methods:

https://github.com/discourse/discourse/blob/85620abb710741ed83c42565b1033c492a756aeb/app/helpers/common_helper.rb#L2-L6

I could refactor this to extract the conditionals as helpers, if that's clearer i.e.

```erb
<%- if render_google_universal_analytics_code? %>
  <%= render partial: "common/google_universal_analytics"  %>
<% end %>
```

app/assets/javascripts/google-tag-manager.js
```diff
@@ -0,0 +1,15 @@
...
+  /* eslint-disable */
+  // prettier-ignore
+  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+  'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+  })(window,document,'script','dataLayer',gtmDataElement.dataset.containerId);
+  /* eslint-enable */
+
+  window.dataLayer = [dataLayerJson];
```
tgxworld: Do we have to set the `dataLayer` attribute before loading the Google Tag Manager script as per https://developers.google.com/tag-manager/devguide#adding-data-layer-variables-to-a-page? 

screenshot from 2018-09-12 10-08-41

app/helpers/application_helper.rb
```diff
@@ -31,7 +31,7 @@ def google_universal_analytics_json(ua_domain_name = nil)
     if SiteSetting.ga_universal_auto_link_domains.present?
       result[:allowLinker] = true
     end
-    result.to_json.html_safe
```
tgxworld: hmm why was this removed?

xrav3nz: We are now rendering this json in a HTML attribute, so it has to be escaped (no longer html safe)

```html
<!-- html_safe -->
<meta data-json="{"key": "value"}">

<!-- without html_safe -->
<meta data-json="{&quot;key&quot;: &quot;value&quot;}">
```

app/assets/javascripts/google-tag-manager.js
xrav3nz: Ah good catch, thanks!

xrav3nz:
re async/preload on the script tags, I think we safe here since we are not changing how the analytic scripts are loading.

Also did a [live test](https://xrav3nz.github.io/ga-gtm-preload-test/) today:

image

app/helpers/application_helper.rb
tgxworld: :+1:

SamSaffron:
Cool, merging this in, I want to see some of this in action 👍

PR #6381, opened 2018-09-10
FIX: Remove null bytes from search terms to prevent ArgumentError(s)

rishabhnambiar:
**Error log**
```ruby
ArgumentError (string contains null byte) /var/www/discourse/lib/search.rb:148:in `escape_string' 
/var/www/discourse/lib/search.rb:148:in `escape_string'
/var/www/discourse/lib/search.rb:148:in `initialize'
/var/www/discourse/app/controllers/search_controller.rb:33:in `new'
/var/www/discourse/app/controllers/search_controller.rb:33:in `show'
```

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -138,6 +138,8 @@ class Search
     term.gsub!(/[\u200B-\u200D\uFEFF]/, '')
     # Replace curly quotes to regular quotes
     term.gsub!(/[\u201c\u201d]/, '"')
+    # Removes any null bytes from search terms
+    term.gsub!("\u0000", '')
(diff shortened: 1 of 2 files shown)


rishabhnambiar:
@SamSaffron I was doing this error for the log game and I just realized you fixed it here :man_facepalming: 
https://github.com/discourse/discourse/commit/2f5c21e28c4f8b8a441b611caefe206a178b6eef

So there would be no need for this PR anymore right?

SamSaffron:
Yeah no need for this change, but good first shot :)

PR #6384, opened 2018-09-11
FIX: use current user color scheme when filling `theme-color` attribute

OsamaSayegh:
https://meta.discourse.org/t/meta-attribute-theme-color-uses-the-default-header-background-color-not-the-active-one/96727?u=osama


SamSaffron:
sure, lets try this, thanks

PR #6387, opened 2018-09-12
FIX: allows forcing timezone displayed in local-dates

jjaffeux:
Use case being you might want to always display a specific timezone, but still have the possibility to show the tooltip with various timezones.


tgxworld:
Merged in https://github.com/discourse/discourse/commit/4b3c2490c38ed6acfe37597a21d5a36e26d54a97

Thanks @jjaffeux!

PR #6388, opened 2018-09-12
Add test to ensure a user cannot see drafts stream of another user

pmusaraj:
As requested by @tgxworld here: https://github.com/discourse/discourse/commit/4a872823e75fe54f752eca12b14a35108b2e871c#r29930758

### DIFF
--- a/app/controllers/drafts_controller.rb
+++ b/app/controllers/drafts_controller.rb
@@ -16,8 +16,6 @@ class DraftsController < ApplicationController
       limit: params[:limit]
     }
 
-    help_key = "user_activity.no_drafts"
-
     if user == current_user
       stream = Draft.stream(opts)
       stream.each do |d|
@@ -31,15 +29,13 @@ class DraftsController < ApplicationController
           end
         end
       end
-
-      help_key += ".self"
     else
-      help_key += ".others"
+      raise Discourse::InvalidAccess
     end
 
     render json: {
       drafts: stream ? serialize_data(stream, DraftSerializer) : [],
-      no_results_help: I18n.t(help_key)
+      no_results_help: I18n.t("user_activity.no_drafts.self")
     }
 
   end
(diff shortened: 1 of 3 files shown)


spec/requests/drafts_controller_spec.rb
```diff
@@ -24,4 +24,15 @@
     parsed = JSON.parse(response.body)
     expect(parsed["drafts"].length).to eq(0)
   end
+
+  it 'does not let userA see drafts by userB' do
+    userB = Fabricate(:user)
+    Draft.set(userB, 'xxx', 0, '{}')
+
+    userA = sign_in(Fabricate(:user))
```
tgxworld: We don't need this variable here :) On the ruby side, we also prefer to use snake_case instead of camelCase.

spec/requests/drafts_controller_spec.rb
```diff
@@ -24,4 +24,15 @@
...
     expect(parsed["drafts"].length).to eq(0)
   end
+
+  it 'does not let a user see drafts stream of another user' do
+    user_b = Fabricate(:user)
+    Draft.set(user_b, 'xxx', 0, '{}')
+
+    sign_in(Fabricate(:user))
+    get "/drafts.json", params: { username: user_b.username }
+    expect(response.status).to eq(200)
```
tgxworld: I'm thinking if we should return a `200` response code here since the `username` param is invalid if not signed in as the current user.

pmusaraj: Would `raise Discourse::InvalidAccess` be appropriate here? It would also mean that in the frontend, users attempting `/u/user_b/activity/drafts` would get an Access Denied screen. 

Seems reasonable to me @tgxworld.

app/controllers/drafts_controller.rb
```diff
@@ -31,15 +29,13 @@ def index
           end
         end
       end
-
-      help_key += ".self"
     else
-      help_key += ".others"
+      raise Discourse::InvalidAccess
```
tgxworld: :bear: with me during this review but I'll prefer that we raise the error as early as possible. IMO, the check and error can be raised after `user = fetch_user_from_params`

tgxworld:
Thank for you making the changes @pmusaraj :+1:

PR #6390, opened 2018-09-12
FIX: Permit unpinning an unlisted topic.

nbianca:


ZogStriP:
Can you also write a test?

SamSaffron:
looks good to me, thanks!

PR #6393, opened 2018-09-13
FIX: ignore and log bad json values for custom fields

techAPJ:


spec/components/concern/has_custom_fields_spec.rb
```diff
@@ -187,6 +187,55 @@ class CustomFieldsTestItemCustomField < ActiveRecord::Base
...
+      field_type = "json_array"
+      CustomFieldsTestItem.register_custom_field_type(field_type, :json)
+
+      item = CustomFieldsTestItem.new
+      item.custom_fields = {
+        "json_array" => [{ a: "test" }, { b: "another" }]
+      }
+      item.save
+
+      item = CustomFieldsTestItem.find(item.id)
```
tgxworld: You can use `item.reload` instead which does the same thing but with less code.

spec/components/concern/has_custom_fields_spec.rb
```diff
@@ -187,6 +187,55 @@ class CustomFieldsTestItemCustomField < ActiveRecord::Base
...
+
+      expect(item.custom_fields[field_type]).to eq(
+          [{ "a" => "test" }, { "b" => "another" }]
```
tgxworld: style: The indentation here is off

spec/components/concern/has_custom_fields_spec.rb
```diff
@@ -187,6 +187,55 @@ class CustomFieldsTestItemCustomField < ActiveRecord::Base
...
+      )
+
+      item.custom_fields["json_array"] = ['a']
+      item.save
+
+      item = CustomFieldsTestItem.find(item.id)
+
+      expect(item.custom_fields[field_type]).to eq(["a"])
+
+      item.custom_fields["json_array"] = ['a', 'b']
```
tgxworld: I'm wondering if we need this test case here since the number of items in the array isn't testing anything new.

spec/components/concern/has_custom_fields_spec.rb
```diff
@@ -187,6 +187,55 @@ class CustomFieldsTestItemCustomField < ActiveRecord::Base
...
+      expect(item.custom_fields[field_type]).to eq(["a", "b"])
+    end
+
+    it "will not fail to load custom fields if json is corrupt" do
+
+      field_type = "bad_json"
+      CustomFieldsTestItem.register_custom_field_type(field_type, :json)
+
+      item = CustomFieldsTestItem.new
+      item.save
```
tgxworld: The two lines above can be replaced with `item = CustomFieldsTestItem.create!`

spec/components/concern/has_custom_fields_spec.rb
techAPJ: I've removed the single item array test case, and kept this instead.

PR #6394, opened 2018-09-13
FIX: Onceoff job to recover missing post uploads.

tgxworld:
This fixes the regression due to https://github.com/discourse/discourse/commit/1f636c445bcbc66b1765bcc3966887adcb14bdbf


tgxworld:
@SamSaffron  @ZogStriP  Can you please review this commit? It'll be great if you can review the `UploadRecovery` class as well. Thank you!

app/jobs/onceoff/recover_post_uploads.rb
```diff
@@ -0,0 +1,10 @@
+require_dependency "upload_recovery"
+
+module Jobs
+  class RecoverPostUploads < Jobs::Onceoff
+    def execute_onceoff(args)
+      grace_period = SiteSetting.purge_deleted_uploads_grace_period_days.days.ago
```
SamSaffron: give this a minimum and maximum I think ... minimum is say 30 days and maximum is 120. I do not wan this to do absolutely everything ...

app/jobs/onceoff/recover_post_uploads.rb
```diff
@@ -0,0 +1,10 @@
...
+      UploadRecovery.new.recover(Post.where("baked_at >= ?", grace_period))
```
SamSaffron: hmm so you pass in a relation... I hope the job is not materializing all those posts cause memory will explode, what happens if a single post explodes doe the whole onceoff fail?

tgxworld: Yup if a single post explodes, the onceoff fail and we can figure out why. I'm doing a `find_each` so no worries about it materializing everything.

PR #6395, opened 2018-09-13
DEV: apply plugin migrations when testing in Travis and Docker

davidtaylorhq:
In `RAILS_ENV=test`, plugins are not loaded by default. Therefore we need to explicitly specify `LOAD_PLUGINS=1` when we want to apply plugin migrations.

This will become essential once @ZogStriP's poll changes are merged (note that travis is currently failing on #6359). It will also be useful for `discourse-anonymous-user` and `discourse-brightcove` (and any other plugins that may want to apply migrations in future).

This PR makes the changes for Travis, and for `docker.rake` (used by our internal test system, and plugins that use Travis)


tgxworld:
Looks good to me.

PR #6396, opened 2018-09-13
FIX: ensures we have a color for reports

jjaffeux:


app/models/report.rb
```diff
@@ -29,8 +29,10 @@ def initialize(type)
     @modes = [:table, :chart]
     @prev_data = nil
     @dates_filtering = true
-    @primary_color = rgba_color(ColorScheme.hex_for_name('tertiary'))
-    @secondary_color = rgba_color(ColorScheme.hex_for_name('tertiary'), 0.1)
+
+    tertiary = ColorScheme.hex_for_name('tertiary') || '0088cc'
```
ZogStriP: Where is `0088cc` coming from? Shouldn't it be from the current color scheme?

jjaffeux: 0088cc is our default and yes I'm trying to get it from current scheme `ColorScheme.hex_for_name('tertiary')` but it seems like sometimes this is returning nil, I can’t reproduce, but I'm quite sure this is what is happening https://meta.discourse.org/t/after-update-to-2-1-0-admin-dashboard-have-errors-and-not-load/97061

PR #6397, opened 2018-09-13
FEATURE: new 'ignore_accents' site setting

ZogStriP:
This adds a new `ignore accents` site setting (defaults to `false`, but enabled for specific locales) to automatically remove diacritics before indexing for search.

https://meta.discourse.org/t/discourse-should-ignore-if-a-character-is-accented-when-doing-a-search/90198

### DIFF
--- a/app/services/search_indexer.rb
+++ b/app/services/search_indexer.rb
@@ -11,8 +11,8 @@ class SearchIndexer
     @disabled = false
   end
 
-  def self.scrub_html_for_search(html)
-    HtmlScrubber.scrub(html)
+  def self.scrub_html_for_search(html, strip_diacritics: SiteSetting.ignore_accents)
+    HtmlScrubber.scrub(html, strip_diacritics: strip_diacritics)
   end
 
   def self.inject_extra_terms(raw)
@@ -169,18 +169,10 @@ class SearchIndexer
 
     DIACRITICS ||= /([\u0300-\u036f]|[\u1AB0-\u1AFF]|[\u1DC0-\u1DFF]|[\u20D0-\u20FF])/
 
-    def self.strip_diacritics(str)
-      s = str.unicode_normalize(:nfkd)
-      s.gsub!(DIACRITICS, "")
-      s.strip!
-      s
-    end
-
     attr_reader :scrubbed
 
     def initialize(strip_diacritics: false)
       @scrubbed = +""
-      # for now we are disabling this per: https://meta.discourse.org/t/discourse-should-ignore-if-a-character-is-accented-when-doing-a-search/90198/16?u=sam
       @strip_diacritics = strip_diacritics
     end
 
@@ -189,7 +181,7 @@ class SearchIndexer
 
       me = new(strip_diacritics: strip_diacritics)
       Nokogiri::HTML::SAX::Parser.new(me).parse("<div>#{html}</div>")
-      me.scrubbed
+      me.scrubbed.squish
     end
 
     ATTRIBUTES ||= %w{alt title href data-youtube-title}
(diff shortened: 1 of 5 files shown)


config/site_settings.yml
```diff
@@ -1445,6 +1444,19 @@ search:
...
     max: 1825 # 5 years
+  ignore_accents:
+    default: false
+    locale_default:
+      ar: true
+      ca: true
+      cs: true
+      el: true
+      es: true
+      fa: true
```
gschlager: Specs are failing. It should be `fa_IR`

config/site_settings.yml
```diff
@@ -1445,6 +1444,19 @@ search:
   search_query_log_max_retention_days:
     default: 365 # 1 year
     max: 1825 # 5 years
+  ignore_accents:
+    default: false
+    locale_default:
+      ar: true
+      ca: true
+      cs: true
```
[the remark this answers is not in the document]
gschlager: In Discourse that would be `hu`, `tr_TR` and `sk`.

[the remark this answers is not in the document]
danielhollas: Oh, right, sorry about that.

app/services/search_indexer.rb
```diff
@@ -189,7 +181,7 @@ def self.scrub(html, strip_diacritics: false)
 
       me = new(strip_diacritics: strip_diacritics)
       Nokogiri::HTML::SAX::Parser.new(me).parse("<div>#{html}</div>")
-      me.scrubbed
+      me.scrubbed.squish
```
SamSaffron:
```
[1] pry(main)> "test\n   \ntest".squish
=> "test test"
```
Do we really want to eat up every newline? What impact will this have on excerpts ?

SamSaffron:
Overall I like, happy that the stripping is not enabled on english by default cause diacritics are just so rare that it is kind of nice to be able to find the rare words that use them. We can revisit the list of on by default vs off by default over time for sure.

SamSaffron:
@erlend-sh what about Danish / Norwegian, should diacritics be stripped for search?

app/services/search_indexer.rb
```diff
@@ -11,8 +11,8 @@ def self.enable
     @disabled = false
   end
 
-  def self.scrub_html_for_search(html)
-    HtmlScrubber.scrub(html)
+  def self.scrub_html_for_search(html, strip_diacritics = SiteSetting.ignore_accents)
```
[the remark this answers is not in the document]
tgxworld: :+1: for keyword arguments.

app/services/search_indexer.rb
ZogStriP: Good point. I just wanted to remove consecutive white spaces (but leave new lines).

ZogStriP: Actually, search results were always squished https://github.com/discourse/discourse/blob/aca195e4a7fb760009e3e57e2a884349ed56cf08/lib/search/grouped_search_results.rb#L66

config/locales/server.en.yml
```diff
@@ -1143,6 +1143,7 @@ en:
     log_search_queries: "Log search queries performed by users"
     search_query_log_max_size: "Maximum amount of search queries to keep"
     search_query_log_max_retention_days: "Maximum amount of time to keep search queries, in days."
+    ignore_accents: "Ignore accents when searching for text."
```
SamSaffron: one last thing... let's call this `search_ignore_accents` ... far clearer name cause this is very search specific. Once done feel free to merge

ZogStriP:
Merged in https://github.com/discourse/discourse/commit/4481836de2feb4813b6282a6ec4ae4fdde509627

PR #6399, opened 2018-09-13
Add async search to tags dropdown

pmusaraj:
This also includes the following small fixes: 

- shows the tag label in the header of the dropdown when visiting `/tags/tag`, even when this tag isn't in the top x tags
- shows the "all tags" link inside the dropdown consistently

### DIFF
--- a/app/assets/javascripts/select-kit/components/tag-drop.js.es6
+++ b/app/assets/javascripts/select-kit/components/tag-drop.js.es6
@@ -1,9 +1,10 @@
 import ComboBoxComponent from "select-kit/components/combo-box";
 import DiscourseURL from "discourse/lib/url";
+import Tags from "select-kit/mixins/tags";
 import { default as computed } from "ember-addons/ember-computed-decorators";
-const { isEmpty } = Ember;
+const { isEmpty, run } = Ember;
 
-export default ComboBoxComponent.extend({
+export default ComboBoxComponent.extend(Tags, {
   pluginApiIdentifiers: ["tag-drop"],
   classNameBindings: ["categoryStyle", "tagClass"],
   classNames: "tag-drop",
@@ -21,6 +22,8 @@ export default ComboBoxComponent.extend({
   fullWidthOnMobile: true,
   caretDownIcon: "caret-right",
   caretUpIcon: "caret-down",
+  allowContentReplacement: true,
+  isAsync: true,
 
   @computed("tagId")
   noTagsSelected() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/select-kit/components/tag-drop.js.es6
```diff
@@ -1,9 +1,10 @@
 import ComboBoxComponent from "select-kit/components/combo-box";
 import DiscourseURL from "discourse/lib/url";
+import Tags from "select-kit/mixins/tags";
```
tgxworld: Let's name this `TagsMixin` instead since most of the other files follow this convention.

app/assets/javascripts/select-kit/components/tag-drop.js.es6
```diff
@@ -119,6 +124,21 @@ export default ComboBoxComponent.extend({
...
+      q: query,
+      limit: this.get("siteSettings.max_tag_search_results")
+    };
+
+    this.searchTags("/tags/filter/search", data, this._transformJson);
+  },
+
+  _transformJson(context, json) {
+    let results = json.results;
+    results = results.sort((a, b) => a.id > b.id);
```
tgxworld: I'm curious why we need to sort this and if possible, can we sort this server side instead?

tgxworld:
@pmusaraj May I have the link to the discussion for this feature? I'll like to understand a little more about this feature.

tgxworld:
@pmusaraj May I have the link to the discussion for this feature? I'll like to understand a little more about this feature. We should also have a look at writing QUnit acceptance test to ensure that we don't regress.

jjaffeux:
@pmusaraj I have one big question: could we avoid hitting network if tag is already in content ?

pmusaraj:
@jjaffeux we could avoid hitting network only if we could know client-side that the total number of tags is not greater than `max_tags_in_filter_list`. So we would need at least a `count` query to server (unless this number is already available client-side, somehow?). 

I do feel this would be overkill. The network call is only triggered when looking for a specific tag using the keyboard. In sites with few tags, in the majority of cases the user will see the tag they're looking for in the dropdown.

SamSaffron:
Agree with @tgxworld on the name change for the mixin (will apply by hand), agree with @pmusaraj that doing massive amounts of gymnastics is not really worth it to avoid the network call, especially since tag groups make it even more complex to have the knowledge that the client has the full list with all dependencies.

PR #6400, opened 2018-09-13
FIX: Sync client & server rules regarding tag names.

nbianca:


SamSaffron:
This looks like pretty complex code to me, can we add a client side test?

PR #6403, opened 2018-09-16
SPEC: reworked specs of existing group behavior

xrav3nz:

This only re-organizes and adds some specs about our **current** group behaviour around joining, leaving, title granting, and primary group changes. 

---

- [x] More specs on existing `Group` behavior (this PR)
- [x] Switch title if exact match and primary changes (https://github.com/discourse/discourse/pull/6404)
- [ ] When stripping a title, we should check if there is another title we can grant the user

### DIFF
--- a/spec/models/group_spec.rb
+++ b/spec/models/group_spec.rb
@@ -454,6 +454,29 @@ describe Group do
       expect(event[:event_name]).to eq(:group_destroyed)
       expect(event[:params].first).to eq(group)
     end
+
+    it "strips the user's title and unsets the user's primary group when exact match" do
+      group.update(title: 'Awesome')
+      user.update(primary_group: group)
+
+      group.destroy!
+
+      user.reload
+      expect(user.title).to eq(nil)
+      expect(user.primary_group).to eq(nil)
+    end
+
+    it "does not strip title or unset primary group when not exact match" do
+      primary_group = Fabricate(:group, primary_group: true, title: 'Different')
+      primary_group.add(user)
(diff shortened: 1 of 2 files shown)


spec/models/group_spec.rb
```diff
@@ -689,7 +653,60 @@ def can_view?(user, group)
...
 
+  describe '#remove' do
+    before { group.add(user) }
+
+    context 'when group has a title' do
+      before { group.update(title: 'Awesome') }
+
+      it "does not strip user's title if not exact match" do
+        user.update_columns(title: 'Different')
+        expect { group.remove(user) }.to_not change { user.reload.title }
```
SamSaffron: one downside of this reworking is that over time it makes the test suite slower.

If we can do 2 assertions after the change it is cheaper than repeating the group creation, adding user, removing user N times. 

Can you rework this change overall to avoid large amounts of non-reused db work?

spec/models/group_spec.rb
```diff
@@ -194,70 +194,30 @@ def real_staff
...
-    group.remove(user)
-
-    user.reload
-    expect(user.title).to eq nil
+  describe '#title=' do
+    it "updates the member's title only if it was blank or exact match" do
+      group.add(user)
 
-    group.add(user)
-    group.destroy
```
SamSaffron: hmm we lost the spec on what happens when the group is destroyed?

SamSaffron:
cool 👍  merged it in

PR #6404, opened 2018-09-16
FEATURE: match user title when primary group changes

xrav3nz:

When primary group changes and the user's title is the previous primary group's title, change the title to the new primary group's title

---

- [x] More specs on existing `Group` behavior (https://github.com/discourse/discourse/pull/6403)
- [x] Switch title if exact match and primary changes (this PR)
- [ ] When stripping a title, we should check if there is another title we can grant the user

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1815,28 +1815,15 @@ describe User do
   end
 
   describe '#match_title_to_primary_group_changes' do
-    let(:previous_primary_group) { Fabricate(:group, title: 'Previous', users: [user]) }
-    let(:new_primary_group) { Fabricate(:group, title: 'New', users: [user]) }
+    let(:primary_group_a) { Fabricate(:group, title: 'A', users: [user]) }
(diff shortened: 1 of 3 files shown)


SamSaffron:
Ideally let's keep "work in progress" PRs in your own branch, you can always link me to it 😄

xrav3nz:
I actually meant to split it down into separate PRs (thus the three checkboxes linking to different PRs). But for sure I will keep that in mind.

spec/models/user_spec.rb
```diff
@@ -1814,4 +1814,29 @@ def filter_by(method)
...
+    let(:new_primary_group) { Fabricate(:group, title: 'New', users: [user]) }
+
+    before { user.update(primary_group: previous_primary_group) }
+
+    def change_primary_group
+      user.update(primary_group: new_primary_group)
+    end
+
+    it "updates user's title when it matches the previous primary group" do
+      user.update(title: 'Previous')
```
SamSaffron: lets merge all these 3 specs if we can cleanly? Overall the change looks good to me. 👍

SamSaffron:
looks good to me!

PR #6408, opened 2018-09-17
FIX: improves themes/components UI on mobile

jjaffeux:


SamSaffron:
I think whatever you do here is probably fine, it is conflicting atm, can you correct and merge in?

PR #6409, opened 2018-09-18
extract client side `Discourse` setup inline JS

xrav3nz:
Manually tested in <kbd>F12</kbd> that all the `Discourse` vars are set up properly 😅

Will be able to ✂️ `common/_discourse_javascript.html.erb` in the next PR.

### DIFF
--- a/app/assets/javascripts/preload-application-data.js
+++ b/app/assets/javascripts/preload-application-data.js
@@ -1,12 +1,42 @@
 (function() {
+  var ps = require("preload-store").default;
   var preloadedDataElement = document.getElementById("data-preloaded");
 
   if (preloadedDataElement) {
-    var ps = require("preload-store").default;
     var preloaded = JSON.parse(preloadedDataElement.dataset.preloaded);
 
     Object.keys(preloaded).forEach(function(key) {
       ps.store(key, JSON.parse(preloaded[key]));
     });
+  }
+
+  var setupData = document.getElementById("data-discourse-setup").dataset;
+
+  Discourse.CDN = setupData.cdn;
+  Discourse.BaseUrl = setupData.baseUrl;
+  Discourse.BaseUri = setupData.baseUri;
+  Discourse.Environment = setupData.environment;
+  Discourse.SiteSettings = ps.get("siteSettings");
+  Discourse.ThemeSettings = ps.get("themeSettings");
+  Discourse.LetterAvatarVersion = setupData.letterAvatarVersion;
+  Discourse.MarkdownItURL = setupData.markdownItUrl;
+  Discourse.ServiceWorkerURL = setupData.serviceWorkerUrl;
+  I18n.defaultLocale = setupData.defaultLocale;
+  Discourse.start();
+  Discourse.set("assetVersion", setupData.assetVersion);
+  Discourse.Session.currentProp(
+    "disableCustomCSS",
+    setupData.disableCustomCss === "true"
+  );
+
+  if (setupData.safeMode) {
+    Discourse.Session.currentProp("safe_mode", setupData.safeMode);
+  }
+
+  Discourse.HighlightJSPath = setupData.highlightJsPath;
+
+  if (setupData.S3BaseUrl) {
+    Discourse.S3CDN = setupData.s3Cdn;
+    Discourse.S3BaseUrl = setupData.s3BaseUrl;
   }
 })();
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -307,13 +307,13 @@ module ApplicationHelper
   end
 
   def normalized_safe_mode
-    safe_mode = nil
-    (safe_mode ||= []) << ApplicationController::NO_CUSTOM if customization_disabled?
-    (safe_mode ||= []) << ApplicationController::NO_PLUGINS if !allow_plugins?
-    (safe_mode ||= []) << ApplicationController::ONLY_OFFICIAL if !allow_third_party_plugins?
-    if safe_mode
-      safe_mode.join(",").html_safe
-    end
+    safe_mode = []
+
+    safe_mode << ApplicationController::NO_CUSTOM if customization_disabled?
+    safe_mode << ApplicationController::NO_PLUGINS if !allow_plugins?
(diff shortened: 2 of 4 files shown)


app/views/layouts/application.html.erb
```diff
@@ -17,11 +17,6 @@
       <meta name="shared_session_key" content="<%= shared_session_key %>">
     <%- end %>
 
-    <script>
-      window.EmberENV = window.EmberENV || {};
-      window.EmberENV['FORCE_JQUERY'] = true;
-    </script>
```
xrav3nz: This is already in `env.js` and included in `ember_query.js`

https://github.com/discourse/discourse/blob/ce6a0a5e9e33e3cf7459c180eb566e5f12ec1371/app/assets/javascripts/ember_jquery.js#L1

app/helpers/application_helper.rb
```diff
@@ -406,4 +406,35 @@ def preloaded_json
...
+      highlight_js_path: HighlightJs.path,
+      enable_s3_uploads: SiteSetting.Upload.enable_s3_uploads
+    }
+
+    if guardian.can_enable_safe_mode? && params["safe_mode"]
+      setup_data[:safe_mode] = normalized_safe_mode
+    end
+
+    if SiteSetting.Upload.enable_s3_uploads
+      setup_data[:s3_cdn] = SiteSetting.Upload.s3_cdn_url.presence
```
tgxworld: Hmm we used to check if `SiteSetting.Upload.s3_cdn_url.present?`

app/helpers/application_helper.rb
```diff
@@ -406,4 +406,35 @@ def preloaded_json
     return '{}' if @preloaded.blank?
     @preloaded.transform_values { |value| escape_unicode(value) }.to_json
   end
+
+  def client_side_setup_data
+    service_worker_url = Rails.env.development? ? 'service-worker.js' : Rails.application.assets_manifest.assets['service-worker.js']
+    current_hostname_without_port = RailsMultisite::ConnectionManagement.current_hostname.sub(/:[\d]*$/, '')
```
tgxworld: I would just name the variable `base_url` as well

app/assets/javascripts/preload-application-data.js
```diff
@@ -1,12 +1,42 @@
...
+    setupData.disableCustomCss === "true"
+  );
+
+  if (setupData.safeMode) {
+    Discourse.Session.currentProp("safe_mode", setupData.safeMode);
+  }
+
+  Discourse.HighlightJSPath = setupData.highlightJsPath;
+
+  if (setupData.enableS3Uploads === "true") {
```
tgxworld: Hmm we already do the check on the server side, maybe we can just check if the keys are in the date set and if so we set it.

app/helpers/application_helper.rb
xrav3nz: Yea, took the liberty to use `.presence` because we run an empty check on the value anyway. 😅 but I can change this back

https://github.com/discourse/discourse/blob/1a64b3a4871cceb1b42d5c54ada58ab1241022f6/app/assets/javascripts/discourse.js.es6#L36

app/assets/javascripts/preload-application-data.js
```diff
@@ -1,12 +1,42 @@
...
 
     Object.keys(preloaded).forEach(function(key) {
       ps.store(key, JSON.parse(preloaded[key]));
     });
   }
+
+  var setupData = document.getElementById("data-discourse-setup").dataset;
+
+  Discourse.CDN = setupData.cdn;
+  Discourse.BaseUrl = setupData.baseUrl;
```
tgxworld: Hmm is there something that automatically coverts the key from snake case to camel case?

app/helpers/application_helper.rb
xrav3nz: Personally I think this explains what the regex does better than reading `/:[\d]*$/` every time (my regex is rusty and I actually dug out the original commit to make sure this regex wants to strip port ...)

app/assets/javascripts/preload-application-data.js
xrav3nz: Yes, it is part of the `dataset` property magic: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

> The name of a custom data attribute in JavaScript is the name of the same HTML attribute but in camelCase and with no dashes, dots, etc.

app/helpers/application_helper.rb
tgxworld: Ahh icic then don't worry about this comment. :+1:

app/helpers/application_helper.rb
tgxworld: OK :+1:

app/assets/javascripts/preload-application-data.js
tgxworld: Nice :+1: learnt something new today

tgxworld:
I'm alittle afraid to merge this right before the Singapore meetup. Perhaps we can merge this once we all get back from Singapore?

SamSaffron:
Yay for the change, but it looks like it is conflicting now, can you rebase?

app/views/layouts/application.html.erb
```diff
@@ -58,6 +53,8 @@
     <%= yield :head %>
 
     <%= build_plugin_html 'server:before-head-close' %>
+
+    <%= tag.meta id: 'data-discourse-setup', data: client_side_setup_data %>
```
xrav3nz: Rebased and kept this in `<head>` since it's a lot smaller and a lot more visually pleasing than `preloaded-data`:

```html
<meta id="data-discourse-setup" data-base-url="localhost" data-base-uri="" 
data-environment="development" data-letter-avatar-version="5_06ddb670fcf086fa527f7887a59f15f8" 
data-markdown-it-url="http://localhost:3000/assets/markdown-it-bundle.js" 
data-service-worker-url="service-worker.js" data-default-locale="en" 
data-asset-version="d41d8cd98f00b204e9800998ecf8427e" 
data-disable-custom-css="false" data-highlight-js-path="/highlight-js/localhost/100444ceef8ca161334f64d3ce49fe7a1556f943.js">
```

(and just a tiny bit longer in production with S3)

Let's merge on Monday tho if all else is 👍

PR #6411, opened 2018-09-18
always `loadScript` with a script tag

xrav3nz:
`loadScript` by default uses `jQuery.ajax({ dataType: script })` which “evaluates the response as JavaScript and returns it as plain text”, i.e. inserts an inline script tag. This will cause a Content Security Policy unsafe inline violation.

This PR changes `loadScript` so that it will always load with a `<script>` tag for JavaScript.

@eviltrout could you take a look please? Thanks!

### DIFF
--- a/app/assets/javascripts/discourse/lib/load-script.js.es6
+++ b/app/assets/javascripts/discourse/lib/load-script.js.es6
@@ -44,7 +44,7 @@ export default function loadScript(url, opts) {
   $("script").each((i, tag) => {
     const src = tag.getAttribute("src");
 
-    if (src && (opts.scriptTag || src !== url)) {
+    if (src && src !== url) {
       _loaded[tag.getAttribute("src")] = true;
     }
   });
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/lib/load-script.js.es6
```diff
@@ -86,22 +86,15 @@ export default function loadScript(url, opts) {
       cdnUrl = Discourse.CDN.replace(/\/$/, "") + url;
     }
 
-    // Some javascript depends on the path of where it is loaded (ace editor)
-    // to dynamically load more JS. In that case, add the `scriptTag: true`
-    // option.
-    if (opts.scriptTag) {
-      if (Ember.testing) {
-        throw new Error(
-          `In test mode scripts cannot be loaded async ${cdnUrl}`
```
xrav3nz: This (and `LOAD_ASYNC` in ace component) was introduced in https://github.com/discourse/discourse/commit/e66fa82f3a1d7fc769fe57b801d7f4c17d1b3e78. Since our test suite has been refactored to use `async/await`, I assumed this is no longer relevant? (plus tests still pass)

eviltrout: I suspect this will still be a problem. `loadScript` happens in `didInsertElement` which doesn't not return a promise, so the test suite has no idea when the work has finished.

It's good the tests are passing, but I suspect it might be leaking promises/information. We could merge this optimistically and keep an eye on it.

test/javascripts/lib/load-script-test.js.es6
```diff
@@ -0,0 +1,23 @@
...
+  async assert => {
+    const src = "/javascripts/ace/ace.js";
+
+    await loadScript(src).then(() => {
+      assert.ok(
+        typeof ace !== "undefined",
+        "callbacks should only be executed after the script has fully loaded"
+      );
+
+      const scriptTags = Array.from(document.getElementsByTagName("script"));
```
eviltrout: I imagine this can't use the ember test helpers because the script tag is injected outside of the test sandbox frame?

eviltrout: (probably should add a comment if so - otherwise someone might copy and paste this test and not be searching the dom with the correct restrictions.)

app/assets/javascripts/discourse/lib/load-script.js.es6
xrav3nz: Should this become a problem in the future, do we fallback to using inline scripts in test environment? In that case, is there any other way to test `loadScript`'s behavior other than clicking around in a live instance?

Will merge this after the meetup.

eviltrout: Yes I think if it becomes a problem in our automed test runners (jenkins/travis) we should revert back to the inline script.

I don't there is a good way to test it except manually. Let's give it a try and see what happens.

PR #6412, opened 2018-09-18
Recover public actions (likes) when recovering a post

pmusaraj:
This restores public actions when a staff member recovers a deleted post. 

To avoid restoring actions previously deleted by the user, this only restores actions deleted at or slightly after (5 seconds) the deleted time of the post.

### DIFF
--- a/lib/post_destroyer.rb
+++ b/lib/post_destroyer.rb
@@ -69,6 +69,7 @@ class PostDestroyer
     topic.recover! if @post.is_first_post?
     topic.update_statistics
     recover_user_actions
+    recover_public_post_actions
     DiscourseEvent.trigger(:post_recovered, @post, @opts, @user)
     if @post.is_first_post?
       DiscourseEvent.trigger(:topic_recovered, topic, @user)
@@ -194,6 +195,19 @@ class PostDestroyer
 
     f = PostActionType.public_types.map { |k, _| ["#{k}_count", 0] }
     Post.with_deleted.where(id: @post.id).update_all(Hash[*f.flatten])
+  end
+
+  def recover_public_post_actions
+    public_post_actions = PostAction.publics
+      .where(post_id: @post.id)
+      .with_deleted
+      .where("deleted_at IS NOT NULL")
+      .where(deleted_by_id: @user.id)
+
+    public_post_actions.each { |pa|
+      pa.recover!
+      pa.save
+    }
   end
 
   def agree_with_flags
--- a/lib/post_destroyer.rb
+++ b/lib/post_destroyer.rb
@@ -60,6 +60,7 @@ class PostDestroyer
   end
 
   def recover
+    post_deleted_at = @post.deleted_at
     if @user.staff? && @post.deleted_at
       staff_recovered
     elsif @user.staff? || @user.id == @post.user_id
@@ -69,7 +70,7 @@ class PostDestroyer
     topic.recover! if @post.is_first_post?
     topic.update_statistics
     recover_user_actions
-    recover_public_post_actions
+    recover_public_post_actions(post_deleted_at)
     DiscourseEvent.trigger(:post_recovered, @post, @opts, @user)
     if @post.is_first_post?
       DiscourseEvent.trigger(:topic_recovered, topic, @user)
@@ -197,17 +198,15 @@ class PostDestroyer
     Post.with_deleted.where(id: @post.id).update_all(Hash[*f.flatten])
   end
 
-  def recover_public_post_actions
-    public_post_actions = PostAction.publics
+  def recover_public_post_actions(post_deleted_at)
+    PostAction.publics
       .where(post_id: @post.id)
       .with_deleted
-      .where("deleted_at IS NOT NULL")
-      .where(deleted_by_id: @user.id)
-
-    public_post_actions.each { |pa|
-      pa.recover!
-      pa.save
-    }
+      .where(deleted_at: post_deleted_at..(post_deleted_at + 5.seconds))
+      .find_each do |post_action|
+      post_action.recover!
+      post_action.save
+    end
   end
 
   def agree_with_flags
--- a/lib/post_destroyer.rb
+++ b/lib/post_destroyer.rb
@@ -60,7 +60,6 @@ class PostDestroyer
   end
 
   def recover
-    post_deleted_at = @post.deleted_at
     if @user.staff? && @post.deleted_at
       staff_recovered
     elsif @user.staff? || @user.id == @post.user_id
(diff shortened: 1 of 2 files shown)


lib/post_destroyer.rb
```diff
@@ -196,6 +198,17 @@ def trash_public_post_actions
     Post.with_deleted.where(id: @post.id).update_all(Hash[*f.flatten])
   end
 
+  def recover_public_post_actions(post_deleted_at)
+    PostAction.publics
```
ZogStriP: This is going to generate N+1 queries. 

Why not do `.update_all(deleted_at: nil, deleted_by_id: nil)` instead of the `.find_each` loop?

SamSaffron:
The 5 second rule here makes me very uneasy. 

Can we change it so we store all the "recoverable" ids in a custom post field on delete, it makes it laser safe to recover without relying on clocks.

lib/post_destroyer.rb
```diff
@@ -196,6 +198,16 @@ def trash_public_post_actions
...
+      .where(post_id: @post.id)
+      .with_deleted
+      .where(deleted_at: post_deleted_at..(post_deleted_at + 5.seconds))
```
tgxworld: A small one but we usually try to avoid chaining multiple `where`. 

```
.where(
  post_id: @post.id,
  deleted_at: post_deleted_at..(post_deleted_at + 5.seconds)
)
```
is faster than

```
.where(post_id: @post.id)
.where(deleted_at: post_deleted_at..(post_deleted_at + 5.seconds))
```

spec/components/post_destroyer_spec.rb
```diff
@@ -217,6 +217,32 @@
...
   end
 
+  describe "recovery and public post actions" do
+    let(:codinghorror) { Fabricate(:coding_horror) }
+    let!(:like) { PostAction.act(codinghorror, post, PostActionType.types[:like]) }
+    let!(:another_like) { PostAction.act(moderator, post, PostActionType.types[:like]) }
+
+    it "restores public post actions" do
+      PostDestroyer.new(moderator, post).destroy
+      expect(PostAction.find_by(id: like.id)).not_to be_present
```
tgxworld: Since we don't need to fetch the record here, `PostAction.exists?(id: like.id)` is going to use less memory and be faster.

lib/post_destroyer.rb
```diff
@@ -196,6 +198,16 @@ def trash_public_post_actions
...
+      .update_all(deleted_at: nil, deleted_by_id: nil)
+
+    @post.post_actions.each(&:update_counters)
```
tgxworld: We need to test that the counters are updated.

tgxworld: @ZogStriP Hmm if we need to update the counters here anyway, it'll probably be better to just do `find_each` and call `recover!`

ZogStriP: @pmusaraj can you find a way to update the counters once per `post_action_type`?

lib/post_destroyer.rb
davidtaylorhq: Why would combining them be more performant? I think ActiveRecord is smart enough to combine multiple `where` into a single SQL query - it doesn't hit the database for each one.

tgxworld: O the SQL is the same. What I meant is that if you benchmark the Ruby code, it'll be faster

lib/post_destroyer.rb
pmusaraj: @ZogStriP I couldn't find a way to update the counters once for `post_action_type`. I tried this below: 

    PostActionType.public_types.each do |z, type_id|
      @post.post_actions.where(post_action_type_id: type_id).last(&:update_counters)
    end

but it didn't work. So I went back to `find_each` and `recover!` as Alan suggested.

lib/post_destroyer.rb
```diff
@@ -78,6 +78,7 @@ def recover
 
   def staff_recovered
     @post.recover!
+    recover_public_post_actions
```
tgxworld: I think this needs to happens in the same transaction as `@post.recover!`.

lib/post_destroyer.rb
```diff
@@ -190,10 +191,28 @@ def feature_users_in_the_topic
...
+    end
+  end
+
+  def recover_public_post_actions
+    PostAction.publics
+      .with_deleted
+      .where(post_id: @post.id, id: @post.custom_fields["deleted_public_actions"])
+      .find_each do |post_action|
+        post_action.recover!
+        post_action.save
```
tgxworld: `save!` :wink:

spec/components/post_destroyer_spec.rb
```diff
@@ -217,6 +217,45 @@
...
+    end
+
+    it "updates post like count" do
+      PostDestroyer.new(moderator, post).destroy
+      PostDestroyer.new(moderator, post).recover
+      post.reload
+      expect(post.like_count).to eq(2)
+    end
+
+    it "cleans up deleted_public_actions custom field after recovery" do
```
tgxworld: Since this test has the same setup as "updates post like count", I think we can just run the ` expect(post.custom_fields["deleted_public_actions"]).to be_nil` in that test and do without this one.

app/models/post.rb
```diff
@@ -375,6 +376,19 @@ def update_flagged_posts_count
     PostAction.update_flagged_posts_count
   end
 
+  def recover_public_post_actions
+    PostAction.publics
+      .with_deleted
+      .where(post_id: self.id, id: self.custom_fields["deleted_public_actions"])
```
ZogStriP: 👍

lib/post_destroyer.rb
```diff
@@ -190,10 +190,15 @@ def feature_users_in_the_topic
 
   def trash_public_post_actions
     public_post_actions = PostAction.publics.where(post_id: @post.id)
```
ZogStriP: You can merge both lines

```ruby
public_post_actions = PostAction.publics.where(post_id: @post.id)
if public_post_actions
```

to

```
if public_post_actions = PostAction.publics.where(post_id: @post.id)
```

spec/components/post_destroyer_spec.rb
```diff
@@ -217,6 +217,38 @@
...
+
+    it "restores public post actions" do
+      PostDestroyer.new(moderator, post).destroy
+      expect(PostAction.exists?(id: like.id)).to eq(false)
+
+      PostDestroyer.new(moderator, post).recover
+      expect(PostAction.exists?(id: like.id)).to eq(true)
+    end
+
+    it "does not recover previously-deleted actions" do
```
ZogStriP: 👍

PR #6416, opened 2018-09-20
FEATURE: auto grant an available title when removing old title

xrav3nz:

When revoking an old title (due to leaving the group that grants the said title), this feature will grant one of the other available titles automatically.

We will pick the first available title in the following order:

1. user's current primary group title
2. title from a user's group that automatically set primary group
3. user's other group titles
4. user's badge titles

### DIFF
--- a/app/models/group_user.rb
+++ b/app/models/group_user.rb
@@ -45,7 +45,7 @@ class GroupUser < ActiveRecord::Base
 
   def grant_other_available_title
     if group.title.present? && group.title == user.title
-      user.update(title: user.available_titles.first)
+      user.update!(title: user.next_best_title)
     end
   end
 
@@ -81,7 +81,6 @@ class GroupUser < ActiveRecord::Base
     user.update!(group_locked_trust_level: highest_level)
     Promotion.recalculate(user)
   end
-
 end
 
 # == Schema Information
--- a/app/models/group_user.rb
+++ b/app/models/group_user.rb
@@ -5,7 +5,7 @@ class GroupUser < ActiveRecord::Base
   belongs_to :user
 
   after_save :update_title
-  after_destroy :remove_title
+  after_destroy :grant_other_available_title
 
   after_save :set_primary_group
   after_destroy :remove_primary_group, :recalculate_trust_level
@@ -43,13 +43,9 @@ class GroupUser < ActiveRecord::Base
     )
   end
 
-  def remove_title
-    if group.title.present?
-      DB.exec("
-        UPDATE users SET title = NULL
-        WHERE title = :title AND id = :id",
-        id: user_id, title: group.title
-      )
+  def grant_other_available_title
+    if group.title.present? && group.title == user.title
+      user.update(title: user.available_titles.first)
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -109,6 +109,7 @@ class User < ActiveRecord::Base
   before_save :update_username_lower
   before_save :ensure_password_is_hashed
   before_save :match_title_to_primary_group_changes
+  before_save :check_if_title_is_badged_granted
 
   after_save :expire_tokens_if_password_changed
   after_save :clear_global_notice_if_needed
@@ -999,13 +1000,6 @@ class User < ActiveRecord::Base
     @user_fields
   end
 
-  def title=(val)
-    write_attribute(:title, val)
-    if !new_record? && user_profile
-      user_profile.update_column(:badge_granted_title, false)
-    end
-  end
-
   def number_of_deleted_posts
     Post.with_deleted
       .where(user_id: self.id)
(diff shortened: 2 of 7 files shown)


app/services/user_updater.rb
```diff
@@ -66,9 +66,6 @@ def update(attributes = {})
       attributes[:title] != user.title &&
       guardian.can_grant_title?(user, attributes[:title])
       user.title = attributes[:title]
-      if user.badges.where(name: user.title).exists?
-        user_profile.badge_granted_title = true
-      end
```
xrav3nz: existing specs to make sure removing this does not regress:

https://github.com/discourse/discourse/blob/d30e45a45832f29cfd41ea4cc144d0032f43c2f7/spec/services/user_updater_spec.rb#L204-L219

app/models/group_user.rb
```diff
@@ -43,13 +43,9 @@ def remove_primary_group
...
-  def remove_title
-    if group.title.present?
-      DB.exec("
-        UPDATE users SET title = NULL
-        WHERE title = :title AND id = :id",
-        id: user_id, title: group.title
-      )
+  def grant_other_available_title
+    if group.title.present? && group.title == user.title
+      user.update(title: user.available_titles.first)
```
tgxworld: Whenever we're not checking for the return value, we want to use `update!` so that this doesn't fail silently.

app/models/user.rb
```diff
@@ -1122,6 +1116,18 @@ def mature_staged?
     from_staged? && self.created_at && self.created_at < 1.day.ago
   end
 
+  def available_titles
+    group_titles_query = groups
+    group_titles_query = group_titles_query.order("groups.id = #{primary_group_id} DESC") if primary_group_id
+    group_titles_query = group_titles_query.order("groups.primary_group DESC")
+
+    group_titles = group_titles_query.pluck(:title).reject(&:blank?)
```
tgxworld: hmm perhaps we can check for a blank title when fetching the recrod?

app/models/user.rb
```diff
@@ -1122,6 +1116,18 @@ def mature_staged?
...
+    badge_titles = badges.where(allow_title: true).pluck(:name)
+      .map { |name| Badge.display_name(name) }
+
+    group_titles + badge_titles
```
tgxworld: Using `concat` will avoid an extra array allocation.

app/models/user.rb
xrav3nz: wanted to do it in the query, but there isn't an index on `title`.

spec/models/badge_spec.rb
```diff
@@ -80,4 +80,21 @@
...
+  describe '.i18n_name' do
+    it 'transforms to lower case letters, and replaces spaces with underscores' do
+      expect(Badge.i18n_name('Basic User')).to eq('basic_user')
+    end
+  end
+
+  describe '.display_name' do
+    it 'fetches from translations when i18n_name key exists' do
+      expect(Badge.display_name('basic_user')).to eq('Basic')
+      expect(Badge.display_name('Basic User')).to eq('Basic')
```
tgxworld: I usually prefer to avoid repetition in the code base by looping instead. 

```
["basic_user", "Basic User"].each { |name| expect(Badge.display_name(name)).to eq('Basic')  }
```

spec/models/group_spec.rb
```diff
@@ -675,13 +675,22 @@ def can_view?(user, group)
   describe '#remove' do
     before { group.add(user) }
 
-    it "only strips user's title if exact match" do
-      group.update(title: 'Awesome')
-      expect { group.remove(user) }.to change { user.reload.title }.from('Awesome').to(nil)
+    context 'when stripping title' do
+      it "only strips user's title if exact match" do
+        group.update(title: 'Awesome')
```
tgxworld: Prefer update! smile_cat

spec/models/user_spec.rb
```diff
@@ -1826,4 +1826,64 @@ def filter_by(method)
...
     end
   end
+
+  describe '#title=' do
+    let(:badge) { Fabricate(:badge, name: 'Badge', allow_title: false) }
+
+    it 'sets badge_granted_title correctly' do
+      BadgeGranter.grant(badge, user)
+
+      user.update(title: 'Badge')
```
tgxworld: instead of repeating the string `Badge`, we should use `badge.name` instead.

spec/models/badge_spec.rb
xrav3nz: 👍 on looping, but I think it may overcomplicate the test case here since there are only two. I don't feel strongly about it tho.

app/models/user.rb
tgxworld: Hmm I think this needs to be optimized further, the only place we're using this method calls `.first` on it. That means that we actually don't need to fetch all the titles out of the DB.

xrav3nz: For sure, I will change the behavior and move this to `GroupUser` then!

spec/models/badge_spec.rb
SamSaffron: Yeah I am not a fan of a loop of 2, 20 yeah, 2 is overkill imo, this reads fine to me

spec/models/user_spec.rb
```diff
@@ -1826,4 +1826,75 @@ def filter_by(method)
...
+
+    it 'only includes groups with title' do
+      group_a.add(user)
+      expect(user.next_best_title).to eq('Group A')
+
+      group_a.update!(title: nil)
+      expect(user.next_best_title).to eq(nil)
+    end
+
+    it 'onlly inclues badges that allow to be set as title' do
```
SamSaffron: tiny typo

spec/models/user_spec.rb
```diff
@@ -1826,4 +1826,75 @@ def filter_by(method)
...
+      BadgeGranter.grant(badge, user)
+      expect(user.next_best_title).to eq('Badge')
+
+      badge.update!(allow_title: false)
+      expect(user.next_best_title).to eq(nil)
+    end
+
+    it "picks the next best title in the order: user's primary group, primary group, groups, and badges" do
+      group_a.add(user)
```
SamSaffron: good test

SamSaffron:
excellent lets try this out!

PR #6419, opened 2018-09-20
FIX: Avoid redirection when not leaving forum.

nbianca:
Please review this one carefully, the code is pretty old & messy.

### DIFF
--- a/app/assets/javascripts/discourse/lib/click-track.js.es6
+++ b/app/assets/javascripts/discourse/lib/click-track.js.es6
@@ -128,16 +128,18 @@ export default {
     const isInternal = DiscourseURL.isInternal(href);
 
     // If we're on the same site, use the router and track via AJAX
-    if (tracking && isInternal && !$link.hasClass("attachment")) {
-      ajax("/clicks/track", {
-        data: {
-          url: href,
-          post_id: postId,
-          topic_id: topicId,
-          redirect: false
-        },
-        dataType: "html"
-      });
+    if (isInternal && !$link.hasClass("attachment")) {
+      if (tracking) {
+        ajax("/clicks/track", {
+          data: {
+            url: href,
+            post_id: postId,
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/lib/click-track.js.es6
```diff
@@ -128,16 +128,18 @@ export default {
...
-        data: {
-          url: href,
-          post_id: postId,
-          topic_id: topicId,
-          redirect: false
-        },
-        dataType: "html"
-      });
+    if (isInternal && !$link.hasClass("attachment")) {
+      if (tracking) {
```
ZogStriP: 👍 but we definitely need a test here.

app/assets/javascripts/discourse/lib/url.js.es6
```diff
@@ -187,13 +187,7 @@ const DiscourseURL = Ember.Object.extend({
...
     const baseUri = Discourse.BaseUri;
 
-    // If we have a baseUri and an absolute URL, make sure the baseUri
-    // is the same. Otherwise we could be switching forums.
-    if (
-      baseUri &&
-      path.indexOf("http") === 0 &&
-      pathname.indexOf(baseUri) !== 0
-    ) {
+    if (!DiscourseURL.isInternal(path)) {
```
ZogStriP: Are there tests ensuring this is the same behavior?

nbianca: Hmm... I think this might behave different in subdirectory installations.

ZogStriP: Yeah, you definitely don't want to break subfolder support ;)

nbianca: I tested this on a subfolder setup and looks good to me.

PR #6420, opened 2018-09-21
DEV: Run prettier.

nbianca:


ZogStriP:
Not sure how we missed those files 🤔

PR #6421, opened 2018-09-21
FEATURE: Display trust level descriptions in site settings

davidtaylorhq:
Adds the translated name of the trust level alongside the number

## Before:
screen shot 2018-09-21 at 12 33 08

## After:
screen shot 2018-09-21 at 12 21 07

### DIFF
--- a/app/models/trust_level_setting.rb
+++ b/app/models/trust_level_setting.rb
@@ -8,7 +8,8 @@ class TrustLevelSetting < EnumSiteSetting
   end
 
   def self.values
-    @values ||= valid_values.map { |x| { name: x.to_s, value: x } }
+    levels = TrustLevel.all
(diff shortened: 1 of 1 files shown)


app/models/trust_level_setting.rb
```diff
@@ -8,7 +8,8 @@ def self.valid_value?(val)
   end
 
   def self.values
-    @values ||= valid_values.map { |x| { name: x.to_s, value: x } }
+    levels = TrustLevel.all
+    @values ||= valid_values.map { |x| { name: "#{x}: #{levels[x.to_i].name}", value: x } }
```
ZogStriP: Is `.name` the translated name? If not, we should use the translated version instead.

davidtaylorhq: It is already translated in https://github.com/discourse/discourse/blob/trust-level-settings/lib/trust_level.rb#L41

ZogStriP:
LGTM 👍

PR #6422, opened 2018-09-21
UX: Improve error messages for minimum and maximum username lengths.

nbianca:


lib/validators/max_username_length_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize(opts = {})
+    @opts = opts
+    @username = ""
+  end
+
+  def valid_value?(val)
+    return false if val < SiteSetting.min_username_length
+
+    users = User.where('length(username) > ?', val).limit(1)
```
ZogStriP: It's even faster to ask postgresql if a record exists?

```ruby
return true unless User.exists?("length(username) > ?", val)
```

lib/validators/max_username_length_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+    return true if users.size == 0
+
+    @username = users[0].username
+    false
+  end
+
+  def error_message
+    if @username == ""
```
ZogStriP: `if @username.blank?` is more readable here ;)

lib/validators/max_username_length_validator.rb
ZogStriP: Actually, nevermind. You use the username later on. Carry on :)

tgxworld:
Merged in https://github.com/discourse/discourse/commit/e0d7cdac12a180697b3939967c8315184899bfa4. @nbianca There are some minor changes I made but nothing major. Nice work :+1:

PR #6423, opened 2018-09-21
PERF: Quit out of the email job quickly if disabled

eviltrout:
This prevents sidekiq from doing a bunch of queries when email is
disabled.

Critical emails are a special case and will be sent.

### DIFF
--- a/app/jobs/regular/user_email.rb
+++ b/app/jobs/regular/user_email.rb
@@ -9,9 +9,19 @@ module Jobs
 
     sidekiq_options queue: 'low'
 
+    # Can be overridden by subclass, for example critical email
+    # should always consider being sent
+    def quit_email_early?
+      SiteSetting.disable_emails == 'yes'
+    end
+
     def execute(args)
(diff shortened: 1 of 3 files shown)


app/jobs/regular/user_email.rb
```diff
@@ -9,10 +9,20 @@ class UserEmail < Jobs::Base
...
+      SiteSetting.disable_emails == 'yes'
+    end
+
     def execute(args)
       raise Discourse::InvalidParameters.new(:user_id) unless args[:user_id].present?
       raise Discourse::InvalidParameters.new(:type)    unless args[:type].present?
 
+      # This is for performance. Quit out fast without doing a bunch
+      # of extra work when emails are disabled.
+      return if quit_email_early?
```
ZogStriP: Should we prevent the job to even be scheduled in the first place when emails are disabled?

eviltrout: Yes but that’s trickier, because some emails will go out if disabled.

PR #6424, opened 2018-09-24
FIX: make theme import's public key field `readonly`

xrav3nz:
> CONTEXT & DISCUSSION: https://meta.discourse.org/t/unable-to-copy-public-key-when-installing-theme-from-private-repository-using-firefox/97849?u=xrav3nz

`disabled` attribute prevents the user from clicking or selecting in the control, whereas `readonly` does not.


xrav3nz:
Tested on Firefox 62.0.2 (macOS) and will merge on CI ✅

PR #6425, opened 2018-09-25
FIX: progress bar position when theme has header

pmusaraj:
When there is content in the theme's header, the position of the topic progress bar on screens below 975px wide is off by the header's height. 

This happens because when calculating the docked topic progress bar's position, we use `#topic-bottom`'s top offset (which relates to the document), and `#main`'s height (which matches document height only if there are no elements between it and the body). We should use the `body` height instead. 

See this meta topic for a longer discussion with screenshots: https://meta.discourse.org/t/adding-header-in-theme-messes-up-topic-progress-bar-when-docked/97393/6

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-progress.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-progress.js.es6
@@ -162,7 +162,7 @@ export default Ember.Component.extend({
     const windowHeight = $(window).height();
     const composerHeight = $("#reply-control").height() || 0;
     const isDocked = offset >= maximumOffset - windowHeight + composerHeight;
-    const bottom = $("#main").height() - maximumOffset;
+    const bottom = $("body").height() - maximumOffset;
 
     if (composerHeight > 0) {
       $wrapper.css("bottom", isDocked ? bottom : composerHeight);


ZogStriP:
Have you tested on desktop/mobile with and without custom headers?

pmusaraj:
Yes, I did, and I did some more tests today. Looks good with/without a custom header on all tests, I've tested the following browsers: 

Desktop, window width less than 975px: Safari, Firefox, Chrome on Mac, IE11 on Windows 8 (via Browserstack). 
Mobile: Safari iOS (in Simulator), Chrome on Google Pixel (via Browserstack).

SamSaffron:
I support merging this but do worry that many have depended on this quirk. 

Does https://forums.thimbleweedpark.com continue to work properly after this ?

pmusaraj:
https://forums.thimbleweedpark.com/ should work fine. I had a quick look at their markup, and they don't have an extra header (whether inside or outside of #main), so this shouldn't affect them. 

I also tested the fix with https://meta.discourse.org/t/brand-header-theme-component/77977 and it doesn't affect that header technique either. 

This could affect sites that have implemented additional workarounds to compensate for the previous discrepancy in the calculation.

SamSaffron:
Ok lets go for it!

PR #6426, opened 2018-09-26
UX: confirmation before changing group membership in admin

xrav3nz:
[Video demo](https://www.youtube.com/watch?v=a9v-7SZE1ok) of the change.

---

Currently, group memberships are granted/revoked automatically when selecting/deselecting a custom group, this is inconsistent with the other fields where the  ✔️ ❌ buttons are available.

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-user-index.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-user-index.js.es6
@@ -11,6 +11,7 @@ export default Ember.Controller.extend(CanCheckEmails, {
   editingName: false,
   editingTitle: false,
   originalPrimaryGroupId: null,
+  customGroupIdsBuffer: null,
   availableGroups: null,
   userTitleValue: null,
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -278,20 +314,22 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+      const bufferedIds = this.get("customGroupIdsBuffer");
+      const availableGroups = this.get("availableGroups");
+
+      bufferedIds.filter(id => !currentIds.includes(id)).forEach(id => {
+        this.groupAdded(availableGroups.findBy("id", id));
+      });
+
+      currentIds
+        .filter(id => !bufferedIds.includes(id))
+        .forEach(id => this.groupRemoved(id));
```
xrav3nz: Note, even though this is issuing the `POST`/`DELETE` requests one by one, I don't think it warrants a new endpoint.

tgxworld:
LGTM :+1:

SamSaffron:
looks great to me, only minor bit of feedback, if you make multiple changes and they all fail this will be bootbox central, and state will be ... indeterminate. 

I think we can probably live with this for now, but it is worth noting.

PR #6428, opened 2018-09-26
move large blobs out of `<head>`

xrav3nz:
Thanks @SamSaffron for bringing this up: the `data-preloaded` JSON unnecessarily bloats the `<head>` section, and increases the payload dramatically for reading open graph tags, so moving this and future large blobs of data out of `<head>`.

Sanity check that everything still works:

image


SamSaffron:
Looks good to me 👍

PR #6429, opened 2018-09-27
FEATURE: replace github links to files with permanent links

majakomel:
https://meta.discourse.org/t/meta-links-to-github-often-broken/85151

### DIFF
--- a/app/jobs/regular/replace_github_non_permalinks.rb
+++ b/app/jobs/regular/replace_github_non_permalinks.rb
@@ -1,0 +1,69 @@
+require 'net/http'
+require 'json'
+require 'uri'
+
+module Jobs
+
+  class ReplaceGithubNonPermalinks < Jobs::Base
+    sidekiq_options queue: 'low'
+
+    def execute(args)
+      post_id = args[:post_id]
+      raise Discourse::InvalidParameters.new(:post_id) unless post_id.present?
+
+      post = Post.find_by(id: post_id)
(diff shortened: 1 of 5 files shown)


SamSaffron:
I like this change, but don't feel like it belongs in core, maybe we create a plugin for this? Also, this behaviour is kind of intrusive so people should be allowed to opt out.

app/jobs/regular/replace_github_non_permalinks.rb
```diff
@@ -0,0 +1,69 @@
...
+      post = Post.find_by(id: post_id)
+      return unless post.present?
+
+      raw = post.raw.dup
+      start_raw = raw.dup
+
+      regex = /github\.com\/(?<user>[^\/]+)\/(?<repo>[^\/\s]+)\/blob\/(?<sha1>[^\/\s]+)\/(?<file>[^#\s]+)(?<from-to>#(L([^-\s]*)(-L(\d*))?))?/i
+
+      matches = post.raw.scan(regex)
+      matches.each do |url|
```
ZogStriP: You can extract the values in the proc parameters ;)

```ruby
matches.each do |user, repo, sha1, file, from_to|
...
end
```

SamSaffron:
I am totally with @gschlager  here let's create a new `discourse-github` plugin and roll this plus the other 2 plugins into it !! 

Going to close this for now and make a separate TODO

PR #6431, opened 2018-09-27
FEATURE: allow multiple secrets for Discourse SSO provider

majakomel:
https://meta.discourse.org/t/supporting-multiple-sso-secrets-with-discourse-as-provider/96693

### DIFF
--- a/lib/single_sign_on.rb
+++ b/lib/single_sign_on.rb
@@ -50,9 +50,14 @@ class SingleSignOn
 
   def self.parse(payload, sso_secret = nil)
     sso = new
-    sso.sso_secret = sso_secret if sso_secret
 
     parsed = Rack::Utils.parse_query(payload)
+    decoded = Base64.decode64(parsed["sso"])
+    decoded_hash = Rack::Utils.parse_query(decoded)
+
+    return_sso_url = decoded_hash['return_sso_url']
+    sso.sso_secret = sso_secret || (provider_secret(return_sso_url) if return_sso_url)
+
     if sso.sign(parsed["sso"]) != parsed["sig"]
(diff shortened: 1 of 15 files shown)


SamSaffron:
Awesome feature @majakomel  can you do a quick screenshot of how this looks with your change?

majakomel:
@SamSaffron I'd like to improve the flow from the user perspective, will consult @jjaffeux tomorrow and and then post screenshots

gschlager:
I think it would be better to use only `sso_provider_secrets` and not fallback to `sso_secret`. Maybe migrate the value of `sso_secret` as `*|old_sso_secret`?

jjaffeux:
Just a note here, I had a call with @majakomel and we will modify the UI here, the basic of the feature are here but we will try to achieve a better experience.

majakomel:
This is how sso provider secrets site setting looks like:

screen shot 2018-10-05 at 14 11 00

lib/single_sign_on.rb
```diff
@@ -83,6 +88,18 @@ def self.parse(payload, sso_secret = nil)
     sso
   end
 
+  def self.provider_secret(return_sso_url)
+    provider_secrets = SiteSetting.sso_provider_secrets.split(/[\|,\n]/)
+    provider_secrets_hash = Hash[*provider_secrets]
+    return_url = URI.parse(return_sso_url).host
+
+    secret = provider_secrets_hash.select do |domain, _|
+      domain_regex = Regexp.escape(domain).gsub("\\*", '\S*')
```
SamSaffron: Is there a bug here ... Say we allow. `*.discourse.org` will `bad-domain.discourse.org.evil.com` slip through? maybe extract some of this logic into a service object of sorts and add some tests for the security aspects?

SamSaffron:
Thanks heaps @majakomel  👍  merging it in!

PR #6433, opened 2018-09-28
FIX: hide Delete Category button while creating a new category

prssanna:
screen shot 2018-09-28 at 6 09 42 pm

### DIFF
--- a/app/assets/javascripts/discourse/templates/modal/edit-category.hbs
+++ b/app/assets/javascripts/discourse/templates/modal/edit-category.hbs
@@ -27,7 +27,7 @@
                  action="deleteCategory"
                  icon="trash-o"
                  label="category.delete"}}
-    {{else}}
+    {{else if model.id}}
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/templates/modal/edit-category.hbs
```diff
@@ -28,17 +28,19 @@
                  icon="trash-o"
                  label="category.delete"}}
     {{else}}
```
gschlager: You don't need all those changes in `edit-category.js.es6` and in this file if you replace the `{{else}}` with `{{else if model.id}}` 😉

prssanna: Ouch...that was so much better 😂
I've updated the code, thanks @gschlager!

gschlager:
Thanks! 👍

PR #6437, opened 2018-10-01
FIX: force enable a user's email_private_messages option when user emails group

pmusaraj:


tgxworld:
Merged in https://github.com/discourse/discourse/commit/da9eee52625cb2d12158cf185a5f983c220aaf0f. 

Thanks @pmusaraj

PR #6438, opened 2018-10-01
FIX: Make error in Discourse Hub more descriptive.

nbianca:


lib/discourse_hub.rb
```diff
@@ -52,14 +52,25 @@ def self.singular_action(action, rel_url, params = {})
...
       {
         body: JSON[params],
         headers: { 'Referer' => referer, 'Accept' => accepts.join(', '), "Content-Type" => "application/json" },
         omit_default_port: true
       }.merge(connect_opts)
-    ).body)
+    )
+
+    if response.status != 200
+      log(:warninig, "Discourse Hub (#{hub_base_url}#{rel_url}) returned a bad status #{response.status}.")
```
SamSaffron: typos there, should be :warning

lib/discourse_hub.rb
```diff
@@ -52,14 +52,25 @@ def self.singular_action(action, rel_url, params = {})
...
+    end
+
+    begin
+      JSON.parse(response.body)
+    rescue
```
tgxworld: When rescuing errors, we want to avoid rescuing all types of errors because errors like a typo will be incorrectly rescued as well. In this case, we should only rescue `JSON::ParserError`.

spec/components/discourse_hub_spec.rb
```diff
@@ -75,4 +75,25 @@
...
+    it 'should log a warning if status is not 200' do
+      stub_request(:get, (ENV['HUB_BASE_URL'] || "http://local.hub:3000/api")).
+        to_return(status: 500, body: "", headers: {})
+
+      Rails.logger.expects(:warn)
+
+      DiscourseHub.collection_action(:get, "")
+    end
+
+    it 'should log an error if status is not 200' do
```
ZogStriP: status is 200 but it's not a valid JSON ;)

PR #6441, opened 2018-10-02
✂️ `discourse_javascript.html.erb`

xrav3nz:
This extracts the last bits of inline JavaScript in our HTML partials, and remove `discourse_javascript.html.erb` 🎉


app/assets/javascripts/authentication-complete.js
```diff
@@ -0,0 +1,17 @@
...
+  const authenticationData = JSON.parse(
+    document.getElementById("data-authentication").dataset.authenticationData
+  );
+
+  Discourse.showingSignup = true;
+  require("discourse/routes/application").default.reopen({
+    actions: {
+      didTransition: function() {
+        Em.run.next(function() {
+          Discourse.authenticationComplete(authenticationData);
```
xrav3nz: Tested this locally with the Linkedin OAuth plugin with [`full_screen_login: true`](https://github.com/discourse/discourse/blob/a515ba86120d605cdd5b77f36ecff94169dd229a/app/controllers/users/omniauth_callbacks_controller.rb#L65).

app/assets/javascripts/preload-application-data.js
```diff
@@ -39,4 +39,26 @@
...
     Discourse.S3BaseUrl = setupData.s3BaseUrl;
   }
+
+  Ember.RSVP.configure("onerror", function(e) {
+    // Ignore TransitionAborted exceptions that bubble up
+    if (e && e.message === "TransitionAborted") {
+      return;
+    }
+
+    if (Discourse.Environment === "development") {
```
xrav3nz: I moved this here because this is the earliest place that we have access to the initialized `Discourse.Environment`.

coding-horror:
"In all browsers" basically means "does it work in IE11" since everything
else we support is at recent / modern versions.

eviltrout:
@coding-horror true but we want to transpile it anyway so we don't have to think about whether a particular feature works. `const` seems to but we use a lot of advanced JS and I'm not sure there's a good reason to limit some files to what browsers support.

xrav3nz:
This was mainly because I couldn't find a way for the transpiled code to run when linked with `<script>` 😅, as they were transpiled into module definitions (not sure if this is the right term).

Roughly along the lines of:

```js
// hello.js.es6
console.log("hello");

// transpiled hello.js
define('hello', [], function() { ... });
```

Is there a way to run them?

eviltrout:
@xrav3nz you can use requirejs to find the module and execute it. For example if it's called `hello` you should be able to say `require('hello').default()` to execute the default exported function.

xrav3nz:
Yes, but wouldn't that need to be a regular non-transpiled `.js`?

eviltrout:
If CSP prevents us from all inline JS then we might need to add a secondary type of sprockets handler that does not produce a module, but does the ES6 transpiling. I believe that is an option to babel.

SamSaffron:
At a minimum here @xrav3nz we got to make sure we use no es6 features at all in any `*.js` file, they all must be in `*.js.es6` files, to avoid needing to do major sprockets rework the easiest path forward is just to manually pass that js file through a transpiler and paste it in to the js file.

xrav3nz:
Absolutely, I made an effort to avoid es6 features (https://github.com/xrav3nz/discourse/commit/f1e25a615d84b8f0db10c624c134af4ab2f68d03) since our uglifier will flip when minimizing any untranspiled es6, but the `const`s slipped through it somehow. 

I have just ran all the new `.js` files through a transpiler.

> add a secondary type of sprockets handler that does not produce a module, but does the ES6 transpiling

That'd be awesome! I'm more than happy to take this as a separate todo, to see how much work is required.

xrav3nz:
Rebased and made the new extracted js a `.no-module.es6`!

SamSaffron:
looks good to me!

PR #6445, opened 2018-10-02
FIX: Correctly censor strings starting or ending with non-word characters

davidtaylorhq:


app/assets/javascripts/pretty-text/censored-words.js.es6
```diff
@@ -28,10 +32,8 @@ export function censorFn(
           "ig"
         );
       } else {
-        censorRegexp = new RegExp(
-          "(\\b(?:" + patterns.join("|") + ")\\b)(?![^\\(]*\\))",
-          "ig"
-        );
+        censorRegexp = createCensorRegexp(patterns);
+        console.log(censorRegexp);
```
danielhollas: extra console.log() :)

PR #6446, opened 2018-10-03
Readme: Remove reference to old dev environment setup blog post

hakusaro:
The old "Discourse as your first rails app" blog post points people in
the direction of Vagrant, which is no longer maintained. That blog post
now points people in the direction of guides that have been wiki-fied on
Meta, which can also be replied to if there are questions. The existing
setup guides appear to be relevant still, but the community wikis are
probably better for new people to reference.

If you ask me, though, there's a great debate here about whether or not
setup documentation should live in this repo or on Meta -- but I don't
have the answer for that. It's more important to get the effectively
dead guide off the front page, since it's Hacktoberfest and I'm sure
people want to contribute, but need a clear starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,13 @@ Browse [lots more notable Discourse instances](https://www.discourse.org/custome
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


coding-horror:
looks good, thank you!!

PR #6447, opened 2018-10-03
extract inline js when baking theme fields

xrav3nz:
**What are extracted?**

- Handlebar template overrides (`type="text/x-handlebars"`)
- scripts with `type="text/discourse-plugin"`, and transpilation errors are also extracted as `console.error`s
- other inline scripts are extracted as-is

I will ship a background job to trigger a rebake of all HTML theme fields in a future PR, so it gives us a bit of time to play with the feature.

### DIFF
--- a/app/controllers/javascripts_controller.rb
+++ b/app/controllers/javascripts_controller.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+class JavascriptsController < ApplicationController
+  DISK_CACHE_PATH = "#{Rails.root}/tmp/javascript-cache"
+
+  skip_before_action(
+    :check_xhr,
+    :handle_theme,
+    :preload_json,
+    :redirect_to_login_if_required,
+    :verify_authenticity_token,
+    only: [:show]
+  )
+
+  before_action :is_asset_path, :no_cookies, only: [:show]
+
+  def show
+    raise Discourse::NotFound unless last_modified.present?
+    return render body: nil, status: 304 if not_modified?
+
+    # Security: safe due to route constraint
+    cache_file = "#{DISK_CACHE_PATH}/#{params[:digest]}.js"
+
+    unless File.exist?(cache_file)
+      content = query.pluck(:content).first
+
+      FileUtils.mkdir_p(DISK_CACHE_PATH)
+      File.write(cache_file, content)
+    end
+
+    set_cache_control_headers
+    send_file(cache_file, disposition: :inline)
+  end
+
+  private
+
+  def query
+    @query ||= JavascriptCache.where(digest: params[:digest]).limit(1)
+  end
+
+  def last_modified
+    @last_modified ||= query.pluck(:updated_at).first
+  end
+
+  def not_modified?
+    cache_time =
+      begin
+        Time.rfc2822(request.env["HTTP_IF_MODIFIED_SINCE"])
+      rescue ArgumentError
+        nil
+      end
+
+    cache_time && last_modified && last_modified <= cache_time
+  end
+
+  def set_cache_control_headers
+    if Rails.env.development?
+      response.headers['Last-Modified'] = Time.zone.now.httpdate
+      immutable_for(1.second)
+    else
+      response.headers['Last-Modified'] = last_modified.httpdate if last_modified
+      immutable_for(1.year)
+    end
+  end
+end
--- a/app/controllers/javascripts_controller.rb
+++ b/app/controllers/javascripts_controller.rb
@@ -22,6 +22,7 @@ class JavascriptsController < ApplicationController
 
     unless File.exist?(cache_file)
       content = query.pluck(:content).first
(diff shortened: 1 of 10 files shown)


app/controllers/javascripts_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  )
+
+  before_action :is_asset_path, :no_cookies, only: [:show]
+
+  def show
+    raise Discourse::NotFound unless last_modified.present?
+    return render body: nil, status: 304 if not_modified?
+
+    # Security: safe due to route constraint
+    cache_file = "#{DISK_CACHE_PATH}/#{params[:digest]}.js"
```
eviltrout: Since this is a potential security vector we should have a test. Otherwise someone could remove it from routes.rb accidentally and any file on the server could be served which is nasty.

config/routes.rb
```diff
@@ -447,6 +447,7 @@
 
   get "stylesheets/:name.css.map" => "stylesheets#show_source_map", constraints: { name: /[-a-z0-9_]+/ }
   get "stylesheets/:name.css" => "stylesheets#show", constraints: { name: /[-a-z0-9_]+/ }
+  get "javascripts/:digest.js" => "javascripts#show", constraints: { name: /[-a-z0-9_]+/ }
```
eviltrout: I'm surprised you don't have to escape values in the bracketed area. Usually I would do `\-` and `\_` but this seems to work?

db/migrate/20180927135248_create_javascript_caches.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJavascriptCaches < ActiveRecord::Migration[5.2]
+  def change
+    create_table :javascript_caches do |t|
+      t.references :theme_field, null: false
+      t.string :digest, null: true, index: true
+      t.text :content, null: false
+      t.timestamps
+    end
```
eviltrout: There are no indexes on this table. It looks like it's mainly queried by `digest` so it should have an index on that, probably a unique index?

config/routes.rb
xrav3nz: Ah thanks, will double check by adding a test, and also test for the security aspect.

db/migrate/20180927135248_create_javascript_caches.rb
xrav3nz: There is one index on `theme_field_id` (added by `t.references` helper), and another on `digest` (added by `index: true`).

`digest` is calculated entirely based on `content`, so it may not be unique (one common case: snippets copied from a Meta post). We could calculate it with `"#{id}_#{content}"` to make it "unique", do you think it's necessary?

config/routes.rb
```diff
@@ -447,6 +447,7 @@
 
   get "stylesheets/:name.css.map" => "stylesheets#show_source_map", constraints: { name: /[-a-z0-9_]+/ }
   get "stylesheets/:name.css" => "stylesheets#show", constraints: { name: /[-a-z0-9_]+/ }
+  get "javascripts/:digest.js" => "javascripts#show", constraints: { digest: /\h{40}/ }
```
xrav3nz: tightened this significantly by only allowing 40 hexdigit chars, [and added tests](https://github.com/discourse/discourse/pull/6447/files#diff-698a1e4385463ecf5c8eac5c8c65b9feR17).

eviltrout: Excellent thanks!

db/migrate/20180927135248_create_javascript_caches.rb
eviltrout: Oh my bad, I did not see the `index: true` - In fact I didn't know you could add one that way. It's fine if it's not unique, I was wrong about that.

app/controllers/javascripts_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    unless File.exist?(cache_file)
+      content = query.pluck(:content).first
```
SamSaffron: can't content be nil here, then you will try to write nil to a file which in theory will explode?

SamSaffron:
How big is the cache? how do we guarantee this table does not keep growing and growing?

app/controllers/javascripts_controller.rb
xrav3nz: Good point. In theory `content` at this point cannot be `nil` because:

1. we check to make sure the database entry exist at the start of the request
2. DB constraint to make sure `content` cannot be `nil` (it is at least an empty string)

But I agree to be safe I will add in another 404 check here.

xrav3nz:
Each theme field only has **one** javascript cache entry, and the entry should (will add a test to verify) be deleted when the corresponding theme field is deleted.

The size of the extracted `content` is not bounded by anything though, it could be as large as the theme field permits.

SamSaffron:
Thanks @xrav3nz  !!! Merging

PR #6448, opened 2018-10-03
Add missing fields to Upload Fabricator

ptrippett:
Now that https://github.com/discourse/discourse/commit/9ab1fb7dfc349293e433d4d3e78518d7d2ba1b06 saves the thumbnail_width and height the fabricator needs updating to prevent fix_dimensions attempting to fetch the image to populate the missing fields.

Test added for sanity, output prior to change to fabricator is below.

```
Failures:

  1) UploadSerializer should render without errors
     Failure/Error:
         def self.resize(width, height, opts = {})
           return if width.blank? || height.blank?

           max_width = (opts[:max_width] || SiteSetting.max_image_width).to_f
           max_height = (opts[:max_height] || SiteSetting.max_image_height).to_f

           w = width.to_f
           h = height.to_f

           return [w.floor, h.floor] if w <= max_width && h <= max_height

     ArgumentError:
       wrong number of arguments (given 0, expected 2..3)
```

### DIFF
--- a/spec/serializers/upload_serializer_spec.rb
+++ b/spec/serializers/upload_serializer_spec.rb
@@ -1,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe UploadSerializer do
+  let(:upload) { Fabricate(:upload) }
+  let(:subject) { UploadSerializer.new(upload, root: false) }
+
+  it 'should render without errors' do
+    json_data = JSON.load(subject.to_json)
+
+    expect(json_data['id']).to eql upload.id
+    expect(json_data['width']).to eql upload.width
+    expect(json_data['height']).to eql upload.height
+    expect(json_data['thumbnail_width']).to eql upload.thumbnail_width
+    expect(json_data['thumbnail_height']).to eql upload.thumbnail_height
+  end
+end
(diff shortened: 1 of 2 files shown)


spec/serializers/upload_serializer_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe UploadSerializer do
+  let(:upload) { Fabricate(:upload) }
+
+  it 'should render without errors' do
+    expect { UploadSerializer.new(upload).to_json }.not_to raise_error
```
tgxworld: `not_to raise_error` is an anti-pattern here because the test will just fail with the error if an error was raised. I will just test for some simple fields returned by the serializer here.

tgxworld: not_to raise_error is an anti-pattern here because the test will just fail with the error if an error was raised. I will just test for some simple fields returned by the serializer here.

ptrippett: Totally agree, i'll test that the result is valid instead.

im-ptrippett:
Test updated as per @tgxworld 's suggestion.

ptrippett:
Test updated as per @tgxworld 's suggestion.

ZogStriP:
Thanks @ptrippett  👍

PR #6450, opened 2018-10-04
FIX: No longer educate users who are editing content

hakusaro:
A user editing a post will no longer get composer messages that are
meant for new users posting replies and threads. These messages don't
make sense in an edit context at all -- they're usually discussing
making salient replies or topics, or adding avatars. They make even less
sense when a user is an admin attempting to change the default topics
for the first time.

Since these messages actually do make sense for a user when they have a
low post count, though, they're still going to occur. They just occur
when a user is creating new content (and thus, more likely to read the
notice), not during edits.

This is in response to this issue:
https://meta.discourse.org/t/education-message-for-editing-wiki-topic/66682

### DIFF
--- a/spec/components/composer_messages_finder_spec.rb
+++ b/spec/components/composer_messages_finder_spec.rb
@@ -479,7 +479,7 @@ describe ComposerMessagesFinder do
     it "returns nothing even if it normally would" do
       user.expects(:post_count).returns(9)
       expect(edit_post_finder.find).to be_blank
-      expect(new_post_finder.find).to be_present
+      expect(new_post_finder.find).to be_instance_of Hash
     end
   end
 


coding-horror:
Aw nice this one has bugged me for a long time 🍻 cheers

spec/components/composer_messages_finder_spec.rb
```diff
@@ -467,4 +467,20 @@
...
+    let(:user) { Fabricate.build(:user) }
+    let(:new_post_finder) { ComposerMessagesFinder.new(user, composer_action: 'reply') }
+    let(:edit_post_finder) { ComposerMessagesFinder.new(user, composer_action: 'edit') }
+
+    before do
+      SiteSetting.educate_until_posts = 10
+    end
+
+    it "returns nothing even if it normally would" do
+      user.expects(:post_count).returns(9)
```
tgxworld: We prefer to avoid stubs in our code base. What I think we can do here is to create a user with 2 posts and reduce the setting for `SiteSetting.educate_until_posts`

spec/components/composer_messages_finder_spec.rb
```diff
@@ -467,4 +467,20 @@
...
+      expect(edit_post_finder.find).to be_blank
+      expect(new_post_finder.find).to be_present
```
tgxworld: `be_present` is abit vague for me because the method can return anything and the test will still pass. It'll be better if we can test for the exact return value here.

lib/composer_messages_finder.rb
```diff
@@ -11,6 +11,7 @@ def self.check_methods
   end
 
   def find
+    return nil if editing_post?
```
tgxworld: `return if editing_post?` will be sufficient here :+1:

spec/components/composer_messages_finder_spec.rb
hakusaro: I can change this to:

```ruby
expect(new_post_finder.find).to be_instance_of Hash
```

Does that work? Sorry for all the questions here.

```ruby
  def find
    return if editing_post?
    self.class.check_methods.each do |m|
      msg = send(m)
      return msg if msg.present?
    end
    nil
  end
```

Find is just nebulous to me because it calls other methods and they all return different hashes of different forms -- so I'm not entirely sure how to accommodate this request other than making sure it returns a hash.

hakusaro:
@tgxworld I added a couple commits -- but I'm not sure they're the right ones. I'm open to better guidance on how this set of specs should work. I can always change the other specs in this file to not use stubs too?

PR #6451, opened 2018-10-04
Replace deprecated dropdown-button with DropdownSelectBoxComponent

pmusaraj:
Fixes dropdowns that relied on bootstrap classes (broken since [this commit](https://github.com/discourse/discourse/commit/dc8a16b43bf264b9001e3515aaba9dc9d5143228)) in groups members list and tags overview.


app/assets/javascripts/discourse/components/tags-admin-dropdown.js.es6
```diff
@@ -1,31 +1,32 @@
...
     return items;
   },
 
   actionNames: {
     manageGroups: "showTagGroups"
   },
 
-  clicked(id) {
+  mutateValue(id) {
     this.sendAction("actionNames." + id);
```
jjaffeux: minor, can we use interpolation here while we are at it ;)

jjaffeux:
LGTM, merge when you are available. I would want to avoid using the mutate function but it's totally fine for now. 👍

PR #6452, opened 2018-10-04
Let plugins decorate banner topic, fixes local dates in banners

pmusaraj:


SamSaffron:
Hmm @eviltrout can you review?

PR #6454, opened 2018-10-04
FEATURE: Mixed case tagging

davidtaylorhq:
- By default, behavior is not changed: tags are made lowercase upon creation and edit.

- If force_lowercase_tags is disabled, then mixed case tags are allowed.

- Tags must remain case-insensitively unique. This is enforced by ActiveRecord and Postgres.

- A migration is added to provide a `UNIQUE` index on `lower(name)`. Migration includes a safety to correct any current tags that do not meet the criteria.

- A scope is added to `models/tag.rb`, to allow easy case-insensitive lookups. This is used instead of `Tag.where(name: "blah")`.

- URLs remain lowercase. Mixed case URLs are functional, but have the lowercase equivalent as the canonical.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1,7 +1,12 @@
 class Tag < ActiveRecord::Base
   include Searchable
 
-  validates :name, presence: true, uniqueness: true
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
+
+  scope :with_name, ->(name) do
+    name = Array(name).map(&:downcase)
+    where("lower(name) IN (?)", name)
+  end
 
   has_many :tag_users # notification settings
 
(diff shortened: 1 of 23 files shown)


app/models/tag.rb
```diff
@@ -1,7 +1,12 @@
 class Tag < ActiveRecord::Base
   include Searchable
 
-  validates :name, presence: true, uniqueness: true
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
+
+  scope :with_name, ->(name) do
+    name = Array(name).map(&:downcase)
```
ZogStriP: Nitpick: since `name` is an array, shouldn't it be called `names`?

davidtaylorhq: It can also take a single string - it works the same way as 

```
tag_name = "hello"
tag_name_array = ["hello", "world"]
Tag.where(name: tag_name)
Tag.where(name: tag_name_array)
```

What do you think about `Tag.where_name`? That would probably be more clear.

PR #6456, opened 2018-10-05
FIX: Safari issue with some emojis

pmusaraj:


lib/pretty_text/shims.js
```diff
@@ -19,6 +19,9 @@ __setUnicode = function(replacements) {
         replacement = "\u200b" + replacement;
       }
       text = text.replace(m[0], replacement);
+      // fixes Safari VARIATION SELECTOR-16 issue with some emojis
+      // https://meta.discourse.org/t/emojis-selected-on-ios-displaying-additional-rectangles/86132
+      text = text.replace('\ufe0f', '');
```
ZogStriP: Do we want to run that in the loop? Can't it be done once before the loop?

pmusaraj: You're of course right, I moved it outside the loop.

tgxworld:
I think we should add a spec for this in https://github.com/discourse/discourse/blob/43512e9865087d10fb08a814e8f75f8b765aab6d/spec/components/pretty_text_spec.rb

pmusaraj:
Sure thing, done!

> I think we should add a spec for this in https://github.com/discourse/discourse/blob/43512e9865087d10fb08a814e8f75f8b765aab6d/spec/components/pretty_text_spec.rb

Added the spec.

tgxworld:
@pmusaraj Nice :+1: :heart_eyes_cat:

PR #6461, opened 2018-10-08
FIX: raise if staged user creation failed when processing incoming email

xrav3nz:
```text
Unrecognized error type (NoMethodError: undefined method `id' for nil:NilClass) when processing incoming email

Backtrace:
  /var/www/discourse/lib/email/receiver.rb:189:in `log_and_validate_user'
  /var/www/discourse/lib/email/receiver.rb:148:in `process_internal'
  /var/www/discourse/lib/email/receiver.rb:72:in `block in process!'
  ...
```

Specifically, `log_and_validate_user` in the below snippet expects an existing user or a newly created staged user.

https://github.com/discourse/discourse/blob/c0bb04d89d50b6d244f8f113b2c63ae1da6fe5e3/lib/email/receiver.rb#L146-L149

When something goes wrong, `find_or_create_user` swallows all exceptions and returns a `nil` user. Having a `find_or_create_user!` variation will help surface and log the actual problem.

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -144,7 +144,7 @@ module Email
       # Lets create a staged user if there isn't one yet. We will try to
       # delete staged users in process!() if something bad happens.
       if user.nil?
-        user = find_or_create_user(@from_email, @from_display_name)
+        user = find_or_create_user!(@from_email, @from_display_name)
         log_and_validate_user(user)
       end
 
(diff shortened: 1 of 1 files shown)


tgxworld:
@xrav3nz I think we should add a test for this :grin:

xrav3nz:
I only didn't add a test because I couldn't find a way to trigger that `create!` exception 😅 

- email validation: we already find by email first,
- username validation: `UsernameSuggester` would try to suggest an available username

Any other suggestions?

tgxworld:
@xrav3nz Ahh my bad I didn't see the use of `rescue` which basically swallows the error. Feel free to merge this :+1:

ZogStriP:
LGTM

PR #6464, opened 2018-10-08
FIX: Display logoff modal when session cookie expired

vinothkannans:


app/assets/javascripts/discourse/lib/ajax.js.es6
```diff
@@ -14,7 +14,14 @@ export function viewTrackingRequired() {
 }
 
 export function handleLogoff(xhr) {
-  if (xhr.getResponseHeader("Discourse-Logged-Out") && !_showingLogout) {
+  const sessionExpired =
+    xhr.getResponseHeader("X-Request-Id") &&
```
SamSaffron: what is this x request id , where is it coming from?

vinothkannans: This is default rails header. I used this to differentiate requests from client side. Anyway you ignore this PR for now. I'm working in better server side fix.

vinothkannans:
False alarm.

PR #6466, opened 2018-10-09
FIX: Generate Onebox for posts of type moderator_action.

nbianca:


lib/oneboxer.rb
```diff
@@ -197,7 +197,7 @@ def self.local_topic_html(url, route, opts)
       topic.posts.where(post_number: post_number).first :
       topic.ordered_posts.first
 
-    return if !post || post.hidden || post.post_type != Post.types[:regular]
+    return if !post || post.hidden || (post.post_type != Post.types[:regular] && post.post_type != Post.types[:moderator_action])
```
ZogStriP: Might be worth extracting to a constant

```ruby
ALLOWED_POST_TYPES ||= [Post.types[:regular], Post.types[:moderator_action])
```

and then

```ruby
return if !post || post.hidden || !ALLOWED_POST_TYPES.include?(post.post_type)
```

PR #6469, opened 2018-10-09
FEATURE: Higher Logic importer

pfaffman:
Here's a higher Logic importer that Hawk asked for.

There was a bunch of import-specific stuff that I removed from the code. I think that it should work, but I don't have a database to test against.


SamSaffron:
@gschlager  can you review / merge ?

gschlager:
Yes, I'll do it when I get some test data in the next few days.

ZogStriP:
We don't have data to test against either and it looks like there still are plenty of hardcoded info.

We'll revisit this if/when we get a migration from Higher Logic.

PR #6472, opened 2018-10-10
FIX: Do not allow revoking the token of current session.

nbianca:
Right now, if `log_out_strict` is enabled trying to revoke any token will result in revoking all the tokens.

This PR changes two things:
* it does not allow the current session to be terminated by any means;
* it does not take into account if `log_out_strict` is enabled anymore (because of the previous point).

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1109,10 +1109,9 @@ class UsersController < ApplicationController
     guardian.ensure_can_edit!(user)
 
     if params[:token_id]
-      cookie = guardian.request.cookies[Auth::DefaultCurrentUserProvider::TOKEN_COOKIE]
       token = UserAuthToken.find_by(id: params[:token_id], user_id: user.id)
       # The user should not be able to revoke the auth token of current session.
-      raise Discourse::NotFound if UserAuthToken.hash_token(cookie) == token.auth_token
+      raise Discourse::NotFound if guardian.auth_token == token.auth_token
       UserAuthToken.where(id: params[:token_id], user_id: user.id).each(&:destroy!)
     else
       UserAuthToken.where(user_id: user.id).each(&:destroy!)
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -1108,7 +1108,13 @@ def revoke_auth_token
     user = fetch_user_from_params
     guardian.ensure_can_edit!(user)
 
-    if !SiteSetting.log_out_strict && params[:token_id]
+    # The user should not be able to revoke the auth token of current session.
+    if params[:token_id]
```
ZogStriP: Why have the same condition twice?

app/serializers/user_auth_token_serializer.rb
```diff
@@ -9,9 +9,7 @@ def include_is_active?
   end
 
   def is_active
-    cookie = scope.request.cookies[Auth::DefaultCurrentUserProvider::TOKEN_COOKIE]
-
-    UserAuthToken.hash_token(cookie) == object.auth_token
+    scope.auth_token == object.auth_token
```
ZogStriP: Does that mean we can remove the `request` attribute from the scope object?

app/controllers/users_controller.rb
```diff
@@ -1108,7 +1108,10 @@ def revoke_auth_token
     user = fetch_user_from_params
     guardian.ensure_can_edit!(user)
 
-    if !SiteSetting.log_out_strict && params[:token_id]
+    if params[:token_id]
+      token = UserAuthToken.find_by(id: params[:token_id], user_id: user.id)
+      # The user should not be able to revoke the auth token of current session.
+      raise Discourse::NotFound if guardian.auth_token == token.auth_token
```
SamSaffron: NotFound is not really the right error message, I would go with a simple 400 here Discourse::InvalidParameters (will change it in a follow up commit)

PR #6474, opened 2018-10-10
FIX: multiple loadScript to the same url may resolve prematurely 

xrav3nz:
> BUG REPORT: https://meta.discourse.org/t/js-plugins-not-loading-correctly/99105?u=xrav3nz

This is how `loadScript(url)` currently deals with multiple concurrent requests:

1. Check existing `<script>` tags, and mark existing scripts (other than the input `url`) as loaded
2. Find "true" `url` of the requested resource (CDN, subfolder relation path, etc)
3. Check if we have loaded the resource with that "true" `url`, and resolve immediately if we have
4. Otherwise insert a `<script>` tag with the "true" `url` to load it

For example, in a subfolder install:

- Input `url` = `/javascripts/script.js`
- "True" `url` = `/subfolder/javascript/script.js`

And the _very_ subtle bug here is that we should use also use the true `url` for step (1), because:

- Since the input and true `url` are different, we mistakenly mark the true `url` as loaded in step one
- After finding the true `url`, and setting `loaded[trueUrl] = true` in (1) , we resolve the promise prematurely, when the resource could still be loading

I have tested the change locally and on a subfolder install 💪

### DIFF
--- a/app/assets/javascripts/discourse/lib/load-script.js.es6
+++ b/app/assets/javascripts/discourse/lib/load-script.js.es6
@@ -11,7 +11,6 @@ function loadWithTag(path, cb) {
   if (Ember.Test) {
     Ember.Test.registerWaiter(() => finished);
   }
-  head.appendChild(s);
 
   s.onload = s.onreadystatechange = function(_, abort) {
     finished = true;
@@ -27,6 +26,8 @@ function loadWithTag(path, cb) {
       }
     }
   };
+
+  head.appendChild(s);
 }
 
 export function loadCSS(url) {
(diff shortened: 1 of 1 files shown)


eviltrout:
Wouldn't this method cause it to double load if two loadScripts come in at the same time? Does the browser handle that case and only request it once? I have a doubt about that.

xrav3nz:
I believe the following part would catch that case? If we are already loading it, we wouldn't insert another script tag.

https://github.com/discourse/discourse/blob/9517df5dd8fa87118e4e9d42c40416053d56d2e1/app/assets/javascripts/discourse/lib/load-script.js.es6#L59-L61

eviltrout:
@xrav3nz right, well if it's working in testing I am cool with it!

xrav3nz:
Awesome, thanks. Will merge this in during my day tomorrow so I can watch the deploy/effect!

SamSaffron:
cool, lets get this merged asap, let me know when it is ready and we can deploy an subfolder site to confirm all is good

PR #6475, opened 2018-10-10
Exit with an error if there are errors with settings import

crispygoth:
If the site_settings import has any errors or settings that are not found, this
will cause the task to exit with a non-zero exit code.

This is useful when using this task as part of automated configuration deployment,
where you may not want to continue with the process if a setting fails to
import.

My specific use-case is that I'm using Puppet to bootstrap and deploy Discourse, and using  site_settings:import to populate some of the settings in the DB. I want my Puppet runs to fail if there's an issue with the settings; at the moment I have to use a hack where I check the output of the rake task for the error/not found counts.


ZogStriP:
Thanks 👍

PR #6476, opened 2018-10-10
FEATURE: automatically hide non-TL4 posts when flagged by a TL4 user

vinothkannans:
`advanced_user_narrative.rb` file change is required because discobot/TL4 user's `:off_topic` flag will auto hide the post.

### DIFF
--- a/spec/models/post_action_spec.rb
+++ b/spec/models/post_action_spec.rb
@@ -501,6 +501,37 @@ describe PostAction do
       expect(post.hidden_reason_id).to eq(Post.hidden_reasons[:flagged_by_tl3_user])
     end
 
+    it "hide non-tl4 posts that are flagged by a tl4 user" do
+      SiteSetting.site_contact_username = admin.username
+
+      post_action_type = PostActionType.types[:spam]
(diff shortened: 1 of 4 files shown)


spec/models/post_action_spec.rb
```diff
@@ -501,6 +501,37 @@ def value_for(user_id, dt)
       expect(post.hidden_reason_id).to eq(Post.hidden_reasons[:flagged_by_tl3_user])
     end
 
+    it "hide non-tl4 posts that are flagged by a tl4 user" do
+      Discourse.stubs(:site_contact_user).returns(admin)
```
SamSaffron: Note... I much prefer to do this via site settings, can you do a follow up commit

spec/models/post_action_spec.rb
```diff
@@ -501,6 +501,37 @@ def value_for(user_id, dt)
...
+
+      post_action_type = PostActionType.types[[:off_topic, :inappropriate, :spam].sample]
```
SamSaffron: definitely never add random to tests, just pick one of the 3, it is a nightmare to debug if this fails 33% of the time

PR #6477, opened 2018-10-10
FIX: Fix open quote links in new window.

nbianca:


ZogStriP:
I know this is hard, but we definitely need some tests here. This has regressed a lot in the past.

ZogStriP:
👍

PR #6478, opened 2018-10-10
FIX: force enable a user's email_private_messages option when user replies via email

pmusaraj:
Builds on: https://github.com/discourse/discourse/pull/6437

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -577,6 +577,7 @@ module Email
 
       case destination[:type]
       when :group
+        enable_email_PM_setting(user)
         group = destination[:obj]
         create_group_post(group, user, body, elided, hidden_reason_id)
 
@@ -601,6 +602,8 @@ module Email
           raise ReplyUserNotMatchingError, "post_reply_key.user_id => #{post_reply_key.user_id.inspect}, user.id => #{user.id.inspect}"
         end
 
+        enable_email_PM_setting(user)
+
         post = Post.with_deleted.find(post_reply_key.post_id)
 
         create_reply(user: user,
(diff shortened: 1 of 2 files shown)


lib/email/receiver.rb
```diff
@@ -1064,6 +1062,13 @@ def delete_staged_users
         end
       end
     end
+
+    def enable_email_PM_setting(user)
```
ZogStriP: Ruby use snake case for methods so that would be `def enable_email_pm_setting`

ZogStriP:
This also needs a test ;)

lib/email/receiver.rb
```diff
@@ -577,6 +577,7 @@ def process_destination(destination, user, body, elided, hidden_reason_id)
 
       case destination[:type]
       when :group
+        enable_email_pm_setting(user)
```
gschlager: `process_destination` won't be called when the `find_related_post_with_key` site setting is `false`.

lib/email/receiver.rb
```diff
@@ -601,6 +602,8 @@ def process_destination(destination, user, body, elided, hidden_reason_id)
           raise ReplyUserNotMatchingError, "post_reply_key.user_id => #{post_reply_key.user_id.inspect}, user.id => #{user.id.inspect}"
         end
 
+        enable_email_pm_setting(user)
```
gschlager: I think this will change the setting even when the user replies to a public topic instead of a PM. Probably not a big deal, but worth fixing if it's easy to fix.

SamSaffron:
Thanks! yes this looks good to me!

PR #6479, opened 2018-10-10
FIX: Do not add lightbox to onebox images

vinothkannans:


coding-horror:
@SamSaffron I think we need this fix deployed everywhere today

SamSaffron:
merging, one thing I want to test though is a direct to flickr ... merging now

PR #6480, opened 2018-10-10
UX: Add link to Groups in admin dashboard.

nbianca:
The placement of the "Groups" link under "Users" is sometimes confusing.

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin.js.es6
@@ -4,6 +4,11 @@ export default Ember.Controller.extend({
   application: Ember.inject.controller(),
 
   @computed
+  showGroups() {
+    return !this.siteSettings.enable_group_directory;
+  },
+
+  @computed
   showBadges() {
     return this.currentUser.get("admin") && this.siteSettings.enable_badges;
   },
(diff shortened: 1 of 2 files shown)


SamSaffron:
I know people struggle to find groups but I worry about adding a top level tab that takes you to an entirely different place

tgxworld:
Hmm where was this feature discussed? We removed the groups links from the admin page intentionally.

ZogStriP:
@eviltrout care to elaborate? ;)

eviltrout:
So here is the problem: if `enable_group_directory` is disabled, we hide the link to Groups from the navigation dropdown. (A big gaming company that uses Discourse does this.)

At that point, there is no obvious way to edit groups. We did find a very hidden link if you go Admin > Users > Groups but that is super confusing.

I asked @nbianca to put the link back. We could compromise and only put the link back if `enable_group_directory` is disabled?

nbianca:
I made the link show up only if `enable_group_dictionary` is disabled.

The other commit is to make the UI a little nicer. The admin will see the effect of him changing the site settings immediately. For example, immediately after the badges are disabled, the pill is hidden (a page refresh was required before).

SamSaffron:
I think this is a fine compromise for now!

PR #6482, opened 2018-10-11
DEV: Use DiscourseIpInfo for all IP queries.

nbianca:


config/locales/client.en.yml
```diff
@@ -597,7 +597,7 @@ en:
       topics_entered: "topics entered"
       post_count: "# posts"
       confirm_delete_other_accounts: "Are you sure you want to delete these accounts?"
-      powered_by: "powered by <a href='https://ipinfo.io'>ipinfo.io</a>"
+      powered_by: "powered by <a href='https://maxmind.com'>maxmind.com</a>"
```
SamSaffron: I don't think we have any attribution requirements here, we can remove this imo. We do not want people to think we are uploading IPs to maxmind

lib/discourse_ip_info.rb
```diff
@@ -16,27 +22,45 @@ def initialize
...
+    end
 
-    {
-      country: result.country.name,
-      country_code: result.country.iso_code,
-      region: result.subdivisions.most_specific.name,
-      city: result.city.name,
-    }
+    begin
+      result = Resolv::DNS.new.getname(ip)
```
SamSaffron: Is this code ran in test, in general this is one of the super rare times I would be ok with mocking

SamSaffron:
do we need any more tests here?

config/locales/client.en.yml
nbianca: According to License section of [this document](https://dev.maxmind.com/geoip/legacy/geolite/), I believe we do.

ZogStriP: That only says we need to say so in ads and documentations ;)

PR #6484, opened 2018-10-11
FEATURE: Upload tags from CSV

davidtaylorhq:
Adds an option to the `/tags` menu:

screenshot 2018-10-11 at 15 18 16

Supports one tag per line, and can optionally add the tag to a tag group. Import is 'hijacked' to prevent performance issues, and runs in a transaction. Restricted access in the same way as `/tag_groups` (admins & moderators).

### DIFF
--- a/app/assets/javascripts/admin/templates/components/tags-uploader.hbs
+++ b/app/assets/javascripts/admin/templates/components/tags-uploader.hbs
@@ -1,0 +1,6 @@
+  <label class="btn {{if addDisabled 'disabled'}}">
+    {{d-icon "upload"}}
+    {{i18n 'admin.watched_words.form.upload'}}
+    <input disabled={{addDisabled}} type="file" accept="text/plain,text/csv" style="visibility: hidden; position: absolute;" />
+  </label>
+  <span class="instructions">{{i18n 'tagging.upload_instructions'}}</span>
--- a/app/assets/javascripts/discourse/templates/modal/tag-upload.hbs
+++ b/app/assets/javascripts/discourse/templates/modal/tag-upload.hbs
@@ -1,0 +1,3 @@
+{{#d-modal-body title='tagging.upload'}}
+    {{tags-uploader closeModal=(action "closeModal") refresh=(route-action "refresh")}}
+{{/d-modal-body}}
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -117,6 +117,35 @@ class TagsController < ::ApplicationController
     end
   end
 
+  def upload
+    guardian.ensure_can_admin_tags!
+
+    file = params[:file] || params[:files].first
+
+    hijack do
+      begin
+        Tag.transaction do
+          CSV.foreach(file.tempfile) do |row|
+            raise Discourse::InvalidParameters.new(I18n.t("tags.upload_row_too_long")) if row.length > 2
+
+            tag_name = DiscourseTagging.clean_tag(row[0])
+            tag_group_name = row[1] || nil
+
+            tag = Tag.find_by_name(tag_name) || Tag.create!(name: tag_name)
+
+            if tag_group_name
+              tag_group = TagGroup.find_by(name: tag_group_name) || TagGroup.create!(name: tag_group_name)
+              tag.tag_groups << tag_group unless tag.tag_groups.include?(tag_group)
+            end
+          end
+        end
+        render json: success_json
+      rescue Discourse::InvalidParameters => e
+        render json: failed_json.merge(errors: [e.message]), status: 422
+      end
+    end
+  end
+
   def destroy
     guardian.ensure_can_admin_tags!
     tag_name = params[:tag_id]
--- a/spec/requests/tags_controller_spec.rb
+++ b/spec/requests/tags_controller_spec.rb
@@ -369,4 +369,50 @@ describe TagsController do
       end
     end
   end
+
+  context '#upload_csv' do
+    it 'requires you to be logged in' do
+      post "/tags/upload.json"
+      expect(response.status).to eq(403)
+    end
+
+    context 'while logged in' do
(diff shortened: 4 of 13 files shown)


app/assets/javascripts/admin/templates/components/tags-uploader.hbs
```diff
@@ -0,0 +1,7 @@
+
```
tgxworld: minor but there is an extra line here

app/assets/javascripts/admin/templates/components/tags-uploader.hbs
```diff
@@ -0,0 +1,7 @@
+  
+  <label class="btn {{if addDisabled 'disabled'}}">
+    {{d-icon "upload"}}
+    {{i18n 'admin.watched_words.form.upload'}}
+    <input disabled={{addDisabled}} type="file" accept="text/plain,text/csv" style="visibility: hidden; position: absolute;" />
```
tgxworld: We should avoid inline styles and move the styling here into a stylesheet.

app/assets/javascripts/discourse/templates/modal/tag-upload.hbs
```diff
@@ -0,0 +1,3 @@
+{{#d-modal-body title='tagging.upload'}}
+    {{tags-uploader closeModal=(action "closeModal") refresh=(route-action "refresh")}}
+{{/d-modal-body}}
```
tgxworld: minor but missing newline here.

app/controllers/tags_controller.rb
```diff
@@ -117,6 +117,35 @@ def update
...
+
+            if tag_group_name
+              tag_group = TagGroup.find_by(name: tag_group_name) || TagGroup.create!(name: tag_group_name)
+              tag.tag_groups << tag_group unless tag.tag_groups.include?(tag_group)
+            end
+          end
+        end
+          render json: success_json
+        rescue => e
+          render json: failed_json.merge(errors: [e.message]), status: 422
```
tgxworld: I think we're missing an `end ` here? Not sure since the indentation is off

spec/requests/tags_controller_spec.rb
```diff
@@ -369,4 +369,49 @@
...
+        expect(Tag.pluck(:name)).to contain_exactly("tag1", "capitaltag2", "spaced-tag", "tag3", "tag4")
+        expect(Tag.find_by_name("tag3").tag_groups.pluck(:name)).to contain_exactly("taggroup1")
+        expect(Tag.find_by_name("tag4").tag_groups.pluck(:name)).to contain_exactly("taggroup1")
+      end
+
+      it "fails gracefully with invalid input" do
+        sign_in(Fabricate(:moderator))
+        post "/tags/upload.json", params: { file: invalid_file, name: filename }
+        expect(response.status).to eq(422)
+        expect(Tag.pluck(:name)).to be_empty
```
tgxworld: I would do a `to not_change { Tag.count }` instead since it should in theory be faster than having to pluck a column.

app/controllers/tags_controller.rb
```diff
@@ -117,6 +117,35 @@ def update
...
+          CSV.foreach(file.tempfile) do |row|
+            raise Discourse::InvalidParameters.new(I18n.t("tags.upload_row_too_long")) if row.length > 2
+
+            tag_name = DiscourseTagging.clean_tag(row[0])
+            tag_group_name = row[1] || nil
+
+            tag = Tag.find_by_name(tag_name) || Tag.create!(name: tag_name)
+
+            if tag_group_name
+              tag_group = TagGroup.find_by(name: tag_group_name) || TagGroup.create!(name: tag_group_name)
```
tgxworld: I think we can use `find_or_create_by_safe!` here

app/assets/javascripts/admin/templates/components/tags-uploader.hbs
davidtaylorhq: I was following the conventions used [elsewhere](https://github.com/discourse/discourse/search?q=style%3D%22visibility%3A+hidden%3B+position%3A+absolute%3B%22&unscoped_q=style%3D%22visibility%3A+hidden%3B+position%3A+absolute%3B%22). But I agree, and will create a separate PR to clean them all up.

app/controllers/tags_controller.rb
davidtaylorhq: Indentation was off, thanks

app/controllers/tags_controller.rb
davidtaylorhq: For tags, we can't because we need to use the case insensitive `find_by_name` function. For tag groups, we can, but I think it would make the code a little more confusing to read since it will be different to tags. What do you think @tgxworld?

app/assets/javascripts/admin/templates/components/tags-uploader.hbs
davidtaylorhq: PR at #6487

app/controllers/tags_controller.rb
tgxworld: I guess this is fine too since `find_or_create_by` is really just a helper method that does the same thing.

app/controllers/tags_controller.rb
```diff
@@ -117,6 +117,35 @@ def update
...
+              tag.tag_groups << tag_group unless tag.tag_groups.include?(tag_group)
+            end
+          end
+        end
+        render json: success_json
+      rescue => e
```
tgxworld: Do we need which errors we want to rescue here? If so we should only rescue those errors and let the rest blow up. Otherwise, `rescue` will rescue things like typos as well.

davidtaylorhq: Thanks, I have explicitly rescued InvalidParameters (the generic `rescue_from` in `application_controller` doesn't apply inside a hijack block)

PR #6487, opened 2018-10-12
DEV: Remove inline styles from upload buttons

davidtaylorhq:
@tgxworld [pointed out](https://github.com/discourse/discourse/pull/6484#discussion_r224695970) that we use inline styles for file inputs, and that it would be better to move into CSS files. @awesomerobot would you mind having a look at this to check it doesn't interfere with your button-cleanup work?

### DIFF
--- a/app/assets/javascripts/admin/templates/components/watched-word-uploader.hbs
+++ b/app/assets/javascripts/admin/templates/components/watched-word-uploader.hbs
@@ -1,7 +1,7 @@
 <label class="btn {{if addDisabled 'disabled'}}">
   {{d-icon "upload"}}
   {{i18n 'admin.watched_words.form.upload'}}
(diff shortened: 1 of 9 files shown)


gschlager:
👍 I did something similar in https://github.com/discourse/discourse/pull/6345/commits/c2d8f5009a2d34085c64f9abd6f1ccd3338c174b, but it got temporarily reverted... 😉

davidtaylorhq:
Oh 😂

Why did that get reverted? Did something break?

gschlager:
It was part of my large "backup locations" PR that was squashed and reverted as a whole.

SamSaffron:
do we need a specific class here? I am not against this atomic approach @awesomerobot's call.

PR #6488, opened 2018-10-12
FIX: keep emoji syntax for custom emojis in quotes

majakomel:
fixes https://meta.discourse.org/t/custom-emoji-are-shown-in-wrong-size-when-quoted/98572


tgxworld:
Nice work! Thank you @majakomel

PR #6490, opened 2018-10-15
Add Cache-Control header to CORS

pdavide:
When trying to use Discourse API with caching clients, I get `Request header field Cache-Control is not allowed by Access-Control-Allow-Headers` error in the browser console.
This PR updates the `Access-Control-Allow-Headers` array to allow the use of the `Cache-Control` header.

This follows #6107 and #6322.


SamSaffron:
This seems safe to me, Thanks!

PR #6491, opened 2018-10-15
Typo in Thai language admin page

chaintng:
Typo in Thai language admin page

image


gschlager:

PR #6492, opened 2018-10-15
FIX: Do not award badges for links in restricted categories.

nbianca:


spec/models/badge_spec.rb
```diff
@@ -97,4 +97,34 @@
       expect(Badge.display_name('Not In Translations')).to eq('Not In Translations')
     end
   end
+
+  context "PopularLink badge" do
+    it "is awarded for 50 links", focus: true do
+      post = create_post(raw: "https://www.discourse.org/")
+
+      49.times do
```
ZogStriP: I would prefer we don't create that much records in the tests. Better to lower the badge query expectations.

SamSaffron:
Nice fix, merging in @nbianca  have a quick look at my follow up commit

PR #6495, opened 2018-10-15
rename route `/javascripts` to `/theme-javascripts`

xrav3nz:
RE: https://github.com/discourse/discourse/pull/6447

I tried rebaking a small theme component to test the javascript extraction, but the resulting extracted javascript 404'ed `/javascripts/263e67c3f1569db61de1d05af93cc7efc73043d5.js`.

image

Looks like Nginx handled the request directly, so it didn't even get to the Rails controller.

This PR renames the route to `/theme-javascripts/...`


xrav3nz:
Thanks! I just deployed and tested the change on a test instance (http://206.81.12.90/test/)

PR #6496, opened 2018-10-15
Fix/Permalinks Admin - Mobile

gutofoletto:
screen shot 2018-10-15 at 15 25 37

screen shot 2018-10-15 at 15 25 49


ZogStriP:
Thanks @gutofoletto. Seems like you need to run prettier ;)

gutofoletto:
@ZogStriP this is strange because I have prettier installed on my vscode. 🧐
it would be good if the project had a `.prettierrc` file that the editor could rely upon.
I also ran prettier with CLI and it changed only one space...hope CI passes now. 😄

ZogStriP:
> it would be good if the project had a `.prettierrc` file that the editor could rely on.

I think that's because we use the defaults.

PR #6497, opened 2018-10-15
Terms of Service v1.0.0

kemitchell:
This PR replaces the current template terms of service with version 1.0.0 of our new terms.

The new terms of service include a few template variables that I couldn't map to current template vars in the YAML file:

1. e-mail address for contacting the company behind the forum

2. governing law, which is "California Law" for CDCK

3. forum for legal disputes, which is "San Francisco, California" for CDCK

Is there an existing template variable for administrator e-mail address?

For 2 and 3, what's the best way of calling out that new forum admins should fill in for their companies?

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1847,9 +1847,9 @@ en:
     remove_muted_tags_from_latest: "Don't show topics tagged with muted tags in the latest topic list."
     force_lowercase_tags: "Force all new tags to be entirely lowercase."
 
-    company_short_name: "Company Name (short)"
-    company_full_name: "Company Name (full)"
-    company_domain: "Company Domain"
+    company_name: "Company Name"
+    governing_law: "Governing Law"
+    city_for_disputes: "City for Disputes"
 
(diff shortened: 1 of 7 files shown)


kemitchell:
Please note that the data file I patched comes with a header note:

```yaml
# This is a "source" file, which is used by Transifex to get translations for other languages.
# After this file is changed, it needs to be pushed by a maintainer to Transifex:
#
#   tx push -s
#
# Read more here: https://meta.discourse.org/t/contribute-a-translation-to-discourse/14882
```

kemitchell:
Looks like we may or may not need conforming changes in `test/javascripts/fixtures/static_fixtures.js.es6`

SamSaffron:
hmmm @gschlager  / @kemitchell  anything else needed here, would love to knock this off the list

SamSaffron:
We should in theory be ok without changing the fixture.. .cla bot can be like this it does notify after you first sign the CLA so we have to check it by hand, going to try to resolve conflicts here and merge

gschlager:
@SamSaffron I wanted to work on this on my Monday. We need to add new inputs in the wizard for

> 2. governing law, which is "California Law" for CDCK
> 
> 3. forum for legal disputes, which is "San Francisco, California" for CDCK

and also replace the contact email address.

PR #6498, opened 2018-10-16
UX: Warn users if the post that's currently edited has changed.

nbianca:


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -762,16 +768,21 @@ const Composer = RestModel.extend({
 
     const props = {
       raw: this.get("reply"),
+      raw_old: !post.get("wiki") || this.get("editConflict") ? null : this.get("originalText"),
```
ZogStriP: Any reasons we're not **always** sending the `originalText`? Conflicts are not limited to wiki posts.

```suggestion
      raw_old: this.get("originalText"),
```

app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -762,16 +768,21 @@ const Composer = RestModel.extend({
...
     };
 
     this.set("composeState", SAVING);
 
-    let rollback = throwAjaxError(() => {
+    let rollback = throwAjaxError(error => {
       post.set("cooked", oldCooked);
       this.set("composeState", OPEN);
+
+      if (error.jqXHR && error.jqXHR.status === 409) {
```
ZogStriP: 👍

app/controllers/posts_controller.rb
```diff
@@ -210,6 +210,13 @@ def update
       edit_reason: params[:post][:edit_reason]
     }
 
+    # Check if there is no edit conflict caused by two users saving different
+    # versions of the post.
+    raw_old = params[:post][:raw_old]
+    if !raw_old.empty? && raw_old != post.raw
```
ZogStriP: The "double negation" feels weird.

```suggestion
    if raw_old.present? && raw_old != post.raw
```

app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -387,9 +387,14 @@ const Composer = RestModel.extend({
     return SAVE_ICONS[action];
   },
 
-  @computed("action", "whisper")
-  saveLabel(action, whisper) {
-    return whisper ? "composer.create_whisper" : SAVE_LABELS[action];
+  @computed("action", "whisper", "editConflict")
+  saveLabel(action, whisper, editConflict) {
```
ZogStriP: There should be a test for that ;)

app/controllers/posts_controller.rb
```diff
@@ -210,6 +210,13 @@ def update
...
+      return render json: { errors: [I18n.t('edit_conflict')] }, status: 409
```
ZogStriP: There should be a spec for that.

config/locales/server.en.yml
```diff
@@ -304,6 +304,7 @@ en:
     badge: "%{display_name} badge on %{site_title}"
 
   too_late_to_edit: "That post was created too long ago. It can no longer be edited or deleted."
+  edit_conflict: "That post was edited by another user and your changes can no longer be edited."
```
ZogStriP: Maybe this instead?

```suggestion
  edit_conflict: "That post was edited by another user and your changes can no longer be saved."
```

app/assets/javascripts/discourse/models/composer.js.es6
nbianca: The absence of `raw_old` will skip the check on server side.

First time `raw_old` will be sent so the server can check if there is a problem. The second time (after a `editConflict` happened) `raw_old` will not be sent anymore which will bypass the check and overwrite the content.

However, I do agree that it could work for all posts, not just wiki. The spec mentioned only these and I meant to ask you about it. :)

app/controllers/posts_controller.rb
```diff
@@ -210,6 +210,13 @@ def update
       edit_reason: params[:post][:edit_reason]
     }
 
+    # Check if there is no edit conflict caused by two users saving different
+    # versions of the post.
```
tgxworld: The code itself explains the behavior quite clearly so I'm not sure if we really need this comment here.

app/controllers/posts_controller.rb
tgxworld: We have a pattern for rendering json errors, you can have a look at `render_json_error`

spec/requests/posts_controller_spec.rb
```diff
@@ -295,6 +295,14 @@
         expect(post.raw).to eq("edited body")
       end
 
+      it 'checks for an edit conflict' do
+        update_params[:post][:raw_old] = 'old body'
+        post.raw = 'newer body than raw_old'
```
tgxworld: Hmm this is never saved so the raw that the controller is seeing is different.

test/javascripts/acceptance/composer-edit-conflict-test.js.es6
```diff
@@ -0,0 +1,23 @@
...
+  loggedIn: true,
+
+  pretend(server, helper) {
+    server.put("/posts/398", () => {
+      return helper.response(409, { errors: ["edit conflict"] });
+    });
+  }
+});
+
+QUnit.test("Edit a post that causes an edit conflict", async assert => {
```
tgxworld: :+1:

PR #6502, opened 2018-10-17
FIX: extracted theme JavaScripts for multisite

xrav3nz:
NOTE: this includes a onceoff job to rebake all HTML theme fields, which will

1. extract inline javascripts in themes, and
2. correct the wrong `theme-javascript` url for multisite

### DIFF
--- a/app/jobs/onceoff/rebake_all_html_theme_fields.rb
+++ b/app/jobs/onceoff/rebake_all_html_theme_fields.rb
@@ -1,0 +1,11 @@
+module Jobs
+  class RebakeAllHtmlThemeFields < Jobs::Onceoff
+    def execute_onceoff(args)
+      ThemeField.where(type_id: ThemeField.types[:html]).find_each do |theme_field|
+        theme_field.update(value_baked: nil)
+      end
+
+      Theme.clear_cache!
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/jobs/onceoff/rebake_all_html_theme_fields.rb
```diff
@@ -0,0 +1,11 @@
+module Jobs
+  class RebakeAllHtmlThemeFields < Jobs::Onceoff
+    def execute_onceoff(args)
+      ThemeField.where(type_id: ThemeField.types[:html]).find_each do |theme_field|
+        theme_field.update(value_baked: nil)
```
xrav3nz: I intentionally picked `update` over `update!`, so that one failure will not throw off the entire job.

tgxworld:
@xrav3nz I think we should use [post deployment migration](https://meta.discourse.org/t/introducing-post-deployment-migration/98974) here instead as onceoff job is in the process of being faced out. Note that in migration we tend to avoid using ActiveRecord and prefer SQL directly. This is so because we don't want future changes in Ruby code to break the migration when a new site is being set up.

SamSaffron:
To me this feels enormously urgent, I am merging in, please do a follow up commit to address @tgxworld's concern regarding the once off.

xrav3nz:
Sorry, I totally mis-evaluated the urgency on this.

ActiveRecord is used because the theme baking process depends on an AR callback to extract and transpile the JavaScripts. I am happy to transform this into a regular (or new `maintenance`?) job, and have a post-deploy migration queue it. Is that okay?

PR #6504, opened 2018-10-17
FEATURE: [Experimental] Content Security Policy

xrav3nz:
This introduces four Site Settings:

- `content_security_policy`: Turn on Content-Security-Policy
- `content_security_policy_report_only`: Turn on Content-Security-Policy-Report-Only
- `content_security_policy_collect_reports`: Enable CSP violation report collection at `/csp_reports` (I added this so reporting can be temporarily disabled to avoid flooding the logs)
- `content_security_policy_script_src`: Additional whitelisted script sources. The current host and CDN are included by default.

Right now, this only adds the `script_src` CSP directive, which specifies the valid sources for JavaScripts.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,7 @@ require_dependency 'global_path'
 require_dependency 'secure_session'
 require_dependency 'topic_query'
 require_dependency 'hijack'
+require_dependency 'content_security_policy'
 
 class ApplicationController < ActionController::Base
   include CurrentUser
@@ -52,6 +53,7 @@ class ApplicationController < ActionController::Base
   before_action :block_if_requires_login
   before_action :preload_json
   before_action :check_xhr
+  before_action :set_content_security_policy_header
   after_action  :add_readonly_header
   after_action  :perform_refresh_session
   after_action  :dont_cache_page
@@ -98,6 +100,16 @@ class ApplicationController < ActionController::Base
     if !response.headers["Cache-Control"] && response.cache_control.blank?
       response.cache_control[:no_cache] = true
       response.cache_control[:extras] = ["no-store"]
+    end
+  end
+
+  def set_content_security_policy_header
+    if SiteSetting.content_security_policy
+      response.headers['Content-Security-Policy'] = content_security_policy
+    end
+
+    if SiteSetting.content_security_policy_report_only
+      response.headers['Content-Security-Policy-Report-Only'] = content_security_policy
     end
   end
 
@@ -500,6 +512,10 @@ class ApplicationController < ActionController::Base
   end
 
   private
+
+  def content_security_policy
+    @content_security_policy ||= ContentSecurityPolicy.new.build
+  end
 
   def check_readonly_mode
     @readonly_mode = Discourse.readonly_mode?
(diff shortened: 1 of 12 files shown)


app/controllers/application_controller.rb
```diff
@@ -501,6 +513,10 @@ def secure_session
 
   private
 
+  def content_security_policy
+    @content_security_policy ||= ContentSecurityPolicy.new.build
```
xrav3nz: Do we need to cache this so we don't have to calculate it for every request? Note the cache needs to be invalidated when the related site settings update

app/controllers/application_controller.rb
```diff
@@ -101,6 +103,16 @@ def dont_cache_page
...
 
+  def set_content_security_policy_header
+    if SiteSetting.content_security_policy
+      response.headers['Content-Security-Policy'] = content_security_policy
+    end
+
+    if SiteSetting.content_security_policy_report_only
+      response.headers['Content-Security-Policy-Report-Only'] = content_security_policy
+    end
+  end
```
xrav3nz: Should we skip this when the request is not a `GET HTML`?

config/site_settings.yml
```diff
@@ -1664,40 +1673,40 @@ uncategorized:
...
   default_email_mailing_list_mode_frequency:
-    enum: 'MailingListModeSiteSetting'
+    enum: "MailingListModeSiteSetting"
     default: 1
   disable_mailing_list_mode:
     default: false
     client: true
   default_email_always: false
   default_email_previous_replies:
-    enum: 'PreviousRepliesSiteSetting'
```
SamSaffron: yeah ... lets not do all of these now :)

SamSaffron:
As a security feature I wonder if we are setting it early enough, in particular omniauth is not going to get it now. 

Maybe just swap out: `use ActionDispatch::ContentSecurityPolicy::Middleware` and place our own implementation there ?

xrav3nz:
Argh how did I not see this 😂 will revert

lib/content_security_policy.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+require_dependency 'global_path'
+
+class ContentSecurityPolicy
+  include GlobalPath
+
+  class Middleware
+    WHITELISTED_PATHS = %w(
+      /logs
```
xrav3nz: Logster still has [quite some inline JS](https://github.com/discourse/logster/blob/0b4b42fe815b333f56bf7daa804b1aa2fb0642cc/lib/logster/middleware/viewer.rb#L204-L248).

SamSaffron: Ouch ... can you work with @OsamaSayegh on getting that removed asap?

lib/content_security_policy.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    def call(env)
+      request = Rack::Request.new(env)
+      _, headers, _ = response = @app.call(env)
+
+      return response unless html_response?(headers)
+      return response if whitelisted?(request.path)
+
+      policy = ContentSecurityPolicy.new.build
```
SamSaffron: why do any building if this is not enabled?

SamSaffron:
overall this looks fine to me sans my tiny new comment, once you address that first thing in your day can you merge this in and enable this on meta? that way you will have all of friday to watch it. and if you feel any kind of risk you can just turn it off on meta at the end of your day.

xrav3nz:
Sadly had to revert :( more info and new PR: https://github.com/discourse/discourse/pull/6514

PR #6505, opened 2018-10-18
Add a plugin entry point to add SEO metadata

notriddle:
Related to https://github.com/discourse/discourse-solved/pull/69

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -18,3 +18,4 @@
 <%= canonical_link_tag %>
 <%= render_sitelinks_search_tag %>
 <link rel="search" type="application/opensearchdescription+xml" href="<%= Discourse.base_url %>/opensearch.xml" title="<%= SiteSetting.title %> Search">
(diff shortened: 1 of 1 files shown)


SamSaffron:
We already have:

https://github.com/discourse/discourse/blob/f0af61da415d0bbb4a49cb0eb461bd7bbeb7ea12/app/views/layouts/application.html.erb#L55-L55 

Are you sure this is needed?

notriddle:
Yeah. `server:before-head-close` isn't included in https://github.com/discourse/discourse/blob/cc851af /app/views/layouts/crawler.html.erb

SamSaffron:
Hmm @eviltrout  what is your call here as you added the original extension, there are 2 options

1. Extend `server:before-head-close` so it hits crawler as well

2. Add the extra plugin html that is unconditional 

From what I can tell in all-the-plugins nobody is using server:before-head-close so this must be a customer specific hook.

SamSaffron:
@notriddle see: https://github.com/discourse/discourse/commit/abaa3f0650058bff083761d5318b640588b7d78d

PR #6506, opened 2018-10-18
FIX: Strip @ when searching for users and groups.

nbianca:


SamSaffron:
Yeah this is good! thanks

PR #6509, opened 2018-10-18
auto is not a valid value for min/max height

angusmcleod:
Discovered through a related style issue with the Custom Wizard Plugin.

https://meta.discourse.org/t/custom-wizard-plugin/73345/209?u=angus

### DIFF
--- a/app/assets/stylesheets/wizard.scss
+++ b/app/assets/stylesheets/wizard.scss
@@ -543,7 +543,7 @@ body.wizard {
     max-height: none;
   }
   .wizard-step-contents {
-    min-height: none !important;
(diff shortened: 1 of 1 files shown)


hellekin:
@angusmcleod according to https://developer.mozilla.org/en-US/docs/Web/CSS/min-height, `auto` is a valid value, but `none` is not -- it's only available to `max-height`. We're not ready for `fit-content` which would be optimal... I guess leaving `min-height: auto` is fine, it's the `max-height: none` that needs to be implemented.

SamSaffron:
@awesomerobot your call on what to do here.

PR #6510, opened 2018-10-19
Add base_url to config locales

dkess:
These links are broken for forums with a base URLs under a subdirectory. I used the following `sed` commands to fix them:

```
sed -i "s|href=\'/|href=\'\%\{base_url\}/|g" server.*
sed -i "s|href=\"/|href=\"\%\{base_url\}/|g" server.*
```

### DIFF
--- a/config/locales/server.ar.yml
+++ b/config/locales/server.ar.yml
@@ -626,8 +626,8 @@ ar:
       email_body: "%{link}\n\n%{message}"
     inappropriate:
       title: 'غير مناسب'
-      description: 'محتوي هذا المنشور أي شخص عاقل سيعتبره عدواني، أو مسيئ، أو انتهاك لـ<a href="/guidelines">مبادئ مجتمعنا التوجيهية</a>.'
-      short_description: 'انتهاك <a href="/guidelines">للقواعد العامة للمجتمع</a>'
+      description: 'محتوي هذا المنشور أي شخص عاقل سيعتبره عدواني، أو مسيئ، أو انتهاك لـ<a href="%{base_url}/guidelines">مبادئ مجتمعنا التوجيهية</a>.'
+      short_description: 'انتهاك <a href="%{base_url}/guidelines">للقواعد العامة للمجتمع</a>'
       long_form: 'ترفع علم هذا عن صورة غير ملائمة'
     notify_user:
       title: 'أرسل رسالة إلى @{{username}}'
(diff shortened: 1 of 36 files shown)


ZogStriP:
We only support changing the `en` locale. All other locales must be changed on [transifex](https://www.transifex.com/discourse/discourse-org/).

dkess:
I don't mind resubmitting this PR with only the `en` locale changed, but it seems silly that other languages aren't able to benefit from this. What do you recommend I do? Is there a way to automatically make this change in Transifex without having to manually go through an interface and change each string?

ZogStriP:
@dkess We could do that replace when we [pull the translations](https://github.com/discourse/discourse/blob/master/script/pull_translations.rb) and ensure the paths are all using `%base_url%` when we [check them](https://github.com/discourse/discourse/blob/master/lib/i18n/locale_file_checker.rb)?

gschlager:
We could push translations to Transifex, but I think @ZogStriP's idea of adding a missing `%base_url%` during pull is even better. That way even plugins would profit (not sure if we have any plugins with that issue though).

@dkess It would be fantastic if you could make the following changes:

* Create a PR for https://github.com/discourse/translations-manager that makes sure the `%base_url%` is there if needed

* Add a check in https://github.com/discourse/discourse/blob/master/lib/i18n/locale_file_checker.rb for missing `%base_url%`

* Add a spec in https://github.com/discourse/discourse/blob/master/spec/integrity/i18n_spec.rb for missing `%base_url%` in English locale files

* Update only the English locale file in this PR

dkess:
Thanks for the guidance! Sorry for being late with these modifications. This PR now only changes the  English locale file.

How do you recommend I make the `translations-manager` PR? I'm looking through that repo and it's not quite obvious where this change would go.

ZogStriP:
> Sorry for being late with these modifications.

No worries ;)



> How do you recommend I make the `translations-manager` PR? I'm looking through that repo and it's not quite obvious where this change would go.

I think we should add a new `prepend_base_url` step after the [`remove_empty_translations`](https://github.com/discourse/translations-manager/blob/fdcb427bb0f11139a7f2ace42f28bed3e182ef64/lib/translations_manager/transifex_updater.rb#L54) that would walk the tree and ensure `href` are properly using `%{base_url}`

gschlager:
@dkess I'm really sorry, I hope you haven't started working on it yet, but I've been thinking about this some more and I believe adding a step to `translations-manager` doesn't make much sense. The changes to the English locale file automatically invalidate the affected translations on Transifex for each language. Those translations need to be updated anyway and translators will be forced to add the `%{base_url}`.

So, basically doing the following steps would be enough:

* Add a check in https://github.com/discourse/discourse/blob/master/lib/i18n/locale_file_checker.rb for missing %base_url%

* Add a spec in https://github.com/discourse/discourse/blob/master/spec/integrity/i18n_spec.rb for missing %base_url% in English locale files

ZogStriP:
> The changes to the English locale file automatically invalidate the affected translations on Transifex for each language. Those translations need to be updated anyway and translators will be forced to add the `%{base_url}`.

Oh good point.

eviltrout:
I had to revert this. It broke the flag modal:

https://meta.discourse.org/t/flag-post-popup-needs-some/101142

dkess:
Interesting. Not sure why that would have happened though? I'm re-reading my changes and it's not clear to me how it would have caused it. Maybe they have problematic characters in their base_url?

PR #6514, opened 2018-10-19
 FEATURE: [Experimental] Content Security Policy

xrav3nz:
https://github.com/discourse/discourse/pull/6504 was merged and reverted because _some_ requests would fail (e.g. loading a post works but editing a post doesn't), thankfully the smoke tests caught this.

The culprit was `Mime::Type.register 'application/csp-report', :json`. It turns out this would overwrite and remove all other JSON types, and `application/json` was no longer recognized.

Instead of registering a new MIME type, https://github.com/discourse/discourse/commit/ad6227716a1c1033e4a7b63894b24170bbdcd8b1 parses the raw JSON body instead.

### DIFF
--- a/app/controllers/csp_reports_controller.rb
+++ b/app/controllers/csp_reports_controller.rb
@@ -14,7 +14,7 @@ class CspReportsController < ApplicationController
   private
 
   def report
-    @report ||= params.require('csp-report').permit(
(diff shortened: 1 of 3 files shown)


SamSaffron:
I support a repeat of the Friday experiment as soon you start your day.

PR #6518, opened 2018-10-22
FIX: Strip accents from search query

danielhollas:
4481836 introduced accent stipping in search_indexer,
but we need to strip it from the query itself as well.

One open questions: The search log still includes accents if the `strip_diacritics` function is called from `Search.prepare_data`. I am not sure that is the desired behaviour since there will be separate logs for queries that are identical from the search perspective.


SamSaffron:
This looks good to me, thanks!

PR #6520, opened 2018-10-23
FEATURE: Warn users via email about suspicious logins.

nbianca:


app/mailers/user_notifications.rb
```diff
@@ -31,6 +33,25 @@ def signup_after_approval(user, opts = {})
                 new_user_tips: tips)
   end
 
+  def suspicious_login(user, opts = {})
+    ipinfo = DiscourseIpInfo.get(opts[:client_ip])
+    location = [ipinfo[:city], ipinfo[:region], ipinfo[:country]].reject { |x| x.blank? }.join(", ")
```
ZogStriP: Slightly shorted notation

```suggestion
    location = [ipinfo[:city], ipinfo[:region], ipinfo[:country]].reject(&:blank?).join(", ")
```

app/models/user_auth_token.rb
```diff
@@ -30,6 +30,40 @@ def self.log(info)
...
+    DiscourseIpInfo.get(ip)[:country]
+  end
+
+  def self.is_suspicious(user_id, user_ip)
+    ips = UserAuthTokenLog.where(user_id: user_id).pluck('DISTINCT client_ip')
+    return false if ips.empty? # first login is never suspicious
+
+    user_location = login_location(user_ip)
+
+    ips.each do |ip|
```
ZogStriP: This can be a one-liner ;)

```suggestion
    ips.any? { |ip| user_location == login_location(ip) }
```

app/models/user_auth_token.rb
```diff
@@ -30,6 +30,40 @@ def self.log(info)
...
+  # exception that for users from huge countries it might not be specific
+  # enoguh. For US users where the real user and the malicious one could
+  # happen to live both in USA, this will not detect any suspicious activity.
+  def self.login_location(ip)
+    DiscourseIpInfo.get(ip)[:country]
+  end
+
+  def self.is_suspicious(user_id, user_ip)
+    ips = UserAuthTokenLog.where(user_id: user_id).pluck('DISTINCT client_ip')
+    return false if ips.empty? # first login is never suspicious
```
ZogStriP: Is this ever hit? Looks like `is_suspicious` is called **after** you've [created](https://github.com/discourse/discourse/blob/651d911a4cfa8beb1975b2e97ae239403ce5cbc8/app/models/user_auth_token.rb#L70-L77) the `UserAuthToken` for the current session

spec/models/user_auth_token_spec.rb
```diff
@@ -282,4 +283,36 @@
     expect(lookup.auth_token_seen).to eq(true)
   end
 
+  context "suspicious login" do
+
+    let(:user) { Fabricate(:user) }
+
+    before do
+      UserAuthToken.stubs(:login_location).with("1.1.1.1").returns("Location 1")
```
ZogStriP: Is there a way we can avoid the stubs?

app/models/user_auth_token.rb
nbianca: Yes, this checks the UserAuthToken**Log**s and is called before log is called.

There is a test for this one.

ZogStriP: My bad. I apparently can't read 🙈

app/models/user_auth_token.rb
nbianca: Is actually `.none?`

Thanks! To my shame, I did not know that `.any?` can receive a block.

ZogStriP: No shame in that ;)

ZogStriP:
Let's try it 👍

PR #6523, opened 2018-10-24
FIX: GlobalPath#upload_cdn_path when S3 bucket has a folder

xrav3nz:
`GlobalPath#upload_cdn_path` does not consider S3 bucket "folder" path, and results in a repeated folder path in the generated URL, e.g. `https://cdn-aws.com/subfolder/subfolder/original/1X/e57963.jpg`


SamSaffron:
Subfolder is the gift that does not stop giving ;) 

Thanks for the fix ❤️

PR #6524, opened 2018-10-24
Add Updated Vagrantfile and screen file

tareko:
The Vagrantfile was previously removed as it was considered abandoned. This made the use of a Vagrant setup for development much more difficult. A new Vagrantfile that points to a new vagrant box (available publicly via dropbox) is now included. It uses an Ubuntu 18.04 server and the most recent Discourse


SamSaffron:
I appreciate this but I do not want to take this on... why not use the docker based setup ... 

We certainly can not point people from our core repository to a third party maintained image.

PR #6525, opened 2018-10-24
FIX: Ensure the like button always has a title, for accessibility

mwcampbell:
The like button previously didn't have a title for anonymous users,
because the `canToggleLike` flag wasn't set, but the `liked` flag wasn't
set either. This made the button inaccessible to blind users.


ZogStriP:
Thanks 👍

PR #6528, opened 2018-10-24
CSP - extract all other inline JavaScripts

xrav3nz:
I tested all the changes locally, and will squash when merging.


xrav3nz:
👀 on the failing tests

app/views/user_api_keys/new.html.erb
```diff
@@ -28,20 +28,8 @@
...
-
-  // prevent double submission which would invalidate the nonce
-  document.getElementById('submit').addEventListener('click', function(e){
-    if (window.__submitted) {
-      e.preventDefault();
-    } else {
-      window.__submitted = true;
-    }
-  });
-</script>
```
xrav3nz: These are simply removed, because the intended behavior is already the default for [`submit_tag` `disable_with` helper](https://api.rubyonrails.org/v5.0/classes/ActionView/Helpers/FormTagHelper.html#method-i-submit_tag).

spec/views/omniauth_callbacks/complete.html.erb_spec.rb
```diff
@@ -6,7 +6,7 @@
 describe "users/omniauth_callbacks/complete.html.erb" do
 
   let :rendered_data do
-    JSON.parse(rendered.match(/var authResult = (.*);/)[1])
+    JSON.parse(rendered.match(/data-auth-result="([^"]*)"/)[1].gsub('&quot;', '"'))
```
SamSaffron: hmmm so this is a bug fix?

SamSaffron:
overall I am fine with this, please merge in as soon as you start your day and watch meta closely

spec/views/omniauth_callbacks/complete.html.erb_spec.rb
xrav3nz: No, simply changed to work with rendering `authResult` in HTML data attribute.

PR #6529, opened 2018-10-24
PERF: remove total unread notifications from message bus

featheredtoast:
User notifications queries and publishes total_unread_notifications on the message bus, but does not appear to use it.


SamSaffron:
This is good, this is no longer used anywhere and is expensive.

PR #6531, opened 2018-10-25
FEATURE: hide muted categories from /categories list

pmusaraj:


SamSaffron:
discussed internally, we will try it out, thanks!

PR #6532, opened 2018-10-26
FIX: ensures reports links are correct on subfolder installs

jjaffeux:


test/javascripts/models/report-test.js.es6
```diff
@@ -515,4 +515,22 @@ QUnit.test("computed labels", assert => {
     "<a href='/t/-/2/3'>This is the beginning of</a>"
   );
   assert.equal(computedPostLabel.value, "This is the beginning of");
+
+  // subfolder support
+  Discourse.BaseUri = "/forum";
+
+  const postLink = computedLabels[5].compute(row).formatedValue;
+  assert.equal(
```
ZogStriP: 😍

PR #6534, opened 2018-10-26
FIX: Do not check for suspicious login when impersonating.

nbianca:


app/models/user_auth_token.rb
```diff
@@ -82,7 +82,7 @@ def self.generate!(user_id: , user_agent: nil, client_ip: nil, path: nil, staff:
         path: path,
         auth_token: hashed_token)
 
-    if staff
+    if staff && path != '/admin/impersonate'
```
ZogStriP: Does that work as well on a subfolder install?

app/models/user_auth_token.rb
```diff
@@ -82,7 +82,7 @@ def self.generate!(user_id: , user_agent: nil, client_ip: nil, path: nil, staff:
         path: path,
         auth_token: hashed_token)
 
-    if staff
+    if staff && path !~ /\/impersonate/
```
ZogStriP: What if a user is called `impersonate`?

nbianca:
@davidtaylorhq I think so!

spec/models/user_auth_token_spec.rb
```diff
@@ -283,4 +283,26 @@
     expect(lookup.auth_token_seen).to eq(true)
   end
 
+  context "suspicious login" do
+
+    let(:user) { Fabricate(:user) }
+    let(:admin) { Fabricate(:admin) }
+
+    it "is not checked when generated for non-staff" do
+      Jobs.expects(:enqueue).with(:suspicious_login, anything).never
```
tgxworld: We generally want to avoid using mocks whenever possible. It will be better if we test for the actual job being en-queued. You can access the jobs by accessing `Jobs:SuspiciousLogin.jobs.size`. For testing Sidekiq jobs, https://github.com/mperham/sidekiq/wiki/Testing provides a very good overview :+1:

lib/current_user.rb
```diff
@@ -13,8 +13,8 @@ def clear_current_user
     @current_user_provider = Discourse.current_user_provider.new({})
   end
 
-  def log_on_user(user)
-    current_user_provider.log_on_user(user, session, cookies)
+  def log_on_user(user, opts)
```
davidtaylorhq: I think `opts` needs to be optional - that's what's causing all the specs to fail in travis.
```suggestion
  def log_on_user(user, opts={})
```

PR #6536, opened 2018-10-26
FIX: Do not leak information about post revisions.

nbianca:


spec/requests/posts_controller_spec.rb
```diff
@@ -1173,6 +1173,25 @@
...
+        post.save
+      }
+
+      it "throws an exception for users" do
+        sign_in(Fabricate(:user))
+        get "/posts/#{post.id}/revisions/#{post_revision.number}.json"
+        expect(response.status).to eq(404)
+      end
+
+      it "throws an exception for users" do
```
ZogStriP:
```suggestion
      it "works for admins" do
```

PR #6537, opened 2018-10-27
Correct Xcode capitalization and add Apple dev info

tekamoment:
Corrected Xcode capitalization from XCode to its proper spelling (as documented [here.](https://developer.apple.com/xcode/))

Also added a small note re: access to the Apple Developer Downloads page, noting that a free Apple Developer account is required.


ZogStriP:
Thanks

PR #6546, opened 2018-10-31
FIX: return 400 for missing required params

oblakeerickson:
If a required param is missing return a 400 and show a message
displaying which param was missing. Added this to the application
controller so that we don't have to add this logic to every controller
action.


SamSaffron:
Looks good to me!

PR #6547, opened 2018-10-31
FIX: Add presentation role to topic map headings, for accessibility

mwcampbell:
The `h4` elements in the topic map are not true headings; that tag is only
used to achieve a particular presentation. The use of a heading tag here
confuses screen reader users, who use headings for quick navigation. To
fix this without breaking any CSS, especially for third-party themes, we
add `role="presentation"` to these elements. That tells screen readers,
and other assistive technologies, to ignore the normal semantics of
these elements.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/topic-map.js.es6
+++ b/app/assets/javascripts/discourse/widgets/topic-map.js.es6
@@ -86,7 +86,7 @@ createWidget("topic-map-summary", {
     const contents = [];
     contents.push(
       h("li", [
-        h("h4", I18n.t("created_lowercase")),
+        h("h4", { attributes: { role: "presentation" } }, I18n.t("created_lowercase")),
         h("div.topic-map-post.created-at", [
           avatarFor("tiny", {
             username: attrs.createdByUsername,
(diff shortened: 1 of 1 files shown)


mwcampbell:
I fixed code formatting issues yesterday after they were pointed out by discoursedangerbot. Is there anything else I need to do before this can be reviewed?

mwcampbell:
Would someone on the team please explain why this PR was ignored? Do I need to mention it on the thread where I reported the bug? I didn't think that would be necessary, because my first PR was merged though I didn't mention it on meta. I don't want to be whiny; I just want to make sure I understand how the process works. I'm also concerned that Joffrey's fix for this problem might break third-party themes, since it requires CSS changes.

eviltrout:
Please be patient, 2 days does not mean a PR is ignored.

jjaffeux:
@mwcampbell Sorry about that, thanks for the PR. You might have noticed I have merged a commit today on another of your suggestions, where I dropped h4 elements.

You are right this might break themes. But I don’t want to carry styling on html elements and there was no class before, so yes it might break a little for some people.

mwcampbell:
If there's any chance this will break things, please take a look at my PR. It doesn't add a style attribute in HTML, just the role attribute. So it fixes accessibility while not affecting anything else. Again, I'm sorry for the confusion, as I didn't mention my PR on the bug topic.

jjaffeux:
Hmmm indeed that's a possibility, ok I will revert my commit and fix this later

jjaffeux:
Merged here: https://github.com/discourse/discourse/pull/6561

Thanks

PR #6549, opened 2018-10-31
FEATURE: Track how many user flags are agreed/disagreed/ignored

eviltrout:
* Display the percentages when reviewing flags (if the user has flagged at least 3 things.)

image

* Also fixes a bug where if you disagreed with a flag it would update flags that were previously agreed with.


eviltrout:
Thanks for the feedback @gschlager I've implemented the translation changes and fixed the spec.

PR #6551, opened 2018-11-01
UX: don't show crawler navigation in print view

hnb-ku:


SamSaffron:
weird that we are using embedded styles here ... we should probably change it

Thanks for the fix!

hnb-ku:
Sure @SamSaffron 

continued in #6581

PR #6553, opened 2018-11-01
FIX: only extract script tags with certain types

xrav3nz:
`script` tags with custom types (e.g. `text/template`) will not be executed by the browser, and should not be extracted into an external theme JavaScript.

### DIFF
--- a/app/models/theme_field.rb
+++ b/app/models/theme_field.rb
@@ -123,9 +123,10 @@ COMPILED
     end
 
     doc.css('script').each do |node|
-      next if node['src'].present?
+      next unless inline_javascript?(node)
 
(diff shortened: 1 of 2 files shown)


app/models/theme_field.rb
```diff
@@ -253,6 +254,24 @@ def target_name
...
     MessageBus.publish "/footer-change/#{theme.id}", self.value if theme && self.name == "footer"
   end
+
+  private
+
+  JAVASCRIPT_TYPES = %w(
+    text/javascript
+    application/javascript
+    application/ecmascript
+  )
```
xrav3nz: This is not an exhaustive list -- only the most common ones and current standard, according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type

PR #6555, opened 2018-11-02
FIX: uses more semantically correct spans in post map

jjaffeux:


mwcampbell:
Will this break any third-party themes? I did [my own PR](https://github.com/discourse/discourse/pull/6547) that fixes the problem without any CSS change, by adding `role="presentation"`. I'm sorry I didn't mention that on meta in the bug report topic.

PR #6557, opened 2018-11-02
FEATURE: Update Font Awesome to v5.4.1 and SVGs

pmusaraj:
This PR makes substantial changes to our Font Awesome (FA) icon set: 

- upgrades from FA 4.7 to FA 5.4.1
- uses SVG icons in a sprite instead of an icon font
- only loads a subset of FA's 1300+ icons for better performance 
- allows plugins, themes, admins to include additional icons in the subset 

**Adding icons to the subset**
Icons can be added to the default subset by: 

- plugins, via `register_svg_icon "user-plus" if respond_to?(:register_svg_icon)` (where `user-plus` is the icon name in FA)
- themes or theme components, via any setting containing `_icon` in its name
- admins, via the site setting `svg icon subset` or via the respective icon fields in group flair and badges

**Referencing changes**
FA 5+ splits icons across three styles: solid (default), regular and brands. To use an icon from the regular or brands bundles, FA 5 introduces new class prefixes: `"far fa-address-book"` or `"fab fa-facebook"`. To allow bundling the three styles into one SVG sprite, icons in the regular and brand styles are converted to `far-icon-name` and `fab-icon-name` references in Discourse. The examples above then can also be referenced as `"far-address-book"` and `"fab-facebook"` in Discourse. 

**Updating Font Awesome**
This dependency in Discourse is now managed by `yarn`. To update, modify the version number in `package.json`, run `yarn install` and use the rake task `svgicons:update` to copy the SVG sprites to their destination.


app/assets/javascripts/discourse-common/lib/icon-library.js.es6
```diff
@@ -68,58 +236,89 @@ export function registerIconRenderer(renderer) {
...
+}
+
 // default resolver is font awesome
 registerIconRenderer({
   name: "font-awesome",
 
-  string(icon, params) {
-    let tagName = params.tagName || "i";
-    let html = `<${tagName} class='${faClasses(icon, params)}'`;
-    if (params.title) {
```
pmusaraj: SVG elements can't have a title attribute. But does this mean we're regressing on accessibility here?

app/assets/javascripts/discourse/helpers/icon-or-image.js.es6
```diff
@@ -1,14 +1,15 @@
...
-import { iconHTML } from "discourse-common/lib/icon-library";
+import { iconHTML, convertIconClass } from "discourse-common/lib/icon-library";
 
 export default htmlHelper(function({ icon, image }) {
   if (!Ember.isEmpty(image)) {
     return `<img src='${image}'>`;
   }
 
-  if (Ember.isEmpty(icon) || icon.indexOf("fa-") !== 0) {
+  if (Ember.isEmpty(icon) || icon.indexOf("fa-") < 0) {
```
pmusaraj: This small change lets users use both FA 4.7 icons (fa-icon) and FA 5+ icons (far fa-icon).

app/assets/javascripts/discourse/initializers/svg-sprite-fontawesome.js.es6
```diff
@@ -0,0 +1,11 @@
+import svgSpriteLoader from "discourse/lib/svg-sprite-loader";
+
+export default {
+  name: "svg-sprite-fontawesome",
+
+  initialize() {
+    if (Discourse && Discourse.SvgSpritePath) {
```
pmusaraj: This line means the SVG sprite isn't loaded when testing (without it, the tests would be inundated with `Unhandled request in test environment` errors). Should it be added as a fixture? Should it check for `Ember.testing` instead?

app/assets/javascripts/discourse/widgets/avatar-flair.js.es6
```diff
@@ -52,20 +52,8 @@ createWidget("avatar-flair", {
...
-                Handlebars.Utils.escapeExpression(
-                  attrs.primary_group_flair_color
-                ) +
-                "; "
-              : ""
-          }
-        })
-      ];
+      let icon = convertIconClass(attrs.primary_group_flair_url);
+      return [iconNode(icon)];
```
pmusaraj: The style attribute here is redundant, it's already set on the parent tag.

app/assets/stylesheets/common/base/colorpicker.scss
```diff
@@ -35,16 +35,10 @@
...
-        align-items: center;
-        justify-content: center;
-        font-family: fontawesome;
-        content: "\f00c";
-        height: 100%;
-        width: calc(1.25em - 2px);
-        color: rgba($secondary, 0.75);
+        top: 0px;
+        left: 2px;
+        content: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="11px" height="11px" viewBox="0 0 512 512" fill="white"><path d="M173.898 439.404l-166.4-166.4c-9.997-9.997-9.997-26.206 0-36.204l36.203-36.204c9.997-9.998 26.207-9.998 36.204 0L192 312.69 432.095 72.596c9.997-9.997 26.207-9.997 36.204 0l36.203 36.204c9.997 9.997 9.997 26.206 0 36.204l-294.4 294.401c-9.998 9.997-26.207 9.997-36.204-.001z" /></svg>');
```
pmusaraj: Here (and in a few other similar cases) it was complicated to replace a font icon used in a pseudoselector with an icon in the DOM, so I opted to use SVG in CSS directly. It's not ideal, but it helps keep down the size of the PR.

app/assets/javascripts/discourse-common/lib/icon-library.js.es6
```diff
@@ -68,58 +238,114 @@ export function registerIconRenderer(renderer) {
...
 registerIconRenderer({
   name: "font-awesome",
 
   string(icon, params) {
-    let tagName = params.tagName || "i";
-    let html = `<${tagName} class='${faClasses(icon, params)}'`;
-    if (params.title) {
-      html += ` title='${I18n.t(params.title).replace(/'/g, "&#39;")}'`;
-    }
+    let id = handleIconId(icon);
```
jjaffeux: should be const

jjaffeux:
```suggestion
    const id = handleIconId(icon);
```

app/assets/javascripts/discourse-common/lib/icon-library.js.es6
```diff
@@ -68,58 +238,114 @@ export function registerIconRenderer(renderer) {
...
   node(icon, params) {
-    let tagName = params.tagName || "i";
+    let id = handleIconId(icon);
+    let classes = iconClasses(icon, params) + " svg-node";
 
-    const properties = {
-      className: faClasses(icon, params),
-      attributes: { "aria-hidden": true }
-    };
+    let svg = h(
```
jjaffeux:
```suggestion
    const svg = h(
```

app/assets/javascripts/discourse-common/lib/icon-library.js.es6
```diff
@@ -68,58 +238,114 @@ export function registerIconRenderer(renderer) {
...
+        params.title
+      ).replace(/'/g, "&#39;")}'>${html}</span>`;
+    }
     return html;
   },
 
   node(icon, params) {
-    let tagName = params.tagName || "i";
+    let id = handleIconId(icon);
+    let classes = iconClasses(icon, params) + " svg-node";
```
jjaffeux:
```suggestion
    const classes = iconClasses(icon, params) + " svg-node";
```

app/assets/javascripts/discourse/lib/svg-sprite-loader.js.es6
```diff
@@ -0,0 +1,17 @@
+import { ajax } from "discourse/lib/ajax";
+
+export default {
+  name: "svg-sprite-loader",
+  load(spritePath, spriteName) {
+    const c = "svg-sprites";
+    const cEl = `#${c}`;
```
jjaffeux: I would write:

```
const $cEl = $(`#${c}`);
```

app/assets/javascripts/discourse/lib/svg-sprite-loader.js.es6
```diff
@@ -0,0 +1,17 @@
...
+    const spriteEl = `${cEl} .${spriteName}`;
```
jjaffeux: I would write:

```
const $spriteEl = $cEl.find(`.${spriteName}`);
```

app/assets/javascripts/discourse/widgets/avatar-flair.js.es6
```diff
@@ -52,20 +52,8 @@ createWidget("avatar-flair", {
...
-              ? "color: #" +
-                Handlebars.Utils.escapeExpression(
-                  attrs.primary_group_flair_color
-                ) +
-                "; "
-              : ""
-          }
-        })
-      ];
+      let icon = convertIconClass(attrs.primary_group_flair_url);
```
jjaffeux:
```suggestion
      const icon = convertIconClass(attrs.primary_group_flair_url);
```

lib/svg_sprite/svg_sprite.rb
```diff
@@ -0,0 +1,218 @@
...
+          @svg_subset << sym.to_xml
+        end
+      end
+    end
+
+    @svg_subset << '</svg>'
+  end
+
+  def self.version(svg_subset)
+    (@svg_subset_cache ||= {})[svg_subset] ||=
```
jjaffeux: 😅 not a big of this double ||=

lib/svg_sprite/svg_sprite.rb
```diff
@@ -0,0 +1,218 @@
...
+      Digest::SHA1.hexdigest(svg_subset)
+  end
+
+  def self.path
+    "/svg-sprite/#{Discourse.current_hostname}/#{version all_icons.to_s}.svg"
```
jjaffeux: if it's not a typo and version is a method I would add parentheses here: version(all_icons.to_s)

lib/svg_sprite/svg_sprite.rb
pmusaraj: It sure isn't easy to read, but the `version` method here and the `path` method just below it were pretty much copy-pasted from [highlight_js.rb](https://github.com/discourse/discourse/blob/48501b0d45f3144fc7063e16f508e29b22ef6ee0/lib/highlight_js/highlight_js.rb#L24-L31).

lib/svg_sprite/svg_sprite.rb
```diff
@@ -0,0 +1,218 @@
+module SvgSprite
```
SamSaffron: https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html
# frozen_string_literal: true

We should probably adopt this globally to be honest

app/views/wizard/index.html.erb
```diff
@@ -16,5 +16,12 @@
   <body class='wizard'>
     <div id='wizard-main'></div>
     <%= preload_script 'wizard-start' %>
+
+    <div id="svg-sprites" style="display:none;">
+      <div class="fontawesome">
+        <%= raw SvgSprite.bundle %>
```
SamSaffron: hmmm so the wizard bundles ? why is is not loading via ajax?

SamSaffron:
are custom icons for badges / staff flair tested here?

app/assets/javascripts/discourse-common/lib/icon-library.js.es6
```diff
@@ -68,58 +238,114 @@ export function registerIconRenderer(renderer) {
...
 }
 
-// Support for font awesome icons
-function faClasses(icon, params) {
-  let classNames = `fa fa-${icon.replacementId || icon.id} d-icon d-icon-${
-    icon.id
-  }`;
+function iconClasses(icon, params) {
+  // "notification." is invalid syntax for classes, use replacement instead
+  let dClass =
```
jjaffeux:
```suggestion
  const dClass =
```

app/views/wizard/index.html.erb
pmusaraj: I opted to use the bundle on the wizard to keep a smaller code footprint. To get the same AJAX-loading into the wizard would require exposing the SvgSprite path to the wizard JS, including the the sprite loader and adding an initializer to the wizard, the equivalent of `svg-sprite-fontawesome.js.es6`.

app/assets/stylesheets/common/base/colorpicker.scss
eviltrout: I don't love this. I think I'd prefer to remove the icon altogether if we can't easily put it in the DOM. It's for the color picker? It doesn't have a spot for HTML?

config/routes.rb
```diff
@@ -446,6 +447,8 @@
   # in most production settings this is bypassed
   get "letter_avatar_proxy/:version/letter/:letter/:color/:size.png" => "user_avatars#show_proxy_letter"
 
+  get "svg-sprite/:hostname/:version.svg" => "svg_sprite#show", format: false, constraints: { hostname: /[\w\.-]+/ }
```
eviltrout: Is there any constraint on the version name possible? I always prefer it if we can.

lib/svg_sprite/svg_sprite.rb
```diff
@@ -0,0 +1,218 @@
...
+  def self.badge_icons
+    Badge.all.pluck(:icon).uniq.each { |icon| process(icon) }
+  end
+
+  def self.group_icons
+    Group.where("flair_url LIKE '%fa-%'").pluck(:flair_url).uniq.each { |icon| process(icon) }
+  end
+
+  def self.theme_icons
+    theme_icon_settings = Array.new
```
eviltrout: Trivial but why not `= []` ?

pmusaraj:
@eviltrout I have replaced another batch of SVGs in the stylesheets, there are four remaining cases: 2 in lightbox.scss, one specific to instagram videos in onebox.scss and one in attachments in topic-post.scss. These three cases each have some difficulties, but all would require rebaking all posts. Let me know if you'd like me to pursue these cases as well.

eviltrout:
I think those are fine if difficult. Perhaps add comments saying so, so
that a future coder doesn’t copy it without thinking.

PR #6560, opened 2018-11-04
FIX: don't allow adding a value containing vertical bar char to the secret list

majakomel:
Prevents users entering vertical bar char in SSO provider secrets site settings and breaking it.


SamSaffron:
looks good to me, client side validation really is the only thing I can think of that can easily work here short of double encoding `|` which is a massive amount of work for such an edge case

PR #6562, opened 2018-11-05
FEATURE: Compute distance between logins to generate login alerts.

nbianca:


app/models/user_auth_token.rb
```diff
@@ -30,23 +30,24 @@ def self.log(info)
...
-  #
-  # On the other hand, if this is too broad it will not report any suspicious
-  # logins at all.
-  #
-  # For example, let's choose the country as the only component in login
-  # locations. In general, this should be a pretty good choce with the
-  # exception that for users from huge countries it might not be specific
-  # enoguh. For US users where the real user and the malicious one could
-  # happen to live both in USA, this will not detect any suspicious activity.
+  MAX_SUSPICIOUS_DISTANCE = 500
```
ZogStriP: What's the unit? Km?

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,24 @@ def self.log(info)
...
+
   def self.login_location(ip)
-    DiscourseIpInfo.get(ip)[:country]
+    ipinfo = DiscourseIpInfo.get(ip)
+    [ipinfo['latitude'], ipinfo['longitude']]
+  end
+
+  def self.distance(loc1, loc2)
+    rad_per_deg = Math::PI / 180
```
ZogStriP: Looks like this is a constant, so this should probably be extracted out that method?

app/models/user_auth_token.rb
nbianca: Yes, kilometres. I will add a comment.

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,24 @@ def self.log(info)
...
+    radius = 6371 # kilometers
+
+    lat1_rad, lon1_rad = loc1[0] * rad_per_deg, loc1[1] * rad_per_deg
+    lat2_rad, lon2_rad = loc2[0] * rad_per_deg, loc2[1] * rad_per_deg
+
+    a = Math.sin((lat2_rad - lat1_rad) / 2)**2 + Math.cos(lat1_rad) * Math.cos(lat2_rad) * Math.sin((lon2_rad - lon1_rad) / 2)**2
+    c = 2 * Math::atan2(Math::sqrt(a), Math::sqrt(1 - a))
+
+    c * radius
```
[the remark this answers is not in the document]
nbianca: It's in kilometres.

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,23 @@ def self.log(info)
...
-  # enoguh. For US users where the real user and the malicious one could
-  # happen to live both in USA, this will not detect any suspicious activity.
+  MAX_SUSPICIOUS_DISTANCE_KM = 500 # kilometers
+  RAD_PER_DEG = Math::PI / 180
+  EARTH_RADIUS_KM = 6371 # kilometers
+
   def self.login_location(ip)
-    DiscourseIpInfo.get(ip)[:country]
+    ipinfo = DiscourseIpInfo.get(ip)
+    [ipinfo['latitude'], ipinfo['longitude']]
```
SamSaffron: Is there any chance this could return `[nil, nil]`

Cause if this can happen:

`loc1[0] * RAD_PER_DEG, loc1[1] * RAD_PER_DEG` is going to:

`NoMethodError (undefined method `*' for nil:NilClass)`

app/models/user_auth_token.rb
```diff
@@ -58,7 +58,7 @@ def self.is_suspicious(user_id, user_ip)
     return false if ips.empty? # first login is never suspicious
 
     user_location = login_location(user_ip)
-    ips.none? { |ip| user_location == login_location(ip) }
+    ips.none? { |ip| distance(user_location, login_location(ip)) < MAX_SUSPICIOUS_DISTANCE_KM }
```
SamSaffron: I would go with a hidden site setting here instead, cause it will allow us to tweak it more easily.

Additionally people who want the distance thing off can just set it to 0.

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,23 @@ def self.log(info)
...
+  end
+
+  def self.distance(loc1, loc2)
+    lat1_rad, lon1_rad = loc1[0] * RAD_PER_DEG, loc1[1] * RAD_PER_DEG
+    lat2_rad, lon2_rad = loc2[0] * RAD_PER_DEG, loc2[1] * RAD_PER_DEG
+
+    a = Math.sin((lat2_rad - lat1_rad) / 2)**2 + Math.cos(lat1_rad) * Math.cos(lat2_rad) * Math.sin((lon2_rad - lon1_rad) / 2)**2
+    c = 2 * Math::atan2(Math::sqrt(a), Math::sqrt(1 - a))
+
+    c * EARTH_RADIUS_KM
```
tgxworld: I think we should improve the names of the variables used for `a` and `c`.

app/models/user_auth_token.rb
nbianca: If there is no MaxMindDb, this might happen. Thanks!

app/models/user_auth_token.rb
nbianca: Hm.. sounds like a great idea. Thank you!

app/models/user_auth_token.rb
nbianca: They do not have any particular meaning and are just partial results of the formula. I have checked now and most implementations I could find on the internet just mention them as `a` and `c`. 🤔

tgxworld: ah icic. This is fine then :+1:

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,24 @@ def self.log(info)
...
-  # exception that for users from huge countries it might not be specific
-  # enoguh. For US users where the real user and the malicious one could
-  # happen to live both in USA, this will not detect any suspicious activity.
+  RAD_PER_DEG = Math::PI / 180
+  EARTH_RADIUS_KM = 6371 # kilometers
+
   def self.login_location(ip)
-    DiscourseIpInfo.get(ip)[:country]
+    ipinfo = DiscourseIpInfo.get(ip)
+    if ipinfo['latitude'] && ipinfo['longitude']
```
ZogStriP: I prefer when we're explicit about the return value of a function. If `ipinfo['latitude']` or `ipinfo['longitude']` is `nil`, then the method will return `nil`. But that's not very clear.

What about?

```ruby
def self.login_location(ip)
  ip_info = DiscourseIpInfo.get(ip)
  ip_info["latitude"] && ip_info["longitude"] ? [ip_info["latitude"], ip_info["longitude"]] : nil
end
```

app/models/user_auth_token.rb
```diff
@@ -57,8 +57,13 @@ def self.is_suspicious(user_id, user_ip)
     ips.uniq!
     return false if ips.empty? # first login is never suspicious
 
-    user_location = login_location(user_ip)
-    ips.none? { |ip| user_location == login_location(ip) }
+    if user_location = login_location(user_ip)
```
SamSaffron: if there are no locations shouldn't everything be suspicious?

app/models/user_auth_token.rb
```diff
@@ -30,23 +30,23 @@ def self.log(info)
...
-  # exception that for users from huge countries it might not be specific
-  # enoguh. For US users where the real user and the malicious one could
-  # happen to live both in USA, this will not detect any suspicious activity.
+  RAD_PER_DEG = Math::PI / 180
+  EARTH_RADIUS_KM = 6371 # kilometers
+
   def self.login_location(ip)
-    DiscourseIpInfo.get(ip)[:country]
+    ipinfo = DiscourseIpInfo.get(ip)
+
```
ZogStriP: Why the empty line?

app/models/user_auth_token.rb
nbianca: In that case, absence of MaxMindDb would cause every login to be suspicious. Is this something we want?

app/models/user_auth_token.rb
nbianca: I just wanted to make the return value more obvious.

PR #6563, opened 2018-11-05
FIX: uses hex to compare images

jjaffeux:
It prevents some terminals from crashing in case of errors and dumping the whole file content into the terminal.

### DIFF
--- a/spec/models/optimized_image_spec.rb
+++ b/spec/models/optimized_image_spec.rb
@@ -17,9 +17,12 @@ describe OptimizedImage do
             5
           )
 
-          expect(File.read(tmp_path)).to eq(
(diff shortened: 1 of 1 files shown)


ZogStriP:
Looks 👍. Are there other specs where we do this?

jjaffeux:
> Looks 👍. Are there other specs where we do this?

No I don’t think so

ZogStriP:
What about https://github.com/discourse/discourse/blob/master/spec/lib/upload_recovery_spec.rb#L96?

eviltrout:
@ZogStriP that one is not changing the file or creating a file, so I think it wouldn't have the same issue.

PR #6566, opened 2018-11-06
Full height swipe-able menus

featheredtoast:
Add full height menu on mobile, animated slide in menus, and swipe-to-dismiss menu gestures. Android additionally gets swipe-to-slide-in gestures for menus. (iOS does not respond to this as it's reserved for forward/back motions. Dismissal swipes still work from the center of the screen).

Adds swipe support for iOS.

### DIFF
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -18,6 +18,8 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
   _isPanning: false,
   _panMenuOrigin: "right",
   _panMenuOffset: 0,
+  _scheduledMovingAnimation: null,
+  _scheduledRemoveAnimate: null,
   _topic: null,
 
   @observes(
@@ -62,7 +64,9 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
         this._animateOpening($panel);
       } else {
         //continue to open or close menu
-        window.requestAnimationFrame(() => this._handlePanDone(offset, event));
+        this._scheduledMovingAnimation = window.requestAnimationFrame(() =>
+          this._handlePanDone(offset, event)
+        );
       }
     });
   },
@@ -244,6 +248,8 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
         .off("pointermove")
         .off("pointercancel");
     }
+    Ember.run.cancel(this._scheduledRemoveAnimate);
+    window.cancelAnimationFrame(this._scheduledMovingAnimation);
   },
 
   buildArgs() {
@@ -361,7 +367,7 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
       if (this._animate) {
         $panel.addClass("animate");
         $headerCloak.addClass("animate");
-        Ember.run.later(() => {
+        this._scheduledRemoveAnimate = Ember.run.later(() => {
           $panel.removeClass("animate");
           $headerCloak.removeClass("animate");
         }, 200);
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -61,13 +61,13 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     const panMenuOrigin = this.get("panMenuOrigin");
     const panMenuOffset = this.get("panMenuOffset");
     if (panMenuOrigin === "right" && panMenuOffset === 0) {
-      return (e.deltaX > 200 && e.velocityX > -0.15) || e.velocityX > 0.15;
+      return (e.deltaX > 200 && e.velocityX > -0.10) || e.velocityX > 0.10;
     } else if (panMenuOrigin === "left" && panMenuOffset === 0) {
-      return (e.deltaX < -200 && e.velocityX < 0.15) || e.velocityX < -0.15;
+      return (e.deltaX < -200 && e.velocityX < 0.10) || e.velocityX < -0.10;
     } else if (panMenuOrigin === "right" && panMenuOffset !== 0) {
-      return (e.deltaX > 200 && e.velocityX > -0.15) || e.velocityX > 0.15;
+      return (e.deltaX > 200 && e.velocityX > -0.10) || e.velocityX > 0.10;
     } else if (panMenuOrigin === "left" && panMenuOffset !== 0) {
-      return (e.deltaX < -200 && e.velocityX < 0.15) || e.velocityX < -0.15;
+      return (e.deltaX < -200 && e.velocityX < 0.10) || e.velocityX < -0.10;
     }
   },
 
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -180,6 +180,7 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
   },
 
   setTopic(topic) {
+    this.eventDispatched("dom:clean", "header");
     this._topic = topic;
     this.queueRerender();
   },
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -138,7 +138,6 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     if (!this.get("isPanning")) {
       return;
     }
-    e.originalEvent.preventDefault();
     const $menuPanels = $(".menu-panel");
     const panMenuOffset = this.get("panMenuOffset");
     $menuPanels.each((idx, panel) => {
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -82,6 +82,7 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
         $centeredElement.parents(".panel-body").length) &&
       (e.direction === "left" || e.direction === "right")
     ) {
+      e.originalEvent.preventDefault();
       this.set("isPanning", true);
     } else if (
       center.x < 30 &&
@@ -137,6 +138,7 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     if (!this.get("isPanning")) {
       return;
     }
+    e.originalEvent.preventDefault();
     const $menuPanels = $(".menu-panel");
     const panMenuOffset = this.get("panMenuOffset");
     $menuPanels.each((idx, panel) => {
@@ -334,8 +336,10 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
 
         const menuTop = this.site.mobileView ? 0 : headerHeight();
 
-        let height;
-        const winHeight = $(window).height() - 16;
+        let height = window.innerHeight
+          ? window.innerHeight
+          : $(window).height();
+        const winHeight = height - 16;
         height = winHeight - menuTop;
 
         if ($panelBody.prop("style").height !== "100%") {
--- a/app/assets/javascripts/discourse/components/site-header.js.es6
+++ b/app/assets/javascripts/discourse/components/site-header.js.es6
@@ -47,8 +47,8 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     const $window = $(window);
     const windowWidth = parseInt($window.width());
     const $menuPanels = $(".menu-panel");
-    this._shouldPanClose(event) ? (offset += velocity) : (offset -= velocity);
     const menuOrigin = this.get("panMenuOrigin");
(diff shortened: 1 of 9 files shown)


coding-horror:
https://media.giphy.com/media/asHT7eh4AwG9G/200.gif

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -23,6 +28,135 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
+      const $headerCloak = $(".header-cloak");
+      $panel.css(menuOrigin, -offset);
+      $headerCloak.css("opacity", Math.min(0.5, (300 - offset) / 600));
+      if (offset > windowWidth) {
+        $panel.css(menuOrigin, -windowWidth);
+        this.set("animate", true);
+        this.eventDispatched("dom:clean", "header");
+        this.set("panMenuOffset", 0);
+      } else if (offset <= 0) {
+        $panel.css("right", "");
```
jjaffeux:
```
$panel.css({ right: "", left: "" });
```

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -180,6 +354,18 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
       $panel.width(width);
+      if (animate) {
+        $panel.addClass("animate");
+        $headerCloak.addClass("animate");
+        Ember.run.later(() => {
+          $panel.removeClass("animate");
+          $headerCloak.removeClass("animate");
+        }, 200);
+      }
+      $panel.css("right", "");
```
jjaffeux: $panel.css({ right: "", left: "" });

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -416,10 +428,32 @@ export default createWidget("header", {
...
 
   toggleHamburger() {
     this.state.hamburgerVisible = !this.state.hamburgerVisible;
+    this.toggleBodyScrolling(this.state.hamburgerVisible);
+  },
+
+  toggleBodyScrolling(bool) {
+    if (!this.site.mobileView) return;
+    if (bool) {
+      document.body.addEventListener('touchmove', this.preventDefault, { passive: false });
```
jjaffeux:
```suggestion
      document.body.addEventListener("touchmove", this.preventDefault, { passive: false });
```

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -416,10 +428,32 @@ export default createWidget("header", {
...
+    } else {
+      document.body.removeEventListener('touchmove', this.preventDefault, { passive: false });
```
jjaffeux:
```suggestion
      document.body.removeEventListener("touchmove", this.preventDefault, { passive: false });
```

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -416,10 +428,32 @@ export default createWidget("header", {
...
+    }
+  },
+
+  preventDefault(e) {
+    // prevent all scrollin on menu panels, except on overflow
+    let height = window.innerHeight
+        ? window.innerHeight
+        : $(window).height();
+    if (!$(e.target).parents('.menu-panel').length ||
```
jjaffeux:
```suggestion
    if (!$(e.target).parents(".menu-panel").length ||
```

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -416,10 +428,32 @@ export default createWidget("header", {
...
+        $('.menu-panel .panel-body-contents').height() <= height) {
```
jjaffeux:
```suggestion
        $(".menu-panel .panel-body-contents").height() <= height) {
```

jjaffeux:
If you can’t add tests to this code (which I understand given the mobile and gesture nature of the feature) please add more comments and maybe more explicit code. ATM, I feel like I wouldn’t want to have to modify this in the future, it's a lot of special cases with very few explanations.

app/assets/javascripts/discourse/widgets/header.js.es6
```diff
@@ -416,10 +428,32 @@ export default createWidget("header", {
...
+    if (bool) {
+      document.body.addEventListener('touchmove', this.preventDefault, { passive: false });
+    } else {
+      document.body.removeEventListener('touchmove', this.preventDefault, { passive: false });
+    }
+  },
+
+  preventDefault(e) {
+    // prevent all scrollin on menu panels, except on overflow
+    let height = window.innerHeight
```
jjaffeux:
```suggestion
    const height = window.innerHeight
```

app/assets/javascripts/discourse/widgets/user-menu.js.es6
```diff
@@ -194,7 +194,23 @@ export default createWidget("user-menu", {
...
+  clickOutside(e) {
+    const $centeredElement = $(document.elementFromPoint(e.clientX, e.clientY));
+    if (
+      $centeredElement.parents(".panel").length &&
+      !$centeredElement.hasClass("header-cloak")
+    ) {
+      this.sendWidgetAction("toggleUserMenu");
+    } else {
+      const $window = $(window);
+      const windowWidth = parseInt($window.width());
```
jjaffeux: It's good practice to always use the radix, even if it should never be an issue here.

```suggestion
      const windowWidth = parseInt($window.width(), 10);
```

app/assets/javascripts/discourse/widgets/hamburger-menu.js.es6
```diff
@@ -323,7 +323,24 @@ export default createWidget("hamburger-menu", {
...
+  clickOutside(e) {
+    const $centeredElement = $(document.elementFromPoint(e.clientX, e.clientY));
+    if (
+      $centeredElement.parents(".panel").length &&
+      !$centeredElement.hasClass("header-cloak")
+    ) {
+      this.sendWidgetAction("toggleHamburger");
+    } else {
+      const $window = $(window);
+      const windowWidth = parseInt($window.width());
```
jjaffeux:
```suggestion
      const windowWidth = parseInt($window.width(), 10);
```

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -160,16 +332,18 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
+        if (this.site.mobileView) {
+          $headerCloak.show();
         }
 
+        const menuTop = this.site.mobileView ? 0 : headerHeight();
+
+        let height = window.innerHeight
+          ? window.innerHeight
+          : $(window).height();
+        const winHeight = height - 16;
```
jjaffeux: can you clarify the meaning of 16 here ?

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -23,6 +28,135 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
     this.queueRerender();
   },
 
+  _panOpenClose(offset, velocity, direction) {
+    const $window = $(window);
+    const windowWidth = parseInt($window.width());
+    const $menuPanels = $(".menu-panel");
```
jjaffeux: used only once

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -23,6 +28,135 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
+        $panel.css(menuOrigin, -windowWidth);
+        this.set("animate", true);
+        this.eventDispatched("dom:clean", "header");
+        this.set("panMenuOffset", 0);
+      } else if (offset <= 0) {
+        $panel.css("right", "");
+        $panel.css("left", "");
+        this.set("panMenuOffset", 0);
+      } else {
+        Ember.run.later(
```
eviltrout: I wonder if this event should check if one is already scheduled and clear it first? 20ms is pretty fast but is there any chance two could be enqueued at once? I would recommend storing the timer in a private variable and calling Ember.run.cancel before.

You should also clear the timeout in `willDestroyElement`

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -23,6 +28,135 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
+    if (
+      ($centeredElement.hasClass("panel-body") ||
+        $centeredElement.hasClass("header-cloak") ||
+        $centeredElement.parents(".panel-body").length) &&
+      (e.direction === "left" || e.direction === "right")
+    ) {
+      e.originalEvent.preventDefault();
+      this.set("isPanning", true);
+    } else if (
+      center.x < 30 &&
```
eviltrout: There are a couple more constants here, 30 and 300 it seems.

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -74,6 +209,14 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
         this.eventDispatched("dom:clean", "header");
       }
     });
+
+    if (this.site.mobileView) {
```
eviltrout: Would it be better to check the capabilities object for touch? Some touch devices like ipad are not in mobileView. `this.capabilities.touch`

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -180,6 +354,18 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
       }
 
       $panel.width(width);
+      if (animate) {
+        $panel.addClass("animate");
+        $headerCloak.addClass("animate");
+        Ember.run.later(() => {
```
eviltrout: This is another event you will want to make sure only happens once and is cleared on destroy (see my earlier comments.)

app/assets/javascripts/discourse/mixins/pan-events.js.es6
```diff
@@ -7,11 +11,23 @@ export default Ember.Mixin.create({
     this._super();
 
     if (this.site.mobileView) {
-      this.$()
-        .on("pointerdown", e => this._panStart(e))
-        .on("pointermove", e => this._panMove(e))
-        .on("pointerup", e => this._panMove(e))
-        .on("pointercancel", e => this._panMove(e));
+      if ("onpointerdown" in document.documentElement) {
```
eviltrout: This could use `this.capabilities.touch` too right? If not, should it be added to capabilities?

featheredtoast:
Thanks for the concerns/feedback - I've implemented most of the feedback given so far.

Two points about the capacities, and rationale behind mobile only - tldr, it'd be nice to get to supporting all touch devices at some point but I'm focusing/optimizing for mobile only here:

onpointerdown seems to be a newer touch API, not (yet) supported in iOS/webkit - I wrote against that API before I realized ontouch is the only thing I've got on iOS. Probably not worth adding to capacities, I added this as a shim to support both APIs, rather than rewrite against ontouch only.

I also wanted this work to be incremental, which is why I opted to not support anything other than mobile views (for now). While we could add support for ipads and tablets, that would be a secondary goal for me. The work as is also includes masking the side of a screen for a menu, and optimizes touch events for mobile specifically where unmasked content is not useful at all.

app/assets/javascripts/discourse/components/site-header.js.es6
```diff
@@ -23,6 +30,146 @@ const SiteHeaderComponent = MountWidget.extend(Docking, {
...
+      return (
+        (e.deltaX < -distanceThreshold && e.velocityX < velocityThreshold) ||
+        e.velocityX < -velocityThreshold
+      );
+    }
+  },
+
+  panStart(e) {
+    //android supports pulling in from the screen edges
+    const screenEdgeMargin = 30;
```
eviltrout: I would recommend pulling these out and putting them at the top of the file like `SCREN_EDGE_MARGIN = 30`. It's nice to collect those kinds of things up top in case you want to export them later.

app/assets/javascripts/discourse/components/topic-navigation.js.es6
```diff
@@ -136,10 +136,11 @@ export default Ember.Component.extend(PanEvents, {
   },
 
   _shouldPanClose(e) {
-    return (e.deltaY > 200 && e.velocityY > -0.15) || e.velocityY > 0.15;
+    return (e.deltaY > 200 && e.velocityY > -0.10) || e.velocityY > 0.10;
```
eviltrout: no constants for this stuff?

featheredtoast: Oops! missed those

featheredtoast:
Refactored the shared ones into the pan events mixin to be used in both site-header and topic-navigation.

Additionally, I've refactored topic-navigation to better match the feedback I got from Joffrey.

davidtaylorhq:
@featheredtoast you may already be aware, but for me the slide-in menus happen on Desktop when my browser is less than about 1200px wide. Is that deliberate?

featheredtoast:
@davidtaylorhq Was pointed out in chat on Tuesday as well I believe. I've changed the behavior so it's only switched for mobile views only.

PR #6570, opened 2018-11-06
UX: Make the edits indicator a real link for accessibility

mwcampbell:
This link didn't have an `href` attribute, so it wasn't in the tab
order. This commit fixes that, while adding a call to `preventDefault`
in the event handler to avoid any regression in event handling.


SamSaffron:
this looks good but can the `href` just be blank? does it need to be `#` ? merging in for now, seems safe.

mwcampbell:
A blank `href` is less safe if, for some reason, the browser does the default action of following the link despite our best efforts. Firefox and Chrome reload the page, while IE apparently opens the URL with the last path segment removed. By comparison, using `#` in the `href` is harmless, and it's a common practice for links that are really just click targets.

PR #6571, opened 2018-11-06
FEATURE: Add new setting to force user edit last post.

nbianca:


coding-horror:
I am not sure this is what was discussed.. there should be a site setting for the maximum number of consecutive replies in the same topic by the same user? 

https://meta.discourse.org/t/allow-only-x-consecutive-replies-in-the-same-topic/69756

So the alert after reaching the threshold would be

> No more than x consecutive replies are allowed. Please edit your previous reply instead, or wait for someone to reply to you.

lib/validators/post_validator.rb
```diff
@@ -141,6 +142,25 @@ def unique_post_validator(post)
     end
   end
 
+  def force_edit_last_validator(post)
+    return if SiteSetting.max_consecutive_replies == 0 || post.id #|| acting_user_is_trusted?(post) || private_message?(post)
```
ZogStriP: I don't think you meant to comment these conditions out?

lib/validators/post_validator.rb
```diff
@@ -141,6 +142,25 @@ def unique_post_validator(post)
...
+
+    posts = post.topic&.posts.order('post_number desc').where('not hidden AND posts.deleted_at IS NULL')
```
ZogStriP: I think you can come up with **one** SQL query that returns the count of the last consecutive posts made by the current user instead of retrieving all the posts and iterating over them in Ruby-land.

nbianca: It's definitely do-able, but I have a feeling it's going to get ugly. 😨

coding-horror:
You can fast exit if the most recent reply was not from the current user.
Maybe that would be a good first check before proceeding further, to use a
cheap query up front and then only go with the second more expensive query
if it is needed?

nbianca:
I do not think it's possible to write a query that is doing this and is also readable. In theory, the best query should check if the last X posts are all made by the same user AND then fetch only the last one.

What I do now is:
* fetch at most `SiteSetting.max_consecutive_replies` user IDs (which is not that much) that have last posted in the current topic
* iterate over those IDs in Ruby and count how many of them were made by current user (usually a few cheap iterations)
* if user has reached the threshold, do another query to fetch the last post that would be passed to the Guardian

If we consider that in most communities users will not post many consecutive replies, in that case this validator is going to query the database only once and the result will usually be small (few integers).

ZogStriP:
> In theory, the best query should check if the last X posts are all made by the same user AND then fetch only the last one.

No, that's not what I meant. I meant you could do one query to "check whether the last poster has made X consecutive replies" ;)

nbianca:
> No, that's not what I meant. I meant you could do one query to "check whether the last poster has made X consecutive replies" ;)

Yes, I understand that. That would be the "gaps and islands" problem and I believe the query will be way too complicated and the benefit will be very small. I will look into it.

coding-horror:
I am not following. How can there be gaps when you are only checking the
last post equals to the current poster? There is no "gap" at the end of a
topic..?

It seems to me the current poster is the only one we care about, and only
the last post in the topic (and possibly last x posts in topic)?

ZogStriP:
Exactly, we only care about knowing whether the last X posts in a topic were done by the current user.

nbianca:
I was considering current user's posts as "islands" and other posts as "gaps" and it kind of made sense. Since the maximum length of an island is known (`SiteSetting.max_consecutive_replies`), I was able to simplify the query.

I believe it should be good.

lib/validators/post_validator.rb
```diff
@@ -141,6 +142,26 @@ def unique_post_validator(post)
     end
   end
 
+  def force_edit_last_validator(post)
+    return if SiteSetting.max_consecutive_replies == 0 || post.id || post.acting_user.try(:staff?) || private_message?(post)
```
ZogStriP: Why `post.id`?

lib/validators/post_validator.rb
```diff
@@ -141,6 +142,26 @@ def unique_post_validator(post)
...
+
+    topic = post.topic
+
+    last_posts_count = DB.query_single(<<~SQL, topic_id: topic.id, max_replies: SiteSetting.max_consecutive_replies, user_id: post.user_id).first
```
ZogStriP:
```suggestion
last_posts_count = Post.visible.where(topic: topic, user: post.acting_user).order(post_number: :desc).limit(SiteSetting.max_consecutive_replies).count
```

lib/validators/post_validator.rb
nbianca: Presence of `post.id` signals that the post already exists and this validator runs for an edited post, which should not be happen.

lib/validators/post_validator.rb
```diff
@@ -141,6 +142,26 @@ def unique_post_validator(post)
...
+      WITH topic_posts AS (
+        SELECT * FROM posts
```
ZogStriP:
```suggestion
        SELECT user_id FROM posts
```

Since that's all you care about ;)

ZogStriP: I also like this style, but that's up to you

```ruby
DB.query_single(<<~SQL, topic_id: post.topic_id, user_id: post.acting_user.id, max_replies: SiteSetting.max_consecutive_replies).first
  SELECT COUNT(*)
  FROM (
    SELECT user_id
      FROM posts
     WHERE deleted_at IS NULL
       AND NOT hidden
       AND topic_id = :topic_id
     ORDER BY post_number DESC
     LIMIT :max_replies
  ) c
  WHERE c.user_id = :user_id
SQL
```

lib/validators/post_validator.rb
tgxworld:
```suggestion
    return if SiteSetting.max_consecutive_replies == 0 || post.id || post.acting_user&.staff? || private_message?(post)
```

spec/components/validators/post_validator_spec.rb
```diff
@@ -219,6 +219,43 @@
...
+    it "should always allow editing" do
+      post = Fabricate(:post, user: user, topic: topic)
+      post = Fabricate(:post, user: user, topic: topic)
+
+      revisor = PostRevisor.new(post)
+      revisor.revise!(post.user, raw: 'hello world123456789')
+    end
+
+    it "should allow posting more than 2 replies" do
+      1.upto(3).each do |i|
```
tgxworld:
```suggestion
      3.times.each do |i|
```

ZogStriP: `times` is 0-based ;)

PR #6574, opened 2018-11-07
FIX: Relative links in translations should work with subfolder

gschlager:
Updates translations which contain relative links (HTML and Markdown) so that they work with subfolders. It adds specs to prevent us from adding new translations with the same problem.

This supersedes https://github.com/discourse/discourse/pull/6510 which has been reverted.

**Do not merge yet.** I need to fix translations in a couple of plugins and to work on pushing fixed translations to Transifex, otherwise this commit will invalidate some important translations like `read_more_MF` and `guidelines_topic`.


app/assets/javascripts/admin/components/admin-report.js.es6
```diff
@@ -41,7 +41,7 @@ export default Ember.Component.extend({
   classNameBindings: ["isEnabled", "isLoading", "dasherizedDataSourceName"],
   classNames: ["admin-report"],
   isEnabled: true,
-  disabledLabel: "admin.dashboard.disabled",
+  disabledLabel: I18n.t("admin.dashboard.disabled"),
```
gschlager: @jjaffeux I had to change `disabledLabel` from being an I18n key to the actual translated string. I hope that doesn't create any problems.

PR #6576, opened 2018-11-07
FIX: Use dynamic name resolution for the nginx letter avatar proxy

mwcampbell:
This is desirable for two reasons:

1. As reported in a [bug
topic](https://meta.discourse.org/t/restarting-container-when-internet-is-down-causes-discourse-to-fail/42615/4),
if nginx resolves Internet hostnames on startup, then the container
won't run if the host is disconnected from the Internet at startup.

2. If the IP address of the avatars.discourse.org hostname ever changes,
then nginx will end p using an old IP address until the administrator
restarts the container, unless we use dynamic resolution.

This commit depends on [a corresponding PR in the discourse_docker
repository](https://github.com/discourse/discourse_docker/pull/413).

### DIFF
--- a/config/nginx.sample.conf
+++ b/config/nginx.sample.conf
@@ -214,7 +214,7 @@ server {
       break;
     }
 
-    location /letter_avatar_proxy/ {
+    location ~ ^/letter_avatar_proxy/(.*)$ {
       # Don't send any client headers to the avatars service
       proxy_method GET;
       proxy_pass_request_headers off;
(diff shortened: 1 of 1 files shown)


mwcampbell:
Sorry for the mangled formatting in the description. Looks like I can't edit the original comment now.

tgxworld:
@SamSaffron What do you think about this?

SamSaffron:
I have not forgotten about this, I just am super worried about making this change, I left a comment here: 

https://github.com/discourse/discourse_docker/pull/413

SamSaffron:
I am closing this, not because I do not want the change ... I do ... I just feel the risk is very high and we almost never see this actual problem in the wild.

gschlager:
I'm reopening this in light of recent problems with avatars.dicourse.org due to a CDN change. We should reevaluate our decision. I really like both PRs!

PR #6578, opened 2018-11-07
FIX: make sure "opener window" is not "current window" during onmiauth callback

xrav3nz:
More details: https://meta.discourse.org/t/oauth2-basic-support/33879/168?u=xrav3nz

### DIFF
--- a/app/assets/javascripts/omniauth-complete.js.no-module.es6
+++ b/app/assets/javascripts/omniauth-complete.js.no-module.es6
@@ -4,7 +4,11 @@
   ).dataset;
   const parsedAuthResult = JSON.parse(authResult);
 
-  if (!window.opener) {
+  if (
+    !window.opener ||
+    !window.opener.Discourse ||
+    !window.opener.Discourse.authenticationComplete
+  ) {
     localStorage.setItem("lastAuthResult", authResult);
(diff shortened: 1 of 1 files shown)


davidtaylorhq:
Did you manage to reproduce the case where `window.opener === window`? It seems super weird that that would happen - maybe a browser bug?

In any case, this change looks pretty safe to me :)

xrav3nz:
I can only repro on the site that's experiencing this (with different browsers), but not locally. 🤕 It's indeed super weird.

app/assets/javascripts/omniauth-complete.js.no-module.es6
```diff
@@ -4,7 +4,7 @@
   ).dataset;
   const parsedAuthResult = JSON.parse(authResult);
 
-  if (!window.opener) {
+  if (!window.opener || window.opener === window) {
```
SamSaffron: Maybe we go with the safer?
```suggestion
  if (!window.opener || !window.opener.Discourse || !window.opener.Discourse.authenticationComplete) {
```

xrav3nz:
Reapplied the changes manually because of `prettier` 😅

SamSaffron:
Looks good to me! thanks

PR #6579, opened 2018-11-07
remove JS no longer needed on 404 page

xrav3nz:
I have tried this on a few pages:

- `/404`
- `/random-nonexistent-path`
- `/t/not-found/12312313123`
- Link to the URLs above in a post and click on them

Removing the listener does not seem to break go-back or other navigation functionalities.

The original commit from 2014: https://github.com/discourse/discourse/commit/a967d4dfba292ee15f5a490e98495e35f27b1ffb. /cc @eviltrout do you happen to remember what else the listener is used for?

### DIFF
--- a/app/views/exceptions/not_found.html.erb
+++ b/app/views/exceptions/not_found.html.erb
@@ -37,12 +37,4 @@
       </p>
     </div>
   </div>
-
-  <script language="Javascript">
-    window.onpopstate = function(event) {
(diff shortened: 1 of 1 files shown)


SamSaffron:
I suspect @eviltrout added it so back -> forward buttons work as expected have you tested back / forward?

xrav3nz:
Yup that’s what I mainly tested!

xrav3nz:
Hmm found the unique case that requires the `onpopstate` listener:

1. click on a 404 link in a post
2. **refresh** the 404 page
3. go back / forward no longer works

I will properly extract this.

PR #6585, opened 2018-11-09
FEATURE: Report edit conflicts when saving draft.

nbianca:


config/locales/client.en.yml
```diff
@@ -1353,6 +1353,7 @@ en:
       saved_local_draft_tip: "saved locally"
       similar_topics: "Your topic is similar to..."
       drafts_offline: "drafts offline"
+      edit_conflict: '<i class="fa fa-exclamation-triangle"></i> edit conflict'
```
davidtaylorhq: I wonder if this will be broken by the FA5 change @pmusaraj? Maybe the icon should be added in the handlebars template rather than the translation string?

app/controllers/draft_controller.rb
```diff
@@ -10,6 +10,14 @@ def show
 
   def update
     Draft.set(current_user, params[:draft_key], params[:sequence].to_i, params[:data])
+
+    if params[:post_id] && params[:original_text]
+      post = Post.find_by(id: params[:post_id])
+      if post && post.raw != params[:original_text]
+        return render json: { success: 'OK', edit_conflict: true }
```
davidtaylorhq: I think other places in the codebase do this like
```suggestion
        return render json: success_json.merge(edit_conflict: true)
```
The difference is negligible though

config/locales/client.en.yml
pmusaraj: Yes, the FA5 change would break this. Best to put the icon in the handlebars template.

app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -1002,8 +1003,12 @@ const Composer = RestModel.extend({
 
     // try to save the draft
     return Draft.save(this.get("draftKey"), this.get("draftSequence"), data)
-      .then(function() {
-        composer.set("draftStatus", I18n.t("composer.saved_draft_tip"));
+      .then(function(result) {
```
jjaffeux:
```suggestion
      .then(result => {
```

config/locales/client.en.yml
nbianca: It looks like there are 3 more icons left in i18n files.

pmusaraj: Yes, thanks, I've addressed those in the fontawesome5 branch.

PR #6586, opened 2018-11-09
FIX user directory time period count

majakomel:
https://meta.discourse.org/t/8-visits-per-week-in-user-list/70141


jjaffeux:
@majakomel would it be reasonably easy to write a test for this please ?

PR #6588, opened 2018-11-09
FIX: Improve performance for `Group.posts_for`

davidtaylorhq:


tgxworld:
Looks good to me :+1:

tgxworld:
@davidtaylorhq Once this is deployed, can you have a look at the performance numbers for the before and after queries? Thank you!

PR #6589, opened 2018-11-10
FEATURE: Better handling of quotation marks in site text search

gschlager:
It also matches 3 dots with the ellipsis symbol.


SamSaffron:
Seems good to me.

PR #6591, opened 2018-11-11
Update client.pl_PL.yml

Tae-Sik:


ZogStriP:
Thanks but all our translations has to be done on [transifex](https://www.transifex.com/discourse/discourse-org/).

PR #6594, opened 2018-11-12
allow CSP reports to be sent when header is set outside of Discourse

LeoMcA:
@xrav3nz one for you

[we set our CSP headers in nginx](https://github.com/mozilla/discourse.mozilla.org/blob/4fe534298e2ea9959b06c283f68887adc5599d3c/includes/headers.yml#L2-L14) (allowing us to set different policies for different directives, and ensuring we always fail-safe to a CSP policy even if Discourse isn't setting headers correctly), but currently `/csp_reports` 404s because Discourse doesn't think we have a CSP header set.

I've added another setting to override this behaviour, and force `/csp_reports` to accept reports, even if (from Discourse's perspective) a CSP header isn't being set

### DIFF
--- a/app/controllers/csp_reports_controller.rb
+++ b/app/controllers/csp_reports_controller.rb
@@ -30,6 +30,10 @@ class CspReportsController < ApplicationController
   end
 
   def report_collection_enabled?
-    ContentSecurityPolicy.enabled? && SiteSetting.content_security_policy_collect_reports
+    (
+      ContentSecurityPolicy.enabled? ||
+      SiteSetting.content_security_policy_collect_reports_no_header
+    ) &&
+    SiteSetting.content_security_policy_collect_reports
   end
 end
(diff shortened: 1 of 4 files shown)


xrav3nz:
Sorry, but I don't think core needs this additional site setting. However,

1. I am comfortable with relaxing the condition to not check whether CSP is enabled
2. Or this can be monkey patched in a plugin

Let me know what you think of the report logging feature tho! It's still very rough and subject to lots of changes.

> set different policies for different directives

FWIW it's planned to ship default policies for other directives to cover the minimum that Discourse needs, and they can all be extended.

xrav3nz:
Thanks for testing the waters ❤️

LeoMcA:
> I am comfortable with relaxing the condition to not check whether CSP is enabled

Perfect, I only added the extra setting because I assumed this wasn't the case - I'll update this PR to do this. Perhaps the setting should be default false if CSP doesn't need to be enabled for it to accept reports?

> Or this can be monkey patched in a plugin

Yeah I hadn't considered how easy it would be to patch `report_collection_enabled?`. Let me know if you're more comfortable with me doing this :)

> Let me know what you think of the report logging feature tho! It's still very rough and subject to lots of changes.

I'll let you know when I've had a chance to test it ;)

> Thanks for testing the waters :heart:

Not at all! Thanks for the amazing work on this :smiley:

LeoMcA:
PR updated

PR #6600, opened 2018-11-13
FEATURE: Log password changes in UserHistory

davidtaylorhq:
Currently none of these "non-staff" UserHistory rows are used in the user interface, and can only be accessed through data-explorer. Nonetheless, I think we should be logging password changes. 

This will also be useful for the discourse-password-expiry plugin I am working on.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -497,6 +497,11 @@ class UsersController < ApplicationController
           Invite.invalidate_for_email(@user.email) # invite link can't be used to log in anymore
           secure_session["password-#{token}"] = nil
           secure_session["second-factor-#{token}"] = nil
+          UserHistory.create!(
+            target_user: @user,
+            acting_user: @user,
+            action: UserHistory.actions[:change_password]
+          )
           logon_after_password_reset
         end
       end
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -235,6 +235,22 @@ describe UsersController do
         expect(response).to redirect_to(wizard_path)
       end
 
+      it "logs the password change" do
+        user = Fabricate(:admin)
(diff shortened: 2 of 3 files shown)


spec/requests/users_controller_spec.rb
```diff
@@ -235,6 +235,17 @@
...
 
+      it "logs the password change" do
+        user = Fabricate(:admin)
+        UserAuthToken.generate!(user_id: user.id)
+        token = user.email_tokens.create(email: user.email).token
+        get "/u/password-reset/#{token}"
+
+        expect do
+          put "/u/password-reset/#{token}", params: { password: 'hg9ow8yhg98oadminlonger' }
+        end.to change { UserHistory.count }.by (1)
```
gschlager: I think it's better to check that it actually logs a `change_password` action.

app/controllers/users_controller.rb
```diff
@@ -497,6 +497,10 @@ def password_reset
           Invite.invalidate_for_email(@user.email) # invite link can't be used to log in anymore
           secure_session["password-#{token}"] = nil
           secure_session["second-factor-#{token}"] = nil
+          UserHistory.create!(
+            target_user: @user, acting_user: @user,
```
jjaffeux: each one should be on a different line, this  kind of formatting is error prone in the future

davidtaylorhq: :+1: can we add this to Rubocop? - seems a waste for us humans to have to deal with this ;)

spec/requests/users_controller_spec.rb
davidtaylorhq: :+1: Done

app/controllers/users_controller.rb
jjaffeux: tgx is working on having prettier working in ruby I think, so we should just wait for this

PR #6602, opened 2018-11-13
FIX: change title when primary group changes

xrav3nz:


app/models/group_user.rb
```diff
@@ -24,14 +24,7 @@ def set_notification_level
...
   def set_primary_group
-    if group.primary_group
-      DB.exec("
-        UPDATE users
-        SET primary_group_id = :id
-        WHERE id = :user_id",
-        id: group.id, user_id: user_id
-      )
-    end
+    user.update!(primary_group: group) if group.primary_group
```
xrav3nz: [`user#match_title_to_primary_group_changes`](https://github.com/discourse/discourse/blob/1570b01184864ab3ef9e8f8e5ce302a10ec14c7b/app/models/user.rb#L1327) is an AR callback. It cannot be moved here because [`Admin::UsersController#primary_group`](https://github.com/discourse/discourse/blob/1570b01184864ab3ef9e8f8e5ce302a10ec14c7b/app/controllers/admin/users_controller.rb#L239) also depends on the behavior.

tgxworld:
Looks great to me. Thanks @xrav3nz :+1:

PR #6603, opened 2018-11-13
Disable invites step in wizard when local_logins are turned off

pmusaraj:


jjaffeux:
LGTM 👍

app/serializers/wizard_step_serializer.rb
```diff
@@ -39,6 +39,8 @@ def translate(sub_key, vars = nil)
   end
 
   def description
+    return translate("disabled") if object.disabled
```
tgxworld:
```suggestion
    translate(object.disabled ? "disabled" : translate("description", base_path: Discourse.base_path))
```

app/serializers/wizard_step_serializer.rb
```diff
@@ -39,6 +39,8 @@ def translate(sub_key, vars = nil)
...
+
     translate("description", base_path: Discourse.base_path)
```
tgxworld: Not related but I don't think we need `base_path` to be passed here.

spec/components/wizard/wizard_builder_spec.rb
```diff
@@ -27,4 +27,14 @@
     expect(wizard.steps).to be_blank
   end
 
+  it "returns wizard with disabled invites step when local_logins are off" do
```
tgxworld: I think we should test that the right translation is being displayed as well.

jjaffeux:
@pmusaraj if you have some time can you merge it when you start your day please? I need it for other works. Thanks ❤️

tgxworld:
@jjaffeux feel free to merge it now. My comments are minor and can be addressed later :)

PR #6604, opened 2018-11-14
DEV: Add test for draft edit conflict.

nbianca:
Add test for c78dcde97349ceba2b8e36e87c8c8b3424ac13b6.

### DIFF
--- a/test/javascripts/acceptance/composer-edit-conflict-test.js.es6
+++ b/test/javascripts/acceptance/composer-edit-conflict-test.js.es6
@@ -1,17 +1,16 @@
 import { acceptance } from "helpers/qunit-helpers";
 
 acceptance("Composer - Edit conflict", {
-  loggedIn: true,
-
-  pretend(server, helper) {
-    server.put("/posts/18", () => {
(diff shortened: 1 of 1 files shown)


test/javascripts/acceptance/composer-actions-test.js.es6
```diff
@@ -324,3 +324,30 @@ QUnit.test(
     );
   }
 );
+
+QUnit.test(
+  "replying to post should not trigger an edit conflict",
+  async assert => {
+    // prettier-ignore
+    server.post("/draft.json", request => { // eslint-disable-line no-undef
+      assert.equal(request.requestBody.indexOf("originalText"), -1);
```
ZogStriP: There's something odd with this `assert`. 

The point of using `async/await` is to write asynchronous code like you would write synchronous code (ie. without callbacks).

Is there a way you could change this so that `assert` is at the end of the function?

nbianca: The problem is that I have to intercept the request and make sure it does include or does not include `originalText` which I do not see how it would be possible otherwise.

PR #6607, opened 2018-11-14
FIX box category style in emails

majakomel:
https://meta.discourse.org/t/digest-categories-labels-have-no-background-color-in-gmail/101446


ZogStriP:
Can you show before/after screenshots?

majakomel:
They are supposed to be the same, I just shuffled things around to remove absolute positioning which doesn't work in most email clients.

Before:
screen shot 2018-11-15 at 12 59 15

After:
screen shot 2018-11-15 at 13 17 34

PR #6608, opened 2018-11-14
FIX: makes sure an hex is always at 6 chars

jjaffeux:
eg: #aaa should be converted to #aaaaaa

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1325,6 +1325,15 @@ class Report
   end
 
   def rgba_color(hex, opacity = 1)
+    if hex.length === 3
+      chars = hex.scan(/\w/)
+      hex = chars.zip(chars).flatten.join.ljust(6)
+    end
+
+    if hex.length < 3
+      hex = hex.ljust(6, hex.last)
+    end
+
     rgbs = hex_to_rgbs(hex)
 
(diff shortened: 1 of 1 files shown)


app/models/report.rb
```diff
@@ -1325,6 +1325,15 @@ def hex_to_rgbs(hex_color)
   end
 
   def rgba_color(hex, opacity = 1)
+    if hex.length === 3
```
ZogStriP: Why not do?

```ruby
hex += hex if hex.size == 3
```

or 

```ruby
hex *= 2 if hex.size == 3
```

ZogStriP: Also, `===` does not mean what you think it means in Ruby ;) I guess too much JavaScript :P

jjaffeux: For #f49 for example the expected result is : `#ff4499`, you double each letter

```
hex = "f49"
chars = hex.scan(/\w/)
p chars.zip(chars).flatten.join.ljust(6) #  "ff4499"

hex = "f49"
p hex += hex # "f49f49"

hex = "f49"
p hex *= 2 # "f49f49"
```

ZogStriP: Right. Was thinking about the easy case "000". Cary on!

PR #6609, opened 2018-11-15
Stop seeding assets for site design topic.

tgxworld:
cc @SamSaffron @coding-horror  Any objections? We don't require this topic now that images can be directly uploaded into the site settings.


ZogStriP:
Love it 👍 

I wonder if there are places where we mention this topic? Like the starting guide or any other doc?

coding-horror:
Yes the seeded quick start guide staff topic needs to.be updated as well.

tgxworld:
Looks like that was already updated when the wizard came along in https://github.com/discourse/discourse/commit/0c4e4406a1bc23ea01cbee4109391dd89594d6f4

PR #6611, opened 2018-11-15
CSP: drop 'self' in `script-src`

xrav3nz:
This version does not repeatedly whitelist assets when using CDN, e.g.

```text
https://host.com/assets-that-do-not-use-cdn/
https://cdn.com/assets-use-cdn/
```

whereas https://github.com/discourse/discourse/pull/6611/commits/0e73bed72e96f1049460a27466a3fe4603eef354 is a looser version that whitelists them repeatedly, e.g.

```text
https://host.com/assets-that-do-not-use-cdn/
https://host.com/assets-use-cdn/
https://cdn.com/assets-that-do-not-use-cdn/
https://cdn.com/assets-use-cdn/
```

### DIFF
--- a/lib/content_security_policy.rb
+++ b/lib/content_security_policy.rb
@@ -60,17 +60,26 @@ class ContentSecurityPolicy
 
   attr_reader :request
 
-  SCRIPT_ASSET_DIRECTORIES = %w(
-    /assets/
-    /brotli_asset/
-    /extra-locales/
-    /highlight-js/
-    /javascripts/
-    /theme-javascripts/
-  )
+  SCRIPT_ASSET_DIRECTORIES = [
+    # [dir, can_use_s3_cdn, can_use_cdn]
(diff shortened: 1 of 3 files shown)


lib/content_security_policy.rb
```diff
@@ -57,14 +58,55 @@ def build
 
   private
 
+  attr_reader :request
+
+  SCRIPT_ASSET_DIRECTORIES = [
+    # [dir, can_use_s3_cdn, can_use_cdn]
+    ['/assets/',             true, true],
+    ['/brotli_asset/',       true, true],
+    ['/extra-locales/',      false, false],
```
xrav3nz: `extra-locales` JS bundle specifically does not use CDN https://github.com/discourse/discourse/commit/daad2291ba420b820e2b007a9944ed9bfb1ff9a0

PR #6617, opened 2018-11-16
DEV: removes _.find from codebase

jjaffeux:


jjaffeux:
@eviltrout @ZogStriP oh yes right, didn’t think of it, we already use find at so many places

PR #6620, opened 2018-11-16
DEV: removes _.include from codebase

jjaffeux:


eviltrout:
Oh yes, if it's polyfilled then great.

PR #6621, opened 2018-11-16
FIX: Do not show nested user cards.

nbianca:


app/assets/javascripts/discourse/mixins/card-contents-base.js.es6
```diff
@@ -31,6 +31,13 @@ export default Ember.Mixin.create({
       return false;
     }
 
+    // Don't show nested
+    if ($target.parents('.card-content').length) {
```
ZogStriP: This can probably be merged with the previous condition ("Don't show on mobile").

PR #6622, opened 2018-11-16
FIX: Fix broken theme field URLs.

nbianca:


lib/stylesheet/importer.rb
```diff
@@ -43,7 +43,7 @@ def self.register_import(name, &blk)
       theme&.all_theme_variables&.each do |field|
         if field.type_id == ThemeField.types[:theme_upload_var]
           if upload = field.upload
-            url = upload_cdn_path(upload.url)
+            url = upload.url
```
ZogStriP: Is that the only place where we use this method?

app/assets/javascripts/discourse/models/category.js.es6
```diff
@@ -47,9 +47,9 @@ const Category = RestModel.extend({
     return Discourse.getURL("/c/") + Category.slugFor(this);
   },
 
-  @computed("url")
-  fullSlug(url) {
-    return url.slice(3).replace("/", "-");
```
nbianca: This does not work well with subdirectory installations (i.e. `/forum/c/my-category` is transformed to `rum-c/my-category`).

Using `replace(/\//g, "-")` would partially fix (transform to `rum-c-my-category`), but it is still not very good.

ZogStriP: We have `Discourse.BaseUri` available so you should be able to do

```javascript
@computed
fullSlug() {
  return Category.slugFor(this).slice(Discourse.BaseUri.length + 3).replace(\\//g, "-");
}
```

lib/stylesheet/importer.rb
ZogStriP: Why are we removing the `cdn`? Was it unnecessary?

nbianca: It looks like `Upload.url` is the full CDN URL when a CDN is used or path to local file otherwise.

I do not see any point in using `upload_cdn_path`. I see it is only used in two more places, in `session_controller` for SSO card and profile background. I also have a feeling that it is not required in those places either, but I did not get around to see how it works.

app/assets/javascripts/discourse/models/category.js.es6
nbianca: That looked hacky imo, but I agree we should do it like that to be the same with the logic on the server-side.

ZogStriP: Why is this "hacky"? This is the same logic as [server-side](https://github.com/discourse/discourse/pull/6622/files#diff-af69b723432608c351fb1dd5db9e8348R538).

ZogStriP: > Well... `url == Discourse.getURL("/c/") + Category.slugFor(this)`

If that's **always** the case, then fine 👍 Might need a comment.



> There is also that `3` which is a magic constant (stands for `'/c/'.length`).

I agree, kill all the magic constants ;)

lib/global_path.rb
```diff
@@ -4,7 +4,7 @@ def path(p)
   end
 
   def cdn_path(p)
-    "#{GlobalSetting.cdn_url}#{path(p)}"
+    GlobalSetting.cdn_url&.empty? ? p : "#{GlobalSetting.cdn_url}#{path(p)}"
```
ZogStriP: Thanks to Rails's magic, this can be reduced to 

```suggestion
"#{GlobalSetting.cdn_url.presence}#{path(p)}"
```

PR #6623, opened 2018-11-16
do not overwrite hostname in development

xrav3nz:
https://github.com/discourse/discourse/blob/10723ba1c259a559ad085bb96109e6fe7ddfdac7/lib/content_security_policy.rb#L103

Otherwise, `port` information is lost and leads to incorrect CSP generated on `localhost`:

> Refused to load the script `http://localhost:3000/assets/locales/en.js` because it violates the following Content Security Policy directive: `"script-src 'unsafe-eval' localhost/logs/ localhost/sidekiq/ localhost/mini-profiler-resources/ localhost/assets/ localhost/brotli_asset/ localhost/extra-locales/ localhost/highlight-js/ localhost/javascripts/ localhost/theme-javascripts/"`.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -198,8 +198,10 @@ module Discourse
     # supports etags (post 1.7)
     config.middleware.delete Rack::ETag
 
-    require 'middleware/enforce_hostname'
-    config.middleware.insert_after Rack::MethodOverride, Middleware::EnforceHostname
+    unless Rails.env.development?
+      require 'middleware/enforce_hostname'
+      config.middleware.insert_after Rack::MethodOverride, Middleware::EnforceHostname
+    end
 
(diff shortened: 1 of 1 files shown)


jjaffeux:
Ah nice, I just noticed I was loosing my ports in my local env

jjaffeux:
@SamSaffron can we merge this please ? it seems harmless

SamSaffron:
hmmm ... I guess since you all use ports I can live with this ... for me local dev is `http://l.discourse` and `https://l.discourse` but I get this is a more ninja setup.

jjaffeux:
I do have this setup with no port too, but on macos it has been brraking regularly for the last years.. So I usually dont fix it until I actually need it 😁

PR #6632, opened 2018-11-20
FEATURE: Scroll-based logo on mobile 

hnb-ku:
This PR adds the ability to show the small logo and topic information in the header on mobile while the user is scrolling down.

Previous related discussions can be found here https://meta.discourse.org/t/smart-mobile-header/99846

Re-ordering the elements in the header-contents widget came up while working on this. So, this PR also re-orders those elements and converts the header to flexbox

### DIFF
--- a/app/assets/javascripts/discourse/components/discourse-topic.js.es6
+++ b/app/assets/javascripts/discourse/components/discourse-topic.js.es6
@@ -145,7 +145,7 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
     // conditions for showing topic title in the header for mobile
     if (
       this.site.mobileView &&
-      scrollDirection != "up" &&
+      scrollDirection !== "up" &&
       offset > this.dockAt
     ) {
       return true;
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/discourse/components/discourse-topic.js.es6
```diff
@@ -94,6 +97,18 @@ export default Ember.Component.extend(AddArchetypeClass, Scrolling, {
...
+    // setup mobile scroll logo
+    if (this.site.mobileView) {
+      this.appEvents.on("topic:scrolled", offset =>
+        this.mobileScrollGaurd(offset)
+      );
+      this.appEvents.on("header:show-topic", () => {
+        $("header.d-header").addClass("scroll-logo");
+      });
+      this.appEvents.on("header:hide-topic", () => {
+        $("header.d-header").removeClass("scroll-logo");
```
awesomerobot: Optional: If you add a new class when the title is hidden on scrollup, you could also animate back in the big logo/nav panel

app/assets/stylesheets/common/base/header.scss
```diff
@@ -195,3 +208,56 @@
...
+    overflow: hidden;
+    width: 100%;
+  }
+  h1 {
+    margin: 0 0 0.25em 0;
+    font-size: $font-up-3;
+    width: 100%;
+  }
+  .topic-header-extra {
+    display: inline-flex;
```
awesomerobot: We should force categories and subcategories onto the same line and truncate if they're too long... will probably need to add an extra wrapper to contain them?

PR #6635, opened 2018-11-20
add deprecation warnings and comments for plugin authors

fantasticfears:
See related post. I don't have time to finish the specs for now :(

### DIFF
--- a/lib/auth/auth_provider.rb
+++ b/lib/auth/auth_provider.rb
@@ -1,3 +1,5 @@
+require 'ostruct'
+
 class Auth::AuthProvider
   include ActiveModel::Serialization
 
@@ -6,9 +8,27 @@ class Auth::AuthProvider
   end
 
   def self.auth_attributes
-    [:pretty_name, :title, :message, :frame_width, :frame_height, :authenticator,
-     :pretty_name_setting, :title_setting, :enabled_setting, :full_screen_login, :full_screen_login_setting,
-     :custom_url]
+    [
+      :pretty_name, # displayed pretty name string, needs to be localized string.
+      :title,
+      :message,
+      :frame_width, # popup frame window size
+      :frame_height,
+      :authenticator, # authenticator class inheriented from lib/auth/authenticator.rb
+      :pretty_name_setting,
+      :title_setting,
+      :enabled_setting,
+      :full_screen_login,
+      :full_screen_login_setting,
+      :custom_url
+    ]
+  end
+
+  def self.deprecated_auth_attributes
+    [
+      OpenStruct.new(attribute: :enabled_setting, version: "2.1.0", drop_version: "2.x.0", comment: "Please use authenticator.enabled? instead."),
+      OpenStruct.new(attribute: :background_color, version: "2.1.0", drop_version: "2.x.0", comment: "Please use CSS for color and icons instead.")
+    ]
   end
 
   attr_accessor(*auth_attributes)
--- a/lib/plugin/instance.rb
+++ b/lib/plugin/instance.rb
@@ -488,6 +488,13 @@ class Plugin::Instance
   def auth_provider(opts)
     provider = Auth::AuthProvider.new
 
+    Auth::AuthProvider.deprecated_auth_attributes.each do |struct|
+      if opts.key?(struct.attribute)
+        message = "[auth_provider] :#{struct.attribute} is deprectated since Discourse #{struct.version} and will not be supported since #{struct.drop_version}. #{struct.comment}"
+        log_warn(message)
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


lib/plugin/instance.rb
```diff
@@ -615,18 +622,30 @@ def register_locales!
...
-          puts msg
-        end
+        log_error(msg)
       end
     end
   end
 
   private
 
+  # When booting/parsing locales from plugins, some actions are invoked before Rails logger.
```
SamSaffron: Feels to me like it makes more sense to sort out boot order here so we have access to a logger. 

If we have no chance here we should log to STDERR vs STDOUT for warnings and errors

lib/auth/auth_provider.rb
```diff
@@ -6,9 +8,27 @@ def initialize(params = {})
...
+      :enabled_setting,
+      :full_screen_login,
+      :full_screen_login_setting,
+      :custom_url
+    ]
+  end
+
+  def self.deprecated_auth_attributes
+    [
+      OpenStruct.new(attribute: :enabled_setting, version: "2.1.0", drop_version: "2.x.0", comment: "Please use authenticator.enabled? instead."),
```
SamSaffron: I am really not a fan of OpenStruct, do you mind using a simple Hash here?

SamSaffron:
@davidtaylorhq any feedback here?

lib/plugin/instance.rb
fantasticfears: That make sense. This is actually a peak of iceberg. I think we should make this module much cleaner since it's plugin API. Some ideas:

1. We probably need a way to document where a plugin method is in Rails booting sequence.
2. And we put all our logic to somewhere else so API is clear for reading.
3. It might be worthy to to versionize API too.

I think the problem is that the module is really hard to test due to weird into booting sequence.

fantasticfears:
There was a discussion on Meta. I'm not going to continue with this PR now. @davidtaylorhq is working on auth providers so he will take over.

lib/plugin/instance.rb
davidtaylorhq: > Feels to me like it makes more sense to sort out boot order here so we have access to a logger.

This is quite tricky... all the auth providers need to be registered in time for the omniauth initializer. Therefore auth providers are registered super early on, at the `activate!` stage:
 https://github.com/discourse/discourse/blob/master/config/application.rb#L245

The plugin `after_initialize` hook is too late.

We can add a specific hook for 'pre-omniauth', or we could do something like this so that `Discourse.deprecate` saves its warning for when the application has initialized:

```
@@ -613,7 +616,11 @@ module Discourse
     redis_key = "deprecate-notice-#{digest}"
 
     if !$redis.without_namespace.get(redis_key)
-      Rails.logger.warn(warning)
+      if Rails.initialized?
+        Rails.logger.warn(warning)
+      else
+        Rails.application.config.after_initialize { Rails.logger.warn(warning) }
+      end
       $redis.without_namespace.setex(redis_key, 3600, "x")
     end
```
^^ This option would allow us to use Discourse.deprecate in places where the app hasn't yet loaded, which could be useful. At the same time, it feels kinda hacky...

What do you think @SamSaffron?

fantasticfears: > This is quite tricky... all the auth providers need to be registered in time for the omniauth initializer. Therefore auth providers are registered super early on, at the activate! stage:
Can we move activation to a initializer? We have logger there.

1. Hopefully it doesn't mess up seed/assets managing
2. Putting it before `009-omniauth.rb` should fix the problem

davidtaylorhq: Yep, we can definitely do that. It could even be done inside the 009-omniauth initializer. I was trying to decide whether to 'fix' `Discourse.deprecate`, or work around it by adding this extra hook to the initializer 🤔 

Now that a few hours have passed I think the initializer is the better option, as you said @fantasticfears. I'm sure I'll change my mind again by tomorrow morning 😆

fantasticfears: Well, this approach comes a lot of work because plugin hooks into every aspects of Discourse. When I started to look at auth module, I didn't know the existence of `Discourse.deprecate`. The good thing is that it's only used in `active_record_base.rb` and `sql_builder.rb`.

Happy hacking :P You have the power to design.

davidtaylorhq:
I've implemented this functionality in https://github.com/discourse/discourse/commit/4e010382ccb623597875b02fb271f2de469318cc

- Exception if you try to register an auth provider too early
- Deprecation notices for `background_color` and `enabled_setting`

Thanks for the suggestions @fantasticfears, this will definitely help plugin developers figure out why stuff isn't working!

I will add some documentation in comments next week.

PR #6637, opened 2018-11-20
narrative bot: add comma in fortune answer

caugner:
Source: https://www.transifex.com/discourse/discourse-org/translate/#de/$/113764397


caugner:
Looks like @discoursedangerbot accidentally identified my change to the *English* locale as a *non-English* translation update.

PR #6638, opened 2018-11-20
Add functionality required by discourse-encrypt

udan11:
These patches add or improve functionality exposed to plugins (especially to discourse-encrypt 💃 ).

- Added a plugin outlet to add top-side notices (such as "You did not activate encryption on this device...").
- Added `data-topic-id` on few elements containing topic titles (plugin needs to know what content belongs to which topic).
- Added `composer-action-after` to add an indicator whether the message is encrypted or not. There is another plugin outlet nearby (`composer-fields`), but that forces me to add a whole line just to show an icon.
- Modals are notified when closed. It is basically the opposite of `onShow`. Hmm... maybe I should have called it `onHide`? 🤔

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-action-title.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-action-title.js.es6
@@ -66,9 +66,9 @@ export default Ember.Component.extend({
   },
 
   _formatReplyToTopic(link) {
-    return `<a class="topic-link" href="${link.href}">${
-      link.anchor
-    }</a>`.htmlSafe();
+    return `<a class="topic-link" href="${link.href}" data-topic-id="${this.get(
+      "model.topic.id"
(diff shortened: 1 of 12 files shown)


SamSaffron:
Overall I am fine with these changes, can you sort out the prettier offense? 

I wonder though why zero tests fail here when we made all these changes to html ? are we missing some tests?

udan11:
@SamSaffron I do not believe that is always the case. Maybe tests account for structure (i.e. `element.class-name` contains a certain text) and not an exact match of the HTML?

SamSaffron:
Odd, 3 JS test failures, I wonder if this is heisnetests or something new, can you check?

SamSaffron:
failure looks related to discourse-assign unrelated to this change, thanks @udan11 !!

PR #6640, opened 2018-11-21
PERF: Move mention lookups out of the V8 context.

tgxworld:
* Also make it such that mention lookups only does
  a single DB query per post.


app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
```diff
@@ -1,32 +1,12 @@
 function addMention(buffer, matches, state) {
   let username = matches[1] || matches[2];
-  let { getURL, mentionLookup, formatUsername } = state.md.options.discourse;
-
-  let type = mentionLookup && mentionLookup(username);
-
-  let tag = "a";
+  let tag = "span";
```
jjaffeux:
```suggestion
  const tag = "span";
```

lib/pretty_text/shims.js
```diff
@@ -97,5 +106,7 @@ function __getCurrentUser(userId) {
 }
 
 I18n = {
-  t: function(a,b) { return __helpers.t(a,b); }
+  t: function(a, b) {
```
jjaffeux:
```suggestion
  t: (a, b) => {
```

tgxworld:
@SamSaffron  Can you review this? Thank you!

PR #6641, opened 2018-11-21
FIX notifications for flag PMs

majakomel:
Meta: https://meta.discourse.org/t/not-seeing-replies-to-report-pms/101844

Similar case as https://github.com/discourse/discourse/pull/6108 - sidekiq job runs before transaction is commited


ZogStriP:
What about the `feature_users_in_the_topic`? Should we run it in the `after_commit` as well?

PR #6642, opened 2018-11-21
DEV: adds test for image-short-url

jjaffeux:


SamSaffron:
Looks good to me! thanks

PR #6645, opened 2018-11-21
Extract inline JS on embedded comments

xrav3nz:
I didn't think CSP would work within an iframe... Testing proved me wrong today.

Tested locally, specifically the JavaScript that supports scrolling to a post after clicking on the post number.

### DIFF
--- a/app/views/embed/loading.html.erb
+++ b/app/views/embed/loading.html.erb
@@ -1,12 +1,9 @@
+<% content_for :head do %>
+  <meta http-equiv="refresh" content="30">
+<% end %>
+
 <header class='discourse' data-embed-state='loading'>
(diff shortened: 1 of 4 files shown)


app/views/embed/loading.html.erb
```diff
@@ -1,12 +1,9 @@
+<% content_for :head do %>
+  <meta http-equiv="refresh" content="30">
```
xrav3nz: The unit here is seconds, while `setTimeout` takes milliseconds. Thus 30_000 => 30.

SamSaffron:
@eviltrout can you review, looks safe to me. the meta tag for the refresh is my only concern here

PR #6646, opened 2018-11-21
FEAT: Fix and improve migrate_to_s3 rake task

brianhelba:
See individual commits for atomic changes:
* Ensure that upload migration to S3 works without a CDN enabled
  * An attempt at optimization in https://github.com/discourse/discourse/commit/7290145641e855c01a03ed505175cda79db3eb12 resulted in the migration task failing entirely when a CDN was not configured for use. This reverts part of https://github.com/discourse/discourse/commit/7290145641e855c01a03ed505175cda79db3eb12 , ensuring it is more reliable in all cases.
* Fix variable naming in migrate_to_s3
  * It's still unclear why prioritizing the migration of non-image uploads first results in any net performance improvement. @tgxworld Can you provide any rationale for this? Otherwise, it may be best to simplify the code and remove the sorting entirely.
* Include remote files in "uploads:migrate_to_s3" task
  * This adds a new feature to the task, allowing migration of existing remote files (for example, when renaming the upload bucket). This should also work across S3-compliant services, allowing migration in and out of AWS to other S3-compliant providers.
* Remove local upload files after moving them to S3
  * Once locally stored uploads are successfully migrated to S3, they should be removed, so they don't continue to consume disk space as 'leaked' resources.
* Use MiniMime to determine S3 upload content types
  * MiniMime is used in other parts of Discourse to determine MIME type from file extensions, and should be used instead of a subprocess call to the `file` utility.

### DIFF
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -239,9 +239,9 @@ def migrate_to_s3
 
   # Migrate all uploads
   file_uploads = Upload.where.not(sha1: nil).where("url NOT LIKE '#{s3.absolute_base_url}%'")
-  image_uploads = file_uploads.where("lower(extension) NOT IN (?)", FileHelper.supported_images.to_a)
+  non_image_uploads = file_uploads.where("lower(extension) NOT IN (?)", FileHelper.supported_images.to_a)
 
-  [image_uploads, file_uploads].each do |uploads|
+  [non_image_uploads, file_uploads].each do |uploads|
     uploads.find_in_batches(batch_size: 100) do |batch|
       batch.each do |upload|
         now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -227,6 +227,7 @@ def migrate_to_s3
   # will throw an exception if the bucket is missing
   s3 = FileStore::S3Store.new
   local = FileStore::LocalStore.new
+  max_file_size = [SiteSetting.max_image_size_kb, SiteSetting.max_attachment_size_kb].max.kilobytes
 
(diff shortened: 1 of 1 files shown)


brianhelba:
@tgxworld Please take a look. Per my PR description, I'd also like to determine whether the non-image upload prioritization is worth keeping, so I'd appreciate any comments you could provide.

brianhelba:
For what it's worth, I've already successfully used this change myself to migrate uploads as part of renaming some S3 buckets. I hope to see this improvement upstreamed, to benefit everyone.

tgxworld:
@brianhelba Thank you for the PR and thank you so much for keeping each commit atomic.

> Ensure that upload migration to S3 works without a CDN enabled

Can you provide me with some example URLs so that I can understand the problem better?

> Can you provide any rationale for this? Otherwise, it may be best to simplify the code and remove the sorting entirely.

Once the `SiteSetting.enable_s3_uploads` has been enabled, non images upload will [return a 404](https://github.com/discourse/discourse/blob/a3f8ef89a64753fa646ac3a6d46f82f36a327595/app/controllers/uploads_controller.rb#L68) which is why we want to prioritize the migration for non images upload first.

 > Include remote files in "uploads:migrate_to_s3" task
Hmm I'm not sure if I see the need for this in the task. If the intention is to migrate between S3 buckets, wouldn't it be better to just sync the two buckets and then do a remap of the bucket name? 

> Remove local upload files after moving them to S3

This should be a separate task. We currently do not verify if the upload has been uploaded correctly to S3. If we nuke the local upload immediately, it'll leave us with no backups should the rake task go wrong.

ZogStriP:
Closing this because I re-worked the task in 5381096bfd4e208657f928601d3792a5f0d1d0c8

PR #6647, opened 2018-11-22
FEATURE: Show change name of user in staff logs

mrfinch:
https://meta.discourse.org/t/admins-changing-users-name-not-username-should-be-logged/99511

This is how it looks:
screenshot from 2018-11-22 06-27-03


tgxworld:
Thank you @mrfinch :hugs:

PR #6648, opened 2018-11-22
FEATURE: If PM email bounced for staged user then alert in whisper reply

vinothkannans:


lib/email/receiver.rb
```diff
@@ -985,6 +1008,19 @@ def create_post(options = {})
       end
 
       user = options.delete(:user)
+
+      if options[:bounce]
+        options[:raw] = "The message to #{user.email} bounced.
```
gschlager: This message should be translatable.

lib/email/receiver.rb
```diff
@@ -444,6 +446,17 @@ def trim_discourse_markers(reply)
     end
 
     def parse_from_field(mail)
+      if mail.bounced? || verp
+        final_recipient = mail.final_recipient
+        return extract_from_address_and_name(final_recipient) if final_recipient.is_a? String
```
gschlager: I like to use `Array.wrap`, because you don't need to distinguish between `String` and `Array`.

lib/email/receiver.rb
```diff
@@ -444,6 +446,17 @@ def trim_discourse_markers(reply)
...
+
+        if final_recipient.is_a? Array
+          final_recipient.each do |from|
```
gschlager: Why loop? It looks like it's always returning the first recipient.

lib/email/processor.rb
```diff
@@ -19,10 +19,10 @@ def process!
       rescue RateLimiter::LimitExceeded
         @retry_on_rate_limit ? Jobs.enqueue(:process_email, mail: @mail) : raise
       rescue Email::Receiver::BouncedEmailError => e
-        # never reply to bounced emails
-        log_email_process_failure(@mail, e)
-        set_incoming_email_rejection_message(@receiver.incoming_email, I18n.t("emails.incoming.errors.bounced_email_error"))
+        handle_bounce(e)
```
gschlager: I guess explicitly handling `BouncedEmailError` isn't necessary anymore?

spec/components/email/receiver_spec.rb
```diff
@@ -103,12 +103,39 @@ def process(email_name)
...
+          reply_key: "4f97315cc828096c9cb34c6f1a0d6fe8",
+          user: user,
+          post: post
+        )
+      end
+      
+      post = process(:bounced_email)
+      
+      expect(post.whisper?).to eq(true)
+      expect(post.raw).to include("The message to linux-admin@b-s-c.co.jp bounced. \n\n### Details\n\n```text\nYour email bounced\n```")
```
gschlager: I prefer using squiggly heredoc for multi-line strings. Makes it more readable.

spec/fixtures/emails/bounced_email.eml
```diff
@@ -2,7 +2,7 @@ Delivered-To: someguy@discourse.org
 Date: Thu,  7 Apr 2016 19:04:30 +0900 (JST)
 From: MAILER-DAEMON@b-s-c.co.jp (Mail Delivery System)
 Subject: Undelivered Mail Returned to Sender
-To: someguy@discourse.org
+To: To: reply+4f97315cc828096c9cb34c6f1a0d6fe8@bar.com
```
gschlager: Duplicate `To:`

lib/email/receiver.rb
```diff
@@ -109,7 +109,7 @@ def create_incoming_email
     end
 
     def process_internal
-      raise BouncedEmailError  if is_bounce?
+      raise BouncedEmailError if is_bounce? && !SiteSetting.enable_whispers?
```
gschlager: All those special cases for `is_bounced?` worry me. Wouldn't it be possible to create the whisper the first time we know it's a bounce? `is_bounced?` tries to lookup an `EmailLog` entry which should have all the information needed (`user_id` and `post_id`). Am I missing something?

lib/email/receiver.rb
vinothkannans: nope. I received an array of recipients in [corresponding rspec test](https://github.com/discourse/discourse/blob/c3bc11b6b9185b6373fde1e02623686ebde4feee/spec/components/email/receiver_spec.rb#L133).

lib/email/receiver.rb
vinothkannans: As per [this IF condition](https://github.com/discourse/discourse/blob/c3bc11b6b9185b6373fde1e02623686ebde4feee/lib/email/receiver.rb#L203) there is a chance to miss bounce_key & email_log. For now I will try to reduce special cases.

Yes, if it is always possible to find `EmailLog` then it's better to create whisper in first time.

vinothkannans:
@gschlager the changes are done.

lib/email/receiver.rb
```diff
@@ -195,24 +196,33 @@ def log_and_validate_user(user)
...
+    end
 
+    def handle_bounce
       @incoming_email.update_columns(is_bounce: true)
 
       if verp && (bounce_key = verp[/\+verp-(\h{32})@/, 1]) && (email_log = EmailLog.find_by(bounce_key: bounce_key))
         email_log.update_columns(bounced: true)
-        email = email_log.user.try(:email).presence
+        user = email_log.user
+        email = user.try(:email).presence
```
gschlager: I know you just copied this line, but it's worth switching to the Safe Navigation Operator in the process.

`email = user&.email.presence`

lib/email/receiver.rb
```diff
@@ -985,6 +1007,21 @@ def create_post(options = {})
       end
 
       user = options.delete(:user)
+
+      if options[:bounce]
+        options[:raw] = <<~MD
```
gschlager: This string still needs to be moved to `config/server.en.yml`.
Then use `options[:raw] = I18n.t("<key>", email: user.email, raw: options[:raw])`

lib/email/receiver.rb
```diff
@@ -157,17 +159,16 @@ def process_internal
...
-                     skip_validations: user.staged?)
+                     skip_validations: user.staged?,
+                     bounce: is_bounce?)
       else
         first_exception = nil
 
         destinations.each do |destination|
           begin
-            process_destination(destination, user, body, elided, hidden_reason_id)
+            return process_destination(destination, user, body, elided, hidden_reason_id)
```
gschlager: 👍

vinothkannans:
@gschlager now the whisper text is moved to server locale 👍

gschlager:
@vinothkannans LGTM. 👍
Feel free to merge when you are around to fix any problems in case we missed something.

PR #6649, opened 2018-11-22
FEATURE: Use translated name for 'your email has been authenticated by'

davidtaylorhq:
Right now we use the auth provider name from the url (e.g. meta.discourse.org/auth/**github**), and simply capitalise the first letter. To quote 2013 @SamSaffron "This feels a tad wrong" 😉.

This change means that we use the proper translated name, using the same method as the user preferences page.

For the case of GitHub, before:
screenshot 2018-11-22 at 14 35 23

After:
screenshot 2018-11-22 at 14 35 39

The server-side changes are very minimal, and there are no changes to the plugin API.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/create-account.js.es6
+++ b/app/assets/javascripts/discourse/controllers/create-account.js.es6
@@ -147,13 +147,13 @@ export default Ember.Controller.extend(
       );
     }.property("accountEmail", "authOptions.email", "authOptions.email_valid"),
 
-    authProviderDisplayName(provider) {
-      switch (provider) {
-        case "Google_oauth2":
-          return "Google";
-        default:
-          return provider;
-      }
+    authProviderDisplayName(providerName) {
+      const matchingProvider = findAll().find(provider => {
+        return provider.name === providerName;
+      });
+      return matchingProvider !== undefined
+        ? matchingProvider.get("prettyName")
+        : providerName;
     },
 
     prefillUsername: function() {
--- a/app/assets/javascripts/discourse/controllers/create-account.js.es6
+++ b/app/assets/javascripts/discourse/controllers/create-account.js.es6
@@ -151,7 +151,7 @@ export default Ember.Controller.extend(
       const matchingProvider = findAll().find(provider => {
         return provider.name === providerName;
       });
-      return matchingProvider !== undefined
+      return matchingProvider
         ? matchingProvider.get("prettyName")
         : providerName;
     },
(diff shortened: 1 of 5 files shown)


test/javascripts/controllers/create-account-test.js.es6
```diff
@@ -84,3 +84,19 @@ QUnit.test("passwordValidation", function(assert) {
   testInvalidPassword("porkchops", I18n.t("user.password.same_as_username"));
   testInvalidPassword("pork@chops.com", I18n.t("user.password.same_as_email"));
 });
+
+QUnit.test("authProviderDisplayName", function(assert) {
+  var subject = this.subject;
```
jjaffeux:
```suggestion
  const subject = this.subject;
```

test/javascripts/controllers/create-account-test.js.es6
```diff
@@ -84,3 +84,19 @@ QUnit.test("passwordValidation", function(assert) {
...
+  var controller = subject({ siteSettings: Discourse.SiteSettings });
```
jjaffeux:
```suggestion
  const controller = subject({ siteSettings: Discourse.SiteSettings });
```

app/assets/javascripts/discourse/controllers/create-account.js.es6
```diff
@@ -147,13 +147,13 @@ export default Ember.Controller.extend(
...
-        case "Google_oauth2":
-          return "Google";
-        default:
-          return provider;
-      }
+    authProviderDisplayName(providerName) {
+      const matchingProvider = findAll().find(provider => {
+        return provider.name === providerName;
+      });
+      return matchingProvider !== undefined
```
jjaffeux: undefined seems overkill here:

```js
matchingProvider ? matchingProvider.get("prettyName") : providerName;
```

test/javascripts/controllers/create-account-test.js.es6
```diff
@@ -84,3 +84,19 @@ QUnit.test("passwordValidation", function(assert) {
   testInvalidPassword("porkchops", I18n.t("user.password.same_as_username"));
   testInvalidPassword("pork@chops.com", I18n.t("user.password.same_as_email"));
 });
+
+QUnit.test("authProviderDisplayName", function(assert) {
+  const subject = this.subject;
```
jjaffeux: I feel like this line is not needed and we can just call this.subject at next line, given we don’t use it at other places

app/assets/javascripts/discourse/controllers/create-account.js.es6
davidtaylorhq: Done

test/javascripts/controllers/create-account-test.js.es6
davidtaylorhq: Done

PR #6656, opened 2018-11-22
FEATURE: Check if draft exists before starting a new one.

nbianca:


app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -775,12 +775,47 @@ export default Ember.Controller.extend({
...
+            }
+
+            // do not show abandon dialog if old draft is clean
+            const draft = JSON.parse(data.draft);
+            if (draft.reply === draft.originalText) {
+              data.draft = null;
+              return data;
+            }
+
+            return new Promise(resv => {
```
jjaffeux: Could we extract this in its own function with an explicit name, like `return this.confirmAbandon()` ?

jjaffeux:
Did you try to write a test for this ?

app/assets/javascripts/discourse/controllers/composer.js.es6
nbianca: Well, it does more than just look for an existent draft and the name of "confirmAbandon" would be misleading (it also gets a draft sequence number in case it is missing).

Initially, it was a different function, but @ZogStriP suggested that I could combine that code with the already existing call to `Draft.get`.

jjaffeux: Totally feel free to find a better name, that was just a suggestion based on my limited understanding of the code.

coding-horror:
is this ready or waiting for review?

nbianca:
I want to get a second opinion about the failing tests.

jjaffeux:
taking care of this today

jjaffeux:
@nbianca 

Having this at the top of your test should fix almost all tests:

```
acceptance("Composer", {
  loggedIn: true,
  pretend(server, helper) {
    server.get("/draft.json", () => {
      return helper.response({
        draft: null,
        draft_sequence: 42
      });
    });
  },
  settings: {
    enable_whispers: true
  }
});
```

Can you try this? and look at the remaining failing test? I think it's unrelated

jjaffeux:
Ok @coding-horror this is mostly done in https://github.com/discourse/discourse/pull/6755

It was more convulted to get right than expected.

I have to go now, will merge it first thing tomorrow morning.

PR #6661, opened 2018-11-23
Update client.en.yml

tobiaseigen:
I think this reads better as two sentences. Happy thanksgiving!


gschlager:
I guess this is fine @coding-horror?

coding-horror:
It'll break all translations, and I don't feel a punctuation change is worthy of that at the moment..

PR #6662, opened 2018-11-24
UX: category images have no sizes

vinothkannans:


app/serializers/category_upload_serializer.rb
```diff
@@ -1,3 +1,3 @@
 class CategoryUploadSerializer < ApplicationSerializer
-  attributes :id, :url
+  attributes :id, :url, :width, :height
```
tgxworld: There aren't any tests now but it'll be good to add one so that someone in the future don't remove the attributes without realizing that it is required on the client side.

PR #6663, opened 2018-11-24
FEAT: split topic_stat_sentence to topic_stat_sentence_week and topic…

mrfinch:
…_stat_sentence_month

this adds two new values for translating but also keeps backward compatibility till translations are not available
https://meta.discourse.org/t/incorrect-usage-of-week-month-year-units-in-a-translated-string/98931/2


gschlager:
Thanks! You are right, `topic_stat_sentence` is only used for `week` and `month`. However, your current solution is too complicated. I guess mostly because you tried adding backward compatibility. You don't have to do that. Adding the new translations for English is enough. The system will fallback to English when it doesn't find the new translations in another language.

Would you like to try again? I'd recommend the following steps:
1. Add `topic_stat_sentence_week` and `topic_stat_sentence_month` (with `one` and `other` sub-keys for pluralization) to `config/locales/client.en.yml` and remove the existing `topic_stat_sentence`.

2. Update the `I18n.t` call in `category-list.js.es6` so that it uses the new strings for week and month.

3. Make sure it works as expected. 🙂 

That should be it.

app/assets/javascripts/discourse/models/category-list.js.es6
```diff
@@ -52,10 +52,13 @@ CategoryList.reopenClass({
           const unit = I18n.t(statPeriod);
           if (stat > 0) {
             c.stat = `<span class="value">${stat}</span> / <span class="unit">${unit}</span>`;
-            c.statTitle = I18n.t("categories.topic_stat_sentence", {
-              count: stat,
-              unit: unit
+
+            const stringToTranslate =
+              "categories.topic_stat_sentence" + "_" + statPeriod;
```
gschlager: You can simplify this even more by using template literals:

```es6
            c.statTitle = I18n.t(
              `categories.topic_stat_sentence_${statPeriod}`,
              {
                count: stat
              }
            );
```

mrfinch:
@gschlager  I tested the new approach. 

I changed the language from discourse admin to `es`.  Since I have not added the translation for it, it didn't work. Is there any other way we change locale? I think it didn't work because the value of `I18n.defaultLocale` was set to `es`. Am I missing anything?

screenshot from 2018-11-27 13-39-08

gschlager:
The development environment behaves a little bit different than production. Instead of falling back to English it will show the missing translation key. I wanted to change that for a long time...

Anyway, you can always test it by adding translations to the language's locale file or in the UI (Admin -> Customize -> Site Text).

app/assets/javascripts/discourse/models/category-list.js.es6
gschlager: @mrfinch Can you change this to use a template literal? The rest of the PR looks fine.

mrfinch: @gschlager  done.

Thanks

PR #6665, opened 2018-11-24
FEATURE: allows html tooltips

jjaffeux:


jjaffeux:
Needed for local-dates refactoring

PR #6669, opened 2018-11-26
FEATURE: Add setting to enable inline emoji translation

nbianca:
There is a minor issue which prevents the autocomplete from showing up on just ":" (but works with ":s" for example).

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -410,12 +410,17 @@ export default Ember.Component.extend({
       },
 
       onKeyUp(text, cp) {
+        // Regular expressions used to extract emoji name from text.
+        // The space version requires a ' ' (space) before the emoji name
+        // (i.e. ' :smile'), while the other one does not and is used
+        // when enable_inline_emoji_translation is true.
+        const noSpaceColonEmoji = /(:(?!:).?[\w-]*:?(?!:)(?:t\d?)?:?) ?$/gi;
+        const spaceColonEmoji = /(?:^|[^a-z])(:(?!:).?[\w-]*:?(?!:)(?:t\d?)?:?) ?$/gi;
+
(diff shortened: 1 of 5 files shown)


SamSaffron:
we are going to need a couple of tests here :)

config/site_settings.yml
```diff
@@ -744,6 +744,14 @@ posting:
     default: 0
     locale_default:
       fr: 1
+  enable_inline_emoji_translation:
+    client: true
+    default: true
```
ZogStriP: It's the opposite. It should be `false` by default and enabled only for CJK locales.

app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -188,7 +195,7 @@ function applyEmoji(content, state, emojiUnicodeReplacer, enableShortcuts) {
 
   for (i = 0; i < content.length - 1; i++) {
     let offset = 0;
-    let emojiName = getEmojiName(content, i, state);
+    let emojiName = getEmojiName(content, i, state, inlineEmoji);
```
jjaffeux:
```suggestion
    const emojiName = getEmojiName(content, i, state, inlineEmoji);
```

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -410,9 +410,11 @@ export default Ember.Component.extend({
       },
 
       onKeyUp(text, cp) {
-        const matches = /(?:^|[^a-z])(:(?!:).?[\w-]*:?(?!:)(?:t\d?)?:?) ?$/gi.exec(
-          text.substring(0, cp)
-        );
+        const regex = self.siteSettings.enable_inline_emoji_translation
```
jjaffeux: Given we are essentially duplicating the regex her with a minor difference, could you try to have two variables with explicit names, and combine them when needed please?

PR #6670, opened 2018-11-26
DEV: add rake task to convert attachments to oneboxes

pmusaraj:
Useful especially after running imports with posts containing attachments (like bbPress).

### DIFF
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -767,3 +767,51 @@ task "uploads:recover" => :environment do
     end
   end
 end
+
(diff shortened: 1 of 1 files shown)


tgxworld:
I know alot of our rake tasks are not tested but it is my hope that we'll be able to start having tests for more rake tasks so that someone else can easily made changes without break things in the future.

SamSaffron:
Hmm ... maybe this is simply a utility for the import scripts, not sure if we want a rake task for it, I don't see us running this outside of import?

PR #6671, opened 2018-11-26
FIX: do not agree flags by default when deleting posts

techAPJ:
When bulk deleting posts we were agreeing with flags on the deleted posts. Now by default we will defer flags, unless the flag is mark as agreed explicitly.

### DIFF
--- a/app/controllers/admin/flags_controller.rb
+++ b/app/controllers/admin/flags_controller.rb
@@ -86,8 +86,7 @@ class Admin::FlagsController < Admin::AdminController
     restore_post = params[:action_on_post] == "restore"
 
     if delete_post
-      # PostDestroy calls PostAction.agree_flags!
-      destroy_post(post)
+      destroy_post(post, agree_flags: true)
     elsif restore_post
       PostAction.agree_flags!(post, current_user, delete_post)
       PostDestroyer.new(current_user, post).recover
@@ -138,12 +137,12 @@ class Admin::FlagsController < Admin::AdminController
 
   private
 
-  def destroy_post(post)
+  def destroy_post(post, agree_flags: false)
     if post.is_first_post?
       topic = Topic.find_by(id: post.topic_id)
       guardian.ensure_can_delete!(topic) if topic.present?
     end
 
-    PostDestroyer.new(current_user, post).destroy
+    PostDestroyer.new(current_user, post, agree_flags: agree_flags).destroy
   end
 end
--- a/lib/post_destroyer.rb
+++ b/lib/post_destroyer.rb
@@ -147,7 +147,11 @@ class PostDestroyer
       update_user_counts
       TopicUser.update_post_action_cache(post_id: @post.id)
       DB.after_commit do
-        agree_with_flags
+        if @opts[:agree_flags]
+          agree_with_flags
+        else
+          defer_flags
+        end
       end
     end
 
@@ -225,7 +229,7 @@ class PostDestroyer
     if @post.has_active_flag? && @user.id > 0 && @user.staff?
       Jobs.enqueue(
         :send_system_message,
-        user_id: @post.user.id,
+        user_id: @post.user_id,
         message_type: :flags_agreed_and_post_deleted,
         message_options: {
           url: @post.url,
@@ -237,8 +241,11 @@ class PostDestroyer
         }
       )
     end
-
     PostAction.agree_flags!(@post, @user, delete_post: true)
+  end
+
+  def defer_flags
+    PostAction.defer_flags!(@post, @user, delete_post: true)
   end
 
   def trash_user_actions
--- a/spec/components/post_destroyer_spec.rb
+++ b/spec/components/post_destroyer_spec.rb
@@ -632,13 +632,13 @@ describe PostDestroyer do
     let!(:flag) { PostAction.act(moderator, second_post, PostActionType.types[:off_topic]) }
 
     before do
-      SiteSetting.queue_jobs = false
+      Jobs::SendSystemMessage.clear
     end
 
     it "should delete public post actions and agree with flags" do
       second_post.expects(:update_flagged_posts_count)
 
-      PostDestroyer.new(moderator, second_post).destroy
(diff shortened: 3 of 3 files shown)


spec/components/post_destroyer_spec.rb
```diff
@@ -657,7 +657,7 @@
 
     it "should not send the flags_agreed_and_post_deleted message if it was deleted by system" do
       second_post.expects(:update_flagged_posts_count)
-      PostDestroyer.new(Discourse.system_user, second_post).destroy
+      PostDestroyer.new(Discourse.system_user, second_post, { agree_flags: true }).destroy
```
eviltrout: If it's the last argument in ruby, you don't need the `{` and `}` - I wonder why rubocop didn't catch this actually.

spec/components/post_destroyer_spec.rb
```diff
@@ -666,7 +666,7 @@
     it "should not send the flags_agreed_and_post_deleted message if it was deleted by author" do
       SiteSetting.delete_removed_posts_after = 0
       second_post.expects(:update_flagged_posts_count)
-      PostDestroyer.new(second_post.user, second_post).destroy
+      PostDestroyer.new(second_post.user, second_post, { agree_flags: true }).destroy
```
eviltrout: Same here, about braces.

spec/components/post_destroyer_spec.rb
```diff
@@ -676,7 +676,15 @@
       second_post.expects(:update_flagged_posts_count)
       PostAction.defer_flags!(second_post, moderator)
       second_post.reload
-      PostDestroyer.new(moderator, second_post).destroy
+      PostDestroyer.new(moderator, second_post, { agree_flags: true }).destroy
```
eviltrout: more braces

spec/components/post_destroyer_spec.rb
```diff
@@ -676,7 +676,15 @@
...
+      expect(
+        Topic.where(title: I18n.t('system_messages.flags_agreed_and_post_deleted.subject_template')).exists?
+      ).to eq(false)
+    end
+
+    it "should not send the flags_agreed_and_post_deleted message if agree_flags is false" do
+      second_post.expects(:update_flagged_posts_count)
+      PostDestroyer.new(moderator, second_post, { agree_flags: false }).destroy
```
eviltrout: last braces I think :)

app/controllers/admin/flags_controller.rb
```diff
@@ -86,8 +86,7 @@ def agree
     restore_post = params[:action_on_post] == "restore"
 
     if delete_post
-      # PostDestroy calls PostAction.agree_flags!
-      destroy_post(post)
+      destroy_post(post, true)
```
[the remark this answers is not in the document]
techAPJ: I made the required change. Thanks for the tip!

lib/post_destroyer.rb
```diff
@@ -221,24 +221,27 @@ def trash_public_post_actions
...
-          flag_reason: I18n.t(
-            "flag_reasons.#{@post.active_flags.last.post_action_type.name_key}",
-            locale: SiteSetting.default_locale,
-            base_path: Discourse.base_path
-          )
-        }
-      )
+  def handle_flags(agree = false)
+    if agree
+      if @post.has_active_flag? && @user.id > 0 && @user.staff?
```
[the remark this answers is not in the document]
techAPJ: Sounds good to me. Added on my longer term to-do list.

spec/components/post_destroyer_spec.rb
techAPJ: I am experimenting with different code editors for the moment which didn't had rubocop configured. I caught this (on manual rubocop run) after I created this PR.

lib/post_destroyer.rb
```diff
@@ -221,24 +221,27 @@ def trash_public_post_actions
...
-        message_options: {
-          url: @post.url,
-          flag_reason: I18n.t(
-            "flag_reasons.#{@post.active_flags.last.post_action_type.name_key}",
-            locale: SiteSetting.default_locale,
-            base_path: Discourse.base_path
-          )
-        }
-      )
+  def handle_flags(agree: false)
```
tgxworld: I personally think we should still keep `agree_with_flags` because the method name is clear and easy to reason about. What we can do here instead is to have two methods `agree_with_flags` and `defer_flags`.

spec/components/post_destroyer_spec.rb
```diff
@@ -676,7 +676,15 @@
...
       second_post.reload
-      PostDestroyer.new(moderator, second_post).destroy
+      PostDestroyer.new(moderator, second_post, agree_flags: true).destroy
+      expect(
+        Topic.where(title: I18n.t('system_messages.flags_agreed_and_post_deleted.subject_template')).exists?
+      ).to eq(false)
+    end
+
+    it "should not send the flags_agreed_and_post_deleted message if agree_flags is false" do
+      second_post.expects(:update_flagged_posts_count)
```
tgxworld: Are we able to set this up via fixtures somehow instead of mocking it?

spec/components/post_destroyer_spec.rb
```diff
@@ -676,7 +676,15 @@
       second_post.expects(:update_flagged_posts_count)
       PostAction.defer_flags!(second_post, moderator)
       second_post.reload
-      PostDestroyer.new(moderator, second_post).destroy
+      PostDestroyer.new(moderator, second_post, agree_flags: true).destroy
+      expect(
+        Topic.where(title: I18n.t('system_messages.flags_agreed_and_post_deleted.subject_template')).exists?
+      ).to eq(false)
```
tgxworld: This spec is no longer accurate because we queue jobs by default now. Since the job is queued and never executed, the topic is never created. We should update this to check for the jobs in the queue.

techAPJ:
Thanks for the review @tgxworld. Update the PR per your feedback.

Note: travis build failure is unrelated.

tgxworld:
LGTM :+1:

PR #6677, opened 2018-11-27
FIX: removes country from timezone indicator

jjaffeux:


plugins/discourse-local-dates/assets/javascripts/discourse-local-dates.js.no-module.es6
```diff
@@ -84,6 +84,15 @@
         .replace("Etc/", "")
         .split("/");
 
+    const _zoneWithoutPrefix = timezone => {
+      const parts = _formatTimezone(timezone);
+      if (parts[1]) {
```
ZogStriP: What about?

```suggestion
      return parts[1] || parts[0];
```

jjaffeux: yes better indeed

PR #6678, opened 2018-11-27
FEAT:  moving likes from deleted post to merged post

mrfinch:
https://meta.discourse.org/t/when-merging-posts-likes-of-the-posts-being-merged-do-not-add-up-to-the-final-post/48918/16

### DIFF
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -604,7 +604,7 @@ class PostRevisor
     @post_successfully_saved && !@topic_changes.errored?
   end
 
-  def revise_likes deleted_post_ids, acting_user
+  def revise_likes(deleted_post_ids, acting_user)
     return if deleted_post_ids.blank?
 
     already_liked_users = UserAction.where(target_post_id: @post.id, action_type: UserAction::LIKE).pluck(:user_id)
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -115,7 +115,7 @@ class PostRevisor
   # - bypass_bump: do not bump the topic, even if last post
   # - skip_validations: ask ActiveRecord to skip validations
   # - skip_revision: do not create a new PostRevision record
-  # - move_likes: moves likes from list of post_ids passed to given post
+  # - move_likes_from_post_ids: moves likes from list of deleted_post_ids to given post
   def revise!(editor, fields, opts = {})
     @editor = editor
     @fields = fields.with_indifferent_access
@@ -157,7 +157,7 @@ class PostRevisor
     old_raw = @post.raw
 
     Post.transaction do
-      revise_likes(opts[:move_likes], editor) if opts[:move_likes]
+      revise_likes(opts[:move_likes_from_post_ids], editor) if opts.has_key?(:move_likes_from_post_ids)
       revise_post
 
       yield if block_given?
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -1,5 +1,6 @@
 require "edit_rate_limiter"
 require 'post_locker'
+require 'onpdiff'
 
 class PostRevisor
 
@@ -114,6 +115,7 @@ class PostRevisor
   # - bypass_bump: do not bump the topic, even if last post
   # - skip_validations: ask ActiveRecord to skip validations
   # - skip_revision: do not create a new PostRevision record
+  # - move_likes: moves likes from list of post_ids passed to given post
   def revise!(editor, fields, opts = {})
     @editor = editor
     @fields = fields.with_indifferent_access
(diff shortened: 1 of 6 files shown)


lib/post_revisor.rb
```diff
@@ -601,4 +604,27 @@ def successfully_saved_post_and_topic
...
+
+    args = {}
+    args[:message] = "Merging posts"
+    args[:take_action] = true
+    args[:flag_topic] = true
+
+    deleted_posts = Post.where(id: deleted_post_ids)
+    new_users_to_like = User.where(id: new_user_ids_to_like)
+
+    new_users_to_like.each do |user|
```
SamSaffron: Is this right? aren't you calling `remove_act` here on some posts with no actions by that user? 

also dates on the post action are going to be skewed + users are not going to be notified of these new likes.

coding-horror:
Users should not be notified about post merge "likes" anyways...

lib/post_revisor.rb
mrfinch: I have pushed code with the new approach, have to test it still though.

mrfinch: @SamSaffron can you have a look now?

spec/fabricators/post_fabricator.rb
```diff
@@ -144,3 +144,12 @@
     incoming_email.user = post.user
   end
 end
+
+Fabricator(:post_with_likes, from: :post) do
```
SamSaffron: I would prefer just to keep this in post_revisor_spec cause its a bit of a magic fabricator and its better to keep the magic all in the file that is doing the testing im

lib/post_revisor.rb
```diff
@@ -601,4 +604,43 @@ def successfully_saved_post_and_topic
...
+    already_liked_users = PostAction.where(post_id: @post.id, post_action_type_id: PostActionType.types[:like]).pluck(:user_id)
+
+    args = {}
+    args[:message] = "Merging posts"
+    args[:take_action] = true
+    args[:flag_topic] = true
+
+    user_with_transferred_likes = Set.new
+    PostAction.where(post_id: deleted_post_ids, post_action_type_id: PostActionType.types[:like]).find_in_batches do |group|
+      group.each do |post_action|
```
SamSaffron: I would personally prefer 4 SQL statements here instead of the looping, there are some odd cases here in the current pattern where stuff will crash out like missing user action and so on. SQL avoids needing the extra gates and is generally faster.

spec/components/post_revisor_spec.rb
```diff
@@ -592,6 +592,66 @@
       end
     end
 
+    context "revise_likes" do
```
SamSaffron: I do like that this is in post revisor and that it is nicely tested!

lib/post_revisor.rb
mrfinch: Can you explain how you would remove the looping using SQL statements?

From what I understand what you mean is removing below and writing corresponding sql for it:
```
user_action_like = UserAction.where(target_post_id: deleted_post_id, user_id: post_action.user_id, action_type: UserAction::LIKE, target_topic_id: deleted_post.topic_id).first
user_action_like.update_column(:target_post_id, @post.id)
```
But how does it remove looping?

SamSaffron:
@mrfinch SQL is a bit fancier but in many ways it is a bit easier to reason about cause it ends up being more concise. 

You want to run a query that "updates all the post actions that are likes in the set of ids, where the post id = to the minimal post id in the set and the user is not in the set of users who acted on the post" 

Then you run "delete of all post actions that are like on the set of ids" 

and repeat for user actions.

mrfinch:
> You want to run a query that "updates all the post actions that are likes in the set of ids, where the post id = to the minimal post id in the set and the user is not in the set of users who acted on the post"

I think this will not work because of `the post id = to the minimal post id in the set`.
Let's say we have three posts with likes as below:
P1 - A
P2 - A, B, C
P3 - B, C, D
So if all are merged into P1, final should be P1 -> A,B,C,D.
But when we just take minimal post_id, then we will only move likes of P2 (assuming it has minimal id) and not move likes of P3 thus giving P1 -> A, B, C only.

I think SQL will be more complicated since we have to move likes only from post we are borrowing that likes. It can happen that we borrow one like from post A and other from post B so we have to delete corresponding actions for post A and post B only.

Let me know if I m missing anything here.

mrfinch:
Hi,
  Can someone help review this?

Thanks

mrfinch:
Bump :bowing_man:

SamSaffron:
I appreciate the attempt here a lot, but the looping here can very easily cause timeouts and so on when merging posts with lots of likes 

I am sadly putting this on the backburner for now, thanks for the attempt here @mrfinch  sorry

PR #6681, opened 2018-11-27
DEV: exposes getPluginApi

jjaffeux:
This allows to make this kind of check in plugin initializers:

```
    if (getPluginApi("0.8.26")) {
      withPluginApi("0.8.26", api => doNewThing(api, container));
    } else {
      withPluginApi("0.8.13", api => doOldThing(api, container));
    }
```


jjaffeux:
@eviltrout any better way to do this? Or anything you don’t like with this?

jjaffeux:
https://github.com/discourse/discourse-assign/pull/19

PR #6682, opened 2018-11-27
FIX: call onSelect plugin callbacks for noop rows

jjaffeux:


ZogStriP:
No tests?

jjaffeux:
@ZogStriP tests added, merging

PR #6687, opened 2018-11-27
FIX: more resilient whisper state between composer states

jjaffeux:


tgxworld:
Perhaps a controller test for the method?

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -194,9 +194,17 @@ export default Ember.Controller.extend({
 
   canUnlistTopic: Em.computed.and("model.creatingTopic", "isStaffUser"),
 
-  @computed("canWhisper", "model.whisper")
-  showWhisperToggle(canWhisper, isWhisper) {
-    return canWhisper && !isWhisper;
+  @computed("canWhisper", "model.post")
+  showWhisperToggle(canWhisper, repliedToPost) {
+    let replyingToWhisper = false;
```
ZogStriP: What about inlining the `if`? ;)

```suggestion
    const replyingToWhisper = repliedToPost && repliedToPost.get("post_type") === this.site.post_types.whisper;
```

PR #6690, opened 2018-11-28
FIX: throw error when link in reason for grant badge is an external link

mrfinch:
https://meta.discourse.org/t/adding-an-external-link-to-the-reason-field-badge-fails-silently/95847

### DIFF
--- a/app/controllers/user_badges_controller.rb
+++ b/app/controllers/user_badges_controller.rb
@@ -50,8 +50,7 @@ class UserBadgesController < ApplicationController
     user = fetch_user_from_params
 
     unless can_assign_badge_to_user?(user)
-      render json: failed_json, status: 403
-      return
+      return render json: failed_json, status: 403
     end
 
     badge = fetch_badge_from_params
@@ -59,8 +58,7 @@ class UserBadgesController < ApplicationController
 
     if params[:reason].present?
       unless is_badge_reason_valid? params[:reason]
-        render json: { failed: 'External link not allowed in reason' }, status: 400
-        return
+        return render json: { failed: I18n.t('invalid_grant_badge_reason_link') }, status: 400
       end
 
       path = begin
@@ -123,12 +121,6 @@ class UserBadgesController < ApplicationController
   end
 
   def is_badge_reason_valid?(reason)
-    uri = URI.parse(reason)
-
-    if uri && uri.scheme && uri.host
-      return Discourse.base_url == "#{uri.scheme}://#{uri.host}"
-    else
-      return false
-    end
+    return Discourse.route_for(reason)
   end
 end
--- a/app/controllers/user_badges_controller.rb
+++ b/app/controllers/user_badges_controller.rb
@@ -58,6 +58,11 @@ class UserBadgesController < ApplicationController
     post_id = nil
 
     if params[:reason].present?
+      unless is_badge_reason_valid? params[:reason]
+        render json: { failed: 'External link not allowed in reason' }, status: 400
+        return
+      end
+
       path = begin
         URI.parse(params[:reason]).path
       rescue URI::Error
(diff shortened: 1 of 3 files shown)


app/controllers/user_badges_controller.rb
```diff
@@ -116,4 +121,14 @@ def can_assign_badge_to_user?(user)
   def ensure_badges_enabled
     raise Discourse::NotFound unless SiteSetting.enable_badges?
   end
+
+  def is_badge_reason_valid?(reason)
+    uri = URI.parse(reason)
+
+    if uri && uri.scheme && uri.host
```
ZogStriP: The whole `if`/`else` can be replace by this one-liner ;)

```suggestion
uri && uri.scheme && uri.host && Discourse.base_url == "#{uri.scheme}://#{uri.host}"
```

app/controllers/user_badges_controller.rb
```diff
@@ -116,4 +121,14 @@ def can_assign_badge_to_user?(user)
   def ensure_badges_enabled
     raise Discourse::NotFound unless SiteSetting.enable_badges?
   end
+
+  def is_badge_reason_valid?(reason)
+    uri = URI.parse(reason)
```
ZogStriP: This might throw exceptions if `reason` is not a valid URI.

app/controllers/user_badges_controller.rb
```diff
@@ -58,6 +58,11 @@ def create
     post_id = nil
 
     if params[:reason].present?
+      unless is_badge_reason_valid? params[:reason]
+        render json: { failed: 'External link not allowed in reason' }, status: 400
```
ZogStriP: You can combine the `return` and the `render` ;)

```suggestion
        return render json: { failed: 'External link not allowed in reason' }, status: 400
```

ZogStriP: The text of the error should be localized (ie. use `I18n.t("name of the string")`)

app/controllers/user_badges_controller.rb
ZogStriP: Also, instead of checking that's the link is on the same host, we should check it's a working link to a post or a topic.

ZogStriP: Yes, we _almost_ do that in `TopicLink`. Here are the relevant lines

https://github.com/discourse/discourse/blob/851ef14096f0b6f3c6726151099483907f2c0805/app/models/topic_link.rb#L142-L158

app/controllers/user_badges_controller.rb
```diff
@@ -116,4 +119,8 @@ def can_assign_badge_to_user?(user)
   def ensure_badges_enabled
     raise Discourse::NotFound unless SiteSetting.enable_badges?
   end
+
+  def is_badge_reason_valid?(reason)
+    return Discourse.route_for(reason)
```
ZogStriP: In Ruby, you don't need the `return` as it's always returning the value of the last statement in a method/function.

You need to check that the route is one that points to a post or a topic. Not all routes in Discourse are valid for this badge ;)

app/controllers/user_badges_controller.rb
```diff
@@ -116,4 +119,9 @@ def can_assign_badge_to_user?(user)
   def ensure_badges_enabled
     raise Discourse::NotFound unless SiteSetting.enable_badges?
   end
+
+  def is_badge_reason_valid?(reason)
+    route = Discourse.route_for(reason)
+    route && (route[:controller] == 'posts' || route[:controller] == 'topics')
```
ZogStriP: 😍

spec/requests/user_badges_controller_spec.rb
```diff
@@ -143,6 +143,51 @@
...
+    it 'grants badge when valid post/topic link is given in reason' do
+      admin = Fabricate(:admin)
+      post_1 = create_post
+
+      sign_in(admin)
+
+      post "/user_badges.json", params: {
+        badge_id: badge.id,
+        username: user.username,
+        reason: Discourse.base_url + '/t/incorrect-usage-of-week-month-year-units-in-a-translated-string/98931'
```
ZogStriP: I would use the url of the topic of the post you've just created for the test otherwise the test will fail if we change the default title.

ZogStriP:
Thanks 👍

PR #6694, opened 2018-11-28
Suppress tar(1)'s output

dinomite:
The `tar` command in this test produces output when running; send that to /dev/null instead.


ZogStriP:
Thanks 👍

PR #6696, opened 2018-11-28
Authenticator Improvements

davidtaylorhq:
https://meta.discourse.org/t/future-social-authentication-improvements/94691

The ManagedAuthenticator is already tested and in-use for the discourse-openid-connect plugin. This PR moves it into core, and migrates the Facebook Authenticator to the new system. `lib/auth/facebook_authenticator.rb` goes from 167 lines to 25 lines. Once we migrate other providers, this should provide a significant reduction in duplicated logic. Long term it makes it easier for us to add new functionality (e.g. "last used at"). 

The only user-facing change is that data-explorer queries referencing facebook_user_infos will need to be updated. Once merged I will update all the queries I can find on Meta.

## DEV: Introduce Auth::ManagedAuthenticator
A generic implementation of Auth::Authenticator which stores data in the
new UserAssociatedAccount model. This should help significantly reduce the duplicated
logic across different auth providers.

## REFACTOR: Migrate FacebookAuthenticator to use ManagedAuthenticator
Changes to functionality
  - Removed syncing of user metadata including gender, location etc.
    These are no longer available to standard Facebook applications.
  - Removed the remote 'revoke' functionality. No other providers have
    it, and it does not appear to be standard practice in other apps.
  - The 'facebook_no_email' event is no longer logged. The system can
    cope fine with a missing email address.

Data is migrated to the new user_associated_accounts table.
facebook_user_infos can be dropped once we are confident the data has
been migrated successfully.


db/migrate/20181108115009_create_user_associated_accounts.rb
```diff
@@ -0,0 +1,18 @@
+class CreateUserAssociatedAccounts < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_associated_accounts do |t|
+      t.string :provider_name, null: false
+      t.string :provider_uid, null: false
+      t.integer :user_id, null: false
+      t.datetime :last_used, null: false, default: -> { "CURRENT_TIMESTAMP" }
+      t.jsonb :info, null: false, default: {}
```
SamSaffron: I guess we are embracing our inner mongo ... 

mini_sql supports this which is nice: 

```
[15] pry(main)> DB.query("select '{\"bar\" : 1}'::json").first.json["bar"]
=> 1
```

lib/auth/facebook_authenticator.rb
```diff
@@ -10,158 +10,17 @@ def enabled?
...
-    true
-  end
-
   def register_middleware(omniauth)
     omniauth.provider :facebook,
            setup: lambda { |env|
              strategy = env["omniauth.strategy"]
               strategy.options[:client_id] = SiteSetting.facebook_app_id
               strategy.options[:client_secret] = SiteSetting.facebook_app_secret
-              strategy.options[:info_fields] = 'gender,email,name,about,first_name,link,last_name,website,location'
```
SamSaffron: all the gender,location etc is stuff we got to kill cause facebook do not provide it anymore?

lib/auth/managed_authenticator.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    # Add the association to the database if it doesn't already exist
+    if association.nil? && result.user
+      association = create_association!(result.extra_data.merge(user: result.user))
+    end
+
+    # Update all the metadata in the association:
+    if association
+      association.update!(
```
SamSaffron: I am not sure, but maybe at some point we want an audit log of N last changes. depends on how hard it is to debug issues here going forward. I would not rush to add one preemptively though

SamSaffron:
When you feel ready to merge please merge, looks good to me, it is quite huge though very hard to tell if there is going to be a regression or not.

lib/auth/facebook_authenticator.rb
davidtaylorhq: Yes, they have [tightened security a lot recently](https://developers.facebook.com/docs/facebook-login/changelog). Most of these fields now require formal App Review, and even then they require an additional `scope` parameter to be sent. So I am confident that no Discourse sites are currently using this functionality.

db/migrate/20181108115009_create_user_associated_accounts.rb
davidtaylorhq: I am torn on using the JSON fields. On the one hand, it has no enforced schema, but on the other hand, each auth provider will be slightly different and this gives us the most flexibility.

I am pretty sure we will never want to do a "lookup" based on any of the info in the json fields, so I think it should be ok performance-wise. Activerecord also has pretty decent support for getting/setting the values.

lib/auth/managed_authenticator.rb
davidtaylorhq: :+1: Yeah I have some ideas for improvements here, which I will look at once all of the auth providers are switched across to the new system

fantasticfears:
@davidtaylorhq Can you share the migration script? I think there are a lot of auth plugins need their migration too. We can make that as a rake test.

davidtaylorhq:
What migration script? Right now only Facebook has been changed to the new table, and I have included the migration in `/db/migrate`

I don’t intend to introduce anything backwards incompatible, so plugins won’t NEED to use ManagedAuthenticator. Of course, they can choose to use the new pattern if they want.

fantasticfears:
> What migration script? Right now only Facebook has been changed to the new table, and I have included the migration in /db/migrate

It's like a lot of plugin uses PluginStore with `name_uid_#{uid}`. That might be okay to be migrated to new table by scripts. Does that make sense?

davidtaylorhq:
For sure, yes. The problem is we can't just implement this in core - it will require plugin developers to make some changes.

Taking the oauth2-basic plugin as an example, we will need to change the implementation [here](https://github.com/discourse/discourse-oauth2-basic/blob/master/plugin.rb#L24) to inherit from `ManagedAuthenticator`, and then add a migration within the plugin to move the data from PluginStoreRows to UserAssociatedAccount. I will be working through the official plugins slowly over the next few weeks. If you'd like to see one that's already implemented, take a look at [openid-connect](https://github.com/discourse/discourse-openid-connect/blob/master/plugin.rb#L9)

fantasticfears:
Yep, I understand plugin authors need to change. That's the right approach. But I'm looking forward to your migration script. Many has used `PluginStoreStore`(https://github.com/discourse/discourse-steam-login/blob/master/plugin.rb#L26). This should be easy to be migrated by a rake task.

davidtaylorhq:
:+1: I will let you know when I’ve migrated one of the plugins.

davidtaylorhq:
@fantasticfears please see https://meta.discourse.org/t/adding-a-new-authentication-method-to-discourse/106695 for some documentation. I will add a plugin migration example as soon as we migrate our first official plugin (once 2.2.0 stable is released).

PR #6697, opened 2018-11-28
FIX: Fix query selecting users not accepting PMs.

nbianca:


tgxworld:
Thanks @nbianca. Merged in https://github.com/discourse/discourse/commit/ddd260941e72b9db405bc521f1cf53d1e22a6851

PR #6698, opened 2018-11-28
FIX: Use CDN for logos and icons.

nbianca:


app/models/site_setting.rb
```diff
@@ -176,35 +178,43 @@ def self.Upload
   }.each { |client_setting| client_settings << client_setting }
 
   def self.site_logo_url
-    self.logo&.url || self.logo_url(warn: false)
+    upload = self.logo
+    upload ? UrlHelper.absolute(upload_cdn_path(upload.url)) : self.logo_url(warn: false)
```
tgxworld: `UrlHelper.absolute` might already be including the CDN I think.

https://github.com/discourse/discourse/blob/c388cacb81af619593e5d75cea679284c95e15f6/lib/url_helper.rb#L11

tgxworld:
I think we need to be careful here. Some spots on the server and client might are expecting the url to be that of `Upload#url` to be used as values for fields. The broken wizard spec is an example of such a case.

app/models/site_setting.rb
nbianca: * This pattern is used in multiple places.
* `UrlHelper.absolute` does nothing if the link is already absolute.

I also spoke with @tgxworld in private and we both agreed that this is fine.

ZogStriP:
@nbianca can you fix the failing specs?

app/models/site_setting.rb
tgxworld: I think we can extract what I did in https://github.com/discourse/discourse/blob/671469bcc7a190d63bf4283a44ee6477cd39f180/app/helpers/user_notifications_helper.rb#L21-L29 to replace the pattern used here.

PR #6700, opened 2018-11-28
FEATURE: Multisite support for S3 backup store

gschlager:
There are some specs missing and there's no migration yet. Still, @tgxworld can you take a quick look and tell me if I'm on the right path? I didn't have to make any big changes to `LocalBackupStore` in order to get the shared specs working, so I'm confident that the specs make sense.

The S3 backup store assumes that backups always are under the `/default` prefix even when there is no multisite. Is there a method to detect if there are multiple sites in case we don't want that behavior?

### DIFF
--- a/lib/backup_restore/backuper.rb
+++ b/lib/backup_restore/backuper.rb
@@ -83,7 +83,7 @@ module BackupRestore
       @timestamp = Time.now.strftime("%Y-%m-%d-%H%M%S")
       @tmp_directory = File.join(Rails.root, "tmp", "backups", @current_db, @timestamp)
       @dump_filename = File.join(@tmp_directory, BackupRestore::DUMP_FILE)
-      @archive_directory = BackupRestore::LocalBackupStore.base_directory(@current_db)
+      @archive_directory = BackupRestore::LocalBackupStore.base_directory(db: @current_db)
       filename = @filename_override || "#{SiteSetting.title.parameterize}-#{@timestamp}"
       @archive_basename = File.join(@archive_directory, "#{filename}-#{BackupRestore::VERSION_PREFIX}#{BackupRestore.current_version}")
 
--- a/lib/backup_restore/local_backup_store.rb
+++ b/lib/backup_restore/local_backup_store.rb
@@ -3,9 +3,11 @@ require_dependency "disk_space"
 
 module BackupRestore
   class LocalBackupStore < BackupStore
-    def self.base_directory(current_db = nil)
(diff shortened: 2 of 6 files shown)


tgxworld:
@gschlager What are the paths that will eventually be generated? 

> Is there a method to detect if there are multiple sites in case we don't want that behavior?

O you can use `Rails.configuration.multisite`

gschlager:
> What are the paths that will eventually be generated?

`<bucket_name_with_prefix>/<current_db>`

* Regular site without prefix: `my_backup_bucket/default`
* Regular site with prefix: `my_backup_bucket/some_prefix/default`
* Multisite without prefix: `my_backup_bucket/site_name`
* Multisite with prefix: `my_backup_bucket/some_prefix/site_name`

gschlager:
I've added an additional commit to store backups of the default site in the bucket without the `default` folder for now. This can easily be removed as soon as the migration is available.

tgxworld:
@gschlager I'm thinking if the file should be `https://s3-backup-bucket.s3.us-stubbed-1.amazonaws.com/backups/second/foo.tar.gz?X` instead of `https://s3-backup-bucket.s3.us-stubbed-1.amazonaws.com/second/foo.tar.gz?X`? What are your thoughts? I generally prefer to keep the path structure the same as the local store.

gschlager:
That's already possible. Admins can include a prefix in the `s3_backup_bucket`. Values like `s3-backup-bucket/backups` work.

I was thinking that it might not be unreasonable for some one to have `s3_backup_bucket` the same as `s3_upload_bucket`
That's a recipe for disaster and disallowing the same value for both buckets is on my todo list.

tgxworld:
Which is why I think we need to retain control of the directory structure for the backups. No matter what the bucket is, we’ll only put the backups in the folder that we create.

gschlager:
OK, so you _always_ want to append `backups` to the `s3_backup_bucket`? What if the bucket already contains `backups` as prefix? 😉
Essentially the `s3_backup_bucket` isn't necessarily needed anymore. Are we forcing the `uploads` prefix for uploads too?

tgxworld:
Yea for uploads we always put it in an uploads directory. I guess we’ll need to mention it in the documentation. Let me do a write up later because there were some changes Sam requested on our call today.

gschlager:
@tgxworld PR updated.

On regular sites, backups are always stored in the `s3_backup_bucket`. On multisite the prefix `/backups/<db_name>` is always appended to the bucket.

tgxworld:
LGTM :+1:  Thank you @gschlager

PR #6702, opened 2018-11-29
FIX: defer flags (only) when handling a flag and deleting replies

techAPJ:
When deleting a flagged post and all child replies if the reply has a flag it will be deferred instead of being agreed with. All the other behaviour will remain same.

### DIFF
--- a/app/assets/javascripts/admin/models/flagged-post.js.es6
+++ b/app/assets/javascripts/admin/models/flagged-post.js.es6
@@ -136,7 +136,7 @@ export default Post.extend({
             label: I18n.t("yes_value"),
             class: "btn-danger",
             callback() {
-              Post.deleteMany(replies.map(r => r.id))
+              Post.deleteMany(replies.map(r => r.id), { deferFlags: true })
                 .then(action)
                 .then(resolve)
                 .catch(error => {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/admin/models/flagged-post.js.es6
```diff
@@ -136,7 +136,7 @@ export default Post.extend({
             label: I18n.t("yes_value"),
             class: "btn-danger",
             callback() {
-              Post.deleteMany(replies.map(r => r.id))
+              Post.deleteMany(replies.map(r => r.id), true)
```
eviltrout: I believe I mentioned this on the previous PR, but dislike APIs that take boolean parameters like this. If you read this line, it's totally unclear what `true` means. 

Please replace this with an object like `{ deferFlags: true }`

techAPJ: I did remember that from last PR. I was looking for named parameters in JavaScript similar to what we have in Ruby and landed on [this post](http://2ality.com/2011/11/keyword-parameters.html).

> JavaScript does not have native support for named parameters like Python and many other languages.

I was not sure about the ideal using object literal in parameters but now that you have confirmed it, I'll go ahead and update the code to use it.

eviltrout: Thanks @techAPJ

techAPJ:
Updated the PR. Thanks for the review @eviltrout. ❤️

PR #6703, opened 2018-11-29
FIX: prevent minimum_required_tags on category being set to null

majakomel:
Fixes `NoMethodError (undefined method '>' for nil:NilClass) /var/www/discourse/lib/topic_creator.rb:161:in 'setup_tags'` from the logs

### DIFF
--- a/spec/requests/categories_controller_spec.rb
+++ b/spec/requests/categories_controller_spec.rb
@@ -91,6 +91,7 @@ describe CategoriesController do
 
       it "raises an exception when the text color is missing" do
         post "/categories.json", params: { name: "hello", color: "ff0" }
+        expect(response.status).to eq(400)
       end
 
       describe "failure" do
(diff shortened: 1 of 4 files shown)


davidtaylorhq:
Nice! This has been reported a couple of times now on meta [1](https://meta.discourse.org/t/email-in-topic-creator-bug/102658?u=david) [2](https://meta.discourse.org/t/internal-server-error-setting-tags-via-the-api/100137/8?u=david)

Can we add a NOT_NULL database constraint here to make absolutely sure the value cannot be null? If we specify a default (0) on the column then that will fix the issue for any sites that are already in a broken state.

spec/requests/categories_controller_spec.rb
```diff
@@ -91,6 +91,7 @@
 
       it "raises an exception when the text color is missing" do
         post "/categories.json", params: { name: "hello", color: "ff0" }
+        expect(response.status).to eq(400)
```
ZogStriP: 😲

PR #6704, opened 2018-11-29
FEATURE: allow plugins and themes to extend the default CSP

xrav3nz:
For plugins:

```rb
# plugin.rb
extend_content_security_policy(
  script_src: ['https://domain.com/script.js', 'https://your-cdn.com/'],
  style_src: ['https://domain.com/style.css']
)
```

For themes and components:

```yml
# settings.yml
extend_content_security_policy:
  type: list
  default: "script_src:https://domain.com/|style_src:https://domain.com"
```

### DIFF
--- a/lib/content_security_policy/extension.rb
+++ b/lib/content_security_policy/extension.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+class ContentSecurityPolicy
+  module Extension
+    extend self
+
+    def site_setting_extension
+      { script_src: SiteSetting.content_security_policy_script_src.split('|') }
+    end
+
+    def plugin_extensions
+      DiscoursePluginRegistry.csp_extensions
+    end
+
+    THEME_SETTING = 'extend_content_security_policy'
+
+    def theme_extensions
+      cache['theme_extensions'] ||= find_theme_extensions
+    end
+
+    def clear_theme_extensions_cache!
+      cache['theme_extensions'] = nil
+    end
+
+    private
+
+    def cache
+      @cache ||= DistributedCache.new('csp_extensions')
+    end
+
+    def find_theme_extensions
+      extensions = []
+
+      Theme.find_each do |theme|
+        theme.cached_settings.each do |setting, value|
+          extensions << build_theme_extension(value) if setting.to_s == THEME_SETTING
+        end
+      end
+
+      extensions
+    end
+
+    def build_theme_extension(raw)
+      {}.tap do |extension|
+        raw.split('|').each do |entry|
+          directive, source = entry.split(':', 2).map(&:strip)
+
+          extension[directive] ||= []
+          extension[directive] << source
+        end
+      end
+    end
+  end
+end
--- a/lib/content_security_policy/extension.rb
+++ b/lib/content_security_policy/extension.rb
@@ -8,7 +8,11 @@ class ContentSecurityPolicy
     end
 
     def plugin_extensions
-      DiscoursePluginRegistry.csp_extensions
+      [].tap do |extensions|
+        Discourse.plugins.each do |plugin|
+          extensions.concat(plugin.csp_extensions) if plugin.enabled?
+        end
+      end
     end
 
     THEME_SETTING = 'extend_content_security_policy'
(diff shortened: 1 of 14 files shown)


lib/content_security_policy/extension.rb
```diff
@@ -0,0 +1,53 @@
...
+    private
+
+    def cache
+      @cache ||= DistributedCache.new('csp_extensions')
+    end
+
+    def find_theme_extensions
+      extensions = []
+
+      Theme.find_each do |theme|
```
xrav3nz: I initially wanted to optimize this into something like 

```rb
Theme.join(:theme_settings).where(theme_settings: { name: THEME_SETTING })
```

But this won't work because we do not create a `theme_setting` entry for empty/non-overwritten values.

lib/content_security_policy/extension.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def plugin_extensions
+      DiscoursePluginRegistry.csp_extensions
+    end
+
+    THEME_SETTING = 'extend_content_security_policy'
+
+    def theme_extensions
+      cache['theme_extensions'] ||= find_theme_extensions
```
xrav3nz: I was wondering if I should cache the final combined CSP, but that means a lot more cache invalidation cases.

xrav3nz: I was wondering if I should cache the final combined CSP, but that means a lot more cache invalidation cases.

lib/content_security_policy/middleware.rb
```diff
@@ -0,0 +1,32 @@
...
+      @app = app
+    end
+
+    def call(env)
+      request = Rack::Request.new(env)
+      _, headers, _ = response = @app.call(env)
+
+      return response unless html_response?(headers)
+
+      ContentSecurityPolicy.base_url = request.host_with_port if Rails.env.development?
```
xrav3nz: A hack to make sure CSP works with `localhost:ports` :(

lib/content_security_policy/extension.rb
xrav3nz: I was also thinking if I should cache the final combined CSP instead, but that means a lot more cache invalidation cases.

xrav3nz:
Hmm where spec failure that I not getting locally. Could be related to the `base_url` hack for dev env. Will 👀  later.

```
expected: "http://test.localhost/csp_reports"
     got: "test.localhost/csp_reports"
```

spec/lib/content_security_policy_spec.rb
```diff
@@ -78,14 +79,50 @@
...
-      expect(script_srcs).to include('example.com')
-      expect(script_srcs).to include('another.com')
-      expect(script_srcs).to include("'unsafe-eval'")
-    end
+  it 'can be extended by plugins' do
+    plugin = Plugin::Instance.new(nil, "#{Rails.root}/spec/fixtures/plugins/csp_extension/plugin.rb")
+
+    plugin.activate!
+    Discourse.plugins << plugin
+    SiteSetting.load_settings("#{Rails.root}/spec/fixtures/plugins/csp_extension/config/settings.yml")
```
SamSaffron: Are you 100% sure state is not going to get leaked here between tests?

lib/content_security_policy/extension.rb
SamSaffron: I would not stress about this, it is cached

SamSaffron: regarding caching the entire CSP... its probably ok not to, but you can measure the impact... if less than say 0.5ms its probably not worth it.

spec/lib/content_security_policy_spec.rb
xrav3nz: Jinx ;) was just about to push the change

SamSaffron:
Overall change looks really nice to me. ❤️  Tiny bit of feedback (we all do this sometimes) but try to keep the big "I moved stuff around" commit in a separate commit, that we its clearer what happened to the middleware.

lib/content_security_policy/extension.rb
xrav3nz: Measured and it's consistently ~0.1ms when theme extensions are cached. Going to leave this as is 💪 Thanks

PR #6705, opened 2018-11-29
FEATURE: Use amazon s3 inventory to manage upload stats

vinothkannans:
TODO

- [ ] Fetch inventory .csv file to get upload stats.
- [ ] Update `uploads:missing` rake task with s3 inventory.
- [ ] Create rspec test cases.
- [ ] Add locale for inventory site settings.

### DIFF
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -8,6 +8,7 @@ class SiteSetting < ActiveRecord::Base
   validates_presence_of :data_type
 
   after_save do |site_setting|
+    SiteSetting.refresh!
     DiscourseEvent.trigger(:site_setting_saved, site_setting)
     true
   end
--- a/lib/s3_inventory.rb
+++ b/lib/s3_inventory.rb
@@ -1,0 +1,86 @@
+require "aws-sdk-s3"
+
+class S3Inventory
+
+  attr_reader :inventory_id, :s3_client, :source_bucket_name, :source_bucket_path, :destination_bucket_name, :destination_bucket_path
+
+  def initialize(s3_client, source_bucket_name, source_bucket_path, inventory_id = "discourse-uploads")
+    @s3_client = s3_client
+    @source_bucket_name = source_bucket_name
+    @source_bucket_path = source_bucket_path
+    @inventory_id = inventory_id
+  end
+
+  def update!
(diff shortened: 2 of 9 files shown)


lib/s3_inventory.rb
```diff
@@ -0,0 +1,187 @@
+require "aws-sdk-s3"
+require "csv"
+
+class S3Inventory
```
tgxworld: I think we need to test this class as much as possible. Otherwise, it is hard for the PR to be reviewed.

app/models/site_setting.rb
```diff
@@ -8,6 +8,7 @@ class SiteSetting < ActiveRecord::Base
   validates_presence_of :data_type
 
   after_save do |site_setting|
+    SiteSetting.refresh!
```
gschlager: Why is the refresh needed?

lib/site_settings/validations.rb
```diff
@@ -53,6 +53,10 @@ def validate_enable_s3_uploads(new_val)
     validate_error :s3_upload_bucket_is_required if new_val == "t" && SiteSetting.s3_upload_bucket.blank?
   end
 
+  def validate_enable_s3_inventory(new_val)
+    validate_error :s3_inventory_bucket_is_required if new_val == "t" && (SiteSetting.s3_inventory_bucket.blank? || SiteSetting.s3_upload_bucket.blank?)
+  end
+
```
gschlager: It's probably a good idea to add a validation to prevent bucket reuse.

https://github.com/discourse/discourse/blob/c928386c3ec6499a608363f855f8713bf4c4f2c9/lib/site_settings/validations.rb#L66-L72

app/models/site_setting.rb
vinothkannans: Without `refresh!` method if I call  like `SiteSetting.UPDATED_SETTING_NAME` then it's returning the old value. I think in my particular case I can use like this https://github.com/discourse/discourse/pull/6705/commits/1091a5b80c51acf45c81b1bd4afc534df523ee01

vinothkannans:
To make reviewing easier, I'm splitting this to two small PRs.

* S3 etag implementation is moved to PR https://github.com/discourse/discourse/pull/6795.

PR #6708, opened 2018-11-30
DEV: upgrades highlight.js from v9.12.0 to v9.13.1

jjaffeux:


SamSaffron:
Yay, hopefully the rake task is still working 👍

PR #6710, opened 2018-12-02
FIX: show generic title when quoting off-topic secure category posts

majakomel:


SamSaffron:
Thanks heaps, this is a great change

PR #6711, opened 2018-12-02
FIX: make staff_edit_locks_post work with download_remote_images_to_local

majakomel:


SamSaffron:
Thanks, this looks good to me!

PR #6712, opened 2018-12-02
FIX: don't steal focus when text in editor is replaced

majakomel:
https://meta.discourse.org/t/topic-title-input-should-not-lose-focus-after-image-upload-succeeds/103050


ZogStriP:
Excellent 👍

PR #6715, opened 2018-12-03
FIX: Require groups to be given when inviting to a restricted category.

tgxworld:


config/locales/server.en.yml
```diff
@@ -190,6 +190,7 @@ en:
     error: "There was an error uploading that file. Please try again later."
 
   topic_invite:
+    failed_to_invite: "The user cannot be invited into this topic without a group membership in either one of the following groups: %{group_names}."
```
SamSaffron: minor but maybe you only have one group, should this take count?

SamSaffron:
I am ok with this being merged, its quite a big change so it is hard to find flaws but the specs look right to me

danielhollas:
Just a question: I vaguely remember that I had issues when inviting people into a category restricted by TL, where I could not do it even though the group was automatically granting necessary TL level. O have no idea if this is relevant to this PR, let me know if you need more info...

PR #6717, opened 2018-12-03
DEV: Enable `Style/SingleLineMethods` and `Style/Semicolon` in Rubocop

davidtaylorhq:
Cops enabled and `rubocop --auto-correct` run.

Following comments on https://github.com/discourse/discourse/commit/71aaed272cb2e8155e771b8055b7eec975b8c66b

@tgxworld I know you are thinking about ruby formatting/linting at the moment. Is this change ok, or will it affect what you're planning?

### DIFF
--- a/lib/autospec/qunit_runner.rb
+++ b/lib/autospec/qunit_runner.rb
@@ -5,8 +5,12 @@ module Autospec
   class QunitRunner < BaseRunner
 
     WATCHERS = {}
-    def self.watch(pattern, &blk); WATCHERS[pattern] = blk; end
-    def watchers; WATCHERS; end
+    def self.watch(pattern, &blk)
+      WATCHERS[pattern] = blk
+    end
+    def watchers
+      WATCHERS
+    end
 
     # Discourse specific
     watch(%r{^app/assets/javascripts/discourse/(.+)\.js.es6$}) { |m| "test/javascripts/#{m[1]}-test.js.es6" }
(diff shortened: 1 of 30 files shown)


lib/autospec/qunit_runner.rb
```diff
@@ -5,17 +5,25 @@ module Autospec
   class QunitRunner < BaseRunner
 
     WATCHERS = {}
-    def self.watch(pattern, &blk); WATCHERS[pattern] = blk; end
-    def watchers; WATCHERS; end
+    def self.watch(pattern, &blk)
+      WATCHERS[pattern] = blk
+    end
```
ZogStriP: Can you add a newline here?

lib/autospec/qunit_runner.rb
```diff
@@ -5,17 +5,25 @@ module Autospec
...
     watch(%r{^app/assets/javascripts/discourse/(.+)\.js.es6$}) { |m| "test/javascripts/#{m[1]}-test.js.es6" }
     watch(%r{^app/assets/javascripts/admin/(.+)\.js.es6$})     { |m| "test/javascripts/admin/#{m[1]}-test.js.es6" }
     watch(%r{^test/javascripts/.+\.js.es6$})
 
     RELOADERS = Set.new
-    def self.reload(pattern); RELOADERS << pattern; end
-    def reloaders; RELOADERS; end
+    def self.reload(pattern)
+      RELOADERS << pattern
+    end
```
ZogStriP: Ditto ^^

ZogStriP:
So many JavaScript in our Ruby; 😉

lib/autospec/qunit_runner.rb
davidtaylorhq: So we should really enable `Layout/EmptyLineBetweenDefs` as well ;)

https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_layout.md#layoutemptylinebetweendefs

PR #6726, opened 2018-12-04
FEATURE: Filter topic and post web hook events by tags

vinothkannans:


spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -125,6 +125,33 @@
...
+      stub_request(:post, "https://meta.discourse.org/webhook_listener")
+        .with(body: "{\"topic\":{\"test\":\"some payload\"}}")
+        .to_return(body: 'OK', status: 200)
+
+      subject.execute(
+        web_hook_id: topic_hook.id,
+        event_type: 'topic',
+        tag_ids: topic.tags.pluck(:id),
+        payload: { test: "some payload" }.to_json
+      )
```
SamSaffron: we should add one more test here that confirms it does not raise any web hook when the tag does not match.

SamSaffron:
Except for 1 more test the change looks good to me... add the test and merge away 👍

PR #6729, opened 2018-12-05
FEATURE: Remove full quotes of direct replies.

nbianca:


davidtaylorhq:
Can we add an `edit_reason` when revising the post? Otherwise people could get confused why their post has been edited.

Maybe "Automatically removed quote of whole previous post"?

spec/services/username_changer_spec.rb
```diff
@@ -335,7 +335,7 @@ def create_post_and_change_username(args = {}, &block)
               Lorem ipsum
 
               [quote="foo, post:1, topic:#{quoted_post.topic.id}"]
-              quoted post
```
tgxworld: Hmm why was this spec changed?

lib/post_creator.rb
```diff
@@ -397,6 +399,27 @@ def unarchive_message
     end
   end
 
+  def remove_useless_quotes
```
tgxworld: I think this should be done in a post processor since we don't want more overhead when creating a post.

tgxworld: Also do we need this to be a public method?

lib/post_creator.rb
```diff
@@ -397,6 +399,27 @@ def unarchive_message
...
+    num_quotes = @post.raw.scan(/\[quote[^\]]*\]/).size
+    return if num_quotes != 1
+
+    prev = @post.topic.posts.order('post_number desc').where('post_number < ? AND not hidden AND posts.deleted_at IS NULL', @post.post_number).first
```
tgxworld: Should we check if `post_number` is 1?

config/locales/server.en.yml
```diff
@@ -274,6 +274,7 @@ en:
   not_accepting_pms: "Sorry, %{username} is not accepting messages at the moment."
   max_pm_recepients: "Sorry, you can send a message to maximum %{recipients_limit} recipients."
   pm_reached_recipients_limit: "Sorry, you can't have more than %{recipients_limit} recipients in a message."
+  remove_useless_quotes: "Automatically removed quote of whole previous post."
```
SamSaffron: I would not call it "useless", instead "remove_direct_reply_full_quotes` useless has a very bad taste.

lib/post_creator.rb
SamSaffron: maybe just `pluck` raw here is is more efficient `.limit(1).pluck(:raw).first` you don't really need any of the other post data

spec/services/username_changer_spec.rb
nbianca: It is a full quote of the previous post, which was unnecessary and thus will be removed.

lib/post_creator.rb
```diff
@@ -192,6 +192,8 @@ def create
       enqueue_jobs unless @opts[:skip_jobs]
       BadgeGranter.queue_badge_grant(Badge::Trigger::PostRevision, post: @post)
 
+      remove_useless_quotes
```
ZogStriP: Make sure you rename this method as well ;)

lib/cooked_post_processor.rb
```diff
@@ -85,6 +86,29 @@ def post_process_quotes
     end
   end
 
+  def removed_direct_reply_full_quotes
+    return if @post.post_number == 1
+
+    num_quotes = @post.raw.scan(/\[quote[^\]]*\]/).size
+    return if num_quotes != 1
+
+    prev = @post.topic.posts.order('post_number desc').where('post_number < ? AND not hidden AND posts.deleted_at IS NULL', @post.post_number).limit(1).pluck(:raw).first
```
ZogStriP: A few notes:
- `@post.topic.posts` will generate an unnecessary query.
- `posts.deleted_at IS NULL` is already checked by the default scope of `Post`.
- Also, we only want to check for `regular` posts.

Here's what I suggest instead:

```suggestion
    prev = Post.where("topic_id = ? AND post_number < ? AND post_type = ? AND NOT hidden", @post.topic_id, @post.post_number, Post.types[:regular]).limit(1).pluck(:raw).first
```

lib/cooked_post_processor.rb
```diff
@@ -85,6 +86,29 @@ def post_process_quotes
...
+    return if !prev
+
+    new_raw = @post.raw.gsub(/\[quote[^\]]*\]\s*#{Regexp.quote(prev.strip)}\s*\[\/quote\]/, '')
+    return if @post.raw == new_raw
+
+    revisor = PostRevisor.new(@post)
```
ZogStriP: Since we're not using `revisor`, we can combine the 2 lines

```ruby
PostRevisor.new(@post).revise!({ ... })
```

lib/cooked_post_processor.rb
```diff
@@ -85,6 +86,29 @@ def post_process_quotes
...
+  def removed_direct_reply_full_quotes
+    return if @post.post_number == 1
+
+    num_quotes = @post.raw.scan(/\[quote[^\]]*\]/).size
+    return if num_quotes != 1
+
+    prev = @post.topic.posts.order('post_number desc').where('post_number < ? AND not hidden AND posts.deleted_at IS NULL', @post.post_number).limit(1).pluck(:raw).first
+    return if !prev
+
+    new_raw = @post.raw.gsub(/\[quote[^\]]*\]\s*#{Regexp.quote(prev.strip)}\s*\[\/quote\]/, '')
```
ZogStriP: I know we add the `full=true` attributes on the quote when they're full quotes. I don't remember how we check it though (if we do it at all).

nbianca: `full=true` attribute is only sometimes added. I noticed this when quoting bigger posts (i.e. including images).

tgxworld:
I’m curious does this remove full quotes that are padded with extra spaces or special characters?

ZogStriP:
Extra spaces should be handled. Not sure what you mean by special characters?

tgxworld:
Hmm ah ignore the special characters as i just realised that they could be part of the text.

spec/components/cooked_post_processor_spec.rb
```diff
@@ -1144,4 +1144,31 @@
 
   end
 
+  context "remove direct reply full quote" do
+    let(:topic) { Fabricate(:topic) }
+
+    it 'works' do
+      post = Fabricate(:post, topic: topic, raw: "this is the first post")
+      hidden = Fabricate(:post, topic: topic, hidden: true, raw: "this is the second post")
```
ZogStriP: Can you add another non-regular post here to ensure the query is working?

PR #6732, opened 2018-12-05
FEATURE: show avatar flair on group, badges and directory pages

majakomel:
https://meta.discourse.org/t/avatar-flair-not-showing-up-on-groups-pages/103079

### DIFF
--- a/app/serializers/directory_item_serializer.rb
+++ b/app/serializers/directory_item_serializer.rb
@@ -1,9 +1,33 @@
 class DirectoryItemSerializer < ApplicationSerializer
 
+  class UserSerializer < UserNameSerializer
+    attributes :primary_group_name,
+               :primary_group_flair_url,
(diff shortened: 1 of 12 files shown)


app/serializers/directory_item_serializer.rb
```diff
@@ -1,9 +1,33 @@
 class DirectoryItemSerializer < ApplicationSerializer
 
+  class UserSerializer < UserNameSerializer
+    attributes :primary_group_name,
+               :primary_group_flair_url,
+               :primary_group_flair_bg_color,
+               :primary_group_flair_color
+
+    def primary_group_name
```
ZogStriP: We should not even serialize these fields when there's no `primary_group`.

You will have to add some `include_<>?` methods like this

```ruby
def primary_group_name
  object.primary_group.name
end

def include_primary_group_name?
  object&.primary_group.present?
end
```

PR #6733, opened 2018-12-05
FIX: Do not reset link counts upon rebake

davidtaylorhq:
This was an indentation mistake introduced in 44eba0b. Pretty understandable, considering we are indented 8 levels deep in this method. The first commit corrects the issue and adds a test, the second commit refactors the method to reduce complexity.

(Please use rebase & merge, not squash, so that we keep the two commits separate)

### DIFF
--- a/app/models/topic_link.rb
+++ b/app/models/topic_link.rb
@@ -223,34 +223,35 @@ SQL
           # If we can't find the route, no big deal
         end
       end
+    end
 
-      # Remove links that aren't there anymore
-      if added_urls.present?
+    # Remove links that aren't there anymore
+    if added_urls.present?
+      TopicLink.where(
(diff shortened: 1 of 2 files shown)


rboy1:
Quick question, so posts which have been rebaked before this patch was rolled out would have lost their link counts? Any way to retrieve them?

davidtaylorhq:
Unfortunately not, the records were erased from the database. The only possible way would be to extract the old values from an old database backup.

rboy1:
Ok. One more question. How long before this rolls out to your enterprise customers ?

davidtaylorhq:
Sorry for the delay in response @rboy1 - this fix has already been deployed across all of our hosting. If you notice any further issues please do let us know.

PR #6740, opened 2018-12-07
FIX: Validate poll arguments.

nbianca:


plugins/poll/lib/polls_validator.rb
```diff
@@ -96,5 +124,32 @@ def valid_multiple_choice_settings?(poll)
...
+    def valid_numbers?(poll)
+      return true if poll["type"] != "number"
+
+      valid = true
+
+      min = (poll["min"].presence || 1).to_i
+      max = (poll["max"].presence || 2_147_483_647).to_i
+      step = (poll["step"].presence || 1).to_i
+
+      if min <= 0
```
ZogStriP: I think `0` is ok here.

plugins/poll/lib/polls_validator.rb
```diff
@@ -96,5 +124,32 @@ def valid_multiple_choice_settings?(poll)
...
+        @post.errors.add(:base, "Min " + I18n.t("errors.messages.greater_than", count: 0))
+        valid = false
+      end
+
+      if max <= 0
```
ZogStriP: `max` should also be greater than `min`

plugins/poll/spec/lib/polls_validator_spec.rb
```diff
@@ -5,6 +5,21 @@
   subject { described_class.new(post) }
 
   describe "#validate_polls" do
+    it "ensures that polls have valid values" do
```
ZogStriP: We need to test that each arguments are properly validated and not just `type`.

plugins/poll/lib/polls_validator.rb
```diff
@@ -21,6 +23,32 @@ def validate_polls
 
     private
 
+    def valid_arguments?(poll)
+      valid = true
+
+      unless [nil, "regular", "multiple", "number"].include?(poll["type"])
```
ZogStriP: We try to avoid duplicating hardcoded values (and creating unnecessary arrays). What about this instead?

```ruby
if poll["type"].present? && !Poll.types.has_key?(poll["type"])
```

plugins/poll/lib/polls_validator.rb
```diff
@@ -96,5 +124,32 @@ def valid_multiple_choice_settings?(poll)
...
+        @post.errors.add(:base, "Min " + I18n.t("errors.messages.greater_than", count: 0))
+        valid = false
+      end
+
+      if max <= 0
+        @post.errors.add(:base, "Max " + I18n.t("errors.messages.greater_than", count: 0))
+        valid = false
+      end
+
+      if step <= 0
```
ZogStriP: We should also validate that the `step` value allows for at least 1 number to be generated.

plugins/poll/lib/polls_validator.rb
ZogStriP: What if `max > 2_147_483_647`?

plugins/poll/lib/polls_validator.rb
ZogStriP: What if `min > 2_147_483_647`?

plugins/poll/lib/polls_validator.rb
```diff
@@ -80,10 +106,10 @@ def valid_number_of_options?(poll)
     def valid_multiple_choice_settings?(poll)
       if poll["type"] == "multiple"
         options = poll["options"].size
-        min = (poll["min"].presence || 1).to_i
+        min = (poll["min"].presence || 0).to_i
```
ZogStriP: Since this is a multiple choice poll, we only allow at least 1 choice. So that was correct before.

plugins/poll/lib/polls_validator.rb
```diff
@@ -96,5 +122,45 @@ def valid_multiple_choice_settings?(poll)
 
       true
     end
+
+    def valid_numbers?(poll)
+      return true if poll["type"] != "number"
+
+      valid = true
+
+      min = (poll["min"].presence || 1).to_i
```
ZogStriP: Here, we should allow for `min >= 0`.

You can also shorten that to

```suggestion
      min = poll["min"].to_i
```

Since `nil.to_i == 0`

plugins/poll/lib/polls_validator.rb
```diff
@@ -96,5 +122,45 @@ def valid_multiple_choice_settings?(poll)
...
+      max = (poll["max"].presence || MAX_VALUE).to_i
+      step = (poll["step"].presence || 1).to_i
+
+      if min <= 0
```
ZogStriP: We should allow `min == 0` for number polls ;)

PR #6742, opened 2018-12-09
FEATURE: Adds site setting to let quotes on direct replies.

nbianca:


lib/cooked_post_processor.rb
```diff
@@ -32,7 +32,7 @@ def initialize(post, opts = {})
   def post_process(bypass_bump = false)
     DistributedMutex.synchronize("post_process_#{@post.id}") do
       DiscourseEvent.trigger(:before_post_process_cooked, @doc, @post)
-      removed_direct_reply_full_quotes
+      removed_direct_reply_full_quotes if SiteSetting.remove_full_quote
```
ZogStriP: The `SiteSetting.remove_full_quote` check should be done in the `removed_direct_reply_full_quotes` instead.

spec/components/cooked_post_processor_spec.rb
```diff
@@ -28,6 +28,20 @@
       expect(PostUpload.exists?(post: post, upload: upload)).to eq(true)
     end
 
+    it "calls removed_direct_reply_full_quotes if site setting is true" do
+      SiteSetting.remove_full_quote = true
+      cpp.expects(:removed_direct_reply_full_quotes).once
```
ZogStriP: We need to test the `raw` wasn't changed. Not that the method wasn't called.

spec/components/cooked_post_processor_spec.rb
```diff
@@ -1146,30 +1146,41 @@
...
-
+      CookedPostProcessor.new(reply).removed_direct_reply_full_quotes
       expect(topic.posts).to eq([post, hidden, small_action, reply])
       expect(reply.raw).to eq("and this is the third reply")
       expect(reply.revisions.count).to eq(1)
       expect(reply.revisions.first.modifications["raw"]).to eq([raw, reply.raw])
       expect(reply.revisions.first.modifications["edit_reason"][1]).to eq(I18n.t(:removed_direct_reply_full_quotes))
     end
 
+    it 'checks site setting' do
```
ZogStriP: This test deserves a better title. What about something like this?

```suggestion
    it "does nothing when 'remove_full_quote' is disabled" do
```

ZogStriP:
@nbianca can you fix the conflict?

nbianca:
@ZogStriP closing this because it was merged with #6759

PR #6745, opened 2018-12-09
FIX: Rescue errors in job.

nbianca:


plugins/poll/jobs/regular/close_poll.rb
```diff
@@ -3,7 +3,10 @@ module Jobs
   class ClosePoll < Jobs::Base
 
     def execute(args)
-      DiscoursePoll::Poll.toggle_status(args[:post_id], args[:poll_name], "closed", Discourse.system_user)
+      begin
+        DiscoursePoll::Poll.toggle_status(args[:post_id], args[:poll_name], "closed", Discourse.system_user)
+      rescue StandardError
```
SamSaffron: in general we are very against this kind of rescue, we need to know what went wrong so we can fix it ... simplest thing you can do is just confirm the poll and post and topic are all there prior to attempting to change it

ZogStriP:
@nbianca can you fix the conflict?

PR #6747, opened 2018-12-10
FIX: delete all posts without hijack - in batches

majakomel:


config/locales/client.en.yml
```diff
@@ -3737,6 +3737,8 @@ en:
         suspended_until: "(until %{until})"
         cant_suspend: "This user cannot be suspended."
         delete_all_posts: "Delete all posts"
+        delete_all_posts_progress: "Deleting posts... (<span id='posts-sum'>{{posts_sum}}</span> out of {{posts_total}}) <div class='spinner'></div>"
```
ZogStriP: Not a fan of UI elements in the translation. Is there a way we can avoid that?

app/assets/javascripts/admin/models/admin-user.js.es6
```diff
@@ -98,6 +98,7 @@ const AdminUser = Discourse.User.extend({
   },
 
   deleteAllPosts() {
+    let deleted_posts_sum = 0;
```
ZogStriP: This probably can just be called `deletedPosts` ;)

app/assets/javascripts/admin/models/admin-user.js.es6
```diff
@@ -114,13 +115,44 @@ const AdminUser = Discourse.User.extend({
...
+            posts_sum: deleted_posts_sum,
+            posts_total: user.get("post_count")
+          })
+        );
+      },
+      performDelete = () => {
+        return ajax(`/admin/users/${user.get("id")}/delete_all_posts`, {
+          type: "PUT"
+        })
+          .then(result => {
```
ZogStriP: You can extract `post_deleted` with `.then({ post_deleted } => { `

app/controllers/admin/users_controller.rb
```diff
@@ -46,12 +46,11 @@ def show
   end
 
   def delete_all_posts
```
ZogStriP: I wonder if we should also rename (or better deprecate) this methods since it's not deleting all posts anymore?

PR #6748, opened 2018-12-10
Add groups to current user serializer

majakomel:
https://meta.discourse.org/t/plugin-development-list-a-user-s-groups-without-making-a-request/103710

### DIFF
--- a/app/serializers/current_user_serializer.rb
+++ b/app/serializers/current_user_serializer.rb
@@ -45,6 +45,8 @@ class CurrentUserSerializer < BasicUserSerializer
              :top_category_ids,
              :hide_profile_and_presence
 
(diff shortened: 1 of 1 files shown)


majakomel:
@ZogStriP Yes, should be fine.

ZogStriP:
Ok, let's try it ;)

eviltrout:
I should note that `BasicGroupSerializer` is not very "basic":

https://github.com/discourse/discourse/blob/master/app/serializers/basic_group_serializer.rb

This adds a fair amount of data that I'm not sure we'll need in there.

ZogStriP:
Very good point @eviltrout. @majakomel, we only want group **names** in there. Not the whole shebang! 😉

majakomel:
Alright, will add names only

PR #6749, opened 2018-12-10
FEATURE: Activate users invited via email when invite is redeemed

gschlager:
Do not send an activation email to users invited via email. They already confirmed their email address by clicking the invite link. Users invited via link will need to confirm their email address before they can login.

https://meta.discourse.org/t/allow-disabling-activation-email-for-invited-users/88426

### DIFF
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -107,10 +107,19 @@ class Invite < ActiveRecord::Base
       invite = nil
     end
 
-    invite.update_columns(created_at: Time.zone.now, updated_at: Time.zone.now) if invite
+    if invite
+      invite.update_columns(
+        created_at: Time.zone.now,
+        updated_at: Time.zone.now,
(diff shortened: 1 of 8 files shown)


app/models/invite.rb
```diff
@@ -107,10 +107,21 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
...
-    invite.update_columns(created_at: Time.zone.now, updated_at: Time.zone.now) if invite
+    if invite
+      invite.update_columns(
+        created_at: Time.zone.now,
+        updated_at: Time.zone.now,
+        via_email: invite.via_email && send_email
+      )
+    end
 
     if !invite
```
ZogStriP: Shouldn't this be an `else`?

gschlager: Yes, it should 🙂

PR #6751, opened 2018-12-10
Only add group names to current user serializer

majakomel:


app/serializers/current_user_serializer.rb
```diff
@@ -43,9 +43,12 @@ class CurrentUserSerializer < BasicUserSerializer
              :link_posting_access,
              :external_id,
              :top_category_ids,
-             :hide_profile_and_presence
+             :hide_profile_and_presence,
+             :groups
 
-  has_many :groups, embed: :object, serializer: BasicGroupSerializer
+  def groups
+    object.groups.order(:id).pluck(:name)
```
ZogStriP: I don't think we need the `order`.

PR #6752, opened 2018-12-11
FIX: always show filtered site settings after navigation

xrav3nz:
This PR fixes a problem that, after filtering the site settings:

If you navigate away by, for example, going to the "Customize" tab, and come back to the "Settings" tab:

You end up in a weird state that the filter is set, but the site settings aren't filtered.

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
@@ -2,6 +2,8 @@ import debounce from "discourse/lib/debounce";
 
 export default Ember.Controller.extend({
   filter: null,
+  allSiteSettings: Ember.computed.alias("model"),
+  visible: null,
   onlyOverridden: false,
 
   filterContentNow(category) {
@@ -14,7 +16,7 @@ export default Ember.Controller.extend({
     }
 
     if ((!filter || 0 === filter.length) && !this.get("onlyOverridden")) {
-      this.set("model", this.get("allSiteSettings"));
+      this.set("visible", this.get("allSiteSettings"));
       this.transitionToRoute("adminSiteSettings");
       return;
     }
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/admin/routes/admin-site-settings.js.es6
```diff
@@ -10,9 +10,10 @@ export default Discourse.Route.extend({
...
   afterModel(siteSettings) {
-    this.controllerFor("adminSiteSettings").set(
-      "allSiteSettings",
-      siteSettings
-    );
+    const controller = this.controllerFor("adminSiteSettings");
+
+    if (!controller.get("visible")) {
+      controller.set("visible", siteSettings);
+    }
```
xrav3nz: I really want to move this to the constructor in `controllers/admin-site-settings.js.es6`, but then  `visible` wouldn't be initialized in time for `routes/admin-site-settings-index.js.es6`:

https://github.com/discourse/discourse/blob/e10236cb6863777c4eb96765b5fb70e1965afa3f/app/assets/javascripts/admin/routes/admin-site-settings-index.js.es6#L9

app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
```diff
@@ -2,6 +2,8 @@ import debounce from "discourse/lib/debounce";
 
 export default Ember.Controller.extend({
   filter: null,
+  allSiteSettings: Ember.computed.alias("model"),
+  visible: null,
```
ZogStriP: To me `visible` is a boolean, not a list of visible settings. Maybe call it `visibleSiteSettings`?

xrav3nz: 👍 done

SamSaffron:
yay, nice fix!

PR #6754, opened 2018-12-11
FEATURE: Allow connecting associated accounts when two-factor is enabled

davidtaylorhq:
Previously the 'reconnect' process was a bit magic - IF you were already logged into discourse, and followed the auth flow, your account would be reconnected and you would be 'logged in again'.

Now, we explicitly check for a `reconnect=true` parameter when the flow is started, store it in the session, and then only follow the reconnect logic if that variable is present. Setting this parameter also skips the 'logged in again' step, which means reconnect now works with 2fa enabled 😁

### DIFF
--- a/app/assets/javascripts/discourse/controllers/preferences/account.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/account.js.es6
@@ -243,7 +243,7 @@ export default Ember.Controller.extend(
       },
 
       connectAccount(method) {
-        method.doLogin();
+        method.doLogin(true);
       }
     }
   }
--- a/app/assets/javascripts/discourse/models/login-method.js.es6
+++ b/app/assets/javascripts/discourse/models/login-method.js.es6
@@ -24,7 +24,7 @@ const LoginMethod = Ember.Object.extend({
     );
   },
 
-  doLogin() {
+  doLogin(reconnect = false) {
     const name = this.get("name");
     const customLogin = this.get("customLogin");
 
@@ -32,6 +32,10 @@ const LoginMethod = Ember.Object.extend({
       customLogin();
     } else {
       let authUrl = this.get("custom_url") || Discourse.getURL("/auth/" + name);
+
+      if (reconnect) {
+        authUrl += "?reconnect=true";
+      }
 
       if (this.get("full_screen_login")) {
         document.cookie = "fsl=true";
@@ -45,7 +49,8 @@ const LoginMethod = Ember.Object.extend({
         const width = this.get("frame_width") || 800;
 
         if (name === "facebook") {
-          authUrl = authUrl + "?display=popup";
+          authUrl += authUrl.includes("?") ? "&" : "?";
+          authUrl += "display=popup";
         }
 
         const w = window.open(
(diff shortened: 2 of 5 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -25,8 +25,18 @@ def complete
     authenticator = self.class.find_authenticator(params[:provider])
     provider = DiscoursePluginRegistry.auth_providers.find { |p| p.name == params[:provider] }
 
-    if authenticator.can_connect_existing_user? && current_user
+    if session.delete(:auth_reconnect) && authenticator.can_connect_existing_user? && current_user
+      # If we're reconnecting, don't actually try and log the user in
       @auth_result = authenticator.after_authenticate(auth, existing_account: current_user)
+      if (provider && provider.full_screen_login) || cookies['fsl']
```
ZogStriP:
```suggestion
      if provider&.full_screen_login || cookies['fsl']
```

spec/requests/omniauth_callbacks_controller_spec.rb
```diff
@@ -338,6 +338,86 @@ def enabled?
...
+        )
+
+        Rails.application.env_config["omniauth.auth"] = OmniAuth.config.mock_auth[:google_oauth2]
+      end
+
+      it 'should not reconnect normally' do
+        # Log in normally
+        get "/auth/google_oauth2"
+        expect(response.status).to eq(302)
+        expect(session[:auth_reconnect]).to be_falsey
```
ZogStriP: I think we prefer to be explicit here (ie. `.to eq(nil)` or `.to eq(false)`)

app/controllers/users/omniauth_callbacks_controller.rb
davidtaylorhq: Done, and changed further down the file as well

spec/requests/omniauth_callbacks_controller_spec.rb
davidtaylorhq: Done

app/assets/javascripts/discourse/controllers/preferences/account.js.es6
```diff
@@ -243,7 +243,7 @@ export default Ember.Controller.extend(
       },
 
       connectAccount(method) {
-        method.doLogin();
+        method.doLogin(true);
```
eviltrout: I am not a fan of unnamed booleans like this. From reading this line of code it's not obvious what `true` does. Please change it to use an object literal like `method.doLogin({ reconnect: true })`

app/assets/javascripts/discourse/models/login-method.js.es6
```diff
@@ -45,7 +49,8 @@ const LoginMethod = Ember.Object.extend({
         const width = this.get("frame_width") || 800;
 
         if (name === "facebook") {
-          authUrl = authUrl + "?display=popup";
+          authUrl += authUrl.includes("?") ? "&" : "?";
```
eviltrout: It's a bit messy to look for the question mark here. If you take my suggestion above of using an object literal for parameters, you could call the argument to doLogin `queryParameters`. Facebook would set `queryParameters.display`. Finally you'd build the query parameters string from the object.

PR #6756, opened 2018-12-11
FEATURE: Lazily Load Images as they scroll into the viewport.

eviltrout:
This feature uses the Intersection Observer API

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

It should be compatible with all modern browsers. Non-Edge IE is *NOT*
supported, so in that particular browser images are loaded by default.

### DIFF
--- a/app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
+++ b/app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
@@ -1,0 +1,49 @@
+const OBSERVER_OPTIONS = {
+  rootMargin: "50%" // load images slightly before they're visible
+};
+
+// We hide an image by replacing it with a transparent gif
+function hide(image) {
+  image.classList.add("d-lazyload");
+  image.classList.add("d-lazyload-hidden");
+  image.setAttribute("data-src", image.getAttribute("src"));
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
```diff
@@ -0,0 +1,49 @@
...
+    "src",
+    "data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=="
+  );
+}
+
+// Restore an image from the `data-src` attribute
+function show(image) {
+  let dataSrc = image.getAttribute("data-src");
+  if (dataSrc) {
+    image.setAttribute("src", image.getAttribute("data-src"));
```
ZogStriP: Any reason you're calling `getAttribute` twice?

```suggestion
    image.setAttribute("src", dataSrc);
```

eviltrout: That's a mistake! Will fix, good catch.

SamSaffron:
This looks fantastic to me!

PR #6761, opened 2018-12-13
FIX:show popup before bulk invite

mrfinch:
https://meta.discourse.org/t/bulk-invitations-need-a-confirmation-dialog/19950/4

### DIFF
--- a/app/assets/javascripts/discourse/components/csv-uploader.js.es6
+++ b/app/assets/javascripts/discourse/components/csv-uploader.js.es6
@@ -28,11 +28,19 @@ export default Em.Component.extend(UploadMixin, {
   },
 
   uploadOptions() {
-    return {
-      autoUpload: false,
-      confirmationMessage: I18n.t(
-        "user.invited.bulk_invite.confirmation_message"
-      )
-    };
-  }
+    return { autoUpload: false };
+  },
+
+  _init: function() {
+    const $upload = this.$();
+
+    $upload.on("fileuploadadd", (e, data) => {
+      bootbox.confirm(
+        I18n.t("user.invited.bulk_invite.confirmation_message"),
+        I18n.t("cancel"),
+        I18n.t("go_ahead"),
+        result => (result ? data.submit() : data.abort())
+      );
+    });
+  }.on("didInsertElement")
 });
--- a/app/assets/javascripts/discourse/components/csv-uploader.js.es6
+++ b/app/assets/javascripts/discourse/components/csv-uploader.js.es6
@@ -25,5 +25,9 @@ export default Em.Component.extend(UploadMixin, {
 
   uploadDone() {
     bootbox.alert(I18n.t("user.invited.bulk_invite.success"));
+  },
+
+  uploadOptions() {
+    return { autoUpload: false };
   }
 });
--- a/app/assets/javascripts/discourse/components/csv-uploader.js.es6
+++ b/app/assets/javascripts/discourse/components/csv-uploader.js.es6
@@ -28,6 +28,12 @@ export default Em.Component.extend(UploadMixin, {
   },
 
   uploadOptions() {
-    return { autoUpload: false };
+    return {
+      autoUpload: false,
+      shouldConfirm: true,
+      confirmationMessage: I18n.t(
+        "user.invited.bulk_invite.confirmation_message"
+      )
+    };
   }
 });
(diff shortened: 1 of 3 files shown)


mrfinch:
This PR is ready for review :bowing_man:

mrfinch:
Bump :bowing_man:

app/assets/javascripts/discourse/mixins/upload.js.es6
```diff
@@ -74,6 +74,31 @@ export default Em.Mixin.create({
...
+      const options = this.uploadOptions();
+      if (options.shouldConfirm) {
+        const confirmationMessage = options.confirmationMessage
+          ? options.confirmationMessage
+          : I18n.t("generic_confirmation");
+        bootbox.confirm(
+          confirmationMessage,
+          I18n.t("cancel"),
+          I18n.t("go_ahead"),
+          result => {
```
ZogStriP: This can be shortened to

```javascript
result => result ? data.submit() : data.abort()
```

app/assets/javascripts/discourse/mixins/upload.js.es6
```diff
@@ -74,6 +74,31 @@ export default Em.Mixin.create({
       }
     });
 
+    // all pre-processing or confirmations come here before file is submitted to server
+    // set autoUpload: false in uploadOptions to allow this function to be called
+    $upload.on("fileuploadadd", (e, data) => {
+      const options = this.uploadOptions();
+      if (options.shouldConfirm) {
```
ZogStriP: I'm not sure we need this boolean. Isn't setting `autoUpload` to `false` enough?

ZogStriP:
Sorry for the delay @mrfinch. I've added a few comments ;)

app/assets/javascripts/discourse/mixins/upload.js.es6
mrfinch: My idea behind adding it was - let's say in future someone else wants to do some other processing or show something other than confirmation box, they can just add flags for it.

But I guess its ok for now since we don't have any other use-case. This can be added later.

config/locales/client.en.yml
```diff
@@ -213,6 +213,8 @@ en:
     submit: "Submit"
     generic_error: "Sorry, an error has occurred."
     generic_error_with_reason: "An error occurred: %{error}"
+    generic_confirmation: "Are you sure?"
```
gschlager: Not needed anymore

app/assets/javascripts/discourse/components/csv-uploader.js.es6
```diff
@@ -25,5 +25,30 @@ export default Em.Component.extend(UploadMixin, {
 
   uploadDone() {
     bootbox.alert(I18n.t("user.invited.bulk_invite.success"));
-  }
+  },
+
+  uploadOptions() {
+    return {
+      autoUpload: false,
+      confirmationMessage: I18n.t(
```
gschlager: Not needed anymore

app/assets/javascripts/discourse/components/csv-uploader.js.es6
```diff
@@ -25,5 +25,30 @@ export default Em.Component.extend(UploadMixin, {
...
+
+  _init: function() {
+    const $upload = this.$();
+
+    $upload.on("fileuploadadd", (e, data) => {
+      const confirmationMessage = I18n.t(
+        "user.invited.bulk_invite.confirmation_message"
+      );
+      bootbox.confirm(
+        confirmationMessage,
```
gschlager: You can replace the variable with the actual `I18n.t()` call

mrfinch: haha, I was already working on fixing them after I saw your comment that I can move `fileuploadadd` part to CSV-uploader.
I didn't know I can override it inside CSV-uploader, new to this framework :slightly_smiling_face: 

Thanks for suggesting, this approach is much cleaner.

app/assets/javascripts/discourse/mixins/upload.js.es6
gschlager: IHMO developing a generic solution is almost always a waste of time unless you already **know** about another use-case. It's highly likely that it's never needed in the future or that the generic solution doesn't work in some kind of unforseen use-case.

PR #6762, opened 2018-12-13
FIX: Strip remote url before import.

nbianca:


spec/requests/admin/themes_controller_spec.rb
```diff
@@ -76,6 +76,15 @@
       expect(theme.theme_fields.first.upload.original_filename).to eq(upload.original_filename)
     end
 
+    it 'can import a theme from Git' do
+      post "/admin/themes/import.json", params: {
+        remote: '    https://github.com/discourse/discourse-brand-header       ',
+        branch: ''
```
ZogStriP: Do you need that parameter for this test?

app/controllers/admin/themes_controller.rb
```diff
@@ -77,7 +77,7 @@ def import
     elsif params[:remote]
       begin
         branch = params[:branch] ? params[:branch] : nil
-        @theme = RemoteTheme.import_theme(params[:remote], current_user, private_key: params[:private_key], branch: branch)
+        @theme = RemoteTheme.import_theme(params[:remote].strip, current_user, private_key: params[:private_key], branch: branch)
```
ZogStriP: I think it might be better to do the `strip` in the `import_theme` method ;)

PR #6775, opened 2018-12-14
FEATURE: Improve backup stats on admin dashboard

gschlager:
* Dashboard doesn't timeout anymore when Amazon S3 is used for backups
* Storage stats are now a proper report with the same caching rules
* Changing the backup_location, s3_backup_bucket or creating and deleting backups removes the report from the cache
* It shows the number of backups and the backup location
* It shows the used space for the correct backup location instead of always showing used space on local storage
* It shows the date of the last backup as relative date

image
image


app/models/report.rb
```diff
@@ -1407,6 +1409,28 @@ def self.report_suspicious_logins(report)
     end
   end
 
+  def self.report_storage_stats(report)
+    backup_stats = begin
+      BackupRestore::BackupStore.create.stats
+    rescue BackupRestore::BackupStore::StorageError
```
tgxworld: :heart:

jjaffeux:
Awesome work @gschlager ❤️hope you didn’t have too much trouble doing it. With this, the dashboard becomes extremely resilient as it was almost the last thing which could crash!

gschlager:
You did great work with the admin dashboard. It's quite self-explanatory. 👍 
Converting it into a report wasn't that hard after I figured out how everything works. Refactoring everything else involved in calculating the disk space was definitely more work than the report itself.

> how does the UI behaves with 429 and 500 errors?

Like every other report I guess.

image image

PR #6778, opened 2018-12-15
FIX: use new key for delete topic to make it lowercased as all other …

mrfinch:
…buttons label around it

https://meta.discourse.org/t/delete-topic-tooltip-is-capitalized-but-nothing-else-is/73252/6


tgxworld:
Thanks @mrfinch

PR #6779, opened 2018-12-15
FEATURE: Allow wildcard in allowed_user_api_auth_redirects setting

pdavide:
### Use case
I need to consume the Discourse APIs using User API Keys from JS in some pages reachable in a different location than the domain root. Instead of listing all the locations in the `allowed_user_api_auth_redirects` (which will be impossibile if I generate pages dynamically), I can use a wildcard to allow redirect to multiple pages: i.e. `www.example.com/site/*`.

### DIFF
--- a/app/services/wildcard_url_checker.rb
+++ b/app/services/wildcard_url_checker.rb
@@ -1,8 +1,6 @@
 module WildcardUrlChecker
 
   def self.check_url(url, url_to_check)
-    puts "url: " + url
-    puts "url_to_check: " + url_to_check
     escaped_url = Regexp.escape(url).sub("\\*", '\S*')
     url_regex = Regexp.new("^#{escaped_url}$", 'i')
 
--- a/app/services/wildcard_url_checker.rb
+++ b/app/services/wildcard_url_checker.rb
@@ -1,0 +1,12 @@
+module WildcardUrlChecker
+
+  def self.check_url(url, url_to_check)
+    puts "url: " + url
+    puts "url_to_check: " + url_to_check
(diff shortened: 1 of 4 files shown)


app/services/wildcard_url_checker.rb
```diff
@@ -0,0 +1,12 @@
+module WildcardUrlChecker
+
+  def self.check_url(url, url_to_check)
+    puts "url: " + url
```
ZogStriP: Those `puts` should probably be removed ;)

pdavide: Ops! I forgot some debug code.

pdavide:
Hi @SamSaffron, resolved some conflicts caused by 19d7545318f8a42168261d5c09b8978d2ae63b90.

pdavide:
Hi @SamSaffron can I help more with this PR?

cc189:
nice feature

pdavide:
Thanks @cc189. Hope it will merged soon!

app/services/wildcard_url_checker.rb
```diff
@@ -0,0 +1,10 @@
+module WildcardUrlChecker
+
+  def self.check_url(url, url_to_check)
+    escaped_url = Regexp.escape(url).sub("\\*", '\S*')
```
ZogStriP: I think we want `gsub` here to support "`*.example.com*`"

```suggestion
    escaped_url = Regexp.escape(url).gsub("\\*", '\S*')
```

ZogStriP:
Sorry for the delay. Thanks @pdavide

PR #6780, opened 2018-12-15
Current user serializer groups (ef7f84b5 follow-up)

majakomel:


app/serializers/current_user_serializer.rb
```diff
@@ -38,14 +38,16 @@ class CurrentUserSerializer < BasicUserSerializer
...
              :link_posting_access,
              :external_id,
              :top_category_ids,
-             :hide_profile_and_presence
+             :hide_profile_and_presence,
+             :groups
 
-  has_many :groups, embed: :object, serializer: BasicGroupSerializer
+  def groups
+    object.groups.select { |group| Guardian.new(object).can_see_group?(group) }.pluck(:id, :name).map { |id, name| { id: id, name: name.downcase } }
```
SamSaffron: this pattern should probably be:

https://github.com/discourse/discourse/blob/d803dfc14ac7d27ac229c9dcad8997b1b74aaf99/app/controllers/groups_controller.rb#L49-L49

The select there is a bit odd cause it will query the object and then pluck is not saving much

I think this will work.


```
DB.query(Group.visible(object,nil).select(:id, :name).to_sql)
```

majakomel: Problem with `Group.visible_groups` is that it queries all groups that user can see, not only those they're a member of.

SamSaffron: aha ... we need an extra join then

app/models/user.rb
```diff
@@ -353,6 +353,10 @@ def group_granted_trust_level
       .maximum("groups.grant_trust_level")
   end
 
+  def visible_groups
+    groups.select { |group| guardian.can_see_group?(group) }
```
SamSaffron:
```suggestion
    groups.visible_groups(self)
```

I think this works, can you confirm... if passes all tests feel free to merge

PR #6781, opened 2018-12-16
FIX: hide emails on admin user list for moderators

majakomel:


app/serializers/admin_user_list_serializer.rb
```diff
@@ -38,8 +38,9 @@ class AdminUserListSerializer < BasicUserSerializer
 
   def include_email?
     # staff members can always see their email
-    (scope.is_staff? && object.id == scope.user.id) || scope.can_see_emails? ||
-      (scope.is_staff? && object.staged?)
+    (scope.is_staff? && object.id == scope.user.id) ||
```
ZogStriP: What about this?

```ruby
def include_email?
  (scope.is_staff? && (object.id == scope.user.id || object.staged?)) ||
  (scope.is_admin? && scope.can_see_email?)
end
```

PR #6782, opened 2018-12-16
FIX: Bypass draft check when switching to shared draft.

nbianca:


app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -786,7 +786,14 @@ export default Ember.Controller.extend({
       // or get a draft sequence number
       if (!opts.draft || opts.draftSequence === undefined) {
         return Draft.get(opts.draftKey)
-          .then(data => self.confirmDraftAbandon(data))
+          .then(data => {
+            if (opts.skipDraftCheck) {
```
ZogStriP: If we're going to skip the draft check, would it be possible to do the query to the server in the first place?

nbianca: We'd still need a draft sequence number which we are getting from the server.

PR #6783, opened 2018-12-16
FEATURE: Prohibit S3 bucket reusage

gschlager:
This validation makes sure that the `s3_upload_bucket` and the `s3_backup_bucket` have different values. The backup bucket is allowed to be a subfolder of the upload bucket. The other way around is forbidden because the backup system searches by prefix and would return all files stored within the backup bucket and its subfolders.

### DIFF
--- a/spec/lib/site_settings/validations_spec.rb
+++ b/spec/lib/site_settings/validations_spec.rb
@@ -1,0 +1,80 @@
+require 'rails_helper'
+require 'site_settings/validations'
+
+describe SiteSettings::Validations do
+  subject { Class.new.include(described_class).new }
+
+  context "s3 buckets reusage" do
+    let(:error_message) { I18n.t("errors.site_settings.s3_bucket_reused") }
+
+    shared_examples "s3 bucket validation" do
+      def change_bucket_value(value)
+        SiteSetting.public_send("#{other_setting_name}=", value)
+      end
+
+      it "shouldn't raise an error when both buckets are blank" do
+        change_bucket_value("")
+        expect { validate("") }.to_not raise_error
+      end
+
+      it "shouldn't raise an error when only one bucket is set" do
+        change_bucket_value("")
+        expect { validate("my-awesome-bucket") }.to_not raise_error
+      end
+
+      it "shouldn't raise an error when both buckets are equal, but use a different path" do
+        change_bucket_value("my-awesome-bucket/foo")
+        expect { validate("my-awesome-bucket/bar") }.to_not raise_error
+      end
+
+      it "should raise an error when both buckets are equal" do
+        change_bucket_value("my-awesome-bucket")
+        expect { validate("my-awesome-bucket") }.to raise_error(Discourse::InvalidParameters, error_message)
+      end
+
+      it "should raise an error when both buckets are equal except for a trailing slash" do
+        change_bucket_value("my-awesome-bucket/")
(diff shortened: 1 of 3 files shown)


spec/lib/site_settings/validations_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+describe SiteSettings::Validations do
+  subject { Class.new.include(described_class).new }
+
+  context "s3 buckets reusage" do
+    let(:error_message) { I18n.t("errors.site_settings.s3_bucket_reused") }
+
+    shared_examples "s3 bucket validation" do
+      def change_other_bucket_value(value)
+        SiteSetting.send("#{other_setting_name}=", value)
```
tgxworld: I usually prefer `public_send` to make sure we're never calling a private method.

spec/lib/site_settings/validations_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      it "shouldn't raise an error when both buckets are blank" do
+        change_other_bucket_value("")
+        expect { validate("") }.to_not raise_error
+      end
+
+      it "shouldn't raise an error when only one bucket is set" do
+        change_other_bucket_value("")
+        expect { validate("my-awesome-bucket") }.to_not raise_error
```
tgxworld: `to_not raise_error` [should be deprecated in a later version](https://github.com/rspec/rspec-expectations/pull/244) if I am not wrong. Anyway you can read more about the problem with it here -> https://github.com/rspec/rspec-expectations/issues/231

spec/lib/site_settings/validations_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+require 'site_settings/validations'
+
+describe SiteSettings::Validations do
+  subject { Class.new.include(described_class).new }
+
+  context "s3 buckets reusage" do
+    let(:error_message) { I18n.t("errors.site_settings.s3_bucket_reused") }
+
+    shared_examples "s3 bucket validation" do
+      def change_other_bucket_value(value)
```
tgxworld: hmm I would just call this `change_bucket_value`. The word `other` isn't really reflective of any actions being down in this method.

spec/lib/site_settings/validations_spec.rb
gschlager: I didn't read anything about deprecating `not_to raise_error` (without an error class). I think the more specific `expect { }.not_to raise_error(SpecificErrorClass)` doesn't even work anymore. I like the `to_not raise_error`, because it's more explicit to me as a reader, but I guess you are right and the "not raise an error" is already in the test name. Yeah, I'm going to remove it.

gschlager: Oops! I forgot to push that change before I merged. 😊 
Fixed in https://github.com/discourse/discourse/commit/2bdbca38012c4b46d8788a11369d968eb8abe885

tgxworld: > I think the more specific expect { }.not_to raise_error(SpecificErrorClass) doesn't even work anymore

OO it was for a specific error. I've been holding this thought in my head that `not_to raise_error` is deprecated.

0xtavian:
@gschlager @tgxworld  Prohibiting S3 bucket reuse should be enforced if not already done, was there ever any movement on this?

gschlager:
@0xtavian I'm not following. That's exactly what this commit is all about.

PR #6787, opened 2018-12-17
Add category link renderer to plugin API

pmusaraj:
This lets themes/plugins override the category link display. I am working on a theme component that would use this plugin function to add its own renderer with icons in category badges.

### DIFF
--- a/app/assets/javascripts/discourse/helpers/category-link.js.es6
+++ b/app/assets/javascripts/discourse/helpers/category-link.js.es6
@@ -5,6 +5,12 @@ import { iconHTML } from "discourse-common/lib/icon-library";
 var get = Em.get,
   escapeExpression = Handlebars.Utils.escapeExpression;
 
+let _renderers = [];
+
+export function registerCategoryLinkRenderer(renderer) {
+  _renderers.unshift(renderer);
+}
+
 function categoryStripe(color, classes) {
   var style = color ? "style='background-color: #" + color + ";'" : "";
   return "<span class='" + classes + "' " + style + "></span>";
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/helpers/category-link.js.es6
```diff
@@ -32,76 +38,12 @@ export function categoryBadgeHTML(category, opts) {
...
-
-  const categoryStyle =
-    opts.categoryStyle || Discourse.SiteSettings.category_style;
-  if (categoryStyle !== "none") {
-    if (parentCat && parentCat !== category) {
-      html += categoryStripe(
-        get(parentCat, "color"),
-        "badge-category-parent-bg"
-      );
+  for (let i = 0; i < _renderers.length; i++) {
```
ZogStriP: What's wrong with `_renderers.each(renderer => {`?

ZogStriP: You can probably be even fancier and do something like

```javascript
let result = _renderers.find(r => r.render(category, opts));
if (results) return result;
```

pmusaraj: I tried both options, but they didn't work well. But thanks for raising this, the loop there is unnecessary, all we need is to run the render of the first item in `_renderers`, so I switched to simply: 

```
return _renderers[0].render(category, opts);
```

ZogStriP: 🤔 if we're only ever using only the first rendered, what's the point of allowing to register more than one?

pmusaraj: To allow theme components / plugins to override the default one. We use a similar pattern for icon renderers in `icon-library.js.es6`.

ZogStriP: But if it's an override, why keep them in an array?

pmusaraj: No specific reason. I'll take a look at refactoring.

PR #6788, opened 2018-12-17
FIX: Copy old status to new poll.

nbianca:


tgxworld:
We need tests for this so that it doesn't regress :)

PR #6789, opened 2018-12-17
FEATURE: Show a blurry preview when lazy loading images

eviltrout:
This generates a 10x10 PNG thumbnail for each lightboxed image.
If Image Lazy Loading is enabled (IntersectionObserver API) then
we'll load the low res version when offscreen. As the image scrolls
in we'll swap it for the high res version.

### DIFF
--- a/app/models/optimized_image.rb
+++ b/app/models/optimized_image.rb
@@ -68,7 +68,7 @@ class OptimizedImage < ActiveRecord::Base
         Rails.logger.error("Could not find file in the store located at url: #{upload.url}")
       else
         # create a temp file with the same extension as the original
-        extension = ".#{upload.extension}"
+        extension = ".#{opts[:format] || upload.extension}"
 
         if extension.length == 1
           return nil
@@ -96,6 +96,7 @@ class OptimizedImage < ActiveRecord::Base
             url: "",
             filesize: File.size(temp_path)
           )
+
           # store the optimized image and update its url
           File.open(temp_path) do |file|
             url = Discourse.store.store_optimized_image(file, thumbnail)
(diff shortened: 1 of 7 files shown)


eviltrout:
Looks like I broke a stubbed test. Will fix.

app/models/optimized_image.rb
```diff
@@ -68,7 +68,7 @@ def self.create_for(upload, width, height, opts = {})
         Rails.logger.error("Could not find file in the store located at url: #{upload.url}")
       else
         # create a temp file with the same extension as the original
-        extension = ".#{upload.extension}"
+        extension = ".#{opts[:format] || upload.extension}"
```
tgxworld: Is this option still required because I don't see it being used in this PR. Also, I think we shouldn't allow this option because the upload extension has to be based on that of the upload which is determined from the actual image that we're holding on this.

app/models/optimized_image.rb
```diff
@@ -87,7 +87,7 @@ def self.create_for(upload, width, height, opts = {})
         end
 
         if resized
-          thumbnail = OptimizedImage.create!(
+          thumbnail = OptimizedImage.new(
```
tgxworld: `create!` should be equivalent to `new..save!`

app/models/optimized_image.rb
```diff
@@ -189,10 +191,14 @@ def self.resize_instructions(from, to, dimensions, opts = {})
...
+
+    if opts[:colors]
+      instructions << "-colors" << opts[:colors].to_s
+    end
+
     # NOTE: ORDER is important!
-    %W{
-      convert
-      #{from}[0]
+    instructions + %W{
```
tgxworld: `concat` will probably be better since since we avoid an extra array allocation.

tgxworld:
I think we're missing some tests for the changes made to the server side code :grinning:

app/models/optimized_image.rb
eviltrout: That was a mistake of mine during a refactor (the code said png: 'png' instead of format: 'png')

For this particular feature we did a series of benchmarks and found that PNG gave the best results, so this is required to force the PNG format.

app/models/optimized_image.rb
eviltrout: Good catch, that was leftover from a pre-refactor where I needed to change attributes.

app/models/optimized_image.rb
eviltrout: Good call. I didn't realize concat mutated the array because it doesn't have a `!` method and returns the object. Changed it.

PR #6794, opened 2018-12-19
DEV: Refactor location string builders.

nbianca:
The [location](https://github.com/discourse/discourse/blob/master/lib/discourse_ip_info.rb#L45) is already provided by DiscourseIpInfo.

### DIFF
--- a/app/serializers/concerns/user_auth_tokens_mixin.rb
+++ b/app/serializers/concerns/user_auth_tokens_mixin.rb
@@ -21,11 +21,7 @@ module UserAuthTokensMixin
 
   def location
     ipinfo = DiscourseIpInfo.get(client_ip, locale: I18n.locale)
-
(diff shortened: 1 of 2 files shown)


ZogStriP:
I think we should keep the `.uniq` though.

app/serializers/concerns/user_auth_tokens_mixin.rb
```diff
@@ -22,10 +22,8 @@ def client_ip
   def location
     ipinfo = DiscourseIpInfo.get(client_ip, locale: I18n.locale)
 
-    location = [ipinfo[:city], ipinfo[:region], ipinfo[:country]].reject { |x| x.blank? }.join(", ")
-    return I18n.t('staff_action_logs.unknown') if location.blank?
-
-    location
+    return I18n.t('staff_action_logs.unknown') if ipinfo[:location].blank?
```
ZogStriP: These two lines can be shortened to this ;)

```suggestion
    ipinfo[:location].presence || I18n.t("staff_action_logs.unknown")
```

PR #6795, opened 2018-12-19
FEATURE: Add S3 etag value to uploads table

vinothkannans:


tgxworld:
@vinothkannans Can I clarify what the ETag is used for in the bigger scheme of S3 inventory check?

vinothkannans:
@tgxworld We're going to use ETag to confirm the object identity (like [this](https://github.com/discourse/discourse/pull/6705/files#diff-94ece895a0dd91f6615a160573faefe1R146)). Previously we only checked the object's key value. As per S3 documents ETag will never change until the object is overwritten.

lib/s3_helper.rb
```diff
@@ -23,10 +23,11 @@ def initialize(s3_bucket_name, tombstone_prefix = '', options = {})
   end
 
   def upload(file, path, options = {})
+    options[:body] = file
     path = get_path_for_s3_upload(path)
     obj = s3_bucket.object(path)
-    obj.upload_file(file, options)
-    path
+    output = obj.put(options)
```
gschlager: This makes me very uncomfortable. We are losing the multipart upload for files > 15MB by using `put` instead of `upload_file` and this restricts the maximum upload size to 5GB. That might be too low for backups.

I understand why you are doing it. It's very unfortunate that `upload_file` does return a boolean instead of the server's response. Still, I think the better option is to use `FileUploader` even though it's marked as "private api". Hopefully our tests will fail when the API changes.

https://github.com/aws/aws-sdk-ruby/blob/f465e9a7500659b74fce7d1cc4f17e6616aa0a4c/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb#L305-L308

tgxworld:
> we're going to use ETag to confirm the object identity (like this). Previously we only checked the object's key value. As per S3 documents ETag will never change until the object is changed.

Thanks for clarifying. Are we planning to backfill the ETag for older uploads?

vinothkannans:
@tgxworld yes, we have to backfilll the etags from first inventory data.

lib/tasks/uploads.rake
```diff
@@ -263,7 +263,8 @@ def migrate_to_s3
         begin
           file = File.open(path)
           content_type = `file --mime-type -b #{path}`.strip
-          to = s3.store_upload(file, upload, content_type)
+          to, upload.etag = s3.store_upload(file, upload, content_type)
+          upload.save
```
tgxworld:
```suggestion
          upload.save!
```

tgxworld:
@vinothkannans The specs have been updated to reflect that an ETag is returned. However, we're not testing that the ETag is now saved in the DB when creating an s3 based upload.

app/models/optimized_image.rb
```diff
@@ -99,7 +99,7 @@ def self.create_for(upload, width, height, opts = {})
 
           # store the optimized image and update its url
           File.open(temp_path) do |file|
-            url = Discourse.store.store_optimized_image(file, thumbnail)
+            url, thumbnail.etag = Discourse.store.store_optimized_image(file, thumbnail)
```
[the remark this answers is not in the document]
vinothkannans: > we can regenerate the optimized image

But, first we've to find to which optimized images are broken right?

lib/s3_helper.rb
vinothkannans: yes I already added the workaround https://github.com/discourse/discourse/commit/1e547c0ebc46a11115bf873e41c9af797d3ac881

lib/s3_helper.rb
```diff
@@ -4,6 +4,8 @@ class S3Helper
 
   class SettingMissing < StandardError; end
 
+  MULTIPART_THRESHOLD = 15 * 1024 * 1024
```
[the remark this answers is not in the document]
vinothkannans: Now I'm using existing constant from gem and forcing the `multipart_threshold` value. So we will receive an error when default threshold value changed. https://github.com/discourse/discourse/pull/6795/commits/23eb353188c88eab272777d007def48431a99116

lib/s3_helper.rb
tgxworld: What are the trade-offs using `obj.load` here?

vinothkannans: Multipart upload won't return etag (just void). So we using `.load` method to do http `head` request. It will update the object data except the body (file content).

tgxworld: Ahh icic. So we’ll need to do an extra request for large uploads. I guess for multipart uploads that isn’t really a huge deal since we’re already doing multiple requests.

app/models/upload.rb
```diff
@@ -234,7 +234,7 @@ def self.migrate_to_new_scheme(limit = nil)
           FileHelper.optimize_image!(path) if FileHelper.is_supported_image?(File.basename(path))
           # store to new location & update the filesize
           File.open(path) do |f|
-            upload.url = Discourse.store.store_upload(f, upload)
+            upload.url, upload.etag = Discourse.store.store_upload(f, upload)
```
tgxworld: I'm curious what happens in this assignment when `Discourse.store` is a different store? The pattern that we're using now requires all new devs to be aware of the fact that only the s3 store will return an etag.

lib/upload_creator.rb
```diff
@@ -123,7 +123,7 @@ def create_for(user_id)
 
       # store the file and update its url
       File.open(@file.path) do |f|
-        url = Discourse.store.store_upload(f, @upload)
+        url, @upload.etag = Discourse.store.store_upload(f, @upload)
```
tgxworld: I'm curious what happens in this assignment when Discourse.store is a different store? The pattern that we're using now requires all new devs to be aware of the fact that only the s3 store will return an etag.

vinothkannans: It's not a problem. If etag is not returned then it will be assigned to `nil` automatically. Anyway I improved the code further.

tgxworld: Can we set the etag in the `S3Store` instead? Since we already pass the `@upload` to `Discourse.store.store_upload`, perhaps we can set the etag within that method and only return the `url` only as it used to?

vinothkannans: @tgxworld now it's done like that.

vinothkannans:
@SamSaffron no I'll merge this soon after a approved review (may be today itself).

db/migrate/20181218071253_add_etag_to_uploads.rb
```diff
@@ -0,0 +1,17 @@
+class AddEtagToUploads < ActiveRecord::Migration[5.2]
+  def change
+    add_column :uploads, :etag, :string
+    add_index :uploads, [:etag]
+
+    add_column :optimized_images, :etag, :string
+    add_index :optimized_images, [:etag]
+  end
+
+  def down
```
tgxworld: We don't need `down` here since you're using `change`. Rails migration can automatically figure out how to migrate down.

vinothkannans: I removed it.

lib/file_store/s3_store.rb
```diff
@@ -19,12 +19,16 @@ def initialize(s3_helper = nil)
 
     def store_upload(file, upload, content_type = nil)
       path = get_path_for_upload(upload)
-      store_file(file, path, filename: upload.original_filename, content_type: content_type, cache_locally: true)
+      url, etag = store_file(file, path, filename: upload.original_filename, content_type: content_type, cache_locally: true)
+      upload.update_attributes!(etag: etag)
```
tgxworld: Do we need to save here? I think the upload record is expected to be saved by the client as it needs to save the url too.

vinothkannans: [here](https://github.com/discourse/discourse/blob/86255faa08ed01930b55888cf52bc75d339f9766/lib/upload_creator.rb#L129) we only updating the url attribute. In this case I'll to update that too.

tgxworld: Yea just make the method do a `save!`

vinothkannans: Done.

tgxworld:
Looks good :+1:

PR #6796, opened 2018-12-19
FIX: validate YAML before save to avoid race condition

xrav3nz:
YAML setting `ThemeField`s are validated in an `after_commit` callback, which does not finish before the setting is used elsewhere, specifically:

https://github.com/discourse/discourse/blob/bb67ca9d21fc0871ea545e5ffb551cac4c20e4dd/app/models/theme.rb#L348-L357

where `Theme#settings` depends on the theme field being validated.

This PR moves the `validate_yaml!` call to `before_save` instead. This is okay because the validation process is not expensive.

### DIFF
--- a/app/models/theme_field.rb
+++ b/app/models/theme_field.rb
@@ -171,10 +171,7 @@ COMPILED
       errors << e.message
     end
 
-    self.error = errors.join("\n").presence unless self.destroyed?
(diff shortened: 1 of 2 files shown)


SamSaffron:
OK let's try this, thanks 👍  and happy holidays

PR #6797, opened 2018-12-19
UX: show lazy loaded images while they are downloading

SamSaffron:
This ensures 2 big things

1. That we progressively load images even if they are right away in view
2. That we progressively load images on top of the existing blurry image

previously due to the way the polyfill is implemented (1) was not the case in safari. Also on chrome we missed out on all progressive loading. 

This still leaves the big piece of adding support for non light boxed images


coding-horror:
Yes let's get this merged quickly if there are no objections @eviltrout

eviltrout:
Looks good, let's try it out!

PR #6798, opened 2018-12-20
UX: adds missing wizard preview icons

hnb-ku:
With our recent move to SVG icons, using font-file based icons is not an option any more.

I've opted for path2D, which accepts an SVG path

Path2D is not supported by IE11 but the chances of admins running the wizard on IE11 are practically none. 

https://caniuse.com/#feat=path2d


jjaffeux:
Note: long term, but I have talked with @eviltrout to eventually move wizard step to classic hbs/css stuff at some point

PR #6801, opened 2018-12-21
FIX: show error msg on grant badge if message available from backend

mrfinch:
https://meta.discourse.org/t/cant-grant-badge-generic-error/66982/9

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-user-badges.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-user-badges.js.es6
@@ -70,9 +70,20 @@ export default Ember.Controller.extend(GrantBadgeController, {
             }
           });
         },
-        function() {
-          // Failure
-          bootbox.alert(I18n.t("generic_error"));
+        function(error) {
+          let err;
+          if (
+            error.jqXHR &&
+            error.jqXHR.responseJSON &&
+            error.jqXHR.responseJSON.failed
+          ) {
+            err = I18n.t("generic_error_with_reason", {
+              error: error.jqXHR.responseJSON.failed
+            });
+          } else {
+            err = I18n.t("generic_error");
+          }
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/controllers/admin-user-badges.js.es6
```diff
@@ -70,9 +70,20 @@ export default Ember.Controller.extend(GrantBadgeController, {
             }
           });
         },
-        function() {
-          // Failure
-          bootbox.alert(I18n.t("generic_error"));
+        function(error) {
```
eviltrout: wouldn't our `popupAjaxError` helper do all this logic?

mrfinch: I didn't know about this function. That function works here perfectly though had to fix a bug in it :slightly_smiling_face:

app/assets/javascripts/discourse/lib/ajax-error.js.es6
```diff
@@ -32,7 +32,7 @@ export function extractError(error, defaultMessage) {
     } else if (parsedJSON.error) {
       parsedError = parsedJSON.error;
     } else if (parsedJSON.failed) {
-      parsedError = parsedJSON.message;
+      parsedError = parsedJSON.failed;
```
SamSaffron: careful ... this is changing internal behaviour, I guess we kind of mucked up not having this documented in a test. 

https://github.com/discourse/discourse/blob/7b253dbe4a92f7656bbc779d68e07ec93c56dfed/app/controllers/admin/backups_controller.rb#L40-L40

mrfinch: I have fixed it now.

PR #6802, opened 2018-12-21
FEATURE: move posts to new/existing PM

techAPJ:
This PR allows splitting/merging of posts from/to personal messages.

When moving posts from a topic:

screen shot 2018-12-21 at 15 04 50

When moving posts from a message:

screen shot 2018-12-21 at 15 15 20

screen shot 2018-12-21 at 15 20 06


app/assets/javascripts/discourse/components/choose-message.js.es6
```diff
@@ -0,0 +1,65 @@
...
+  messagesChanged() {
+    const messages = this.get("messages");
+    if (messages) {
+      this.set("noResults", messages.length === 0);
+    }
+    this.set("loading", false);
+  },
+
+  search: debounce(function(title) {
+    const self = this,
```
eviltrout: I would like us to stop using `self = this` and instead use ES6 arrow functions for keeping scope.

app/assets/javascripts/discourse/components/choose-message.js.es6
```diff
@@ -0,0 +1,65 @@
...
+        self.setProperties({ messages: null, loading: false });
+      }
+    });
+  }, 300),
+
+  actions: {
+    chooseMessage(message) {
+      const messageId = Em.get(message, "id");
+      this.set("selectedTopicId", messageId);
+      Ember.run.next(() =>
```
eviltrout: Could this be an `afterRender` instead?

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
```diff
@@ -0,0 +1,217 @@
...
+      categoryId: null,
+      topicName: "",
+      tags: null,
+      participants: null
+    });
+
+    const isPrivateMessage = this.get("model.isPrivateMessage");
+    const canSplitTopic = this.get("canSplitTopic");
+    if (isPrivateMessage) {
+      if (canSplitTopic) {
```
eviltrout: This would be a good candidate for the ternary operator since most of the line is the same.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
```diff
@@ -0,0 +1,217 @@
...
+    }
+  },
+
+  @computed("selectedAllPosts", "selectedPosts", "selectedPosts.[]")
+  canSplitTopic(selectedAllPosts, selectedPosts) {
+    return (
+      !selectedAllPosts &&
+      selectedPosts.length > 0 &&
+      selectedPosts.sort((a, b) => a.post_number - b.post_number)[0]
+        .post_type === 1
```
eviltrout: What is this 1? Is there any way it could be replaced with a constant?

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
```diff
@@ -0,0 +1,217 @@
...
+    );
+  },
+
+  actions: {
+    performMove() {
+      if (this.get("newTopic")) {
```
eviltrout: This is quite a few if statements that could be DRYer. If you had a constant above of the different modes `["newTopic", "existingTopic", etc]` you could loop through them, check if true and then send to the string name for the action. 

This could also clean up `buttonTitle` above by doing something similar.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
```diff
@@ -0,0 +1,217 @@
...
+        : movePosts(topicId, {
+            destination_topic_id: this.get("selectedTopicId"),
+            post_ids: this.get("topicController.selectedPostIds")
+          });
+
+      promise
+        .then(result => {
+          this.send("closeModal");
+          this.get("topicController").send("toggleMultiSelect");
+          Ember.run.next(() => DiscourseURL.routeTo(result.url));
```
eviltrout: Why is `Ember.run.next` here required?

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
```diff
@@ -0,0 +1,217 @@
...
+      } else if (this.get("existingTopic")) {
+        this.send("movePostsToExistingTopic");
+      } else if (this.get("newMessage")) {
+        this.send("movePostsToNewMessage");
+      } else if (this.get("existingMessage")) {
+        this.send("movePostsToExistingMessage");
+      }
+    },
+
+    movePostsToExistingTopic() {
```
eviltrout: All these actions repeat themselves a lot. Would it be possible to make one action that takes a parameter rather?

app/assets/stylesheets/desktop/modal.scss
```diff
@@ -113,19 +113,30 @@
...
   #move-selected {
+    width: 475px;
+
     p {
       margin-top: 0;
     }
 
-    input[type="radio"] {
-      margin-right: 10px;
+    .radios {
```
eviltrout: I noticed in the screenshot that the labels don't line up with the radio buttons. This would be a good place to use flexbox so you could align them properly.

app/controllers/topics_controller.rb
```diff
@@ -607,20 +607,37 @@ def set_notifications
...
 
   def merge_topic
-    params.require(:destination_topic_id)
+    topic_id = params.require(:topic_id)
+    destination_topic_id = params.require(:destination_topic_id)
+    params.permit(:participants)
+    params.permit(:archetype)
 
-    topic = Topic.find_by(id: params[:topic_id])
+    raise Discourse::InvalidAccess if params[:archetype].present? && params[:archetype] == "private_message" && !guardian.is_staff?
```
eviltrout: You shouldn't need to check if archetype is present. Checking `== "private_message"` would work even if archetype is `nil`.

test/javascripts/controllers/topic-test.js.es6
```diff
@@ -281,49 +281,32 @@ QUnit.test("Can split/merge topic", function(assert) {
   const controller = this.subject({ model });
   const selectedPostIds = controller.get("selectedPostIds");
 
-  assert.not(
```
eviltrout: I see you're removing a lot of JS tests even though we're adding functionality. I think the new functionality should be tested.

app/assets/stylesheets/desktop/modal.scss
```diff
@@ -113,19 +113,30 @@
...
+      margin-bottom: 10px;
+
+      label {
```
jjaffeux: use class instead of styling on html attributes if possible

app/assets/javascripts/discourse/components/choose-message.js.es6
techAPJ: I tried replacing it with

```
Ember.run.scheduleOnce("afterRender", () => {
  $(`#choose-message-${messageId}`).prop("checked", "true")
});
```

but it's not working in this case, the topic is not getting selected.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
techAPJ: Good catch, done.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
techAPJ: Good idea, done. Regarding cleaning up `buttonTitle` above I tried using array constant for that as well but since locale keys and constants we are using here have different names I couldn't use it above.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
techAPJ: Good catch, not needed here. Removed.

app/assets/javascripts/discourse/controllers/move-to-topic.js.es6
techAPJ: Done 👍

test/javascripts/controllers/topic-test.js.es6
techAPJ: JavaScript tests added.

techAPJ:
@eviltrout @jjaffeux Thanks a lot for the review! ❤️ 

Updated the PR per your feedback.

PR #6803, opened 2018-12-21
FIX: Accounts for MarkdownIt's typographer replacements.

nbianca:


coding-horror:
We decided this is a "best effort" kind of thing and we'll leave it alone for now while evaluating the feature

PR #6804, opened 2018-12-21
FEATURE: image resizing discoverability

majakomel:


app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -862,6 +862,64 @@ export default Ember.Component.extend({
     }
   },
 
+  _placeImageScaleButtons($preview) {
+    const images = $preview.find("img.upload-img");
+
+    const regexString = `(\\!\\[\\S[^\\]]+\\|\\d{3,6}x\\d{3,6})(?:,(\\d+)\\%)?([^\\]]*?\\]\\(upload\\:\\/\\/(?:.)*?\\))`;
```
ZogStriP: The double escape is making this regexp really hard to read. Why aren't you declaring the regex directly?

```js
const uploadRegexp = /(!\[\S[^\]]+|\d+x\d+)(?::,(\d+)%)?([^\]]*?\]\(update:\/\/(?:.)*?\)))/g;
```

SamSaffron:
This is going to be awesome, screenshots please?

jjaffeux:
@SamSaffron 

screenshot 2018-12-25 at 15 33 35

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -862,6 +862,89 @@ export default Ember.Component.extend({
     }
   },
 
+  _appenedImageScaleButtons($images, imageScaleRegex) {
```
ZogStriP: I think there's a small typo here ;)

```suggestion
   _appendImageScaleButtons($images, imageScaleRegex) {
```

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -862,6 +862,89 @@ export default Ember.Component.extend({
...
+    const buttonScales = [100, 75, 50];
+    const imageWrapperTemplate = `<div class="image-wrapper"></div>`;
+    const buttonWrapperTemplate = `<div class="button-wrapper"></div>`;
+    const scaleButtonTemplate = `<a href class="scale-btn"></a>`;
+
+    $images.each((i, e) => {
+      const $e = $(e);
+      const matches = this.get("composer.reply").match(imageScaleRegex);
```
ZogStriP: I think this should be done outside the loop?

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -862,6 +862,89 @@ export default Ember.Component.extend({
...
+
+      if (!matches) {
+        return;
+      }
+
+      const match = matches[i];
+      const matchingPlaceholder = imageScaleRegex.exec(match);
```
ZogStriP: We should probably do this inside the next `if` ?

jjaffeux:
Did you manage to find a way to some of the edge cases we discussed?

majakomel:
@jjaffeux Yes, all discussed edge cases are now covered: 
- multiple uploads on the same line
- upload markup in a codeblock

Additionally I found out BBCode and HTML syntax could break it too - those are fixed as well.

This is updated layout - buttons are moved under the image since we figured out with Joffrey they don't work well with small images. They appear only when hovering image.

screen shot 2019-01-07 at 09 50 30

ZogStriP:
I LOLed 🤣 so much at that image!

jjaffeux:
I think this is kind of risky and we should wait 2.3 for this if that's ok with you @SamSaffron @majakomel

coding-horror:
Yes definitely agreed, let's wait on this.

SamSaffron:
I want to get to merging this, so it needs a bit of a rebase to deal with conflicts. It is a nice change! 

One change I would like to see prior to merging is to have the extra resizable class only added when you are previewing markdown, the engine should know that we are in preview mode and then inform the renderer to add the extra classes. That way we will not carry this extra markdown in cooked.

SamSaffron:
hmmm can we tail debounce, `this._placeImageScaleButtons($preview);` it only runs once a second max (with a 1 second delay) ... or maybe even just run it on hover in the desktop case?  

I don't think we want any perf impact here for cases where we are typing and this does move nodes around. Also we should confirm that on mobile this is only called when you explicitly hit preview.

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -767,6 +767,116 @@ export default Ember.Component.extend({
...
+        const currentScale = matchingPlaceholder[2] || 100;
+
+        $e.data("index", i).wrap(imageWrapperTemplate);
+        $e.parent().append(
+          $(buttonWrapperTemplate).attr("data-image-index", i)
+        );
+
+        buttonScales.forEach((buttonScale, buttonIndex) => {
+          const activeClass =
+            parseInt(currentScale) === buttonScale ? "active" : "";
```
jjaffeux:
```suggestion
            parseInt(currentScale, 10) === buttonScale ? "active" : "";
```

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -253,7 +261,7 @@ export default Ember.Component.extend({
       const lastMatch = matchingPlaceholder[matchingPlaceholder.length - 1];
       const regex = new RegExp(regexString);
       const orderNr = regex.exec(lastMatch)[1]
```
ZogStriP: There's no point in executing the regex twice ;)

majakomel: Will address this in a separate commit since it's not related to this feature

app/assets/javascripts/discourse/components/composer-editor.js.es6
```diff
@@ -774,6 +782,116 @@ export default Ember.Component.extend({
...
+    // match 2 `5`
+    // match 3 `](upload://ceEfx3vO7bx7Cecv2co1SrnoTpW.png)`
+    const imageScaleRegex = /(!\[(?:\S*?(?=\|)\|)*?(?:\d{1,6}x\d{1,6})+?)(?:,?(\d{1,3})?%?)?(\]\(upload:\/\/\S*?\))/g;
+
+    // wraps previewed upload markdown in a codeblock in its own class to keep a track
+    // of indexes later on to replace the correct upload placeholder in the composer
+    if ($preview.find(".codeblock-image").length === 0) {
+      this.$(".d-editor-preview *")
+        .contents()
+        .filter(function() {
```
ZogStriP:
```suggestion
        .filter(() => this.nodeType === 3)
```

jjaffeux:
Awesome @majakomel 🎉

PR #6805, opened 2018-12-21
UX: improves dates display in emails

jjaffeux:
discourse_local_dates_email_format site setting is now available to tweak the format.

### DIFF
--- a/plugins/discourse-local-dates/config/settings.yml
+++ b/plugins/discourse-local-dates/config/settings.yml
@@ -1,4 +1,7 @@
 plugins:
+  discourse_local_dates_email_format:
+    default: "YYYY-MM-DDTHH:mm:ss[Z]"
+    client: true
   discourse_local_dates_enabled:
     default: true
     client: true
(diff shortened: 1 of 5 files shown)


plugins/discourse-local-dates/config/settings.yml
```diff
@@ -1,4 +1,7 @@
 plugins:
+  discourse_local_dates_email_format:
+    default: "YYYY/MM/DD HH:mm:ss"
```
gschlager: I think the default should be the ISO 8601 date format.

```suggestion
    default: "YYYY-MM-DD HH:mm:ss"
```

I wonder if this should be a localizable string instead of a site setting.

SamSaffron: isn't it automatically localisable due to moment.js?

SamSaffron:
I want something merged cause what we have now does not make sense ... 

I prefer `17:23 22-01-2016` but really anything is better than what we have now and admins can change this. 

In the US I never ever got used to the crazy "month before day" notation, my brain does not parse it. 

Anyway ... @coding-horror  can make the call here on formats of dates in email. and lets get this merged.

coding-horror:
International format is fine

PR #6808, opened 2018-12-23
DEV: upgrades from Ember 2.13 to Ember 3.5.1

jjaffeux:
Co-Authored-By: Bianca Nenciu <nbianca@users.noreply.github.com>


db/migrate/20181218143317_force_theme_compilation_for_ember3.rb
```diff
@@ -0,0 +1,6 @@
+class ForceThemeCompilationForEmber3 < ActiveRecord::Migration[5.2]
+  def up
+    ThemeField.force_recompilation!
+    Theme.expire_site_cache!
```
davidtaylorhq: I don't think a migration is the best solution here. We already have a method to force recompilation of themes when things change.

https://github.com/discourse/discourse/blob/master/app/models/theme_field.rb#L35

Just bump the `COMPILER_VERSION`, and `ThemeField#ensure_baked!` will ensure the field is recompiled  before being served to the client.

davidtaylorhq: Just to follow up here @jjaffeux I'm happy to make this change - we need to wait to see which PR will be merged first out of this one, and theme localisations (both need to bump the COMPILER_VERSION version).

jjaffeux: hmmm

PR #6809, opened 2018-12-24
FEATURE: Add `Top Uploads` report

khalilovcmded:
## Why do we need this feature?

This is part of: https://meta.discourse.org/t/gain-understanding-of-file-uploads-usage/104994

I am introducing a new report: `Top Uploads` to take a first step in solving the problem of: `an Admin wants to gain understanding of their forum file uploads usage`.

## How does this work?

1. This is building on top of the `report.rb` class by providing a new method: `report_top_uploads` which in turn will be automatically detected as a new report by the `Admin::ReportsController#index` ...

2. The report will show the largest (by filesize) N (default: 250) uploads, and it will show by which user is the upload created.

## How does it look like?

top uploads

### DIFF
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -79,7 +79,7 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-          let aValue = label.compute(a, { useSortProperty: true }).value;
+           const aValue = label.compute(a, { useSortProperty: true }).value;
           let bValue = label.compute(b, { useSortProperty: true }).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-          const aValue = label.compute(a).value;
-          const bValue = label.compute(b).value;
+          let aValue = label.compute(a).value;
+          let bValue = label.compute(b).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
         };
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-          let aValue = label.compute(a).value;
-          let bValue = label.compute(b).value;
+          let aValue = label.compute(a, { useSortProperty: true }).value;
+          let bValue = label.compute(b, { useSortProperty: true }).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
         };
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-          const aValue = label.compute(a, {useSortProperty: true}).value;
-          const bValue = label.compute(b, {useSortProperty: true}).value;
+          const aValue = label.compute(a).value;
+          const bValue = label.compute(b).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
         };
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-           const aValue = label.compute(a, { useSortProperty: true }).value;
-           const bValue = label.compute(b, { useSortProperty: true }).value;
+          const aValue = label.compute(a, {useSortProperty: true}).value;
+          const bValue = label.compute(b, {useSortProperty: true}).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
         };
--- a/app/assets/javascripts/admin/components/admin-report-table.js.es6
+++ b/app/assets/javascripts/admin/components/admin-report-table.js.es6
@@ -80,7 +80,7 @@ export default Ember.Component.extend({
       const compare = (label, direction) => {
         return (a, b) => {
            const aValue = label.compute(a, { useSortProperty: true }).value;
-          let bValue = label.compute(b, { useSortProperty: true }).value;
+           const bValue = label.compute(b, { useSortProperty: true }).value;
           const result = aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
           return result * direction;
         };
--- a/app/assets/javascripts/admin/models/report.js.es6
+++ b/app/assets/javascripts/admin/models/report.js.es6
@@ -266,7 +266,7 @@ const Report = Discourse.Model.extend({
         compute: (row, opts = {}) => {
           let value = null;
 
-          if(opts.useSortProperty) {
+          if (opts.useSortProperty) {
             value = row[label.sort_property || mainProperty];
           } else {
             value = row[mainProperty];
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/admin/components/admin-report-table.js.es6
```diff
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
     if (sortLabel) {
       const compare = (label, direction) => {
         return (a, b) => {
-          let aValue = label.compute(a).value;
-          let bValue = label.compute(b).value;
+          let aValue = label.compute(a, { useSortProperty: true }).value;
```
khalilovcmded: We wish to allow Admins to sort the uploads table by file size, yet, our column value is based on the `filesize_label` property which is of type String. We also return the `filesize` property which is a number.

If [you look at this line](https://github.com/discourse/discourse/pull/6809/files#diff-3f97cbb8726f3310e0b0c386dbe89e22R1451), you can see that we are setting the `sort_property` to be `filesize`. Therefore, I found that within the `compute` method, we still end up using the `main property for the column`, not the `sort_property`. 

I exposed this new option to allow depending on the `sort_property` if supplied by the backend.

app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -264,7 +264,13 @@ const Report = Discourse.Model.extend({
         mainProperty,
         type,
         compute: (row, opts = {}) => {
-          const value = row[mainProperty];
+          let value = null;
+
+          if(opts.useSortProperty) {
```
khalilovcmded: [Look here](https://github.com/discourse/discourse/pull/6809/files#r243835413) ☝️

app/models/report.rb
```diff
@@ -1,6 +1,8 @@
 require_dependency 'topic_subtype'
 
 class Report
+  extend ActionView::Helpers::NumberHelper
```
khalilovcmded: I am not very happy about extending the model class with a helper, so I am open to suggestions 😃

app/models/report.rb
```diff
@@ -1418,6 +1420,73 @@ def self.report_storage_stats(report)
...
+    DB.query(sql).each do |row|
+      data = {}
+      data[:author_id] = row.user_id
+      data[:author_username] = row.username
+      data[:author_avatar_template] = User.avatar_template(row.username, row.uploaded_avatar_id)
+      data[:filesize_label] = number_to_human_size(row.filesize)
+      data[:filesize] = row.filesize
+      data[:extension] = row.extension
+      data[:file_url] = "#{Discourse.base_url}/#{row.url}"
+      data[:file_name] = row.original_filename.truncate(25)
```
khalilovcmded: This is to make sure the table view isn't messed up. I am not happy about this too, I believe this should be the responsibility of the view, Ember should easily [handle this](https://github.com/romulomachado/ember-cli-string-helpers#truncate).

app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -264,7 +264,13 @@ const Report = Discourse.Model.extend({
         mainProperty,
         type,
         compute: (row, opts = {}) => {
-          const value = row[mainProperty];
+          let value = null;
```
jjaffeux: This should be enough

```suggestion
           let value;
```

app/models/report.rb
jjaffeux: yes we could handle this client side, lt me think about it. However it's cool that we don’t send over he network unnecessary things

app/assets/javascripts/admin/components/admin-report-table.js.es6
```diff
@@ -79,8 +79,8 @@ export default Ember.Component.extend({
...
+          let bValue = label.compute(b, { useSortProperty: true }).value;
```
jjaffeux:
```suggestion
           const bValue = label.compute(b, { useSortProperty: true }).value;
```

app/assets/javascripts/admin/components/admin-report-table.js.es6
jjaffeux:
```suggestion
           const aValue = label.compute(a, { useSortProperty: true }).value;
```

jjaffeux: :+1:

app/assets/javascripts/admin/components/admin-report-table.js.es6
khalilovcmded: Nice! I believe we should also change `aValue` to be `const`.

app/assets/javascripts/admin/components/admin-report-table.js.es6
jjaffeux: 👍

app/models/report.rb
jjaffeux: A simple solution would be to give this report its own file, we already do this for some reports. This way we would only include it for this specific report and not all reports.

khalilovcmded: I think this is a good approach. Looking at the `report.rb`, I could imagine, even from a directory structure, it would be better having a `lib/reporting/...` directory where each report type perhaps have its own class.

I am happy to go by your suggestion in this PR, or leave the refactor to a subsequent one. What do you think?

app/models/report.rb
```diff
@@ -1418,6 +1420,73 @@ def self.report_storage_stats(report)
...
+    up.filesize,
+    up.original_filename,
+    up.extension,
+    up.url
+    FROM uploads up
+    JOIN users u
+    ON u.id = up.user_id
+    WHERE up.created_at >= '#{report.start_date}' AND up.created_at <= '#{report.end_date}'
+    ORDER BY up.filesize DESC
+    LIMIT #{report.limit || 250}
```
khalilovcmded: The `limit` value always defaults at `50`, I have an incoming fix in a separate PR to support this.

khalilovcmded:
@jjaffeux I am trying to figure out this CI failure, but it seems super cryptic:

job__49486_3_-_discourse_discourse_-_travis_ci

I ran both `.es6` files with Discourse's prettier version, and it didn't yield any problems. (prettier's output is the same as the committed code).

Do you have any ideas?

SamSaffron:
Note from looking at that screenshot ... sorting columns is going to have limited value since it appears you are doing this all from the client side? For context many of our customers have 100s of thousands of records in the uploads table.  

The only particular use case I can think about for wanting sorting here is ... find all the largest PNG, or find largest DOC upload by bob. Sorting alone will not get us there we would need filters as well.  

I think adding more generic filtering support to the reporting framework may be interesting, but I would be very careful to get something merged in prior to adding even more features.

app/models/report.rb
jjaffeux: I think you can change this one to work like this, no need to change everything yet. I have let multiple refactorings on the side waiting for old dashboard to be removed. It would have forced me to have duplicated logic in some cases, once we don’t have the old one we can go aggressive on refactoring without fearing of breaking old features.

Also I think `lib/reports/` might be slightly better`

jjaffeux:
> @jjaffeux I am trying to figure out this CI failure, but it seems super cryptic:
> 
> job__49486_3_-_discourse_discourse_-_travis_ci
> 
> I ran both `.es6` files with Discourse's prettier version, and it didn't yield any problems. (prettier's output is the same as the committed code).
> 
> Do you have any ideas?

Make sure you have rebase from discourse master and it should work. My workflow is often something like this:

```
cd my-discours-dir
git checkout top-uploads-report
// discourse here is the main discourse git repo
git fetch discourse && git rebase discourse/master
// X is the number of commits on my working branch I want to merge into one
git reset --soft HEAD~X
// origin here should your khalilovcmded/discourse git repo, force updating the branch to last revision of your work
git push origin +top-uploads-report
```

khalilovcmded:
@SamSaffron 

> sorting columns is going to have limited value ....

Very good point!

> I think adding more generic filtering support to the reporting framework may be interesting, but I would be very careful to get something merged in prior to adding even more features.

If I understand you correctly, then I think it makes sense to introduce the generic filtering support in a separate PR first, that will have no functional changes but it will prepare us to jump back to this PR and add the file extension filtering.

app/models/report.rb
khalilovcmded: > I think you can change this one to work like this, no need to change everything yet.

That makes sense 👍

app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -264,7 +264,13 @@ const Report = Discourse.Model.extend({
         mainProperty,
         type,
         compute: (row, opts = {}) => {
-          const value = row[mainProperty];
+           let value;
```
ZogStriP: Looks like the indentation is off by one here ;)

app/models/report.rb
```diff
@@ -1418,6 +1420,73 @@ def self.report_storage_stats(report)
...
+
+    DB.query(sql).each do |row|
+      data = {}
+      data[:author_id] = row.user_id
+      data[:author_username] = row.username
+      data[:author_avatar_template] = User.avatar_template(row.username, row.uploaded_avatar_id)
+      data[:filesize_label] = number_to_human_size(row.filesize)
+      data[:filesize] = row.filesize
+      data[:extension] = row.extension
+      data[:file_url] = "#{Discourse.base_url}/#{row.url}"
```
ZogStriP: That might not work when using S3 as file storage. You should instead use something like `Discourse.store.cdn_url(row.url)`

khalilovcmded: Nice catch 👍

khalilovcmded: I am thinking if this is worth an extra test, just to be safe :)

jjaffeux:
Merged in https://github.com/discourse/discourse/commit/f1269fa807ba5c36c4d3c73b5b45e45457cd05f7

:tada:

Awesome @khalilovcmded  thanks!

PR #6816, opened 2018-12-27
FEATURE: Exposing a way to add a generic report filter

khalilovcmded:
## Why do we need this change?

Part of the work discussed [here](https://meta.discourse.org/t/gain-understanding-of-file-uploads-usage/104994), and implemented a first spike [here](https://github.com/discourse/discourse/pull/6809), I am trying to expose a single generic filter selector per report.

## How does this work?

We basically expose a simple, single generic filter that is computed and displayed based on backend values passed into the report.

This would be a simple contract between the frontend and the backend.

**Backend changes:** we simply need to return a list of dropdown / select options, and enable the report's newly introduced `custom_filtering` property.

For example, for our [Top Uploads](https://github.com/discourse/discourse/pull/6809/files#diff-3f97cbb8726f3310e0b0c386dbe89e22R1423) report, it can look like this on the backend:

```ruby
report.custom_filtering = true
report.custom_filter_options = [{ id: "any", name: "Any" }, { id: "jpg", name: "JPEG" } ]
```

In our javascript report HTTP call, it will look like:

```js
{
  "custom_filtering": true,
  "custom_filter_options": [
    {
      "id": "any",
      "name": "Any"
    },
    {
      "id": "jpg",
      "name": "JPG"
    }
  ]
}
```

**Frontend changes:** We introduced a generic `filter` param and a `combo-box` which hooks up into the existing framework for fetching a report.

This works alright, with the limitation of being a single custom filter per report. If we wanted to add, for an instance a `filesize filter`, this will not work for us. _I went through with this approach because it is hard to predict and build abstractions for requirements or problems we don't have yet, or might not have._

## How does it look like?

a1ktg1odde

## More on the bigger picture

The major concern here I have is the solution I introduced might serve the `think small` version of the reporting work, but I don't think it serves the `think big`, I will try to shed some light into why.

Within the current design, It is hard to maintain QueryParams for dynamically generated params (based on the idea of introducing more than one custom filter per report).

To allow ourselves to have more than one generic filter, we will need to:

a. Use the Route's model to retrieve the report's payload (we are now dependent on changes of the QueryParams via computed properties)
b. After retrieving the payload, we can use the `setupController` to define our dynamic QueryParams based on the custom filters definitions we received from the backend
c. Load a custom filter specific Ember component based on the definitions we received from the backend

I would also say that a lot of code can be simplified / removed if we are able to jump across to the latest Ember version. (yet, this is hard work)


jjaffeux:
Do we need `custom_filtering`? if checking if there are `custom_filter_options` should be enough no ?

jjaffeux:
"I would also say that a lot of code can be simplified / removed if we are able to jump across to the latest Ember version. (yet, this is hard work)"

This will happen very soon.

khalilovcmded:
@jjaffeux 

> Do we need custom_filtering? Checking if there are custom_filter_options should be enough no ?

> And support this client side:
xxxxx?filter=[size=small&file-extension=gif]
> It's all pseudo code, but I hope you get the idea.

This sounds like a much better idea, it has been circulating in my head but I was unsure if I should go for it before I get some feedback on a simpler approach. It feels like a good approach, at least for simple, option based filters 😃 

I will have to check how this plays out with QueryParams but I believe it should be alright.

> This will happen very soon. Inspect ember meta version. ;)

I saw the [PR](https://github.com/discourse/discourse/pull/6808), monumental effort! 😻If you need help, I would be happy to.

app/assets/javascripts/admin/components/admin-report.js.es6
```diff
@@ -226,10 +231,33 @@ export default Ember.Component.extend({
   },
 
   actions: {
+    filter(filterOptionId, value) {
+      let params = [];
+      let paramPairs = {};
+      let newParams = [];
+
+      if (this.get("filter")) {
+        let filter = this.get("filter").slice(1, -1);
```
jjaffeux:
```suggestion
         const filter = this.get("filter").slice(1, -1);
```

app/assets/javascripts/admin/components/admin-report.js.es6
```diff
@@ -226,10 +231,33 @@ export default Ember.Component.extend({
...
+        params = filter.split("&") || [];
+        params.map(p => {
+          let pair = p.split("=");
```
jjaffeux:
```suggestion
           const pair = p.split("=");
```

app/models/report.rb
```diff
@@ -1434,8 +1439,27 @@ def self.report_storage_stats(report)
   end
 
   def self.report_top_uploads(report)
-    report.modes = [:table]
+    if report.filter.present?
```
jjaffeux: We need to find a better solution here. We can’t have  to add so much code for each report to handle filters.


I suggest you look at what we do for categories.

We might also have to rewrite sql queries using the builder: https://github.com/discourse/mini_sql

khalilovcmded: > We need to find a better solution here. We can’t have to add so much code for each report to handle filters.

That's a good point, it seems that this can be easily extracted out into a method that can be used by any report.

> We might also have to rewrite sql queries using the builder: https://github.com/discourse/mini_sql

Awesome, I will look into that 👍

app/models/report.rb
```diff
@@ -1480,12 +1497,17 @@ def self.report_top_uploads(report)
...
     JOIN users u
     ON u.id = up.user_id
-    WHERE up.created_at >= '#{report.start_date}' AND up.created_at <= '#{report.end_date}'
+    /*where*/
     ORDER BY up.filesize DESC
     LIMIT #{report.limit || 250}
     SQL
 
-    DB.query(sql).each do |row|
+    extension_filter = report.filter_values["file-extension"]
```
khalilovcmded: @jjaffeux is this what you had in mind? Also, I used the SQL Builder below 👇 

I believe any new report that uses custom filters will need to add a single line per filter, I feel it is alright, but happy to hear your feedback 😃 

```ruby
builder.where("up.extension = :extension", extension: extension_filter) if extension_filter.present?
```

app/models/report.rb
```diff
@@ -1480,12 +1497,17 @@ def self.report_top_uploads(report)
...
+    builder = DB.build(sql)
+    builder.where("up.created_at >= :start_date", start_date: report.start_date)
+    builder.where("up.created_at < :end_date", end_date: report.end_date)
+    builder.where("up.extension = :extension", extension: extension_filter) if extension_filter.present?
```
jjaffeux: do we need this `.present?`  we should build filter_values, to make sure `report.filter_values["xxx"]` returns nil of not present to avoid this everywhere

app/models/report.rb
```diff
@@ -1435,7 +1445,14 @@ def self.report_storage_stats(report)
 
   def self.report_top_uploads(report)
     report.modes = [:table]
-
+    report.filter_options = [
+      {
+        id: "file-extension",
+        selected: report.filter_values.fetch("file-extension", "any"),
+        choices: ["jpeg", "gif", "png"],
+        allowAny: true
```
jjaffeux: I wonder three things here :p

- do we need a `allowAll` ? as in "any file extension"
- should `allowAny` means you can enter your own value? Maybe we need a better name ?
- for this specific case, could choices use the allowed extensions file setting?

app/models/report.rb
```diff
@@ -29,6 +29,8 @@ def initialize(type)
     @modes = [:table, :chart]
     @prev_data = nil
     @dates_filtering = true
+    @filter_options = nil
+    @filter = nil
```
jjaffeux: maybe it should be plural ?

jjaffeux:
Generally speaking I think we need better naming here, the difference between "filter", "filter_values", "filter_options" is not clear without reading the code in details. maybe some of it should be private too.

jjaffeux:
Also we will need at least one acceptance test to check clicking on col head is actually setting the filter. And I see no trace of direction handling, filesize ASC/DESC ?

app/models/report.rb
khalilovcmded: 1. Yes, I think we do. That being said, I should use: `SiteSetting.authorized_extensions` for `choices` value
2. The assumption now is that it has a default value that is [fetched from client.en.yml](https://github.com/discourse/discourse/pull/6816/files#diff-7ac2736614f0a5678e71cecf1563384eR2867), the default value is: `any` ... I would say this is okay.
3. Yes, totally 😃

app/models/report.rb
khalilovcmded: I think it would be okay to make it plural, though, the value of `filter` is a string, and not an array / collection / enumerable. So it is a single value, from correctness point of view, it is a singular value.

khalilovcmded:
> Generally speaking I think we need better naming here, the difference between "filter", "filter_values", "filter_options" is not clear without reading the code in details. maybe some of it should be private too.

Perhaps yes, naming is hard 😸 I am thinking that the naming problem here is because we want to make sure we mentally connect all these three variables for an Engineer reading the code. I personally find it mentally connectable once an Engineer inspects how `filter` is used in the `reports.rb`. Do you have a better naming on your mind?

> Also we will need at least one acceptance test to check clicking on col head is actually setting the filter. 

Can you elaborate more on this? which acceptance test? which `col head`?

> And I see no trace of direction handling, filesize ASC/DESC ?

I am not sure I understand `direction handling`, the sorting is client side only, and in a case a filter is applied, the sorting behaviour isn't available. This is how it works 👇 

gif_reports

app/models/report.rb
khalilovcmded: > report.filter_values["xxx"] returns nil of not present to avoid this everywhere

So this is already the case, `filter_values` will return a `nil` if the key doesn't exist. Yet, the `.present?` here is a conditional for allowing the builder to specify the `where` statement itself.

If we don't have `.present?` then even if `report.filter_values["xxx"] returns nil`, we are still going to have a where condition that we don't need, which will result in wrong results.

jjaffeux:
> Can you elaborate more on this? which acceptance test? which `col head`?

Sorry I meant when selecting filter

app/models/report.rb
jjaffeux: For me any would mean I could type: "something" and it would use it the where query.

So basically I don’t think this report should support "any". "all" or a choice in the list should be what we need. "any" should also not be a selectable entry in the list but we should only implement combo-box filtering handling.

jjaffeux:
> I am not sure I understand `direction handling`, the sorting is client side only, and in a case a filter is applied, the sorting behaviour isn't available. This is how it works 👇

I feel like filter should be handling filesize sorting server side for example. Indeed no point in sorting extension here. 

Adding a filter on author could be cool too. But let's keep this for another PR.

SamSaffron:
@jjaffeux is this ready for merging? can you go ahead and merge?

SamSaffron:
I worry a lot that we are sitting on this big PR for too long @khalilovcmd  can you make it your #1 priority to merge this in.

khalilovcmded:
@SamSaffron makes sense! I am prioritising this today.

app/models/report.rb
```diff
@@ -73,9 +76,13 @@ def prev_end_date
     self.start_date
   end
 
+  def filter_values
+    return self.filter.delete_prefix("[").delete_suffix("]").split("&").map { |param| param.split("=") }.to_h if self.filter.present?
```
jjaffeux: can we put the if before and not inline ?

app/assets/javascripts/admin/templates/components/admin-report.hbs
```diff
@@ -173,6 +173,17 @@
           </div>
         {{/if}}
 
+        {{#each filterOptions as |filterOption|}}
```
jjaffeux:
```
              {{combo-box content=filterOption.choices
                          filterable=true
                          value=filterOption.selected
                          allowAny=true
                          onSelect=(action "filter" filterOption.id)}}
```

I suggest we try this, so we can filter by any extension if needed, eg: mp4 used to be allowed before and I want to find them

PR #6822, opened 2018-12-28
FEATURE: adds support for unicode usernames

jjaffeux:


app/models/username_validator.rb
```diff
@@ -65,21 +80,21 @@ def username_length_max?
 
   def username_char_valid?
     return unless errors.empty?
-    if username =~ /[^\w.-]/
+    unless username[/^[#{UNICODE_RANGES}\.\d\-_]+$/]
```
ZogStriP: Why not use `[[:alphanumeric:]]` instead?

lib/route_format.rb
```diff
@@ -1,7 +1,7 @@
 module RouteFormat
 
   def self.username
-    /[\w.\-]+?/
+    /[%\w.\-]+?/
```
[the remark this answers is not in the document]
jjaffeux: This is because encoding url client side will create this kind of usernames:

```js
encodeURI("東京都") // %E6%9D%B1%E4%BA%AC%E9%83%BD
```

If you don’t you will have this kind of errors:

```
No route matches [GET] "/u/%E6%9D%B1%E4%BA%AC%E9%83%BD/preferences"
```

app/models/username_validator.rb
jjaffeux: I guess you are talking of `/[[:alnum]]/` that seems like a good idea to me. We have ranges here because originally I was given a defined list of accepted unicode ranges. But that would probably work, will change the PR to use this and see if it creates issues.

jjaffeux:
@SamSaffron are you ok if we merge this ?

SamSaffron:
This is very complicated, a lot more complicated than this. 

In particular we don't even have punycode going and given how much we love: https://meta.discourse.org/u/sam/activity vs https://meta.discourse.org/u/1/sam/activity swinging the URLs right is going to be super hard. cause we will need one of these puppies in our code https://www.punycoder.com/ 

So my bits of feedback here are 

1. How are URLs going to work for all user routes? punycode vs slug+id like we do for categories? 
2. I think this should be optional per site in the multisite, some sites may want to support unicode usernames / some not.

@eviltrout thoughts?

SamSaffron:
Correction to my previous deleted comment :)

I am good merging this provided 

- We test autocomplete after @ still works nicely (I think it does) 

- We ensure this is default off in English (can be default on in Hebrew / Japanese / Chinese and many other languages)

- Works in a multisite so some sites may have this on, others not

gschlager:
Should the `UserNameSuggester` be updated as well?

SamSaffron:
The only blocker here merging is the site setting, we got to get it in.

@gschlager , @jjaffeux  is mega busy can you help finish up the PR I would like to merge it in the next couple of days.

SamSaffron:
@gschlager  I am closing this PR not cause I don't want it, I would love it, but more cause I don't want it sitting in the queue while it is pending work. Once you are done here send through a new PR :)

PR #6837, opened 2018-12-31
Adding a setting for redis client id

domingusj:
Some redis providers, such as Google Memorystore do not allow the CLIENT
command.

https://github.com/mperham/sidekiq/wiki/Using-Redis#disabled-client-command

Discussion: https://meta.discourse.org/t/is-there-a-way-to-change-the-redis-connection-timeout/104476/11

### DIFF
--- a/app/models/global_setting.rb
+++ b/app/models/global_setting.rb
@@ -142,6 +142,7 @@ class GlobalSetting
         c = {}
         c[:host] = redis_host if redis_host
         c[:port] = redis_port if redis_port
+        c[:id] = redis_client_id if redis_client_id || "*"
 
         if redis_slave_host && redis_slave_port
           c[:slave_host] = redis_slave_host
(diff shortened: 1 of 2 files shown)


app/models/global_setting.rb
```diff
@@ -142,6 +142,7 @@ def self.redis_config
         c = {}
         c[:host] = redis_host if redis_host
         c[:port] = redis_port if redis_port
+        c[:id] = redis_client_id if redis_client_id || "*"
```
SamSaffron: this line does not look right to me:

```
irb(main):001:0> 1 if nil || "*"
(irb):1: warning: string literal in condition
=> 1
```

SamSaffron:
I would prefer this is handled using a plugin. I feel this is just too much of an edge case to carry this code in Discourse.

PR #6839, opened 2019-01-01
FIX: composer full screen shortcut doesn't work if inputs have focus

hnb-ku:
History:

https://meta.discourse.org/t/fullscreen-composer-keyboard-shortcut-works-only-once/101920/3

The full screen shortcut in the composer does not work if either the title or body inputs have focus. Since we auto-focus the title field, this creates an extra step where the user has to click outside of that input to be able to trigger the shortcut and enter full screen mode.

In the transition between full screen mode and normal mode, the title field is removed and reinserted and it gets auto-focused every time. 

To avoid the extra step where the user has to click outside, we bind the shortcut to the inputs so that the user will be able to use it even if the inputs have focus.

### DIFF
--- a/app/assets/javascripts/discourse/components/composer-title.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-title.js.es6
@@ -1,3 +1,4 @@
+/* global Mousetrap:true */
 import {
   default as computed,
   observes,
--- a/app/assets/javascripts/discourse/components/composer-title.js.es6
+++ b/app/assets/javascripts/discourse/components/composer-title.js.es6
@@ -1,11 +1,13 @@
 import {
   default as computed,
-  observes
+  observes,
+  on
 } from "ember-addons/ember-computed-decorators";
 import InputValidation from "discourse/models/input-validation";
 import { load, lookupCache } from "pretty-text/oneboxer";
 import { ajax } from "discourse/lib/ajax";
 import afterTransition from "discourse/lib/after-transition";
+import { getOwner } from "discourse-common/lib/get-owner";
 
 export default Ember.Component.extend({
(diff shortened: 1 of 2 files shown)


SamSaffron:
I really worry about having to carry so much code just for one shortcut.

hnb-ku:
Yeah @SamSaffron I feel the same way. I'm not happy about this either but it's a tricky problem. 

The shortcut works fine if the inputs are not focused so I think it's OK to close this and revisit in the future.

SamSaffron:
Hmmm ... but this shortcut appears to work without any extra magic

https://github.com/discourse/discourse/blob/8b7a2d1cb725df422a34ea4a1082a6edd756abae/app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6#L9-L9

Maybe if we use something else it will not be handled by other handlers?

pmusaraj:
I stumbled on a possible alternative to this PR while looking at our mousetrap fork: it's mousetrap that blocks the shortcut from working when the focus is on `input` elements, and it's that way by design to avoid triggering shortcuts while typing. There is, however, a mousetrap plugin that allows binding a shortcut globally. I think we can use this plugin (happy to look into it). 

In fact, it would be useful for `ctrl+alt+f` too, that shortcut only works now because of this hack: 

https://github.com/discourse/discourse/blob/4459665deeb84fb43621be90d64378ebc5fbd365/app/assets/javascripts/discourse/components/d-editor.js.es6#L258-L265

jjaffeux:
Should we close this PR ?

hnb-ku:
Yes. I'm closing this in favor of @pmusaraj PR.

PR #6840, opened 2019-01-01
FEATURE: Make auth_redirect param options on user_api_keys

cfitz:
This is a possible solution for https://meta.discourse.org/t/user-api-keys-specification/48536/19
This allows for user-api-key requests to not require a redirect url.
Instead, the encypted payload will just be displayed after creation  ( which can be copied
pasted into an env for a CLI, for example  )

### DIFF
--- a/app/controllers/user_api_keys_controller.rb
+++ b/app/controllers/user_api_keys_controller.rb
@@ -61,12 +61,13 @@ class UserApiKeysController < ApplicationController
     raise Discourse::InvalidAccess unless meets_tl?
 
     validate_params
+    @application_name = params[:application_name]
 
     # destroy any old keys we had
     UserApiKey.where(user_id: current_user.id, client_id: params[:client_id]).destroy_all
 
     key = UserApiKey.create!(
-      application_name: params[:application_name],
+      application_name: @application_name,
       client_id: params[:client_id],
       user_id: current_user.id,
       push_url: params[:push_url],
@@ -76,7 +77,7 @@ class UserApiKeysController < ApplicationController
 
     # we keep the payload short so it encrypts easily with public key
     # it is often restricted to 128 chars
-    payload = {
+    @payload = {
       key: key.key,
       nonce: params[:nonce],
       push: key.has_push?,
@@ -84,12 +85,15 @@ class UserApiKeysController < ApplicationController
     }.to_json
 
     public_key = OpenSSL::PKey::RSA.new(params[:public_key])
-    payload = Base64.encode64(public_key.public_encrypt(payload))
+    @payload = Base64.encode64(public_key.public_encrypt(@payload))
 
-    if params[:auth_redirect]
-      redirect_to "#{params[:auth_redirect]}?payload=#{CGI.escape(payload)}"
-    else
-      render plain: payload
+    (redirect_to("#{params[:auth_redirect]}?payload=#{CGI.escape(@payload)}") && return) if params[:auth_redirect]
+    respond_to do |format|
+      format.html { render :show }
+      format.json do
+        instructions = I18n.t("user_api_key.instructions", application_name: @application_name)
+        render json: { payload: @payload, instructions: instructions }
+      end
     end
   end
 
--- a/app/controllers/user_api_keys_controller.rb
+++ b/app/controllers/user_api_keys_controller.rb
@@ -87,12 +87,15 @@ class UserApiKeysController < ApplicationController
     public_key = OpenSSL::PKey::RSA.new(params[:public_key])
     @payload = Base64.encode64(public_key.public_encrypt(@payload))
 
-    (redirect_to("#{params[:auth_redirect]}?payload=#{CGI.escape(@payload)}") && return) if params[:auth_redirect]
-    respond_to do |format|
-      format.html { render :show }
-      format.json do
-        instructions = I18n.t("user_api_key.instructions", application_name: @application_name)
-        render json: { payload: @payload, instructions: instructions }
+    if params[:auth_redirect]
+      redirect_to("#{params[:auth_redirect]}?payload=#{CGI.escape(@payload)}")
+    else
+      respond_to do |format|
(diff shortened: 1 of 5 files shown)


vsoch:
Huge +1 to this addition! The first working example (also to be merged soon!) is via the helpme client :point_right: https://github.com/vsoch/helpme/pull/34 when you write:

```
helpme discourse https://stage.neurostars.org
```
(or some other board) it will walk you through posting an issue, meaning a title, text, optional asciinema and environment, and then post without leaving the command line! This PR here allows for the generation of the user token on the first try (when the user doesn't have one yet).

SamSaffron:
Just rendering the payload seems somewhat surprising, I would expect some sort of instructions on the final page, eg:

> We just generated a new user API key for you, please paste the following key into your application:
>
> abc12312ab12312312321

vsoch:
The generation of the token happens by way of the client, and the user needs to copy paste the payload into the terminal. This PR redirects the user to the page with the payload, and the token generation needs to happen by the client so we can use their credentials files. The instructions are also clear when the interaction is happening. It also only needs to happen once, so it's not too much of a hassle.

SamSaffron:
I follow, but a real page here is very beneficial. It is then clear you are on site cause the header is still there, and there are a more instructions to go with cause we don't simply render a white page with a code. 

It is a very simple page rendering this with some canned message and the no-ember layout.

cfitz:
Adding a view makes sense to me. But also, if the request is expecting a JSON response, should we return JSON? I added both, but let me know if I should just return the html version.

Also I added a new i18n entry to the server.en.yml. Is there a process I should follow to add these to the other translations?

Lastly, should I be squashing my commits?

vsoch:
If the user needs to physically click to approve, how can we expect Json? Would they approve, leave the window open, then press enter (and then the client does a GET to the page with json?) Is that a better interaction than copy paste?

cfitz:
I think the login and the approval will still need to be done via the browser. The only change is if a user post's to user-api-key.json they'll get a JSON response with the payload and the instructions message instead of the html page.
Biut, yeah, that might be unnecessary.

vsoch:
Is this change live with stage.neurostars.org? If so I'd like to run it through testing again to see the new interaction (and try out the json).

cfitz:
@vsoch just updated it. try it out and see if that works for you.

app/controllers/user_api_keys_controller.rb
```diff
@@ -76,17 +77,24 @@ def create
...
       api: AUTH_API_VERSION
     }.to_json
 
     public_key = OpenSSL::PKey::RSA.new(params[:public_key])
-    payload = Base64.encode64(public_key.public_encrypt(payload))
-
-    redirect_to "#{params[:auth_redirect]}?payload=#{CGI.escape(payload)}"
+    @payload = Base64.encode64(public_key.public_encrypt(@payload))
+
+    (redirect_to("#{params[:auth_redirect]}?payload=#{CGI.escape(@payload)}") && return) if params[:auth_redirect]
```
SamSaffron: style wise ... the && return is just so easy to miss ... 

if params[:auth_redirect]
   ... do one thing
else
   ... do another thing
end

app/controllers/user_api_keys_controller.rb
```diff
@@ -51,22 +51,23 @@ def create
 
     require_params
 
-    unless SiteSetting.allowed_user_api_auth_redirects
+    if params.key?(:auth_redirect) && SiteSetting.allowed_user_api_auth_redirects
```
SamSaffron: yeah ... unless is just so hard to reason about ... agree with this change and sorry about using it there in the first place :)

SamSaffron:
one minor comment and we are good to merge 🎊

vsoch:
Testing now.

vsoch:
hey @cfitz the page is totally blank
image

vsoch:
Here is the actual page, looks like some kind of blank 404? 
[notfound.html.txt](https://github.com/discourse/discourse/files/2722161/notfound.html.txt)

vsoch:
Does the flow need to change at all?

vsoch:
I'll try the .json version and see if that shows anything.

vsoch:
It's not clear to me if I'm supposed to do user-api-key.json/new, or user-api-key/new.json - both trigger an error. Here is doing:

```python
    url = (board + "/user-api-key.json/new?scopes=write&application_name=HelpMe&public_key=" + 
           self.public_key.replace("'", "") + 
           "&client_id=" + self.client_id +
           "&nonce=" + nonce )
```

image

And then doing the other way:

```python
    url = (board + "/user-api-key/new.json?scopes=write&application_name=HelpMe&public_key=" + 
           self.public_key.replace("'", "") + 
           "&client_id=" + self.client_id +
           "&nonce=" + nonce )
```
image

And if I leave as it was before (not asking for json) I get an empty page. Let me know what to try!

cfitz:
@vsoch  Sorry, that's my bad! I realized the last git pull on the stage server actually errored because the git history diverged. I just reset it and pulled. I think it should be working now.

 Sorry, I think I confused things about the JSON handler...the test suite POSTs directly to the user_api_keys controller.  The /user-api-key/new ( without the .json ) action is the form that the user needs to approve.

vsoch:
okay, I will try again (with the original flow)

vsoch:
Oh it's lovely! So much better!

image

vsoch:
Still testing though :) Hang tight!

vsoch:
okay it looks good! Here is the post that I just made:

https://stage.neurostars.org/t/why-are-jelly-beans-so-funky-looking/29

And it has it's own asciinema, but I also recorded a second one to see the client interaction (so @SamSaffron  can see too!)

https://asciinema.org/a/219484?speed=2

vsoch:
@SamSaffron did you have any other issues we need to work on?

SamSaffron:
Looks good to me! thanks heaps!

PR #6848, opened 2019-01-04
FEATURE: Support for localized themes

davidtaylorhq:
> - Themes can supply translation files in a format like `/locales/{locale}.yml`. These files should be valid YAML, with a single top level key equal to the locale being defined. For now these can only be defined using the `discourse_theme` CLI, importing a `.tar.gz`, or from a GIT repository.
> 
> - Fallback is handled on a global level (if the locale is not defined in the theme), as well as on individual keys (if some keys are missing from the selected interface language).
> 
> - Administrators can override individual keys on a per-theme basis in the /admin/customize/themes user interface.
> 
> - Theme developers should access defined translations using the new theme prefix variables:
>   JavaScript: `I18n.t(themePrefix + "my_translation_key")`
>   Handlebars: `{{i18n (concat theme-prefix "my_translation_key")}}`
> 
> - To design for backwards compatibility, theme developers can check for the presence of the `themePrefix` variable in JavaScript

Key things to review (because they will be difficult to change later):
- Variable names for theme developers:
  - `theme-prefix` in handlebars. Used like `{{i18n (concat theme-prefix "sometranslationkey")}}`
  - `themePrefix` in javascript. Used like `i18n.t(themePrefix + "sometranslationkey)`
- Translation files: 
  - Translations are stored under `/locales` in a theme
  - Files have names in the format `en.yml`, `fr.yml`, etc.
  - Files should have a single top level key which is is the name of the locale

For an example of adding this to a theme, see https://github.com/davidtaylorhq/discourse-media-overlay-theme/commit/a39f554a042f8e1b91d2f1272f3790dec38a5a71


app/assets/javascripts/admin/components/theme-translation.js.es6
```diff
@@ -0,0 +1,21 @@
...
+export default Ember.Component.extend(BufferedContent, SettingComponent, {
+  layoutName: "admin/templates/components/site-setting",
+  setting: Ember.computed.alias("translation"),
+
+  @computed("translation.key")
+  settingName(key) {
+    return key;
+  },
+
+  type: "string",
```
eviltrout: Bit weird for this variable to be initialized down here. I'd suggest moving it up top.

app/assets/javascripts/admin/models/theme.js.es6
```diff
@@ -188,6 +188,12 @@ const Theme = RestModel.extend({
     const settings = {};
     settings[name] = value;
     return this.save({ settings });
+  },
+
+  saveTranslation(name, value) {
+    const translations = {};
+    translations[name] = value;
+    return this.save({ translations });
```
eviltrout: I think we support the ES6 dynamic key syntax. If not we should enable that on babel! Would simplify this to:

`this.save({ translations: { [name]: value } });`

app/assets/javascripts/discourse/helpers/theme-injector.es6
```diff
@@ -0,0 +1,46 @@
...
+    value = value.replace(/\\u0022/g, '"');
+  }
+
+  if (!(context instanceof Ember.Object)) {
+    injectPlainObject(context, parent, key, value);
+    return;
+  }
+
+  if (!parent) {
+    context.set(`${key}`, value);
```
eviltrout: Is `key` not a string already?

app/models/theme_field.rb
```diff
@@ -136,6 +159,65 @@ def process_html(html)
...
+    javascript_cache || build_javascript_cache
+    javascript_cache.content = ''
+    begin
+      data = translation_data
+
+      js = <<~JS
+        /* Translation data for theme #{self.theme_id} (#{self.name})*/
+        const data = #{data.to_json};
+
+        for (var lang in data){
```
eviltrout: We should always use `let`

app/models/theme_field.rb
```diff
@@ -136,6 +159,65 @@ def process_html(html)
...
+          var cursor = I18n.translations;
+          for (var key of [lang, "js", "theme_translations"]){
```
eviltrout: `let` here too

db/migrate/20181221121805_create_theme_translation_override.rb
```diff
@@ -0,0 +1,11 @@
+class CreateThemeTranslationOverride < ActiveRecord::Migration[5.2]
+  def change
+    create_table :theme_translation_overrides do |t|
+      t.integer :theme_id, null: false
+      t.string :locale, length: 30, null: false
+      t.string :translation_key, null: false
+      t.string :value, null: false
+      t.timestamps null: false
+    end
```
eviltrout: You will almost certainly want to create an index based on `theme_id` here at least.

It also seems to me that a unique index on `[:theme_id, :locale, :translation_key]` would be prudent.

lib/theme_translation_manager.rb
```diff
@@ -0,0 +1,57 @@
...
+        record.save!
+      else
+        record = create_record!(new_value.to_s)
+      end
+      record.value
+    end
+  end
+
+  def db_record
+    theme.theme_translation_overrides.to_a.find { |i| i.locale.to_s == @locale.to_s && i.translation_key.to_s == key.to_s }
```
eviltrout: This line is a bit long, I'd suggest splitting it up.

lib/theme_translation_manager.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def has_record?
+    db_record.present?
+  end
+
+  def create_record!(value)
+    record = ThemeTranslationOverride.new(locale: @locale, translation_key: @key, theme: @theme, value: value)
```
eviltrout: These three lines could be one `ThemeTranslationOverride.create`

davidtaylorhq:
Thanks @eviltrout, I've made those changes

app/assets/javascripts/admin/components/theme-translation.js.es6
```diff
@@ -0,0 +1,21 @@
...
+
+export default Ember.Component.extend(BufferedContent, SettingComponent, {
+  layoutName: "admin/templates/components/site-setting",
+  setting: Ember.computed.alias("translation"),
+
+  @computed("translation.key")
+  settingName(key) {
+    return key;
+  },
+
```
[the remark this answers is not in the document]
davidtaylorhq: I'm guessing you're referring to the `  type: "string",` line that used to be here? Fixed that following Robin's review 😄

app/assets/javascripts/discourse/helpers/theme-injector.es6
```diff
@@ -0,0 +1,46 @@
...
+
+  if (parent && !context.get(parent)) {
+    context.set(parent, {});
+  }
+  context.set(`${parent}.${key}`, value);
+}
+
+function injectPlainObject(context, parent, key, value) {
+  if (!parent) {
+    _.assign(context, { [key]: value });
```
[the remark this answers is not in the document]
davidtaylorhq: This isn't a change made in this PR - I renamed a file and GIT didn't detect the rename. Will try and fix that so history is preserved.

Not against changing this, but outside the scope of this PR.

app/assets/javascripts/discourse/helpers/theme-injector.es6
```diff
@@ -0,0 +1,46 @@
...
+  }
+  _.assign(context[parent], { [key]: value });
+}
+
+registerHelper("theme-injector", function(arr, hash) {
+  inject(hash.context, hash.parent, hash.key, hash.value);
+});
+
+Handlebars.registerHelper("theme-injector", function(hash) {
+  inject(this, hash.hash.parent, hash.hash.key, hash.hash.value);
```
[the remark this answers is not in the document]
davidtaylorhq: Very confusing - but it does seem to work. See other comment on this file, it's not supposed to be a part of this PR.

SamSaffron:
Everything sounds good but I would like to ship dedicated helpers here to make it easier.

```
{{theme-i18n "media-popup"}}

// ? maybe ? .... there is slight less importance to this one compared to `theme-i18n` raw helper
I18n.theme("my_translation_key")
```

Also can you confirm this is usable in raw handlebars as well as regular handlebars.

davidtaylorhq:
> Can you confirm this is usable in raw handlebars as well as regular handlebars.

Yep, `theme-prefix` is accessible in raw templates. However I just discovered that the `concat` helper is not 😢

---

```
{{theme-i18n "media-popup"}}
```
The problem here is that handlebars helpers cannot access the parent context, so the `theme-i18n` helper would not be able to identify the theme ID. We _could_ do something like 
```
{{theme-i18n .. "media-popup"}}
```
(where the `..` passes in the current context to the helper). Or for a bit less magic we could inject `theme-id` and use this syntax:
```
{{theme-i18n theme-id "media-popup"}}
```

Either of those options would also solve the lack of `concat` in raw templates. What do you think @SamSaffron?

davidtaylorhq:
This PR is now updated to inject the theme id during template compilation. The same approach has been applied to themeSettings. Tests are included for both raw and ember templates. `theme_field.rb` has been refactored to make this cleaner.

For theme developers, the new behaviour is:
```
Translations:
<h1>{{theme-i18n 'somestring'}}</h1>
{{d-button label=(theme-prefix 'somestring')}}

Settings, new syntax:
{{theme-setting 'some_setting'}}

Old syntax (automatically transformed):
{{themeSetting.some_setting}}
```

In JavaScript:
```
themePrefix("somestring") // Outputs theme_translation.##.somestring
I18n.t(themePrefix("somestring")) // Can be used like this

themeSetting.some_setting // Setting syntax remains the same (but now references a singleton ember service to retrieve settings)
```

app/assets/javascripts/discourse-common/lib/helpers.js.es6
```diff
@@ -45,20 +45,25 @@ function resolveParams(ctx, options) {
...
 }
 
-export function registerUnbound(name, fn) {
-  const func = function(property, options) {
-    if (
-      options.types &&
-      (options.types[0] === "ID" || options.types[0] === "PathExpression")
-    ) {
-      property = get(this, property, options);
+export function registerUnbound(name, fn, argCount = 1) {
```
eviltrout: I don't think we should lightly change the signature of this function, since it's used in many places.

Is it not possible to get the arity by checking `options.types.length` perhaps?

lib/theme_javascript_compiler.rb
```diff
@@ -0,0 +1,213 @@
...
+            "hash": {
+              "type": "Hash",
+              "pairs": [],
+            }
+        }
+      }
+
+      function manipulatePath(path){
+        // Override old themeSetting syntax when it's a param inside another node
+        if(path.parts[0] == "themeSetting"){
```
eviltrout: Trivial, but I imagine this code isn't linted? Seems well tested I just noticed the spacing is off :)

app/assets/javascripts/discourse/services/theme-settings.js.es6
```diff
@@ -0,0 +1,20 @@
+export default Ember.Service.extend({
+  settings: null,
+
+  init() {
+    this._super(...arguments);
+    this.set("settings", {});
+  },
+
+  registerSettings(themeId, settingsObject) {
+    this.get("settings")[themeId] = settingsObject;
```
eviltrout: If everything is accessed via the helpers, it's slightly faster to use plain old javascript objects. For example `this._settings = {}` in the constructor, and `this._settings[themeId] = settingsObject` here.

In general I recommend this unless something observes or computes a property based on the object, which doesn't seem to be the case in this interface.

app/helpers/application_helper.rb
```diff
@@ -402,6 +402,11 @@ def theme_lookup(name)
     lookup.html_safe if lookup
   end
 
+  def theme_translations_lookup
+    lookup = Theme.lookup_field(theme_ids, :translations, I18n.locale)
```
eviltrout: Trivial but you could use the safe navigation operator here.

spec/models/theme_field_spec.rb
```diff
@@ -188,4 +187,117 @@ def create_yaml_field(value)
...
     expect(field.error).to be_nil
   end
+
+  describe "locale fields" do
+
+    let!(:theme) { Fabricate(:theme) }
+    let!(:theme2) { Fabricate(:theme) }
+    let!(:theme3) { Fabricate(:theme) }
+
+    let!(:en1) { ThemeField.create!(theme: theme, target_id: Theme.targets[:translations], name: "en",
```
eviltrout: These lines are super long.

app/assets/javascripts/discourse-common/lib/helpers.js.es6
davidtaylorhq: The new `argCount` argument has a default, so it shouldn't have any impact on existing uses. I originally tried using `options.length`, but it didn't work because some existing places register like this:

```
registerUnbound('test', function(value){ ... });
```
while others want to use the keyword arguments, so they register like this:
```
registerUnbound('test', function(value, params){ ... });
```
So number of arguments of the handler is not a good indication of the number of arguments passed to the handlebars helper.

davidtaylorhq:
Thanks @eviltrout, I've made most of those changes. Not sure what the best path is for the first comment 🤔

app/assets/javascripts/discourse-common/lib/helpers.js.es6
davidtaylorhq: Resolved in d7daf54

PR #6850, opened 2019-01-04
PWA alerts

xfalcox:


app/models/admin_dashboard_data.rb
```diff
@@ -211,6 +211,15 @@ def github_config_check
     end
   end
 
+  def pwa_config_check
+    unless SiteSetting.large_icon.present? && SiteSetting.large_icon.width == 512 && SiteSetting.large_icon.height == 512
+      return I18n.t('dashboard.pwa_config_icon_warning', base_path: Discourse.base_path)
+    end
+    unless SiteSetting.short_title.present? && SiteSetting.short_title.size <= 12
```
gschlager: Should we force a max length for that setting instead of adding a warning? `site_settings.yml` has an option to define the `max` length.

xfalcox: Good Idea!

Added a max to the setting, but letting the warning there for now so we can guide sites who already filled it.

xfalcox:
@tgxworld I think this is ready for a review

config/locales/server.en.yml
```diff
@@ -1192,6 +1192,8 @@ en:
...
     twitter_config_warning: 'The server is configured to allow signup and log in with Twitter (enable_twitter_logins), but the key and secret values are not set. Go to <a href="%{base_path}/admin/site_settings">the Site Settings</a> and update the settings. <a href="https://meta.discourse.org/t/configuring-twitter-login-for-discourse/13395" target="_blank">See this guide to learn more</a>.'
     github_config_warning: 'The server is configured to allow signup and log in with GitHub (enable_github_logins), but the client id and secret values are not set. Go to <a href="%{base_path}/admin/site_settings">the Site Settings</a> and update the settings. <a href="https://meta.discourse.org/t/configuring-github-login-for-discourse/13745" target="_blank">See this guide to learn more</a>.'
+    pwa_config_icon_warning: 'The server is missing a proper large icon which allow users to add a homescreen shortcut to this site on Android. Go to <a href="%{base_path}/admin/site_settings/category/all_results?filter=large_icon">the Site Settings</a> and upload a icon of the recommended size.'
```
tgxworld:
```suggestion
    pwa_config_icon_warning: 'The server is missing a proper large icon which allows users to add a homescreen shortcut to this site on Android devices. Go to <a href="%{base_path}/admin/site_settings/category/all_results?filter=large_icon">the Site Settings</a> and upload an icon of the recommended size.'
```

tgxworld: `%{base_path}/admin/site_settings/category/all_results?filter=large_icon` One thing I learnt when building Discobot is to keep full paths out of translations. Translators can always find ways to mess this up and this make updating the URLs less brittle.

config/locales/server.en.yml
```diff
@@ -1192,6 +1192,8 @@ en:
...
+    pwa_config_title_warning: 'The server is missing a short title which allow users to add a homescreen shortcut to this site on Android. Go to <a href="%{base_path}/admin/site_settings/category/all_results?filter=short_title">the Site Settings</a> and fill a title of the recommended length.'
```
tgxworld:
```suggestion
    pwa_config_title_warning: 'The server is missing a short title which allows users to add a homescreen shortcut to this site on Android devices. Go to <a href="%{base_path}/admin/site_settings/category/all_results?filter=short_title">the Site Settings</a> and configure a title of the recommended length.'
```

spec/models/admin_dashboard_data_spec.rb
```diff
@@ -192,6 +192,52 @@ def my_test_method
     end
   end
 
+  describe 'pwa_config_check' do
+    subject { described_class.new.pwa_config_check }
+
+    it 'alerts for large_icon missing' do
+      SiteSetting.large_icon = nil
+      expect(subject).to_not be_nil
```
tgxworld: Perhaps we can test for the exact object here? `to_not be_nil` is too generic and returning an incorrect object will cause this test to pass.

tgxworld:
Looks good to me :+1:

PR #6854, opened 2019-01-07
FIX: show lock glyph to a user without permissions when expanding a q…

majakomel:
…uote

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
@@ -163,9 +163,10 @@ export default class PostCooked {
           $blockQuote.showHtml(div, "fast", finished);
         })
         .catch(e => {
-          if (e.jqXHR.status === 404) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -169,6 +169,12 @@ export default class PostCooked {
               "fast",
               finished
             );
+          } else if (e.jqXHR.status === 403) {
+            $blockQuote.showHtml(
+              $(`<div class='expanded-quote'>${iconHTML("lock")}</div>`),
+              "fast",
+              finished
+            );
```
tgxworld: Perhaps something like this to avoid duplicating the code above?

```
if ([403, 404].includes(e.jqXHR.status)) {
  const icon = e.jqXHR.status === "403" ? "lock" : "trash-o"; 

  $blockQuote.showHtml(
    $(`<div class='expanded-quote'>${iconHTML(icon)}</div>`),
    "fast",
    finished
  );
}
```

PR #6855, opened 2019-01-07
FIX: Add compatibility for bucket folder paths in migrate_to_s3 task

rishabhnambiar:
**Example** without bucket folder path

```text
DISCOURSE_S3_ACCESS_KEY_ID="" \
DISCOURSE_S3_SECRET_ACCESS_KEY="" \
DISCOURSE_S3_REGION="us-west-2" \
DISCOURSE_S3_BUCKET="rishabhmultisite" \
DISCOURSE_S3_CDN_URL="https://discourse-cdn-aws1.com/pocketgems" \
rake uploads:migrate_to_s3
```

screenshot from 2019-01-07 17-40-23

**Example** with bucket folder path

```text
DISCOURSE_S3_ACCESS_KEY_ID="" \
DISCOURSE_S3_SECRET_ACCESS_KEY="" \
DISCOURSE_S3_REGION="us-west-2" \
DISCOURSE_S3_BUCKET="rishabhmultisite/pocketgems" \
DISCOURSE_S3_CDN_URL="https://discourse-cdn-aws1.com/pocketgems" \
rake uploads:migrate_to_s3
```

screenshot from 2019-01-07 17-36-22

### DIFF
--- a/lib/s3_helper.rb
+++ b/lib/s3_helper.rb
@@ -12,7 +12,7 @@ class S3Helper
 
     @s3_bucket_name, @s3_bucket_folder_path = begin
       raise Discourse::InvalidParameters.new("s3_bucket_name") if s3_bucket_name.blank?
-      s3_bucket_name.downcase.split("/".freeze, 2)
+      self.class.get_bucket_and_folder_path(s3_bucket_name)
     end
 
     @tombstone_prefix =
@@ -21,6 +21,10 @@ class S3Helper
       else
         tombstone_prefix
       end
+  end
+
+  def self.get_bucket_and_folder_path(s3_bucket_name)
+    s3_bucket_name.downcase.split("/".freeze, 2)
   end
 
   def upload(file, path, options = {})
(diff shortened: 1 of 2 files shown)


ZogStriP:
I wonder if there's a way to extract that logic so that both these rake tasks and the s3_helper can use it instead of duplicating it?

rishabhnambiar:
> I wonder if there's a way to extract that logic so that both these rake tasks and the s3_helper can use it instead of duplicating it?

Sure, that makes sense. Do you have anything specific technique in mind, maybe an s3 [module](https://ruby-doc.org/core-2.2.0/Module.html)? If not, I will try to come up with something else :+1:

ZogStriP:
Nothing in mind, maybe just some class methods on the s3_helper class since it already has some?

lib/s3_helper.rb
```diff
@@ -23,6 +23,10 @@ def initialize(s3_bucket_name, tombstone_prefix = '', options = {})
       end
   end
 
+  def self.get_bucket_and_folder_path(s3_bucket_name)
+    return s3_bucket_name.downcase.split("/".freeze, 2)
```
ZogStriP: You don't need the `return` in Ruby when it's the last statement ;)

rishabhnambiar: Wow, I did that again :man_facepalming: 
Fixed, thanks!

PR #6858, opened 2019-01-08
DEV: Re-introduce composer upload acceptance tests that were removed …

rishabhnambiar:
…temporarily

These tests were removed temporarily in https://github.com/discourse/discourse/commit/41eacd859edac8598e380ff284c9f62a33884ab7, they pass for me locally 100% of the time, should we move them back in now or is this intentional?


majakomel:
@rishabhnambiar They're passing locally/travis, but were breaking the build.

rishabhnambiar:
Oh okay, thanks!
I was just wondering if we forgot to put them back! :grinning:

PR #6859, opened 2019-01-08
FEATURE: Allow the base font size to be changed on a per-user basis

davidtaylorhq:

### DIFF
--- a/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
@@ -58,7 +58,7 @@ export default Ember.Controller.extend(PreferencesTabController, {
     return currentThemeId();
   },
 
-  @computed()
+  @computed
   fontSizes() {
     return FONT_SIZES.map(value => {
       return { name: I18n.t(`user.font_size.${value}`), value };
@@ -86,7 +86,7 @@ export default Ember.Controller.extend(PreferencesTabController, {
     const classList = document.documentElement.classList;
     if (!classList.contains(`font-size-${newSize}`)) {
       FONT_SIZES.forEach(name => {
-        let className = `font-size-${name}`;
+        const className = `font-size-${name}`;
         if (newSize === name) {
           classList.add(className);
         } else {
--- a/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
@@ -80,25 +80,6 @@ export default Ember.Controller.extend(PreferencesTabController, {
     previewTheme([id]);
   },
 
-  @observes("model.user_option.font_size")
-  fontSizeChanged() {
-    const newSize = this.get("model.user_option.font_size");
-    const classList = document.documentElement.classList;
-    if (!classList.contains(`font-size-${newSize}`)) {
-      FONT_SIZES.forEach(name => {
-        const className = `font-size-${name}`;
-        if (newSize === name) {
-          classList.add(className);
-        } else {
-          classList.remove(className);
-        }
-      });
-
-      // Force refresh when leaving this screen
-      Discourse.set("assetVersion", "forceRefresh");
-    }
-  },
-
   homeChanged() {
     const siteHome = this.siteSettings.top_menu.split("|")[0].split(",")[0];
     const userHome = USER_HOMES[this.get("model.user_option.homepage_id")];
@@ -143,6 +124,22 @@ export default Ember.Controller.extend(PreferencesTabController, {
           this.homeChanged();
         })
         .catch(popupAjaxError);
+    },
+
+    selectFontSize(newSize) {
+      const classList = document.documentElement.classList;
+
+      FONT_SIZES.forEach(name => {
+        const className = `font-size-${name}`;
+        if (newSize === name) {
+          classList.add(className);
+        } else {
+          classList.remove(className);
+        }
+      });
+
+      // Force refresh when leaving this screen
+      Discourse.set("assetVersion", "forceRefresh");
     }
   }
 });
--- a/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
@@ -20,6 +20,8 @@ const USER_HOMES = {
   5: "top"
 };
 
+const FONT_SIZES = ["normal", "larger", "largest"];
+
 export default Ember.Controller.extend(PreferencesTabController, {
   @computed("makeThemeDefault")
   saveAttrNames(makeDefault) {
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/common/foundation/base.scss
```diff
@@ -13,6 +13,14 @@ html {
...
   overflow-y: scroll;
   direction: ltr;
+
+  &.font-size-larger {
+    font-size: $base-font-size-larger;
+  }
+
+  &.font-size-largest {
+    font-size: $base-font-size-largest;
+  }
```
davidtaylorhq: @awesomerobot are you happy for these dynamic sizes to be in `foundation/base.scss`, or would you rather they were moved somewhere else?

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
```diff
@@ -70,6 +80,25 @@ export default Ember.Controller.extend(PreferencesTabController, {
     previewTheme([id]);
   },
 
+  @observes("model.user_option.font_size")
+  fontSizeChanged() {
+    const newSize = this.get("model.user_option.font_size");
+    const classList = document.documentElement.classList;
+    if (!classList.contains(`font-size-${newSize}`)) {
+      FONT_SIZES.forEach(name => {
+        let className = `font-size-${name}`;
```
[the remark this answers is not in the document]
jjaffeux: Generally speaking observers are bad.

I would prefer if you would implement, `onSelect` on the combo-box and the associated action in the controller, pseudo code:

```
// hbs
{{combo-box onSelect=(action "selectFontSize")}}

//controller
actions: {
  selectFontSize(newSize) {
    this._changeFontSize(newSize);
  }
},

_changeFontSize(newSize) {
  //
}
```

Let me know if you need help.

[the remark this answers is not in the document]
davidtaylorhq: > Generally speaking observers are bad.

Why do you say that? (Not against changing it, just interested in the reasoning)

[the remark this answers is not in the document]
jjaffeux: It's against the "Data Down, Actions Up" (DDAU) philosophy, which embers tries to follow.

There is a good article about it here: https://dockyard.com/blog/2015/11/16/best-practices-functional-programming-and-the-observer-effect

Happy to talk more about this if you want.

app/models/user_option.rb
```diff
@@ -28,6 +28,12 @@ def self.like_notification_frequency_type
     @like_notification_frequency_type ||= Enum.new(always: 0, first_time_and_daily: 1, first_time: 2, never: 3)
   end
 
+  def self.font_sizes
+    @font_sizes = Enum.new(normal: 0, larger: 1, largest: 2)
+  end
+
+  validates :font_size_id, inclusion: { in: UserOption.font_sizes.values }
```
tgxworld: I don't think we should name this `font_size_id` as this usually means it `belongs_to :font_size`. Perhaps we can just name this column `font_size`? This will be similar to how we do it for posts where have `Post.types` and `Post#post_type`.

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
```diff
@@ -20,6 +20,8 @@ const USER_HOMES = {
   5: "top"
 };
 
+const FONT_SIZES = ["normal", "larger", "largest"];
```
tgxworld: @jjaffeux taught me a neat trick. You can add the `erb` extension on this and do `const FONT_SIZE = <%= UserOption.font_sizes.keys.map(&:to_s) %>`

config/site_settings.yml
```diff
@@ -1811,6 +1811,14 @@ user_preferences:
...
     type: category_list
     default: ''
+  
+  default_font_size:
+    type: enum
+    default: normal
+    choices:
+      - normal
+      - larger
+      - largest
```
tgxworld: I'm just thinking out loud here but will this allow us room for expansion if we're already at the `largest`? I was thinking if using the actual pixel size as the name will be better since it is more intuitive to think in terms of the number of pixels instead of thinking about "how large is large".

SamSaffron: nahhh we need to keep the names simple here, I think this is fine for now.

SamSaffron:
@coding-horror  wanted to call this "text size" vs "font size", so we should sweep through and make a consistent rename here.

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
SamSaffron: note one side effect here is that its an erb file then and caching can be a bit annoying. I am on the fence on this ...

tgxworld: Why would caching be affected?

app/models/user_option.rb
davidtaylorhq: I was following the same convention as other enums in this same user_options table. e.g. `homepage_id`

I think having inconsistencies within the same table could be even more confusing than the `belongs_to` confusion 🤔

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
davidtaylorhq: One problem I can see is if individual sites wanted to extend this using a plugin - we can only compile one javascript file **per cluster**. 

I think if we really want to declare the sizes only on the server, we should use the Site Serializer. Writing javascript using erb templates doesn't feel like the 'right' solution to me, and might cause us problems down the line (e.g. if/when we move to ember-cli)

config/site_settings.yml
davidtaylorhq: I wanted to avoid hard-coding pixel values, because they can easily be changed by a site using some basic CSS. If we decide to change this later it would be a trivial migration to change the names.

davidtaylorhq:
Renamed everything to `text size`, and renamed the `font_size_id` column to `text_size_key` per @tgxworld's suggestion.

app/models/user_option.rb
```diff
@@ -185,6 +201,7 @@ def update_tracked_topics
 #  homepage_id                      :integer
 #  theme_ids                        :integer          default([]), not null, is an Array
 #  hide_profile_and_presence        :boolean          default(FALSE), not null
+#  font_size_id                     :integer          default(0), not null
```
SamSaffron: leftover :)

app/assets/stylesheets/common/foundation/variables.scss
```diff
@@ -33,6 +33,8 @@ $bronze: #cd7f32 !default;
 // --------------------------------------------------
 
 $base-font-size: 14px !default;
+$base-font-size-larger: 16px !default;
```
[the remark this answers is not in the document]
awesomerobot: Since these values are being set on the HTML tag, they're what all ems further down the cascade are being calculated from, so px values make sense. Using px at least ensures that we're using the same base size across all browsers. 

If we used ems they'd be based on the default browser font size (which I think is 16px in most cases?).

PR #6861, opened 2019-01-08
[docs] Add instructions for CentOS/RHEL users in install guide

jwflory:
I'm preparing to install Discourse on a CentOS 7 node and figured I would improve the install guide's post-install section to include instructions for CentOS/RHEL users, alongside the Ubuntu instructions.

I have signed the Discourse CLA.


tgxworld:
Thank you!

PR #6862, opened 2019-01-08
FEATURE: Remove full quotes only from new posts.

nbianca:


SamSaffron:
sorry, I think it needs to be revised I changed post.rb today so somehow this no longer merges cleanly.

lib/cooked_post_processor.rb
```diff
@@ -33,10 +33,10 @@ def initialize(post, opts = {})
     @disable_loading_image = !!opts[:disable_loading_image]
   end
 
-  def post_process(bypass_bump = false)
+  def post_process(bypass_bump = false, new_post = false)
```
eviltrout: It might be a good time to convert these arguments to be ruby keywords. Right now the method invocation would look like:

`post_process(true, false)` and I'd have no idea what those args would be.

`post_process(bypass_bump: true, new_post: false)` is much more clear.

PR #6863, opened 2019-01-08
FEATURE: set CSP base-uri and object-src to none

xrav3nz:
`object-src`: 
- For `<object>` , `<embed>` , and `<applet>` elements
- This is for Flash and family, so default to `'none'`

`base-uri` :

- Prevents the `<base>` of the page to be set to another domain which changes relative urls
- Discourse does not use `<base>` , so default to `'none'`

Setting them can prevent trivial bypasses to our CSP, and they are should quite safe.

### DIFF
--- a/lib/content_security_policy/default.rb
+++ b/lib/content_security_policy/default.rb
@@ -7,6 +7,8 @@ class ContentSecurityPolicy
 
     def initialize
       @directives = {}.tap do |directives|
+        directives[:base_uri] = [:none]
+        directives[:object_src] = [:none]
         directives[:script_src] = script_src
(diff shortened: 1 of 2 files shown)


pmusaraj:
Looks good @xrav3nz! Just one minor issue: these are currently not extendable, and I would like to at least merge this with `object_src` extendable. It's a little tricky currently, because the `none` keyword cannot be combined with sources, it would generate an invalid policy.

pmusaraj:
Extendability added via  https://github.com/discourse/discourse/commit/e11c6ffa89ee4b9416bc18a6eb41109109549e6b

PR #6864, opened 2019-01-09
FEATURE: Remove option for Google Plus sharing

zjwhitehead:
Not sure if this falls under a feature but given that Google Plus is being shut down (https://www.blog.google/technology/safety-security/expediting-changes-google-plus/) soon this sharing option should be removed.


tgxworld:
We'll need to remove this https://github.com/discourse/discourse/blob/4232d32699526dfe7592e39d009e529c519280a6/config/locales/client.en.yml#L143 too

zjwhitehead:
@tgxworld thanks, I removed the related translations

tgxworld:
Looks good to me :+1:  Thank you

PR #6867, opened 2019-01-09
FEATURE: Use amazon s3 inventory to manage upload stats

vinothkannans:


lib/s3_inventory.rb
```diff
@@ -0,0 +1,195 @@
...
+    @archive_filename = File.join(@tmp_directory, File.basename(file.key))
+    @csv_filename = @archive_filename[0...-3]
+
+    FileUtils.mkdir_p(@tmp_directory)
+    copy_archive_to_tmp_directory
+    unzip_archive
+
+    etags = []
+    connection = ActiveRecord::Base.connection.raw_connection
+    connection.exec('CREATE TEMP TABLE etags(val text PRIMARY KEY)')
```
tgxworld: What if two processes call this method at the same time? Would it fail because we can't create two temp tables of the same name? Also, what was the rational for using a temp table instead of holding the records in memory?

lib/s3_inventory.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    CSV.foreach(csv_filename, headers: false) do |row|
+      next if row[1]["/optimized/"].present? && skip_optimized
+
+      etags << row[S3Inventory::CSV_ETAG_INDEX]
+
+      if etags.length == 1000
```
tgxworld: Is there something special about etags with a length of `1000`? It'll probably be better to extract this into a variable so that it is clear what `1000` represents.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,195 @@
...
+        etags_clause = etags.map { |i| "('#{PG::Connection.escape_string(i.to_s)}')" }.join(",")
+        connection.exec("INSERT INTO etags VALUES #{etags_clause}")
+        etags = []
+      end
+    end
+
+    if etags.present?
```
tgxworld: If we're going to be inserting the etag into the empty table, can't we do it in the block above? We only have to not reset `etags = []` if `etag.length == 1000`

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let(:csv_filename) { File.new("#{Rails.root}/spec/fixtures/csv/s3_inventory.csv") }
+
+  before do
+    SiteSetting.enable_s3_uploads = true
+    SiteSetting.s3_access_key_id = "abc"
+    SiteSetting.s3_secret_access_key = "def"
+    SiteSetting.s3_inventory_bucket = "ghi"
+    SiteSetting.enable_s3_inventory = true
+
+    s3_helper = store.inventory.instance_variable_get(:@s3_helper)
```
tgxworld: Maybe we should add `s3_helper` to `attr_reader` if the instance variable needs to be accessed?

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+            id: "12345example25102679df27bb0ae12b3f85be6f290b936c4393484be31bebcc", 
+          }, 
+          size: 713193, 
+          storage_class: "STANDARD", 
+        }
+      ],
+      next_marker: "eyJNYXJrZXIiOiBudWxsLCAiYm90b190cnVuY2F0ZV9hbW91bnQiOiAyfQ=="
+    })
+    s3_helper.stubs(:s3_client).returns(client)
+    Discourse.stubs(:store).returns(store)
```
tgxworld: Hmm why do we need to stub this?

lib/s3_inventory.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    s3_options = S3Helper.s3_options(SiteSetting)
+    @s3_helper = S3Helper.new(SiteSetting.s3_inventory_bucket, '', s3_options)
+    @s3_client = @s3_helper.s3_client
+
+    @inventory_id = inventory_id
+  end
+
+  def file
```
tgxworld: Does this method need to be public?

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  it "will return recent inventory file name" do
+    expect(inventory.file.key).to eq("example1.csv.gz")
+  end
+
+  it "will raise storage error if inventory file not found" do
+    inventory.stubs(:file).returns(nil)
+    expect { inventory.list_missing_uploads }.to raise_error(S3Inventory::StorageError)
+  end
+
+  it "will raise storage error if inventory file not found" do
```
tgxworld: I think the description here is incorrect.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    CSV.foreach(csv_filename, headers: false) do |row|
+      Fabricate(:upload, etag: row[S3Inventory::CSV_ETAG_INDEX])
+    end
+
+    upload = Fabricate(:upload, etag: "ETag")
+    inventory.stubs(:unzip_archive)
+    inventory.stubs(:log)
+    inventory.stubs(:csv_filename).returns(csv_filename)
+    STDOUT.expects(:puts).with(upload.url)
+    STDOUT.expects(:puts).with("1 of 4 uploads are missing")
```
tgxworld: You can capture outputs to `STDOUT` without mocking with something like https://github.com/discourse/discourse/blob/ec58c33e9e5fde6c452eedbcb605b34b59ccc4e5/spec/components/migration/safe_migrate_spec.rb#L14-L22

spec/jobs/update_s3_inventory_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    store.inventory.stubs(:s3_client).returns(@client)
+    Discourse.stubs(:store).returns(store)
+  end
+
+  it "picks gravatar if system avatar is picked and gravatar was just downloaded" do
+    @client.expects(:put_bucket_policy).with(
+      bucket: "ghi",
+      policy: "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Sid\":\"InventoryAndAnalyticsPolicy\",\"Effect\":\"Allow\",\"Principal\":{\"Service\":\"s3.amazonaws.com\"},\"Action\":[\"s3:PutObject\"],\"Resource\":[\"arn:aws:s3:::ghi/*\"],\"Condition\":{\"ArnLike\":{\"aws:SourceArn\":\"arn:aws:s3:::bucket\"},\"StringEquals\":{\"s3:x-amz-acl\":\"bucket-owner-full-control\"}}}]}"
+    )
+    @client.expects(:put_bucket_inventory_configuration).with(
```
tgxworld: Instead of mocking, can we capture the request and inspect the params?

config/site_settings.yml
```diff
@@ -1050,6 +1050,11 @@ files:
   s3_configure_tombstone_policy:
     default: true
     shadowed_by_global: true
+  enable_s3_inventory:
+    default: false
+  s3_inventory_bucket:
```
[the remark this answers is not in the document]
vinothkannans: I think it's good idea. It should be possible. I'll check that.

lib/s3_inventory.rb
vinothkannans: nothing. I'm just doing the batch insert after every 1000 records. I'll create a variable for it.

lib/s3_inventory.rb
vinothkannans: > What if two processes call this method at the same time? Would it fail

I think so. I have to fix it.

> what was the rational for using a temp table instead of holding the records in memory?

The reason is to check missing objects [using simple `JOIN` query](https://github.com/discourse/discourse/pull/6867/files/6fa012dbf08131ac757ad2583a9f1c45ee402a14#diff-7fc755e5e9dc2118547f063b20bbb2ddR86).

spec/components/s3_inventory_spec.rb
vinothkannans: Yes, I also think we need few more refactoring here.

spec/components/s3_inventory_spec.rb
vinothkannans: Because `Discourse.store` method currently returning new `s3_store` object on every calls. Yes, the existing code needs to be changed. I'll improve this while s3_store and s3_helper refactoring.

lib/s3_inventory.rb
vinothkannans: I need this for rspec test cases [here](https://github.com/discourse/discourse/pull/6867/files/6fa012dbf08131ac757ad2583a9f1c45ee402a14#diff-43df0e63549fdfd6a4400759c68ce74eR51).

lib/backup_restore/s3_backup_store.rb
```diff
@@ -32,9 +32,7 @@ def delete_file(filename)
     end
 
     def download_file(filename, destination_path, failure_message = nil)
-      unless @s3_helper.object(filename).download_file(destination_path)
-        raise failure_message&.to_s || "Failed to download file"
```
tgxworld: I'm curious as to why we need to remove this when this is related to the backup store.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+  class StorageError < RuntimeError; end
+
+  ::DiscourseEvent.on(:site_setting_saved) do |site_setting|
+    name = site_setting.name.to_s
+    Jobs.enqueue(:update_s3_inventory) if name.include?("s3_inventory") || name == "s3_upload_bucket"
+  end
+
+  attr_reader :inventory_id, :csv_filename, :model
+
+  CSV_KEY_INDEX ||= 1.freeze
```
tgxworld: Perhaps we can use `# frozen_string_literal: true` instead?

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+
+  def list_missing
+    if file.blank?
+      Rails.logger.warn("Failed to list inventory from S3")
+      raise StorageError
+    end
+
+    DistributedMutex.synchronize("s3_inventory_list_missing_#{inventory_id}") do
+      current_db = RailsMultisite::ConnectionManagement.current_db
+      timestamp = Time.now.strftime("%Y-%m-%d-%H%M%S")
```
tgxworld:
```suggestion
      timestamp = Time.zone.now.strftime("%Y-%m-%d-%H%M%S")
```

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+            connection.put_copy_data("#{row[CSV_KEY_INDEX]},#{row[CSV_ETAG_INDEX]}\n")
+          end
+        end
+
+        missing_uploads = model.joins("LEFT JOIN #{table_name} ON #{table_name}.etag = #{model.table_name}.etag").where("#{table_name}.etag is NULL")
+        missing_count = missing_uploads.count
+
+        if missing_count > 0
+          missing_uploads.find_each do |upload|
+            puts upload.url
```
tgxworld: I think we can just pluck the `url` column here or only select the `:url` field.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+
+          puts "#{missing_count} of #{model.count} #{model.name.underscore.pluralize} are missing"
+        end
+      ensure
+        connection.exec("DROP TABLE #{table_name}") unless connection.nil?
+      end
+    end
+  end
+
+  def copy_archive_to_tmp_directory
```
tgxworld: While this is being used in the `backup/restorer` code, I don't agree with the name used. The file that is downloaded here is a gzipped CSV file so I dpn't think we should refer to it as an archive.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+    log "Downloading archive to tmp directory..."
+    failure_message = "Failed to download archive to tmp directory."
+
+    @s3_helper.download_file(file.key, @archive_filename, failure_message)
+  end
+
+  def unzip_archive
```
tgxworld:
```suggestion
  def decompress_inventory_file
```

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+      },
+      filter: {
+        prefix: filter_prefix
+      },
+      is_enabled: SiteSetting.enable_s3_inventory,
+      id: inventory_id,
+      included_object_versions: "Current",
+      optional_fields: ["ETag"],
+      schedule: {
+        frequency: "Daily"
```
tgxworld: Because the inventory is only taken daily, our check for missing uploads should filter uploads that are created before the inventory file was created.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+  def update_bucket_inventory_configuration
+    @s3_helper.s3_client.put_bucket_inventory_configuration(
+      bucket: bucket_name,
+      id: inventory_id,
+      inventory_configuration: inventory_configuration,
+      use_accelerate_endpoint: false
+    )
+  end
+
+  def inventory_configuration
```
tgxworld: I don't think this method needs to be public.

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+    @s3_helper.list(File.join(inventory_path, "data")).each do |obj|
+      if obj.key.match?(/\.csv\.gz$/i)
+        objects << obj
+      end
+    end
+
+    objects
+  rescue Aws::Errors::ServiceError => e
+    Rails.logger.warn("Failed to list inventory from S3: #{e.message.presence || e.class.name}")
+    raise StorageError
```
tgxworld: Same as above, I don't feel like this custom error helps us in debugging the error in anyway.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+            display_name: "myname",
+            id: "12345example25102679df27bb0ae12b3f85be6f290b936c4393484be31bebcc",
+          },
+          size: 713193,
+          storage_class: "STANDARD",
+        }
+      ],
+      next_marker: "eyJNYXJrZXIiOiBudWxsLCAiYm90b190cnVuY2F0ZV9hbW91bnQiOiAyfQ=="
+    )
+    store.s3_helper.stubs(:s3_client).returns(client)
```
tgxworld: There is no need to stubb here, We can pass a custom `S3Helper` to `S3Store`. `S3Helper` accepts `client: client` in the options param.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def capture_stdout
```
tgxworld: This is being used in https://github.com/discourse/discourse/blob/7e4edcfae8a3c0e664b836ee7c5f28b47853a2f8/spec/components/migration/safe_migrate_spec.rb#L14-L22 as well so we should avoid duplicating code and extract the method into a helper or `rails_helper`.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  ensure
+    $stdout = old_stdout
+  end
+
+  it "will return recent inventory file name" do
+    expect(inventory.file.key).to eq("example1.csv.gz")
+  end
+
+  it "will raise storage error if inventory file not found" do
+    inventory.stubs(:file).returns(nil)
```
tgxworld: We're already stubbing `list_objects` so we should just make that stub return an empty list instead of stubbing again here.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    old_stdout = $stdout
+    io = StringIO.new
+    $stdout = io
+    yield
+    io.string
+  ensure
+    $stdout = old_stdout
+  end
+
+  it "will return recent inventory file name" do
```
tgxworld:
```suggestion
  it "should return the latest inventory file name" do
```

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    expect(inventory.file.key).to eq("example1.csv.gz")
+  end
+
+  it "will raise storage error if inventory file not found" do
```
tgxworld:
```suggestion
  it "should raise the right error if an inventory file is not found" do
```

lib/s3_inventory.rb
tgxworld:
```suggestion
            log upload.url
```

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        missing_uploads = model.joins("LEFT JOIN #{table_name} ON #{table_name}.etag = #{model.table_name}.etag").where("#{table_name}.etag is NULL")
+        missing_count = missing_uploads.count
+
+        if missing_count > 0
+          missing_uploads.find_each do |upload|
+            puts upload.url
+          end
+
+          puts "#{missing_count} of #{model.count} #{model.name.underscore.pluralize} are missing"
```
tgxworld:
```suggestion
          log "#{missing_count} of #{model.count} #{model.name.underscore.pluralize} are missing"
```

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  it "should display missing uploads correctly" do
+    CSV.foreach(csv_filename, headers: false) do |row|
+      Fabricate(:upload, etag: row[S3Inventory::CSV_ETAG_INDEX])
+    end
+
+    upload = Fabricate(:upload, etag: "ETag")
+    inventory.stubs(:unzip_archive)
+    inventory.stubs(:log)
```
tgxworld: We have a `silence_stdout` helper so we should use that instead of stubbing `log` away.

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    expect { inventory.list_missing }.to raise_error(S3Inventory::StorageError)
+  end
+
+  it "should display missing uploads correctly" do
+    CSV.foreach(csv_filename, headers: false) do |row|
+      Fabricate(:upload, etag: row[S3Inventory::CSV_ETAG_INDEX])
+    end
+
+    upload = Fabricate(:upload, etag: "ETag")
+    inventory.stubs(:unzip_archive)
```
tgxworld:
```suggestion
    inventory.expects(:unzip_archive)
```

spec/components/s3_inventory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    inventory.stubs(:log)
+    inventory.stubs(:csv_filename).returns(csv_filename)
```
tgxworld:
```suggestion
    inventory.stubs(:csv_filename).expects(csv_filename)
```

lib/s3_inventory.rb
```diff
@@ -0,0 +1,192 @@
...
+      @csv_filename = @archive_filename[0...-3]
+
+      FileUtils.mkdir_p(@tmp_directory)
+      copy_archive_to_tmp_directory
+      unzip_archive
+
+      begin
+        table_name = "#{inventory_id}_inventory"
+        connection = ActiveRecord::Base.connection.raw_connection
+        connection.exec("CREATE TEMP TABLE #{table_name}(key text UNIQUE, etag text PRIMARY KEY)")
```
tgxworld: Do we need an index on the `etag` column?

vinothkannans: I thought `PRIMARY KEY` itself an index. no?

spec/jobs/update_s3_inventory_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    SiteSetting.s3_secret_access_key = "def"
+    SiteSetting.enable_s3_inventory = true
+
+    store = FileStore::S3Store.new
+    @client = Aws::S3::Client.new(stub_responses: true)
+    store.s3_helper.stubs(:s3_client).returns(@client)
+    Discourse.stubs(:store).returns(store)
+  end
+
+  it "picks gravatar if system avatar is picked and gravatar was just downloaded" do
```
tgxworld: I don't think the description of this test is correct .

spec/jobs/update_s3_inventory_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    @client.expects(:put_bucket_policy)
+    @client.expects(:put_bucket_policy).with(
+      bucket: "bucket",
+      policy: "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Sid\":\"InventoryAndAnalyticsPolicy\",\"Effect\":\"Allow\",\"Principal\":{\"Service\":\"s3.amazonaws.com\"},\"Action\":[\"s3:PutObject\"],\"Resource\":[\"arn:aws:s3:::bucket/inventory/uploads/*\"],\"Condition\":{\"ArnLike\":{\"aws:SourceArn\":\"arn:aws:s3:::bucket\"},\"StringEquals\":{\"s3:x-amz-acl\":\"bucket-owner-full-control\"}}}]}"
+    )
+    @client.expects(:put_bucket_inventory_configuration)
```
tgxworld: Is this line duplicated?

spec/jobs/update_s3_inventory_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    SiteSetting.enable_s3_inventory = true
+
+    store = FileStore::S3Store.new
+    @client = Aws::S3::Client.new(stub_responses: true)
+    store.s3_helper.stubs(:s3_client).returns(@client)
+    Discourse.stubs(:store).returns(store)
+  end
+
+  it "picks gravatar if system avatar is picked and gravatar was just downloaded" do
+    @client.expects(:put_bucket_policy)
```
tgxworld: I don't think we need this line too.

lib/s3_inventory.rb
tgxworld: > The reason is to check missing objects using simple JOIN query.

This is quite a nice idea :+1:  Let's try it out and see how it works.

spec/jobs/update_s3_inventory_spec.rb
vinothkannans: No, it will be called twice for uploads and optimized images. Since both are identical I'm just checking the one (uploads) with the parameters.

spec/jobs/update_s3_inventory_spec.rb
vinothkannans: same as [above](https://github.com/discourse/discourse/pull/6867#discussion_r252837340).

PR #6871, opened 2019-01-10
FIX: Use correct version when generating file path for optimized image

vinothkannans:


lib/file_store/base_store.rb
```diff
@@ -116,7 +116,8 @@ def get_path_for_upload(upload)
 
     def get_path_for_optimized_image(optimized_image)
       upload = optimized_image.upload
-      extension = "_#{OptimizedImage::VERSION}_#{optimized_image.width}x#{optimized_image.height}#{optimized_image.extension}"
+      version = optimized_image.version || 1
```
tgxworld: Should it be `|| 1` or maybe `|| OptimizedImage::VERSION`?

PR #6872, opened 2019-01-10
FIX: show user avatar on User summary page

techAPJ:
Fixes the issue where user avatar is not showing up on "Most replied to", "Most liked by", "Most liked" section on [user summary page](https://meta.discourse.org/u/techAPJ/summary).

The UX was regressed in [this commit](https://github.com/discourse/discourse/commit/f0af61da415d0bbb4a49cb0eb461bd7bbeb7ea12#diff-6cc61be7ac2ff6a06ffe7dcc9dccee8d).

Basically the existing condition was just checking for an object and in case of user summary page the object was `UserSummary::UserWithCount` which does not have `avatar_template` method anymore. This commit ensures that we check for `User` object explicitly and if that's not the case fetch the `avatar_template` from User model.

### DIFF
--- a/spec/serializers/user_summary_serializer_spec.rb
+++ b/spec/serializers/user_summary_serializer_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+
+describe UserSummarySerializer do
+  it "returns expected data" do
+    UserActionCreator.enable
+    user = Fabricate(:user)
+    liked_post = create_post
+    PostAction.act(user, liked_post, PostActionType.types[:like])
+
+    guardian = Guardian.new(user)
+    summary = UserSummary.new(user, guardian)
+    serializer = UserSummarySerializer.new(summary, scope: guardian, root: false)
+    json = serializer.as_json
+
+    expect(json[:likes_given]).to be_present
+    expect(json[:likes_received]).to be_present
+    expect(json[:posts_read_count]).to be_present
+    expect(json[:topic_count]).to be_present
+    expect(json[:time_read]).to be_present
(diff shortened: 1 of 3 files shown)


techAPJ:
@eviltrout Note that I was adding a missing spec in `basic_user_serializer_spec.rb`. For testing the latest changes I added a new spec for `UserSummarySerializer` which will fail before this change and pass after this change.

eviltrout:
@techAPJ thanks for the additional test that shows the PR works. What confused me before is that the fix involves treating the object like a hash (using symbol methods like `:username`) even though we're no longer checking if it's a Hash. Those work because it's an `OpenStruct`.

I think the design mistake we've made here is to use `BasicUserSerializer` for something that isn't a `User`. 

I believe a better solution would be to change `UserWithCountSerializer` to not extend that and return the correct fields. Then `BasicUserSerializer` needn't have this odd logic.

spec/serializers/user_summary_serializer_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+require_dependency 'user'
+
+describe UserSummarySerializer do
+
+  context "works" do
```
eviltrout: I don't think it makes sense to have a context here. Contexts are only important for separating things.

techAPJ:
> I believe a better solution would be to change UserWithCountSerializer to not extend that and return the correct fields. Then BasicUserSerializer needn't have this odd logic.

@eviltrout Excellent point. 👍 

Made the required changes.

PR #6873, opened 2019-01-10
FEATURE: enable CSP by default on new sites

pmusaraj:
- adds migration to enable CSP for new sites

- removes EXPERIMENTAL from setting labels

- CSP report site setting set to default off

- adds CSP-related note to GTM setting

### DIFF
--- a/db/migrate/20190110142917_enable_content_security_policy_for_new_sites.rb
+++ b/db/migrate/20190110142917_enable_content_security_policy_for_new_sites.rb
@@ -1,0 +1,25 @@
+class EnableContentSecurityPolicyForNewSites < ActiveRecord::Migration[5.2]
+  def up
+    return if Rails.env.test?
+    return if row_exists?
+
+    if instance_is_new?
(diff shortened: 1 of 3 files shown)


coding-horror:
We should merge this

db/migrate/20190110142917_enable_content_security_policy_for_new_sites.rb
```diff
@@ -0,0 +1,25 @@
...
+  def down
+    # Don't undo, up method only enables CSP if row isn't already there and if instance is new
+  end
+
+  def row_exists?
+    return DB.query("SELECT 1 AS one FROM site_settings where name='content_security_policy'").present?
+  end
+
+  def instance_is_new?
+    post = DB.query("SELECT created_at FROM posts WHERE id=1")
```
davidtaylorhq: What if the post with id=1 has been deleted for some reason? I think it would be better to do something like (untested)

```SQL
SELECT created_at FROM posts ORDER BY created_at ASC LIMIT 1
```

PR #6874, opened 2019-01-10
DEV: s/Em/Ember

jjaffeux:


davidtaylorhq:
Can we enforce this in eslint/prettier? Otherwise I imagine we will naturally slip back into having a mix.

jjaffeux:
@davidtaylorhq not that I know eslint/prettier (at least fairly EASY way)

The place for this is Dangerfile... sadly the way we do things in Travis CI makes Danger almost useless. As we clone plugins before doing danger in a before_install hook. I need to investigate a way to make it more useful. Or feel free to have look if you have some time and/or are interested in this. I fully agree on the comment.

PR #6876, opened 2019-01-11
FIX: do not show PM topics when moving posts to an existing public topic

techAPJ:
https://meta.discourse.org/t/trying-to-move-post-to-existing-pm-results-in-discourse-invalidparameters/106117/6?u=techapj

### DIFF
--- a/app/assets/javascripts/discourse/components/choose-topic.js.es6
+++ b/app/assets/javascripts/discourse/components/choose-topic.js.es6
@@ -32,20 +32,22 @@ export default Ember.Component.extend({
       return;
     }
 
-    searchForTerm(title, { typeFilter: "topic", searchForId: true }).then(
-      function(results) {
-        if (results && results.posts && results.posts.length > 0) {
-          self.set(
-            "topics",
-            results.posts
-              .mapBy("topic")
-              .filter(t => t.get("id") !== currentTopicId)
-          );
-        } else {
-          self.setProperties({ topics: null, loading: false });
-        }
+    searchForTerm(title, {
+      typeFilter: "topic",
+      searchForId: true,
+      restrictToArchetype: "regular"
+    }).then(function(results) {
+      if (results && results.posts && results.posts.length > 0) {
+        self.set(
+          "topics",
+          results.posts
+            .mapBy("topic")
+            .filter(t => t.get("id") !== currentTopicId)
+        );
+      } else {
+        self.setProperties({ topics: null, loading: false });
       }
-    );
+    });
   }, 300),
 
   actions: {
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -227,7 +227,7 @@ class Search
     end
 
     # If the term is a number or url to a topic, just include that topic
-    if @opts[:search_for_id] && @results.type_filter == 'topic'
+    if @opts[:search_for_id] && (@results.type_filter == 'topic' || @results.type_filter == 'private_messages')
       if @term =~ /^\d+$/
         single_topic(@term.to_i)
       else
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/discourse/components/choose-topic.js.es6
```diff
@@ -32,20 +32,22 @@ export default Ember.Component.extend({
...
-              .mapBy("topic")
-              .filter(t => t.get("id") !== currentTopicId)
-          );
-        } else {
-          self.setProperties({ topics: null, loading: false });
-        }
+    searchForTerm(title, {
+      typeFilter: "topic",
+      searchForId: true,
+      restrict_to_archetype: "regular"
```
tgxworld: I don't think `searchForTerm` will process the `restrict_to_archetype` key in this object. Also, we're mixing snake_case with camelCase here which can be quite confusing.

lib/search.rb
```diff
@@ -629,7 +629,12 @@ def add_more_topics_if_expected
 
   # If we're searching for a single topic
   def single_topic(id)
-    post = Post.find_by(topic_id: id, post_number: 1)
+    if @opts[:restrict_to_archetype].present?
+      topic = Topic.find_by(id: id, archetype: @opts[:restrict_to_archetype] == "regular" ? Archetype.default : Archetype.private_message)
```
tgxworld:
```suggestion
      topic = Topic.find_by(id: id, archetype: @opts[:restrict_to_archetype] == Archetype.default ? Archetype.default : Archetype.private_message)
```

spec/components/search_spec.rb
```diff
@@ -341,12 +341,22 @@ def new_post(raw, topic = nil)
...
         expect(result.posts.length).to eq(1)
         expect(result.posts.first.id).to eq(post.id)
       end
+
+      context 'restrict_to_archetype' do
+        let(:personal_message) { Fabricate(:private_message_topic) }
+        let!(:p1) { Fabricate(:post, topic: personal_message, post_number: 1) }
+
+        it 'restricts result to specified archetype' do
+          result = Search.execute(personal_message.relative_url, search_for_id: true, type_filter: 'topic', guardian: Guardian.new(Fabricate(:admin)), restrict_to_archetype: 'regular')
```
tgxworld:
```suggestion
          result = Search.execute(personal_message.relative_url, search_for_id: true, type_filter: 'topic', guardian: Guardian.new(Fabricate(:admin)), restrict_to_archetype: ArcheType.default)
```

spec/components/search_spec.rb
```diff
@@ -341,12 +341,22 @@ def new_post(raw, topic = nil)
...
+        result = Search.execute(topic.relative_url, search_for_id: true, type_filter: 'topic')
         expect(result.posts.length).to eq(1)
         expect(result.posts.first.id).to eq(post.id)
       end
+
+      context 'restrict_to_archetype' do
+        let(:personal_message) { Fabricate(:private_message_topic) }
+        let!(:p1) { Fabricate(:post, topic: personal_message, post_number: 1) }
+
+        it 'restricts result to specified archetype' do
```
tgxworld: We need to add a test that shows searching for a PM when restricting the archetype to "private_message" works

lib/search.rb
```diff
@@ -629,7 +629,12 @@ def add_more_topics_if_expected
...
+      post = topic&.first_post
```
tgxworld: This ends up executing two queries as opposed to one. I think we can do a joins on the topics table to look for the post we need.

spec/components/search_spec.rb
techAPJ: Made the change in latest commit. Notice `ArcheType` vs `Archetype` :)

techAPJ:
@tgxworld Thanks for review. Updated the PR.

PR #6879, opened 2019-01-14
FEATURE: Consolidate likes notifications.

tgxworld:
screenshot from 2019-01-14 17-17-24

### DIFF
--- a/app/assets/javascripts/discourse/models/user-stream.js.es6
+++ b/app/assets/javascripts/discourse/models/user-stream.js.es6
@@ -30,14 +30,16 @@ export default RestModel.extend({
     "/user_actions.json?offset=%@&username=%@"
   ),
 
-  filterBy(filter, noContentHelpKey) {
+  filterBy(filter, noContentHelpKey, actingUsername) {
     this.setProperties({
       filter,
       itemsLoaded: 0,
       content: [],
-      noContentHelpKey: noContentHelpKey,
-      lastLoadedUrl: null
+      noContentHelpKey,
+      lastLoadedUrl: null,
+      actingUsername
     });
+
     return this.findItems();
   },
 
@@ -75,6 +77,10 @@ export default RestModel.extend({
     }
     if (this.get("noContentHelpKey")) {
       findUrl += "&no_results_help_key=" + this.get("noContentHelpKey");
+    }
+
+    if (this.get("actingUsername")) {
+      findUrl += `&acting_username=${this.get("actingUsername")}`;
     }
 
     // Don't load the same stream twice. We're probably at the end.
--- a/app/assets/javascripts/discourse/widgets/notification-item.js.es6
+++ b/app/assets/javascripts/discourse/widgets/notification-item.js.es6
@@ -16,6 +16,7 @@ import { iconNode } from "discourse-common/lib/icon-library";
 const LIKED_TYPE = 5;
 const INVITED_TYPE = 8;
 const GROUP_SUMMARY_TYPE = 16;
+export const LIKED_CONSOLIDATED_TYPE = 19;
 
 createWidget("notification-item", {
   tagName: "li",
@@ -61,6 +62,14 @@ createWidget("notification-item", {
       return userPath(data.display_username);
     }
 
+    if (attrs.notification_type === LIKED_CONSOLIDATED_TYPE) {
+      return userPath(
+        `${
+          this.currentUser.username
+        }/notifications/likes-received?acting_username=${data.display_username}`
+      );
+    }
+
     if (data.group_id) {
       return userPath(data.username + "/messages/group/" + data.group_name);
     }
@@ -77,7 +86,16 @@ createWidget("notification-item", {
       return this.attrs.fancy_title;
     }
 
-    const title = data.topic_title;
+    let title;
+
+    if (this.attrs.notification_type === LIKED_CONSOLIDATED_TYPE) {
+      title = I18n.t("notifications.liked_consolidated_description", {
+        count: parseInt(data.count)
+      });
+    } else {
+      title = data.topic_title;
+    }
+
     return Ember.isEmpty(title) ? "" : escapeExpression(title);
   },
 
@@ -95,9 +113,11 @@ createWidget("notification-item", {
 
     const username = formatUsername(data.display_username);
     const description = this.description();
+
     if (notificationType === LIKED_TYPE && data.count > 1) {
       const count = data.count - 2;
       const username2 = formatUsername(data.username2);
+
       if (count === 0) {
         return I18n.t("notifications.liked_2", {
           description,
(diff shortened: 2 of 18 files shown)


coding-horror:
😍😍😍😍😍😍😍😍

test/javascripts/widgets/user-menu-test.js.es6
```diff
@@ -15,6 +15,30 @@ widgetTest("basics", {
   }
 });
 
+widgetTest("notifications", {
+  template: '{{mount-widget widget="user-menu"}}',
+
+  test(assert) {
+    const $links = this.$(".notifications li a");
```
jjaffeux:
```suggestion
    const $links = find(".notifications li a");
```

config/locales/client.en.yml
```diff
@@ -1515,6 +1515,8 @@ en:
       liked_many:
         one: "<span>{{username}}, {{username2}} and 1 other</span> {{description}}"
         other: "<span>{{username}}, {{username2}} and {{count}} others</span> {{description}}"
+      liked_consolidated_description: "liked {{number}} of your posts"
```
gschlager: This needs proper pluralization:

```suggestion
      liked_consolidated_description:
        one: "liked {{count}} of your posts"
        other: "liked {{count}} of your posts"
```

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -77,7 +86,16 @@ createWidget("notification-item", {
       return this.attrs.fancy_title;
     }
 
-    const title = data.topic_title;
+    let title;
+
+    if (this.attrs.notification_type === LIKED_CONSOLIDATED_TYPE) {
+      title = I18n.t("notifications.liked_consolidated_description", {
+        number: data.count
```
gschlager: Should be `count` because the string needs to be pluralized.

```suggestion
        count: data.count
```

test/javascripts/widgets/user-menu-test.js.es6
```diff
@@ -15,6 +15,30 @@ widgetTest("basics", {
...
+    assert.ok(
+      $links[1].href.includes(
+        "/u/eviltrout/notifications/likes-received?acting_username=aquaman"
+      )
+    );
+
+    assert.equal(
+      $links[1].text,
+      `aquaman ${I18n.t("notifications.liked_consolidated_description", {
+        number: 5
```
gschlager: Should be `count` because the string needs to be pluralized.

```suggestion
        count: 5
```

config/locales/client.en.yml
tgxworld: The min for the site setting is 3 which is why  didn’t pluralize this.

app/assets/javascripts/discourse/models/user-stream.js.es6
```diff
@@ -30,14 +30,16 @@ export default RestModel.extend({
     "/user_actions.json?offset=%@&username=%@"
   ),
 
-  filterBy(filter, noContentHelpKey) {
+  filterBy(filter, noContentHelpKey, actingUsername) {
```
eviltrout: This API could be a good candidate for using object literals for the options. I always think of the call site:

`filterBy(123, 'some_default_key', user.username)`

In this case I can still somewhat understand it but I almost always prefer:

`filterBy(123, { defaultKey: 'some_key, actingUsername: user.username })`.

config/locales/client.en.yml
gschlager: I agree, it wouldn't be necessary for English. Unfortunately other languages have pluralization rules that are more complicated than the English ones.

jjaffeux: eg, Russian:

Quantity ends in 1 - As per position in the sentence.
Quantity ends in 2,3,4 - Genitive Singular.
Quantity ends in 5,6,7,8,9,0,-надцать - Genitive Plural
General Quantity - Genitive Case (Singular or Plural depending on context.)
Quantity not specified - Use the case that is appropriate to the sentence position.

source: http://www.russianlessons.net/lessons/lesson11_main.php

tgxworld: Ahh icic. Thanks for clarifying, I'll make the changes.

spec/models/notification_spec.rb
```diff
@@ -250,6 +250,32 @@ def process_alerts(post)
...
+      expect(Notification.get_liked_by(user)).to eq([])
+
+      expect do
+        PostAlerter.post_created(Fabricate(:basic_reply,
+          user: user,
+          topic: post.topic
+        ))
+
+        PostAction.act(user, post, PostActionType.types[:like])
+      end.to change { Notification.count }.by(3)
```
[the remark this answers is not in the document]
tgxworld: I don't care about the content of the other two notifications but I need to assert for the count to enforce that the test is filtering for a particular notification from 3 available ones. That is important because if we're filtering for a single notification when only that notification has been created, someone replacing the method with `Notification.all` will make the test pass.

app/assets/javascripts/discourse/models/user-stream.js.es6
tgxworld: I feel that it makes the resulting function harder to understand considering that the function is expecting 3 exact inputs.

`filterBy(filter, noContentHelpKey, actingUsername)`

vs

`filterBy(filter, options)`

Since the latter requires us to read through the function to figure out the keys in the options object is going to be, I find it less readable then before.

app/models/notification.rb
```diff
@@ -13,6 +13,12 @@ class Notification < ActiveRecord::Base
   scope :visible , lambda { joins('LEFT JOIN topics ON notifications.topic_id = topics.id')
     .where('topics.id IS NULL OR topics.deleted_at IS NULL') }
 
+  scope :get_liked_by, ->(user) {
+    where("data::json ->> 'original_username' = ?", user.username_lower)
```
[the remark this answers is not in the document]
tgxworld: Yup this can be slow because we need to cast to `json` in Postgres. Technically, `Notification#data` needs to be migrated to use a `jsonb` column type but that is going to be a whole can of worms. The consideration I made here is that the use case restricts this query to `liked` notifications that have been made within the last 2 hours. Therefore, we'll have to cast the string to json for a limited set of records.

app/assets/javascripts/discourse/models/user-stream.js.es6
eviltrout: I see what you mean, but you aren't using `noContentHelpKey` or `actingUsername` in any meaningful way here. You are passing them through to `setProperties`.

It would be simpler and less code if you took an argument called `props`. You could either merge it into the literal before calling `setProperties` or could call `setProperties` a second time with the props.

spec/models/notification_spec.rb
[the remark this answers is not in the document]
eviltrout: That makes sense, thanks.

app/assets/javascripts/discourse/models/user-stream.js.es6
tgxworld: Followed up in https://github.com/discourse/discourse/commit/40533bc570685e3102f2df81d92d7a3c52f9bcb8.

PR #6881, opened 2019-01-14
FIX: Fix URL to user profile.

nbianca:


plugins/poll/test/javascripts/acceptance/polls-test-mobile.js.es6
```diff
@@ -0,0 +1,65 @@
...
+  server.get("/t/13.json", () => { // eslint-disable-line no-undef
+    return [200, { "Content-Type": "application/json" }, Fixtures["t/13.json"]];
+  });
+
+  // prettier-ignore
+  server.get("/polls/voters.json", request => { // eslint-disable-line no-undef
+    let body = {};
+
+    if (
+      request.queryParams["post_id"] === "16" &&
```
eviltrout: You needn't quote properties in Javascript unless they contain characters that don't work with dot notation. `request.queryParameters.poll_id` is preferred.

PR #6883, opened 2019-01-15
rake emails:test fix

Supermathie:
https://meta.discourse.org/t/amazon-ses-must-issue-a-starttls-command-first/106372/11?u=supermathie


Supermathie:
I'm not sure if we want to spit out a message if we're bypassing this check. It might just add noise?

I opted to *not* add a message.

PR #6884, opened 2019-01-15
DEV: replaces DOMNodeInserted by the better MutationObserver

jjaffeux:
@eviltrout I noticed we had this violation in chrome console on mobile:

```
discourse/lib/safari-hacks:196
[Violation] Added synchronous DOM mutation listener to a 'DOMNodeInserted' event. Consider using MutationObserver to make the page more responsive.
```

So I went ahead and did this. It seems rather safe https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

Could you have a look at some point please ?

### DIFF
--- a/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
+++ b/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
@@ -191,7 +191,14 @@ function positioningWorkaround($fixedElement) {
     });
   }, 100);
 
-  fixedElement.addEventListener("DOMNodeInserted", checkForInputs);
+  const config = {
+    childList: true,
+    subtree: true,
+    attributes: false,
+    characterData: false
+  };
+  const observer = new MutationObserver(checkForInputs);
+  observer.observe(fixedElement, config);
 }
 
 export default positioningWorkaround;


eviltrout:
Seems fine to me, but I believe @SamSaffron  was the one who originally added that safari workaround so you might want him to take a peek.

SamSaffron:
I am not sure the mutation observer will be cleaned up unless we call disconnect. The events get auto cleaned up per: 

https://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory

You are going to need to test this ultra carefullyt to confirm we do remove the observer from memory when willDestroyElement is called on composer-body.js

jjaffeux:
> I am not sure the mutation observer will be cleaned up unless we call disconnect. The events get auto cleaned up per:
> 
> https://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory
> 
> You are going to need to test this ultra carefully to confirm we do remove the observer from memory when willDestroyElement is called on composer-body.js

Sorry for the delay I missed your answer.

From the documentation: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/disconnect#Usage_notes

```
If the element being observed is removed from the DOM and then subsequently released by the browser's garbage collection mechanism, the MutationObserver is likewise deleted.
```

SamSaffron:
OK feel free to merge, after you merge in, test posting on the deployed side via your iPhone.

PR #6886, opened 2019-01-16
rm period

angusmcleod:
``topic.suggest_create_topic`` ends in a question mark:

```
Why not create a topic?
```

Meaning that this sentence has double punctuation if the period is included in the template.


ZogStriP:
Thanks!. 😉

PR #6888, opened 2019-01-16
FIX: phpbb3 import - Display Post Id while importing

aurawindsurfing:
Displays posts id while importing for better error reference. At the moment it displays only table row which is not the same as post_id and it is harder for user to track.

### DIFF
--- a/script/import_scripts/phpbb3/importers/post_importer.rb
+++ b/script/import_scripts/phpbb3/importers/post_importer.rb
@@ -25,6 +25,8 @@ module ImportScripts::PhpBB3
       is_first_post = row[:post_id] == row[:topic_first_post_id]
 
       attachments = import_attachments(row, user_id)
+      
+      puts "Processing post id: #{row[:post_id]}"
 
       mapped = {
         id: row[:post_id],


ZogStriP:
Thanks but it's going to be very verbose. Why would you need to show the imported post id when everything is good? I think it would be much better to show the imported post id only when an error is happening (and I think we're already displaying it).

aurawindsurfing:
HI! At the moment it does not display post_id only table row_id which is not the same when posts do not have continuous numbering. The errors I encounter happen in text_processing and will not show post_id. 

IMHO it does not matter if the output is verbose, it matters if import works and is easy to troubleshoot.

ZogStriP:
> HI! At the moment it does not display post_id on an error only table row_id which is not the same when posts do not have continuous numbering.

Why not change that then?

aurawindsurfing:
Do you expect me to go and change every post_id on a few million posts table and all its references only because posts ids are not incremental?

aurawindsurfing:
I'm not really familiar with rails tbh. To change whole exception handling only for that small issue that you encounter only once while doing your import seems like a bit over the top job to me, unless someone familiar with rails can do it in 5 minutes.

To me this PR's does the job of getting you one step forward with problems in your import and that is all this script is for. I understand it does not look pretty when you do that.

gschlager:
You could actually make this quite helpful and less verbose by moving the `puts` into a `rescue` block at the end of the `map_post` method and outputting a message that tells the user what has gone wrong.

gschlager:
I added my proposed solution in https://github.com/discourse/discourse/commit/24369a81663b25c16b8fccd7e6e33fd4a44284f8

PR #6894, opened 2019-01-17
UX: admin badge page improvements 

hnb-ku:
1. adds intro section to admin badges
2. makes badge list scrollable
3. makes admin badges "usable" on mobile
4. CSS cleanup


jjaffeux:
@hnb-ku I'm having a look at this soon and will merge it following

app/assets/stylesheets/common/admin/badges.scss
```diff
@@ -1,31 +1,91 @@
...
+    padding: 10px 0;
+    border-top: 1px solid $primary-low;
+    border-bottom: 1px solid $primary-low;
+    .badges-heading {
+      margin: 0;
+      padding: 0 10px;
+    }
+  }
+  .content-list {
+    ul {
```
jjaffeux: would prefer a class here

app/assets/stylesheets/common/admin/badges.scss
```diff
@@ -1,31 +1,91 @@
...
+      height: 70vh;
+      overflow-y: scroll;
+      border-right: 1px solid $primary-low;
+      border-left: 1px solid $primary-low;
+      li {
```
jjaffeux: would prefer a class here

app/assets/stylesheets/common/admin/badges.scss
```diff
@@ -122,28 +178,26 @@
...
-  .badge-grouping {
-    .badge-grouping-name-input {
-      margin: 0;
+// mobile specific styles - mostly commmon style overrides
+// TODO move to mobile sheet once mobile view has a seprate template.
+.mobile-view {
+  .admin-badges {
+    .content-list {
+      flex: 0 0 100%;
+      ul {
```
jjaffeux: would prefer a class here

app/assets/stylesheets/common/admin/badges.scss
```diff
@@ -122,28 +178,26 @@
...
-    .btn {
-      padding: 3px 6px;
+    .badge-intro {
+      flex: 0 1 75%;
+    }
+    .current-badge {
+      margin: 20px 0;
+      width: 100%;
+    }
+    input[type="text"] {
```
jjaffeux: would prefer a class here

jjaffeux:
Few comments, make sure you test this on IE11 please.

app/assets/stylesheets/common/admin/badges.scss
hnb-ku: I'm with you 100% on this but I'm afraid that adding classes to the inputs here is only a temporary fix to the problem. It might be better to do this systematically for all inputs where we can define the different widths based on a class like so (rough example)

```
.w25 {
  width: 25%;
}
.w50 {
  width: 25%;
}
.w75 {
  width: 25%;
}
.w100 {
  width: 100%;
}
```
<input class="w25">
<input class="w50">
<input class="w75">
<input class="w100">
```

or some refined version of that.

jjaffeux: feel free to open an internal topic to discuss this with kris

PR #6895, opened 2019-01-17
DEV: Do the draft conflict check async.

nbianca:


SamSaffron:
Merging this in cause it feels rather urgent, but I would like an acceptance test here, historically we have had lots of regressions around drafts it is super hard to get it to work perfect.

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -848,17 +847,26 @@ export default Ember.Controller.extend({
...
             }
-
+            opts.draftSequence = data.draft_sequence;
             self._setModel(composerModel, opts);
           })
           .then(resolve, reject);
       }
+      // otherwise, do the draft check async
+      else if (!opts.draft && !opts.skipDraftCheck) {
+        Draft.get(opts.draftKey)
```
SamSaffron: technically there is a timing issue here, if get for draft is mega delayed

nbianca: That was pretty much the point of these changes: open composer asap and do not wait for draft conflict check, because the response might be delayed.

PR #6898, opened 2019-01-17
FIX: since Ember 3 interceptClick was not working on buffered links

jjaffeux:


eviltrout:
This never used to be required, as those links were handled by the global link handler. Why does it need to happen in here now?

jjaffeux:
@eviltrout Don’t have an answer so far...

What I know:

- This is correctly initialized https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/initializers/click-interceptor.js.es6

- https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/initializers/click-interceptor.js.es6#L7 is never trirggering on mobile
- another event like `touchstart` would trigger it

eviltrout:
I'd like to take a look at it myself if that's OK. I'll try to get to it soon.

jjaffeux:
Sure, anyways it's making a test fail. I'm reading more at ember internals meanwhile.

PR #6899, opened 2019-01-17
PERF: delete potentially large associated tables before user_destroyer.destroy transaction

majakomel:


app/services/user_destroyer.rb
```diff
@@ -21,6 +21,8 @@ def destroy(user, opts = {})
     # default to using a transaction
     opts[:transaction] = true if opts[:transaction] != false
 
+    prepare_for_destroy(user) if opts[:prepare_for_destroy] == true
```
tgxworld: If this happens outside the transaction, what would happen if the transaction fails?

SamSaffron: This is fine, if it halfway fails, user can still be deleted, so all that will happen is you will have a half deleted user when you wanted a full deleted user

app/services/user_destroyer.rb
```diff
@@ -111,6 +113,16 @@ def destroy(user, opts = {})
 
   protected
 
+  def prepare_for_destroy(user)
+    PostAction.where(user_id: user.id).delete_all
+    UserAction.where(user_id: user.id).delete_all
```
SamSaffron: Minor change, have a look at the user model, there are 2 other columns to match, feel free to merge once you change that, change is good!

PR #6907, opened 2019-01-18
FIX: make composer full screen shortcut work when inputs have focus

pmusaraj:
This is a fix for https://github.com/discourse/discourse/pull/6839 using a mousetrap.js plugin that allows binding keyboard shortcuts globally. 

* adds `global` parameter to keyboard shortcut definitions for `shift+f11` (composer full screen) and `ctrl+alt+f` (toggle search)
* sets focus on composer editor when going fullscreen


SamSaffron:
I like this fix a lot but we should wait till 2.3 here.

SamSaffron:
I would love to see this change land now, @pmusaraj  can you rebase and merge?

PR #6909, opened 2019-01-21
FEATURE: Add page for all group membership requests. 

nbianca:
Work in progress. Needs tests.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/group-requests.js.es6
+++ b/app/assets/javascripts/discourse/controllers/group-requests.js.es6
@@ -82,8 +82,8 @@ export default Ember.Controller.extend({
     }
   },
 
-  replyRequest(data) {
-    ajax(`/groups/${this.get("model.id")}/membership_request.json`, {
+  handleRequest(data) {
+    ajax(`/groups/${this.get("model.id")}/handle_membership_request.json`, {
       data,
       type: "PUT"
     }).catch(popupAjaxError);
@@ -95,7 +95,7 @@ export default Ember.Controller.extend({
     },
 
     acceptRequest(user) {
-      this.replyRequest({ user_id: user.get("id"), accept: true });
+      this.handleRequest({ user_id: user.get("id"), accept: true });
       user.setProperties({
         request_accepted: true,
         request_denied: false
@@ -112,7 +112,7 @@ export default Ember.Controller.extend({
     },
 
     denyRequest(user) {
-      this.replyRequest({ user_id: user.get("id") });
+      this.handleRequest({ user_id: user.get("id") });
       user.setProperties({
         request_accepted: false,
         request_denied: true
--- a/app/assets/javascripts/discourse/controllers/group-requests.js.es6
+++ b/app/assets/javascripts/discourse/controllers/group-requests.js.es6
@@ -96,12 +96,27 @@ export default Ember.Controller.extend({
 
     acceptRequest(user) {
       this.replyRequest({ user_id: user.get("id"), accept: true });
-      user.set("status", "groups.requests.accepted");
+      user.setProperties({
+        request_accepted: true,
+        request_denied: false
+      });
+    },
+
+    undoAcceptRequest(user) {
+      ajax("/groups/" + this.get("model.id") + "/members.json", {
+        type: "DELETE",
+        data: { user_id: user.get("id") }
+      }).then(() => {
+        user.set("request_undone", true);
+      });
     },
 
     denyRequest(user) {
       this.replyRequest({ user_id: user.get("id") });
-      user.set("status", "groups.requests.denied");
+      user.setProperties({
+        request_accepted: false,
+        request_denied: true
+      });
     }
   }
 });
--- a/app/assets/javascripts/discourse/controllers/group-requests.js.es6
+++ b/app/assets/javascripts/discourse/controllers/group-requests.js.es6
@@ -1,0 +1,107 @@
+import { ajax } from "discourse/lib/ajax";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+import Group from "discourse/models/group";
+import {
+  default as computed,
+  observes
+} from "ember-addons/ember-computed-decorators";
+import debounce from "discourse/lib/debounce";
+
+export default Ember.Controller.extend({
+  queryParams: ["order", "desc", "filter"],
+  order: "",
+  desc: null,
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/discourse/controllers/group-requests.js.es6
```diff
@@ -0,0 +1,97 @@
...
+      {
+        order: this.get("order"),
+        desc: this.get("desc"),
+        filter: this.get("filter"),
+        requesters: true
+      }
+    ).then(result => {
+      const requesters = (!force && this.get("model.requesters")) || [];
+      requesters.addObjects(result.members.map(m => Discourse.User.create(m)));
+      this.set("model.requesters", requesters);
```
jjaffeux: you can group this with the following setProperties

app/assets/javascripts/discourse/controllers/group-requests.js.es6
```diff
@@ -0,0 +1,97 @@
...
+    if (
+      !force &&
+      this.get("count") &&
+      this.get("model.requesters.length") >= this.get("count")
+    ) {
+      this.set("application.showFooter", true);
+      return;
+    }
+
+    this.set("loading", true);
```
jjaffeux: setProperties

SamSaffron:
Can we merge this in now, or is more work pending?

ZogStriP:
@nbianca can you fix the merge conflicts?

SamSaffron:
can I see some screen shots?

nbianca:
Sure! Here they are:
Screenshot 2019-03-13 at 22 49 44

Screenshot 2019-03-13 at 22 49 57
Screenshot 2019-03-13 at 22 50 09

app/controllers/groups_controller.rb
```diff
@@ -362,6 +412,8 @@ def request_membership
     group = find_group(:id)
     group_name = group.name
 
+    GroupRequest.create(group: group, user: current_user, reason: params[:reason])
```
SamSaffron: technically this should be `create!` cause you are not checking for errors

SamSaffron:
Since we already have a custom UI I am not sure we need the PM per request maybe only a notification and weekly PM to remind. That said, we can follow up on that. 

At a minimum the PM should include a link to the group page so when you get it you simply click on the link and approve the user.

```
---
<a href=''>handle membership request</a>
```

nbianca:
The link to requests does not work really well because the user will also see the message.

db/migrate/20190117191606_create_group_requests.rb
```diff
@@ -0,0 +1,11 @@
+class CreateGroupRequests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :group_requests do |t|
+      t.integer :group_id
+      t.integer :user_id
```
tgxworld: We'll probably need an index or two here for `user_id` and `group_id`

app/controllers/groups_controller.rb
```diff
@@ -294,6 +327,23 @@ def add_members
     end
   end
 
+  def membership_request
+    group = Group.find(params[:id])
+    guardian.ensure_can_edit!(group)
+
+    user = User.find(params[:user_id])
```
tgxworld: This is going to raise an error if the `id` is invalid. As a result, the next line isn't going to work.

app/controllers/groups_controller.rb
```diff
@@ -294,6 +327,23 @@ def add_members
     end
   end
 
+  def membership_request
+    group = Group.find(params[:id])
```
tgxworld: Same as https://github.com/discourse/discourse/pull/6909/files#r267216316

app/controllers/groups_controller.rb
```diff
@@ -362,16 +412,27 @@ def request_membership
     group = find_group(:id)
     group_name = group.name
 
+    GroupRequest.create!(group: group, user: current_user, reason: params[:reason])
```
tgxworld: I think this should probably happen in the same transaction as `PostCreate.create!`

db/migrate/20190117191606_create_group_requests.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :reason
```
tgxworld: Is this a custom message? We might need `text` here instead.

app/controllers/groups_controller.rb
```diff
@@ -294,6 +327,23 @@ def add_members
     end
   end
 
+  def membership_request
```
tgxworld: hmm the method name here  doesn't seem to reflect what the method actually does:
1. Adds a user to the group 
2. Delete all the user's existing requests to a group.

app/controllers/groups_controller.rb
nbianca: There is a transaction inside PostCreator and it is not something I would want to pollute with group requests.

tgxworld: O I don’t mean that it should use the transaction inside. I meant that it should happen in the same transaction which you can wrap around post creator

SamSaffron:
@tgxworld is this good to merge?

app/controllers/groups_controller.rb
```diff
@@ -294,6 +327,23 @@ def add_members
     end
   end
 
+  def handle_membership_request
+    group = Group.find_by(id: params[:id])
+    guardian.ensure_can_edit!(group)
```
tgxworld: If `group` is `nil`, we'll end up raising an `Discourse::InvalidAccess` error here instead of a `Discourse::InvalidParameter`.

app/controllers/groups_controller.rb
```diff
@@ -294,6 +327,23 @@ def add_members
...
+
+    user = User.find_by(id: params[:user_id])
+    raise Discourse::InvalidParameters.new(:user_id) if user.blank?
+
+    if params[:accept]
+      group.add(user)
+      GroupActionLogger.new(current_user, group).log_add_user_to_group(user)
+    end
+
+    GroupRequest.where(group_id: group.id, user_id: user.id).delete_all
```
tgxworld: I think the addition of the user of the group + logging + deletion of the group request should happen in the same DB transaction.

tgxworld:
@nbianca I left some final comments but otherwise feel free to merge once those comments have been resolved :+1:

PR #6912, opened 2019-01-21
SECURITY: fix possible XSS with badges

jjaffeux:


eviltrout:
Shouldn't we add a test since it's SECURITY?

PR #6916, opened 2019-01-21
FEATURE: Import and export themes in a .tar.gz format

davidtaylorhq:
## User-facing changes
- Removes support for `.dcstyle.json` exports. This has been partially broken for some time
- Allows exporting themes in a `.tar.gz` format. The file/folder structure is the same as used for git-sourced themes
- Allows importing themes in a `.tar.gz`. We already had this functionality via the `discourse_theme` CLI, but this adds the functionality to the UI, and fixes a number of bugs
- Improves display of "non-git remote themes". Previously imported `.tar.gz` themes would show the "update" button, which would throw an error when clicked

## Backwards Compatability
- The partially-broken `.dcstyle.json` importer remains in place, so that users have the opportunity to easily convert their old themes. We should plan to remove this at the next major release.

## Technical Notes
- The `tar.gz` file is created by building a temporary folder on the host, building an archive, then serving it to the user. Safeguards are in place to prevent malicious user input writing files outside of the temporary directory.
- The mapping of folder/file names to and from `ThemeField`s is now handled in one location, so it should be relatively easy to add support for more files in future.
- All color schemes which 'belong to' the theme are included in `about.json`. Additionally, the presently selected color scheme is included, even if it 'belongs to' a different theme. (there is currently no way to change color scheme ownership within the UI, so this seems a good compromise for now)


app/models/remote_theme.rb
```diff
@@ -14,15 +16,22 @@ class RemoteTheme < ActiveRecord::Base
...
     importer.import!
 
-    theme_info = JSON.parse(importer["about.json"])
-
-    theme = Theme.find_by(name: theme_info["name"])
+    theme_info = RemoteTheme.extract_theme_info(importer)
+    theme = Theme.find_by(name: theme_info["name"]) if match_theme
     theme ||= Theme.new(user_id: user&.id || -1, name: theme_info["name"])
 
+    theme.component = [true, "true"].include?(theme_info["component"])
```
eviltrout: Slightly simpler to do `theme_info["component"].to_s == "true"`

app/models/remote_theme.rb
```diff
@@ -214,6 +205,11 @@ def github_repo_url
       "https://github.com/#{org_repo}"
     end
   end
+
+  def is_git
```
eviltrout: Minor but in ruby we prefer boolean methods to end with `?`. Also the body could be `remote_url.present?`

app/models/theme_field.rb
```diff
@@ -277,6 +277,82 @@ def target_name
     Theme.targets.invert[target_id].to_s
   end
 
+  class ThemeFileMatcher
+    OPTIONS = %i{name type target}
+    def initialize(regex, canonical:, target:, name:, type:)
+      @allowed_values = {}
+      @allowed_values[:name] = Array(name) if name
```
eviltrout: you can use `[name]` instead of `Array(name)` to be more "ruby"-like here. (and for the following two lines too!)

app/models/theme_field.rb
```diff
@@ -277,6 +277,82 @@ def target_name
...
+        hash[option] = @allowed_values[option][0] if @allowed_values[option].length == 1
+        hash[option] = match[option] if hash[option].nil?
+      end
+      hash
+    end
+
+    def filename_from_opts(opts)
+      is_match = true
+      OPTIONS.each do |option|
+        next if @allowed_values[option] == nil
```
eviltrout: Could do this in one line if you prefer:

`next if (@allowed_values[option] || []).include?(opts[option])`

spec/components/theme_store/tgz_exporter_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      cs2.update(theme_id: theme.id)
+
+      theme.save!
+    end
+  end
+
+  let(:dir) do
+    tmpdir = Dir.tmpdir
+    dir = "#{tmpdir}/#{SecureRandom.hex}"
+    `mkdir #{dir}`
```
eviltrout: Any reason you aren't using `FileUtils` for creating and removing the directories? I know these are tests but it still seems safer to me.

app/models/theme_field.rb
davidtaylorhq: I think the behaviour is slightly different. Here I wanted to have the option of passing either a single symbol, or an array of symbols into the function. Later on it should always be an array. Using the `[name]` syntax would end up with a two-dimensional array:
```
[1] pry(main)> test = :hello
=> :hello
[2] pry(main)> [test]
=> [:hello]
[3] pry(main)> test2 = [:hello, :world]
=> [:hello, :world]
[4] pry(main)> [test2]
=> [[:hello, :world]]
```
Using `Array()`:
```
[5] pry(main)> Array(test)
=> [:hello]
[6] pry(main)> Array(test2)
=> [:hello, :world]
```

app/models/theme_field.rb
davidtaylorhq: The idea was that it should always match when the value is `nil`. But this was a very confusing piece of code, so I've refactored it to be clearer, and added a couple of comments.

spec/components/theme_store/tgz_exporter_spec.rb
davidtaylorhq: No reason, other than copy/paste from some existing theme import specs. I've switched to using `FileUtils` 👍

app/models/theme_field.rb
eviltrout: I see. However, then I think the code has a clarity problem. It's not obvious by reading the signature that `name` could be a single symbol or an array of symbols. Maybe something like `names` or `types` would make more sense here.

davidtaylorhq: :+1: I've pluralised the three arguments to make it clearer. Thanks

app/models/theme_field.rb
davidtaylorhq: The `names`/`targets`/`types` values can only be set via the initialiser, where they are required keyword arguments. So it's impossible to leave them "unset".

PR #6924, opened 2019-01-22
FEATURE: List unused theme components

davidtaylorhq:
Makes it easier to see which components are not being used by any theme:

screenshot 2019-01-22 at 13 25 39

### DIFF
--- a/app/assets/javascripts/admin/adapters/theme.js.es6
+++ b/app/assets/javascripts/admin/adapters/theme.js.es6
@@ -14,6 +14,10 @@ export default RestAdapter.extend({
       let mapped = theme.get("child_themes") || [];
       mapped = mapped.map(t => map[t.id]);
       theme.set("childThemes", mapped);
+
+      let mappedParents = theme.get("parent_themes") || [];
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/admin/adapters/theme.js.es6
```diff
@@ -14,6 +14,10 @@ export default RestAdapter.extend({
       let mapped = theme.get("child_themes") || [];
       mapped = mapped.map(t => map[t.id]);
       theme.set("childThemes", mapped);
+
+      let mappedParents = theme.get("parent_themes") || [];
```
eviltrout: I wonder if it would be possible to use the rest serializer to automatically wire these objects up using the identity map. Not urgent, just a thought.

davidtaylorhq: Agreed. The other major problem here is that we currently ship ALL the themes and ALL the components and ALL of their source code in a single payload. This needs to be refactored so we only load basic information for the index page, then load a separate payload for each theme.

This has the potential to introduce issues on the JS side, so I'd rather do it after the release I think.

PR #6928, opened 2019-01-23
Replace embed on site link

sau226:
Move embed on site link to meta.discourse per the blog that was previously used (users are redirected to this meta.discourse link now).


jjaffeux:
👍

PR #6935, opened 2019-01-23
FIX: Lazy Loading: copy only present srcset

fuzzykiller:
When showing a lazy-loaded image, copy the `srcset` property only when it is actually set. `copyImg.srcset = copyImg.srcset` is not actually a noop but creates an empty `srcset`, changing content security rules on the image.


eviltrout:
Nice find!

PR #6938, opened 2019-01-23
FIX: allow sending PMs to staff via flag even when PMs are disabled

arpitjalan:
FIX: allow sending PMs to staff via flag even if the user trust level is insufficient

### DIFF
--- a/lib/topic_creator.rb
+++ b/lib/topic_creator.rb
@@ -261,7 +261,7 @@ class TopicCreator
 
   def check_can_send_permission!(topic, obj)
     unless @opts[:skip_validations] ||
-      @guardian.can_send_private_message?(obj, notify_moderators: (topic.present? && topic.subtype == TopicSubtype.notify_moderators))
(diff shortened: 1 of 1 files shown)


coding-horror:
nice @eviltrout can you review? How risky is this? Would be nice to get this in as it affected T-nation etc

lib/topic_creator.rb
```diff
@@ -260,7 +260,10 @@ def add_groups(topic, groups)
   end
 
   def check_can_send_permission!(topic, obj)
-    rollback_with!(topic, :cant_send_pm) unless @opts[:skip_validations] || @guardian.can_send_private_message?(obj)
+    unless @opts[:skip_validations] ||
+      @guardian.can_send_private_message?(obj, notify_moderators: (topic.present? && topic.subtype == TopicSubtype.notify_moderators))
```
ZogStriP:
```suggestion
      @guardian.can_send_private_message?(obj, notify_moderators: topic&.subtype == TopicSubtype.notify_moderators))
```

PR #6939, opened 2019-01-24
Compose PM Button Class Fix

angusmcleod:
The compose-pm button in user profiles was inserted as an inline component, wrapping the icon and the label. This meant that the discourse button class system was not being applied correctly, i.e. this button has the class "no-text" when it has text. This converts the button to a standard component.

screenshot at jan 24 17-18-10


jjaffeux:
👍  thanks

PR #6941, opened 2019-01-24
FIX: displays user-card when clicking on avatar of post small action

jjaffeux:


app/assets/javascripts/discourse/widgets/post-small-action.js.es6
```diff
@@ -102,7 +102,7 @@ export default createWidget("post-small-action", {
     contents.push(
       avatarFor.call(this, "small", {
         template: attrs.avatar_template,
-        username: attrs.avatar,
```
eviltrout: Wow, that's a weird bug. How long has that been there?

jjaffeux: Don't you remember doing it 3 years ago ?

screenshot 2019-01-24 at 23 01 29

:trollface:

jjaffeux: In fact it was probably working when you did it maybe ? maybe `attrs.avatar` was different at that time ?

eviltrout: Totally possible.

PR #6942, opened 2019-01-24
FIX: ensures correct scroll position of textarea after autocomplete

jjaffeux:


eviltrout:
BTW I see that we got the meta link here, but the PR/Commit itself did not make clear what this is fixing, only how it it is fixed. An additional line saying why would go a long way for future code explorers who stumble upon it.

PR #6944, opened 2019-01-24
FEATURE: Support additional metadata in theme about.json

davidtaylorhq:
New `about.json` fields (all optional):
 - `authors`: An arbitrary string describing the theme authors
 - `theme_version`: An arbitrary string describing the theme version
 - `minimum_discourse_version`: Theme will be auto-disabled for lower versions. Must be a valid version descriptor.
 - `maximum_discourse_version`: Theme will be auto-disabled for lower versions. Must be a valid version descriptor.

A localized description for a theme can be provided in the language files under the `theme_metadata.description` key

The admin UI has been re-arranged to display this new information, and give more prominence to the remote theme options.

screenshot 2019-01-22 at 17 54 38


jjaffeux:
I know this is not coming from this PR but a `remoteTheme` alias to `model.remote_theme` would be cleaner for the template

app/models/remote_theme.rb
```diff
@@ -16,6 +17,8 @@ class ImportError < StandardError; end
     joins("JOIN themes ON themes.remote_theme_id = remote_themes.id").where.not(remote_url: "")
   }
 
+  validates_format_of :minimum_discourse_version, :maximum_discourse_version, with: /\A\d+\.\d+\.\d+(\.beta\d+)?\z/, allow_nil: true
```
eviltrout: I feel like this Regexp might find a better home in `lib/version.rb`. Would encourage more re-use.

app/models/remote_theme.rb
```diff
@@ -3,6 +3,7 @@
 require_dependency 'upload_creator'
 
 class RemoteTheme < ActiveRecord::Base
+  METADATA_PROPERTIES = %i{license_url about_url authors theme_version minimum_discourse_version maximum_discourse_version}
```
eviltrout: Super low priority but this line seems very long.

app/models/theme.rb
```diff
@@ -125,16 +121,23 @@ def self.clear_default!
   end
 
   def self.transform_ids(ids, extend: true)
-    return [] if ids.blank?
+    get_set_cache "transformed_ids_#{ids.join("_")}" do
+      next [] if ids.blank?
+
+      ids.uniq!
+      parent = ids.first
```
eviltrout: One thing to consider here is using `Array#shift`. It will return the first value and mutate the array to remove it which is I think cleaner than the way you're doing it.

app/models/theme.rb
```diff
@@ -125,16 +121,23 @@ def self.clear_default!
...
 
-    ids.uniq!
-    parent = ids.first
+      all_ids = [parent, *components]
 
-    components = ids[1..-1]
-    components.push(*components_for(parent)) if extend
-    components.sort!.uniq!
+      enabled_ids = Theme.where(id: all_ids).includes(:remote_theme)
+        .select(&:enabled?).map(&:id)
```
eviltrout: Trivial but `#pluck` is slightly better here.

app/models/theme.rb
```diff
@@ -151,6 +154,23 @@ def default?
...
   end
 
+  def enabled?
+    enabled = true
+
+    minimum_version = remote_theme&.minimum_discourse_version
+    maximum_version = remote_theme&.maximum_discourse_version
+
+    if minimum_version
+      enabled = false unless Discourse.has_needed_version?(Discourse::VERSION::STRING, minimum_version)
```
eviltrout: If you're not depending on `enabled` later it would make more sense to `return false` and avoid any further checks. You could probably avoid the `enabled` variable altogether by returning `true` if you make it until the end of the method.

app/models/theme.rb
```diff
@@ -435,25 +435,30 @@ def translation_override_hash
...
-      name: name,
-      about_url: remote_theme&.about_url,
-      license_url: remote_theme&.license_url,
-      component: component,
-      assets: {}.tap do |hash|
+    {}.tap do |meta|
+      meta[:name] = name
+      meta[:component] = component
+
+      RemoteTheme::METADATA_PROPERTIES.each do |property|
```
eviltrout: Could you use `slice` here?  `meta = remote_theme.slice(RemoteTheme::METADATA_PROPERTIES)`

davidtaylorhq: We want the output to contain all the keys, even if the values are `nil`, AND even if there is no `remote_theme`. 

This works for missing values:
```
meta.merge! remote_theme.slice(RemoteTheme::METADATA_PROPERTIES)
```

But to also account for a missing remote_theme I'd need:

```
meta.merge! (remote_theme || {})&.slice(RemoteTheme::METADATA_PROPERTIES) ||
            Hash[RemoteTheme::METADATA_PROPERTIES.map { |x| [x, nil] }]
```

I think the original looping method is probably easier to read and understand

davidtaylorhq:
Thanks for the review @eviltrout, I've made those changes.

> few new tests considering how much code changed

I think the diff is slightly deceptive here - one big 'change' is re-arranging the hbs template, which won't actually impact any functionality/tests.

The `transform_ids` function is not new, even though the diff makes it look like that. I only added 1 line, and surrounded it in `get_set_cache`. So all the old tests still apply there, plus I added some for my new line.

>  are there any tests for making sure remote themes correctly update the DB fields and provide the metadata

There were no tests specifically for the new fields - all the metadata follows the same code path (loop over `METADATA_FIELDS`). I've now added a couple of the new ones to some existing tests, to be double sure.

PR #6946, opened 2019-01-24
FIX: ensures ordering set in tags-show queryString works at first load

jjaffeux:
This commit also refactors a significant part of tags-show


app/assets/javascripts/discourse/controllers/tags-show.js.es6
```diff
@@ -52,7 +55,7 @@ export default Ember.Controller.extend(BulkTopicSelection, {
   additionalTags: null,
   list: null,
   canAdminTag: Ember.computed.alias("currentUser.staff"),
-  filterMode: null,
+  filterMode: "",
```
eviltrout: Why an empty string instead of `null`?

PR #6947, opened 2019-01-25
FEATURE: Allow setting font size per-device using a cookie

davidtaylorhq:
This allows changing the font size per-device, just like for themes. cc. @coding-horror 

screenshot 2019-01-25 at 00 23 54


coding-horror:
Can you review this with high priority @eviltrout

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
```diff
@@ -58,6 +59,16 @@ export default Ember.Controller.extend(PreferencesTabController, {
     return currentThemeId();
   },
 
+  @computed()
+  textSize() {
+    let cookie = $.cookie("text_size");
```
jjaffeux:
```suggestion
    const cookie = $.cookie("text_size");
```

app/assets/javascripts/discourse/controllers/preferences/interface.js.es6
```diff
@@ -109,6 +120,11 @@ export default Ember.Controller.extend(PreferencesTabController, {
         this.set("model.user_option.theme_ids", [this.get("themeId")]);
       }
 
+      const makeTextSizeDefault = this.get("makeTextSizeDefault");
+      if (makeTextSizeDefault) {
+        this.set("model.user_option.text_size", this.get("textSize"));
```
jjaffeux: I don’t know how @eviltrout feels about this but I try more and more to avoid using "set" inside a computed property. It makes code harder to read and might lead to weird bugs; eg: you have to make sure you are not using  `model.user_option.text_size` for something which is used to define the property you are in, which in some cases might be in other functions.

In this case can’t we use some init hook?

davidtaylorhq: Agreed, I blindly followed the existing convention for the theme setting here. I will see about moving both to an init hook.

app/assets/javascripts/discourse/routes/preferences-interface.js.es6
```diff
@@ -4,8 +4,10 @@ export default RestrictedUserRoute.extend({
   showFooter: true,
 
   setupController(controller, user) {
+    let textSize = $.cookie("text_size") || user.get("user_option.text_size");
```
jjaffeux:
```suggestion
    const textSize = $.cookie("text_size") || user.get("user_option.text_size");
```

PR #6955, opened 2019-01-26
FEATURE: Allow overriding text size from a different device

davidtaylorhq:
This brings the feature in line with the theme selection system

### DIFF
--- a/app/assets/javascripts/discourse/models/user.js.es6
+++ b/app/assets/javascripts/discourse/models/user.js.es6
@@ -710,9 +710,7 @@ const User = RestModel.extend({
   @computed("user_option.text_size_seq", "user_option.text_size")
   currentTextSize(serverSeq, serverSize) {
     if ($.cookie("text_size")) {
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -705,6 +705,24 @@ const User = RestModel.extend({
     });
 
     return _.uniq(titles).sort();
+  },
+
+  @computed("user_option.text_size_seq", "user_option.text_size")
+  currentTextSize(serverSeq, serverSize) {
+    if ($.cookie("text_size")) {
+      const parts = $.cookie("text_size").split("|");
```
tgxworld: Perhaps something like `const [cookieSize, cookieSeq] = $.cookie("text_size").split("|");`?

tgxworld:
LGTM :+1:

davidtaylorhq:
Thanks for the comments @tgxworld, I've changed both those things. Merging now

PR #6957, opened 2019-01-28
PERF: Cache build_not_found_page.

udan11:


SamSaffron:
Let's rename this `PERF: `

I would like to see a test here, I know it is somewhat annoying but you can add a helper method on app controller to flush the cache so you can properly test caching works as expected. 

Really important to cover this codepath cause we don't want any regressions where 404 pages only intermittently work.

davidtaylorhq:
It's worth double checking that this doesn't regress https://github.com/discourse/discourse/commit/37300d67777da838c9c5be658d92a0ac552bc239 (showing topic titles to anon users on login_required sites)

SamSaffron:
one more subtle issue here, what if user A is using a light theme and user B is using a dark theme?

app/controllers/application_controller.rb
```diff
@@ -717,15 +717,25 @@ def build_not_found_page(status = 404, layout = false)
...
 
-    category_topic_ids = Category.pluck(:topic_id).compact
-    @container_class = "wrap not-found-container"
-    @top_viewed = TopicQuery.new(nil, except_topic_ids: category_topic_ids).list_top_for("monthly").topics.first(10)
-    @recent = Topic.includes(:category).where.not(id: category_topic_ids).recent(10)
-    @slug =  params[:slug].class == String ? params[:slug] : ''
-    @slug =  (params[:id].class == String ? params[:id] : '') if @slug.blank?
+    key = "not_found_page_#{status}_#{layout}_#{!!current_user}"
+
+    if !(page = $redis.get(key))
```
ZogStriP:
```suggestion
    unless page = $redis.get(key)
```

app/controllers/application_controller.rb
```diff
@@ -717,15 +717,25 @@ def build_not_found_page(status = 404, layout = false)
...
+      @container_class = "wrap not-found-container"
+      @top_viewed = TopicQuery.new(nil, except_topic_ids: category_topic_ids).list_top_for("monthly").topics.first(10)
+      @recent = Topic.includes(:category).where.not(id: category_topic_ids).recent(10)
+      @slug = '$_SLUG_PLACEHOLDER_$'
+      @hide_search = true if SiteSetting.login_required
+      page = render_to_string status: status, layout: layout, formats: [:html], template: '/exceptions/not_found'
+      $redis.setex(key, 10.minutes, page)
+    end
+
+    @slug = params[:slug].class == String ? params[:slug] : ''
```
ZogStriP:
```suggestion
    @slug = params[:slug].presence || ""
```

app/controllers/application_controller.rb
udan11: I usually avoid `unless`, but in this case it looks better indeed.

SamSaffron: For the record, don't get me started on `unless` I always unconditionally prefer not to use it. I find `unless` a keyword I wish I could cancel from Ruby.

ZogStriP:
This is one of the rare case where I think it makes sense. But I'm fine banning `unless`.

app/controllers/application_controller.rb
```diff
@@ -717,12 +717,19 @@ def build_not_found_page(status = 404, layout = false)
       layout = 'application' if layout == 'no_ember'
     end
 
-    category_topic_ids = Category.pluck(:topic_id).compact
+    key = "not_found_page_topics"
```
tgxworld: Probably the following so that we are consistent with the naming.

```suggestion
    key = "page_not_found_topics"
```

Also, I think the key should be moved into a constant or a public method so that we're not duplicating this key in multiple places of the code base.

spec/requests/application_controller_spec.rb
```diff
@@ -143,6 +143,21 @@
...
+      it 'should cache results' do
+        $redis.set('not_found_page_topics', 'first response')
+        get '/t/nope-nope/99999999'
+        expect(response.status).to eq(404)
+        expect(response.body).to include('first response')
+
+        SiteSetting.login_required = true
+        sign_in(Fabricate(:user))
+
+        $redis.set('not_found_page_topics', 'second response')
```
tgxworld: We need to ensure that the key in Redis is cleaned up at the end of the test to ensure that it doesn't leak the state to other tests that follow.

spec/requests/application_controller_spec.rb
```diff
@@ -143,6 +143,21 @@
         expect(response.status).to eq(404)
         expect(response.body).to_not include('google.com/search')
       end
+
+      it 'should cache results' do
+        $redis.set('not_found_page_topics', 'first response')
```
tgxworld: I personally think that we're not sufficiently testing the changes here. How I will go about testing this change will be as follows:

1. Create a topic and test that a not found will show that topic. 
1. Assert that the right key has been set in Redis
1. Create a new topic and test the same route.
1. Assert that the response does not include the new topic because the previous response should have been cached.

app/controllers/application_controller.rb
udan11: I think the name **`not_found_page`**`_topics` was consistent with `build_`**`not_found_page`**, but I changed the name.

The key is used in one place right now so I do not believe it makes sense to move it.

spec/requests/application_controller_spec.rb
```diff
@@ -143,6 +143,22 @@
         expect(response.status).to eq(404)
         expect(response.body).to_not include('google.com/search')
       end
+
+      it 'should cache results' do
+        $redis.flushall
```
tgxworld: Small one but we usually assume that each test cleans up after itself. Therefore, there shouldn't be a need to flush redis here.

spec/requests/application_controller_spec.rb
```diff
@@ -143,6 +143,22 @@
...
+
+        topic1 = Fabricate(:topic)
+        get '/t/nope-nope/99999999'
+        expect(response.status).to eq(404)
+        expect(response.body).to include(topic1.title)
+
+        topic2 = Fabricate(:topic)
+        get '/t/nope-nope/99999999'
+        expect(response.status).to eq(404)
+        expect(response.body).to_not include(topic2.title)
```
tgxworld: oops sorry I think we need to assert that the body contains `topic1` but not `topic2`.

spec/requests/application_controller_spec.rb
tgxworld: > A quick look for expect(response.status).to eq(404) returns 122 matches.

Hmm that does mean that all those tests are already leaking a state in Redis. I can't think of a clean way to handle this right now without using some sort of mock Redis like you mentioned xD. Perhaps we can set the `TTL` for the key to be really low in the test env? We leak state but we leak it for a much shorter period of time.

SamSaffron: Yeah lets not flushall in the test, instead expose a helper method from the controller that the test can consume to clean up the cache first. It is a good API to have anyway.

udan11: `$redis.del("page_not_found_topics")` seems to be better. I do not see any point in creating a different method that would only be used by the test.

app/controllers/application_controller.rb
```diff
@@ -717,12 +717,21 @@ def build_not_found_page(status = 404, layout = false)
...
-    category_topic_ids = Category.pluck(:topic_id).compact
+    if !SiteSetting.login_required? || current_user
+      key = "page_not_found_topics"
+      if @topics_partial = $redis.get(key)
+        @topics_partial = @topics_partial.html_safe
+      else
+        category_topic_ids = Category.pluck(:topic_id).compact
+        @top_viewed = TopicQuery.new(nil, except_topic_ids: category_topic_ids).list_top_for("monthly").topics.first(10)
+        @recent = Topic.includes(:category).where.not(id: category_topic_ids).recent(10)
+        @topics_partial = render_to_string partial: '/exceptions/not_found_topics'
```
ZogStriP: Does `render_to_string` call `html_safe`?

udan11: `render_to_string` builds an `ActionView::OutputBuffer` which inherits `ActiveSupport::SafeBuffer` and `html_safe` constructs a new `ActiveSupport::SafeBuffer`.

In other words, `render_to_string` does not call `html_safe`, but they have a common ancestor.

spec/requests/application_controller_spec.rb
tgxworld: Was there a change made for this? It seems like the we’re not cleaning up the cache properly for the other tests that results in a 404.

PR #6959, opened 2019-01-28
FIX: tags added while editing a pending approval topic do not show up

techAPJ:
https://meta.discourse.org/t/tags-added-while-editing-a-pending-approval-topic-do-not-show-up/106349

### DIFF
--- a/app/models/queued_post.rb
+++ b/app/models/queued_post.rb
@@ -91,6 +91,12 @@ class QueuedPost < ActiveRecord::Base
       end
     end
 
+    if create_options[:tags].present? &&
+      created_post.post_number == 1 &&
+      created_post.topic.tags.blank?
+      DiscourseTagging.tag_topic_by_names(created_post.topic, Guardian.new(approved_by), create_options[:tags])
+    end
+
     # Do sidekiq work outside of the transaction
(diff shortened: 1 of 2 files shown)


SamSaffron:
I would like @eviltrout to think this one through and approve given it overlaps with his work. 

The tricky thing here is if admins are not careful when approving they can inject new tags without knowing they are. 

For example.... 

Say only TL3 are allowed to create new tags. 
Then a TL0 gets a new post into approval with random brand new tags
The security check is performed as the "admin" and new tag sneaks in 

Maybe we want to give some extra safety here to ensure the user technically can add these tags... not sure... but there is a bit of a trap here. 

Maybe new review queue can simply warn about new tags that will be created on approval. Not sure.

eviltrout:
The philosophy for approval has been "if the person approving it says it's ok, it's OK." 

We skip validations. 

Why? Because otherwise the approver might click "Approve" and get an error because the person posting it cannot for some trivial reason and the approver would have to figure it out. 

In this case, it seems like the approver is letting something through that does not work. It does seem like a bug. I am not sure why this is needed but it will be replaced by the new review queue.

SamSaffron:
OK we can merge this in for the time being

PR #6960, opened 2019-01-29
FIX: Add multisite support to Sidekiq::Pausable.

tgxworld:
Having a global Sidekiq pause switch is problematic because a site in
the cluster can pause Sidekiq for the entire cluster.


spec/multisite/pausable_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        Sidekiq.pause!
+        expect(Sidekiq.paused?).to eq(true)
+      end
+    end
+  end
+end
+
+RSpec.describe Sidekiq::Pausable do
+  after do
+    $redis.flushall
```
SamSaffron: I don't like this flushall ... we should simply ensure we don't leak state.

tgxworld: Is there a reason for not liking flushall? The operation is cheap in the test env.

ZogStriP: I don't think the issue is about how fast/slow this operation is, but rather, this might prevent us from detecting tests that leak state. Each tests should take care of cleaning up their own mess instead of blowing up the whole world.

tgxworld: I’m not sure if I follow given the number of flushall that is present in our test suite. I personally think that the test shouldn’t have to be exposed to the underlying Redis key that is being used as I consider that an internal implementation detail that is not critical to what we’re testing for.

SamSaffron: yeah we should work to remove all flushalls, they mean we are not in a good habit of cleaning up and our apis are lacking. I also don't think you need to expose any keys here, just `unpause!` the full multisite after each test if you wish.  also we got to make sure extend lease thread in no longer running after every test here.

SamSaffron:
we really need this change now.... merging it in will clean up test afterwards

PR #6962, opened 2019-01-29
FIX: do not treat TIFF, BMP, WEBP as images

pmusaraj:
TIFF support was partially removed in https://github.com/discourse/discourse/commit/71caf7521d217babdcd2b87c2d9275d07de2d17c, this PR removes a few more TIFF occurrences as well as BMP and WEBP support. 

TIFF, BMP, WEBP uploads in the composer will now be treated like regular attachments and an `<a>` tag will be inserted instead of an `` tag.

### DIFF
--- a/app/assets/javascripts/discourse/lib/utilities.js.es6
+++ b/app/assets/javascripts/discourse/lib/utilities.js.es6
@@ -282,7 +282,7 @@ export function validateUploadedFile(file, opts) {
   return true;
 }
 
-const IMAGES_EXTENSIONS_REGEX = /(png|jpe?g|gif|bmp|svg|webp|ico)/i;
+const IMAGES_EXTENSIONS_REGEX = /(png|jpe?g|gif|svg|ico)/i;
 
 function extensionsToArray(exts) {
   return exts
@@ -376,7 +376,7 @@ export function authorizesOneOrMoreImageExtensions() {
 }
 
 export function isAnImage(path) {
-  return /\.(png|jpe?g|gif|bmp|svg|webp|ico)$/i.test(path);
+  return /\.(png|jpe?g|gif|svg|ico)$/i.test(path);
 }
 
 function uploadTypeFromFileName(fileName) {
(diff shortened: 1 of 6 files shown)


tgxworld:
Do we need to remove `webp` and `bmp` too?

pmusaraj:
`bmp` files work on Safari, Chrome, Firefox, Edge. `webp` only works with Chrome and Edge, so I would remove `webp` as well.

coding-horror:
Sam removed bmp earlier, so we need to follow through on that.

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -348,7 +348,7 @@ export function authorizedExtensions() {
 
 export function authorizedImagesExtensions() {
   return authorizesAllExtensions()
-    ? "png, jpg, jpeg, gif, bmp, tiff, svg, webp, ico"
+    ? "png, jpg, jpeg, gif, bmp, svg, webp, ico"
```
ZogStriP: Forgot to remove some in here as well?

```suggestion
    ? "png, jpg, jpeg, gif, svg, ico"
```

lib/upload_creator.rb
```diff
@@ -3,7 +3,7 @@
 
 class UploadCreator
 
-  TYPES_CONVERTED_TO_JPEG ||= %i{bmp png}
+  TYPES_CONVERTED_TO_JPEG ||= %i{png}
```
ZogStriP: There's no point in having an array with only 1 item...

spec/models/optimized_image_spec.rb
```diff
@@ -155,8 +155,8 @@
   describe ".safe_path?" do
 
     it "correctly detects unsafe paths" do
-      expect(OptimizedImage.safe_path?("/path/A-AA/22_00.TIFF")).to eq(true)
-      expect(OptimizedImage.safe_path?("/path/AAA/2200.TIFF")).to eq(true)
+      expect(OptimizedImage.safe_path?("/path/A-AA/22_00.jpg")).to eq(true)
```
ZogStriP: Might want to keep the uppercase

SamSaffron:
Looks good to me!

PR #6963, opened 2019-01-30
DEV: Allow theme CLI to specify which theme to synchronize

davidtaylorhq:
Currently the theme is matched by name, which can be fragile when there are many themes with the same name. The new functionality in this commit will be used by the next version of the theme CLI. The change is backwards compatible, so existing theme CLI versions continue to work.

This is a pretty low risk change, and it's well tested. I'd like to get it in before the release, so that the upcoming theme cli changes can be used with the stable branch. What do you think @SamSaffron?

Sneak peek of how this works with the new theme CLI:

screenshot 2019-01-30 at 11 38 03


coding-horror:
I think we should try to get it in.

PR #6964, opened 2019-01-30
DEV: Remove ignored columns

gschlager:
A little bit of cleanup. Those columns have been ignored for quite some time.

I didn't touch the following, because it's quite new. 
https://github.com/discourse/discourse/blob/d8bd3c32ca43407499dea414c4f17e5c37d6d71d/app/models/category.rb#L6-L8


SamSaffron:
looks good to me, merge in by the end of next week

PR #6965, opened 2019-01-30
FIX: Fix delete button for Tag Groups.

nbianca:


app/assets/javascripts/discourse/models/tag-group.js.es6
```diff
@@ -9,6 +9,11 @@ export default RestModel.extend({
     return saving || Ember.isEmpty(name) || Ember.isEmpty(tagNames);
   },
 
+  @computed("id")
+  disableDelete(id) {
+    return !Number.isInteger(id);
```
eviltrout: I don't believe this works in IE, unfortunately :(

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Browser_compatibility

PR #6966, opened 2019-02-01
Add a permalink normalization prefix setting for phpbb3

curiousdannii:
[As discussed in the forum](https://meta.discourse.org/t/importing-from-phpbb3/30810/421), I used this in a test migration to enable you to set a prefix for the permalink normalisations.

It's not very rigorous though. Is there a good way to strip slashes from the beginning and end, and to escape any slashes found in the middle?

Edit: or considering all this does is add some text to the regexs, maybe it would be better to just add some docs, rather than trying to make the code foolproof.

### DIFF
--- a/script/import_scripts/phpbb3/importers/permalink_importer.rb
+++ b/script/import_scripts/phpbb3/importers/permalink_importer.rb
@@ -47,6 +47,9 @@ module ImportScripts::PhpBB3
     protected
 
     def add_normalization(normalizations, normalization)
+      if @settings.normalization_prefix
+        normalization = "/" + @settings.normalization_prefix + "\\" + normalization
+      end
       normalizations << normalization unless normalizations.include?(normalization)
     end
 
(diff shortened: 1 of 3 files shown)


gschlager:
I think the following code snippet should do the trick. It removes the leading/trailing slashes and escapes the remaining slashes in the prefix. There's no need for the warning about slashes in the settings file anymore.

```ruby
      if @settings.normalization_prefix.present?
        prefix = @settings.normalization_prefix[%r|^/?(.*?)/?$|, 1]
        normalization = "/#{prefix.gsub('/', '\/')}\\#{normalization}"
      end
```

SamSaffron:
@gschlager can you merge this in with that amendment?

curiousdannii:
I am intending to add the suggested patch, but if someone else gets there first I won't mind!

gschlager:
@curiousdannii Thanks for the initial implementation. I added my proposed solution in https://github.com/discourse/discourse/commit/24369a81663b25c16b8fccd7e6e33fd4a44284f8

PR #6967, opened 2019-02-03
Touch only pan events

featheredtoast:
- Pan listens for touch events ONLY. This should fix edge cases where browsers with a pointer device is registering a pan for just hovering.
- Cleanup duplicate code - add touch register helpers
- Add request animation frame after initial render to help "stutter" animations where we're occasionally waiting for afterRender events to fire on swipe open events.


jjaffeux:
Seem way better to me

PR #6971, opened 2019-02-05
FEATURE: adds an API to register topic footer buttons

jjaffeux:
See plugin-api registerTopicFooterButton function for details about usage.

### DIFF
--- a/app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
+++ b/app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
@@ -1,0 +1,143 @@
+import {
+  registerTopicFooterButton,
+  clearTopicFooterButtons
+} from "discourse/lib/register-topic-footer-button";
+
+export default {
+  name: "topic-footer-buttons",
+
+  initialize() {
+    clearTopicFooterButtons();
+
+    registerTopicFooterButton({
+      id: "share",
+      icon: "link",
+      priority: 999,
+      label: "topic.share.title",
+      title: "topic.share.help",
+      action() {
+        this.appEvents.trigger(
+          "share:url",
+          this.get("topic.shareUrl"),
+          $("#topic-footer-buttons")
+        );
+      },
+      dropdown() {
+        return this.site.mobileView;
+      },
+      classNames: ["share"],
+      dependentKeys: ["topic.shareUrl", "topic.isPrivateMessage"],
+      displayed() {
+        return !this.get("topic.isPrivateMessage");
+      }
+    });
+
+    registerTopicFooterButton({
+      id: "flag",
+      icon: "flag",
+      priority: 998,
+      label: "topic.flag_topic.title",
+      title: "topic.flag_topic.help",
+      action: "showFlagTopic",
+      dropdown() {
+        return this.site.mobileView;
+      },
+      classNames: ["flag-topic"],
+      dependentKeys: ["topic.details.can_flag_topic", "topic.isPrivateMessage"],
+      displayed() {
+        return (
+          this.get("topic.details.can_flag_topic") &&
+          !this.get("topic.isPrivateMessage")
+        );
+      }
+    });
+
+    registerTopicFooterButton({
+      id: "invite",
+      icon: "users",
+      priority: 997,
+      label: "topic.invite_reply.title",
+      title: "topic.invite_reply.help",
+      action: "showInvite",
+      dropdown() {
+        return this.site.mobileView;
+      },
+      classNames: ["invite-topic"],
+      dependentKeys: ["canInviteTo", "inviteDisabled"],
+      displayed() {
+        return this.get("canInviteTo");
+      },
+      disabled() {
+        return this.get("inviteDisabled");
+      }
+    });
+
+    registerTopicFooterButton({
+      dependentKeys: ["topic.bookmarked", "topic.isPrivateMessage"],
+      id: "bookmark",
+      icon: "bookmark",
+      priority: 1000,
+      classNames() {
+        const bookmarked = this.get("topic.bookmarked");
+        return bookmarked ? ["bookmark", "bookmarked"] : ["bookmark"];
+      },
+      label() {
+        const bookmarked = this.get("topic.bookmarked");
+        return bookmarked ? "bookmarked.clear_bookmarks" : "bookmarked.title";
+      },
+      title() {
+        const bookmarked = this.get("topic.bookmarked");
+        return bookmarked
+          ? "bookmarked.help.unbookmark"
+          : "bookmarked.help.bookmark";
+      },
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,116 @@
...
+
+  initialize() {
+    withPluginApi("0.8", api => {
+      api.registerTopicFooterButton({
+        id: "share",
+        icon: "link",
+        priority: 999,
+        label: "topic.share.title",
+        title: "topic.share.help",
+        action: context => {
```
eviltrout: I think this is one of those cases where applying `this` makes more sense. Something like:

```
action() {
  this.appEvents(...);
}
```
I think it's nicer and clearer than `context`

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,116 @@
...
+      });
+
+      api.registerTopicFooterButton({
+        id: "edit-message",
+        priority: 750,
+        icon: "pencil-alt",
+        label: "topic.edit_message.title",
+        title: "topic.edit_message.help",
+        action: context => context.get("editFirstPost"),
+        dropdown: false,
```
eviltrout: I would assume the default is no dropdown and that this would be unnecessary.

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,116 @@
...
+        title: "topic.share.help",
+        action: context => {
+          context.appEvents.trigger(
+            "share:url",
+            context.get("topic.shareUrl"),
+            $("#topic-footer-buttons")
+          );
+        },
+        dropdown: context => context.site.mobileView,
+        class: "share",
```
eviltrout: I thought `class` was a reserved word? It might be safer to use classNames or similar.

app/assets/javascripts/discourse/lib/register-topic-footer-button.js.es6
```diff
@@ -0,0 +1,124 @@
...
+    );
+    return;
+  }
+
+  _topicFooterButtons.push(action);
+}
+
+export function getTopicFooterButtons() {
+  const dependentKeys = [].concat(
+    ..._topicFooterButtons.map(tfb => tfb.dependentKeys).filter(x => x)
```
eviltrout: What is the `filter` filtering here? That's a bit confusing.

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,116 @@
...
+        dependentKeys: ["topic.shareUrl", "topic.isPrivateMessage"],
```
eviltrout: I am not crazy about the `dependentKeys` thing. Part of me thinks it should use the `@computed` helper on the methods that need it, but I realize that is super complicated.

A compromise might be to call it something closer to what it does. `refreshIfChanged` perhaps? `triggerRefresh`?

app/assets/javascripts/discourse/lib/register-topic-footer-button.js.es6
jjaffeux: That's some pattern to ensure we dont have any empty element

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
jjaffeux: Yes I think that would be better, class is reserved in some cases only, but probably better to go with `classNames` I like it :+1:

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
jjaffeux: Yep was thinking the same :+1:

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
jjaffeux: Not sure, I like the fact that it does express the fact that an array is kind of expected. Also it's not exactly related to refresh, it's keys needed to build this button.

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
jjaffeux: indeed, I thought being explicit would be nice, but maybe I could just rely on this being what one would expect.

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
eviltrout: Okay cool, keep it.

app/assets/javascripts/discourse/lib/register-topic-footer-button.js.es6
eviltrout: I see, it's like `compact` in ruby. Makes sense.

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,148 @@
...
+
+  initialize() {
+    withPluginApi("0.8.28", api => {
+      api.registerTopicFooterButton({
+        id: "share",
+        icon: "link",
+        priority: 999,
+        label: "topic.share.title",
+        title: "topic.share.help",
+        action: function() {
```
eviltrout:
```suggestion
        action() {
```

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,148 @@
...
+          this.appEvents.trigger(
+            "share:url",
+            this.get("topic.shareUrl"),
+            $("#topic-footer-buttons")
+          );
+        },
+        dropdown: function() {
```
eviltrout:
```suggestion
        dropdown() {
```

app/assets/javascripts/discourse/initializers/topic-footer-buttons.js.es6
```diff
@@ -0,0 +1,148 @@
...
+          return this.site.mobileView;
+        },
+        classNames: ["share"],
+        dependentKeys: ["topic.shareUrl", "topic.isPrivateMessage"],
+        displayed: function() {
```
eviltrout:
```suggestion
        displayed() {
```

PR #6973, opened 2019-02-05
FEATURE: Calculate sprite-sheet based on currently active themes

davidtaylorhq:
Previously there was only one sprite sheet, which always included icons from all themes even if they were disabled.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -382,7 +382,7 @@ module ApplicationHelper
 
   def theme_ids
     if customization_disabled?
-      nil
+      [nil]
     else
       request.env[:resolved_theme_ids]
     end
@@ -454,11 +454,11 @@ module ApplicationHelper
       asset_version: Discourse.assets_digest,
       disable_custom_css: loading_admin?,
       highlight_js_path: HighlightJs.path,
-      svg_sprite_path: SvgSprite.path,
+      svg_sprite_path: SvgSprite.path(theme_ids),
     }
 
     if Rails.env.development?
-      setup_data[:svg_icon_list] = SvgSprite.all_icons
+      setup_data[:svg_icon_list] = SvgSprite.all_icons(theme_ids)
     end
 
     if guardian.can_enable_safe_mode? && params["safe_mode"]
--- a/app/views/wizard/index.html.erb
+++ b/app/views/wizard/index.html.erb
@@ -18,7 +18,7 @@
 
     <div id="svg-sprites" style="display:none;">
       <div class="fontawesome">
-        <%= raw SvgSprite.bundle([]) %>
+        <%= raw SvgSprite.bundle %>
       </div>
     </div>
 
--- a/app/views/wizard/index.html.erb
+++ b/app/views/wizard/index.html.erb
@@ -18,7 +18,7 @@
 
     <div id="svg-sprites" style="display:none;">
       <div class="fontawesome">
-        <%= raw SvgSprite.bundle %>
+        <%= raw SvgSprite.bundle([]) %>
       </div>
     </div>
 
--- a/lib/svg_sprite/svg_sprite.rb
+++ b/lib/svg_sprite/svg_sprite.rb
@@ -342,10 +342,7 @@ Discourse SVG subset of #{fa_license}
   end
 
   def self.get_set_cache(key, &blk)
-    if val = cache[key]
-      return val
-    end
-    cache[key] = blk.call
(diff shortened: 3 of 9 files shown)


pmusaraj:
This looks very nice! 

My only comment is stylistic: can we avoid having to pass an empty `[]` for the default values, e.g. `SvgSprite.bundle([])`?

davidtaylorhq:
I didn't want to set a default value, because the functions should almost always be called with a list of theme ids. Providing a default would make it very easy to forget - I'd prefer that we have to specifically ask for "sprites without any themes". 

That said, I agree it looks pretty messy in the tests. I'll add a helper function there to tidy it up :+1:.

lib/svg_sprite/svg_sprite.rb
```diff
@@ -347,4 +340,15 @@ def self.process(icon_name)
     FA_ICON_MAP.each { |k, v| icon_name.sub!(k, v) }
     fa4_to_fa5_names[icon_name] || icon_name
   end
+
+  def self.get_set_cache(key, &blk)
+    if val = cache[key]
```
ZogStriP: What about?

```suggestion
    cache[key] ||= blk.call
```

lib/svg_sprite/svg_sprite.rb
```diff
@@ -188,39 +188,37 @@ module SvgSprite
...
-      .merge(plugin_icons)
-      .merge(badge_icons)
-      .merge(group_icons)
-      .merge(theme_icons)
-      .delete_if { |i| i.blank? || i.include?("/") }
-      .map! { |i| process(i.dup) }
-      .merge(SVG_ICONS)
-      .sort
+  def self.version(theme_ids)
+    get_set_cache "version_#{Theme.transform_ids(theme_ids).join(',')}" do
```
ZogStriP: We should be consistent with using `(` and `)`

```suggestion
    get_set_cache("version_#{Theme.transform_ids(theme_ids).join(',')}") do
```

app/views/wizard/index.html.erb
```diff
@@ -18,7 +18,7 @@
 
     <div id="svg-sprites" style="display:none;">
       <div class="fontawesome">
-        <%= raw SvgSprite.bundle %>
+        <%= raw SvgSprite.bundle([]) %>
```
eviltrout: I agree with @pmusaraj that this seems odd to me. I understand the rationale that you almost always want to call it with a value, but we have at least this one example of not doing that.

If you want to keep the required parameter, then I would suggest adding a new method `empty_bundle` or something for this particular case (and maybe in your tests too.)

app/helpers/application_helper.rb
```diff
@@ -382,7 +382,7 @@ def topic_featured_link_domain(link)
 
   def theme_ids
     if customization_disabled?
-      nil
+      [nil]
```
eviltrout: I believe we've discussed this before but an empty array with a single `nil` value is really weird to me. (I believe last time you said it is required by some existing code.)

You should make a note that this is in need of a refactor to a more sensible structure. If there's a reason why an empty array or nil wouldn't work, maybe the answer is creating a helper class that can establish the difference with some other properties.

lib/svg_sprite/svg_sprite.rb
```diff
@@ -347,4 +340,12 @@ def self.process(icon_name)
     FA_ICON_MAP.each { |k, v| icon_name.sub!(k, v) }
     fa4_to_fa5_names[icon_name] || icon_name
   end
+
+  def self.get_set_cache(key, &blk)
+    cache[key] ||= blk.call
```
eviltrout: If you're not doing anything funky with the block, I'd suggest to use `yield` instead and omit the `&blk` param.

app/views/wizard/index.html.erb
davidtaylorhq: :+1: ok, I'm persuaded - I've added a default value, and removed all the helper stuff I added to the tests

app/helpers/application_helper.rb
davidtaylorhq: Yes, this is because of safe mode - the array needs to have at least one element, since it gets looped over to generate the CSS. No elements = absolutely no CSS 😢 

I'll look into refactoring this

lib/svg_sprite/svg_sprite.rb
davidtaylorhq: Done, thanks for the tip

PR #6974, opened 2019-02-05
DEV: Plugin scss errors should break precompile

pmusaraj:
Previously, if a plugin had a SCSS syntax error, the app would rebuild and the error would be displayed in the browser when loading the app post-rebuild. 

This PR changes the behaviour by raising an error immediately on SCSS syntax errors, thus triggering a rebuild failure. 

Note that while testing these changes in several environments, the `rescue => PG::UndefinedColumn` line would trigger skipping CSS precompilation due to a `already initialized constant PG::UndefinedColumn` warning (the updated `raise` syntax no longer triggers the warning).

### DIFF
--- a/lib/discourse.rb
+++ b/lib/discourse.rb
@@ -105,6 +105,8 @@ module Discourse
 
   class Deprecation < StandardError; end
 
+  class ScssError < StandardError; end
+
   def self.filters
     @filters ||= [:latest, :unread, :new, :read, :posted, :bookmarks]
   end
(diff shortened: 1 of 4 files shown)


davidtaylorhq:
How does this work in development? (i.e. when assets are not precompiled) Do we still get the SCSS compilation errors appearing in the browser?

screenshot 2019-02-05 at 19 37 39

pmusaraj:
In development, you will now get a regular Rails error in the browser: 

image

and an error in the CSS watcher on your server console: 

image

PR #6977, opened 2019-02-06
Upgrade to Ember 3.7.0

majakomel:


app/assets/javascripts/admin/components/ace-editor.js.es6
```diff
@@ -100,7 +100,7 @@ export default Ember.Component.extend({
 
         if (this.appEvents) {
           // xxx: don't run during qunit tests
-          this.appEvents.on("ace:resize", () => this.resize());
+          this.appEvents.on("ace:resize", this, this.resize);
```
eviltrout: Is this an internal Ember API change? Because I believe a lot of plugins depend on this working. We might need to do something about that.

majakomel: This is underlying deprecation that was added in 3.6 https://www.emberjs.com/deprecations/v3.x/#toc_events-remove-all-listeners

eviltrout: Unfortunately then we have a fair bit more work to do here, because we have exposed `appEvents` as a common interface for plugins to use.  For example if you search through our [all-the-plugins](https://github.com/discourse/all-the-plugins) repo, you'll find many uses of this API.

It looks like you did an excellent job refactoring the internal use of it the API, but we need to take it a little farther:

AppEvents needs to respond and work with the old API. We can do this by overwriting the `on` and `off` methods and checking the number of arguments. If the argument count is 2, we need to log a deprecation warning and make the old code compatible.
 
  * Making the old code compatible is the tricky part here. When called, I would suggest putting a new reference object in an array for that given event name. When `off` is called, it could loop through all those objects and `off` them all.

For example:
```js

let _events = {};

on() {
  if (arguments.length == 2) { 
    let [name, fn] = arguments;
     console.log( deprecation message );
     let reference = {};
     _events[name] = _events[name] || [];
     _events[name].push(reference);
     this._super(name, reference, fn);
];
  } else if (arguments.length == 3) {
   this._super(...arguments);  
  }
}

off() {
  if (arguments.length == 2) { 
     let [name, fn] = arguments;
     _events[name].forEach(ref => this._super(name, ref, fn);
   } else if (arguments.length == 3) {
     this._super(...arguments);  
   }
}
```

I understand this might be a lot to swallow in a git commit message. If you'd prefer to move it to to our dev/meta forums for longer form discussion that's cool too!

SamSaffron:
Super happy to see this change, feel free to deploy meta on a branch with this once you feel it is ready, then we can merge into master 1-2 days later.

ZogStriP:
@majakomel can you merge this tomorrow when you start your day?

PR #6980, opened 2019-02-06
REFACTOR: Do not use `[nil]` as a list of theme_ids when in safe mode

davidtaylorhq:
I've removed this unusual behaviour, so now `[nil]` is only used inside a specific function of the stylesheet compiler.

cc @eviltrout

### DIFF
--- a/lib/stylesheet/manager.rb
+++ b/lib/stylesheet/manager.rb
@@ -44,6 +44,7 @@ class Stylesheet::Manager
     theme_ids = [theme_ids] unless Array === theme_ids
     theme_ids = [theme_ids.first] unless target =~ THEME_REGEX
     theme_ids = Theme.transform_ids(theme_ids, extend: false)
+    theme_ids = [nil] if theme_ids.empty? # Build at least one stylesheet without any theme
 
     current_hostname = Discourse.current_hostname
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -380,7 +380,10 @@ class ApplicationController < ActionController::Base
     return if request.method != "GET"
 
     resolve_safe_mode
-    return if request.env[NO_CUSTOM]
+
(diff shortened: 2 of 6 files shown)


lib/stylesheet/manager.rb
```diff
@@ -44,6 +44,7 @@ def self.stylesheet_details(target = :desktop, media = 'all', theme_ids = :missi
     theme_ids = [theme_ids] unless Array === theme_ids
     theme_ids = [theme_ids.first] unless target =~ THEME_REGEX
     theme_ids = Theme.transform_ids(theme_ids, extend: false)
+    theme_ids = [nil] if theme_ids.empty? # Build at least one stylesheet without any theme
```
eviltrout: It's still necessary here?

davidtaylorhq: Yes, because we loop over the list to generate the stylesheets

https://github.com/discourse/discourse/blob/583177dd2defea318d04d547cea02d2af9aff10e/lib/stylesheet/manager.rb#L57-L80

In the case where we want 'no themes', we still need to generate at least one stylesheet.

SamSaffron:
seems reasonably safe to me but this code is very high risk to touch, if you are feeling confident it is good, go ahead and merge it.

lib/stylesheet/manager.rb
OsamaSayegh: @davidtaylorhq in case you're going to refactor the CSS procompile logic, we have test case for it that we don't run by default cause it's too slow (10+ seconds if I remember correctly). It should help you catch regressions:

https://github.com/discourse/discourse/blob/2e48c43546cc5b988770170927dfebe07ce5b320/spec/components/stylesheet/manager_spec.rb#L167

davidtaylorhq:
Closing this out for now, I would like to do this properly and remove all trace of `[nil]`

PR #6981, opened 2019-02-06
FIX: translate bootbox confirmation dialog

featheredtoast:
Localize bootbox confirmation buttons - pull from Discourse translation keys

### DIFF
--- a/app/assets/javascripts/discourse/initializers/localization.js.es6
+++ b/app/assets/javascripts/discourse/initializers/localization.js.es6
@@ -55,7 +55,7 @@ export default {
     bootbox.addLocale(I18n.currentLocale(), {
       OK: I18n.t("composer.modal_ok"),
       CANCEL: I18n.t("composer.modal_cancel"),
-      CONFIRM: I18n.t("composer.modal_cancel")
+      CONFIRM: I18n.t("composer.modal_ok")
     });
     bootbox.setLocale(I18n.currentLocale());
   }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/initializers/localization.js.es6
```diff
@@ -51,5 +51,12 @@ export default {
         node[segs[segs.length - 1]] = v;
       }
     });
+
+    bootbox.addLocale(I18n.currentLocale(), {
+      OK: I18n.t("composer.modal_ok"),
+      CANCEL: I18n.t("composer.modal_cancel"),
+      CONFIRM: I18n.t("composer.modal_cancel")
```
ZogStriP: Confirm == cancel?

eviltrout:
Don't we normally pass these values to bootbox? Is this for the case when defaults are not supplied?

I suppose if defaults are supplied we could go back and refactor things to remove extra code.

app/assets/javascripts/discourse/initializers/localization.js.es6
featheredtoast: oops, good catch! was supposed to be modal_ok 🤦‍♂️

featheredtoast:
yeah, I think this is for cases where the buttons are not supplied, which we do at least in ajax-error.js

SamSaffron:
Cool, yeah this is a good change. That said I think we should look at removing bootbox and putting our own abstraction in place longer term that is built in ember and so on so we can control our destiny.

PR #6982, opened 2019-02-07
FIX: Allow delete method in write scope

pdavide:
Simple use case: I want to delete a like in a post using the Discourse API and my User API Key.


SamSaffron:
Good change! I would like to see an integration test here for it though, can you add one? You hand out a token, then try to delete a like.

SamSaffron:
thanks @pdavide ! was missing one more little case so I took it.

PR #6984, opened 2019-02-07
FIX: Disable 'Create Topic' button if tag is staff-only.

nbianca:


app/serializers/tag_serializer.rb
```diff
@@ -1,3 +1,7 @@
 class TagSerializer < ApplicationSerializer
-  attributes :id, :name, :topic_count
+  attributes :id, :name, :topic_count, :staff
+
+  def staff
+    DiscourseTagging.staff_tag_names.include?(name)
```
ZogStriP: Did you check this isn't creating an N+1?

SamSaffron: yeah this looks pretty risky to me:

https://github.com/discourse/discourse/blob/74d2d4f658171bcab41ccd7257495e7e39fb45e4/lib/discourse_tagging.rb#L196-L202

SamSaffron:
got to sort out the N+1 here prior to merging also double check where we are using this serializer, we want to make sure it is not being used on standard topic list pages and so on.

app/assets/javascripts/discourse/routes/tags-show.js.es6
```diff
@@ -111,13 +111,15 @@ export default Discourse.Route.extend({
     ).then(list => {
       if (list.topic_list.tags) {
         tag.set("id", list.topic_list.tags[0].name); // Update name of tag (case might be different)
+        tag.set("staff", list.topic_list.tags[0].staff);
```
tgxworld: `setProperties` here instead?

SamSaffron:
This looks good to me, let's try it out!

PR #6986, opened 2019-02-07
FIX: When disagreeing with a flag that silenced a user, unsilence them

eviltrout:
Previously it would unhide their post but leave them silenced.

This fix also cleans up some of the helper classes to make it easier
to pass extra data to the silencing code (for example, a link to the
post that caused the user to be silenced.)

There used to be a SpamRulesEnforcer class, which took either a post or a
user as an argument, then depending on which one would call another class.
I refactored it to call those classes directly instead.

This patch also refactors the auto_silence specs to avoid using
stubs. I changed an internal method name and everything still passed, so 
that was a bad  omen.


app/services/spam_rule/auto_silence.rb
```diff
@@ -1,37 +1,38 @@
...
+  def should_autosilence?
+    return false if @user.blank?
     return false if @user.staged?
     return false if @user.has_trust_level?(TrustLevel[1])
 
-    if SiteSetting.num_spam_flags_to_silence_new_user > (0) &&
-        SiteSetting.num_users_to_silence_new_user > (0) &&
-        num_spam_flags_against_user >= (SiteSetting.num_spam_flags_to_silence_new_user) &&
-        num_users_who_flagged_spam_against_user >= (SiteSetting.num_users_to_silence_new_user)
+    if SiteSetting.num_spam_flags_to_silence_new_user > 0 &&
```
eviltrout: This is a minor style change. I don't know why you'd want to wrap the right side of a compare in parens.

ZogStriP: Why didn't Rubocop pick that up?

app/services/spam_rule/flag_sockpuppets.rb
```diff
@@ -5,11 +5,13 @@ def initialize(post)
   end
 
   def perform
-    if SiteSetting.flag_sockpuppets && reply_is_from_sockpuppet?
-      flag_sockpuppet_users
-      true
-    else
-      false
+    I18n.with_locale(SiteSetting.default_locale) do
```
ZogStriP: 👍

app/services/user_silencer.rb
```diff
@@ -16,6 +16,12 @@ def self.unsilence(user, by_user = nil, opts = {})
     UserSilencer.new(user, by_user, opts).unsilence
   end
 
+  def self.was_silenced_for?(post)
+    return false unless post.present?
```
ZogStriP: Why not?

```suggestion
    return false if post.blank?
```

app/services/spam_rule/auto_silence.rb
eviltrout: Not sure, I'll take a look today and see if that's an option.

PR #6994, opened 2019-02-08
Remove Elastic email as recommended provided

jlsjonas:
They updated their [pricing](https://elasticemail.com/pricing/) and only allow 100mails/day now.
Given the [integration issues](https://meta.discourse.org/t/remove-or-merge-elastic-email-unsubscribe/70236) & the only advantage towards others being the amount of free emails/day I believe it should be removed.

### DIFF
--- a/docs/INSTALL-email.md
+++ b/docs/INSTALL-email.md
@@ -14,15 +14,6 @@ Enter these values when prompted by `./discourse-setup` per the [install guide](
     SMTP user name?      [SMTP credentials for your domain under domains tab]
     SMTP password?       [SMTP credentials for your domain under domains tab]
 
-#### [Elastic Email][ee] &mdash; 150k emails/month (5k max/day)
-
-    SMTP server address? smtp.elasticemail.com
-    SMTP user name?      [Your registered email address]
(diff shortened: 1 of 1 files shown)


jlsjonas:
(CLA signed)

notstephen:
Their paid plans aren’t widely out of line with the others listed. Those suggestions aren’t only for small communities interested in a free ride.

Would recommend updating their details to reflect the change, rather than unilaterally deciding to remove them completely.

jlsjonas:
>Their paid plans aren’t widely out of line with the others listed. Those suggestions aren’t only for small communities interested in a free ride.

I understand, however as mentioned in the for-removal section and on Discourse meta, the main issue with elastic is the double-footer (which also causes re-subscribe issues); as such I don't think it should be considered as a recommendation (unless that issue changes).

I believe that the main reason it was kept on the list is that of the generous free plan; given the multiple integration issues.

SamSaffron:
I am making the executive call of pulling this documentation. If this thing needs large amounts of TLC documentation can live in meta I don't want to carry this around in the repo.

PR #6998, opened 2019-02-12
FIX: remove extra periods

featheredtoast:
First phase of getting proper translations -- removing hardcoded fullstop characters.

Getting less segmented translations for the bottom footers might be a little trickier than originally anticipated since we're segmenting sentences around components. This at least removes the assumption of a fullstop characters for the time being until we can do something better for this case.


SamSaffron:
This is clearly a good change, lets see what shakes out!

PR #6999, opened 2019-02-13
UX: make adding permissions more visible and intuitive to click.

peterlamber:
changed button-text to "+"
changed "Add Permission" button to appear in the same line as
permission value combo-boxes.
referencing: https://meta.discourse.org/t/auto-add-permissions-unsaved-permissions-prompt/100919


peterlamber:
> 1 Error
> 	Please submit your non-English translation updates via [Transifex](https://www.transifex.com/discourse/discourse-org/). You can read more on how to contribute translations [here](https://meta.discourse.org/t/contribute-a-translation-to-discourse/14882).
> Generated by  [Danger](https://danger.systems/)

I deleted the text, therefore this might be ignored?

ZogStriP:
No, this should not be ignored. When you delete a translation key, you should only delete it in the `en` locale. The other locales will be kept in sync via Transifex.

peterlamber:
thank you, I already signed it

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

peterlamber:
> No, this should not be ignored. When you delete a translation key, you should only delete it in the `en` locale. The other locales will be kept in sync via Transifex.
> 
> Also, are you sure this string is not used anywhere else?

Ok thanks, I reversed the change and deleted the key only in the `en` locale.

Yes I'm sure, I searched the project for this key and it's the only place it is used.

SamSaffron:
hold tight here, I worry that this will not fit in other locales, Add can be quite a few pixels longer.

SamSaffron:
Discussing here: https://meta.discourse.org/t/auto-add-permissions-unsaved-permissions-prompt/100919/7?u=sam

app/assets/stylesheets/common/base/modal.scss
```diff
@@ -432,7 +436,12 @@
   }
 
   .add-permission {
-    margin-top: 12px;
+    margin-left: 0.25em;
+    position: absolute;
```
awesomerobot: Shouldn't need `position: absolute;` here, it causes vertical overflow at narrow widths (~510px). It's ok that the button can wrap without it. 

Also, because the button can wrap the `margin-left` should be changed to a `margin-right` on the `permission-selector` dropdown. This way if the button wraps onto the next line there's not an unneeded space to the left of it.

peterlamber: Thanks a lot for the review, I changed it accordingly. 
In order to make the add-button align with the combo-boxes, I did the following:
`  .add-permission {
    position: relative;
    top: 0.1em;
  }`

PR #7003, opened 2019-02-13
FEATURE: Automatically update MaxMindDb every 7 days.

nbianca:


app/jobs/scheduled/update_maxminddb.rb
```diff
@@ -0,0 +1,12 @@
+require_dependency 'discourse_ip_info'
+
+module Jobs
+  class UpdateMaxMindDb < Jobs::Scheduled
+    every 7.days
+
+    def execute(args)
+      DiscourseIpInfo.update!
+      DiscourseIpInfo.reload
```
ZogStriP: I think `update!` should take care of calling `reload`.

lib/discourse_ip_info.rb
```diff
@@ -83,11 +90,29 @@ def get(ip, locale: :en, resolve_hostname: false)
...
-  def self.open_db(path)
-    instance.open_db(path)
-  end
-
   def self.get(ip, locale: :en, resolve_hostname: false)
     instance.get(ip, locale: locale, resolve_hostname: resolve_hostname)
   end
+
+  def self.download_mmdb(name)
+    uri = URI("http://geolite.maxmind.com/download/geoip/database/#{name}.tar.gz")
```
ZogStriP: We should use https if it's available.

SamSaffron:
I am so so sorry, but I am closing this, I don't feel we need to carry a job for this, if people want to update out of band they can run the rake task. We ship a new image every few months and I feel that is good enough cadence for regular updates.

itsbhanusharma:
@SamSaffron I can't find the appropriate rake task to update maxminddb Please help?

PR #7004, opened 2019-02-14
FEATURE: Add localized timezone names

pmusaraj:
Uses https://github.com/discourse/moment-timezone-names-translations to translate timezone names (Moment Timezone has no translations for timezone names). 

Easiest place to see this in action is the timezone dropdown when inserting a date in the composer (under the Advanced toggle). 

For example: 

image


lib/js_locale_helper.rb
```diff
@@ -156,13 +156,18 @@ def self.output_locale(locale)
     result << File.read("#{Rails.root}/vendor/assets/javascripts/moment.js")
     result << File.read("#{Rails.root}/vendor/assets/javascripts/moment-timezone-with-data.js")
     result << moment_locale(locale_str)
+    result << moment_locale(locale_str, true) # load timezones name locales
```
gschlager: The fact that you feel the need to add a comment is a good indicator that the method signature isn't clear enough. It would be a good case for a keyword argument.

```ruby
def self.find_moment_locale(locale_chain, timezone_names: false)
```

pmusaraj: Thanks, that's much better.

pmusaraj:
> I think plugins should be able to provide such translations too, so it would be a good idea to add support for an additional `moment_js_timezones` folder similar to the existing `moment_js`. 

I spent a fair bit of time today trying to get the current `moment_js` implementation to work locally, but it didn't work. Is it possible we have recently regressed here? (I did move the core moment locale files under `vendor/assets/javascripts/moment-locale` recently, though that doesn't seem to be the issue).
 
Maybe this is because the moment-specific naming scheme we use here: 
https://github.com/discourse/discourse/blob/master/lib/js_locale_helper.rb#L167-L168
is missing from the plugin instance code? 

In my tests, I used: 
```
register_locale("es_MX", name: "Spanish (Mexico)", nativeName: "Español (México)", fallbackLocale: "es")
```

and placed the relevant file in: 
```
[plugin]/lib/javascripts/locale/moment_js
```

Tried both `es-mx.js` and `es.js` unsuccessfully.

gschlager:
You were right. `moment_js` locale files from plugins were ignored when the locale name contained an underscore or uppercase letters. I fixed it in https://github.com/discourse/discourse/commit/4799cf281131e1bf34fcb6aad49b00769cff469e

Putting the locale file in `[plugin]/lib/javascripts/locale/moment_js/es_MX.js` should work now. You might need to clear the `tmp` folder before starting the rails server in your development environment if it doesn't work immediately.

pmusaraj:
Thanks @gschlager ! I have updated the PR to add support for timezone translations to plugins.

PR #7006, opened 2019-02-14
FIX: Use CDN for optimized loading images.

udan11:


lib/cooked_post_processor.rb
```diff
@@ -412,7 +412,7 @@ def add_lightbox!(img, original_width, original_height, upload, cropped: false)
       end
 
       if small_upload = loading_image(upload)
-        img["data-small-upload"] = small_upload.url
+        img["data-small-upload"] = UrlHelper.cook_url(small_upload.url)
```
ZogStriP: Nice 👍 Is that the only spot we missed?

udan11: It's the only spot I did not miss too. :)

spec/components/cooked_post_processor_spec.rb
```diff
@@ -291,7 +291,7 @@
 
         # Fake a loading image
         optimized_image = OptimizedImage.create!(
-          url: 'http://a.b.c/10x10.png',
+          url: '/uploads/default/10x10.png',
```
tgxworld: Since we're changing this in the PR, I think it is a good time to fix it such that the URL actually matches the URL format for an optimized image that the app creates. I've been bitten many time by bugs with uploads where we mock URLs with random format in the tests.

SamSaffron:
I want this fix in now, so Dan can follow up a bit more if needed next week.

PR #7009, opened 2019-02-14
FIX: Use field name to name files in theme export.

udan11:


app/models/theme_field.rb
```diff
@@ -281,6 +281,13 @@ def target_name
     Theme.targets.invert[target_id].to_s
   end
 
+  def export_path
```
davidtaylorhq: I think it would be nice to keep a single api for `file_path` - it might get confusing having both `export_path` and `file_path`. How about moving this logic into the lambda here: 

https://github.com/discourse/discourse/blob/49c172e2242394b4567102a4504e78d3bb0507d2/app/models/theme_field.rb#L349

That lambda is already passed a hash which includes `name` and `filename`.

udan11: That turned out a lot better.

spec/components/theme_store/tgz_exporter_spec.rb
```diff
@@ -11,6 +11,9 @@
       image = file_from_fixtures("logo.png")
       upload = UploadCreator.new(image, "logo.png").create_for(-1)
       theme.set_field(target: :common, name: :logo, upload_id: upload.id, type: :theme_upload_var)
+      image = file_from_fixtures("logo.png")
+      other_upload = UploadCreator.new(image, "logo.png").create_for(-1)
```
eviltrout:
```suggestion
      other_upload = UploadCreator.new(image, "logo.png").create_for(Discourse.system_user.id)
```

app/models/theme_field.rb
```diff
@@ -339,7 +339,7 @@ def filename_from_opts(opts)
                          canonical: -> (h) { "locales/#{h[:name]}.yml" }),
     ThemeFileMatcher.new(regex: /(?!)/, # Never match uploads by filename, they must be named in about.json
                          names: nil, types: :theme_upload_var, targets: :common,
-                         canonical: -> (h) { "assets/#{h[:filename]}" }),
+                         canonical: -> (h) { "assets/#{h[:name].gsub(/[^0-9A-Za-z.\-]/, '_')}#{File.extname(h[:filename])}" }),
```
eviltrout: I don't see a test here for the filename substitution.

spec/components/theme_store/tgz_exporter_spec.rb
ZogStriP: I tend to prefer 

```suggestion
      other_upload = UploadCreator.new(image, "logo.png").create_for(Discourse::SYSTEM_USER_ID)
```

app/models/theme_field.rb
udan11: I originally added that substitution more as a precaution, because the name of a theme field is already validated [here](https://github.com/discourse/discourse/blob/cfddfa6de2bba1f631cde9054a90bf99910ae2a7/app/models/theme_field.rb#L65).

PR #7010, opened 2019-02-15
FIX: make it possible to use backup code everywhere where 2FA required

majakomel:
https://meta.discourse.org/t/disable-two-factor-auth-with-only-backup-codes/108588

### DIFF
--- a/app/assets/javascripts/discourse/components/second-factor-form.js.es6
+++ b/app/assets/javascripts/discourse/components/second-factor-form.js.es6
@@ -32,11 +32,7 @@ export default Ember.Component.extend({
   actions: {
     toggleSecondFactorMethod() {
       const secondFactorMethod = this.get("secondFactorMethod");
-      if (this.get("loginSecondFactor")) {
-        this.set("loginSecondFactor", "");
-      } else if (this.get("secondFactorToken")) {
-        this.set("secondFactorToken", "");
-      }
+      this.set("secondFactorToken", "");
       if (secondFactorMethod === SECOND_FACTOR_METHODS.TOTP) {
         this.set("secondFactorMethod", SECOND_FACTOR_METHODS.BACKUP_CODE);
       } else {
--- a/app/assets/javascripts/discourse/components/second-factor-form.js.es6
+++ b/app/assets/javascripts/discourse/components/second-factor-form.js.es6
@@ -16,17 +16,27 @@ export default Ember.Component.extend({
       : I18n.t("login.second_factor_backup_description");
   },
 
-  @computed("secondFactorMethod")
-  linkText(secondFactorMethod) {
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -367,20 +367,20 @@ const User = RestModel.extend({
     });
   },
 
-  toggleSecondFactor(token, enable, method) {
+  toggleSecondFactor(authToken, authMethod, targetMethod, enable) {
     return ajax("/u/second_factor.json", {
       data: {
-        second_factor_token: token,
-        second_factor_method: method,
+        second_factor_auth: [authToken, authMethod],
```
tgxworld: I think that changing this to an array makes it confusing in terms of what values the API expects. Is there a reason we need to make it an array instead of using one key per value?

majakomel: I used an array to group together 2fa method and key we're authenticating with and 2fa that should be modified. I will split the array in separate params then to avoid confusion.

SamSaffron:
I am good to have this merged in @majakomel  once the array is split back to 2 params.

app/assets/javascripts/discourse/components/second-factor-form.js.es6
```diff
@@ -16,17 +16,27 @@ export default Ember.Component.extend({
...
+        ? "user.second_factor_backup.use"
+        : "user.second_factor.use";
+    }
   },
 
   actions: {
     toggleSecondFactorMethod() {
       const secondFactorMethod = this.get("secondFactorMethod");
-      this.set("loginSecondFactor", "");
+      if (this.get("loginSecondFactor")) {
```
featheredtoast: Wondering what the purpose of checking for if loginSecondFactor before setting it. Is it performance or functional?

app/assets/javascripts/discourse/controllers/password-reset.js.es6
```diff
@@ -9,7 +9,7 @@ export default Ember.Controller.extend(PasswordValidation, {
   isDeveloper: Ember.computed.alias("model.is_developer"),
   admin: Ember.computed.alias("model.admin"),
   secondFactorRequired: Ember.computed.alias("model.second_factor_required"),
-  backupEnabled: Ember.computed.alias("model.second_factor_backup_enabled"),
+  backupEnabled: Ember.computed.alias("model.backup_enabled"),
```
ZogStriP: Was this changed?

majakomel: No, this was wrong and not used until now (it comes from here https://github.com/discourse/discourse/blob/master/app/controllers/users_controller.rb#L528)

PR #7011, opened 2019-02-15
FEATURE: allow blocking emojis

techAPJ:
https://meta.discourse.org/t/blocking-emojis-wont-work/105853


techAPJ:
Thanks for the review @eviltrout ❤️

PR #7012, opened 2019-02-15
REFACTOR: Proxy letter avatars in rails instead of nginx

davidtaylorhq:
- This change gives us more control over the request. In particular we can easily lookup DNS dynamically, instead of only upon nginx startup.
- The proxy path has been renamed from `letter_avatar_proxy` to `letter_proxy`, so that existing nginx config doesn't prevent use. The old path remains functional.
- Nginx config has been updated to add caching for both paths. This change will require a container rebuild.
- The proxy will now function in development environments, so the patch for `letter_avatar_proxy` has been removed

### DIFF
--- a/app/controllers/user_avatars_controller.rb
+++ b/app/controllers/user_avatars_controller.rb
@@ -35,7 +35,7 @@ class UserAvatarsController < ApplicationController
   def show_proxy_letter
     is_asset_path
 
-    if SiteSetting.external_system_avatars_url !~ /^\/letter_avatar_proxy/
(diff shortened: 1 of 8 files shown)


SamSaffron:
hmmmm why is it removing the direct letter avatar generation? technically that is still supported if people want 100% local

SamSaffron:
I see ... yeah ... hmmm maybe there is more code we should nuke ... I am going to amend this a tiny bit and keep the old path... all the rest can stay.

davidtaylorhq:
Changed and merged in in https://github.com/discourse/discourse/commit/f04471e4229218903df17232db955cd583a423ef

PR #7022, opened 2019-02-18
UX: Allow users to see filename in image-uploader component.

tgxworld:
https://meta.discourse.org/t/downsides-of-the-new-logo-ui-in-site-settings/102247/23?u=tgxworld

screenshot from 2019-02-18 21-40-43

screenshot from 2019-02-18 21-40-35

@awesomerobot @hnb-ku I opted for a button toggle instead of hover like what we did for lightboxes because hover doesn't work on mobile. Would love to get your thoughts on the changes here :)

### DIFF
--- a/app/assets/stylesheets/common/base/upload.scss
+++ b/app/assets/stylesheets/common/base/upload.scss
@@ -1,6 +1,53 @@
 .uploaded-image-preview {
   background: $primary center;
   background-size: cover;
+  position: relative;
+
+  .image-uploader-info {
+    position: absolute;
+    bottom: 0;
+    width: 100%;
+    background: $primary;
+    text-align: center;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    opacity: 0.6;
+
+    a {
+      color: $secondary;
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+  }
+
+  .image-upload-controls {
+    display: flex;
+
+    .btn {
+      margin-right: 5px;
+    }
+
+    .image-uploader-info-btn {
+      background: none;
+      margin-right: 0;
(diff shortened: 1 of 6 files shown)


jjaffeux:
Maybe we can make it use d-popover at some point

app/assets/stylesheets/common/base/upload.scss
```diff
@@ -1,6 +1,32 @@
...
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    opacity: 0.6;
+  }
+
+  .image-upload-controls {
+    display: flex;
+
+    .image-uploader-info-btn {
+      background: none;
```
awesomerobot: I think this icon could be made a little more visible with

```
.d-icon {
  color: $primary-low;
}
```

app/assets/stylesheets/common/base/upload.scss
```diff
@@ -1,6 +1,32 @@
...
+    text-align: center;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    opacity: 0.6;
+  }
+
+  .image-upload-controls {
+    display: flex;
+
```
awesomerobot: The delete button could use a bit of space

screen shot 2019-02-18 at 3 00 47 pm


```
.btn-danger {
  margin-left: .5em;
}
```

tgxworld:
> Maybe we can make it use d-popover at some point

@jjaffeux Has `d-popover` been implemented? I can't seem to find the component.

> Would it be easy to make the filename link directly to the image? then there'd be a simple way to view the full path without trying to cram it in to the limited space available.

@awesomerobot Do you mean that we should make the filename trigger a lightbox? Thanks for the review :+1:

PR #7025, opened 2019-02-19
DEV: Allow validity of lock to be customizable for `DistributedMutex`.

tgxworld:
- Allows a user to override the default lock validity of 60 seconds.
- Also clean up test which was leaking a redis key

### DIFF
--- a/lib/distributed_mutex.rb
+++ b/lib/distributed_mutex.rb
@@ -1,8 +1,9 @@
 # Cross-process locking using Redis.
 class DistributedMutex
+  DEFAULT_VALIDITY = 60
 
-  def self.synchronize(key, redis = nil, &blk)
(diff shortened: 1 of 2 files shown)


lib/distributed_mutex.rb
```diff
@@ -1,8 +1,9 @@
 # Cross-process locking using Redis.
 class DistributedMutex
+  DEFAULT_VALIDITY = 60
 
-  def self.synchronize(key, redis = nil, &blk)
-    self.new(key, redis).synchronize(&blk)
+  def self.synchronize(key, redis = nil, validity = DEFAULT_VALIDITY, &blk)
```
eviltrout: Might be a good candidate for a keyword arg

tgxworld:
Thanks for reviewing @eviltrout :+1:

PR #7026, opened 2019-02-19
FEATURE: Added settings/translations support to theme editor UI

davidtaylorhq:
- These advanced fields are hidden behind an 'advanced' button, so will not affect normal use
- The editor has been refactored into a component, and styling cleaned up so menu items do not overlap on small screens
- Styling has been added to indicate which fields are in use for a theme
- Icons have been added to identify which fields have errors

screenshot 2019-02-19 at 11 22 20


app/assets/javascripts/admin/templates/components/admin-theme-editor.hbs
```diff
@@ -0,0 +1,94 @@
+<div class='edit-main-nav admin-controls'>
+    <nav>
+    <ul class='nav nav-pills target'>
+        {{#each visibleTargets as |target|}}
+        <li>
+            {{#link-to editRouteName
+            theme.id
```
eviltrout: The spacing is quite odd here.

app/assets/javascripts/admin/templates/components/admin-theme-editor.hbs
```diff
@@ -0,0 +1,94 @@
...
+            {{/link-to}}
+        </li>
+        {{/each}}
+        <li>
+        <a {{action "toggleShowAdvanced"}} 
+            class='no-text' 
+            title="{{i18n (concat "admin.customize.theme." (if showAdvanced "hide_advanced" "show_advanced"))}}"
+        >
+            {{#if showAdvanced}}
+            {{d-icon "angle-double-left"}}
```
eviltrout: I think we could use the ember if helper here.

`{{d-icon (if showAdvanced "angle-double-left" "angle-double-right")}}`

PR #7028, opened 2019-02-19
FIX: removes legacy browsers-refresh button

jjaffeux:


gschlager:
Just a small note: It's fine to remove translations from all locale files, but you don't need to. The next pull from Transifex has the same effect. Removing the English translations is enough.

jjaffeux:
> Just a small note: It's fine to remove translations from all locale files, but you don't need to. The next pull from Transifex has the same effect. Removing the English translations is enough.

ahaha I was about to send you a message about this :p Thanks

PR #7030, opened 2019-02-19
FIX: Create and consume temp table inside a transaction

vinothkannans:
To prevent access issue in pgbouncer which runs in transaction pooling.


SamSaffron:
This should be good, thanks for the fix, note ... indenting is a bit our of control deep here, maybe a good idea to split up the method a bit.

PR #7031, opened 2019-02-19
FIX: Sanitize theme settings properly before injecting into stylesheets

davidtaylorhq:
We were not fully escaping variables when injecting them into the SASS compiler. This can cause some unusual errors. Note this is **not** a security issue, it can just cause unexpected theme compilation errors if admins enter certain characters into theme settings.

The complication is that some themes have made use of this bug, so we should wait for them to be patched before merging this PR. 

https://github.com/hnb-ku/discourse-vincent-theme/pull/2

### DIFF
--- a/lib/stylesheet/importer.rb
+++ b/lib/stylesheet/importer.rb
@@ -144,9 +144,8 @@ COMMENT
     end
 
     def to_scss_variable(name, value)
-      escaped = value.to_s.gsub('"', "\\22")
-      escaped.gsub!("\n", "\\A")
-      "$#{name}: unquote(\"#{escaped}\");\n"
+      escaped = SassC::Script::Value::String.quote(value, sass: true)
+      "$#{name}: unquote(#{escaped});\n"
     end
 
     def imports(asset, parent_path)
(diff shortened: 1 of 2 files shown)


SamSaffron:
We can't sit on this for so long, can you merge this in on Monday once @hnb-ku  is back and coordinate with him?

lib/stylesheet/importer.rb
```diff
@@ -144,9 +144,8 @@ def category_css(category)
     end
 
     def to_scss_variable(name, value)
-      escaped = value.to_s.gsub('"', "\\22")
-      escaped.gsub!("\n", "\\A")
-      "$#{name}: unquote(\"#{escaped}\");\n"
+      escaped = SassC::Script::Value::String.quote(value, sass: true)
```
khalilovcmded: Do we need to do this with other parts of the import process that has the potential to need escaping?
In other words, do we need do this at the top of the import funnel?

davidtaylorhq: Which "import funnel"? As far as I know, this is the only place that we inject user input directly into SCSS variables. If there is somewhere else, then yes we should use this escaping method.

davidtaylorhq:
> We can't sit on this for so long

Is this causing an issue somewhere? I already fixed the error handling, so this shouldn't be causing any problems. Will co-ordinate with @hnb-ku to get it merged next week without affecting too many users.

lib/stylesheet/importer.rb
khalilovcmded: > As far as I know, this is the only place that we inject user input directly into SCSS variables.

👍 My bad, I thought that there will be other places within the stylesheet importer where we need to escaping.

PR #7033, opened 2019-02-20
DEV: add function to check whether job with same params has been enqueued

mrfinch:


mrfinch:
This is in relation to this PR: https://github.com/discourse/discourse-akismet/pull/19

jjaffeux:
I think `Jobs.enqueued?` would be better than `Jobs.has_same_job_been_enqueued?`

@ZogStriP ?

app/jobs/base.rb
```diff
@@ -253,20 +253,38 @@ def self.scheduled_for(job_name, opts = {})
...
-        end
-        matched
+        has_required_params?(job_params, opts)
       else
         false
       end
     end
   end
+
+  def self.has_same_job_been_enqueued?(job_name, opts)
```
ZogStriP:
```suggestion
  def self.enqueued?(job_name, opts)
```

mrfinch: @ZogStriP there has been a discussion on whether we should commit or not this: https://github.com/discourse/discourse-akismet/pull/19 , can you give your input?

mrfinch: I like the idea of adding this function but it can cause issues too if you load a very high number of jobs in memory. I m not going to use it in Akismet plugin but this code is independent of it and could be added and used later by someone.

SamSaffron:
I am sorry I am a bit uneasy about this change it feels somewhat high risk to me due to the performance profile. I am uneasy about adding this API for now.

PR #7043, opened 2019-02-21
UX: Create 32x32 thumbnail for favicons.

tgxworld:
- Also read file directly from disk for local store instead of having to
download the file over the network.

- External stores will still have to be downloaded.

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -102,6 +102,7 @@ class StaticController < ApplicationController
   end
 
   FAVICON ||= -"favicon"
+  FAVICON_SIZE = 32
 
   # We need to be able to draw our favicon on a canvas, this happens when you enable the feature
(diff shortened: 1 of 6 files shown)


app/controllers/static_controller.rb
```diff
@@ -118,22 +119,36 @@ def favicon
...
-            max_file_size: 50.kilobytes,
-            tmp_file_name: FAVICON,
-            follow_redirect: true
-          )
-          file ||= Tempfile.new([FAVICON, ".png"])
-          data = file.read
-          file.unlink
-          data
-        rescue => e
-          AdminDashboardData.add_problem_message('dashboard.bad_favicon_url', 1800)
```
tgxworld: The chance of the file missing is very low here so I don't think we need this message anymore.

tgxworld:
Closing this first as I need to rethink how we can support this for all uploads site settings.

PR #7044, opened 2019-02-21
DEV: remove user_profile events added for akismet

mrfinch:
cc @tgxworld

### DIFF
--- a/app/models/user_profile.rb
+++ b/app/models/user_profile.rb
@@ -8,8 +8,6 @@ class UserProfile < ActiveRecord::Base
   validates :user, presence: true
   before_save :cook
   after_save :trigger_badges
-  after_commit :trigger_profile_created_event, on: :create
-  after_commit :trigger_profile_updated_event, on: :update
 
   validates :profile_background, upload_url: true, if: :profile_background_changed?
   validates :card_background, upload_url: true, if: :card_background_changed?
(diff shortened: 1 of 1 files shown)


tgxworld:
This will break code that is still subscribed to the existing event. For what you're working on, you can do the following.

```
on(:user_profile_updated) do |user|
  if self.bio_raw_previous_changed?
    # do stuff
  end
end
```

mrfinch:
> This will break code that is still subscribed to the existing event. For what you're working on, you can do the following.
> 
> ```
> on(:user_profile_updated) do |user|
>   if self.bio_raw_previous_changed?
>     # do stuff
>   end
> end
> ```

@tgxworld , it is not used anywhere. I added it for Akismet plugin itself a week back.

tgxworld:
In that case, we don’t need the events. You can just add callbacks directly in the plugin. There is no need to use callbacks to trigger events that is then consumed.

mrfinch:
> In that case, we don’t need the events. You can just add callbacks directly in the plugin. There is no need to use callbacks to trigger events that is then consumed.

I didn't know I could do that directly. I have made changes accordingly in that PR: https://github.com/discourse/discourse-akismet/pull/19 , can you have a look if it's fine? I'm updating this PR to remove the user_profile events. @tgxworld

ZogStriP:
Let's remove them since they're not useful.

PR #7056, opened 2019-02-23
FIX: allows {{d-button}} to set a form attribute

jjaffeux:
This attribute is used when a submit button is out of a form. It makes it explicit which form this button is submitting.

It's currently used in our login modal form.

### DIFF
--- a/app/assets/javascripts/discourse/components/d-button.js.es6
+++ b/app/assets/javascripts/discourse/components/d-button.js.es6
@@ -4,9 +4,12 @@ export default Ember.Component.extend({
   // subclasses need this
   layoutName: "components/d-button",
 
+  form: null,
+
   tagName: "button",
   classNameBindings: [":btn", "noText", "btnType"],
   attributeBindings: [
+    "form",
     "disabled",
     "translatedTitle:title",
     "translatedLabel:aria-label",
--- a/app/assets/javascripts/discourse/templates/modal/login.hbs
+++ b/app/assets/javascripts/discourse/templates/modal/login.hbs
@@ -37,11 +37,13 @@
 
   <div class="modal-footer">
     {{#if canLoginLocal}}
-      {{d-button action=(action "login")
-          icon="unlock"
-          label=loginButtonLabel
-          disabled=loginDisabled
-          class='btn btn-large btn-primary'}}
+      {{d-button
+        action=(action "login")
+        form="login-form"
+        icon="unlock"
+        label=loginButtonLabel
+        disabled=loginDisabled
+        class='btn btn-large btn-primary'}}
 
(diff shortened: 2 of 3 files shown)


maximbaz:
Hi @jjaffeux, I'm working on a password manager and would like to support Discourse login forms. Thanks for this fix, having the `form` attribute set correctly is important to identify that the button does indeed belong to the login form.

Can I also ask you to add some additional metadata that would help me identify that this button is indeed a "login" button? Especially in cases where a browser extension clicks the "login" button for user, I would like to be _really sure_ that the button I'm about to click is indeed a login button.

Contrary to `input[type=submit]` there could be many `button`s in a form, and even if there's only one, it isn't necessarily a "submit" button.

Currently I'm identifying "login" buttons based on `name`, `id` or `class` attribute, so if you can make sure that any of these attributes contains any of these strings, it would really help password managers: "login", "log-in", "log_in", "signin", "sign-in", "sign_in".

Basically I want your login button to match any of these selectors using `document.querySelectorAll`:

- "button[name*=login i]",
- "button[name*=log-in i]",
- "button[name*=log_in i]",
- "button[name*=signin i]",
- "button[name*=sign-in i]",
- "button[name*=sign_in i]",
- "button[id*=login i]",
- "button[id*=log-in i]",
- "button[id*=log_in i]",
- "button[id*=signin i]",
- "button[id*=sign-in i]",
- "button[id*=sign_in i]",
- "button[class*=login i]",
- "button[class*=log-in i]",
- "button[class*=log_in i]",
- "button[class*=signin i]",
- "button[class*=sign-in i]",
- "button[class*=sign_in i]",

cc @mcnesium who initially discovered the issues with Discourse login form, and who has been communicating with you guys before [here](https://meta.discourse.org/t/login-form-is-invalid-html/87735) and [here](https://meta.discourse.org/t/login-button-form-attribute/109828).

coding-horror:
If it works for the other popular password managers, which many people in
our company currently use .. why does it need to be changed just for you?

maximbaz:
I can only speculate how other password managers work, maybe they are not so strict and choose to click random button in the login form, but because on Discourse login buttons are not identifiable in any way, I will not let my password manager click it. Therefore the ask to identify via "id", "name" or "class" attribute that this is indeed a login button 😉 

image

The fact that it belongs to "login-form" does not mean that this is a login button.

mcnesium:
@coding-horror this is not just about this one password manager, but also a general accessibility issue. 

also: [why is my initial report gone](https://meta.discourse.org/t/login-form-is-invalid-html/87735)?

coding-horror:
Why does it work for the other common password managers? What are you doing
differently that requires this change?

PR #7058, opened 2019-02-24
FIX: users list show was loading multiple times with different params

jjaffeux:
A first load was happening in route, which was setting properties on controller. These properties were observed on the controller and were triggering a reload of the AdminUser model.

Not only was it doing loading two times it was also sometimes resulting on the controller model refresh end to happen after route has been changed, resulting in a wrong model.

### DIFF
--- a/app/assets/javascripts/admin/routes/admin-users-list-show.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-users-list-show.js.es6
@@ -4,6 +4,8 @@ export default Discourse.Route.extend({
     ascending: { refreshModel: true }
   },
 
+  // TODO: this has been introduced to fix a bug in admin-users-list-show
+  // loading AdminUser model multiple times without refactoring the controller
   beforeModel(transition) {
     const routeName = "adminUsersList.show";
 
--- a/app/assets/javascripts/admin/routes/admin-users-list-show.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-users-list-show.js.es6
@@ -1,17 +1,26 @@
-import AdminUser from "admin/models/admin-user";
-
 export default Discourse.Route.extend({
-  model: function(params) {
-    this.userFilter = params.filter;
-    return AdminUser.findAll(params.filter);
+  queryParams: {
+    order: { refreshModel: true },
+    ascending: { refreshModel: true }
   },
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/admin/routes/admin-users-list-show.js.es6
```diff
@@ -1,17 +1,26 @@
...
   },
 
-  setupController: function(controller, model) {
-    controller.setProperties({
-      model: model,
-      query: this.userFilter,
-      showEmails: false,
-      refreshing: false
-    });
+  beforeModel(transition) {
```
jjaffeux: This is questionably not the best solution but changing this would I think  require an heavy refactoring of how admin-users-list-show is working. I think this is a good tradeoff.

eviltrout: It's not ideal, but I am cool with it as a compromise. I would suggest that you add a comment in the code explaining why it's done this way though for future developers.

test/javascripts/acceptance/admin-users-list-test.js.es6
```diff
@@ -8,3 +8,42 @@ QUnit.test("lists users", async assert => {
   assert.ok(exists(".users-list .user"));
   assert.ok(!exists(".user:eq(0) .email small"), "escapes email");
 });
+
+QUnit.test("switching tabs", async assert => {
+  const activeUser = "<small>eviltrout@example.com</small>";
```
eviltrout: Weird, you are calling `text` and it's returning the `<small>` tags? I thought it only ever returned text.

Since this is the case, think it would be better if your selector was `... .email small` instead so you can compare the email address without markup.

jjaffeux: This is done this way because this is escaped and I would need to htmlilfy it first before being able to search inside it. See the existing test:

```
QUnit.test("lists users", async assert => {
  await visit("/admin/users/list/active");

  assert.ok(exists(".users-list .user"));
-->  assert.ok(!exists(".user:eq(0) .email small"), "escapes email");
});
```

eviltrout: Understood, thanks for clarifying!

PR #7059, opened 2019-02-25
FIX: 'guidelines_topic.body' in Brazillian Portuguese

dukex:
Hi everyone, I'm building a community in Brazil, and when I was making the community setup, I saw some wrong things in `guidelines_topic.body` in Brazilian Portuguese translation.

To make the review and my explanation easier,  I splitted this PR in the follow parts/commit:

### Commit d5053cd 

I moved `guidelines_topic.body` to multi line version, looking others translation like [zh_TW](https://github.com/discourse/discourse/blob/55cc5ab4b8fb52bfea05659149fdcf6437d33a3c/config/locales/server.zh_TW.yml), [sk](https://github.com/discourse/discourse/blob/55cc5ab4b8fb52bfea05659149fdcf6437d33a3c/config/locales/server.sk.yml), [zh_CN](https://github.com/discourse/discourse/blob/55cc5ab4b8fb52bfea05659149fdcf6437d33a3c/config/locales/server.zh_CN.yml) and many others, I believe it's better to maintenance keep this long html/markdown content in multi line version instead of one line, will be easy see the next changes in the multi line version too .
  
### Commit 11174172eed5ef6d5f8cab0416f1c8833b22ec8b 

The first wrong thing is the links in titles, I don't know why, but it's was translated, but the anchors was not, of course it'll not work. Again, looking the others translations, I keep the links as the anchors.

### Commit  96591ebf0b8704ec60d7a54fc1c7d31272f0df4b

Maybe you notice, in last commit, we don't have a real link in markdown to the titles, because there is a space between `]`(square brackets) and `(`(round brackets), this commit fix it.

### Commit 36072a8f52534d244363cdafd05bb30b0a103fe0

Looking the text a bit closer, I found some htmlentities with a space between the `&`(amp) and the next letter, the result is the  htmlentities will not be rendered. Fixed!

### Commit b37ea70fe567a82aab85dcdad6d58812992d175e

Fixing the htmlentities, I found some links with space between the `/`, I don't know why, but it's wrong and I fixed it.

--- 

The final result was something like it:

screenshot from 2019-02-24 20-48-09

screenshot from 2019-02-24 20-48-24

### DIFF
--- a/config/locales/server.pt_BR.yml
+++ b/config/locales/server.pt_BR.yml
@@ -2532,7 +2532,7 @@ pt_BR:
     body: | 
       <a name="civilized"></a>
       
-      ## [Este é um lugar civilizado para discussão pública] (#civilized)
+      ## [Este é um lugar civilizado para discussão pública](#civilized)
       
       Por favor, trate este fórum de discussão com o mesmo respeito que você faria com um parque público. Nós também somos um recurso da comunidade compartilhada & mdash; um lugar para compartilhar habilidades, conhecimento e interesses por meio de conversas contínuas.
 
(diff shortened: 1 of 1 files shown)


ZogStriP:
Thanks @dukex but all the non-English translations have to go through https://www.transifex.com/discourse/discourse-org/ otherwise they'd be lost on the next update.

PR #7060, opened 2019-02-25
FIX: ensures topic’s category allows topics tags

jjaffeux:


app/controllers/topics_controller.rb
```diff
@@ -288,6 +288,14 @@ def update
       else
         return render_json_error(I18n.t('category.errors.not_found'))
       end
+
+      if category && topic_tags = params[:tags] || topic.tags.pluck(:name)
+        allowed_tags = category.tags.pluck(:name)
+
+        if !(topic_tags.blank? || allowed_tags.blank?) && !(topic_tags - allowed_tags).empty?
```
eviltrout: Since you're performing the set difference twice `topic_tags - allowed_tags`) it would be more efficient to store the result in a variable.

spec/requests/topics_controller_spec.rb
```diff
@@ -1011,6 +1011,46 @@ def topic_user_post_timings_count(user, topic)
...
+            result = ::JSON.parse(response.body)
+
+            expect(response.status).to eq(422)
+            expect(result['errors']).to eq([I18n.t('category.errors.disallowed_topic_tags', tags: [tag1].map(&:name).join(", "))])
+            expect(topic.reload.category_id).not_to eq(restricted_category.id)
+          end
+
+          it 'can change to a category allowing this topic current tags' do
+            restricted_category.allowed_tags = [tag1.name]
+            restricted_category.reload
```
eviltrout: These `restricted_category.reload` calls are confusing to me. Why would this be necessary unless your test is checking the reloaded value?

spec/requests/topics_controller_spec.rb
```diff
@@ -1011,6 +1011,46 @@ def topic_user_post_timings_count(user, topic)
...
+          it 'can change to a category disallowing this topic current tags' do
+            restricted_category.allowed_tags = [tag2.name]
+            restricted_category.reload
+
+            put "/t/#{topic.slug}/#{topic.id}.json", params: { category_id: restricted_category.id }
+
+            result = ::JSON.parse(response.body)
+
+            expect(response.status).to eq(422)
+            expect(result['errors']).to eq([I18n.t('category.errors.disallowed_topic_tags', tags: [tag1].map(&:name).join(", "))])
```
eviltrout: We probably don't need to be this specific in the test.

spec/requests/topics_controller_spec.rb
jjaffeux: Yes for some reason I thought it was needed, but no. Thanks.

app/controllers/topics_controller.rb
```diff
@@ -288,6 +288,18 @@ def update
       else
         return render_json_error(I18n.t('category.errors.not_found'))
       end
+
+      if category && topic_tags = params[:tags] || topic.tags.pluck(:name)
```
ZogStriP: I think we should add the paranthesis to ensure we get the priority right

```suggestion
      if category && topic_tags = (params[:tags] || topic.tags.pluck(:name))
```

app/controllers/topics_controller.rb
```diff
@@ -288,6 +288,18 @@ def update
...
+        allowed_tags = category.tags.pluck(:name)
+
+        if !(topic_tags.blank? || allowed_tags.blank?)
```
ZogStriP: How about using the opposite condition?

```suggestion
        if topic_tags.present? && allowed_tags.present?
```

PR #7066, opened 2019-02-25
UX: puts back share-panel as floating pane on post actions

jjaffeux:


jjaffeux:
@SamSaffron  I'm merging this tomorrow

PR #7070, opened 2019-02-26
FIX: rtl topic progress wrapper position

scossar:
Checks for the `rtl` class on the `html` element and sets the left/right inline CSS for the `#topic-prograss-wrapper` div conditionally.

Removes an unnecessary CSS rule from `rtl.scss`

Related meta topic: https://meta.discourse.org/t/problem-with-topic-control-on-arabic-forum-mobile-view/109911


SamSaffron:
@awesomerobot can you review and merge?

PR #7072, opened 2019-02-26
FEATURE: Introduce Ignore user

khalilovcmded:
## Why?

Stemming from this [feature request](https://meta.discourse.org/t/will-discourse-ever-gain-ignore-user-functionality/1525), we are beginning to look at ways to support ignoring a certain user.

The breakdown of why we want to ignore a user is:

1. I am a user, I want to ignore a specific user’s posts due to X reasons.
2. I can choose to ignore the other user’s posts:
    a. For a certain topic
    b. For a certain period of time
    c. For certain category
    d. Just in general – I don’t want to see any of that user’s posts anywhere

In this PR, we are aiming to target the implementation for (1).

## How?

First stab at supporting a simplified version of the ignore feature (1):

1. Introduce a new table `ignored_users` which has a one to many relationship to the User model. One User can have many `ignored_users`
2. Introduce an `ignore` and `watch` buttons on the User's profile and the User's pop-up dialog
3. Two new routes, `ignore` and `watch` on the [UserController](https://github.com/discourse/discourse/blob/6ea9f5c9c5976cabe7cb1e09a9e7f97b6697b529/app/controllers/users_controller.rb)

## Constraints

1. We haven't considered how we will model the schema of ignore feature modifiers (certain topic, certain time)
2. We haven't considered [User Merging](https://github.com/discourse/discourse/pull/5620) and how to merge two users `ignored_users` data, the backend and UI doesn't support that at all now (cc: @gschlager )
3. We now simply hide all posts for an ignored user, **EXCEPT** the first post by the ignored user in a certain Topic

## UI

ignore 2
ignore 1
ignore 3


app/assets/javascripts/discourse/controllers/user.js.es6
```diff
@@ -145,6 +145,20 @@ export default Ember.Controller.extend(CanCheckEmails, {
 
     adminDelete() {
       this.get("adminTools").deleteUser(this.get("model.id"));
+    },
+
+    ignoreUser() {
+      const user = this.get("model");
+      user.ignore().then(() => {
```
jjaffeux: This can be written as a one liner:

```
user.ignore().then(() => user.set("ignored", true));
```

app/assets/javascripts/discourse/controllers/user.js.es6
```diff
@@ -145,6 +145,20 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
+        user.set("ignored", true);
+      });
+    },
+
+    followUser() {
+      const user = this.get("model");
+      user.follow().then(() => {
```
jjaffeux: Can be a one liner too

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -48,6 +48,15 @@
             icon="envelope"
             label="user.private_message"}}
         </li>
+        {{#if user.can_ignore_user}}
+          <li>
+            {{d-button
+              class="btn btn-danger"
```
jjaffeux: I think "btn" class is added automatically here, no ?

app/controllers/users_controller.rb
```diff
@@ -995,6 +995,26 @@ def destroy
...
+  def ignore
+    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
+
+    ::IgnoredUser.find_or_create_by!(
+      user: current_user,
+      ignored_user: User.find_by(id: params[:ignored_user_id]))
+    render json: success_json
+  end
+
+  def follow
```
jjaffeux: Maybe this has been discussed before, but for me follow is not the opposite of ignore

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
     }
   },
 
+  ignore: function() {
```
jjaffeux:
```suggestion
  ignore() {
```

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get('id') }
+    });
+  },
+
+  follow: function() {
```
jjaffeux:
```suggestion
  follow() {
```

jjaffeux:
Few minor things, also I think this will need some tests server side and front side

app/controllers/users_controller.rb
khalilovcmded: @jjaffeux how about `watch`?

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get('id') }
+    });
+  },
+
+  watch: function() {
+    return ajax(`${userPath(this.get("username"))}/watch.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get('id') }
```
ZogStriP: I think it would be better to match the name of the variable with the endpoint

```suggestion
      data: { watched_user_id: this.get('id') }
```

app/controllers/users_controller.rb
```diff
@@ -995,6 +995,26 @@ def destroy
...
+      ignored_user: User.find_by(id: params[:ignored_user_id]))
+    render json: success_json
+  end
+
+  def watch
+    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
+
+    ignored_user = ::IgnoredUser.find_by(
+      user: current_user,
+      ignored_user: User.find_by(id: params[:ignored_user_id]))
```
ZogStriP: You don't need to do this query.

```ruby
IgnoredUser.where(user: current_user, ignored_user_id: params[:ignored_user_id]).delete_all
```

app/serializers/user_serializer.rb
```diff
@@ -274,6 +276,14 @@ def stats
     UserAction.stats(object.id, scope)
   end
 
+  def ignored
+    IgnoredUser.find_by(user_id: scope.user.try(:id), ignored_user_id: object.id).present?
```
ZogStriP: Does this generate an N+1?

db/migrate/20190225133654_add_ignored_users_table.rb
```diff
@@ -0,0 +1,12 @@
+class AddIgnoredUsersTable < ActiveRecord::Migration[5.2]
+  def change
+    create_table :ignored_users, force: true do |t|
```
ZogStriP: Why the `force: true`?

db/migrate/20190225133654_add_ignored_users_table.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.integer :user_id, null: false
```
ZogStriP: Do we also need to add a foreign key?

db/migrate/20190225133654_add_ignored_users_table.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.integer :ignored_user_id, null: false
```
ZogStriP: Ditto ^^ (FK)

lib/topic_view.rb
```diff
@@ -602,6 +602,14 @@ def setup_filtered_posts
     @contains_gaps = false
     @filtered_posts = unfiltered_posts
 
+    if SiteSetting.ignore_user_enabled
+      if IgnoredUser.where(user_id: @user.id).present?
+        @filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?",
+                                                    IgnoredUser.where(user_id: @user.id).map(&:ignored_user_id),
```
ZogStriP: Might want to do a `select` instead of a `map` to prevent loading all the `IgnoredUser` records in memory and just let SQL do it's work 💪

app/controllers/users_controller.rb
khalilovcmded: Totally 👍

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
 
+  ignore() {
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get('id') }
+    });
+  },
+
+  watch: function() {
+    return ajax(`${userPath(this.get("username"))}/watch.json`, {
```
ZogStriP: Instead of creating a different endpoint, I think it would be better to use the `DELETE` verb on `<username>/ignore`.

db/migrate/20190225133654_add_ignored_users_table.rb
khalilovcmded: We definitely don't need it here, it will be dropping the table first before creating it. 😃

db/migrate/20190225133654_add_ignored_users_table.rb
khalilovcmded: Yes, I wonder what's our stance in general on foreign keys. Different Rails codebases get to be opinionated about using foreign keys at all in their databases (they have their reasons).

Do we default on using them? Or only when it makes sense?

config/site_settings.yml
```diff
@@ -529,6 +529,10 @@ users:
     default: false
     client: true
   log_personal_messages_views: false
+  ignore_user_enabled:
+    hidden: true
+    default: true
```
khalilovcmded: Oh no, I didn't 🤒

app/serializers/user_serializer.rb
khalilovcmded: For a single User loaded, we will try to find ONE `user_ignored` db record that belongs to the current logged in User.

I think potentially N+1 if we are loading a list of Users and using this Serializer, but I see that we are using this Serializer for loading a single user, like in the `show` method of the UserController.

db/migrate/20190225133654_add_ignored_users_table.rb
ZogStriP: > Or only when it makes sense?

This ^^

gschlager:
> 2\. We haven't considered [User Merging](https://github.com/discourse/discourse/pull/5620) and how to merge two users `ignored_users` data, the backend and UI doesn't support that at all now

I think you can handle it similar to muted users. Ping me if you need any help.

https://github.com/discourse/discourse/blob/1d4e9b08ade008d2ba593213cf20ec57ea287ac8/app/services/user_merger.rb#L260-L261

https://github.com/discourse/discourse/blob/1d4e9b08ade008d2ba593213cf20ec57ea287ac8/app/services/user_merger.rb#L367

db/migrate/20190225133654_add_ignored_users_table.rb
khalilovcmded: Makes sense. IMO, I see in this case we might not need to add a foreign key. Looking at another table with similar usage and context `muted_users`, it was not need in that case either.

db/migrate/20190225133654_add_ignored_users_table.rb
khalilovcmded: Same as above.

khalilovcmded:
@gschlager Nice one!

I think since we are having this as disabled by default, and we are going to definitely revisit and add more functionality to it, perhaps it is better to not write more code for migration until we solidify the data models.

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
+  ignore() {
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get("id") }
+    });
+  },
+
+  watch: function() {
+    return ajax(`${userPath(this.get("username"))}/watch.json`, {
+      type: "DELETE",
```
ZogStriP: You changed the verb, but not the endpoint ;)

app/serializers/user_serializer.rb
ZogStriP:
```suggestion
    IgnoredUser.find_by(user_id: scope.user&.id, ignored_user_id: object.id).present?
```

config/site_settings.yml
```diff
@@ -529,6 +529,10 @@ users:
     default: false
     client: true
   log_personal_messages_views: false
+  ignore_user_enabled:
+    hidden: true
+    default: false
+    client: true
```
ZogStriP: Is the site setting used on the client-side?

app/assets/javascripts/discourse/models/user.js.es6
khalilovcmded: Yes 🤒

lib/topic_view.rb
```diff
@@ -602,6 +602,14 @@ def setup_filtered_posts
     @contains_gaps = false
     @filtered_posts = unfiltered_posts
 
+    if SiteSetting.ignore_user_enabled
+      if IgnoredUser.where(user_id: @user.id).present?
```
ZogStriP: Do we need to do this request?

plugins/discourse-narrative-bot/public/public
```diff
@@ -0,0 +1 @@
+/Users/tarek/Discourse/discourse/plugins/discourse-narrative-bot/public
```
ZogStriP: Did you mean to commit this file?

khalilovcmded: No, removing it 🔪

lib/topic_view.rb
khalilovcmded: Not strongly opinionated on it, but I guess it is useful since if there are no ignored users at all for the current signed in User, we avoid appending the below slightly more complicated `where` query to the `@filtered_posts`

config/site_settings.yml
khalilovcmded: No 👍

lib/topic_view.rb
ZogStriP: The fastest queries are the ones you don't make ;)

khalilovcmded: Makes sense 👍

spec/requests/users_controller_spec.rb
```diff
@@ -2020,6 +2020,76 @@ def post_user
     end
   end
 
+  describe '#ignore' do
+    it 'raises an error when not logged in' do
+      put "/u/ignore.json"
```
ZogStriP: Is that the right endpoint?

spec/requests/users_controller_spec.rb
```diff
@@ -2020,6 +2020,76 @@ def post_user
...
+          expect(IgnoredUser.find_by(user_id: user.id,
+                                     ignored_user_id: another_user.id)).to be_present
+        end
+      end
+    end
+  end
+
+  describe '#watch' do
+    it 'raises an error when not logged in' do
+      put "/u/ignore.json"
```
ZogStriP: Shouldn't this be a delete?

app/assets/javascripts/discourse/templates/user.hbs
```diff
@@ -43,12 +43,27 @@
...
             <ul>
               {{#if model.can_send_private_message_to_user}}
-              <li>
-                {{d-button class="btn btn-primary compose-pm"
-                  action=(route-action "composePrivateMessage" model)
-                  icon="envelope"
-                  label="user.private_message"}}
-              </li>
+                <li>
+                  {{d-button class="btn btn-primary compose-pm"
```
jjaffeux:
```suggestion
                  {{d-button class="btn-primary compose-pm"
```

spec/requests/users_controller_spec.rb
khalilovcmded: Ah this is bad. I totally forgot to add these two controller methods behind `requires_login` 👀 
Specs to the rescue!

spec/requests/users_controller_spec.rb
khalilovcmded: Yes, same reply as above 👍

app/serializers/user_serializer.rb
```diff
@@ -274,6 +276,14 @@ def stats
     UserAction.stats(object.id, scope)
   end
 
+  def ignored
+    IgnoredUser.find_by(user_id: scope.user&.id, ignored_user_id: object.id).present?
```
ZogStriP: This is slightly faster and requires less memory since it doesn't load the whole object in memory ;)

```suggestion
    IgnoredUser.where(user_id: scope.user&.id, ignored_user_id: object.id).exists?
```

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get("id") }
+    });
+  },
+
+  watch: function() {
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "DELETE",
+      data: { watched_user_id: this.get("id") }
```
ZogStriP: I guess we can change this name back to the original now the endpoint is `ignore`

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -611,6 +611,20 @@ const User = RestModel.extend({
...
   },
 
+  ignore() {
+    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
+      type: "PUT",
+      data: { ignored_user_id: this.get("id") }
+    });
+  },
+
+  watch: function() {
```
ZogStriP:
```suggestion
  watch() {
```

PR #7077, opened 2019-02-26
FIX: staff/admin shouldn’t be able to create uncategorized topics

jjaffeux:


jjaffeux:
@ZogStriP for context

This feature has been through a lot of changes over the years:

- no check
- staff allowed to bypass uncategorized on front and back
- admin allowed on front, staff allowed on back

And per https://meta.discourse.org/t/default-category/34481/27?u=joffreyjaffeux (and previous post) I got requested to deny any topic without a category if setting is disabled.

PR #7081, opened 2019-02-27
REFACTOR: Migrate InstagramAuthenticator to use ManagedAuthenticator

jjaffeux:


jjaffeux:
@davidtaylorhq  Will  give it another look tomorrow, but I think this is already working well.

spec/components/auth/instagram_authenticator_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+describe Auth::InstagramAuthenticator do
+
+  it "takes over account if email is supplied" do
+    auth = Auth::InstagramAuthenticator.new
+
+    user = Fabricate(:user)
```
khalilovcmded: Can we move this to before to a `let(:user)` above the `it ..`, I can see we are using another `fabricate(:user)` [below](https://github.com/discourse/discourse/pull/7081/files#diff-431c322d77be415c017de4b9f5567c9bR43).

khalilovcmded:
It will be great to have a description here for this piece of work, just in case someone would like to get a little context of where this all fits, or was looking at this PR to gain some historical context.

spec/components/auth/instagram_authenticator_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(result.user.id).to eq(user.id)
+
+    info = UserAssociatedAccount.find_by(provider_name: "instagram", user_id: user.id)
+    expect(info.info["email"]).to eq(user.email)
+  end
+
+  it 'can connect to a different existing user account' do
+    authenticator = Auth::InstagramAuthenticator.new
+    user1 = Fabricate(:user)
+    user2 = Fabricate(:user)
```
khalilovcmded: Same as above, I think we might use want to `fabricate` for `user1` once at the top of this spec.

spec/components/auth/instagram_authenticator_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    it 'raises exception if no entry for user' do
+      expect { authenticator.revoke(user) }.to raise_error(Discourse::NotFound)
+    end
+
+    it 'revokes correctly' do
+      UserAssociatedAccount.create!(provider_name: "instagram", user_id: user.id, provider_uid: 100)
+      expect(authenticator.can_revoke?).to eq(true)
+      expect(authenticator.revoke(user)).to eq(true)
+      expect(authenticator.description_for_user(user)).to eq("")
```
khalilovcmded: Should this `be_empty` instead of `eq("")`?

spec/components/auth/instagram_authenticator_spec.rb
jjaffeux: I'm kind of reluctant to do this, I followed the same patterns we have in related tests files for other authenticators

spec/components/auth/instagram_authenticator_spec.rb
jjaffeux: same comment it's same pattern we have in all other specs of this kind

davidtaylorhq:
@khalilovcmded the context here is https://meta.discourse.org/t/future-social-authentication-improvements/94691/3

In terms of tests, I think there is very little point testing every different provider to such a large extent - we end up just testing ManagedAuthenticator repeatedly. We're not even testing the omniauth parts of the authenticators, we trust they are tested upstream. However, I agree @jjaffeux that we can tidy this up later in one go. I'll add it to the todo list on Meta

PR #7083, opened 2019-02-28
FIX: only profile owner can see links to watched/tracked/muted topics

majakomel:
https://meta.discourse.org/t/see-and-set-user-notification-level-for-topics-and-messages/108838/4


app/assets/javascripts/discourse/controllers/preferences/categories.js.es6
```diff
@@ -20,12 +20,18 @@ export default Ember.Controller.extend(PreferencesTabController, {
...
   },
 
-  canSave: function() {
+  @computed
+  canSee() {
+    return this.get("currentUser.id") === this.get("model.id");
+  },
+
+  @computed
+  canSave() {
```
ZogStriP:
```suggestion
  canSave: Ember.computed.or("canSee", "currentUser.admin")
```

PR #7084, opened 2019-02-28
FEATURE: Use smooth scrolling for J/K keyboard shortcuts.

udan11:
Moves ["jk smooth mooves" theme component](https://github.com/udan11/discourse-jk-smooth-mooves) into core.

### DIFF
--- a/app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6
+++ b/app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6
@@ -417,53 +417,112 @@ export default {
   },
 
   _moveSelection(direction) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/lib/keyboard-shortcuts.js.es6
```diff
@@ -417,53 +417,112 @@ export default {
...
+        .find(article => article.getBoundingClientRect().top > offset);
       direction = 0;
     }
 
-    const $article = $articles.eq(index + direction);
+    const index = $articles.index($selected);
+    let $article = $articles.eq(index);
+
+    /*
+     * Try doing a page scroll in the context of current post.
```
eviltrout: Trivial but a single line comment should use `//` rather than three lines.

PR #7085, opened 2019-02-28
FIX: when posts are moved to a message then small action post should not be publicly visible

techAPJ:
This fix is inspired from what we do in discourse-assigned plugin.

https://github.com/discourse/discourse-assign/blob/master/lib/topic_assigner.rb#L184
https://github.com/discourse/discourse-assign/blob/master/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6#L125-L133


techAPJ:
Thanks for the review, @ZogStriP ❤️

PR #7086, opened 2019-02-28
FIX: prevents exception when mod creates category

jjaffeux:


jjaffeux:
@ZogStriP I'm unsure what we want to do here, this is keeping exact same behavior.

Should we have "everyone" group permission show for mods ? This fix will show the category form, but permissions will be empty,  and they wll only have admins, staff, moderators, tl{1,2,3,4}

SamSaffron:
Are you sure about this, I suppressed this on purpose in the past. what is the exact repro now?

ZogStriP:
Moderators are getting a JS error when trying to create a new category because the code is trying to find the "everyone" group which is now only visible to admins.

jjaffeux:
My first commit was not touching this, and only preventing the js exception. What it meant was that a moderator creating a category would not have "everyone" in permissions, and would have to add:

TL{1,2,3,4}, staff... permissions when creating a category.

ZogStriP:
I wonder if we should use `DB.build` here instead of a bunch of `+=`?

SamSaffron:
Fixed per: 

https://github.com/discourse/discourse/commit/8d7c10f7f2a13020ccd6af649509591d4aff9643 

Sorry it took me so long here to address this.

PR #7089, opened 2019-03-01
FIX: anonymous user filtering bug in PostView

khalilovcmded:
## Why?

1. When we are attempting to filter posts for ignored users, we assumed that there is always a current User with an ID. Sometimes, there will be anonymous users which don't have an ID. By anonymous user here, I mean, a non-signed in User who is viewing a certain Topic.

2. I am also adding support for `X hidden replies` in the UI.

## UI

After I ignored the example user `Matt`, all of Matt's posts are now hidden, and UI gaps with the description: `X hidden reply` are inserted:

how_to_hike_mount_fuji__-_japan_mountains_ _-_trekup


spec/components/topic_view_spec.rb
```diff
@@ -31,6 +31,50 @@
     expect { TopicView.new(topic.id, admin) }.not_to raise_error
   end
 
+  context "setup_filtered_posts" do
```
khalilovcmded: Tried to cover this with more specs.

lib/topic_view.rb
```diff
@@ -604,8 +604,9 @@ def setup_filtered_posts
 
     if SiteSetting.ignore_user_enabled
       @filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?",
-                                                  IgnoredUser.where(user_id: @user.id).select(:ignored_user_id),
+                                                  IgnoredUser.where(user_id: @user&.id).select(:ignored_user_id),
                                                   first_post_id)
+      @contains_gaps = true
```
khalilovcmded: This one liner basically enables us to invoke the [Gaps](https://github.com/discourse/discourse/blob/03a70ef69c415d561dedcf3c5d453ce8944cd117/lib/topic_view.rb#L110) calculator, which basically will take a full set of posts for that topic and a subset of filtered post, and then it calculates where the gaps should be.

spec/components/topic_view_spec.rb
```diff
@@ -31,6 +31,61 @@
...
+      let!(:ignored_user) { Fabricate(:ignored_user, user: evil_trout, ignored_user: user) }
+      let!(:post) { Fabricate(:post, topic: topic, user: first_poster) }
+      let!(:post2) { Fabricate(:post, topic: topic, user: evil_trout) }
+      let!(:post3) { Fabricate(:post, topic: topic, user: user) }
+
+      describe "when SiteSetting.ignore_user_enabled is false" do
+        it "does not filter out ignored user posts" do
+          SiteSetting.ignore_user_enabled = false
+
+          tv = TopicView.new(topic.id, evil_trout, print: true)
```
eviltrout: Is `print: true` necessary here?

lib/topic_view.rb
```diff
@@ -604,8 +604,9 @@ def setup_filtered_posts
 
     if SiteSetting.ignore_user_enabled
       @filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?",
-                                                  IgnoredUser.where(user_id: @user.id).select(:ignored_user_id),
+                                                  IgnoredUser.where(user_id: @user&.id).select(:ignored_user_id),
```
eviltrout: A few thoughts here:

1. `pluck` is much more efficient than `select` if you want an array of values returned. `select` will create object instances.

2. If the ignored user query returns nothing, we don't need to execute the `where` query. Most users will not have any ignored users so you can often save a query by first retrieving the ignored user ids into a variable, and then only executing the filter query if that array is not empty.

3. At some point we should benchmark the cost of querying the ignored users versus per topic view storing them in a cache and see if we can win some performance that way.

eviltrout: (also sorry I know none of those thoughts have to do with this FIX, but I just noticed them!)

spec/components/topic_view_spec.rb
khalilovcmded: No, I don't think so ☺️

lib/topic_view.rb
khalilovcmded: > 1. `pluck` is much more efficient than `select`, ....

Makes sense 👍 

> 2. If the ignored user query returns nothing, .....

That makes sense, it is quick to change but since this is experimental and I aim to do some changes in this path again, would it be okay to revisit later?

> 3.  At some point we should benchmark

👍

lib/topic_view.rb
```diff
@@ -603,9 +603,12 @@ def setup_filtered_posts
     @filtered_posts = unfiltered_posts
 
     if SiteSetting.ignore_user_enabled
-      @filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?",
-                                                  IgnoredUser.where(user_id: @user.id).select(:ignored_user_id),
-                                                  first_post_id)
+      ignored_user_ids = IgnoredUser.where(user_id: @user&.id).pluck(:ignored_user_id)
```
khalilovcmded: @eviltrout is that what you were thinking of?

lib/topic_view.rb
ZogStriP: > * `pluck` is much more efficient than `select` if you want an array of values returned. `select` will create object instances.

Except in this case, the `select` won't return an array. It'll be directly transformed into SQL 😉 

At some point we should benchmark the differences between

- Doing a small request with a `pluck` and then adding the `where` if there are ignored users

``` ruby
ignored_user_ids = IgnoredUser.where(user_id: @user&.id).pluck(:ignored_user_id)

if ignored_user_ids.present?
  @filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?", ignored_user_ids, first_post_id)
end
```

- Using `select` and always adding the `where` clause

``` ruby
ignored_users = IgnoredUser.where(user_id: @user&.id).select(:ignored_user_id)
@filtered_posts = @filtered_posts.where.not("user_id IN (?) AND id <> ?", ignored_users, first_post_id)
```

@SamSaffron any thoughts about this?

ZogStriP: Although, my point is kinda moot if we need to know if there are ignored users to set the `@contains_gap` to `true`.

khalilovcmded: > Except in this case, the select won't return an array. It'll be directly transformed into SQL 😉

`select` will result in an active relation, but `pluck` will directly return an array of IDs.

ZogStriP: Yes, but that's way less memory that loading the `IgnoredUser` records in memory ;) (which happens if you iterate over the results of the `select`)

khalilovcmded: I am happy to go with either approaches. I am unsure, in this PR's context (besides future bench marking) is there any changes that we need to do?

ZogStriP: I think it's fine as is for now. We can revisit later.

PR #7090, opened 2019-03-01
FEATURE: displays groups in menu search

jjaffeux:


jjaffeux:
screenshot 2019-03-01 at 11 49 16

app/assets/javascripts/discourse/lib/search.js.es6
```diff
@@ -43,6 +44,17 @@ export function translateResults(results, opts) {
     })
     .compact();
 
+  results.groups = results.groups
+    .map(group => {
+      const groupName = Handlebars.Utils.escapeExpression(group.name);
+      return Ember.Object.create({
```
eviltrout: I'm curious if we're observing these attributes. If not, it's often faster to return a POJO.

lib/search.rb
```diff
@@ -683,6 +684,16 @@ def user_search
     end
   end
 
+  def groups_search
+    groups = Group
+      .visible_groups(@guardian.user, "name ASC", include_everyone: false)
+      .where("groups.name ILIKE ?", "%#{@term}%")
+
+    groups.each do |group|
```
eviltrout: Trivial but I'd prefer a one liner for this kind of simple operation.

PR #7091, opened 2019-03-01
FEATURE: Optional detailed performance logging for Sidekiq jobs

davidtaylorhq:
By default, this does nothing. Two environment variables are available:

- `DISCOURSE_LOG_SIDEKIQ`

  Set to `"1"` to enable logging. This will log all completed jobs to `log/rails/sidekiq.log`, along with various db/redis/network statistics. This is useful to track down poorly performing jobs.

- `DISCOURSE_LOG_SIDEKIQ_INTERVAL`

  (seconds) Check running jobs periodically, and log their current duration. They will appear in the logs with `status:pending`. This is useful to track down jobs which take a long time, then crash sidekiq before completing.

### DIFF
--- a/app/jobs/base.rb
+++ b/app/jobs/base.rb
@@ -16,34 +16,86 @@ module Jobs
   end
 
   class Base
+    class JobInstrumenter
+      def initialize(job_class:, opts:, db:)
+        return unless enabled?
+        @data = {}
 
-    class Instrumenter
+        @data["hostname"] = `hostname`.strip # Hostname
+        @data["pid"] = Process.pid # Pid
+        @data["database"] = db # DB name - multisite db name it ran on
+        @data["job_name"] = job_class.name # Job Name - eg: Jobs::AboutStats
+        @data["job_type"] = job_class.try(:scheduled?) ? "scheduled" : "regular" # Job Type - either s for scheduled or r for regular
+        @data["opts"] = opts # Params - json encoded params for the job
 
-      def self.stats
-        Thread.current[:db_stats] ||= Stats.new
+        @data["status"] = 'pending'
+        @start_timestamp = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+        self.class.ensure_interval_logging!
+        @@active_jobs ||= []
+        @@active_jobs << self
+
+        MethodProfiler.ensure_discourse_instrumentation!
+        MethodProfiler.start
       end
 
-      class Stats
-        attr_accessor :query_count, :duration_ms
+      def stop(exception:)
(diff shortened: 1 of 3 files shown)


app/jobs/base.rb
```diff
@@ -16,35 +16,87 @@ def self.num_email_retry_jobs
...
+
+        write_to_log
       end
 
-      def call(name, start, finish, message_id, values)
-        stats = Instrumenter.stats
-        stats.query_count += 1
-        stats.duration_ms += (((finish - start).to_f) * 1000).to_i
+      def logger
+        @@logger ||= Logger.new("#{Rails.root}/log/sidekiq.log")
```
SamSaffron: I think we want an async logger here so there is no chance logging blocks sidekiq, I think active support has one of those.

lib/method_profiler.rb
```diff
@@ -61,4 +61,26 @@ def self.stop
     end
     data
   end
+
+  def self.ensure_discourse_instrumentation!
```
SamSaffron: nice refactor!

app/jobs/base.rb
```diff
@@ -16,35 +16,87 @@ def self.num_email_retry_jobs
...
+        ENV["DISCOURSE_LOG_SIDEKIQ"] == "1"
+      end
+
+      def self.ensure_interval_logging!
+        interval = ENV["DISCOURSE_LOG_SIDEKIQ_INTERVAL"]
+        return if !interval
+        @@interval_thread ||= Thread.new do
+          loop do
+            sleep interval.to_i
+            @@active_jobs.each { |j| j.write_to_log if j.current_duration > interval }
```
SamSaffron: got to output something somewhere if this thread crashes.

SamSaffron:
overall looks good to merge pending the 2 comment. Feel free to merge in once you address.

app/jobs/base.rb
davidtaylorhq: I don't think ActiveSupport has anything. It [used to have](https://github.com/rails/rails/commit/572c3d517899524c2a7c4c84ad9646660168d4cd#diff-03c8fca389d378ba1a40492d96b5fdc5) `BufferedLogger`, but that was removed a while ago.

I've implemented a queue, with logging in a separate thread (3b84d98), but I'm not sure if the performance benefit is worth the extra code, especially bearing in mind that this entire thing is opt-in? Let me know what you think.

app/jobs/base.rb
davidtaylorhq: Done in f9cf37b

PR #7096, opened 2019-03-03
FIX: Return actual errors if PostCreator fails

maulkin:
https://meta.discourse.org/t/insufficient-checking-of-postcreator-in-importer-base/110570
Extract:
It seems that if (for example) you mistakenly try and add
target_usernames to a post with user_ids rather than usernames,
create_post in scripts/import_scripts/base.rb won’t return an error.

This is because it returns: post ? post :
post_creator.errors.full_messages. However, post has already been
initialised as a Post type on line 564. Checking for the post_creator
errors instead returns the correct behaviour.

### DIFF
--- a/script/import_scripts/base.rb
+++ b/script/import_scripts/base.rb
@@ -563,7 +563,7 @@ class ImportScripts::Base
     post_creator = PostCreator.new(user, opts)
     post = post_creator.create
     post_create_action.try(:call, post) if post
-    post_creator.errors.full_messages.empty? ? post : post_creator.errors.full_messages
+    if !post || post_creator.errors&.length > 0 ? post : post_creator.errors.full_messages
   end
 
   def create_upload(user_id, path, source_filename)
--- a/script/import_scripts/base.rb
+++ b/script/import_scripts/base.rb
@@ -563,7 +563,7 @@ class ImportScripts::Base
     post_creator = PostCreator.new(user, opts)
     post = post_creator.create
     post_create_action.try(:call, post) if post
-    post ? post : post_creator.errors.full_messages
+    post_creator.errors.full_messages.empty? ? post : post_creator.errors.full_messages
   end
 
   def create_upload(user_id, path, source_filename)


script/import_scripts/base.rb
```diff
@@ -563,7 +563,7 @@ def create_post(opts, import_id)
     post_creator = PostCreator.new(user, opts)
     post = post_creator.create
     post_create_action.try(:call, post) if post
-    post ? post : post_creator.errors.full_messages
+    post_creator.errors.full_messages.empty? ? post : post_creator.errors.full_messages
```
SamSaffron: can `post_creator.errors` be `nil` ? I am pretty sure it can be. Instead:

```
if !post || post_creator.errors&.length > 0 ? post : post_creator.errors.full_messages
```

maulkin: I don't think `post_creator.errors` can be `nil`, but it certainly could be `blank`:
`PostCreator` seems to include `HasErrors` which initialises:
`@errors ||= ActiveModel::Errors.new(self)`

Within the `valid?` function, it then checks:
`valid = @post.errors.blank?`

However, your suggestion would also catch this. I've had a look to see if `full_messages` **must** be present if `errors` exist, and I think it does.
```suggestion
    if !post || post_creator.errors&.length > 0 ? post : post_creator.errors.full_messages
```

maulkin:
Sure - Did you want it adding as a test case specifically, or should I just add some code here to show what happens/should happen?

eviltrout:
@maulkin I would add a test that fails if your patch is not present.

gschlager:
@eviltrout We don't really have tests for import scripts and the few that we have are not executed by either Jenkins nor Travis.

eviltrout:
Ah my bad @gschlager I thought the change was in PostCreator which absolutely needs tests. If not, no big deal.

gschlager:
Thanks! 👍

gschlager:
For the record: Your first solution was the correct one. 😉 
I fixed it in https://github.com/discourse/discourse/commit/941e096df4b88943079f751082c3708da5da449b

maulkin:
Wonderful, thanks folks!

PR #7097, opened 2019-03-03
FIX: any subcategory permissions are valid if everyone has access to parent category

majakomel:
https://meta.discourse.org/t/parent-category-with-more-restrictions-than-subcategory/20913/10


SamSaffron:
Good fix! thanks!

PR #7098, opened 2019-03-03
UX: focus on search box when emoji picker is opened

majakomel:
https://meta.discourse.org/t/opening-emoji-popover-should-focus-on-emoji-search-box/110309

### DIFF
--- a/app/assets/javascripts/discourse/components/emoji-picker.js.es6
+++ b/app/assets/javascripts/discourse/components/emoji-picker.js.es6
@@ -7,6 +7,7 @@ import {
   isSkinTonableEmoji,
   emojiSearch
 } from "pretty-text/emoji";
+import { safariHacksDisabled } from "discourse/lib/utilities";
 const { run } = Ember;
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/components/emoji-picker.js.es6
```diff
@@ -58,6 +58,9 @@ export default Ember.Component.extend({
       this._scrollTo();
       this._updateSelectedDiversity();
       this._checkVisibleSection(true);
+
+      if (!this.site.isMobileDevice || this.get("isEditorFocused"))
+        this.$filter.find("input[name='filter']").focus();
```
SamSaffron: I think we also need to test for !safariHacksDisabled() in utils, otherwise stuff can get bouncy on iPad

ZogStriP:
@majakomel, Danger is right, can you change it so we don't use `self`?

ZogStriP:
Ok, let's try this 👍

PR #7100, opened 2019-03-04
Ruby version update

sau226:
Commit 8745f78277f80250032352343bde59f7bba12bec stated that Ruby 2.6.1 and above should be used for development operations.

Per that commit bump all references to Ruby 2.5 to Ruby version 2.6.


ZogStriP:
Thanks 👍

PR #7103, opened 2019-03-04
DEV: Add `pry-remote` gem

khalilovcmded:
## Why?

I find it easy to use `pry-remote` when I want to debug and not wanting to depend on `binding.pry` which gets overridden most of the times by the log lines in the terminal. I am not sure if there is a better way but I find this gem handy.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -137,6 +137,7 @@ group :test, :development do
   gem 'shoulda', require: false
   gem 'rspec-html-matchers'
   gem 'pry-nav'
+  gem 'pry-remote'
   gem 'byebug', require: ENV['RM_INFO'].nil?
   gem 'rubocop', require: false
 end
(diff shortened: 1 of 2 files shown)


eviltrout:
Have you tried using byebug which we have installed already? I wonder if we should remove it although @tgxworld uses it I believe.

khalilovcmded:
I did try to look at `byebug` before, maybe it is a personal preference, but it didn't feel at all as natural as `pry` or `pry-remote`.

SamSaffron:
Hmmm ...  I never use pry remote or bye bug ... I usually just do `puts` debugging as I need it. 

I think though we should as a team pick a winner here instead of having to maintain 2, maybe do a dev topic and we can make a call next week.

ZogStriP:
I'm also a [puts debuggerer](https://tenderlovemaking.com/2016/02/05/i-am-a-puts-debuggerer.html) and I agree with @SamSaffron, we should pick one gem.

tgxworld:
Looks like @khalilovcmded might not need this afterall. Closing this :)

PR #7106, opened 2019-03-04
FIX: Fixed editing whispers bumps topic

venarius:
Editing a whisper not longer changes the topic's bump date.

Resolves: https://meta.discourse.org/t/editing-whisper-bumps-topic/108890

This is also my first PR for discourse. Feedback is greatly appreciated!

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -236,9 +236,6 @@ class PostsController < ApplicationController
       opts[:skip_validations] = true
     end
 
-    # Bypass bump for whisper edits
-    opts[:bypass_bump] = true if post.whisper?
-
     topic = post.topic
     topic = Topic.with_deleted.find(post.topic_id) if guardian.is_staff?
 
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -236,6 +236,9 @@ def update
       opts[:skip_validations] = true
     end
 
+    # Bypass bump for whisper edits
+    opts[:bypass_bump] = true if post.whisper?
```
eviltrout: Rather than passing it through from the controller, I think it would make more sense if the `PostRevisor` had this logic built into its `bump_topic` method. `PostRevistor` is used outside of controllers and we want this behaviour to run everywhere.

venarius: That was also where I had put it in the first place but since `bypass_bump?` already makes a decision based on the `@opts[:bypass_bump]` it receives, I thought it would be a better way in just setting this variable to true instead of adding more logic to it.

ZogStriP:
Thanks 👍

PR #7107, opened 2019-03-04
FEATURE: Add ignored user list to the User's preference page

khalilovcmded:
## Why?

Part of: https://meta.discourse.org/t/ability-to-ignore-a-user/110254

We want to add list of Ignored users under or along with the muted users preferences section.
This way Users can find and update their list of ignored users.

## UI

gif

## Open questions

Two of many options to represent a list of ignored users is that we can:

1. We can represent the ignored user list as a table with the ability to `un-ignore` but NOT to add new ignored users.
2. We can keep it functioning as the `muted user list` where you can `un-ignore` or `ignore` users.


app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -0,0 +1,18 @@
...
+
+export default Ember.Controller.extend(PreferencesTabController, {
+  saveAttrNames: ["muted_usernames", "ignored_usernames"],
+
+  actions: {
+    save() {
+      this.set("saved", false);
+      return this.get("model")
+        .save(this.get("saveAttrNames"))
+        .then(() => {
```
jjaffeux: you can online this:

`.then(() => this.set("saved", true))`

app/assets/javascripts/discourse/templates/preferences/notifications.hbs
```diff
@@ -39,14 +39,16 @@
...
-    <label>{{d-icon "d-muted" class="muted icon"}} {{i18n 'user.muted_users'}}</label>
-    {{user-selector excludeCurrentUser=true usernames=model.muted_usernames class="user-selector"}}
+{{#unless siteSettings.ignore_user_enabled}}
+  <div class="control-group muting">
+    <label class="control-label">{{i18n 'user.users'}}</label>
+    <div class="controls category-controls">
+      <label>{{d-icon "d-muted" class="muted icon"}} {{i18n 'user.muted_users'}}</label>
+      {{user-selector excludeCurrentUser=true usernames=model.muted_usernames class="user-selector"}}
+    </div>
+    <div class="instructions">{{i18n 'user.muted_users_instructions'}}</div>
```
jjaffeux: where is the `</div>` ?

I think you can improve html formatting.

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
khalilovcmded: Cool, I think we need to change it in a couple other places under `preferences/*` too.

app/assets/javascripts/discourse/templates/preferences/notifications.hbs
khalilovcmded: Which `</div>`?

I can see the it at [the end of the line](https://github.com/discourse/discourse/pull/7107/files#diff-203c4b2a392348e0564b713df30f9d35R49).

`....{{i18n 'user.muted_users_instructions'}}   </div>`

Also, this is really unchanged since I just wrapped into a `{{#unless siteSettings.ignore_user_enabled}}`

app/assets/javascripts/discourse/templates/preferences/notifications.hbs
```diff
@@ -39,14 +39,16 @@
   </div>
 {{/if}}
 
-<div class="control-group muting">
-  <label class="control-label">{{i18n 'user.users'}}</label>
-  <div class="controls category-controls">
-    <label>{{d-icon "d-muted" class="muted icon"}} {{i18n 'user.muted_users'}}</label>
-    {{user-selector excludeCurrentUser=true usernames=model.muted_usernames class="user-selector"}}
+{{#unless siteSettings.ignore_user_enabled}}
```
jjaffeux: I would just always move this to users all the time

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -0,0 +1,29 @@
+{{#if siteSettings.ignore_user_enabled}}
```
jjaffeux: and then only do this check for the ignored part

khalilovcmded: Makes sense. I was hesitated to make the UI change (it might seem to be dramatic) but I guess it is okay 👍

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -0,0 +1,29 @@
...
+  <div class="control-group muting">
+    <label class="control-label">{{i18n 'user.users'}}</label>
+    <div class="controls category-controls">
+      <label>{{d-icon "d-muted" class="muted icon"}} {{i18n 'user.muted_users'}}</label>
```
jjaffeux: do we "muted icon" in css somewhere ?

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -0,0 +1,29 @@
...
+      {{user-selector excludeCurrentUser=true usernames=model.muted_usernames class="user-selector"}}
+    </div>
+    <div class="instructions">{{i18n 'user.muted_users_instructions'}}</div>
+
+    <div class="controls category-controls">
+      <label>{{d-icon "eye-slash" class="muted icon"}} {{i18n 'user.ignored_users'}}</label>
```
jjaffeux: do we "muted icon" in css somewhere ?

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
jjaffeux: It's ok if it's not in your PR, and it's not something very important, but if we can avoid adding more that's cool :p

ZogStriP:
😍 that you always add screenshots/gifs when you change UI elements ❤️

featheredtoast:
This page has the potential to get really, really long on some instances.

What's not clear in the screencaptures (thanks for those btw!) is how it handles multiline entries, and long lists of users.

ZogStriP:
Multiline entries is already supported (try it on categories or tags) but long list of users might require another UI.

khalilovcmded:
> and long lists of users.

I find it hard to imagine that one will ever a _really_ long list of ignored users that will be hard to represent. In my head, this falls back to the question of: `Why do I need to ignore so many users?`.

app/assets/javascripts/discourse/templates/preferences/users.hbs
khalilovcmded: Good catch. We do use it in the top muted users `d-icon`, I don't see `muted` class defined or used anywhere else. I will remove it 👍

app/assets/javascripts/discourse/templates/preferences/users.hbs
khalilovcmded: Good catch. We do use it in the top muted users d-icon, I don't see muted class defined or used anywhere else. I will remove it 👍

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
khalilovcmded: Let's avoid it ☺️

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -0,0 +1,29 @@
...
+      {{user-selector excludeCurrentUser=true
+                      usernames=model.ignored_usernames class="user-selector"}}
+    </div>
+    <div class="instructions">{{i18n 'user.ignored_users_instructions'}}</div>
+  {{/if}}
+</div>
+
+{{plugin-outlet name="user-preferences-notifications" args=(hash model=model save=(action "save"))}}
+
+<br/>
```
jjaffeux: Not a big fan of <br /> to position elements. Please use css if possible. Less markup, easier to maintain, easier to customise.

jjaffeux:
Other than this very minor nitpick, LGTM 👍

app/assets/javascripts/discourse/templates/preferences/users.hbs
khalilovcmded: Totally 👍 I'd rather not touch the `plugin-outlet` for now.

PR #7108, opened 2019-03-04
FEATURE: Skip small actions when counting replies in PMs.

udan11:


app/models/topic.rb
```diff
@@ -527,7 +527,7 @@ def update_status(status, enabled, user, opts = {})
   end
 
   # Atomically creates the next post number
-  def self.next_post_number(topic_id, reply = false, whisper = false)
+  def self.next_post_number(topic_id, reply = false, whisper = false, post = true)
```
ZogStriP: 3 is a crowd. Can you change these optional parameters to named ones?

spec/components/post_creator_spec.rb
```diff
@@ -776,6 +776,22 @@
 
       expect(post.topic.topic_allowed_users.where(user_id: admin2.id).count).to eq(0)
     end
+
+    it 'does not increase posts count for small actions' do
```
ZogStriP: Can you also test for the number of small actions in that PM? This would make sure we are actually skipping them.

udan11: Did that and also added a few more posts to ensure they are counted.

PR #7109, opened 2019-03-04
Improve Uncategorized category

gschlager:
* Adds warnings to the "Edit Category" dialog
* Doesn't hide the "Security" tab on the "Edit Category" dialog anymore. Instead, it shows an explanation why permissions can't be changed.
* Makes the category name translatable
* Hides the category name from the edit dialog (it can be customized by overriding the translation)
* Creates a translation override if the category has been renamed in the past
* Doesn't concatenate the title on the "Edit Category" dialog anymore. That's bad practice.

image

image


app/assets/javascripts/discourse/templates/components/edit-category-security.hbs
```diff
@@ -1,19 +1,25 @@
...
-  </ul>
+  {{#unless category.isUncategorizedCategory}}
+    <ul class='permission-list'>
+      {{#each category.permissions as |p|}}
+        <li>
+          <span class="name"><span class="badge-group">{{p.group_name}}</span></span>
+          {{{i18n "category.can"}}}
+          <span class="permission">{{p.permission.description}}</span>
+          {{#if editingPermissions}}
+          <a class="remove-permission" href {{action "removePermission" p}}>{{d-icon "times-circle"}}</a>
```
eviltrout: Trivial but I would indent this. (I wonder if we can ever lint templates?)

PR #7111, opened 2019-03-05
FEATURE: defer flags when deleting child replies

techAPJ:
When deleting multiple posts in a topic only auto-agree with the flag on the actual post you pressed delete on and ignore all other flags in child deletes.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -336,7 +336,7 @@ class PostsController < ApplicationController
 
   def destroy_many
     params.require(:post_ids)
-    defer_flags = params[:defer_flags] || false
+    agree_with_first_reply_flag = (params[:agree_with_first_reply_flag] || true).to_s == "true"
 
     posts = Post.where(id: post_ids_including_replies)
     raise Discourse::InvalidParameters.new(:post_ids) if posts.blank?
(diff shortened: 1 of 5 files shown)


app/controllers/posts_controller.rb
```diff
@@ -345,7 +345,9 @@ def destroy_many
     posts.each { |p| guardian.ensure_can_delete!(p) }
 
     Post.transaction do
-      posts.each { |p| PostDestroyer.new(current_user, p, defer_flags: defer_flags).destroy }
+      posts.each_with_index do |p, i|
+        PostDestroyer.new(current_user, p, defer_flags: (agree_with_first_reply_flag.to_s == "true" && i == 0) ? "false" : "true").destroy
```
eviltrout: It's a bit odd that we supply a string based boolean here to `PostDestroyer`.

I'm ok with the parameter to the controller being "true" but `PostDestoyer` should take `true`. Would you consider making that change? I bet it would not be significant.

app/controllers/posts_controller.rb
```diff
@@ -345,7 +345,9 @@ def destroy_many
     posts.each { |p| guardian.ensure_can_delete!(p) }
 
     Post.transaction do
-      posts.each { |p| PostDestroyer.new(current_user, p, defer_flags: defer_flags).destroy }
+      posts.each_with_index do |p, i|
+        PostDestroyer.new(current_user, p, defer_flags: (agree_with_first_reply_flag.to_s == "true" && i == 0) ? false : true).destroy
```
ZogStriP: `agree_with_first_reply_flag.to_s == "true"` should be moved outside the loop so that this line becomes

``` ruby
PostDestroyer.new(current_user, p, defer_flags: !agree_with_first_reply_flag || i > 0).destroy
```

techAPJ:
@eviltrout @ZogStriP Thanks for the review. I made the required changes.

Note: Travis build failure seems unrelated.

PR #7112, opened 2019-03-05
REFACTOR: Change `watch` wording to `unignore`

khalilovcmded:
## Why?

This is part of the [Ability to ignore a user feature](https://meta.discourse.org/t/ability-to-ignore-a-user/110254/8), we found that the wording `watch` is misleading.

## UI

profile_-_john_-_trekup

### DIFF
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -195,11 +195,6 @@ export default Ember.Component.extend(
         this._close();
       },
 
-      watchUser() {
-        this.get("user").watch();
-        this._close();
-      },
-
       showUser() {
         this.showUser(this.get("user"));
         this._close();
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/discourse/components/user-card-contents.js.es6
```diff
@@ -195,11 +195,6 @@ export default Ember.Component.extend(
         this._close();
       },
 
-      watchUser() {
```
khalilovcmded: Cleaning this up, I forgot it in a previous PR.

config/locales/client.en.yml
```diff
@@ -637,7 +637,7 @@ en:
       private_message: "Message"
       private_messages: "Messages"
       ignore: "Ignore"
-      watch: "Watch"
+      unignore: "Un-ignore"
```
```suggestion
      unignore: "Unignore"
```

I don't believe the hyphen is necessary here.

PR #7113, opened 2019-03-05
FEATURE: Hide ignored user's Original Post content

khalilovcmded:
## Why?

We need a method in the UI bit for collapsing the OP of an ignored user, following @ZogStriP suggestion, we think this can be suitable:
for_the_2_3_release_--_an_ignore_feature_-_todo_-_discourse_dev

## How?

This works in a simple way, but I feel it might be a little hacky, so I am open to all feedback. In the `basic_post_serializer.rb`, we basically check if this is an original post (within the stream of posts we are serializing), and if so, we check if it is ignored by the current logged in user.

**Note: _I want to write some tests but thought to expose this for discussion first._**

## UI

how_to_hike_mount_fuji__-_japan_mountains_ _-_trekup_-__private_browsing_


config/locales/server.en.yml
```diff
@@ -868,6 +868,9 @@ en:
     you_must_edit: '<p>Your post was flagged by the community. Please <a href="%{path}">see your messages</a>.</p>'
     user_must_edit: "<p>This post was flagged by the community and is temporarily hidden.</p>"
 
+  ignored:
+    hidden_content: '<i>Hidden content</i>'
```
ZogStriP: I think it would be better to remove the formatting from the string and instead add a class on the post saying it has hidden content so that we can style it properly using CSS

khalilovcmded: > add a class on the post saying it has hidden content

How do you do that? Does adding a class occur on the backend?

khalilovcmded: Maybe you meant:

`<p class="hidden-content">.....</p>`

khalilovcmded: Just wrapping up our offline discussion, we thought it would be better to:

1. Expose `ignored` attribute on the PostSerializer
2. Let the front-end handle the state of an ignored post

khalilovcmded:
@ZogStriP I will merge, this failing test (in the second last commit) has already a fix on master, so it is not a problem.

PR #7115, opened 2019-03-05
FEATURE: Special call-out for new / returning posters.

udan11:


SamSaffron:
Edge case:

- user creates post ... gets notice
- admin deletes post
- user creates another post... gets notice
- admin undeletes old post, now 2 posts have a notice 

Simplest workaround is to blow up the notice on undelete, trickier one is just to remove other dupe notices on undelete.

lib/post_creator.rb
```diff
@@ -508,6 +509,24 @@ def update_user_counts
...
 
+  def create_post_notice
+    last_post_time = Post.where(user_id: @user.id)
+      .where('created_at < ?', 1.day.ago)
+      .order(:created_at)
+      .limit(1)
+      .pluck(:created_at)
+      .first
+
+    if !last_post_time
```
eviltrout:
```suggestion
   unless last_post_time
```

lib/post_creator.rb
```diff
@@ -508,6 +509,24 @@ def update_user_counts
     @user.update_attributes(last_posted_at: @post.created_at)
   end
 
+  def create_post_notice
+    last_post_time = Post.where(user_id: @user.id)
```
ZogStriP: Why check for posts older than 1 day? Also, the order doesn't seem right 😉 

``` ruby
last_post_time = Post.where(user_id: @user.id)
  .order(created_at: :desc)
  .limit(1)
  .pluck(:created_at)
  .first
```

PR #7116, opened 2019-03-06
FIX: Fixed custom emoji circumventing "max emojis in topic title" set…

venarius:
This addresses the issue reported in the link below, where custom emojis weren't counted in the "max emojis" validation.

Resolves: https://meta.discourse.org/t/custom-emoji-circumvent-max-emojis-in-topic-title-setting/109425


ZogStriP:
Thanks 👍

PR #7117, opened 2019-03-06
 FEATURE: Save ignored usernames in user preferences

khalilovcmded:
## Why?

We are adding the backend logic for _actually_ saving ignored usernames.

### DIFF
--- a/app/services/user_updater.rb
+++ b/app/services/user_updater.rb
@@ -161,13 +161,8 @@ class UserUpdater
         INSERT into muted_users(user_id, muted_user_id, created_at, updated_at)
         SELECT :user_id, id, :now, :now
         FROM users
-        WHERE
-          id in (:desired_ids) AND
-          id NOT IN (
-            SELECT muted_user_id
-            FROM muted_users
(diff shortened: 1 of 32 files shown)


app/services/user_updater.rb
```diff
@@ -168,6 +172,30 @@ def update_muted_users(usernames)
...
+      IgnoredUser.where(user_id: user.id).destroy_all
+    else
+      IgnoredUser.where('user_id = ? AND ignored_user_id not in (?)', user.id, desired_ids).destroy_all
+
+      # SQL is easier here than figuring out how to do the same in AR
+      DB.exec(<<~SQL, now: Time.now, user_id: user.id, desired_ids: desired_ids)
+        INSERT into ignored_users(user_id, ignored_user_id, created_at, updated_at)
+        SELECT :user_id, id, :now, :now
+        FROM users
+        WHERE
```
ZogStriP: This is right but I wonder if using `ON CONFLICT DO NOTHING` would be any different? It sure is easier to read.

PR #7119, opened 2019-03-06
FIX: Notify on tag change.

nbianca:


lib/post_revisor.rb
```diff
@@ -84,6 +84,12 @@ def self.track_topic_field(field, &block)
         next
       end
       tc.record_change('tags', prev_tags, tags) unless prev_tags.sort == tags.sort
+
+      DB.after_commit do
```
ZogStriP: Will this happen if the tags are the same?

nbianca: A new notification will not be send, but that statement was indeed called again when the order of tags was changed. I've put this under the previous `unless` now to cover all cases.

ZogStriP:
🎉

PR #7121, opened 2019-03-06
FEATURE: Log user approvals.

nbianca:


app/controllers/admin/users_controller.rb
```diff
@@ -290,6 +290,7 @@ def trust_level_lock
   def approve
     guardian.ensure_can_approve!(@user)
     @user.approve(current_user)
+    StaffActionLogger.new(current_user).log_user_approve(@user)
```
ZogStriP: Is this the only spot where we approve an user? If not, it might be better to move the log in the `user.approve`  method.

spec/requests/admin/users_controller_spec.rb
```diff
@@ -67,6 +67,7 @@
     end
 
     it 'calls approve' do
+      StaffActionLogger.new(admin).expects(:log_user_approve).once
```
ZogStriP: It's better to check that the proper `UserHistory` was created rather than the `log_user_approve` method was called.

SamSaffron:
nice change!

PR #7122, opened 2019-03-06
FIX: Loading draft also replaces the recipients

nbianca:


SamSaffron:
Sure, looks good, I had to read it a few times but basically what this is saying is if we open the composer with a "username" but already have it from the draft prefer the draft.

PR #7128, opened 2019-03-07
FEATURE: New Review Queue

eviltrout:
This is a large refactor that merges all types of reviewable items in core Discourse (Users, Queued Posts and Flags) into one consolidated review queue.

## New Features

* One convenient place to review all content
* Reviewable items are scored and the highest scored items are shown first. Staff members and higher trust level users score items higher than new users.
* You can filter the review queue by status, score, reviewable type, category
* The API for reviewables returns the list of actions that can be performed on an item, which means the front end application can be simpler and not duplicate all the logic.
* Data structures are built to enable non-staff level moderation of items in the near future

## Software Design Notes

The `Reviewable` model uses single table inheritance (STI), and each type of reviewable must declare a subclass, serializer and ember component. Three types currently exist: `ReviewableQueuedPost`, `ReviewableUser` and `ReviewableFlaggedPost`. Developers are encouraged to look at how other reviewable items are implemented and build their own.

Post Actions (such as flags) must now be created using the `PostActionCreator` service object. It has a much friendlier API now and returns a result object which will return the associated reviewable.

## Backwards Compatibility

REST API is maintained with the old endpoints, however deprecation warnings are logged. Third party code that is pointed at the old APIs should be updated to the new review queue API. All of the Ember.js/front end code for the old user approval / flagging interface / queued posts is removed.

In a future release we will remove the old APIs and backwards compatibility.


app/assets/javascripts/discourse/components/reviewable-bundled-action.js.es6
```diff
@@ -0,0 +1,22 @@
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  tagName: "",
+
+  @computed("bundle.actions.length")
```
jjaffeux: Can be rewrite with:

```
multiple: Ember.computed.gt("bundle.actions.length", 1)
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+
+  // Find a component to render, if one exists. For example:
+  // `ReviewableUser` will return `reviewable-user`
+  @computed("reviewable.type")
+  reviewableComponent(type) {
+    if (_components[type] !== undefined) {
+      return _components[type];
+    }
+
+    let dasherized = Ember.String.dasherize(type);
```
jjaffeux:
```suggestion
    const dasherized = Ember.String.dasherize(type);
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+      return dasherized;
+    }
+    _components[type] = null;
+  },
+
+  _performConfirmed(action) {
+    let reviewable = this.get("reviewable");
+
+    let performAction = () => {
+      let version = reviewable.get("version");
```
jjaffeux:
```suggestion
      const version = reviewable.get("version");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+          method: "PUT"
+        }
+      )
+        .then(result => {
+          this.attrs.remove(
+            result.reviewable_perform_result.remove_reviewable_ids
+          );
+        })
+        .catch(popupAjaxError)
+        .finally(() => {
```
jjaffeux:
```
.finally(() => this.set("updating", false))
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
jjaffeux:
```suggestion
    const dasherized = Ember.String.dasherize(type);
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+  // Find a component to render, if one exists. For example:
+  // `ReviewableUser` will return `reviewable-user`
+  @computed("reviewable.type")
+  reviewableComponent(type) {
+    if (_components[type] !== undefined) {
+      return _components[type];
+    }
+
+    let dasherized = Ember.String.dasherize(type);
+    let template = Ember.TEMPLATES[`components/${dasherized}`];
```
jjaffeux:
```suggestion
    const template = Ember.TEMPLATES[`components/${dasherized}`];
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+    if (template) {
+      _components[type] = dasherized;
+      return dasherized;
+    }
+    _components[type] = null;
+  },
+
+  _performConfirmed(action) {
+    let reviewable = this.get("reviewable");
```
jjaffeux:
```suggestion
    const reviewable = this.get("reviewable");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+
+    let performAction = () => {
```
jjaffeux:
```suggestion
    const performAction = () => {
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
jjaffeux:
```suggestion
      const version = reviewable.get("version");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
jjaffeux:
```
.finally(() => this.set("updating", false));
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+          );
+        })
+        .catch(popupAjaxError)
+        .finally(() => {
+          this.set("updating", false);
+        });
+    };
+
+    if (action.client_action) {
+      let actionMethod = this[`client${action.client_action.classify()}`];
```
jjaffeux:
```suggestion
      const actionMethod = this[`client${action.client_action.classify()}`];
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+  clientSuspend(reviewable, performAction) {
+    this._penalize("showSuspendModal", reviewable, performAction);
+  },
+
+  clientSilence(reviewable, performAction) {
+    this._penalize("showSilenceModal", reviewable, performAction);
+  },
+
+  _penalize(adminToolMethod, reviewable, performAction) {
+    let adminTools = this.get("adminTools");
```
jjaffeux:
```suggestion
    const adminTools = this.get("adminTools");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+    if (adminTools) {
+      let createdBy = reviewable.get("target_created_by");
```
jjaffeux:
```suggestion
      const createdBy = reviewable.get("target_created_by");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+      let postId = reviewable.get("post_id");
```
jjaffeux:
```suggestion
      const postId = reviewable.get("post_id");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+      let postEdit = reviewable.get("raw");
```
jjaffeux:
```suggestion
      const postEdit = reviewable.get("raw");
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+      return this.get("adminTools")[adminToolMethod](createdBy, {
```
jjaffeux:
```suggestion
      return adminTools[adminToolMethod](createdBy, {
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+          delete updates[name];
+        }
+      });
+
+      this.set("updating", true);
+      return this.get("reviewable")
+        .update(updates)
+        .then(() => this.set("editing", false))
+        .catch(popupAjaxError)
+        .finally(() => {
```
jjaffeux:
```
.finally(() => this.set("updating", false));
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,162 @@
...
+    valueChanged(fieldId, event) {
+      Ember.set(this._updates, fieldId, event.target.value);
+    },
+
+    perform(action) {
+      if (this.get("updating")) {
+        return;
+      }
+
+      let msg = action.get("confirm_message");
```
jjaffeux:
```suggestion
      const msg = action.get("confirm_message");
```

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -0,0 +1,58 @@
...
+    });
+  },
+
+  actions: {
+    remove(ids) {
+      if (!ids) {
+        return;
+      }
+
+      let newList = this.get("reviewables").reject(reviewable => {
```
jjaffeux:
```suggestion
      const newList = this.get("reviewables").reject(reviewable => {
```

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -0,0 +1,58 @@
...
+      this.set("reviewables", newList);
+    },
+
+    resetTopic() {
+      this.set("topic_id", null);
+      this.send("refreshRoute");
+    },
+
+    refresh() {
+      this.set("type", this.get("filterType"));
```
jjaffeux:
```
this.setProperties({
  type: this.get("filterType"),
  min_score: this.get("filterScore"),
  status: this.get("filterStatus"),
  category_id: this.get("filterCategoryId")
})
```

app/assets/javascripts/discourse/models/reviewable.js.es6
```diff
@@ -0,0 +1,49 @@
...
+    });
+  },
+
+  update(updates) {
+    // If no changes, do nothing
+    if (Object.keys(updates).length === 0) {
+      return Ember.RSVP.resolve();
+    }
+
+    let adapter = this.store.adapterFor("reviewable");
```
jjaffeux:
```suggestion
    const adapter = this.store.adapterFor("reviewable");
```

app/assets/javascripts/discourse/routes/review-index.js.es6
```diff
@@ -0,0 +1,26 @@
+export default Discourse.Route.extend({
+  model(params) {
+    return this.store.findAll("reviewable", params);
+  },
+
+  setupController(controller, model) {
+    let meta = model.resultSetMeta;
```
jjaffeux:
```suggestion
    const meta = model.resultSetMeta;
```

app/assets/javascripts/discourse/routes/review-topics.js.es6
```diff
@@ -0,0 +1,11 @@
+export default Discourse.Route.extend({
+  model() {
+    return this.store.findAll("reviewable-topic");
+  },
+
+  setupController(controller, model) {
+    controller.setProperties({
```
jjaffeux: controller.set("reviewableTopics", model);

app/assets/javascripts/select-kit/components/select-kit.js.es6
```diff
@@ -461,6 +463,7 @@ export default Ember.Component.extend(
     clearSelection() {
       this.deselect(this.get("selection"));
       this.focusFilterOrHeader();
+      this.didClearSelection();
```
jjaffeux: ❤️

app/assets/stylesheets/common/base/reviewables.scss
```diff
@@ -0,0 +1,221 @@
...
+    align-items: center;
+    flex-wrap: wrap;
+  }
+
+  .reviewable-filter {
+    display: flex;
+    align-items: center;
+    margin-right: 1em;
+
+    label {
```
jjaffeux: as much as possible I prefer when we avoid styling on html elements, it always causes pain later

app/assets/stylesheets/common/base/reviewables.scss
```diff
@@ -0,0 +1,221 @@
...
+
+  .user {
+    display: flex;
+    align-items: center;
+
+    .user-flag-percentage {
+      margin-left: 0.5em;
+    }
+  }
+  th {
```
jjaffeux: same remark on html styles

app/assets/stylesheets/common/base/reviewables.scss
```diff
@@ -0,0 +1,221 @@
...
+  }
+}
+
+.reviewable-flagged-post {
+  .post-body {
+    max-width: $topic-body-width;
+    max-height: 300px;
+    overflow-y: auto;
+
+    p,
```
jjaffeux: html styling

app/assets/stylesheets/common/base/reviewables.scss
```diff
@@ -0,0 +1,221 @@
...
+}
+
+.editable-fields {
+  width: 100%;
+  .editable-field {
+    .mini-tag-chooser {
+      margin: 0;
+    }
+
+    input[type="text"] {
```
jjaffeux: html styling

app/controllers/admin/flags_controller.rb
```diff
@@ -19,47 +18,39 @@ def index
...
-        meta: meta
-      )
-    else
-      render_json_dump(
-        posts: posts,
-        topics: serialize_data(topics, FlaggedTopicSerializer),
-        users: serialize_data(users, FlaggedUserSerializer)
+    }
+
+    if (total_rows || 0) > (offset + per_page)
```
jjaffeux: Maybe we can store `(offset + per_page)` as we use it later

app/models/report.rb
```diff
@@ -541,10 +541,19 @@ def self.report_flags(report)
     report.icon = 'flag'
     report.higher_is_better = false
 
-    basic_report_about report, PostAction, :flag_count_by_date, report.start_date, report.end_date, report.category_id
-    countable = PostAction.where(post_action_type_id: PostActionType.flag_types_without_custom.values)
-    countable = countable.joins(post: :topic).merge(Topic.in_category_and_subcategories(report.category_id)) if report.category_id
-    add_counts report, countable, 'post_actions.created_at'
+    basic_report_about(
```
jjaffeux: ❤️

app/assets/javascripts/discourse/components/reviewable-item.js.es6
eviltrout: I don't feel as strongly about making everything `const` as I used to. I find it doesn't catch many bugs, and nothing seems faster as a result (which is initially what I'd hoped.)

app/assets/stylesheets/common/base/reviewables.scss
eviltrout: In this case I think it's more defensible, because a table has a lot of elements th, td that would all need a new class attribute. Instead I made it more explicit by introducing `> th > tr` so that the styles only apply to the format of this table. If there's a nested table they should not apply.

app/assets/stylesheets/common/base/reviewables.scss
eviltrout: I have no other control over this formatting since it's the `cooked` of a post.

app/assets/stylesheets/common/base/reviewables.scss
jjaffeux: makes sense :+1:

app/assets/javascripts/discourse/components/reviewable-item.js.es6
jjaffeux: No it mostly doesn't do anything, I agree. In the end all is transpiled into `var`. The only thing it does is hitting the developer about the fact that this is not going to be mutated later, which I do appreciate, but not strong feelings about this.

eviltrout: My current feeling on it is it's up to the developer writing the code. Using `let` or `const` fixes the weird Javascript scope issues, and they should choose what they think is more appropriate. I'll review/approve either form personally.

jjaffeux: works for me :+1:

app/models/reviewable_flagged_post.rb
```diff
@@ -0,0 +1,284 @@
+require_dependency 'reviewable'
+
+class ReviewableFlaggedPost < Reviewable
+
+  def post
+    @post ||= (target || Post.with_deleted.find_by(id: target_id))
+  end
+
+  def build_actions(actions, guardian, args)
```
romanrizzi: Is there any chance we could move action creation to: `lib/reviewable/actions.rb`.It's not directly related to the actions :thinking:

app/controllers/admin/flags_controller.rb
```diff
@@ -71,65 +62,70 @@ def agree
...
-      post: post,
-      user: current_user
-    )
+    if reviewable = post&.reviewable_flag
+      DiscourseEvent.trigger(
+        :before_staff_flag_action,
+        type: 'defer',
+        post: post,
+        user: current_user
+      )
```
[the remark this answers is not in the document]
eviltrout: That's code I didn't change from before. I do agree it could be put into a method but in this case the entire controller is deprecated and going to be deleted anyway, so probably not worth it.

app/controllers/reviewables_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+        load_more_reviewables: review_path(offset: offset + PER_PAGE, min_score: min_score),
+        min_score: min_score,
+        type: params[:type],
+        category_id: category_id,
+        topic_id: topic_id,
+        types: meta_types,
+        status: status
+      }
+    }
+    json.merge!(hash)
```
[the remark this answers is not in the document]
eviltrout: Hmmm I think the `hash` is leftover from some old code I ended up throwing away. I will do a commit here to remove it altogether, thanks.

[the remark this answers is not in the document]
eviltrout: Oh wait it's not empty. It's populated by the serializer. It's how I set up the side loading of data. It has to stay this way.

app/controllers/reviewables_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+    reviewables = Reviewable.list_for(
+      current_user,
+      status: status,
+      category_id: category_id,
+      topic_id: topic_id,
+      limit: PER_PAGE,
+      offset: offset,
+      type: params[:type],
+      min_score: min_score
+    ).to_a
```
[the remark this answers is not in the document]
eviltrout: I considered that, but also figured you then have to deal with the overhead of a merge. It might be worth it? It was a "borderline" thing for me.

app/controllers/reviewables_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+    end
+
+    status = (params[:status] || 'pending').to_sym
+    raise Discourse::InvalidParameter.new(:status) unless Reviewable.statuses[status].present?
+
+    category_id = nil
+    category_id = params[:category_id].to_i if params[:category_id]
+
+    topic_id = nil
+    topic_id = params[:topic_id].to_i if params[:topic_id]
```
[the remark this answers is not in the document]
eviltrout: Thanks, I didn't accept your suggestion here because it can eliminate the `topic_id = nil` line too so I made my own commit.

app/controllers/reviewables_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+      root: 'reviewable',
+      meta: {
+        types: meta_types
+      }
+    )
+  end
+
+  def update
+    reviewable = Reviewable.viewable_by(current_user).where(id: params[:reviewable_id]).first
+    raise Discourse::NotFound.new if reviewable.blank?
```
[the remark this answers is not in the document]
eviltrout: Good idea, I did so in 8affec2cf. I called it `find_reviewable` to avoid accidentally shadowing it since `reviewable` is a common name.

app/models/reviewable.rb
```diff
@@ -0,0 +1,394 @@
...
+    EditableFields.new(self, guardian, args).tap { |a| build_editable_fields(a, guardian, args) }
+  end
+
+  # subclasses implement "build_actions" to list the actions they're capable of
+  def build_actions(actions, guardian, args)
+  end
+
+  # subclasses implement "build_editable_fields" to list stuff that can be edited
+  def build_editable_fields(actions, guardian, args)
+  end
```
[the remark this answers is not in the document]
eviltrout: It doesn't *have* to be overridden though. It's valid to leave it empty when your item is not editable, such as users / flags right now. Instead I've changed the comment to be `subclasses can implement` to make it more clear.

lib/flag_query.rb
```diff
@@ -12,34 +12,26 @@ def self.register_plugin_post_custom_field(field, plugin)
...
-
-    post_ids_relation = actions.limit(per_page)
-      .offset(offset)
-      .group(:post_id)
-      .order('MIN(post_actions.created_at) DESC')
-
-    if opts[:filter] != "old"
-      post_ids_relation = PostAction.apply_minimum_visibility(post_ids_relation)
-    end
+    total_rows = reviewables.count
```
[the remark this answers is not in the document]
eviltrout: If you did that you'd get back the count of the rows by `limit / offset` and not the total rows, which is used for pagination.

Also I should mention this is the deprecated API so performance isn't the biggest concern.

app/models/reviewable.rb
```diff
@@ -0,0 +1,394 @@
+require_dependency 'enum'
+require_dependency 'reviewable/actions'
+require_dependency 'reviewable/conversation'
+require_dependency 'reviewable/editable_fields'
+require_dependency 'reviewable/perform_result'
+require_dependency 'reviewable_serializer'
+
+class Reviewable < ActiveRecord::Base
```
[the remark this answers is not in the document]
eviltrout: Can you be more specific? You think all the class methods should be up top and instance methods below for example?

app/models/topic_converter.rb
```diff
@@ -52,13 +39,24 @@ def convert_to_private_message
...
 
   private
 
+  def update_user_actions
+    UserAction.where(
+      target_topic_id: @topic.id,
+      action_type: [UserAction::GOT_PRIVATE_MESSAGE, UserAction::NEW_PRIVATE_MESSAGE]).find_each do |ua|
+        UserAction.remove_action!(ua.attributes.symbolize_keys.slice(:action_type, :user_id, :acting_user_id, :target_topic_id, :target_post_id))
+      end
+    @topic.posts.each { |post| UserActionManager.post_created(post) }
```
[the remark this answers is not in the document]
eviltrout: This is a good point, I will make a note about this and will follow up. Topic Conversion was already pretty slow, but on a large topic this could be a time bomb.

lib/flag_query.rb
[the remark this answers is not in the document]
romanrizzi: You're right. I Missed that part

lib/flag_query.rb
```diff
@@ -71,68 +62,56 @@ def self.flagged_posts_report(current_user, opts = nil)
...
-          user_ids << response.user_id
-          if reply = related_topic.ordered_posts[1]
-            conversation[:reply] = {
-              excerpt: excerpt(reply.cooked),
-              user_id: reply.user_id
+    reviewables.each do |r|
+      post = post_lookup[r.target_id]
+      post.post_action_ids ||= []
+
+      r.reviewable_scores.order('created_at desc').each do |rs|
```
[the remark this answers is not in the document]
eviltrout: `reviewable_scores`  should be preloaded by the `viewable_by` class method already?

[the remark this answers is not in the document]
romanrizzi: Great! Closing this one

app/models/reviewable_flagged_post.rb
eviltrout: Can you elaborate? The actions are built dynamically per reviewable type, so these actions are `ReviewableFlaggedPost` specific. You'd have to have `reviewable/actions.rb` test the type of object in some kind of case statement I think.

app/controllers/reviewables_controller.rb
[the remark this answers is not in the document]
romanrizzi: IMO it depends if both calls are highly connected and the common options are unlikely to be changed only in one place. It's better to avoid duplication when possible

eviltrout:
> Sorry for the number of comments 

No problem! Keep them coming. Every suggestion helps out here.

app/models/reviewable.rb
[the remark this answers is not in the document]
romanrizzi: Exactly. I tend to organize my code that way. what do you think?

app/controllers/reviewables_controller.rb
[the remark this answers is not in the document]
romanrizzi: Oh, I see. The hash gets populated as a side effect of the serializer call and that makes it pretty confusing

app/models/reviewable_flagged_post.rb
romanrizzi: This is a pretty complex class that not only perform the review actions but also decides which actions will be visible in the UI. I was asking if we could move some of that complexity away by moving that responsibility to `actions.rb` (if it makes sense)

app/controllers/reviewables_controller.rb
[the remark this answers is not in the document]
eviltrout: Yes unfortunately that's the AMS API for handling this. I had to use this hacky way because of STI and I'm not sure there's a better way of doing it.

app/models/reviewable.rb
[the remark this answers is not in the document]
eviltrout: It's never been a rule of ours before. I wonder if we should add this to rubocop because it seems like a style change (lots of our code does this.)

app/models/reviewable_flagged_post.rb
eviltrout: The actions are used for display in the UI, but it's also used for access checks. For example, if a reviewable is in the `approved` state, it won't return an `approve` action, and if an API request comes in that attempts to review it, it will fail because that action does not exist.

The problem with moving it to actions is then the actions class needs to know about all types of reviewables and how they should respond to things so it could build the appropriate actions.

app/models/reviewable.rb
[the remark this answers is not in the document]
romanrizzi: It's definitely something to discuss. This will help us get cleaner interfaces. It's ok to leave this way for now if you want to.

app/models/reviewable_flagged_post.rb
romanrizzi: That's why I bring this up. This class is taken care of:
  - UI elements
  - Authorization
  - Persistence

and is also an entity of our business domain.

My main intention is just to raise awareness that tight coupling leads to systems that are harder to extend and evolve :stuck_out_tongue_winking_eye:

eviltrout: I think in this case I'm still seeing how you would improve things. If you like, you could create a follow up PR showing how you'd split things up and how it would improve the code.

app/models/reviewable.rb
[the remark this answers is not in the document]
eviltrout: I would suggest a follow up post on our dev or meta forums about this. You can suggest the change and if others are onboard enable the rubocop rule and convert the old code to work.

romanrizzi:
Will do!

romanrizzi:
I will try to move the actions to the Action class in a follow up PR after
finishing my work on the akismet plugin

app/controllers/reviewables_controller.rb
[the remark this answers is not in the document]
eviltrout: I made this change and it does seem a lot easier to follow, at the minor cost of two extra `merge` calls which should be fairly fast.

eviltrout:
Okay I've squashed all the commits, and rebased on top of `master` with all the feedback above. There's still time for more feedback as the target merge day would be Wednesday morning.

pmusaraj:
I pulled this locally, and noticed some issues while playing around with it (mostly with queued posts): 

1. When a new user posts a reply (site is configured to queue all posts from new users), and an admin approves the reply, the admin is displayed as the poster of the reply. This happens on replies and new topics. 

2. The count of pending posts for a user doesn't reset after approvals. For example, I submitted several test replies and posts as a new user, and approved some of them by the admin account, but the count in the modal keeps incrementing. 

3. There is a delay before the red badge notification shows up for admins. (This might be a local issue, though I don't see a delay on master locally.)

romanrizzi:
Looks like the status it's not changing :thinking: 

```ruby
spam_reporter = Discourse.system_user
post = Fabricate(:post)
reviewable = ReviewableFlaggedPost.needs_review!(target: post, topic: post.topic, created_by: spam_reporter)
reviewable.status # 0
reviewable.agree(spam_reporter, {})
reviewable.reload.status # 0
```

Am I doing something wrong?

eviltrout:
Thanks Penar, some really good finds. 

1 + 2 have been fixed, legit bugs.

3. I believe is because I am now updating those counts asynchronously. The issue here is previously we could broadcast to all staff members, but with non-staff members approving I needed to do some more logic. I'm hoping it'll be faster outside of dev mode or otherwise not a problem. Let's keep an eye on it.

app/serializers/queued_post_serializer.rb
```diff
@@ -1,36 +1,74 @@
...
+      reviewable_histories.
+      where(
+        reviewable_history_type: ReviewableHistory.types[:transitioned],
+        status: Reviewable.statuses[status]
+      ).
+      order(:created_at)
+      .last&.created_by_id
+  end
+
+  #
```
tgxworld: Stray `#` here :grin:

app/serializers/queued_post_serializer.rb
```diff
@@ -1,36 +1,74 @@
...
+  end
+
+  def raw
+    object.payload['raw']
+  end
+
+  def post_options
+    json = object.payload.dup
+    json.delete('raw')
+    json
```
tgxworld: You can use `Hash#except` in the future :)

app/models/post.rb
```diff
@@ -462,8 +456,49 @@ def active_flags
...
+    reason ||= hidden_at ?
+      Post.hidden_reasons[:flag_threshold_reached_again] :
+      Post.hidden_reasons[:flag_threshold_reached]
+
+    hiding_again = hidden_at.present?
+
+    self.hidden = true
+    self.hidden_at = Time.zone.now
+    self.hidden_reason_id = reason
+    save
```
tgxworld:
```suggestion
    save!
```

app/models/reviewable.rb
```diff
@@ -0,0 +1,395 @@
...
+      user: user,
+      status: ReviewableScore.statuses[:pending],
+      reviewable_score_type: reviewable_score_type,
+      score: ReviewableScore.user_flag_score(user) + type_bonus + take_action_bonus,
+      meta_topic_id: meta_topic_id,
+      take_action_bonus: take_action_bonus,
+      created_at: created_at || Time.zone.now
+    )
+
+    update(score: self.score + rs.score, latest_score: rs.created_at)
```
tgxworld:
```suggestion
    update!(score: self.score + rs.score, latest_score: rs.created_at)
```

app/models/reviewable.rb
```diff
@@ -0,0 +1,395 @@
...
+    topic.update(reviewable_score: topic.reviewable_score + rs.score) if topic
```
tgxworld:
```suggestion
    topic.update!(reviewable_score: topic.reviewable_score + rs.score) if topic
```

app/models/reviewable.rb
```diff
@@ -0,0 +1,395 @@
...
+#  target_id               :integer
+#  target_type             :string
+#  target_created_by_id    :integer
+#  payload                 :json
+#  version                 :integer          default(0), not null
+#  latest_score            :datetime
+#  created_at              :datetime         not null
+#  updated_at              :datetime         not null
+#
+# Indexes
```
tgxworld: Do we need indexes for `topic_id`, `category_id`, `created_by_id`, `reviewable_by_group_id` and `claimed_by_id`? It seems like some of this columns are used for filtering so an index is going to speed up the query once the reviewables table grows.

app/models/reviewable_flagged_post.rb
```diff
@@ -0,0 +1,292 @@
...
+  def perform_ignore(performed_by, args)
+    actions = PostAction.active
+      .where(post_id: target_id)
+      .where(post_action_type_id: PostActionType.notify_flag_type_ids)
+
+    actions.each do |action|
+      action.deferred_at = Time.zone.now
+      action.deferred_by_id = performed_by.id
+      # so callback is called
+      action.save
```
tgxworld:
```suggestion
      action.save!
```

app/models/reviewable_flagged_post.rb
```diff
@@ -0,0 +1,292 @@
...
+        PostActionType.notify_flag_type_ids
+      end
+
+    actions = PostAction.active.where(post_id: target_id).where(post_action_type_id: action_type_ids)
+
+    actions.each do |action|
+      action.disagreed_at = Time.zone.now
+      action.disagreed_by_id = performed_by.id
+      # so callback is called
+      action.save
```
tgxworld:
```suggestion
      action.save!
```

app/models/reviewable_history.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# == Schema Information
+#
+# Table name: reviewable_histories
+#
+#  id                      :bigint(8)        not null, primary key
+#  reviewable_id           :integer          not null
+#  reviewable_history_type :integer          not null
+#  status                  :integer          not null
+#  created_by_id           :integer          not null
```
tgxworld: I think we need an index here as well.

app/models/reviewable_queued_post.rb
```diff
@@ -0,0 +1,132 @@
...
+    ))
+    created_post = creator.create
+
+    unless created_post && creator.errors.blank?
+      return create_result(:failure) { |r| r.errors = creator.errors }
+    end
+
+    payload['created_post_id'] = created_post.id
+    payload['created_topic_id'] = created_post.topic_id unless topic_id
+    save
```
tgxworld:
```suggestion
    save!
```

app/models/reviewable_queued_post.rb
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  def perform_delete_user(performed_by, args)
+    delete_options = {
+      context: I18n.t('reviewables.actions.delete_user.reason'),
+      delete_posts: true,
+      delete_as_spammer: true
+    }
+
+    if Rails.env.production? && ENV["Staging"].nil?
```
tgxworld: I'm curious about this conditional here. Is there a specific use case for this? My worry here is that the ENV is very generic and its use case isn't documented.

app/models/reviewable_score.rb
```diff
@@ -0,0 +1,86 @@
...
+#  status                :integer          not null
+#  score                 :float            default(0.0), not null
+#  take_action_bonus     :float            default(0.0), not null
+#  reviewed_by_id        :integer
+#  reviewed_at           :datetime
+#  meta_topic_id         :integer
+#  created_at            :datetime         not null
+#  updated_at            :datetime         not null
+#
+# Indexes
```
tgxworld: I feel like most of the `*_id` here will benefit from an index.

app/models/web_hook.rb
```diff
@@ -55,6 +55,10 @@ def self.enqueue_hooks(type, event, opts = {})
   end
 
   def self.enqueue_object_hooks(type, object, event, serializer = nil)
+    if type == :flag
+      Discourse.deprecate("The flags webhook is deprecated. Please use reviewable instead.")
```
tgxworld: Wouldn't this end up spamming our logs in production since it is still being used? Also, I think it'll be helpful to include the version that we're removing this method in.

db/migrate/20190110201340_remove_google_plus_from_share_links_site_setting.rb
```diff
@@ -8,6 +8,6 @@ def up
   end
 
   def down
-    raise ActiveRecord::IrreversibleMigration
+    # raise ActiveRecord::IrreversibleMigration
```
tgxworld: I think this shouldn't be commented?

lib/tasks/user_actions.rake
```diff
@@ -2,13 +2,38 @@ desc "rebuild the user_actions table"
...
-  Topic.all.each { |i| UserActionCreator.log_topic(i) }
-  Post.all.each { |i| UserActionCreator.log_post(i) }
+  PostAction.all.each do |i|
+    if i.deleted_at.nil?
+      UserActionManager.post_action_created(i)
+    else
+      UserActionManager.post_action_destroyed(i)
+    end
+  end
+  Topic.all.each { |i| UserActionManager.log_topic(i) }
```
tgxworld:
```suggestion
  Topic.all.find_each { |i| UserActionManager.log_topic(i) }
```

lib/tasks/user_actions.rake
```diff
@@ -2,13 +2,38 @@ desc "rebuild the user_actions table"
 task "user_actions:rebuild" => :environment do
   MessageBus.off
   UserAction.delete_all
-  PostAction.all.each { |i| UserActionCreator.log_post_action(i) }
-  Topic.all.each { |i| UserActionCreator.log_topic(i) }
-  Post.all.each { |i| UserActionCreator.log_post(i) }
+  PostAction.all.each do |i|
```
tgxworld:
```suggestion
  PostAction.all.find_each do |i|
```

tgxworld:
@eviltrout I left some minor comments after reading through the code. The PR is huge so it was more like a skim rather than a review. My only worry here is that we're missing indexes on a couple of relational columns which are most likely going to be used for filtering purposes.

app/jobs/regular/topic_action_converter.rb
```diff
@@ -0,0 +1,18 @@
+class Jobs::TopicActionConverter < Jobs::Base
+
+  # Re-creating all the user actions could be very slow, so let's do it in a job
+  # to avoid a N+1 query on a front facing operation.
+  def execute(args)
+    topic = Topic.find_by(id: args[:topic_id])
```
romanrizzi: I was wondering if an includes `:posts` will help us here. We only moved the N+1 issue to sidekiq

eviltrout: By default in Rails that would not change anything, because it would select all the posts in the `.posts` method. The real N+1 concern here was that for every post we are creating a new user history object.

app/serializers/queued_post_serializer.rb
eviltrout: nice tip, following up with a fix.

app/models/reviewable.rb
eviltrout: @tgxworld Good question! I didn't add them because they are generally applied *after* all the other indexes that are more important. 

Almost every time you view the reviewable queue you are viewing pending (status) ordered by (score DESC). On most forums that would return less than 10 records I'd imagine unless the mods were absent, so the other filters needn't have indexes.

There might be a case where we find all the reviewables for a topic that I've forgotten, but honestly I'd prefer to deploy experimentally and watch the DB stats and add indexes as needed, rather than adding indexes to columns that might not need them.

app/models/reviewable_queued_post.rb
eviltrout: This isn't new code, it's old Queued Posts code that I migrated over. I looked back and I wrote that code many years ago. It's possible we don't need it but it's probably safer to leave for now.

app/models/reviewable_score.rb
eviltrout: There is a performance penalty on indexing columns where inserts become slower. I would rather show that we are using those columns for a particular thing rather than adding indexes on everything. I am not as certain that we will filter a score based on who reviewed it, for example. You'll almost always want to show a score associated with a reviewable (which has an index.)

app/models/reviewable_queued_post.rb
tgxworld: Ah I’m quite sure we can nuke this code now. I don’t think we have any clusters with the Staging ENV. Plus, I would consider this a private API so we don’t have to worry about deprecating.

app/models/web_hook.rb
eviltrout: I'm not sure how many people are actively using this particular web hook, but we should very quickly migrate them to the alternative. I would personally like to see the logs and take that responsibility on.

db/migrate/20190110201340_remove_google_plus_from_share_links_site_setting.rb
eviltrout: Oops, I had to do that temporarily to re-build my database in dev. Fixed in an upcoming commit. Good catch.

app/models/reviewable_queued_post.rb
eviltrout: Sure thing, nuked in an upcoming commit.

app/models/reviewable_score.rb
tgxworld: I guess I’ve always preferred to add indexes to foreign key columns since they are very commonly used in filters and we tend to forget about the index when adding filters in the future. My view is that the overhead on insertion is there but not expensive though. Most of the problems I’ve seen with slow DB queries are usually due to missing indexes on read queries and almost never on write. However, whether we have the additional indexes here or not is not a show stopper :P

eviltrout:
Okay, rebased, force pushed up latest batch of fixes. Thanks for excellent feedback @tgxworld, @pmusaraj, @jjaffeux, @romanrizzi - I still plan on merging it tomorrow morning but there's still time for more tweaks.

app/models/web_hook.rb
SamSaffron: I think deprecate is pretty safe it has callsite based protection so the flooding should not be severe.

app/assets/javascripts/admin/templates/users-list-show.hbs
```diff
@@ -84,15 +67,10 @@
...
-            {{#if showApproval}}
-              <td>
-                {{#if user.approved}}
-                  {{i18n 'yes_value'}}
-                {{else}}
-                  {{i18n 'no_value'}}
-                {{/if}}
-              </td>
+            {{#if siteSettings.must_approve_users}}
+              <td>{{i18n-yes-no user.approved}}</td>
```
ZogStriP: 😍

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,159 @@
...
+  @computed("reviewable.type")
+  reviewableComponent(type) {
+    if (_components[type] !== undefined) {
+      return _components[type];
+    }
+
+    let dasherized = Ember.String.dasherize(type);
+    let templatePath = `components/${dasherized}`;
+    let template = Ember.TEMPLATES[`${templatePath}`] || Ember.TEMPLATES[`javascripts/${templatePath}`];
+    if (template) {
```
ZogStriP:
```suggestion
    _components[type] = template ? dasherized : null;
    return _components[type];
```

app/assets/javascripts/discourse/components/reviewable-item.js.es6
```diff
@@ -0,0 +1,159 @@
...
+  },
+
+  clientSilence(reviewable, performAction) {
+    this._penalize("showSilenceModal", reviewable, performAction);
+  },
+
+  _penalize(adminToolMethod, reviewable, performAction) {
+    let adminTools = this.get("adminTools");
+    if (adminTools) {
+      let createdBy = reviewable.get("target_created_by");
```
ZogStriP: I wonder if `getProperties` is faster?

app/assets/javascripts/discourse/helpers/editable-value.js.es6
```diff
@@ -0,0 +1,17 @@
+export function formatCurrency([reviewable, fieldId]) {
```
ZogStriP: Why this name? Are we dealy with currencies?

app/assets/javascripts/discourse/initializers/subscribe-user-notifications.js.es6
```diff
@@ -21,18 +21,12 @@ export default {
...
-          if (data.post_queue_new_count > 0) {
-            user.set("show_queued_posts", 1);
-          }
-        });
-      }
-
+      bus.subscribe("/reviewable_counts", data => {
+        user.set("reviewable_count", data.reviewable_count);
+        if (data.reviewable_count > 0) {
+          user.set("show_reviewables", 1);
```
ZogStriP: What does `1` mean? Is `show_reviewables` used anywhere?

app/assets/javascripts/discourse/templates/components/reviewable-queued-post.hbs
```diff
@@ -0,0 +1,36 @@
+
```
ZogStriP: Empty line 🤷‍♂️

app/assets/javascripts/discourse/templates/review-topics.hbs
```diff
@@ -0,0 +1,42 @@
+<h3 class='reviewable-title'>
+  {{#link-to 'review.index' (query-params topic_id=null)}}{{i18n "review.view_all"}}{{/link-to}}
+  |
+  {{i18n "review.grouped_by_topic"}}
+</h3>
+
+{{#if reviewableTopics}}
+	<table class='reviewable-topics'>
```
ZogStriP: Identation is weird

app/controllers/post_actions_controller.rb
```diff
@@ -48,21 +36,25 @@ def create
...
+    result = PostActionDestroyer.new(
+      current_user,
+      Post.find_by(id: params[:id].to_i),
+      @post_action_type_id
+    ).perform
+
+    if result.failed?
+      render_json_error(result)
+    else
+      render_post_json(result.post, _add_raw = false)
```
ZogStriP: I know it's not your change, but we should probably move the `add_raw` optional parameter to a named parameter?

app/controllers/reviewables_controller.rb
```diff
@@ -0,0 +1,164 @@
...
+  PER_PAGE = 10
+
+  before_action :version_required, only: [:update, :perform]
+
+  def index
+    min_score = params[:min_score].nil? ? SiteSetting.min_score_default_visibility : params[:min_score].to_i
+    offset = params[:offset].to_i
+
+    if params[:type].present?
+      unless ['ReviewableUser', 'ReviewableQueuedPost', 'ReviewableFlaggedPost'].include?(params[:type])
```
ZogStriP: Should we extract that array into a constant so that it's not re-created every time this method is called?

romanrizzi: discourse-akismet will add a new Class called `ReviewableAkismetPost`. We need a way to discover this dynamically :thinking:

app/models/reviewable_score.rb
eviltrout: This sounds like a dev topic to me "Should we index all foreign key columns?" :)

For now I've been watching the perf closely (and will continue to do so after merge) and I believe we will catch any missing indexes. If you want to open that dev topic was can make it a bigger discussion which GitHub is not ideal for!

app/assets/javascripts/discourse/components/reviewable-item.js.es6
eviltrout: I don't think it's faster when you consider it has to be de-structured too since the result is a JS object. Also it's been a while since I reviewed `getProperties` but previously it was a for loop on the properties, which is certainly slower than a few `get`s!

app/controllers/post_actions_controller.rb
eviltrout: Sure it's a small thing so I'll add it.

app/controllers/reviewables_controller.rb
eviltrout: I'll add a helper method for this shortly after we merge.

app/assets/javascripts/discourse/helpers/format-score.js.es6
```diff
@@ -0,0 +1,5 @@
+import { registerUnbound } from "discourse-common/lib/helpers";
+
+registerUnbound("format-score", function(score) {
+  return (score || 0).toFixed(1);
```
gschlager: This number should be localized:

```suggestion
  return I18n.toNumber(score || 0, { precision: 1 });
```

PR #7133, opened 2019-03-08
FIX: experimental fix to Ember.ENV.EXTEND_PROTOTYPES rare exception

jjaffeux:


jjaffeux:
It's another attempt to fix: 

```
TypeError: undefined is not an object (evaluating 'Ember.ENV.EXTEND_PROTOTYPES')
```

I can’t reproduce this and it's the only place in the code where we have this code. If `.w()` is defined on string prototype it means that ember has been extending the String prototype, so it should have the exact same behavior.

I want to at least live with this a few days, to see if the exceptions still appears, it would mean it's coming from another source I haven’t identified yet (maybe something internal to ember, called by another piece of code?)

PR #7136, opened 2019-03-08
FIX: Text content search warning if more than 50 results

venarius:
This adds a warning to site texts search if more than 50 results were found.

Resolves: https://meta.discourse.org/t/customize-text-content-search-results-missing/65327

### DIFF
--- a/app/assets/stylesheets/common/admin/admin_base.scss
+++ b/app/assets/stylesheets/common/admin/admin_base.scss
@@ -260,6 +260,9 @@ $mobile-breakpoint: 700px;
       margin-top: 1em;
     }
   }
+  .warning {
+    color: #e45735;
+  }
 }
 
 .content-list {
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -4033,6 +4033,7 @@ en:
         go_back: "Back to Search"
         recommended: "We recommend customizing the following text to suit your needs:"
         show_overriden: "Only show overridden"
+        more_than_50_results: "There are more than 50 results. Please refine your search."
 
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/common/admin/admin_base.scss
```diff
@@ -260,6 +260,9 @@ $mobile-breakpoint: 700px;
       margin-top: 1em;
     }
   }
+  .warning {
+    color: #e45735;
```
eviltrout: This should be based on our SCSS variables for theme support. I'd also suggest making it `p.warning` in case the generic `warning` class is used elsewhere.

config/locales/client.en.yml
```diff
@@ -4033,6 +4033,7 @@ en:
         go_back: "Back to Search"
         recommended: "We recommend customizing the following text to suit your needs:"
         show_overriden: "Only show overridden"
+        more_than_50_results: "There are more than 50 results. Please refine your search."
```
eviltrout: I would suggest something like "Many results were found. The first 50 results are shown." to be more clear about what was shown.

venarius: In the locales there is also the key `site_settings.more_than_30_results` which is `"There are more than 30 results. Please refine your search or select a category."`

Maybe keep it like this so the wording is consistent for this type of error?

eviltrout: Sure sounds good. I would change it but it's better to change both so  let's leave as is.

venarius:
Had to move this to #7145 because my forgotten cronjob deleted my forked repository ._. Sorry!

PR #7137, opened 2019-03-08
FEATURE: unhide the embed_whitelist_selector setting

scossar:
This unhides the `embed_whitelist_selector` Site Setting and moves it to the Setting's 'posting' section.

This is mainly intended to be used by WordPress sites. Since the recent update to WordPress, markup is added to images that is preventing them from displaying when a post is expanded on Discourse.


scossar:
I see where this is failing. Will fix.

PR #7138, opened 2019-03-10
FIX: Do not trigger post alerts for empty posts.

nbianca:


spec/components/post_creator_spec.rb
```diff
@@ -1141,6 +1141,20 @@
...
+      SiteSetting.queue_jobs = false
+      user2 = Fabricate(:user)
+      topic = Fabricate(:private_message_topic,
+        topic_allowed_users: [
+          Fabricate.build(:topic_allowed_user, user: user),
+          Fabricate.build(:topic_allowed_user, user: user2)
+        ],
+      )
+
+      PostAlerter.any_instance.expects(:after_save_post).never
```
tgxworld: Is there another way to test this? We generally do not want to use stubs or mocks in our test suite.

spec/components/post_creator_spec.rb
```diff
@@ -1141,6 +1141,20 @@
...
+      PostCreator.create(user, raw: "", topic_id: topic.id, skip_validations: true)
```
tgxworld:
```suggestion
      PostCreator.create!(user, raw: "", topic_id: topic.id, skip_validations: true)
```

app/jobs/regular/post_alert.rb
```diff
@@ -5,7 +5,7 @@ class PostAlert < Jobs::Base
 
     def execute(args)
       post = Post.find_by(id: args[:post_id])
-      if post&.topic
+      if post&.topic && !post.raw.blank?
```
tgxworld: Can't really remember now but are post actions represented as post with empty raw?

ZogStriP: It's better to always use the positive `post.present?` is easier to understand than `!post.blank?`

```suggestion
      if post&.topic && post.raw.present?
```

SamSaffron:
The trick here is to create it for real with someone that would be notified normally and then dig into the notifications table.

spec/components/post_creator_spec.rb
```diff
@@ -1141,6 +1141,26 @@
       post_creator = PostCreator.new(user, title: '', raw: '')
       expect { post_creator.create! }.to raise_error(ActiveRecord::RecordNotSaved)
     end
+
+    it "does not generate an alert for empty posts" do
+      SiteSetting.queue_jobs = false
```
SamSaffron: oh ... sorry ... but @eviltrout just killed this setting, can you migrate to the new pattern?

spec/components/post_creator_spec.rb
```diff
@@ -1141,6 +1141,26 @@
...
+      )
+      Fabricate(:topic_user,
+        topic: topic,
+        user: user2,
+        notification_level: TopicUser.notification_levels[:watching]
+      )
+
+      expect {
+        PostCreator.create!(user, raw: "", topic_id: topic.id, skip_validations: true)
+      }.to change { user2.notifications.count }.by(0)
```
SamSaffron: I vaguely recall we may need to enable something else... to be extra safe here check that count changes by 1 and add

PostCreator.create!(user, raw: "hello world", topic_id: topic.id, skip_validations: true)

PR #7139, opened 2019-03-11
FIX: Fixed vanishing select boxes on full search as new content loads

venarius:
When more than 50 results were found on full page search and more results were loaded while in bulk select mode, the checkboxes would vanish and bulk select mode would turn off. This has been fixed.

Resolves: https://meta.discourse.org/t/bulk-actions-in-search-results-checkboxes-vanish-as-new-content-loads/111235

### DIFF
--- a/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
+++ b/app/assets/javascripts/discourse/controllers/full-page-search.js.es6
@@ -216,7 +216,7 @@ export default Ember.Controller.extend({
       return;
     }
 
-    var args = { q: searchTerm, page: this.get("page") };
+    let args = { q: searchTerm, page: this.get("page") };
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/controllers/full-page-search.js.es6
```diff
@@ -216,12 +216,14 @@ export default Ember.Controller.extend({
       return;
     }
 
+    var args = { q: searchTerm, page: this.get("page") };
```
eviltrout:
```suggestion
    let args = { q: searchTerm, page: this.get("page") };
```

venarius:
Wanted to write a test but you would need at least 51 test topics for it.

eviltrout:
Thanks!

PR #7143, opened 2019-03-11
FEATURE: add more granular user option levels for email notifications

pmusaraj:
<del>This PR adds a new `email_private_messages_always` option for user email preferences, which defaults to `TRUE`. </del>

This PR migrates email user options to a new data structure: 

* `email_messages_level`, with options: `always`, `only_when_away` and `never` (defaults to `always`)
* `email_level`, with options: `always`, `only_when_away` and `never` (defaults to `only_when_away`)

The respective default site settings have also been migrated. 

The UI now looks like this: 
image


app/assets/javascripts/discourse/controllers/preferences/emails.js.es6
```diff
@@ -42,6 +43,76 @@ export default Ember.Controller.extend(PreferencesTabController, {
     { name: I18n.t("user.email_previous_replies.never"), value: 2 }
   ],
 
+  @computed()
+  emailMessageValue() {
+    return JSON.stringify(
```
eviltrout: Converting the options to a JSON string and back seems a bit cumbersome here.

My suggestion would be to use a single string value could be any of the three values. In your `emailOptionsChanged` callback use a `switch` to set the appropriate values based on the string.

I am also curious if you considered using a single value for this setting on the server side rather than adding a second option, which seems simpler in many ways?

pmusaraj: I have been going back and forth between all three options (JSON conversion, setting a string value and updating the options server-side) and ended up with the JSON conversion because it seemed cleaner (i.e. less code). 

I just noticed that we have a "checkbox + dropdown" in the UI for Activity Summary (in the same screen) so I will try using that pattern for the email preferences, and do any conversion, if necessary, in `emailOptionsChanged`.

app/assets/javascripts/discourse/templates/preferences/emails.hbs
```diff
@@ -7,23 +7,45 @@
...
+      {{combo-box valueAttribute="value" content=emailAlwaysOptions value=model.user_option.email_private_messages_always}}
+    </div>
+    {{#unless model.user_option.email_private_messages_always}}
+      <div class='instructions'>
+        {{#if siteSettings.email_time_window_mins}}
+          {{i18n 'user.email.frequency' count=siteSettings.email_time_window_mins}}
+        {{else}}
+          {{i18n 'user.email.frequency_immediately'}}
+        {{/if}}
+    </div>
```
tgxworld: Small one. Indentation is off here.

app/assets/javascripts/discourse/templates/preferences/emails.hbs
```diff
@@ -7,23 +7,45 @@
...
+      {{combo-box valueAttribute="value" content=emailAlwaysOptions value=model.user_option.email_always}}
+    </div>
+    {{#unless model.user_option.email_always}}
+      <div class='instructions'>
+        {{#if siteSettings.email_time_window_mins}}
+          {{i18n 'user.email.frequency' count=siteSettings.email_time_window_mins}}
+        {{else}}
+          {{i18n 'user.email.frequency_immediately'}}
+        {{/if}}
+    </div>
```
tgxworld: Same as https://github.com/discourse/discourse/pull/7143/files#r264484239

app/jobs/regular/user_email.rb
```diff
@@ -130,7 +131,7 @@ def message_for_email(user, post, type, notification, args = nil)
           return [nil, nil]
         end
 
-        unless user.user_option.email_always?
+        unless (user.user_option.email_always? && type != :user_private_message) || (user.user_option.email_private_messages_always? && type == :user_private_message)
```
tgxworld: There is some code duplication between here and line 103. Perhaps we can extract the logic into a method?

tgxworld:
@pmusaraj I made some changes to how we test `Jobs::UserEmail` so you'll need to rebase. Sorry! Aside from the minor comments, the PR looks good to me.

app/jobs/regular/user_email.rb
pmusaraj: Thanks for raising this, there was a bug just above line 103 too, where `email_always` would affect all types of posts, including private messages. Fixed and added a test. 

(Will need to update tests to avoid stubs and fix merge conflict.)

pmusaraj:
@tgxworld not a problem! Will rebase shortly.

SamSaffron:
I love this feature but the data model bugs me and the UX also should be simplified a bit:

```
Send me an email when some messages me 
[always]
only when I am offline
never

Send me an email when someone quotes me, replies to my post, mentions my @username, or invites me to a topic
[always]
only when I am offline
never
```

Notice how there is no checkbox before `Send me an email when some messages me` and the second one. 

We have 2 fields on user options table that are in one of the 3 states. `always`, `only when offline`, `never`

So we got to move away from booleans here, to simplify everything. I know the migration is a tiny bit painful but the data model will be a lot cleaner

````

if user_option.email_messages_level == UserOption.email_levels[:offline]

# AND 

if user_option.email_level == UserOption.email_levels[:always]

```

Seems like a cleaner data model to me that is easier to reason about.

eviltrout:
Yes exactly what @SamSaffron said - that was what I wanted too.

SamSaffron:
> Where are the trooleans?

Sadly I can only find one

image

Oh ... I said trooleans not trolleans.

pmusaraj:
Ok, I have updated the PR with server-side changes. It's considerably bigger than the first PR, so will require re-reviewing (mostly server-side, client-side it's a lot simpler). 

I initially wrote reversible migrations, but I don't think the changes here can be reversed because in the new data model, `email_messages_level = :always` is decoupled from `email_level = always`.

gschlager:
I'd prefer if you could use `UserOption.email_level_types` as much as possible instead of using magic numbers. I see a lot of `0` and `2` in the code and have no idea what they mean unless I go looking.

app/assets/javascripts/discourse/controllers/preferences/emails.js.es6
```diff
@@ -51,6 +56,25 @@ export default Ember.Controller.extend(PreferencesTabController, {
...
 
+  @computed("model.user_option.email_messages_level")
+  emailMessageLevelAway(level) {
+    return level === 1;
+  },
+
+  @computed("model.user_option.email_level")
+  emailLevelAway(level) {
+    return level === 1;
+  },
```
eviltrout: This should be a `const` in javascript rather than `1`.

Also it's a good candidate for `Ember.computed.equal`. I'd suggest replacing both computed properties with that macro.

app/assets/javascripts/discourse/controllers/preferences/emails.js.es6
```diff
@@ -51,6 +56,25 @@ export default Ember.Controller.extend(PreferencesTabController, {
...
+
+  @computed()
+  emailFrequencyInstructions() {
+    if (this.siteSettings.email_time_window_mins)
```
eviltrout: Not sure why our es6 linter didn't flag this but we should use braces for if statements.

app/models/email_level_site_setting.rb
```diff
@@ -0,0 +1,22 @@
...
+    val.to_i.to_s == val.to_s &&
+    values.any? { |v| v[:value] == val.to_i }
+  end
+
+  def self.values
+    @values ||= [
+      { name: 'user.email_level.always',  value:  0 },
+      { name: 'user.email_level.only_when_away',   value:  1 },
+      { name: 'user.email_level.never', value:  2 },
+    ]
```
eviltrout: The lines above have some extra space (before and after `value`)

db/migrate/20190312181641_migrate_email_user_options.rb
```diff
@@ -0,0 +1,26 @@
+class MigrateEmailUserOptions < ActiveRecord::Migration[5.2]
+  def up
+    add_column :user_options, :email_level, :integer, default: UserOption.email_level_types[:only_when_away], null: false
+    add_column :user_options, :email_messages_level, :integer, default: UserOption.email_level_types[:always], null: false
+
+    execute <<~SQL
+      UPDATE user_options
+      SET email_level = CASE
+        WHEN email_direct IS TRUE AND email_always IS TRUE
```
eviltrout: In SQL you don't need `IS TRUE`, you can do a comparison like `WHEN email_direct AND (NOT email_always)`. Just a little shorter :)

db/migrate/20190313134642_migrate_default_user_email_options.rb
```diff
@@ -0,0 +1,31 @@
+class MigrateDefaultUserEmailOptions < ActiveRecord::Migration[5.2]
+  def up
+    email_always = DB.query_single("SELECT value FROM site_settings WHERE name = 'default_email_always'").first
+    email_direct = DB.query_single("SELECT value FROM site_settings WHERE name = 'default_email_direct'").first
+    email_personal_messages = DB.query_single("SELECT value FROM site_settings WHERE name = 'default_email_personal_messages'").first
+
+    default_email_level = nil
+    default_email_level = UserOption.email_level_types[:always] if email_direct != 'f' && email_always == 't'
```
eviltrout: This is one of those cases where enums are not great. 

It is *probably* safe but we've found if an enum changes in the future (say a key is deleted) the migration breaks. It's counter intuitive but it's probably safer here if you hard code the values.

SamSaffron:
If @eviltrout and @ZogStriP  are good I say merge it in first thing in your morning, then make sure you deploy it by hand to at least one multisite and watch that no crazy ensues. I really like the change a lot!

PR #7144, opened 2019-03-11
FEATURE: Add `IgnoredUsersSummary` daily job

khalilovcmded:
## Why?

This is part of the [Ability to ignore a user feature](https://meta.discourse.org/t/ability-to-ignore-a-user/110254/8), please note, this is a WIP.

We want to:

1. Send an automatic group PM that goes out to moderators
2. When {x} users have Ignored the same user, threshold defined by a site setting, default of 5
3. Only send this message every X days which is defined by another site setting

This job will run every 24 hours, we can re-think of that, and either make it triggered on the [UsersController#ignore](https://github.com/discourse/discourse/blob/da7406e567be10a5d1bcec0af3a11bea6ec6a85f/app/controllers/users_controller.rb#L998-L1005) action.

## UI

A_user_is_being_ignored_by_many_other_users_-_Discourse

## Things to consider

1. Since we just recently introduced the ignored user feature, I don't expect the first run of this job will create _many_ systems message to moderators all at once.
2. ~We are depending on the `post_custom_fields` table to ensure idempotency for system message creation.~
3. I thought we might need a mutex or distributed mutex to deal with race conditions but I think this is handled by the base job
4. I am waiting for the copy for the system message from @coding-horror 
5. ~I will be writing tests soon~


config/locales/server.en.yml
```diff
@@ -2819,6 +2819,18 @@ en:
       %{raw}
       ```
 
+    ignored_users_summary:
```
khalilovcmded: ⚠️ We will definitely change the copy.

config/site_settings.yml
```diff
@@ -532,6 +532,14 @@ users:
   ignore_user_enabled:
     default: false
     client: true
+  ignored_users_count_message_threshold:
+    default: 5
```
khalilovcmded: ⚠️ The defaults to be reviewed.

config/site_settings.yml
```diff
@@ -532,6 +532,14 @@ users:
...
+    client: true
+    min: 1
+  ignored_users_message_gap_days:
+    default: 365
```
khalilovcmded: ⚠️ The defaults to be reviewed.

eviltrout:
BTW Github has a draft pull request feature now, so I would prefer to create than an open one with a "DO NOT MERGE" tag.

khalilovcmded:
@eviltrout makes sense! Next time I shall do so 👍

config/locales/server.en.yml
```diff
@@ -2819,6 +2819,18 @@ en:
...
+      title: "Ignored User passed threshold"
+      subject_template: "New Ignored user passed threshold"
+      text_body_template: |
+        Hello,
+
+        This is an automated message from %{site_name} to inform you about a new ignored user who passed the number of ignores threshold which is: %{ignores_threshold}
```
```suggestion
        This is an automated message from %{site_name} to inform you about a potentially problematic user who has been ignored by %{ignores_threshold} users. It would be smart to review their forum activity.
```

config/locales/server.en.yml
```diff
@@ -2819,6 +2819,18 @@ en:
       %{raw}
       ```
 
+    ignored_users_summary:
+      title: "Ignored User passed threshold"
+      subject_template: "New Ignored user passed threshold"
```
```suggestion
      subject_template: "A user is being ignored by many other users "
```

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def execute(args)
+      return unless SiteSetting.ignore_user_enabled
+
+      user_ids = DB.query_single <<~SQL
+        SELECT u.id AS user_id
+        FROM users AS u
+        INNER JOIN ignored_users AS ig ON ig.ignored_user_id = u.id
+        GROUP BY u.id
+        HAVING COUNT(u.id) >= #{SiteSetting.ignored_users_count_message_threshold}
```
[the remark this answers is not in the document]
ZogStriP: As a general rule of thumb, never do string interpolation in SQL queries ;) Always use `:keyword` or `?` interpolation which are properly escaped.

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    private
+
+    def should_notify_user?(user)
+      custom_field = PostCustomField.where("name = 'summary_sent_for_ignored_user' AND value = ?", user.id.to_s).select(:created_at).first
+      custom_field.blank? || (custom_field.created_at - SiteSetting.ignored_users_message_gap_days.days) > Time.now.utc
+    end
+
+    def notify_user(user)
+      params = SystemMessage.new(User.last).defaults.merge({ignores_threshold: SiteSetting.ignored_users_count_message_threshold})
```
khalilovcmded: ⚠️ `User.last` is not correct.

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,45 @@
...
+        gap_days: SiteSetting.ignored_users_message_gap_days
+      }
+      user_ids = DB.query_single(<<~SQL, params)
+        SELECT ig.ignored_user_id AS user_id
+        FROM ignored_users AS ig
+        WHERE NOT EXISTS (SELECT 1
+                          FROM post_custom_fields as pcf
+                          WHERE pcf.name = 'summary_sent_for_ignored_user'
+                            AND pcf.value = user_id::text
+                            AND pcf.created_at < CURRENT_TIMESTAMP - ':gap_days DAYS'::INTERVAL
```
khalilovcmded: If it is 1 day, I believe this Postgres syntax: `1 DAYS` is still okay?

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,45 @@
+module Jobs
+  class IgnoredUsersSummary < Jobs::Scheduled
```
khalilovcmded: Still testing this, please give me sometime.

app/jobs/scheduled/ignored_users_summary.rb
ZogStriP: Only way to know for sure is to test it ;)

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'rails_helper'
+
+require_dependency 'jobs/scheduled/ignored_users_summary'
+
+describe Jobs::IgnoredUsersSummary do
+  before do
+    SiteSetting.ignore_user_enabled = true
+    SiteSetting.ignored_users_count_message_threshold = 1
+    SiteSetting.ignored_users_message_gap_days = 365
```
ZogStriP: You don't need that line if that's the default 😉

khalilovcmded: Yes, I intentionally wrote it for other Engineers reading the specs and deciphering how this works.

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  subject { Jobs::IgnoredUsersSummary.new.execute({}) }
+
+  context "when no ignored users" do
```
ZogStriP:
```suggestion
  context "with no ignored users" do
```

or 

```suggestion
  context "when there are no ignored users" do
```

?

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  context "when some ignored users exist" do
+    let(:tarek) { Fabricate(:user, username: "tarek") }
+    let(:matt) { Fabricate(:user, username: "matt") }
+    let(:john) { Fabricate(:user, username: "john") }
+
+    before do
+      Fabricate(:ignored_user, user_id: tarek.id, ignored_user_id: matt.id)
```
ZogStriP: I *think* you can be clever a do this instead

```suggestion
      Fabricate(:ignored_user, user: tarek, ignored_user: matt)
```

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    SiteSetting.ignored_users_count_message_threshold = 1
+    SiteSetting.ignored_users_message_gap_days = 365
+  end
+
+  subject { Jobs::IgnoredUsersSummary.new.execute({}) }
+
+  context "when no ignored users" do
+    it "does nothing" do
+      subject
+      expect(Post.count).to eq(0)
```
ZogStriP: This should test that you're (not) creating a system message to the moderators group instead of "just" a post.

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        it "does nothing" do
+          subject
+          expect(Post.count).to eq(0)
+        end
+      end
+
+      context "when threshold is hit" do
+        before do
+          SiteSetting.ignored_users_count_message_threshold = 1
```
ZogStriP: I don't think that line is necessary since you're already setting it to `1` for these tests.

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    SiteSetting.ignore_user_enabled = true
+    SiteSetting.ignored_users_count_message_threshold = 1
+    SiteSetting.ignored_users_message_gap_days = 365
+  end
+
+  subject { Jobs::IgnoredUsersSummary.new.execute({}) }
+
+  context "with no ignored users" do
+    it "does nothing" do
+      subject
```
ZogStriP: Might be better to do something like

``` ruby
expect { subject }.to_not change(Post.count)
```

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    context "when no system message exists for the ignored users" do
+      context "when threshold is not hit" do
+        before do
+          SiteSetting.ignored_users_count_message_threshold = 5
+        end
+
+        it "does nothing" do
+          subject
```
ZogStriP: I think here, we can also do

``` ruby
expect { subject }.to_not change { Post.count }
```

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    context "when a system message already exists for the ignored users" do
+      context "when threshold is not hit" do
+        before do
+          SiteSetting.ignored_users_count_message_threshold = 5
+        end
+
+        it "does nothing" do
+          subject
```
ZogStriP: Same here,

``` ruby
expect { subject }.to_not change { Post.count }
```

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+      end
+
+      context "when threshold is hit" do
+        before do
+          Fabricate(:post_custom_field, name: "summary_sent_for_ignored_user", value: matt.id.to_s)
+          Fabricate(:post_custom_field, name: "summary_sent_for_ignored_user", value: john.id.to_s)
+        end
+
+        it "does nothing" do
```
ZogStriP: Or does it? 😉

khalilovcmded: It actually does nothing 😃 The `expect(Post.count).to eq(2)` is because the `Fabricate` above creates them part of them relationship.

spec/jobs/ignored_users_summary_spec.rb
khalilovcmded: Done

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,45 @@
...
+      return unless SiteSetting.ignore_user_enabled
+
+      params = {
+        threshold: SiteSetting.ignored_users_count_message_threshold,
+        gap_days: SiteSetting.ignored_users_message_gap_days,
+      }
+      user_ids = DB.query_single(<<~SQL, params)
+        SELECT ig.ignored_user_id AS user_id
+        FROM ignored_users AS ig
+        WHERE NOT EXISTS (SELECT 1
```
eviltrout: I just noticed something quite odd here. There's no `post_id` restriction in your subquery.

It seems like you are using a custom field on the `Post` to avoid notifying moderators twice about the same user? If anything, wouldn't you want to add the custom field on the user rather than the post?

I am also not a huge fan of using Custom Fields in the core application to track data, as you can see the query is a little clumsy. Is there a reason you didn't add a field to `IgnoredUser` such as `summarized_at`?

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    private
+
+    def notify_user(user)
+      params = SystemMessage.new(user).defaults.merge({ ignores_threshold: SiteSetting.ignored_users_count_message_threshold })
+      title = I18n.t("system_messages.ignored_users_summary.subject_template")
+      raw = I18n.t("system_messages.ignored_users_summary.text_body_template", params)
+
+      PostCreator.create(
```
tgxworld:
```suggestion
      PostCreator.create!(
```

spec/jobs/ignored_users_summary_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    SiteSetting.ignore_user_enabled = true
+    SiteSetting.ignored_users_count_message_threshold = 1
+    SiteSetting.ignored_users_message_gap_days = 365
+  end
+
+  subject { Jobs::IgnoredUsersSummary.new.execute({}) }
+
+  context "with no ignored users" do
+    it "does nothing" do
+      subject
```
tgxworld: I think we're executing the job twice here.

app/jobs/scheduled/ignored_users_summary.rb
khalilovcmded: Thanks Robin for the feedback 🙌 

> I just noticed something quite odd here. There's no post_id restriction in your subquery.

Since we are depending on the `post_custom_fields` table to ensure idempotency for system message creation, and the idempotency key here being: `'summary_sent_for_ignored_user' + 'user_id' + 'created_at'`, can you elaborate more on why do we need `post_id`? Or rather will there be a data integrity issue if we didn't bind it by `post_id`? 

> It seems like you are using a custom field on the Post to avoid notifying moderators twice about the same user? 

Yes 👍 

> If anything, wouldn't you want to add the custom field on the user rather than the post?

I think that's cleaner, or rather the relationship is more straight forward. I have no strong opinions of going with one of the two approaches, I think it yields the same result for us, one more rational than the other.

> I am also not a huge fan of using Custom Fields in the core application to track data

Yes, I didn't feel super good about it too, we had few choices being:

1. Find a way to track this data by maybe introducing a generic table for tracking idempotent operations, use that table for this and everything else of the same nature in the future.
2. Build upon and use an already existing persistence model (aka: `PostCustomField` or `UserCustomField` tables).
3. Use volatile storage medium (Redis) but I feel that's unreliable.

Perhaps you have other options?

> Is there a reason you didn't add a field to IgnoredUser such as summarized_at?

The relationship is as: `User --> has many --> Ignored Users`, if I add a `summarised_at` on the `IgnoredUser: X`, for example, I need to add that on all `IgnoredUser: X` records, right? Or how would that work?

I think we chose the `PostCustomField` because we know that one Post (System message) is going to ever have one unique combination of:  `'summary_sent_for_ignored_user' + 'user_id' + 'created_at'`.

khalilovcmded: To summarise the above and move this forward,

1. Should we go with the `UserCustomField`?
2. Any other possible or cleaner options than ensuring idempotency through `*CustomField` tables?

eviltrout: > can you elaborate more on why do we need post_id

To be clear, you don't for your code to work. But it indicates to me that the fields are not `Post` specific, they are user specific. That's why I brought it up :)

(BTW while speaking of idempotent operations, custom fields is not great. It has known issues where you can insert the same value twice concurrently and Discourse treats it as an array. We are going to fix this.) 

> The relationship is as: User --> has many --> Ignored Users, if I add a summarised_at on the IgnoredUser: X, for example, I need to add that on all IgnoredUser: X records, right? Or how would that work?

Yes, let's say you added a column `summarized_at DATETIME` to the `ignored_users` table. When a summary is sent for that user you can update it atomically:

`IgnoredUser.where(ignored_user_id: user_id).update_all(summarized_at: Time.zone.now)`

And your query is a lot simpler now

```sql
SELECT ig.ignored_user_id AS user_id
FROM ignored_users AS ig
WHERE (ig.created_at + ':gap_days DAYS'::INTERVAL) > CURRENT_TIMESTAMP
GROUP BY ig.ignored_user_id
HAVING COUNT(*) > :threshold
```

The minor downside of this design is the extra column, but it's significantly less overhead than using the user custom fields table.

ZogStriP: I think the query should be 

``` sql
SELECT ignored_user_id AS user_id
FROM ignored_users
WHERE (summarized_at + ':gap_days DAYS'::INTERVAL) > CURRENT_TIMESTAMP
GROUP BY ignored_user_id
HAVING COUNT(*) > :threshold
```

I changed the `where` condition from using `created_at` to using `summarized_at` since we only want the user_ids of users that haven't already been "summarized" in the past X days.

eviltrout: Yes oops, I meant `summarized_at` (wrote this from memory.)

khalilovcmded: I love this solution ❤️ , though this will fail if:

1. A User has been ignored by 5 other users (passing threshold)
2. We send a System message, and do the `update_all`
3. The same User has been un-ignored from the same 5 other users before the `gap_days` end
4. With every un-ignore, we delete the matching `ignored_users` db record.
5. Now, if the user to be ignored _yet again_ by 5 other users, we will send a System message which we aren't supposed to sent (because we have no records in the `ignored_user` table to give us a hint).

I wouldn't say it is a big deal because of the rare possibility 😃 but it is a scenario that can happen.

ZogStriP: We could soft delete the records instead of hard deleting them ;) That way we'd keep the `summarized_at`. Another solution would be to have a new table for summarized users.

khalilovcmded: I thought of soft delete, but didn't want to mention it 😅

eviltrout: Nice catch, but I agree it seems so unlikely. Let's go with this, and if in the future we get any reports of this will split it off to be a new table.

khalilovcmded: Let's do it 🙌

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,42 @@
...
```
khalilovcmded: ℹ️  I was tempted to wrap these two statements into a transaction, but then I thought if the Post creation, then it will all fail together.

app/jobs/scheduled/ignored_users_summary.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      params = {
+        threshold: SiteSetting.ignored_users_count_message_threshold,
+        gap_days: SiteSetting.ignored_users_message_gap_days,
+        coalesced_gap_days: SiteSetting.ignored_users_message_gap_days + 1,
+      }
+      user_ids = DB.query_single(<<~SQL, params)
+        SELECT ignored_user_id
+        FROM ignored_users
+        WHERE COALESCE(summarized_at, CURRENT_TIMESTAMP + ':coalesced_gap_days DAYS'::INTERVAL) - ':gap_days DAYS'::INTERVAL > CURRENT_TIMESTAMP
```
khalilovcmded: `COALESCE` is to account for `summarized_at` being null.

config/locales/server.en.yml
```diff
@@ -1823,6 +1823,10 @@ en:
 
     ignore_user_enabled: "[Beta] Allow ignoring users."
 
+    ignored_users_count_message_threshold: "[Beta] Count of user ignores before sending a moderator notification."
```
khalilovcmded: ⚠️ Please review this copy.

config/locales/server.en.yml
```diff
@@ -1823,6 +1823,10 @@ en:
...
+
+    ignored_users_message_gap_days: "[Beta] Number of gap days between sending user ignores moderator notifications."
```
khalilovcmded: ⚠️ Please review this copy.

config/locales/server.en.yml
```diff
@@ -2829,6 +2829,18 @@ en:
       %{raw}
       ```
 
+    ignored_users_summary:
+      title: "Ignored User passed threshold"
+      subject_template: "A user is being ignored by many other users "
```
```suggestion
      subject_template: "A user is being ignored by many other users"
```

config/locales/server.en.yml
```suggestion
    ignored_users_count_message_threshold: "Notify moderators if a particular user is ignored by this many other users."
```

config/locales/server.en.yml
```suggestion
    ignored_users_message_gap_days: "How long wait before notifying moderators again about a user who has been ignored by many others"
```

(This one is super hard to describe, I did my best.)

PR #7145, opened 2019-03-11
FIX: Text content search warning if more than 50 results

venarius:
This adds a warning to site texts search if more than 50 results were found.

Resolves: https://meta.discourse.org/t/customize-text-content-search-results-missing/65327


eviltrout:
Thanks!

PR #7127, opened 2019-03-07
UX: Added tooltips to topic admin menu

venarius:
Added tooltips to the topic admin menu.

Resolves: https://meta.discourse.org/t/tooltips-for-topic-admin-actions/96086

PR #7146, opened 2019-03-11
UX: Added tooltips to topic admin menu

venarius:
Added tooltips to the topic admin menu.

Resolves: https://meta.discourse.org/t/tooltips-for-topic-admin-actions/96086

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1932,7 +1932,7 @@ en:
         close: "Close Topic"
         close_tooltip: "Marks the topic as closed and prevents new replies."
         multi_select: "Select Posts…"
-        multi_select_tooltip: "Toggles multi select mode in which multiple posts can be edited."
+        multi_select_tooltip: "Toggles multiselect mode in which multiple posts can be selected."
         timed_update: "Set Topic Timer..."
         timed_update_tooltip: "Opens a new window in which a timer for different actions can be set."
         pin: "Pin Topic…"
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1924,21 +1924,36 @@ en:
 
       actions:
         recover: "Un-Delete Topic"
+        recover_tooltip: "Restores the topic."
         delete: "Delete Topic"
+        delete_tooltip: "Marks the topic as deleted."
         open: "Open Topic"
+        open_tooltip: "Marks the topic as open and allows new replies."
         close: "Close Topic"
+        close_tooltip: "Marks the topic as closed and prevents new replies."
         multi_select: "Select Posts…"
+        multi_select_tooltip: "Toggles multi select mode in which multiple posts can be edited."
         timed_update: "Set Topic Timer..."
+        timed_update_tooltip: "Opens a new window in which a timer for different actions can be set."
         pin: "Pin Topic…"
+        pin_tooltip: "Opens a window in which the topic can be featured in different ways."
         unpin: "Un-Pin Topic…"
+        unpin_tooltip: "Unpins the topic and it will no longer be featured."
         unarchive: "Unarchive Topic"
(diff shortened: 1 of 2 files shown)


config/locales/client.en.yml
```diff
@@ -1924,21 +1924,36 @@ en:
...
         recover: "Un-Delete Topic"
+        recover_tooltip: "Restores the topic."
         delete: "Delete Topic"
+        delete_tooltip: "Marks the topic as deleted."
         open: "Open Topic"
+        open_tooltip: "Marks the topic as open and allows new replies."
         close: "Close Topic"
+        close_tooltip: "Marks the topic as closed and prevents new replies."
         multi_select: "Select Posts…"
+        multi_select_tooltip: "Toggles multi select mode in which multiple posts can be edited."
```
tgxworld:
```suggestion
        multi_select_tooltip: "Toggles multi select mode in which multiple posts can be selected."
```

tgxworld:
The code looks good to me. @coding-horror Do you have any thoughts about the copy? It can feel alittle verbose for some.

config/locales/client.en.yml
```diff
@@ -1924,21 +1924,36 @@ en:
...
         recover: "Un-Delete Topic"
+        recover_tooltip: "Restores the topic."
         delete: "Delete Topic"
+        delete_tooltip: "Marks the topic as deleted."
         open: "Open Topic"
+        open_tooltip: "Marks the topic as open and allows new replies."
         close: "Close Topic"
+        close_tooltip: "Marks the topic as closed and prevents new replies."
         multi_select: "Select Posts…"
+        multi_select_tooltip: "Toggles multiselect mode in which multiple posts can be selected."
```
ZogStriP: I don't think we ever mention this as the "multiselect mode" anywhere else.

venarius: I just took the same name the locale variable has. But I think, "multiselect mode" is pretty easy to understand :) Other ideas on how to call it?

SamSaffron:
I am merging this for now, @coding-horror  can follow up to improve copy! thanks @venarius !

coding-horror:
I don't think this should be merged @sam it's a LOT of extra copy that has to be translated and most of it adds little value over the words that are already there... My vote is that this gets rolled back.

SamSaffron:
I will roll back for now ... but some of the tooltips are handy

1. What does "archive" do? hover on archive to figure out
2. What does "reset bump" do? 
3. What does "unlisting" do? 

I think the failure here is that the change was too far reaching.

pgorod:
I'm on my honeymoon with Discourse, I'm a new admin, and I went looking for this on the Feature Requests, and it brought me here.

It's true that these tooltips have to be translated and that they will rarely be seen, but I don't feel it's true that they don't bring much value. 

My rationale is this: Topic admin actions are things with consequences - you can annoy users, you can make content disappear, you can generate email notifications. I feel the need for extra security before clicking these. Some of these concepts are new to me.

I would vote (recommend? ask?) that this (or something like this) actually gets into the product. I would focus the copy on a logic of "explain consequences, if any, especially if they are irreversible or impact others; provide security to a noobie admin operating this menu".

ZogStriP:
@pgorod if you want to talk about Discourse, highly recommended that you head over https://meta.discourse.org 😉

pgorod:
Ok. But that's where I came from :-) I searched "Feature requests", then I found [somebody had already asked for it](https://meta.discourse.org/t/tooltips-for-topic-admin-actions/96086), then there was [a PR](https://github.com/discourse/discourse/pull/7127) (closed) then there was this PR (reverted). I thought it was better to post here because the latest developments are only here, not on those other places. 

But ok, I get the nudge, I will explain he latest developments on that Meta thread and post there. Thanks.

PR #7148, opened 2019-03-12
FEATURE: Add more control over post notices.

udan11:


app/serializers/post_serializer.rb
```diff
@@ -370,7 +370,7 @@ def post_notice_type
   end
 
   def include_post_notice_type?
-    return false if scope.user&.id == object.user_id || !scope.user&.has_trust_level?(TrustLevel[2])
+    return false if scope.user&.id == object.user_id || !scope.user&.has_trust_level?(SiteSetting.min_post_notice_tl)
```
eviltrout: Rather than using the safe navigation operator twice I think it would be nicer to have `scope.user.blank?` as a first short-circuit.

udan11: Resolved in 476d0050ab80b2dfcaee6368c529f5094f14d79a.

PR #7149, opened 2019-03-12
UX: make name optional for confirmation user field

majakomel:


app/models/user_field.rb
```diff
@@ -2,7 +2,8 @@ class UserField < ActiveRecord::Base
 
   include AnonCacheInvalidator
 
-  validates_presence_of :name, :description, :field_type
+  validates_presence_of :description, :field_type
+  validates_presence_of :name, unless: -> { field_type == "confirm" }
```
ZogStriP: This should be tested.

SamSaffron:
I like how simple this change is, can you include a screenshot with all the permutations now?

majakomel:
I left `*` next to confirmation field name when there is one


Screen Shot 2019-03-13 at 17 55 53

PR #7151, opened 2019-03-12
FIX: invite approval `StaffActionLogger` bug

khalilovcmded:
## Why?

We send `approved_by` to the `StaffActionLogger` which sometimes can be an integer, and that blows it up during its construction.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -410,6 +410,7 @@ class User < ActiveRecord::Base
 
     if approved_by.is_a?(Integer)
       self.approved_by_id = approved_by
+      self.approved_by = User.find(approved_by)
     else
       self.approved_by = approved_by
     end
@@ -421,7 +422,7 @@ class User < ActiveRecord::Base
       DiscourseEvent.trigger(:user_approved, self)
     end
 
-    StaffActionLogger.new(approved_by).log_user_approve(self)
+    StaffActionLogger.new(self.approved_by).log_user_approve(self)
 
     result
   end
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -410,6 +410,7 @@ def approve(approved_by, send_mail = true)
 
     if approved_by.is_a?(Integer)
       self.approved_by_id = approved_by
+      self.approved_by = User.find(approved_by)
```
tgxworld: Hmm I'm not sure if this line is required here. Since the bug is due to using a variable that is named the same as a column in the table, perhaps the variable should be renamed? Maybe `approver`?

khalilovcmded: > Since the bug is due to using a variable that is named the same as a column in the table

Yes, that's a good point.

Let's take a step back, and look at why the argument `approved_by` is not ALWAYS a User object.

discourse____discourse_discourse__-_____app_models_user_rb__discourse_

It is only because of the `invite_redeemer.rb`, I think we can:

1. Change the argument name to `approver` which should always be a User object
2. Clean this up a bit here.

But let's first fix this bug now, and I can refactor. WDYT?

tgxworld: hehe I would just go for the refactor and remove the nightmare where a variable can be an id or an ActiveRecord object.

khalilovcmded: 😂 Let's do it. 🏃 🏃 🏃

khalilovcmded: @tgxworld How about now? 👀

tgxworld:
Looks good :+1:

PR #7152, opened 2019-03-12
FEATURE: Add site contact group.

nbianca:


lib/validators/group_setting_validator.rb
```diff
@@ -0,0 +1,21 @@
+class GroupSettingValidator
```
tgxworld: It'll be good if we can add a test for this too.

lib/validators/group_setting_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  include RegexSettingValidation
+
+  def initialize(opts = {})
+    @opts = opts
+    initialize_regex_opts(opts)
+  end
+
+  def valid_value?(val)
+    !val.present? || (Group.where(name: val).exists? && regex_match?(val))
```
ZogStriP: `val.blank?` is easier to grasp than `!val.present?`

Also, you can shorten the `exists?` call ;)

```suggestion
    val.blank? || (Group.exists?(name: val) && regex_match?(val))
```

SamSaffron:
Oh there is an edge case here that concerns me a bit. 

What if `site_contact_group_name` is valid and saved, later on someone renames the group... 

Last thing we want is for all system messages to be rejected! I think we need to make sure we always hand post creator a valid group name at the time of post creation to avoid weirdness.

PR #7153, opened 2019-03-12
FEATURE: Add `Top Ignored Users` report

khalilovcmded:
## Why?

This is part of the [Ability to ignore a user feature](https://meta.discourse.org/t/ability-to-ignore-a-user/110254/8), and also building towards [this PR](https://github.com/discourse/discourse/pull/7144).

We want to send a System Message daily when a specific count threshold for an ignored is reached. To make this system message informative, we want to link to a report for the Top Ignored Users too.

## UI

Admin_-_Discourse

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1573,27 +1573,26 @@ class Report
     report.data = []
 
     sql = <<~SQL
-    SELECT
-    u.id as user_id,
-    u.username,
-    u.uploaded_avatar_id,
-    COUNT(*) as ignores_count
-    FROM users AS u
-    INNER JOIN ignored_users AS ig ON ig.ignored_user_id = u.id
-    WHERE ig.created_at >= '#{report.start_date}' AND ig.created_at <= '#{report.end_date}'
-    GROUP BY u.id
-    ORDER BY COUNT(*) DESC
-    LIMIT #{report.limit || 250}
+      SELECT
+      u.id as user_id,
+      u.username,
+      u.uploaded_avatar_id,
+      COUNT(*) as ignores_count
+      FROM users AS u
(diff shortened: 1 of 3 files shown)


app/models/report.rb
```diff
@@ -1548,6 +1548,55 @@ def self.report_top_uploads(report)
...
+    FROM users AS u
+    INNER JOIN ignored_users AS ig ON ig.ignored_user_id = u.id
+    WHERE ig.created_at >= '#{report.start_date}' AND ig.created_at <= '#{report.end_date}'
+    GROUP BY u.id
+    ORDER BY COUNT(*) DESC
+    LIMIT #{report.limit || 250}
+    SQL
+
+    DB.query(sql).each do |row|
+      data = {}
```
ZogStriP: You can do all that in "operation"

``` ruby
report.data << {
  ignored_user_id: row.user_id,
  ignored_username: row.username,
  # ...
}
```

app/models/report.rb
```diff
@@ -1548,6 +1548,55 @@ def self.report_top_uploads(report)
...
+          :ignores_count,
+        ],
+        title: I18n.t("reports.top_ignored_users.labels.ignores_count")
+      }
+    ]
+
+    report.data = []
+
+    sql = <<~SQL
+    SELECT
```
ZogStriP: This would look better if the SQL query was indented ▶️

eviltrout: Agreed, and this comment is the most trivial thing I can ever think of bringing up, but `as` should be capitalised since those are SQL keywords we generally capitalise all of them.

khalilovcmded: I think so 👍  I am unsure why do we have the indentation in all the other report methods this same way though.

PR #7154, opened 2019-03-12
remove this.get example

jjaffeux:


app/assets/javascripts/discourse/components/d-button.js.es6
```diff
@@ -19,36 +19,33 @@ export default Ember.Component.extend({
   btnIcon: Ember.computed.notEmpty("icon"),
 
   @computed("icon", "translatedLabel")
-  btnType(icon, translatedLabel) {
```
SamSaffron: I dislike this particular example cause you are arguably adding code here with all the `this.` and its slightly less clear.

app/assets/javascripts/discourse/components/d-button.js.es6
```diff
@@ -19,36 +19,33 @@ export default Ember.Component.extend({
...
-      this.sendAction("action", this.get("actionParam"));
-    } else if (
-      typeof this.get("action") === "object" &&
-      this.get("action").value
-    ) {
-      this.get("action").value(this.get("actionParam"));
-    } else if (typeof this.get("action") === "function") {
-      this.get("action")(this.get("actionParam"));
+    if (typeof this.action === "string") {
+      this.sendAction("action", this.actionParam);
```
SamSaffron: this though, I like a lot and we should use it.

PR #7156, opened 2019-03-12
FIX: Better emoji escaping for topic title

venarius:
This fixes a problem with escaping unicode emojis that consist of multiple combined emojis. Also enhanced `performEmojiUnescape` to now support unicode emojis.

Resolves: https://meta.discourse.org/t/unicode-emoji-modifiers-are-handled-differently-in-titles-and-post-content/111328

### DIFF
--- a/app/assets/javascripts/pretty-text-bundle.js
+++ b/app/assets/javascripts/pretty-text-bundle.js
@@ -2,6 +2,7 @@
 //= require ./pretty-text/guid
 //= require ./pretty-text/censored-words
 //= require ./pretty-text/emoji/data
+//= require ./pretty-text/emoji/version
 //= require ./pretty-text/emoji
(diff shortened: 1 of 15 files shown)


SamSaffron:
I am good with merge this @jjaffeux  feel free to. 

One thing bugs me here though... instead of `emoji.js.es6.erb` it would be nice to have

1. `emoji.js.es6` AND
2. `emoji_version.js.es6.erb` 

Then we erb file only had a boring `Emoj::EMOJI_VERSION` and exports that, we import it into `emoji.js.es6` and we can pass this rather big file through prettier which is wins all around.

venarius:
Then wait a bit before merging this. I will split the files.

jjaffeux:
Yes LGTM merging. Thanks @venarius

PR #7158, opened 2019-03-12
DEV: Allow topics to register default topic order criteria.

udan11:


SamSaffron:
@udan11 prettier needs to be appeased, but feel free to merge once that is done.

PR #7161, opened 2019-03-13
UX: Enabled pinch to zoom for lightbox

venarius:
This replaces the meta viewport tag while looking at a lightbox. This enables pinch to zoom on various devices.

Resolves: https://meta.discourse.org/t/adjusting-image-zoom-level-on-mobile-devices/111440

Wanted to write Qunit tests for it but the box for component tests has no meta tags.

### DIFF
--- a/app/assets/javascripts/discourse/lib/lightbox.js.es6
+++ b/app/assets/javascripts/discourse/lib/lightbox.js.es6
@@ -6,7 +6,7 @@ export default function($elem) {
   if (!$elem) {
     return;
   }
-  const original_meta = $("meta[name=viewport]");
+  const originalMeta = $("meta[name=viewport]").attr("content");
   loadScript("/javascripts/jquery.magnific-popup.min.js").then(function() {
     const spoilers = $elem.find(".spoiler a.lightbox, .spoiled a.lightbox");
     $elem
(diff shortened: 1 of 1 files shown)


venarius:
CI frontend tests timed out... Someone to restart them?

app/assets/javascripts/discourse/lib/lightbox.js.es6
```diff
@@ -6,6 +6,7 @@ export default function($elem) {
   if (!$elem) {
     return;
   }
+  const original_meta = $("meta[name=viewport]");
```
eviltrout: Rather than replacing the entire tag, couldn't we replace the `content` attribute?

Additionally javascript variables in our codebase should be `camelCase`.

venarius: If I would only go about changing the content attribute of the meta tag, the selector would no long be stored in `originalMeta`. I think this would decrease performance then because it has to select `$("meta[name=viewport]")` on more time.

eviltrout: I think the performance of finding a meta tag is negligible compared to the memory churn of replacing a DOM node.

venarius: Good to know. Thanks! :)

eviltrout:
Thanks!

PR #7162, opened 2019-03-13
FEATURE: Skip sending emails to domains on the `.invalid` TLD

davidtaylorhq:
This is a reserved TLD which we use when importing users without an email address. https://tools.ietf.org/html/rfc2606

### DIFF
--- a/spec/components/email/sender_spec.rb
+++ b/spec/components/email/sender_spec.rb
@@ -58,6 +58,13 @@ describe Email::Sender do
     Email::Sender.new(message, :hello).send
   end
 
+  it "doesn't deliver when the to address uses the .invalid tld" do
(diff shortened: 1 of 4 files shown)


spec/components/email/sender_spec.rb
```diff
@@ -58,6 +58,12 @@
     Email::Sender.new(message, :hello).send
   end
 
+  it "doesn't deliver when the to address uses the .invalid tld" do
+    message = Mail::Message.new(body: 'hello', to: 'myemail@example.invalid')
+    message.expects(:deliver_now).never
+    Email::Sender.new(message, :hello).send
```
ZogStriP: I think we should also expect that a `SkippedEmailLog` record was created with the reason `:sender_message_to_invalid`

PR #7163, opened 2019-03-13
FEATURE: Fetch email from auth provider if current user email is invalid

davidtaylorhq:
If the existing email address for a user ends in `.invalid`, we should take the email address from an authentication payload, and replace the invalid address.

This commit also adds some extensibility so that plugin authenticators can define `always_update_user_email?`


davidtaylorhq:
It happens when data is imported from a site which does not store email addresses. This PR allows users to "reclaim" their accounts, if we have a social auth association for them.

PR #7166, opened 2019-03-13
UX: order categories based on recent activity when using categories_and_latest_topics layout

majakomel:
https://meta.discourse.org/t/category-order-by-activity/107905

### DIFF
--- a/app/models/category_list.rb
+++ b/app/models/category_list.rb
@@ -77,6 +77,8 @@ class CategoryList
 
     if SiteSetting.fixed_category_positions
       @categories = @categories.order(:position, :id)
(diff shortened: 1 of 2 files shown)


app/models/category_list.rb
```diff
@@ -77,6 +77,8 @@ def find_categories
 
     if SiteSetting.fixed_category_positions
       @categories = @categories.order(:position, :id)
+    elsif !SiteSetting.fixed_category_positions && SiteSetting.desktop_category_page_style == "categories_and_latest_topics"
+      @categories = @categories.includes(:latest_post).order("posts.created_at DESC NULLS LAST").order('categories.id ASC')
```
eviltrout: TIL about `NULLS LAST`

SamSaffron:
only slight concern I have here is perf, but lets try it out, nice fix!

PR #7167, opened 2019-03-13
FEATURE: custom colors for default letter avatars

majakomel:
https://meta.discourse.org/t/custom-colors-for-default-letter-avatar-generator/111414

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -738,8 +738,21 @@ class User < ActiveRecord::Base
 
   def self.letter_avatar_color(username)
     username ||= ""
-    color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
-    color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    if SiteSetting.restrict_letter_avatar_colors.present?
+      hex_length = 6
+      colors = SiteSetting.restrict_letter_avatar_colors
+      length = colors.count("|") + 1
+      num = color_index(username, length)
+      index = (num * hex_length) + num
+      colors[index, hex_length]
+    else
+      color = LetterAvatar::COLORS[color_index(username, LetterAvatar::COLORS.length)]
+      color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    end
+  end
+
+  def self.color_index(username, length)
+    Digest::MD5.hexdigest(username)[0...15].to_i(16) % length
   end
 
   def avatar_template
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -1100,6 +1100,11 @@ files:
     client: true
     regex: '^((https?:)?\/)?\/.+[^\/]'
     shadowed_by_global: true
+  restrict_letter_avatar_colors:
+    default: ''
+    type: list
+    list_type: compact
+    validator: "ColorListValidator"
   selectable_avatars_enabled:
     default: false
     client: true
--- a/lib/validators/color_list_validator.rb
+++ b/lib/validators/color_list_validator.rb
@@ -1,0 +1,14 @@
+class ColorListValidator
+  def initialize(opts = {})
+    @opts = opts
+  end
+
+  def valid_value?(value)
+    hex_regex = /\A\h{6}\z/
+    value.split("|").all? { |c| c =~ hex_regex }
+  end
+
+  def error_message
+    I18n.t('site_settings.errors.invalid_hex_value')
+  end
+end
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1569,6 +1569,7 @@ en:
 
     external_system_avatars_enabled: "Use external system avatars service."
     external_system_avatars_url: "URL of the external system avatars service. Allowed substitutions are {username} {first_letter} {color} {size}"
+    restrict_letter_avatar_colors: "A list of 6-digit hexadecimal color values to be used for letter avatar background."
 
     selectable_avatars_enabled: "Force users to choose an avatar from the list."
     selectable_avatars: "List of avatars users can choose from."
@@ -2019,6 +2020,7 @@ en:
       min_username_length_range: "You cannot set the minimum above the maximum."
       max_username_length_exists: "You cannot set the maximum username length below the longest username (%{username})."
       max_username_length_range: "You cannot set the maximum below the minimum."
+      invalid_hex_value: "Color values have to be 6-digit hexadecimal codes."
 
     placeholder:
       sso_provider_secrets:
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1138,6 +1138,20 @@ describe User do
 
   end
 
+  describe "#letter_avatar_color" do
+    before do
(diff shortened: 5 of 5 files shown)


app/models/user.rb
```diff
@@ -738,8 +738,13 @@ def self.system_avatar_template(username)
 
   def self.letter_avatar_color(username)
     username ||= ""
-    color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
-    color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    if SiteSetting.restrict_letter_avatar_colors.present?
+      colors = SiteSetting.restrict_letter_avatar_colors.split("|")
+      color = colors[Digest::MD5.hexdigest(username)[0...15].to_i(16) % colors.length]
+    else  
+      color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
```
tgxworld: `Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length`

This line of logic is being duplicated which means we have two places to update if the logic should change in the future. I think it'll be better if we can refactor this into a method of its own so that we won't ever have a case where the logic will differ for the two different code paths.

config/site_settings.yml
```diff
@@ -1100,6 +1100,10 @@ files:
     client: true
     regex: '^((https?:)?\/)?\/.+[^\/]'
     shadowed_by_global: true
+  restrict_letter_avatar_colors:
```
tgxworld: Is there an expect format for the values here? Should it only be HEX codes?

majakomel: Yes, only 6-digit hex codes. I have added a validator for it.

app/models/user.rb
```diff
@@ -738,8 +738,17 @@ def self.system_avatar_template(username)
 
   def self.letter_avatar_color(username)
     username ||= ""
-    color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
-    color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    if SiteSetting.restrict_letter_avatar_colors.present?
```
ZogStriP: Since we're only change the list of available colors, here's how I would have written it 😉 

``` ruby
colors = SiteSetting.restrict_letter_avatar_colors.present? ? 
  SiteSetting.restrict_letter_avatar_colors.split("|") : 
  LetterAvatar::COLORS
color = select_color(username, colors)
color.map { |c| c.to_s(16).rjust(2, '0') }.join
```

lib/validators/color_list_validator.rb
```diff
@@ -0,0 +1,14 @@
+class ColorListValidator
+  def initialize(opts = {})
+    @opts = opts
+  end
+  
+  def valid_value?(value)
+    hex_regex = /\A[0-9a-fA-F]{6}\z/
```
ZogStriP: Ruby has a nice shortcut for hexadecimal characters 😉 

```suggestion
    hex_regex = /\A\h{6}\z/
```

lib/validators/color_list_validator.rb
```diff
@@ -0,0 +1,14 @@
...
+    value.split("|").all? {|c| c =~ hex_regex}
```
ZogStriP: Missing some whitespaces

```suggestion
    value.split("|").all? { |c| c =~ hex_regex }
```

app/models/user.rb
majakomel: `color.map` step is required only when using default colors to turn them into hex, so there is another check for `SiteSetting.restrict_letter_avatar_colors` needed anyway

app/models/user.rb
```diff
@@ -738,8 +738,16 @@ def self.system_avatar_template(username)
...
+    colors = SiteSetting.restrict_letter_avatar_colors.present? ?
+             SiteSetting.restrict_letter_avatar_colors.split("|") :
+             LetterAvatar::COLORS
+
+    select_color(username, colors)
+  end
+
+  def self.select_color(username, available_colors)
+    color = available_colors[Digest::MD5.hexdigest(username)[0...15].to_i(16) % available_colors.length]
+    SiteSetting.restrict_letter_avatar_colors.present? ? color : color.map { |c| c.to_s(16).rjust(2, '0') }.join
```
ZogStriP: That line feels weird. Having `select_color` reaching out to the SiteSetting feels odd. Plus, it means we'll be testing the presence of that site setting twice every time we need a letter avatar.

I like your first solution better (now that I understand the differences between the site setting and the COLORS array)

app/models/user.rb
```diff
@@ -738,8 +738,17 @@ def self.system_avatar_template(username)
 
   def self.letter_avatar_color(username)
     username ||= ""
-    color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
-    color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    if SiteSetting.restrict_letter_avatar_colors.present?
+      colors = SiteSetting.restrict_letter_avatar_colors.split("|")
```
SamSaffron: I think we should avoid the splitting here since we already know the value in site settings is valid. 

So.... `aabbcc|ccddee` has 2 colors cause it is (6+6+1) long

Then we decide on an index ... and do a substring from index * 6 + 1 for length 6. 

This means we don't have to keep splitting the string over and over for every user on the page.

app/models/user.rb
```diff
@@ -738,8 +738,21 @@ def self.system_avatar_template(username)
 
   def self.letter_avatar_color(username)
     username ||= ""
-    color = LetterAvatar::COLORS[Digest::MD5.hexdigest(username)[0...15].to_i(16) % LetterAvatar::COLORS.length]
-    color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    if SiteSetting.restrict_letter_avatar_colors.present?
+      hex_length = 6
+      colors = SiteSetting.restrict_letter_avatar_colors
+      length = colors.length / hex_length
```
ZogStriP: This won't work when there are more than 6 colors.

```suggestion
      length = colors.count("|") + 1
```

app/models/user.rb
```diff
@@ -738,8 +738,21 @@ def self.system_avatar_template(username)
...
+      color_order_nr = color_order_nr(username, length)
```
ZogStriP: We should avoid naming variables the same name as methods.

```suggestion
      num = color_order_nr(username, length)
```

app/models/user.rb
```diff
@@ -738,8 +738,21 @@ def self.system_avatar_template(username)
...
+      index = (color_order_nr * hex_length) + color_order_nr
+      colors[index, hex_length]
+    else
+      color = LetterAvatar::COLORS[color_order_nr(username, LetterAvatar::COLORS.length)]
+      color.map { |c| c.to_s(16).rjust(2, '0') }.join
+    end
+  end
+
+  def self.color_order_nr(username, length)
```
ZogStriP: Since we're basically computing an index, I think the name of the method should tell that

```suggestion
  def self.color_index(username, length)
```

PR #7168, opened 2019-03-14
FIX: Outsourced erb part from emoji.js.es6

venarius:
This is the first part of the "FIX: Better emoji escaping for topic title" in which the erb part is split from emoji.js.es6

More Information: https://github.com/discourse/discourse/commit/d32557ea326735c33a420dee7cc3de856c683dc7


SamSaffron:
excellent, thanks for this!

PR #7169, opened 2019-03-14
FIX: optimise querying Post's serializer ignored attribute

khalilovcmded:


khalilovcmded:
There is a failure in the JS tests that are not related to this PR:

```
Module Failed: multi-select
  Test Failed: with forceEscape
    Assertion Failed: You expected select-kit to be expanded but it is collapsed.    
    Assertion Failed: Error: You expected select-kit to be expanded but it is collapsed.
      Expected: true, Actual: false
```

I have seen this in only one another commit though I am not seeing failures on build, so I will merge.

khalilovcmded:
All good now 👍 Merging!

PR #7170, opened 2019-03-14
FIX: don't redirect incorrectly after full screen login

LeoMcA:
fixes two issues:
1. redirecting to an external origin's path after login
2. redirecting to the external origin after logout

https://meta.discourse.org/t/109755


SamSaffron:
@davidtaylorhq  can you review and merge?

davidtaylorhq:
This looks good to me, thanks @LeoMcA

PR #7171, opened 2019-03-14
FIX: Respect permalinks starting with "/category"

pmusaraj:
Adds a check for permalinks matching "category/path" before redirecting to "c/path". 

Useful when migrating to Discourse from other forum software that use "category/path" in their URL structure.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -499,6 +499,13 @@ class ApplicationController < ActionController::Base
     SecureSession.new(session["secure_session_id"] ||= SecureRandom.hex)
   end
 
+  def handle_permalink(path)
+    permalink = Permalink.find_by_url(path)
+    if permalink && permalink.target_url
+      redirect_to permalink.target_url, status: :moved_permanently
+    end
+  end
+
   private
 
   def check_readonly_mode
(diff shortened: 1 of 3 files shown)


SamSaffron:
structure wise I feel this should simply be implemented in AppicationController, so your code becomes:

```
  def redirect
    return if handle_permalink
    redirect_to path("/c/#{params[:path]}")
  end
```

I also wonder if https://github.com/discourse/discourse/blob/819d4facda2192b64dc18d874ff52c22ec157f71/app/controllers/tags_controller.rb#L385-L396 can then use the same helper ... I am confused at why it only works for permalinks to category_id seems ... odd. 

```
  def category_redirect_or_not_found
    # automatic redirects for renamed categories
    url = params[:parent_category] ? "c/#{params[:parent_category]}/#{params[:category]}" : "c/#{params[:category]}"
     
    if !handle_permalink(url)
      raise Discourse::NotFound
    end
  end
```

SamSaffron:
at least that way we have an easy way of adding this to other spots

pmusaraj:
Sure, I've added the helper in AppicationController. 

The tags controller redirect is for category tag URLs after a category gets renamed. So, `/tags/c/help/tag1` will redirect to `/tags/c/support/tag1` if the Help category slug has been renamed to `support`. That's why it checks for `category_id`, to make double sure that it's a category slug permalink since it's part of the new path.   

```
"#{Discourse::base_uri}/tags#{permalink.target_url}/#{params[:tag_id]}"
```

Since this doesn't redirect to `permalink.target_url`, it would be tricky to integrate with the ApplicationController helper. Maybe we should leave it as is?

PR #7172, opened 2019-03-14
FEATURE: Clean up inactive users.

nbianca:


app/jobs/scheduled/clean_up_inactive_users.rb
```diff
@@ -0,0 +1,28 @@
...
+  class CleanUpInactiveUsers < Jobs::Scheduled
+    every 1.day
+
+    def execute(args)
+      destroyer = UserDestroyer.new(Discourse.system_user)
+
+      User.joins("LEFT JOIN posts ON posts.user_id = users.id")
+        .where("posts.user_id IS NULL")
+        .where("users.last_seen_at < ?", 2.years.ago)
+        .where(trust_level: 0)
```
SamSaffron: I think this is probably ok, maybe add `where(last_posted_at: nil)` to speed up the query a tiny bit (you can still keep the join)

SamSaffron: `2.years.ago` should be a site setting, some sites may want this to be more aggressive than 2 years, others may want to disable it.

app/jobs/scheduled/clean_up_inactive_users.rb
```diff
@@ -0,0 +1,29 @@
+module Jobs
+
+  class CleanUpInactiveUsers < Jobs::Scheduled
+    every 1.day
+
+    def execute(args)
+      if SiteSetting.inactive_time_before_cleaning > 0
```
ZogStriP: We usually do

``` ruby
return if SiteSetting.inactive_time_before_cleaning <= 0
```

That way, we don't have to indent the whole code inside an `if`

app/jobs/scheduled/clean_up_inactive_users.rb
```diff
@@ -0,0 +1,29 @@
...
+        destroyer = UserDestroyer.new(Discourse.system_user)
+
+        User.joins("LEFT JOIN posts ON posts.user_id = users.id")
+          .where(last_posted_at: nil)
+          .where("posts.user_id IS NULL")
+          .where("users.last_seen_at < ?", SiteSetting.inactive_time_before_cleaning.months.ago)
```
ZogStriP: I think days is fine here. Name is hard, what about something like `clean_up_inactive_users_after_days`? (cc @eviltrout)

config/locales/server.en.yml
```diff
@@ -1828,6 +1828,8 @@ en:
 
     ignored_users_message_gap_days: "How long to wait before notifying moderators again about a user who has been ignored by many others."
 
+    clean_up_inactive_users_after_days: "Number of months before an inactive user (trust level 0 and without any posts) is removed."
```
ZogStriP:
```suggestion
    clean_up_inactive_users_after_days: "Number of days before an inactive user (trust level 0 without any posts) is removed."
```

spec/jobs/clean_up_inactive_users_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe Jobs::CleanUpInactiveUsers do
+  let(:user) { Fabricate(:user) }
```
ZogStriP: These should be moved inside the `it 'should clean up the user' do` test since they're not being used elsewhere

ZogStriP: Or at least inside the `context 'when user is inactive' do` block.

PR #7175, opened 2019-03-15
FIX: Topic filters persist when navigating to home page

vinothkannans:


app/assets/javascripts/discourse/controllers/discovery/topics.js.es6
```diff
@@ -25,6 +25,9 @@ const controllerOpts = {
 
   resetParams() {
     this.setProperties({ order: "default", ascending: false });
```
vinothkannans: This line may not required after this. I'm not sure.

vinothkannans: I can confirm it.. It was added for the same purpose https://github.com/discourse/discourse/commit/b2b7f4d90592a70a32c748130f0f82a732cd3090.

SamSaffron: yeah remove that line and go ahead and merge.

PR #7184, opened 2019-03-15
FEATURE: Display notification count in title for logged in users

davidtaylorhq:
This replaces the 'contextual' counters which we still use for anon user

### Existing behavior:

On topic list:
- Regardless of focus, browser title shows the count of topics waiting to be loaded (matches the "See N new or updated topic" banner)

On topic:
- IF tab is in the background, then increment the counter for each new post that arrives
- When focus is restored, reset the counter

### New behavior

For anon, same as above

For logged in users:
- If tab is in the background, **and** a new notification arrives while the tab is in the background, display the total notification count in the title
- Hide the count when the tab is in the foreground

cc @coding-horror

### DIFF
--- a/app/assets/javascripts/discourse.js.es6
+++ b/app/assets/javascripts/discourse.js.es6
@@ -41,7 +41,7 @@ const Discourse = Ember.Application.extend({
 
   Resolver: buildResolver("discourse"),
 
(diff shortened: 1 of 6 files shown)


coding-horror:
sounds good to me, as planned 🎉

PR #7187, opened 2019-03-15
FIX: Show 'emails disabled' to staff users when disabled for non-staff

davidtaylorhq:
At the moment, the "All outgoing email has been globally disabled by an administrator" banner doesn't show to staff users if SiteSetting.disable_emails === "non-staff". That means site admins might not realise emails are disabled until a user reports it.


coding-horror:
ohh dear important oversight, thanks for this!

PR #7190, opened 2019-03-16
FIX: Fixed reorder categories by number not working correctly

venarius:
Current Behavior:
Peek 2019-03-16 22-12

How it should be:
Peek 2019-03-16 22-13

Resolves: https://meta.discourse.org/t/reorder-categories-behaving-strangely/104190


eviltrout:
Thanks this looks good!

PR #7191, opened 2019-03-17
FIX: Fixed changing composer mode escaping title

venarius:
This fixes changing composer mode escaping the title.

Resolves: https://meta.discourse.org/t/toggling-shared-draft-causes-encoding-issue-when-title-has-quotes/111798

I removed the `escapeExpression` function from the `topicTitle` which was originally added by @techAPJ in commit c28c5083e0e7db96ec27f27c0107292e7469b214 to sanitize tags when the topic is created from url.

But similar to what tgxworld commented under the commit, I was also not able to inject html or js into the title using the url method. Tried a lot of variations and from my point of view its safe to remove the escape function.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -910,7 +910,7 @@ export default Ember.Controller.extend({
       opts.topicTitle &&
       opts.topicTitle.length <= this.siteSettings.max_topic_title_length
     ) {
-      this.set("model.title", escapeExpression(opts.topicTitle));
+      this.set("model.title", opts.topicTitle);
     }
 
     if (opts.topicCategoryId) {
--- a/test/javascripts/acceptance/composer-actions-test.js.es6
+++ b/test/javascripts/acceptance/composer-actions-test.js.es6
@@ -139,7 +139,10 @@ QUnit.test("shared draft", async assert => {
   await visit("/");
   await click("#create-topic");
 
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -910,7 +910,7 @@ export default Ember.Controller.extend({
       opts.topicTitle &&
       opts.topicTitle.length <= this.siteSettings.max_topic_title_length
     ) {
-      this.set("model.title", escapeExpression(opts.topicTitle));
```
ZogStriP: Any chances this could introduce an XSS? I think this was added in c28c5083e0e7db96ec27f27c0107292e7469b214

venarius: This is what I also wrote in the top. 

"But similar to what tgxworld commented under the commit, I was also not able to inject html or js into the title using the url method. Tried a lot of variations and from my point of view its safe to remove the escape function."

ZogStriP: My bad, went directly to the "Files changed" tab...

tgxworld:
@techAPJ Maybe you can have a look at this? I was unable to reproduce the XSS previously so we need to be careful here.

techAPJ:
This looks good. The [original commit](https://github.com/discourse/discourse/commit/c28c5083e0e7db96ec27f27c0107292e7469b214) was aimed at sanitizing tags, I am not sure why I sanitized title in that PR.

PR #7192, opened 2019-03-17
FIX: Fixed edit category description redirecting without saving

venarius:
When a user clicks on `Edit Description` in the edit category modal, he got redirected to the post without saving the edit on the category. This fix will open up a new tab when the user clicks on `Edit Description` and the edit category modal will stay open on the first page.

Resolves: https://meta.discourse.org/t/when-editing-a-category-clicking-edit-description-redirects-without-saving/111797


ZogStriP:
Thanks 👍

PR #7194, opened 2019-03-18
FEATURE: Ability to exclude category from search results.

tgxworld:
This commit also adds `Category#search_priority` which sets the ground
work to enable prioritizing of posts for certain categories when searching.

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -243,7 +243,7 @@ class Search
       end
     end
 
-    find_grouped_results unless @results.posts.present?
+    find_grouped_results if @results.posts.blank?
 
     if preloaded_topic_custom_fields.present? && @results.posts.present?
       topics = @results.posts.map(&:topic)
@@ -608,7 +608,6 @@ class Search
   end
 
   def find_grouped_results
-
     if @results.type_filter.present?
       raise Discourse::InvalidAccess.new("invalid type filter") unless Search.facets.include?(@results.type_filter)
       send("#{@results.type_filter}_search")
@@ -727,6 +726,7 @@ class Search
 
   def posts_query(limit, opts = nil)
     opts ||= {}
+
     posts = Post.where(post_type: Topic.visible_post_types(@guardian.user))
       .joins(:post_search_data, :topic)
       .joins("LEFT JOIN categories ON categories.id = topics.category_id")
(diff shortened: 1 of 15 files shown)


tgxworld:
@SamSaffron This is the first part of the search improvements where the ground work like adding the `Category#search_priority` column is made.

spec/components/search_spec.rb
```diff
@@ -576,7 +621,10 @@ def search
     end
 
     it 'can use category as a search context' do
-      category = Fabricate(:category)
+      category = Fabricate(:category,
+        search_priority: Searchable::PRIORITIES[:ignore]
+      )
+
```
SamSaffron: it is fine to sit on this ... but also test this returns stuff. 

```
 search = Search.execute("hello #{category.slug}")
```

lib/search.rb
```diff
@@ -847,6 +851,23 @@ def posts_query(limit, opts = nil)
     posts.limit(limit)
   end
 
+  def categories_ignored(posts)
+    posts.where(<<~SQL, Searchable::PRIORITIES[:ignore])
```
SamSaffron: I don't know, this feels way too fancy to me and uneeded. 

Force people to explicitly set on each category and child category. I can see cases where people may for whatever reason ignore parent, but keep children, nothing wrong with that.

tgxworld: @SamSaffron If I do it in Ruby the code is actually quite simple but it comes at the expense of two extra connections to the DB. 

> Force people to explicitly set on each category and child category. I can see cases where people may for whatever reason ignore parent, but keep children, nothing wrong with that.

Not sure about this. If a parent category has 10 child category, it'll be quite troublesome to have to set ignore on the 10 child category. Ignoring parent but keeping children seems like a case where the parent category has very little relationship to the children categories.

SamSaffron: Oh the SQL here is 100% fine for the proposed feature. BUT I want to keep this simple for now force people to be explicit per category with no inheritance.

Cause otherwise is it going to get crazy when you have other things except for "ignore"

low priority parent
   nothing set on child 1
   high priority on child 2 

Good luck explaining what happens here :)

tgxworld: Ah OK :+1: This is done.

spec/components/search_spec.rb
tgxworld: O this is already tested in https://github.com/discourse/discourse/blob/0a8f95028176c62b06194804be23bfd944ac5873/spec/components/search_spec.rb#L874

tgxworld:
Merging this first since the changes here are pretty safe and can be reverted easily.

PR #7197, opened 2019-03-18
FEATURE: Post notices become old after 14 days.

udan11:


app/assets/javascripts/discourse/widgets/post.js.es6
```diff
@@ -432,12 +432,22 @@ createWidget("post-notice", {
...
     if (attrs.postNoticeType === "first") {
-      return ["new-user"];
+      classes.push("new-user");
     } else if (attrs.postNoticeType === "returning") {
-      return ["returning-user"];
+      classes.push("returning-user");
+    }
+
+    if (
+      +new Date() - +new Date(attrs.created_at) >
```
eviltrout: This is really weird syntax. `+ newDate()` - why the `+`?

ZogStriP: The `+` is a shortcut to convert a date into an `Integer` in JS.

eviltrout: `getTime()` is WAY more clear to me personally, but if we can remove it and have it still work let's do that.

PR #7202, opened 2019-03-19
[FEATURE] Disallow ignoring self, admins or moderators users

khalilovcmded:
## Why?

We do not want to allow Users ignoring:

1. Themselves
2. Any Admins or Moderators

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -997,10 +997,7 @@ class UsersController < ApplicationController
 
   def ignore
     raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
-
-    unless current_user.id != params[:ignored_user_id] && User.where(id: params[:ignored_user_id], admin: false, moderator: false).exists?
-      return render json: failed_json, status: 422
-    end
+    ensure_can_ignore_user!(params[:ignored_user_id])
 
     IgnoredUser.find_or_create_by!(user: current_user, ignored_user_id: params[:ignored_user_id])
     render json: success_json
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -996,9 +996,7 @@ class UsersController < ApplicationController
   end
 
   def ignore
-    unless SiteSetting.ignore_user_enabled
-      raise Discourse::NotFound
-    end
+    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
 
     unless current_user.id != params[:ignored_user_id] && User.where(id: params[:ignored_user_id], admin: false, moderator: false).exists?
       return render json: failed_json, status: 422
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -996,11 +996,15 @@ class UsersController < ApplicationController
   end
 
   def ignore
-    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
+    unless SiteSetting.ignore_user_enabled
+      raise Discourse::NotFound
+    end
 
-    ::IgnoredUser.find_or_create_by!(
-      user: current_user,
-      ignored_user_id: params[:ignored_user_id])
+    unless current_user.id != params[:ignored_user_id] && User.where(id: params[:ignored_user_id], admin: false, moderator: false).exists?
+      return render json: failed_json, status: 422
+    end
+
+    IgnoredUser.find_or_create_by!(user: current_user, ignored_user_id: params[:ignored_user_id])
     render json: success_json
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -997,7 +997,7 @@ class UsersController < ApplicationController
 
   def ignore
     raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
-    ensure_can_ignore_user!(params[:ignored_user_id])
+    guardian.ensure_can_ignore_user!(params[:ignored_user_id])
 
     IgnoredUser.find_or_create_by!(user: current_user, ignored_user_id: params[:ignored_user_id])
     render json: success_json
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -46,7 +46,7 @@ class BasicPostSerializer < ApplicationSerializer
   def ignored
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
-      !object.user.staff? &&
+      !object.user&.staff? &&
       IgnoredUser.where(user_id: scope.current_user&.id,
                         ignored_user_id: object.user_id).exists?
   end
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -46,8 +46,7 @@ class BasicPostSerializer < ApplicationSerializer
   def ignored
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
-      !object.user.admin? &&
-      !object.user.moderator? &&
+      !object.user.staff? &&
       IgnoredUser.where(user_id: scope.current_user&.id,
                         ignored_user_id: object.user_id).exists?
   end
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -44,8 +44,12 @@ class BasicPostSerializer < ApplicationSerializer
   end
 
   def ignored
-    object.is_first_post? && IgnoredUser.where(user_id: scope.current_user&.id,
-                                               ignored_user_id: object.user_id).exists?
+    object.is_first_post? &&
+      !object.user.admin? &&
+      !object.user.moderator? &&
+      scope.current_user&.id != object.user_id &&
+      IgnoredUser.where(user_id: scope.current_user&.id,
+                        ignored_user_id: object.user_id).exists?
   end
 
   def include_name?
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -45,9 +45,9 @@ class BasicPostSerializer < ApplicationSerializer
 
   def ignored
     object.is_first_post? &&
+      scope.current_user&.id != object.user_id &&
       !object.user.admin? &&
       !object.user.moderator? &&
-      scope.current_user&.id != object.user_id &&
       IgnoredUser.where(user_id: scope.current_user&.id,
                         ignored_user_id: object.user_id).exists?
   end
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -282,7 +282,7 @@ class UserSerializer < BasicUserSerializer
   end
 
   def can_ignore_user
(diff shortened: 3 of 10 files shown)


khalilovcmded:
I will be adding specs, but I would love to get some feedback first to see if that's a suitable approach.

app/serializers/user_serializer.rb
```diff
@@ -282,7 +282,7 @@ def ignored
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled
+    SiteSetting.ignore_user_enabled && (scope.user&.admin? || scope.user&.moderator?)
```
khalilovcmded: ⚠️ Reviewing the logic of this.

app/controllers/users_controller.rb
```diff
@@ -996,11 +996,15 @@ def destroy
   end
 
   def ignore
-    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
+    unless SiteSetting.ignore_user_enabled
```
eviltrout: I actually preferred this on one line. Also it reads better if you do `ignore_user_enabled?`

app/controllers/users_controller.rb
```diff
@@ -996,11 +996,15 @@ def destroy
...
+      raise Discourse::NotFound
+    end
+
+    unless current_user.id != params[:ignored_user_id] && User.where(id: params[:ignored_user_id], admin: false, moderator: false).exists?
```
eviltrout: The double negative here is somewhat confusing to read. Also, this logic is probably best put into the `Guardian` class as a `can_ignore_user(user)` method. Then in here you could say `ensure_can_ignore_user!(user)`

app/serializers/basic_post_serializer.rb
```diff
@@ -44,8 +44,12 @@ def cooked
   end
 
   def ignored
-    object.is_first_post? && IgnoredUser.where(user_id: scope.current_user&.id,
-                                               ignored_user_id: object.user_id).exists?
+    object.is_first_post? &&
+      scope.current_user&.id != object.user_id &&
+      !object.user.admin? &&
```
eviltrout: You can use `user.staff?` to handle `admin` and `moderator` cases at the same time.

app/serializers/basic_post_serializer.rb
```diff
@@ -44,8 +44,12 @@ def cooked
...
+      !object.user.moderator? &&
+      IgnoredUser.where(user_id: scope.current_user&.id,
```
eviltrout: I think I saw some discussion about this, but I'd like to confirm what we decided - isn't this a n+1 query here? Each serialized post will perform a query?

app/serializers/user_serializer.rb
```diff
@@ -282,7 +282,7 @@ def ignored
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled
+    SiteSetting.ignore_user_enabled && !object.admin? && !object.moderator?
```
eviltrout: You can use the `staff?` helper (and `ignore_user_enabled?` reads better)

lib/topic_view.rb
```diff
@@ -603,7 +603,18 @@ def setup_filtered_posts
     @filtered_posts = unfiltered_posts
 
     if SiteSetting.ignore_user_enabled
-      ignored_user_ids = IgnoredUser.where(user_id: @user&.id).pluck(:ignored_user_id)
+
+      sql = <<~SQL
+            SELECT ignored_user_id
```
eviltrout: The indentation here is odd. Should be two spaces in from `sql`

app/controllers/users_controller.rb
ZogStriP: I think a one-liner is fine her

``` ruby
raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
```

app/controllers/users_controller.rb
ZogStriP: Can we get this to us an `if` instead of an `unless`?

khalilovcmded: Sure, I can look into that 👍

app/serializers/basic_post_serializer.rb
ZogStriP: You can use `staff?` here instead of `admin? || moderator?` :wink:

app/serializers/user_serializer.rb
ZogStriP: Ditto ^^

app/controllers/users_controller.rb
khalilovcmded: Just to be clear, the code that I can relocate in guardian is:
```
User.where(id: params[:ignored_user_id], admin: false, moderator: false).exists?
```

?

lib/topic_view.rb
```diff
@@ -603,7 +603,18 @@ def setup_filtered_posts
...
+            FROM ignored_users as ig
+            JOIN users as u ON u.id = ig.ignored_user_id
+            WHERE ig.user_id = :current_user_id
+              AND ig.ignored_user_id <> :current_user_id
+              AND u.admin = FALSE
```
ZogStriP:
``` sql
AND NOT u.admin
AND NOT u.moderator
```

app/controllers/users_controller.rb
eviltrout: You can also move the current user check too! A guardian knows which user it's checking for.

app/serializers/basic_post_serializer.rb
khalilovcmded: So I think this is making one extra call for every post, and we can work on fixing/optimising this. But I don't think  that this is making N calls for every post we serialise. In my head, this feels more like: `1+1` rather than `N+1`.

If this is a `N+1` situation, it would be that for every _1_ **Post**, we are trying to fetch _N_ **Ignored Users**, which is commonly referred to as: `N+1` but it is actually a `1+N`.

ZogStriP: 1+1 is a valid N+1 with N=1 😉

ZogStriP: Yup, it's fine. As long as you don't forget about it.

khalilovcmded:
@eviltrout no worries at all! I always find your feedback super useful ☺️ Thank you for that!

app/serializers/basic_post_serializer.rb
eviltrout: If it is indeed making one query per post we have to fix this ASAP as that's a huge performance penalty on every topic.

(BTW this is exactly what is known as N+1 in rails. You are retrieving N posts to render in a topic. This code will cause N queries (1 per post) + 1 which is the query that retrieved the list of posts in the first place.)

eviltrout: Ah totally my bad then. If it's only the first post it is indeed NOT a n+1 and I was wrong. But I think both @tgxworld and myself got confused by this. Even if we're both wrong in the end, I think a comment would save future programmers some time: `NOTE: This is not a n+1 because it's only on the first post`. Make it stupid proof :)

ZogStriP: I definitely think we should fix it, because if/when we remove that test, it'll become a bad N+1.

spec/services/user_updater_spec.rb
```diff
@@ -23,6 +23,15 @@
...
       expect(MutedUser.where(user_id: u3.id).count).to eq 0
     end
+
+    it 'excludes acting user' do
+      u1 = Fabricate(:user)
+      u2 = Fabricate(:user)
+      updater = UserUpdater.new(u1, u1)
+      updater.update_muted_users("#{u1.username},#{u2.username}")
+
+      expect(MutedUser.where(user_id: u1.id).count).to eq 1
```
ZogStriP: We should test that we're only muting `u2`, not just we're muting 1 user.

khalilovcmded:
## Constraints

One thing that I feel will be off here is that we don't filter out the `admin or moderators` in  preference -> users -> ignored users section.

mgHxdmdGZZ

- `Teddy` is a moderator user
- Even if I selected Teddy, on the backend, we never same him as an ignored user
- Though when I click save, it appears to be that Teddy is saved as an ignored user
- If I refresh, he is not there anymore

spec/components/guardian_spec.rb
```diff
@@ -2640,6 +2640,32 @@
     end
   end
 
+  describe '#can_ignore_user?' do
+    let(:guardian) { Guardian.new(user) }
+
+    context "when ignored user is the same as guardian user" do
+      it 'does not allow ignoring user' do
+        expect(guardian.can_ignore_user?(user.id)).to be_falsey
```
ZogStriP: We prefer `.to eq(false)` over `.to be_falsey`

khalilovcmded: Is there a reason for the preference? – Also `be_falsey` is used everywhere in this spec, should we change it?

spec/components/topic_view_spec.rb
```diff
@@ -92,6 +92,18 @@
...
         end
+
+        describe "when a staff user is ignored" do
+          let!(:admin) { Fabricate(:user, admin: true) }
+          let!(:admin_ignored_user) { Fabricate(:ignored_user, user: evil_trout, ignored_user: admin) }
+          let!(:post4) { Fabricate(:post, topic: topic, user: admin_ignored_user) }
+
+          it "filters out ignored user excluding the staff user" do
+            tv = TopicView.new(topic.id, nil)
+            expect(tv.filtered_post_ids.size).to eq(3)
```
ZogStriP: You don't need to test the size if you're testing the whole array 😉

spec/components/topic_view_spec.rb
```diff
@@ -92,6 +92,18 @@
...
+            expect(tv.filtered_post_ids).to match_array([post.id, post2.id, post4.id])
```
ZogStriP: I think the order matters here, right?

khalilovcmded: In general, post ordering matters, though in this context, we never touch ordering, or we rather we don't want to be testing for order.

What's your suggestion?

spec/components/guardian_spec.rb
ZogStriP: `falsey` is not explicit enough as it'll match both `nil` and `false`.

spec/components/topic_view_spec.rb
ZogStriP: I think `.to eq([post.id, post2.id, post4.id])` should work?

khalilovcmded: I get your point now, that makes sense ☺️

app/serializers/basic_post_serializer.rb
tgxworld: In search, this can become an N+1 in the worst case scenario where every post is a first post.

app/serializers/user_serializer.rb
```diff
@@ -282,7 +282,7 @@ def ignored
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled
+    SiteSetting.ignore_user_enabled? && !object.staff?
```
ZogStriP: We should also check that `scope.current_user != object`

PR #7204, opened 2019-03-19
Update flag stats inside Reviewable

romanrizzi:
Updating the flag stats now take place inside `Reviewable#perform` and let us reuse this behavior.

### DIFF
--- a/app/models/reviewable.rb
+++ b/app/models/reviewable.rb
@@ -189,6 +189,8 @@ class Reviewable < ActiveRecord::Base
 
       if result.success?
         transition_to(result.transition_to, performed_by) if result.transition_to
+        update_flag_stats(**result.update_flag_stats) if result.update_flag_stats
+
         recalculate_score if result.recalculate_score
       end
     end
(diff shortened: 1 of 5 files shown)


app/models/reviewable.rb
```diff
@@ -189,6 +189,8 @@ def perform(performed_by, action_id, args = nil)
 
       if result.success?
         transition_to(result.transition_to, performed_by) if result.transition_to
+        update_flag_stats(result.flag_to_update, result.informant_ids)
```
eviltrout: To be consistent with `recalculate_scores`, I think we should change the names of these members. My suggestion would be:

`result.update_flag_stats`

Also `informant_ids` is not clear that it relates to this flag update from its name alone. I would suggest (at the call site):

`result.update_flag_stats = { status: :agreed, user_ids: [1, 2, 3] }`

Then this line could be:

`update_flag_stats(result.update_flag_stats) if result.update_flag_stats`

I realize that the `if` condition is redundant because your method currently looks for a `nil` value, but I think it's clearer to future readers of the code that flag stats are only meant to be updated if that member is present.

PR #7205, opened 2019-03-19
Faster Travis tests

pmusaraj:
Reorganize our Travis matrix to speed up the total time it takes to run the Travis tests by: 

* splitting Ruby tests into 2 nodes using Knapsack. 
* running linting and qunit tests together in the same node -- linting is quite quick (~ 2mins) and it's not worth running it in a separate node because it takes too long to prep the node

If Travis nodes run in parallel (they don't always), this should take under 20 mins (compared to 28-29 mins previously).


pmusaraj:
Hmm, this is actually slower than before ([26 mins](https://travis-ci.org/discourse/discourse/builds/508567754?utm_source=github_status&utm_medium=notification) vs [25 mins](https://travis-ci.org/discourse/discourse/builds/508552558?utm_source=github_status&utm_medium=notification), as an example). 

A similar PR to this in my fork came in at 18 mins, not sure why the discrepancy.

image

I'll try running the builds a few more times in the next few days.

SamSaffron:
I worry about carrying knapsack files in our repo, it feels like something that is never 100% accurate and always in flux.

SamSaffron:
@pmusaraj I am closing this sadly on the grounds that I don't want us to carry `knapsack_rspec_report.json` in our repo, totally open to the knapsack approach maybe if we grab a report via a quick http call or something, but uneasy to carry this report and maintain it long term

PR #7207, opened 2019-03-19
FIX: Improve UX for second factor enforcement.

nbianca:


SamSaffron:
Merging it in, very hard to review this without a screenshot of the change, but I trust it is good :+1:

PR #7208, opened 2019-03-19
UX: Enabled sorting for more columns in admin user list

venarius:
This enables sorting for `email`, `username` and `last emailed` in the admin user list.

Also changed the `showEmail` button to a permanent toggle since the old behavior felt like a bug.

### Before:
Peek 2019-03-19 22-17

### After:
Peek 2019-03-19 22-16

I also wanted to write Qunit tests for both the sorting and the show / hide email function but somehow both of these functions weren't working in Qunit. Sorting wasn't clickable and the show / hide email function somehow always showed the emails.

If someone knows why that is, I will happily add tests to this PR!

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
@@ -8,7 +8,7 @@ export default Ember.Controller.extend(CanCheckEmails, {
   query: null,
   order: null,
   ascending: null,
-  showEmails: false,
+  showEmailsToggle: false,
   refreshing: false,
   listFilter: null,
   selectAll: false,
@@ -51,7 +51,7 @@ export default Ember.Controller.extend(CanCheckEmails, {
 
     AdminUser.findAll(this.get("query"), {
       filter: this.get("listFilter"),
-      show_emails: this.get("showEmails"),
+      show_emails: this.get("showEmailsToggle"),
       order: this.get("order"),
       ascending: this.get("ascending")
     })
@@ -88,7 +88,12 @@ export default Ember.Controller.extend(CanCheckEmails, {
     },
 
     showEmails: function() {
-      this.set("showEmails", true);
+      this.set("showEmailsToggle", true);
+      this._refreshUsers();
+    },
+
+    hideEmails: function() {
+      this.set("showEmailsToggle", false);
       this._refreshUsers();
     }
   }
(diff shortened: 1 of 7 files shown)


venarius:
Hmm, seems like the first CI build hung itself. No need for prettier, since I already prettified it.

tgxworld:
> Sorting wasn't clickable and the show / hide email function somehow always showed the emails. 

Perhaps you can write out some of the tests and we can see why it might not have worked?

venarius:
cc @tgxworld just added them if you want to have a look

app/assets/javascripts/admin/controllers/admin-users-list-show.js.es6
```diff
@@ -88,7 +88,12 @@ export default Ember.Controller.extend(CanCheckEmails, {
     },
 
     showEmails: function() {
-      this.set("showEmails", true);
+      this.set("showEmailsToggle", true);
```
tgxworld: are we able to utilize `this.toggleProperty` here instead of having two actions?

tgxworld:
@venarius Ah I think I know why. QUnit tests uses fixtures which returns a fixed set of response. Looking at the pretender for `this.get("/admin/users/list/active.json")`, you're always going to get the users in the same set of order which is why it seems like the sort isn't working. The sort is done server side so the fixtures need to be updated.

venarius:
@tgxworld yeah, that was the problem. I added a sort function the fixture and the sorting test is now working.

test/javascripts/acceptance/admin-users-list-test.js.es6
```diff
@@ -9,6 +9,50 @@ QUnit.test("lists users", async assert => {
...
+  await visit("/admin/users/list/active");
+
+  assert.ok(exists(".users-list .user"));
+
+  await click(".users-list .sortable:nth-child(1)");
+
+  assert.ok(
+    find(".users-list .user:nth-child(1) .username")
+      .text()
+      .includes("eviltrout")
```
tgxworld: It is a good practice to include a message about what the test is expecting. Otherwise, it won't be clear which test fails when it happens.

tgxworld:
Looks good to me @venarius :+1: Thanks for the contribution!

venarius:
errr @tgxworld you might want to undo this since the "hide email" qunit test is still not working. We only fixed the "sort users" test.

tgxworld:
@venarius no worries I've fixed it

PR #7209, opened 2019-03-20
FEATURE: Allow categories to be prioritized/prioritized in search.

tgxworld:


tgxworld:
@SamSaffron Can you review this for me? Thanks :+1:

lib/search.rb
```diff
@@ -837,8 +837,26 @@ def posts_query(limit, opts = nil)
...
+          #{ts_query(weight_filter: weights)},
+          0|32
+        ) *
+        (
+          CASE categories.search_priority
+          WHEN #{Searchable::PRIORITIES[:very_low]}
+          THEN #{SiteSetting.category_search_priority_very_low_weight}
+          WHEN #{Searchable::PRIORITIES[:low]}
+          THEN #{SiteSetting.category_search_priority_low_weight}
+          WHEN #{Searchable::PRIORITIES[:normal]} THEN 1
```
ZogStriP: Do you need this line when it's the default and covered by the `ELSE`?

lib/validators/category_search_priority_weights_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+      val < 1 && val > SiteSetting.category_search_priority_very_low_weight
+    when "category_search_priority_high_weight"
+      val > 1 && val < SiteSetting.category_search_priority_very_high_weight
+    when "category_search_priority_very_high_weight"
+      val > SiteSetting.category_search_priority_high_weight
+    end
+  end
+
+  def error_message
+    key =
```
ZogStriP: You could also do 

``` ruby
key = @name[/category_search_priority_(\w+)_weight/, 1]
```

lib/search.rb
```diff
@@ -837,8 +837,26 @@ def posts_query(limit, opts = nil)
       end
     else
       data_ranking = <<~SQL
-      TS_RANK_CD(
-        post_search_data.search_data, #{ts_query(weight_filter: weights)}
+      (
+        TS_RANK_CD(
+          post_search_data.search_data,
+          #{ts_query(weight_filter: weights)},
+          0|32
```
ZogStriP: These magic numbers (flags) need at the very least a comment explaining what they do

tgxworld: Hmm these numbers are in the documentation for the function being used here and I would prefer not to have to duplicate the Postgres docs here.

ZogStriP: I don't mean to copy-paste the postgres doc, but why do you use this combination of flags?

config/site_settings.yml
```diff
@@ -1614,6 +1614,18 @@ search:
       ro: true
       sk: true
       tr_TR: true
+  category_search_priority_very_low_weight:
```
SamSaffron: I am uneasy with all these toggles in the UX it feels like something that will just cause confusion to admins, can we start off with just having these as hidden site settings?

SamSaffron:
feel free to merge once my comment is addressed :+1:

tgxworld:
Thanks for reviewing @ZogStriP and @SamSaffron

PR #7210, opened 2019-03-20
DEV: Remove SiteSetting.s3_force_path_style

rishabhnambiar:
- s3_force_path_style was added as a Minio specific url scheme but it has never been well supported in our code base.
- Our new migrate_to_s3 rake task does not work reliably with path style urls too
- Minio has also added support for virtual style requests i.e the same scheme as AWS S3/DO Spaces so we can rely on that instead of using path style requests.


tgxworld:
You'll need to write a migration to drop the record from the `site_settings` table as well :)

PR #7211, opened 2019-03-20
FEATURE: Add scheduled job to purge expired ignored users

khalilovcmded:
## Why?

> The maximum ignore duration will be 4 months, so let’s go ahead and set up a task to clear these ignores after 4 months, as we have to assume all beta ignores are “max duration”.

### DIFF
--- a/spec/jobs/purge_expired_ignored_users_spec.rb
+++ b/spec/jobs/purge_expired_ignored_users_spec.rb
@@ -7,7 +7,6 @@ describe Jobs::PurgeExpiredIgnoredUsers do
 
   context "with no ignored users" do
     it "does nothing" do
-      subject
       expect { subject }.to_not change { IgnoredUser.count }
     end
   end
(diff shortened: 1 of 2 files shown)


spec/jobs/purge_expired_ignored_users_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+require_dependency 'jobs/scheduled/purge_expired_ignored_users'
+
+describe Jobs::PurgeExpiredIgnoredUsers do
+  subject { Jobs::PurgeExpiredIgnoredUsers.new.execute({}) }
+
+  context "with no ignored users" do
+    it "does nothing" do
+      subject
```
ZogStriP: Do we need that line?

spec/jobs/purge_expired_ignored_users_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    let(:john) { Fabricate(:user, username: "john") }
+
+    before do
+      Fabricate(:ignored_user, user: tarek, ignored_user: matt)
+      Fabricate(:ignored_user, user: tarek, ignored_user: john)
+    end
+
+    context "when no expired ignored users" do
+      it "does nothing" do
+        subject
```
ZogStriP: Ditto ^^

spec/jobs/purge_expired_ignored_users_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    context "when there are expired ignored users" do
+      it "purges expired ignored users" do
+        freeze_time(5.months.ago) do
+          fred = Fabricate(:user, username: "fred")
+          Fabricate(:ignored_user, user: tarek, ignored_user: fred)
+        end
+
+        expect { subject }.to change { IgnoredUser.count }.from(3).to(2)
```
ZogStriP: I would also make sure we can't find the old "IgnoreUser" record anymore.

khalilovcmded: Yes, makes sense. I will actually remove this check, and just check we can't find that specific IgnoredUser anymore.

PR #7212, opened 2019-03-20
FEATURE: Only allow TL2 Users to ignore other users

khalilovcmded:
## Why?

We do not want to allow Users with trust level below TL2 to ignore other users.

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -282,7 +282,10 @@ class UserSerializer < BasicUserSerializer
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled? && !object.staff? && scope.current_user != object && (scope.current_user.staff? || scope.current_user.trust_level >= 2)
+    SiteSetting.ignore_user_enabled? &&
+      !object.staff? &&
+      scope.current_user != object &&
+      (scope.current_user.staff? || scope.current_user.trust_level >= TrustLevel.levels[:member])
   end
 
   # Needed because 'send_private_message_to_user' will always return false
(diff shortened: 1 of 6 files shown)


app/services/user_updater.rb
```diff
@@ -168,6 +168,8 @@ def update_muted_users(usernames)
   end
 
   def update_ignored_users(usernames)
+    return if user.trust_level < 2
```
khalilovcmded: ⚠️ I could have use guardian here, but then we will need to invoke it for every Ignored User (`usernames`)

app/serializers/user_serializer.rb
```diff
@@ -282,7 +282,7 @@ def ignored
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled? && !object.staff? && scope.current_user != object
+    SiteSetting.ignore_user_enabled? && !object.staff? && scope.current_user != object && (scope.current_user.staff? || scope.current_user.trust_level >= 2)
```
ZogStriP: That line is getting too long. Might want to put each conditions on its own line 😉 

Also, can you use `TrustLevel.levels[:member]` instead of the magic number `2`?

khalilovcmded: Makes sense 👍

app/services/user_updater.rb
ZogStriP: You could have a guardian method `can_ignore_users?` which would do something like

``` ruby
def can_ignore_users?
  SiteSetting.ignore_user_enabled? && (is_staff? || user.trust_level >= TrustLevel.levels[:member])
end
```

khalilovcmded: Yes, though, do we want to have yet another method in the guardian for a seemingly check?

ZogStriP:
```suggestion
    return if user.trust_level < TrustLevel.levels[:member]
```

PR #7214, opened 2019-03-20
Support for Reviewable types discoverability

romanrizzi:
Removed hardcoded reviewable types. Expose a plugin method to add new reviewable types

### DIFF
--- a/lib/plugin/instance.rb
+++ b/lib/plugin/instance.rb
@@ -604,6 +604,15 @@ class Plugin::Instance
     end
   end
 
+  def register_reviewable_type(reviewable_type_class)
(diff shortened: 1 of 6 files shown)


lib/plugin/instance.rb
```diff
@@ -604,6 +604,15 @@ def each_globbed_asset
     end
   end
 
+  def register_reviewable_type(reviewable_type)
```
eviltrout: I noticed in the Akismet plugin you called this with `MyType.name` - what do you think about passing the class in instead of the string?

romanrizzi: Sounds good!

PR #7215, opened 2019-03-20
 FEATURE: Mute user when ignoring the same user

khalilovcmded:
## Why?

If we ignore a user, either from the preferences or user summary or API , we will mute the same user too.

## Constraints

This can yield a janky behaviour on the UI, but I have a plan to fix that in a subsequent PR.

### DIFF
--- a/spec/services/user_updater_spec.rb
+++ b/spec/services/user_updater_spec.rb
@@ -51,7 +51,7 @@ describe UserUpdater do
 
       expect(IgnoredUser.where(user_id: u2.id).pluck(:ignored_user_id)).to match_array([u3.id, u1.id])
       expect(IgnoredUser.where(user_id: u1.id).pluck(:ignored_user_id)).to match_array([u2.id, u3.id])
-      expect(IgnoredUser.where(user_id: u3.id)).to match_array([])
+      expect(IgnoredUser.where(user_id: u3.id).count).to eq(0)
       expect(MutedUser.where(user_id: u2.id).pluck(:muted_user_id)).to match_array([u3.id, u1.id])
       expect(MutedUser.where(user_id: u1.id).pluck(:muted_user_id)).to match_array([u2.id, u3.id])
       expect(MutedUser.where(user_id: u3.id).count).to eq(0)
--- a/spec/services/user_updater_spec.rb
+++ b/spec/services/user_updater_spec.rb
@@ -78,6 +78,17 @@ describe UserUpdater do
         expect(MutedUser.where(muted_user_id: u2.id).count).to eq 0
       end
     end
+
+    context 'when acting user\'s trust level is below tl2' do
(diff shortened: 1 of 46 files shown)


spec/services/user_updater_spec.rb
```diff
@@ -52,6 +52,9 @@
       expect(IgnoredUser.where(user_id: u2.id).count).to eq 2
       expect(IgnoredUser.where(user_id: u1.id).count).to eq 2
       expect(IgnoredUser.where(user_id: u3.id).count).to eq 0
+      expect(MutedUser.where(user_id: u2.id).count).to eq 2
```
ZogStriP: Can we check that we have the proper Muted/Ignored user ids instead of just the #?

khalilovcmded: I agree. Then I will also refactor the `MutedUser` specs, not so safe.

spec/services/user_updater_spec.rb
```diff
@@ -19,9 +19,9 @@
       updater = UserUpdater.new(u3, u3)
       updater.update_muted_users("")
 
-      expect(MutedUser.where(user_id: u2.id).count).to eq 2
-      expect(MutedUser.where(user_id: u1.id).count).to eq 2
-      expect(MutedUser.where(user_id: u3.id).count).to eq 0
+      expect(MutedUser.where(user_id: u2.id).pluck(:muted_user_id)).to match_array([u3.id, u1.id])
+      expect(MutedUser.where(user_id: u1.id).pluck(:muted_user_id)).to match_array([u2.id, u3.id])
+      expect(MutedUser.where(user_id: u3.id)).to match_array([])
```
ZogStriP: It's fine to use count when it's 0 😉

spec/services/user_updater_spec.rb
```diff
@@ -49,9 +49,12 @@
       updater = UserUpdater.new(u3, u3)
       updater.update_ignored_users("")
 
-      expect(IgnoredUser.where(user_id: u2.id).count).to eq 2
-      expect(IgnoredUser.where(user_id: u1.id).count).to eq 2
-      expect(IgnoredUser.where(user_id: u3.id).count).to eq 0
+      expect(IgnoredUser.where(user_id: u2.id).pluck(:ignored_user_id)).to match_array([u3.id, u1.id])
+      expect(IgnoredUser.where(user_id: u1.id).pluck(:ignored_user_id)).to match_array([u2.id, u3.id])
+      expect(IgnoredUser.where(user_id: u3.id)).to match_array([])
```
ZogStriP:
```suggestion
      expect(IgnoredUser.where(user_id: u3.id).count).to eq(0)
```

ZogStriP:
We'll be updating the queries to ensure we check for `ignored_users` whenever we check for `muted_users` instead of mucking with the lists.

PR #7217, opened 2019-03-20
UX: show if webhook is disabled

majakomel:
+ show in staff logs when webhook is created/updated/destroyed

Screenshot 2019-03-20 at 16 54 08

### DIFF
--- a/app/controllers/admin/web_hooks_controller.rb
+++ b/app/controllers/admin/web_hooks_controller.rb
@@ -34,6 +34,7 @@ class Admin::WebHooksController < Admin::AdminController
     web_hook = WebHook.new(web_hook_params)
 
     if web_hook.save
+      StaffActionLogger.new(current_user).log_webhook(web_hook, action: "create")
       render_serialized(web_hook, AdminWebHookSerializer, root: 'web_hook')
     else
       render_json_error web_hook.errors.full_messages
@@ -42,6 +43,7 @@ class Admin::WebHooksController < Admin::AdminController
 
   def update
     if @web_hook.update_attributes(web_hook_params)
+      StaffActionLogger.new(current_user).log_webhook(@web_hook, action: "update")
       render_serialized(@web_hook, AdminWebHookSerializer, root: 'web_hook')
     else
       render_json_error @web_hook.errors.full_messages
@@ -50,6 +52,7 @@ class Admin::WebHooksController < Admin::AdminController
 
   def destroy
     @web_hook.destroy!
+    StaffActionLogger.new(current_user).log_webhook(@web_hook, action: "destroy")
     render json: success_json
   end
 
--- a/app/controllers/admin/web_hooks_controller.rb
+++ b/app/controllers/admin/web_hooks_controller.rb
@@ -34,7 +34,7 @@ class Admin::WebHooksController < Admin::AdminController
     web_hook = WebHook.new(web_hook_params)
 
     if web_hook.save
-      StaffActionLogger.new(current_user).log_webhook(web_hook, action: "create")
+      StaffActionLogger.new(current_user).log_web_hook(web_hook, UserHistory.actions[:web_hook_create])
       render_serialized(web_hook, AdminWebHookSerializer, root: 'web_hook')
     else
       render_json_error web_hook.errors.full_messages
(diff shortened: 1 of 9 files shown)


app/controllers/admin/web_hooks_controller.rb
```diff
@@ -42,6 +43,7 @@ def create
 
   def update
     if @web_hook.update_attributes(web_hook_params)
+      StaffActionLogger.new(current_user).log_webhook(@web_hook, action: "update")
```
ZogStriP: Can we show what was updated/changed?

app/controllers/admin/web_hooks_controller.rb
```diff
@@ -34,6 +34,7 @@ def create
     web_hook = WebHook.new(web_hook_params)
 
     if web_hook.save
+      StaffActionLogger.new(current_user).log_webhook(web_hook, action: "create")
```
ZogStriP: Would be nice to test all these logs 😉

app/models/user_history.rb
```diff
@@ -85,7 +85,8 @@ def self.actions
       entity_export: 66,
       change_password: 67,
       topic_timestamps_changed: 68,
-      approve_user: 69
+      approve_user: 69,
+      webhook: 70
```
tgxworld:
```suggestion
      web_hook: 70
```

app/models/user_history.rb
```diff
@@ -149,7 +150,8 @@ def self.staff_actions
       :entity_export,
       :change_name,
       :topic_timestamps_changed,
-      :approve_user
+      :approve_user,
+      :webhook
```
tgxworld:
```suggestion
      :web_hook
```

app/models/user_history.rb
tgxworld: Not sure why but we named the model `WebHook`. I think we should keep the consistency here.

config/locales/client.en.yml
```diff
@@ -3702,6 +3703,7 @@ en:
             change_name: "change name"
             topic_timestamps_changed: "topic timestamps changed"
             approve_user: "approved user"
+            webhook: "webhook"
```
tgxworld:
```suggestion
            web_hook: "webhook"
```

app/services/staff_action_logger.rb
```diff
@@ -583,6 +583,18 @@ def log_post_rejected(rejected_post, opts = {})
     ))
   end
 
+  def log_webhook(webhook, opts = {})
```
tgxworld: I think the parameters for the method can be improved here. We're passing in the action as part of `opts` only to override it with `UserHistory.actions[:webhook]`. For post history events, each action is assigned a key like `post_edit` and `delete_post`. Consolidating all the different events under a single action means that the admin page will only show `webhook` making it difficult to filter between each individual event.

Screenshot from 2019-03-21 14-59-13

app/services/staff_action_logger.rb
```diff
@@ -583,6 +583,30 @@ def log_post_rejected(rejected_post, opts = {})
...
+      changes.each do |k, v|
+        old_values << "#{k}: #{v[0]}"
+        new_values << "#{k}: #{v[1]}"
+      end
+    end
+
+    UserHistory.create!(params(opts).merge(
+      action: action,
+      context: details.join(", "),
+      previous_value: old_values&.join(", "),
```
ZogStriP: Since you're initializing `old_values` and `new_values` to `[]`, it can never be `nil` so the `&.` is not necessary 😉

app/services/staff_action_logger.rb
```diff
@@ -583,6 +583,30 @@ def log_post_rejected(rejected_post, opts = {})
...
   end
 
+  def log_web_hook(web_hook, action, opts = {})
+    details = [
+      "webhook_id: #{web_hook.id}",
+      "payload_url: #{web_hook.payload_url}"
+    ]
+
+    if changes = opts[:changes]
+      changes.reject! { |k,v| k == "updated_at" }
```
ZogStriP: You're missing a space 😉 

```suggestion
      changes.reject! { |k, v| k == "updated_at" }
```

majakomel: ugh 😄

PR #7201, opened 2019-03-19
Revert "UX: Better emoji escaping for topic title"

ZogStriP:
Reverts discourse/discourse#7176 as this is breaking the https://github.com/discourse/discourse-prometheus-alert-receiver plugin (specs are failing)

PR #7218, opened 2019-03-20
UX: Better emoji escaping for topic title

venarius:
This is the second part of the "FIX: Better emoji escaping for topic title" in which the main change was applied to better escape emojis in topic titles and enhance the performEmojiUnescape function to now also support unicode emojis.

More Information: d32557e

This also includes a fix for failing specs on the "discourse-prometheus-alert-receiver" plugin mentioned in https://github.com/discourse/discourse/pull/7201

### DIFF
--- a/lib/pretty_text.rb
+++ b/lib/pretty_text.rb
@@ -227,7 +227,7 @@ module PrettyText
   end
 
   def self.unescape_emoji(title)
-    return title unless SiteSetting.enable_emoji?
+    return title unless SiteSetting.enable_emoji? && title
 
     set = SiteSetting.emoji_set.inspect
     custom = Emoji.custom.map { |e| [e.name, e.url] }.to_h.to_json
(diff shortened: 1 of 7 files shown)


lib/pretty_text.rb
```diff
@@ -227,7 +227,7 @@ def self.avatar_img(avatar_template, size)
   end
 
   def self.unescape_emoji(title)
-    return title unless SiteSetting.enable_emoji?
+    return title unless SiteSetting.enable_emoji? && title
```
ZogStriP: Why add `&& title`? If `title` is blank?, we should bail out immediately, shouldn't we?

venarius: But its bailing there. We only continue if `enable_emoji?` is true and a `title` is given.

ZogStriP: Nevermind, it's `unless`, not `if`... 🤯

ZogStriP:
Looks good. I'll merge this as soon as the build go back to green.

ZogStriP:
Let's try this 👍

PR #7219, opened 2019-03-20
REFACTOR: Move redundant ignored user check into guardian

khalilovcmded:
Both: [basic_post_serializer.rb](https://github.com/discourse/discourse/blob/master/app/serializers/basic_post_serializer.rb#L48-L51) and [user_serializer.rb](https://github.com/discourse/discourse/blob/master/app/serializers/user_serializer.rb#L286-L288) can re-use the logic in guardian.

Are we happy to instantiate guardian in those serializers? and refactor accordingly?

### DIFF
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -44,6 +44,7 @@ class BasicPostSerializer < ApplicationSerializer
   end
 
   def ignored
+    return false if SiteSetting.ignore_user_enabled?
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
       IgnoredUser.where(user_id: scope.current_user&.id,
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -44,7 +44,7 @@ class BasicPostSerializer < ApplicationSerializer
   end
 
   def ignored
-    return false if SiteSetting.ignore_user_enabled?
+    return false unless SiteSetting.ignore_user_enabled?
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
       IgnoredUser.where(user_id: scope.current_user&.id,
--- a/app/serializers/basic_post_serializer.rb
+++ b/app/serializers/basic_post_serializer.rb
@@ -46,7 +46,6 @@ class BasicPostSerializer < ApplicationSerializer
   def ignored
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
-      !object.user&.staff? &&
       IgnoredUser.where(user_id: scope.current_user&.id,
                         ignored_user_id: object.user_id).exists?
   end
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -282,10 +282,7 @@ class UserSerializer < BasicUserSerializer
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled? &&
-      !object.staff? &&
-      scope.current_user != object &&
-      (scope.current_user.staff? || scope.current_user.trust_level >= TrustLevel.levels[:member])
+    SiteSetting.ignore_user_enabled? && scope.can_ignore_user?(object.id)
   end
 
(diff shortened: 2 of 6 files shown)


app/serializers/basic_post_serializer.rb
```diff
@@ -46,7 +46,6 @@ def cooked
   def ignored
     object.is_first_post? &&
       scope.current_user&.id != object.user_id &&
-      !object.user&.staff? &&
```
khalilovcmded: Come to think about it, this is not needed here.

app/serializers/user_serializer.rb
```diff
@@ -282,10 +282,7 @@ def ignored
   end
 
   def can_ignore_user
-    SiteSetting.ignore_user_enabled? &&
-      !object.staff? &&
-      scope.current_user != object &&
-      (scope.current_user.staff? || scope.current_user.trust_level >= TrustLevel.levels[:member])
+    SiteSetting.ignore_user_enabled? && scope.can_ignore_user?(object.id)
```
ZogStriP: Should we move that check in the `guardian.can_ignore_users?` method?

khalilovcmded: We can I think. I don't see much strong pros or cons out of moving it.

lib/guardian.rb
```diff
@@ -391,9 +391,11 @@ def can_export_entity?(entity)
...
 
   def can_ignore_user?(user_id)
-    @user.id != user_id &&
-      (@user.staff? || @user.trust_level >= TrustLevel.levels[:member]) &&
-      User.where(id: user_id, admin: false, moderator: false).exists?
+    can_ignore_users? && @user.id != user_id && User.where(id: user_id, admin: false, moderator: false).exists?
+  end
+
+  def can_ignore_users?
+    SiteSetting.ignore_user_enabled? && (@user.staff? || @user.trust_level >= TrustLevel.levels[:member])
```
eviltrout: I believe it's valid to create a `Guardian` with a `nil` user, for anonymous users. It looks like you'd get a `nil` error here.

app/serializers/basic_post_serializer.rb
eviltrout: Right because it was enforced when the `IgnoredUser` was created?

ZogStriP: I think we should keep it since it'll avoid making the SQL query when the owner of the 1st is a staff member.

app/serializers/basic_post_serializer.rb
```diff
@@ -46,7 +46,6 @@ def cooked
   def ignored
```
eviltrout: Should this return false if the site setting is not enabled? Seems like a fast way to exit without doing much work.

app/serializers/basic_post_serializer.rb
khalilovcmded: This actually is not correct, it is a bug. We are checking here if this Post should be ignored or not. It shouldn't matter if this is a `staff`'s post or not.

lib/guardian.rb
khalilovcmded: If a nil user is passed, there is this line:

```ruby
    @user = user.presence || AnonymousUser.new
```

When I do: `::Guardian::AnonymousUser.new.staff?`, it returns false, but it fails for `trust_level`

khalilovcmded: Perhaps I can exit early: `return false if anonymous?`

eviltrout: So you could fix it either by adding `0` to the Anonymous user's trust level, or by doing what we usually do which is `return false if @user.blank?`

khalilovcmded: Won't it be better as: `return false if anonymous?`

app/serializers/basic_post_serializer.rb
khalilovcmded: Yes, good point.

lib/guardian.rb
eviltrout: Sure, I forgot about that helper method. even better :)

PR #7223, opened 2019-03-20
UX: Use a less technical term ("replace") for reseeding

gschlager:
image

image

image


ZogStriP:
`Go ahead` feels weird... Looks like it's only being used here and in CSV uploads. Can we make it consistent with all other modals?

gschlager:
Sure. I changed it.

image

ZogStriP:
Another feedback. I tried it on an instance on a low-end VPS and it took quite a while to run (like 5-ish seconds). Should we show a loading spinner or at least warn the user it might take a little while?

gschlager:
Good idea. I'll add that tomorrow.

PR #7224, opened 2019-03-21
Correct the mistake in client.zh_CN.yml

hundan2020:
correct the mistake, change "影藏" to "隐藏"


SamSaffron:
So sorry, we need you to update this transifex: see https://meta.discourse.org/t/contribute-a-translation-to-discourse/14882

PR #7228, opened 2019-03-21
UX: add more tag classes to topic view

majakomel:
https://meta.discourse.org/t/post-notices-for-first-time-and-returning-users/112078/12

### DIFF
--- a/app/assets/javascripts/discourse/components/add-category-tag-classes.js.es6
+++ b/app/assets/javascripts/discourse/components/add-category-tag-classes.js.es6
@@ -13,20 +13,24 @@ export default Ember.Component.extend({
       return;
     }
     const slug = this.get("category.fullSlug");
+    const tags = this.get("tags");
+
     this._removeClass();
-    if (slug) {
-      $("body").addClass(`category-${slug}`);
-    }
+
+    let classes = [];
+    if (slug) classes.push(`category-${slug}`);
+    if (tags) tags.forEach(t => classes.push(`tag-${t}`));
+    if (classes.length > 0) $("body").addClass(classes.join(" "));
   },
 
-  @observes("category.fullSlug")
+  @observes("category.fullSlug", "tags")
   refreshClass() {
     Ember.run.scheduleOnce("afterRender", this, this._updateClass);
   },
 
   _removeClass() {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/components/add-category-tag-classes.js.es6
```diff
@@ -26,7 +28,7 @@ export default Ember.Component.extend({
 
   _removeClass() {
     $("body").removeClass((_, css) =>
-      (css.match(/\bcategory-\S+/g) || []).join(" ")
+      (css.match(/\b(category-\S+|tag-\S+)/g) || []).join(" ")
```
ZogStriP: I _think_ you can make the regex shorter like this

```suggestion
      (css.match(/\b(category|tag)-\S+/g) || []).join(" ")
```

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -225,6 +225,13 @@ const Topic = RestModel.extend({
 
   categoryClass: fmt("category.fullSlug", "category-%@"),
 
+  @computed
```
ZogStriP: Should this be listening to `tags`?

```
@computed("tags")
tagClasslses(tags) {
  return tags.map(t => `tag-${t}`).join(" ");
}
```

app/assets/javascripts/discourse/components/add-category-tag-classes.js.es6
```diff
@@ -13,20 +13,22 @@ export default Ember.Component.extend({
...
     }
     const slug = this.get("category.fullSlug");
+    const tags = this.get("tags");
+
     this._removeClass();
-    if (slug) {
-      $("body").addClass(`category-${slug}`);
-    }
+
+    if (slug) $("body").addClass(`category-${slug}`);
```
ZogStriP: I think it would be much faster to only do one call to `addClass`. Something like

```javascript
let classes = [];
if (slug) classes.push(`category-${slug}`);
if (tags) tags.forEach(t => classes.push(`tag-${t}`));
if (classes.length > 0) $("body").addClass(classes.join(" "));
```

PR #7230, opened 2019-03-21
FEATURE: Include muted users count within the ignored users report

khalilovcmded:
## Why?

We want to add include the muted users count in the ignored users report.
We will also order by the `SUM(Ignored Users + Muted Users)`.

## UI

Admin_-_Discourse

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1591,23 +1591,50 @@ class Report
           :ignores_count,
         ],
         title: I18n.t("reports.top_ignored_users.labels.ignores_count")
+      },
+      {
+        type: :number,
+        properties: [
+          :mutes_count,
+        ],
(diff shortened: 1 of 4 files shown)


app/models/report.rb
```diff
@@ -1591,23 +1591,50 @@ def self.report_top_ignored_users(report)
...
-      LIMIT #{report.limit || 250}
+        WITH ignored_users AS (
+          SELECT
+          ignored_user_id as user_id,
+          COUNT(*) AS ignores_count
+          FROM ignored_users
+          WHERE created_at >= '#{report.start_date}' AND created_at <= '#{report.end_date}'
+          GROUP BY ignored_user_id
+          ORDER BY COUNT(*) DESC
+          LIMIT #{report.limit || 250}
```
ZogStriP: I think this should be extracted as a parameter to the query

PR #7233, opened 2019-03-21
See reviewable flagged posts from the user page

romanrizzi:
This PR updates the `user.flaggedPosts` routes in the user page and now points to the `ReviewableFlaggedPosts` of that user.

- Should we add any UI component to let the admin know they're seeing reviews of that user and/or filter by user in the search box?
- We're pointing to the flagged posts that were already *reviewed*. Should we add an option to retrieve all the flagged post?


spec/models/reviewable_spec.rb
```diff
@@ -106,6 +106,15 @@
         user.update_columns(moderator: false, admin: true)
         expect(Reviewable.list_for(user, status: :pending)).to eq([reviewable])
       end
+
+      it 'Let us filter by the target_created_by_id attribute' do
```
eviltrout: This test only shows that the filter will exclude. We should also have a test that returns the particular item for that id.

eviltrout:
> Should we add any UI component to let the admin know they're seeing reviews of that user and/or filter by user in the search box?

If you filter by topic something appears to let you remove that filter, so that is one option. Adding a user search would actually be quite powerful though, maybe we should do that instead for extra functionality?

> We're pointing to the flagged posts that were already reviewed. Should we add an option to retrieve all the flagged post?

A status of `reviewed` is probably good enough but I see the argument. A staff member might want to see everything a user did, regardless of status. Let's add an `(all)` option!

spec/models/reviewable_spec.rb
romanrizzi: I changed this test with the one that you suggest

eviltrout: Now you have the opposite problem. The filter might not do anything! :)

I was trying to suggest you test two things. The new test below + a test that passes a weird id and then you check for nothing.

romanrizzi: I added a second test to check  for exclusion :+1:

PR #7236, opened 2019-03-21
FIX: Clear unsaved groups when switching user

venarius:
This clears unsaved groups when admin switches to another user without saving.

Resolves: https://meta.discourse.org/t/admin-user-group-membership-edition-persists-between-users/112239

Didn't write a test for it since I didn't find any acceptance tests for custom groups. Just tell me when I should write some.

### DIFF
--- a/app/assets/javascripts/admin/routes/admin-user-index.js.es6
+++ b/app/assets/javascripts/admin/routes/admin-user-index.js.es6
@@ -19,6 +19,7 @@ export default Discourse.Route.extend({
     controller.setProperties({
       originalPrimaryGroupId: model.get("primary_group_id"),
       availableGroups: this._availableGroups,
+      customGroupIdsBuffer: null,
       model
     });
   }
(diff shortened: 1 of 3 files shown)


eviltrout:
Unfortunately we will rarely accept a fix without a test. If you need some clues on how to add a test @jjaffeux can probably help you out.

ZogStriP:
> Didn't write a test for it since I didn't find any acceptance tests for custom groups. Just tell me when I should write some.

Fixing a bug or a regression is a very good reason to add/fix a test 😉

venarius:
cc @ZogStriP @eviltrout took me quite some time but I got it figured out 🙂

app/assets/javascripts/admin/routes/admin-user-index.js.es6
```diff
@@ -19,6 +19,7 @@ export default Discourse.Route.extend({
     controller.setProperties({
       originalPrimaryGroupId: model.get("primary_group_id"),
       availableGroups: this._availableGroups,
+      customGroupIdsBuffer: null,
```
ZogStriP: I wonder, are these properties the only ones we need to reset?

venarius: I will have a look at it and check if there are more things that need to be reset 👍

venarius: @ZogStriP I tried a lot of things but all other user settings seem fine to me

PR #7237, opened 2019-03-21
 FIX: discourse client should know about Logster (simpler approach)

OsamaSayegh:
A simpler way to achieve the same thing as https://github.com/discourse/discourse/commit/bfcbc4d2d679a4ddc45cdbf6cab164707bd4ff5e

Thanks to @davidtaylorhq for the tip ❤️


SamSaffron:
Nice, yes way simpler, thanks!

PR #7238, opened 2019-03-21
FIX: Prevent critical emails bypassing disable, and improve email test logic

davidtaylorhq:
- The test_email job is removed, because it was always being run synchronously (not in sidekiq)
- 34b29f62 added a bypass for critical emails, to match the (previously broken) spec. This removes the bypass, and removes the spec.
- This adapts the specs for 72ffabf6, so that they check for emails being sent
- This reimplements c2797921, allowing test emails to be sent even when emails are disabled


spec/components/email/sender_spec.rb
```diff
@@ -22,6 +22,12 @@
         message = Mail::Message.new(to: moderator.email , body: "hello")
         Email::Sender.new(message, :admin_login).send
       end
+
+      it "delivers mail when mails are disabled but the email_type is test_message" do
+        Mail::Message.any_instance.expects(:deliver_now).once
```
tgxworld: Let's test for the actual mail delivery here instead of using stubs.

app/controllers/admin/email_controller.rb
```diff
@@ -10,14 +10,10 @@ def index
...
-      Jobs::TestEmail.new.execute(to_address: params[:email_address])
-      if SiteSetting.disable_emails == "yes"
-        render json: { sent_test_email_message: I18n.t("admin.email.sent_test_disabled") }
-      elsif SiteSetting.disable_emails == "non-staff" && !User.find_by_email(params[:email_address])&.staff?
-        render json: { sent_test_email_message: I18n.t("admin.email.sent_test_disabled_for_non_staff") }
-      else
-        render json: { sent_test_email_message: I18n.t("admin.email.sent_test") }
-      end
+      message = TestMailer.send_test(params[:email_address])
+      Email::Sender.new(message, :test_message).send.inspect
```
tgxworld: Do we need to `inspect` here?

tgxworld:
Thanks for fixing what I broke @davidtaylorhq :heart_eyes: Merged in https://github.com/discourse/discourse/commit/a9d5ffbe3d9b40db92a745dd588f2eaedc849c5f

app/controllers/admin/email_controller.rb
davidtaylorhq: Oooh, that was left over from debuggging. Thanks for fixing!

PR #7241, opened 2019-03-22
FIX: adds prominent focus indicator, uniform across browsers

edieblu:
A detailed explanation of the changes made here: https://meta.discourse.org/t/any-space-for-a-junior-developer/111365/35


SamSaffron:
Thanks @edieblu  ! 

@awesomerobot are there any things you want to strategically merge here or should we wait on it?

PR #7242, opened 2019-03-22
FIX: check if user is already allowed before adding to topic allowed users

majakomel:


app/models/invite_redeemer.rb
```diff
@@ -106,14 +106,14 @@ def get_existing_user
 
   def add_to_private_topics_if_invited
     invite.topics.private_messages.each do |t|
-      t.topic_allowed_users.create(user_id: invited_user.id)
+      t.topic_allowed_users.create(user_id: invited_user.id) unless t.topic_allowed_users.exists?(user_id: invited_user.id)
```
ZogStriP: Could we do this in 1 SQL query instead of ~2N (N = # of PMs) ?

app/models/invite_redeemer.rb
```diff
@@ -106,14 +106,14 @@ def get_existing_user
...
     end
   end
 
   def add_user_to_invited_topics
     Invite.where('invites.email = ? and invites.id != ?', invite.email, invite.id).includes(:topics).where(topics: { archetype: Archetype::private_message }).each do |i|
       i.topics.each do |t|
-        t.topic_allowed_users.create(user_id: invited_user.id)
+        t.topic_allowed_users.create(user_id: invited_user.id) unless t.topic_allowed_users.exists?(user_id: invited_user.id)
```
ZogStriP: Same here. Nested `each` scares me 😨

app/models/invite_redeemer.rb
```diff
@@ -106,18 +106,27 @@ def get_existing_user
...
   def add_user_to_invited_topics
     Invite.where('invites.email = ? and invites.id != ?', invite.email, invite.id).includes(:topics).where(topics: { archetype: Archetype::private_message }).each do |i|
       i.topics.each do |t|
-        t.topic_allowed_users.create(user_id: invited_user.id)
+        insert_topic_allowed_users(t.id)
       end
     end
   end
 
+  def insert_topic_allowed_users(topic_id)
```
ZogStriP: Oh, that's not what I meant. I meant one SQL query to avoid the each loop(s).

PR #7245, opened 2019-03-23
FIX: Onebox discourse user not respecting enable names

venarius:
This fixes a bug in which the onebox preview of a discourse user displays the user's real name even if the `enable names` site setting is set to false.

Resolves: https://meta.discourse.org/t/users-full-name-are-shown-in-preview-even-with-enable-names-off/112386


techAPJ:
Looks good 👍

coffeeneed:
@venarius Hi. Is this fix released as of 2.3.0.beta9 ?

PR #7246, opened 2019-03-24
FEATURE: user and group cards on mobile

hnb-ku:
This PR adds the ability to show user and group cards on mobile instead of directly navigating to user profiles or group pages.

**history:**
https://meta.discourse.org/t/mobile-user-cards/105009

**Screenshots:**

User card with details 

User card with details and background

User card new user

Group card

A background cloak is added for two purposes:

1- darken the content behind the cards to give them more focus
2- capture all clicks - touches - outside the card. A click anywhere on the cloak closes the card - matches modal behaviour.

The cards are always centered in the center of the viewport horizontally and are positioned 10% away from the top of the viewport vertically. 

There are two reasons for the vertical position:

1- it leaves more space for the cards to expand vertically on small devices
2- it matches modal behaviour.

The PR adds a subtle animation as visual feedback in case the card takes a while to load - for bad connections. Users on good connections will not notice it.

https://meta-s3-cdn.freetls.fastly.net/original/3X/6/0/601a702f735822f7d3a9e6a556e8bad2a949ec21.mp4

This PR also includes a fix for an issue where large / small group flair images display incorrectly as discussed here

https://meta.discourse.org/t/team-avatar-broken/112237


SamSaffron:
@jjaffeux  can you help @hnb-ku  here with the tests and get this merged in?

PR #7248, opened 2019-03-25
FEATURE: Introducing new UI for changing User's notification levels

khalilovcmded:
## Why?

Also, the control should not appear for TL0, control should contain only normal/mute for TL1, and all 3 options for TL2.

## UI

NNN_1


app/assets/javascripts/discourse/controllers/user.js.es6
```diff
@@ -147,14 +147,9 @@ export default Ember.Controller.extend(CanCheckEmails, {
...
-    ignoreUser() {
+    trackUser(state) {
       const user = this.get("model");
-      user.ignore().then(() => user.set("ignored", true));
-    },
-
-    unignoreUser() {
-      const user = this.get("model");
-      user.unignore().then(() => user.set("ignored", false));
+      return user.track(state);
```
eviltrout: We probably need a `catch(popupAjaxError)` for this, in case the request fails.

app/assets/javascripts/select-kit/components/user-tracking-dropdown.js.es6
```diff
@@ -0,0 +1,76 @@
+import DropdownSelectBox from "select-kit/components/dropdown-select-box";
+
+export default DropdownSelectBox.extend({
+  pluginApiIdentifiers: [""],
```
eviltrout: I think this line is unnecessary.

app/assets/javascripts/select-kit/components/user-tracking-dropdown.js.es6
```diff
@@ -0,0 +1,76 @@
...
+  classNames: ["user-tracking", "user-tracking-dropdown"],
+  nameProperty: "label",
+  allowInitialValueMutation: false,
+
+  computeHeaderContent() {
+    let content = this._super(...arguments);
+    if(this.get("user.ignored")) {
```
eviltrout: I thought prettier was supposed to catch this but we want spaces before `(` - can you confirm this code was linted?

app/assets/javascripts/select-kit/components/user-tracking-dropdown.js.es6
```diff
@@ -0,0 +1,76 @@
...
+      this.set("headerIcon", "user");
+      content.name = `${I18n.t("user.user_tracking_normal_option")}`;
+    }
+    return content;
+  },
+
+  computeContent() {
+    const content = [];
+
+    content.push({
```
eviltrout: Rater than creating an array and calling `push`, you could return the array as a literal:

```
return [
  firstThing,
  secondThing,
  etc
]

File: app/controllers/users_controller.rb [hunk 5]
```diff
@@ -995,18 +995,21 @@ def destroy
...
     raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
-    guardian.ensure_can_ignore_user!(params[:ignored_user_id])
 
-    IgnoredUser.find_or_create_by!(user: current_user, ignored_user_id: params[:ignored_user_id])
-    render json: success_json
-  end
-
-  def unignore
-    raise Discourse::NotFound unless SiteSetting.ignore_user_enabled
+    if params[:state] == "ignore"
```
eviltrout: This method definitely needs some tests :)

khalilovcmded: Sure 👍 It is in Draft :) Needed to get some opinions first.

app/assets/javascripts/select-kit/components/user-tracking-dropdown.js.es6
khalilovcmded: Yea, I left here just to get thoughts, what do we need or use this line for?

app/assets/javascripts/select-kit/components/user-tracking-dropdown.js.es6
khalilovcmded: Yes, sure, just wanted to leave linting, etc out since this is a draft. I am more interested in overall approach validation 👍

app/controllers/users_controller.rb
```diff
@@ -995,18 +995,21 @@ def destroy
     render json: success_json
   end
 
-  def ignore
+  def track
```
ZogStriP: I think this should be changed to `notification_level` to maintain consistency with "Topic Notification Levels".

Plus we're not really "tracking" any users here. Quite the opposite actually 😉

app/serializers/user_serializer.rb
```diff
@@ -281,6 +282,10 @@ def ignored
     IgnoredUser.where(user_id: scope.user&.id, ignored_user_id: object.id).exists?
   end
 
+  def muted
+    MutedUser.where(user_id: scope.user&.id, muted_user_id: object.id).exists?
```
ZogStriP: Any chances this could introduce an N+1?

khalilovcmded: We have this for `ignored` too. I think this recapping from our discussion about introducing similar query for `ignored`, this serializer seems to be used in the User profile but not `topics` for an instance (which is using a lighter serializer version).

ZogStriP: Might want to check that out

khalilovcmded: discourse____discourse_discourse__-_____lib_oneboxer_rb__discourse_

Screenshot_25_03_2019__3_10_p_m_

khalilovcmded: I feel it is safe.

SamSaffron:
I am not a big fan of this UX, thing on top is a "button", thing below is a "button", thing in the middle is a dropdown that looks like a button. 

I feel we should just say "what the status is" and link that directly to `/my/preferences/users` which is where you manage this stuff. I like forcing people to that page, cause when they see it they are forced to think that they are ignoring / muting lots of users. 

I get that twitter loves making ignores and mutes super duper easy, but not sure if this is a good comparison cause they are not really a community. 

@coding-horror ?

coding-horror:
It's not really surprising though because clicking it takes no action. I
don't view this as a problem, we are teaching people they have levels to
work with here, from nothing to mute to ignore.. proceed with the mockup
for now.

khalilovcmded:
@SamSaffron from a UX point of view, this can be solved via adding a arrow-down separator next to the button that clearly hints at this being a dropdown button with options:

Dropdowns_·_Bootstrap

app/controllers/users_controller.rb
khalilovcmded: Good suggestion, changed it 👍

khalilovcmded:
@awesomerobot do you reckon I should keep the `ignore` selection in red like this;

Profile_-_matt_-_Discourse

lib/guardian.rb
```diff
@@ -390,6 +390,17 @@ def can_export_entity?(entity)
     UserExport.where(user_id: @user.id, created_at: (Time.zone.now.beginning_of_day..Time.zone.now.end_of_day)).count == 0
   end
 
+  def can_mute_user?(user_id)
+    can_mute_users? && @user.id != user_id && User.where(id: user_id,
```
ZogStriP: Indentation is a bit weird. I would do each condition on its own line.


``` ruby
def can_mute_user?(user_id)
  can_mute_users? &&
  @user.id != user_id &&
  User.where(id: user_id, admin: false, moderator: false).exists?
end
```

spec/components/guardian_spec.rb
```diff
@@ -2698,6 +2699,61 @@
...
+      it 'allows muting user' do
+        expect(guardian.can_mute_user?(another_user.id)).to eq(true)
+      end
+    end
+
+    context "when muters's trust level is tl1" do
+      let(:guardian) { Guardian.new(trust_level_1) }
+      let!(:another_user) { Fabricate(:user) }
+
+      it 'allows mutring user' do
```
ZogStriP:
```suggestion
      it 'allows muting user' do
```

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -615,17 +615,13 @@ const User = RestModel.extend({
...
-    });
-  },
-
-  unignore() {
-    return ajax(`${userPath(this.get("username"))}/ignore.json`, {
-      type: "DELETE",
-      data: { ignored_user_id: this.get("id") }
+      data: {
+        notification_level: level,
+        notification_level_user_id: this.get("id")
```
ZogStriP: Do we actually need that parameter since we already know what user we're muting via their username in the route?

khalilovcmded: Good point. On the `UsersController`, it seems that we depend on `fetch_user_from_params`, I will do so.

PR #7252, opened 2019-03-26
UX: show which groups are missing permissions for parent category

majakomel:


app/models/category.rb
```diff
@@ -666,7 +666,10 @@ def check_permissions_compatibility(parent_permissions, child_permissions)
     child_groups = child_permissions.map(&:first)
     only_in_subcategory = child_groups - parent_groups
 
-    errors.add(:base, I18n.t("category.errors.permission_conflict")) if only_in_subcategory.present?
+    if only_in_subcategory.present?
```
ZogStriP: `only_in_subcategory` does not look like it's an array. Maybe something like `only_subcategory_groups`?

SamSaffron:
better error messages are nice, thanks

PR #7255, opened 2019-03-26
FIX: flaky test in reports

khalilovcmded:


spec/models/report_spec.rb
```diff
@@ -1128,10 +1121,20 @@ def expect_row_to_be_equal(row, user, ignores, mutes)
...
         it "works" do
           expect(report.data.length).to eq(2)
-          expect_row_to_be_equal(report.data[0], john, 1, 1)
-          expect_row_to_be_equal(report.data[1], matt, 1, 1)
+          expect_row_to_be_equal(report.data, john, 1, 1)
+          expect_row_to_be_equal(report.data, matt, 1, 1)
         end
       end
+
+      def expect_row_to_be_equal(data, user, ignores, mutes)
```
ZogStriP: 2 methods with the same name but with different signatures? That might be confusing.

khalilovcmded: These methods are defined within a their own scopes though, are we worried about scope leakage? or just the confusion from the same name?

PR #7261, opened 2019-03-26
FIX: Switching composer action shouldn't trigger abandon popup

venarius:
Switching the composer action shouldn't trigger the abandon popup.

Resolves: https://meta.discourse.org/t/abandoning-draft-when-switching-from-user-reply-to-general-reply-doesnt-remove-text/112482


ZogStriP:
Thanks 👍

PR #7263, opened 2019-03-27
FIX: Don't index posts with empty `Post#raw` for search.

tgxworld:


app/jobs/scheduled/reindex_search.rb
```diff
@@ -60,8 +61,15 @@ def rebuild_problem_tags(limit = 10000)
 
     private
 
+    def clean_post_search_data
```
tgxworld: Putting this in a scheduled job for now instead of a migration because I'm predicting that we'll need to do a bunch of clean up while I look into search.

spec/models/post_spec.rb
```diff
@@ -1311,4 +1311,18 @@ def post_with_body(body, user = nil)
...
+    it "should return true for regular posts and whispers" do
+      results = {}
+
+      Post.types.each do |name, post_type|
+        post = Fabricate(:post, post_type: post_type)
+        results[name] = post.indexable?
+      end
+
+      expect(results.reject { |_, v| v == true }.keys)
+        .to contain_exactly(:moderator_action, :small_action)
```
tgxworld: This is intentional so the spec will fail when someone adds a new post action type in the future.

PR #7266, opened 2019-03-27
FEATURE: Introduce ignore duration selection

khalilovcmded:
## Why?

## UI

Ignore User 1


app/assets/javascripts/discourse/templates/modal/ignore-duration.hbs
```diff
@@ -0,0 +1,17 @@
+{{#d-modal-body title="user.user_notifications.ignore_duration_title" autoFocus="false"}}
+  {{future-date-input
+    label="user.user_notifications.ignore_duration_when"
+    input=ignoredUntil
+    statusType=selection
+    includeWeekend=true}}
+  <p>Please note that all ignores are automatically removed after the ignore duration expires.</p>
```
ZogStriP: This should be internationalized.

app/assets/javascripts/select-kit/components/user-notifications-dropdown.js.es6
```diff
@@ -73,13 +74,14 @@ export default DropdownSelectBox.extend({
...
     ignore() {
-      this.get("updateNotificationLevel")("ignore")
-        .then(() => {
-          this.set("user.ignored", true);
-          this.set("user.muted", false);
+      const controller = showModal("ignore-duration", {
+        model: this.get("user")
+      });
+      controller.setProperties({
+        refreshHeaderContent: () => {
```
ZogStriP: Why is this a property and not just a function?

khalilovcmded: So I need to be a property that's containing a function because the `modal` controller doesn't have access to: `computeHeaderContent` of the current component.

What's your suggestion?

app/assets/javascripts/discourse/controllers/ignore-duration.js.es6
```diff
@@ -0,0 +1,30 @@
...
+        );
+        return;
+      }
+      this.set("loading", true);
+      this.get("model")
+        .updateNotificationLevel("ignore", this.get("ignoredUntil"))
+        .then(() => {
+          this.set("model.ignored", true);
+          this.set("model.muted", false);
+          this.get("refreshHeaderContent")();
```
jjaffeux: Does it work:

```
this.refreshHeaderContent()
```

PR #7267, opened 2019-03-27
FEATURE: Let users delete their own topics.

nbianca:


lib/guardian/post_guardian.rb
```diff
@@ -168,7 +168,7 @@ def can_edit_post?(post)
 
   # Deleting Methods
   def can_delete_post?(post)
-    can_see_post?(post)
+    return false if !can_see_post?(post)
```
khalilovcmded: Maybe `return false unless can_see_post?(post)` ?

lib/guardian/topic_guardian.rb
```diff
@@ -86,13 +86,17 @@ def can_edit_topic?(topic)
 
   # Recovery Method
   def can_recover_topic?(topic)
-    topic && topic.deleted_at && topic.user && is_staff?
+    if is_staff?
+      topic && topic.deleted_at && topic.user
```
eviltrout: Although this is truthy, I always prefer apis that return `true` or `false` values.

lib/post_destroyer.rb
```diff
@@ -157,10 +157,17 @@ def mark_for_deletion(delete_removed_posts_after = SiteSetting.delete_removed_po
...
+        @post.revise(@user,
+          { raw: I18n.t('js.topic.deleted_by_author', count: delete_removed_posts_after) },
+          force_new_version: true
+        )
+      else
+        @post.revise(@user,
+          { raw: I18n.t('js.post.deleted_by_author', count: delete_removed_posts_after) },
+          force_new_version: true
+        )
+      end
```
tgxworld: One slight nitpick here, the only deference between the lines added here is the key of the translation. I would just go with something like this to remove the deplication.

```
        @post.revise(@user,
          { raw: I18n.t("js.#{@post.is_first_post? ? "topic" : "post"}.deleted_by_author", count: delete_removed_posts_after) },
          force_new_version: true
        )
```

spec/components/guardian_spec.rb
```diff
@@ -1759,11 +1759,17 @@
...
       end
 
-      it "returns false when it's the OP, even as a moderator" do
+      it "returns true when it's the OP" do
         post.update_attribute :post_number, 1
         expect(Guardian.new(moderator).can_delete?(post)).to be_falsey
       end
 
+      it "returns false when it's the OP, even as a moderator if there are at least two posts" do
+        post.update_attribute :post_number, 1
```
tgxworld:
```suggestion
        post.update!(post_number: 1)
```

lib/post_destroyer.rb
gschlager: I'm not a fan of concatenating translation keys. That makes it so much harder to search for them. I guess the middle ground would be a solution like this. 😉 

```ruby
        raw_key = @post.is_first_post? ? 'js.topic.deleted_by_author' : 'js.post.deleted_by_author'
        @post.revise(@user,
          { raw: I18n.t(raw_key, count: delete_removed_posts_after) },
          force_new_version: true
        )
```

config/locales/client.en.yml
```diff
@@ -2135,6 +2135,10 @@ en:
           one: You have selected <b>1</b> post.
           other: "You have selected <b>{{count}}</b> posts."
 
+      deleted_by_author:
```
gschlager: `post.deleted_by_author` is used by https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/post.js.es6

Should the client use `topic.deleted_by_author` as well? Otherwise `server.en.yml` would be a better place for this string.

lib/post_destroyer.rb
tgxworld: This is ok too :)

Crare:
Still can't delete topic.
Is there some kind of rule that if it has replies, you can't delete the topic, like discussed in meta?
I think it's a terrible rule, user's should have the right to delete the whole topic they made if they ever wanted to.

PR #7268, opened 2019-03-27
UX: improves local-dates form

jjaffeux:


plugins/discourse-local-dates/assets/javascripts/discourse/components/discourse-local-dates-create-form.js.es6
```diff
@@ -15,23 +15,154 @@ export default Ember.Component.extend({
...
+  },
+
+  _renderPreview: debounce(function() {
+    const markup = this.get("markup");
+
+    if (markup) {
+      cookAsync(markup).then(result => {
+        this.set("currentPreview", result);
+
+        Ember.run.next(() =>
```
eviltrout: I always dislike this kind nested `Ember.run` - is the issue that when the renderPreview is triggered the element it renders into might not exist yet?

plugins/discourse-local-dates/assets/javascripts/discourse/templates/components/discourse-local-dates-create-form.hbs
```diff
@@ -22,20 +39,30 @@
...
             <div class="controls">
-              {{input type="time" value=time class="time-input"}}
+              {{input input=(mut time) type="time" value=time class="time-input"}}
             </div>
           </div>
         </div>
 
-        <span class="to-indicator"><span>{{i18n "discourse_local_dates.create.form.to"}}</span></span>
+        <div class="to-indicator">
+          {{#if site.mobileView}}
```
eviltrout: This looks like a good candidate for ember's inline-if

ZogStriP:
Where are the before / after screenshots? 😉

plugins/discourse-local-dates/assets/javascripts/discourse/templates/components/discourse-local-dates-create-form.hbs
jjaffeux: indeed, I never use it 👍

plugins/discourse-local-dates/assets/javascripts/discourse/components/discourse-local-dates-create-form.js.es6
jjaffeux: I think this is not needed anymore, it's the result of a refactoring where I have removed something in between

jjaffeux:
> Where are the before / after screenshots? 😉

well in the dedicated topic :p

PR #7269, opened 2019-03-27
 FEATURE: Add ignore duration selection for user notifications preferences page

khalilovcmded:
## Why?

## UI

Smoother un-ignore


app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -1,10 +1,44 @@
...
 import { popupAjaxError } from "discourse/lib/ajax-error";
+import showModal from "discourse/lib/show-modal";
+import User from "discourse/models/user";
 
 export default Ember.Controller.extend(PreferencesTabController, {
   saveAttrNames: ["muted_usernames", "ignored_usernames"],
-
+  ignoredUsernames: Ember.computed.alias("model.ignored_usernames"),
   actions: {
+    ignoredUsernamesChanged(previous, current) {
```
jjaffeux: If possible I would change naming here. It feels like an event, when it's actually an action. This name doesn’t tell me what it's doing.

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -10,7 +10,9 @@
     <div class="controls tracking-controls">
       <label>{{d-icon "eye-slash" class="icon"}} {{i18n 'user.ignored_users'}}</label>
       {{user-selector excludeCurrentUser=true
-                      usernames=model.ignored_usernames class="user-selector"}}
+                      usernames=model.ignored_usernames
+                      onChangeCallback=(action "ignoredUsernamesChanged")
```
jjaffeux: I think you can get rid of "Callback" here. `onChange` is descriptive enough

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -1,10 +1,44 @@
...
+            if (user.get("ignored")) {
+              return;
+            }
+            const controller = showModal("ignore-duration", {
+              model: user
+            });
+            controller.setProperties({
+              onClose: () => {
+                if (!user.get("ignored")) {
+                  const usernames = this.get("ignoredUsernames")
```
jjaffeux: can we do the split only once ?

app/assets/javascripts/discourse/components/user-selector.js.es6
```diff
@@ -94,11 +94,20 @@ export default TextField.extend({
             }
             return i.username ? i.username : i;
           });
+
+          let previouslySelected = [];
```
jjaffeux: can you add a comment explaining what it's doing here ? it's not very clear at first sight.

user-selector will be replaced at some point and it would help ensure wee keep consistent behavior.

khalilovcmded: That makes sense, let me add it 👍

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
khalilovcmded: Good point 👍

khalilovcmded:
Had a chat with @jjaffeux , I will merge this now and apply his feedback in a different PR.

PR #7270, opened 2019-03-27
FEATURE: change layout when default page is category to tabular for _…

mrfinch:
…escaped_fragment_ page

This is a part of https://meta.discourse.org/t/improving-discourse-static-html-archive/112497

The old layout can be viewed at https://meta.discourse.org/categories/?_escaped_fragment_

I have removed `itemprop='item'` since it was not needed (all information is provided in ListItem already) and was showing up as an error here: https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fmeta.discourse.org%2Fcategories%2F%3F_escaped_fragment_

Attaching screenshot with the new layout.

Screenshot from 2019-03-27 22-24-51

### DIFF
--- a/app/assets/stylesheets/desktop/category-list.scss
+++ b/app/assets/stylesheets/desktop/category-list.scss
@@ -109,6 +109,14 @@
     .latest {
       padding: 0 0 10px 10px;
     }
+
+    .blue-border {
+      border-color: #0088cc;
+    }
+
+    .grey-border {
+      border-color: #808281;
+    }
   }
 }
 
--- a/app/assets/stylesheets/desktop/category-list.scss
+++ b/app/assets/stylesheets/desktop/category-list.scss
@@ -110,12 +110,12 @@
       padding: 0 0 10px 10px;
     }
 
-    .blue-border {
-      border-color: #0088cc;
+    .secondary-border {
+      border-color: $secondary-high;
     }
 
-    .grey-border {
-      border-color: #808281;
+    .primary-border {
+      border-color: $primary-high;
     }
   }
 }
--- a/app/assets/stylesheets/desktop/category-list.scss
+++ b/app/assets/stylesheets/desktop/category-list.scss
@@ -109,14 +109,6 @@
     .latest {
       padding: 0 0 10px 10px;
     }
-
-    tr:nth-child(odd) {
-      border-color: $secondary-high;
-    }
-
-    tr:nth-child(even) {
-      border-color: $primary-high;
-    }
   }
 }
 
--- a/app/assets/stylesheets/desktop/category-list.scss
+++ b/app/assets/stylesheets/desktop/category-list.scss
@@ -110,11 +110,11 @@
       padding: 0 0 10px 10px;
     }
 
-    .secondary-border {
+    tr:nth-child(odd) {
       border-color: $secondary-high;
     }
 
-    .primary-border {
+    tr:nth-child(even) {
       border-color: $primary-high;
     }
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/desktop/category-list.scss
```diff
@@ -109,6 +109,14 @@
     .latest {
       padding: 0 0 10px 10px;
     }
+
+    .blue-border {
+      border-color: #0088cc;
```
ZogStriP: Here, we should be using the colors in the scss variables. That way, if the admins change their theme, the colors will change as well.

mrfinch: i have fixed it now @ZogStriP

app/views/categories/index.html.erb
```diff
@@ -1,17 +1,34 @@
...
+  <table class='category-list'>
+    <thead>
+      <tr>
+        <th class='category'><%= t 'js.categories.category' %></th>
+        <th class='topics'><%= t 'js.topic.list' %></th>
+      </tr>
+    </thead>
+    <tbody>
+      <meta itemprop='itemListOrder' content='http://schema.org/ItemListOrderDescending'>
+      <% @category_list.categories.each_with_index do |c, index| %>
```
ZogStriP: You don't need to keep track of the index just to set an odd/even class. You can do that using CSS only 😉 

https://www.w3.org/Style/Examples/007/evenodd.en.html

mrfinch: Haha, actually I didn't add an index in the loop. It's already used in this line: ` <meta itemprop='position' content='<%= index %>'>` , we need it for SEO @ZogStriP

ZogStriP: My bad, I missed it. You should still use CSS to do the odd/even class ;)

mrfinch: Haha, I have made the change @ZogStriP

ZogStriP: Sorry, I just looked at your screenshot and I wonder: why aren't we using the category color for the border?

mrfinch: Actually, I didn't know that the category had colour stored. I have fixed it now, thanks @ZogStriP  :bowing_man:

app/views/categories/index.html.erb
```diff
@@ -1,17 +1,34 @@
...
+        <tr>
+          <td class='category' style='border-color: #<%= c.color %>;'>
+            <div itemprop='itemListElement' itemscope itemtype='http://schema.org/ListItem'>
+              <meta itemprop='position' content='<%= index %>'>
+              <meta itemprop='url' content='<%= "#{c.url}" %>'>
```
ZogStriP:
```suggestion
              <meta itemprop='url' content='<%= c.url %>'>
```

app/views/categories/index.html.erb
```diff
@@ -1,17 +1,34 @@
...
+              <h3>
+                <a href='<%= "#{c.url}" %>'>
```
ZogStriP:
```suggestion
                <a href='<%= c.url %>'>
```

app/views/categories/index.html.erb
```diff
@@ -1,17 +1,34 @@
...
+                  <span itemprop='name'><%= c.name %></span>
+                </a>
+              </h3>
+              <div itemprop='description'><%= c.description&.html_safe %></div>
+            </div>
+          </td>
+          <td class='topics'>
+            <div title='<%= "#{c.topic_count}" %> <%= t 'js.topic.list' %>'><%= c.topic_count %></div>
```
ZogStriP:
```suggestion
            <div title='<%= c.topic_count %> <%= t 'js.topic.list' %>'><%= c.topic_count %></div>
```

app/views/categories/index.html.erb
mrfinch: Actually, it's an old code. I had just changed the indentation :sweat_smile:, I have fixed all your suggestions @ZogStriP

ZogStriP:
Thanks 👍

PR #7271, opened 2019-03-27
DEV: Add simplecov as test dependency

gschlager:
@SamSaffron Are you okay with adding `simplecov` as dependency for tests? I've been using it quite often lately and having it in the `Gemfile` would make my life a little bit easier. Currently I'm always manually editing the `Gemfile` and I have to make sure that I do not commit those changes by accident.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,6 +121,7 @@ group :test do
   gem 'fakeweb', '~> 1.3.0', require: false
   gem 'minitest', require: false
   gem 'danger'
+  gem 'simplecov', require: false
 end
 
 group :test, :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,6 +115,7 @@ GEM
       image_size (~> 1.5)
       in_threads (~> 1.3)
       progress (~> 3.0, >= 3.0.1)
+    docile (1.3.1)
     email_reply_trimmer (0.1.12)
     ember-data-source (3.0.2)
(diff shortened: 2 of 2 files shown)


eviltrout:
I'm not Sam but it does seem a bit odd to me to introduce a library that is not used anywhere in our codebase. What's to stop someone in the future from saying, oh this isn't used *delete*.

If we want to start tracking code coverage I think it should be exposed somehow.

gschlager:
Well, it's a test dependency. You could argue that other libs like danger aren't "used" in the codebase either. Anyway, there's actually a reference to `simplecov` in our codebase. But it's hidden behind an env variable.

https://github.com/discourse/discourse/blob/fa5a15868320b3f45777db25b72ac5293db3ff89/spec/rails_helper.rb#L1-L4

There was also a request on Meta to add that dependency a few months back.
https://meta.discourse.org/t/setting-up-plugin-continuous-integration-tests-on-travis-ci/59612/13

tgxworld:
Perhaps something like what we do for the bullet gem? 

https://github.com/discourse/discourse/blob/master/Gemfile#L146

tgxworld:
> Unfortunately that will create a changed Gemfile.lock on bundle install which isn't really better than the alternative of manually editing the Gemfile everytime

O that is right. I don't see an additional Gem in development being a big problem. We already have quite a number of gems that is only required with an ENV.

SamSaffron:
fine, merging it in.

PR #7272, opened 2019-03-27
FEATURE: Delegated authentication via user api keys

pmusaraj:
For details, see: https://meta.discourse.org/t/delegated-authentication-for-discourse-mobile-app/110414

This adds a new capability to the user API: delegated authentication via a `one time password` mechanism. This allows apps (like the DiscourseHub mobile app) to request a one-use password and use that to authenticate the user in an isolated browser session (like SafariViewController, for example).


app/controllers/application_controller.rb
```diff
@@ -359,6 +360,25 @@ def set_mobile_view
     session[:mobile_view] = params[:mobile_view] if params.has_key?(:mobile_view)
   end
 
+  def handle_safari_svc_session
+    # Workaround for dropped cookies in DiscourseHub SafariViewController iOS 12+.
```
SamSaffron: do we want to call this "safari" specific I agree safari is the symptom but really any client could opt for the same magic no?

pmusaraj: You're right, of course, no need to call this Safari anything. My initial hunch was to explain the use case for this now, but it's already usable by any client. (Which is also why there is an `auth_redirect` parameter, to let clients use their own URL here.)

app/controllers/application_controller.rb
```diff
@@ -46,6 +46,7 @@ def handle_unverified_request
   before_action :clear_notifications
   before_action :set_locale
   before_action :set_mobile_view
+  before_action :handle_delegated_auth
```
SamSaffron: can we avoid adding an extra `before action` I am pretty protective of having too many ? maybe just do this direct in current user provider ?

config/site_settings.yml
```diff
@@ -1919,7 +1919,7 @@ user_api:
   allow_user_api_keys:
     default: true
   allow_user_api_key_scopes:
-    default: "read|write|message_bus|push|notifications|session_info"
+    default: "read|write|message_bus|push|notifications|session_info|one_time_password"
```
SamSaffron: what is the transition period going to look like now that existing keys kind of want to have one_time_password?

SamSaffron:
Overall I am very happy with this change, feel free to merge once you address my small niggles

config/site_settings.yml
pmusaraj: It's pretty straightforward: sites without the one_time_password scope won't get the redirect treatment. This is valid for both sites that aren't up to date as well as sites that decide to disallow that scope. 

The PR also bumps the user API version to 4, so apps can check whether `one_time_password` is supported by checking the API version.

PR #7273, opened 2019-03-27
DEV: Add vagrant directory to gitignore

venarius:
This adds the vagrant directory to the gitignore. Currently I always have to watch to not commit it and probably other people use vagrant too.


eviltrout:

PR #7274, opened 2019-03-28
FIX: Use all flag types in moderators activity query

scossar:
Currently, the Moderators Activity report is excluding `notify_moderators` flag types. This is the type of flag that is created when a user selects 'Something Else' from the flagging modal.

Using `PostActionType.flag_types` instead of `PostActionType.flag_types_without_custom` in the `period_actions` query will allow these types of flags to be counted when a moderator agrees or disagrees with them.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -885,7 +885,7 @@ class Report
         SELECT agreed_by_id,
         disagreed_by_id
         FROM post_actions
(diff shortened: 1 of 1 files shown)


eviltrout:
This report might need to be re-written now that the reviewable queue is merged in.

scossar:
Right! I just saw the PR to move each report in its own file. I'll close this now and create a new PR when that has been merged. It's not something that's urgent to get changed.

PR #7276, opened 2019-03-28
FIX: Uncategorized pm not allowing edit

venarius:
When your Site Settings disallowed uncategorized posts and you edit the first post of a private message and removed all the text, Composer would then block input from happening because of "You must select a category before typing here". This has been fixed.

Resolves: https://meta.discourse.org/t/you-must-select-category-notice-when-editing-pm-op/112741

### DIFF
--- a/test/javascripts/acceptance/composer-test.js.es6
+++ b/test/javascripts/acceptance/composer-test.js.es6
@@ -674,7 +674,7 @@ QUnit.test(
 
     assert.equal(
       find(".d-editor-container textarea").attr("placeholder"),
-      "Type here. Use Markdown, BBCode, or HTML to format. Drag or paste images.",
+      I18n.t("composer.reply_placeholder"),
       "it should not block because of missing category"
     );
   }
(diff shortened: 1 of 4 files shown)


test/javascripts/acceptance/composer-test.js.es6
```diff
@@ -659,6 +659,27 @@ QUnit.test("Loading draft also replaces the recipients", async assert => {
...
+    await visit("/t/34");
+
+    await click("#post_1 .d-icon-ellipsis-h");
+
+    await click("#post_1 .d-icon-pencil-alt");
+
+    await fillIn(".d-editor-input", "");
+
+    assert.equal(
+      find(".d-editor-container textarea").attr("placeholder"),
```
ZogStriP: Is this actually testing anything? I don't think we ever change the value of the `placeholder' attribute of the composer `textarea` element.

venarius: Of course it is testing something. Wenn you comment out the fix the placeholder will be "You must select a category before typing here". I mosty write test driven.

ZogStriP: I see, can this call `I18n.t` instead of copy-pasting the string? If we ever change that string, this test will fail for no good reason.

tgxworld:
Looks good to me. Thank you for the contribution @venarius  :+1:

PR #7280, opened 2019-03-28
FEATURE: Send user activation reminders.

nbianca:


app/jobs/scheduled/resend_validation_emails.rb
```diff
@@ -0,0 +1,17 @@
+module Jobs
+  class ResendValidationEmails < Jobs::Scheduled
+    every 2.days
+
+    def execute(args)
+      User.inactive.where('created_at < ?', 2.days.ago).each do |user|
```
ZogStriP: Isn't this going to send a reminder email to all inactive users every 2 days?

app/jobs/scheduled/resend_validation_emails.rb
```diff
@@ -0,0 +1,22 @@
+module Jobs
+  class ResendValidationEmails < Jobs::Scheduled
+    every 2.days
```
ZogStriP: This should run more frequently otherwise we might send reminder after (almost) 4 days.

```suggestion
    every 2.hours
```

app/jobs/scheduled/resend_validation_emails.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def execute(args)
+      User.joins("LEFT JOIN user_custom_fields ON users.id = user_id AND user_custom_fields.name = 'activation_reminder'")
+        .where(active: false, user_custom_fields: { value: nil })
+        .where('users.created_at < ?', 2.days.ago)
+        .each do |user|
```
ZogStriP: We should not be loading too much users with these conditions but it's better to always use `find_each` when a specific order is not required.

```suggestion
        .find_each do |user|
```

app/models/user.rb
```diff
@@ -180,6 +180,7 @@ class User < ActiveRecord::Base
   scope :suspended, -> { where('suspended_till IS NOT NULL AND suspended_till > ?', Time.zone.now) }
   scope :not_suspended, -> { where('suspended_till IS NULL OR suspended_till <= ?', Time.zone.now) }
   scope :activated, -> { where(active: true) }
+  scope :inactive, -> { where(active: false) }
```
ZogStriP: Is this used anywhere?

spec/jobs/resend_validation_emails_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it 'should email inactive users' do
+    user = Fabricate(:user, active: false, created_at: 3.days.ago)
+
+    expect { described_class.new.execute({}) }
+      .to change { EmailLog.count }.by(1)
+      .and change { user.email_tokens.count }.by(1)
+
+    expect(user.custom_fields['activation_reminder']).to be_truthy
```
ZogStriP: We try not to use `be_truthy` and `be_falsy` anymore

```suggestion
    expect(user.custom_fields['activation_reminder']).to eq(true)
```

spec/jobs/resend_validation_emails_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+describe Jobs::ResendValidationEmails do
+  before do
+    Jobs.run_immediately!
+  end
+
+  it 'should email inactive users' do
+    user = Fabricate(:user, active: false, created_at: 3.days.ago)
+
+    expect { described_class.new.execute({}) }
+      .to change { EmailLog.count }.by(1)
```
ZogStriP: This is not testing that we've sent the email but rather that we've created a new `EmailLog` record.

Can we instead write a test that will properly test that we've sent an email?

app/jobs/scheduled/resend_validation_emails.rb
```diff
@@ -0,0 +1,22 @@
+module Jobs
+  class ResendValidationEmails < Jobs::Scheduled
```
ZogStriP: I would change the name of the job to be more relevant to what it's doing. Something like

```suggestion
  class ActivationReminderEmails < Jobs::Scheduled
```

Don't forget to also change the name of the file to match the name of the job.

spec/jobs/resend_validation_emails_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+describe Jobs::ResendValidationEmails do
```
ZogStriP: That job doesn't exist anymore ;)

PR #7281, opened 2019-03-28
FIX: Rescue file not found hightlight.min.js

venarius:
I had users suddenly report that they were unable to access my forum. I quick look at `/logs` returned me the following error:

```
ActionView::Template::Error (No such file or directory @ rb_sysopen - /home/tla/discourse/vendor/assets/javascripts/highlightjs/highlight.min.js)
lib/highlight_js/highlight_js.rb:15:in `read'
```

I don't know why that exactly happened but my highlight.min.js went missing. I fixed the error by downloading the js file from the git repo and re-adding it.

I think it would be best just to rescue this line so that Discourse can continue with "halfway" normal operations and not totally crash.

### DIFF
--- a/lib/highlight_js/highlight_js.rb
+++ b/lib/highlight_js/highlight_js.rb
@@ -12,7 +12,12 @@ module HighlightJs
   end
 
   def self.bundle(langs)
-    result = File.read(HIGHLIGHTJS_DIR + "highlight.min.js")
+    begin
+      result = File.read(HIGHLIGHTJS_DIR + "highlight.min.js")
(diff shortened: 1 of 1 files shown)


eviltrout:
I am honestly more curious as to how it went missing. Was it following an upgrade or something manual? Did you try a `launcher rebuild`?

venarius:
Didn't try a launcher rebuild but that would have probably also fixed it.

ZogStriP:
That's the first time I've heard about this error. Let's just wait for at least 3 occurences before doing anything about it.

PR #7282, opened 2019-03-28
UX: Unified emoji escape regex

venarius:
This sets a unified regex for `emojiUnescape` which is also used in `lib/pretty-text/shims.js`

Related: https://meta.discourse.org/t/unicode-emoji-modifiers-are-handled-differently-in-titles-and-post-content/111328


eviltrout:
Thanks this looks good!

PR #7283, opened 2019-03-28
FIX: category list order using category featured topics

majakomel:
follow up on https://review.discourse.org/t/ux-order-categories-based-on-latest-activity-for-all-page-styles-7196/2271

### DIFF
--- a/app/models/category_list.rb
+++ b/app/models/category_list.rb
@@ -78,7 +78,12 @@ class CategoryList
     if SiteSetting.fixed_category_positions
       @categories = @categories.order(:position, :id)
     else
-      @categories = @categories.includes(:latest_post).order("posts.created_at DESC NULLS LAST").order('categories.id ASC')
+      allowed_category_ids = @categories.pluck(:id) << nil # `nil` is necessary to include categories without any associated topics
+      @categories = @categories.left_outer_joins(:featured_topics)
+        .where(topics: { category_id: allowed_category_ids })
+        .group('categories.id')
+        .order("max(topics.bumped_at) DESC NULLS LAST")
+        .order('categories.id ASC')
     end
 
     @categories = @categories.to_a
--- a/app/models/category_list.rb
+++ b/app/models/category_list.rb
@@ -43,6 +43,19 @@ class CategoryList
     "categories_list".freeze
   end
 
+  def self.order_categories(categories)
+    if SiteSetting.fixed_category_positions
+      categories.order(:position, :id)
+    else
+      allowed_category_ids = categories.pluck(:id) << nil # `nil` is necessary to include categories without any associated topics
+      categories.left_outer_joins(:featured_topics)
+        .where(topics: { category_id: allowed_category_ids })
+        .group('categories.id')
+        .order("max(topics.bumped_at) DESC NULLS LAST")
+        .order('categories.id ASC')
+    end
+  end
+
   private
 
   def find_relevant_topics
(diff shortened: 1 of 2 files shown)


app/models/category_list.rb
```diff
@@ -78,7 +78,12 @@ def find_categories
     if SiteSetting.fixed_category_positions
       @categories = @categories.order(:position, :id)
     else
-      @categories = @categories.includes(:latest_post).order("posts.created_at DESC NULLS LAST").order('categories.id ASC')
+      allowed_category_ids = @categories.pluck(:id) << nil
```
eviltrout: This line is a bit odd to me. It's taking the array of `id`s and then adding a value for `nil`? Why is that necessary?

spec/models/category_list_spec.rb
```diff
@@ -116,7 +116,8 @@
   end
 
   describe 'category order' do
-    let(:category_ids) { CategoryList.new(Guardian.new(admin)).categories.map(&:id) - [SiteSetting.uncategorized_category_id] }
```
eviltrout: This might be one of those few cases were adding a helper to a test could simplify things a lot. These lines are very similar and could use a function.

app/models/category_list.rb
majakomel: `nil` is necessary to include categories without any topics, equivalent for `.where("topics.category_id IN (?) OR topics.category_id IS null", allowed_category_ids)`

ZogStriP: Looks like a trick that is worth adding a comment about

spec/models/category_list_spec.rb
```diff
@@ -150,18 +155,42 @@
...
+
+        CategoryFeaturedTopic.feature_topics
+
+        expect(category_ids_admin).to eq([cat3.id, cat2.id, cat1.id, cat4.id])
       end
 
       it "returns categories in order of id when there's no activity" do
-        cat1, cat2 = Fabricate(:category, position: 1), Fabricate(:category, position: 0)
-        expect(category_ids).to eq([cat1.id, cat2.id])
+        cat1, cat2, cat3 = Fabricate(:category, position: 2), Fabricate(:category, position: 1), Fabricate(:category, position: 0)
```
ZogStriP: Small nitpick, but one category per line is much easier to read.

spec/models/category_list_spec.rb
```diff
@@ -131,16 +136,16 @@
 
       it "returns categories in specified order" do
         cat1, cat2 = Fabricate(:category, position: 1), Fabricate(:category, position: 0)
-        expect(category_ids).to eq([cat2.id, cat1.id])
+        expect(category_ids_admin).to eq([cat2.id, cat1.id])
       end
 
       it "handles duplicate position values" do
         cat1, cat2, cat3, cat4 = Fabricate(:category, position: 0), Fabricate(:category, position: 0), Fabricate(:category, position: nil), Fabricate(:category, position: 0)
```
ZogStriP: Can you do the same here? One declaration per line?

spec/models/category_list_spec.rb
```diff
@@ -116,7 +116,12 @@
   end
 
   describe 'category order' do
-    let(:category_ids) { CategoryList.new(Guardian.new(admin)).categories.map(&:id) - [SiteSetting.uncategorized_category_id] }
+    def category_list(user)
+      CategoryList.new(Guardian.new(user)).categories.map(&:id) - [SiteSetting.uncategorized_category_id]
```
SamSaffron: I worry that using category list here is too expensive, it does a lot of work including pruning empty dealing with unpinned per category and so on. We end up throwing all this work away.

PR #7284, opened 2019-03-28
FIX: Detect SNS notifications for SES correctly

SystemZ:
This fixes detection of email bounce by:
- removing hard requirement for email msg ID, msg ID in webhook msg never equals to ID in email_log
- gets bounce_score from user stats instead of nonexistent field in webhook msg

It's a bit better patch than shown in this post:
https://meta.discourse.org/t/handling-bouncing-e-mails/45343/100?u=systemz

### DIFF
--- a/app/jobs/regular/process_sns_notification.rb
+++ b/app/jobs/regular/process_sns_notification.rb
@@ -22,7 +22,7 @@ module Jobs
       return unless Aws::SNS::MessageVerifier.new.authentic?(raw)
 
       message.dig("bounce", "bouncedRecipients").each do |r|
-        if email_log = EmailLog.order("created_at DESC").where(to_address: r["emailAddress"])[0]
+        if email_log = EmailLog.order("created_at DESC").where(to_address: r["emailAddress"]).first
           email_log.update_columns(bounced: true)
 
           if email_log.user&.email.present?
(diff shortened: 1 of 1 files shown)


app/jobs/regular/process_sns_notification.rb
```diff
@@ -22,11 +22,11 @@ def execute(args)
       return unless Aws::SNS::MessageVerifier.new.authentic?(raw)
 
       message.dig("bounce", "bouncedRecipients").each do |r|
-        if email_log = EmailLog.find_by(message_id: message_id, to_address: r["emailAddress"])
+        if email_log = EmailLog.order("created_at DESC").where(to_address: r["emailAddress"])[0]
```
SamSaffron:
```suggestion
        if email_log = EmailLog.order("created_at DESC").where(to_address: r["emailAddress"]).first
```

SamSaffron:
@ZogStriP are we good to merge this?

app/jobs/regular/process_sns_notification.rb
SystemZ: Thanks, it looks better :)

SamSaffron:
OK I am merging this, let me know how you go!

PR #7289, opened 2019-03-29
FEATURE: Ignoring users from the preferences page in a cleaner way

khalilovcmded:
## UI

Revised UI

Comparing it to the PM participants:

Naka-Karuizawa_trip__-_Discourse


tgxworld:
I'm wondering if a list will be better here for the ignored users. Instead of trying to fit everything in an input field, we can list the users alongside the duration that the user will be ignored for.

coding-horror:
We don’t need to list duration though, so I think a simpler approach is
better.

tgxworld:
Ah ok :) I thought the duration would be helpful.

app/assets/javascripts/discourse/components/ignored-user-list-item.js.es6
```diff
@@ -0,0 +1,14 @@
+/* You might be looking for navigation-item. */
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  classNames: ['item'],
+  tagName: "div",
+  router: Ember.inject.service(),
```
ZogStriP: Is that needed?

app/assets/javascripts/discourse/components/ignored-user-list-item.js.es6
```diff
@@ -0,0 +1,14 @@
+/* You might be looking for navigation-item. */
```
ZogStriP: What is that comment for?

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
```diff
@@ -0,0 +1,33 @@
...
+      this.get("items").removeObject(item);
+      User.findByUsername(item).then(user => {
+        user
+          .updateNotificationLevel("normal")
+          .catch(popupAjaxError)
+          .finally(() => this.set("saved", true));
+      });
+    },
+    newItem() {
+      const controller = showModal("ignore-duration-with-username", {
```
ZogStriP: `controller` is too generic and should be rarely used.

```suggestion
      const modal = showModal("ignore-duration-with-username", {
```

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
```diff
@@ -0,0 +1,33 @@
...
+        model: this.get("model")
+      });
+      controller.setProperties({
+        onSuccess: (username) => {
+          this.get("items").addObject(username);
+        },
+        onClose: () => {
```
ZogStriP: Is that needed?

app/assets/javascripts/discourse/templates/components/ignored-user-list.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="ac-wrap clearfix" style="width: 480px;">
```
ZogStriP: No inline styling ;)

app/assets/javascripts/discourse/templates/modal/ignore-duration-with-username.hbs
```diff
@@ -0,0 +1,26 @@
+{{#d-modal-body title="user.user_notifications.ignore_duration_title" autoFocus="false"}}
+  <div class="controls tracking-controls">
+    <label>{{d-icon "eye-slash" class="icon"}} Username</label>
```
ZogStriP: `Username` should be localized

khalilovcmded:
@awesomerobot Can you please take a look at this proposed pattern and let me know what you think?

My concern here is that I am introducing a new UX pattern that might not be very consistent or transferable to other parts of Discourse. I wonder if there are any tweaks we could do to improve this. Also, an acceptable tweak is to leave the current version in production as it is 😉 

@jjaffeux made a good point about having a quick chat with you.

coding-horror:
if we can re-use existing code (e.g. the PM participants) I am for it, otherwise simple is better.. no need to gold plate this feature, we just need it to exist, not be perfect

khalilovcmded:
@awesomerobot can you please take a look at [this](https://github.com/discourse/discourse/pull/7289#issue-265787516)? 

It is not much code and maps to the behaviour we have in the PM participants.

khalilovcmded:
@awesomerobot that's a good point Kris 👍 I have changed it:

Profile_-_tarek_-_Discourse

khalilovcmded:
I have just added one more detail, when there are no ignored users, it looks strange as there is a void between the `ignored` title and the space below it.

It used to look like this:

Profile_-_tarek_-_Discourse

I changed it to look like this:

Profile_-_tarek_-_Discourse

ZogStriP:
I wonder if we should instead move the gray description next to the title

image

khalilovcmded:
That looks or will look very strange, I don't think it is good.

It also breaks the pattern that we established: `Title -> input -> instruction` like:
Profile_-_tarek_-_Discourse

khalilovcmded:
@ZogStriP @awesomerobot shall we try and merge this?

app/assets/javascripts/discourse/components/ignored-user-list-item.js.es6
```diff
@@ -0,0 +1,9 @@
+export default Ember.Component.extend({
+  tagName: "div",
+  items: null,
+  actions: {
+    removeItem(item) {
+      this.get("onRemoveItem")(item);
```
ZogStriP: Why isn't this directly sending the proper action?

```suggestion
      this.send("removeItem", item);
```

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
```diff
@@ -0,0 +1,29 @@
...
+          .catch(popupAjaxError)
+          .finally(() => this.set("saved", true));
+      });
+    },
+    newItem() {
+      const modal = showModal("ignore-duration-with-username", {
+        model: this.get("model")
+      });
+      modal.setProperties({
+        onSuccess: username => {
```
ZogStriP: Why isn't this an action defined in the `ignore-duration-with-username` controller?

app/assets/javascripts/discourse/controllers/ignore-duration-with-username.js.es6
```diff
@@ -0,0 +1,31 @@
...
+          "alert-error"
+        );
+        return;
+      }
+      this.set("loading", true);
+      User.findByUsername(this.get("ignoredUsername")).then(user => {
+        user
+          .updateNotificationLevel("ignore", this.get("ignoredUntil"))
+          .then(() => {
+            this.get("onSuccess")(this.get("ignoredUsername"));
```
ZogStriP: Why isn't this sending an action instead?

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
khalilovcmded: The `ignore-duration-with-username` controller / modal doesn't have access to `items`. The idea here is that the controller doesn't need to be aware of `items` since its responsibility is to `take care of the ignore user action` and on its success `we` (the `ignored-user-list`) are responsible of what to do next.

app/assets/javascripts/discourse/controllers/ignore-duration-with-username.js.es6
khalilovcmded: Communication from a controller to its route can be done using `send`, but here we are communicating from a controller to its calling component.

app/assets/javascripts/discourse/components/ignored-user-list-item.js.es6
```diff
@@ -0,0 +1,9 @@
+export default Ember.Component.extend({
+  tagName: "div",
+  items: null,
+  actions: {
+    removeItem(item) {
+      this.send("removeItem", item);
```
khalilovcmded: Just realised, we can't actually do that, we are communicating back to the parent component. We should keep it as it originally was.

app/assets/javascripts/discourse/controllers/ignore-duration-with-username.js.es6
khalilovcmded: We would / could have used `sendAction` but that's deprecated.

app/assets/javascripts/discourse/templates/components/ignored-user-list-item.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="ignored-user-list-item">
+  <span class="ignored-user-name">{{item}}</span>
+  <a class="remove-ignored-user no-text btn-icon btn" {{action "removeItem" item}}>{{d-icon "times"}}</a>
```
jjaffeux:
```suggestion
  {{d-button class="remove-ignored-user no-text btn-icon" action=(action "removeItem" item) icon="times"}}
```

app/assets/javascripts/discourse/templates/components/ignored-user-list.hbs
```diff
@@ -0,0 +1,11 @@
...
+  {{#if items}}
+    {{#each items as |item|}}
+      {{ignored-user-list-item item=item onRemoveItem=(action "removeItem")}}
+    {{/each}}
+  {{else}}
+    {{i18n 'user.user_notifications.ignore_no_users'}}
+  {{/if}}
+</div>
+<div class="instructions">{{i18n 'user.ignored_users_instructions'}}</div>
+<div>{{#d-button action=(action "newItem") class="btn-default" icon="plus"}}Add ...{{/d-button}}</div>
```
jjaffeux: Should be localised

app/assets/stylesheets/common/components/ignored-user-list.scss
```diff
@@ -0,0 +1,37 @@
...
+  box-sizing: border-box;
+}
+
+.ignored-user-list-item {
+  border: 1px solid #e9e9e9;
+  border-radius: 0.25em;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  height: 26px;
```
jjaffeux: can we avoid this kind of arbitrary height ?

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
```diff
@@ -0,0 +1,29 @@
+import { popupAjaxError } from "discourse/lib/ajax-error";
+import showModal from "discourse/lib/show-modal";
+import User from "discourse/models/user";
+
+export default Ember.Component.extend({
+  item: null,
+  actions: {
+    removeItem(item) {
```
jjaffeux: removeItem and newItem seem too generic to me

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
jjaffeux: onSuccess seems to generic to me, what about something like: `onUserIgnored` ?

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
```diff
@@ -0,0 +1,29 @@
...
+      this.get("items").removeObject(item);
+      User.findByUsername(item).then(user => {
+        user
+          .updateNotificationLevel("normal")
+          .catch(popupAjaxError)
+          .finally(() => this.set("saved", true));
+      });
+    },
+    newItem() {
+      const modal = showModal("ignore-duration-with-username", {
```
jjaffeux: depending on your needs you could rewrite it as:

```
const modal = showModal("ignore-duration-with-username", {
  model: {
    betterModelName: this.get("model"),
    onUserIgnored: username => {
      this.get("items").addObject(username);
    }
  }
});
```

app/assets/javascripts/discourse/controllers/ignore-duration-with-username.js.es6
jjaffeux: if you are sure onSuccess is always defined here, you can do: `this.onSucess(this.get("ignoredUsername")`

The get here won’t protect you from undefined object as you call a function on it:

```
this.get("unknown")()
// Uncaught TypeError: this.get(...) is not a function
```

app/assets/javascripts/discourse/templates/preferences/users.hbs
```diff
@@ -1,20 +1,16 @@
 <label class="control-label">{{i18n 'user.users'}}</label>
 <div class="control-group">
   <div class="controls tracking-controls">
-    <label>{{d-icon "d-muted" class="``icon"}} {{i18n 'user.muted_users'}}</label>
+    <label>{{d-icon "d-muted" class="icon"}} {{i18n 'user.muted_users'}}</label>
```
jjaffeux:
```suggestion
    <label>
      {{d-icon "d-muted" class="icon"}}
      <span>{{i18n 'user.muted_users'}}</span>
    </label>
```

app/assets/javascripts/discourse/templates/components/ignored-user-list.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="ignored-list">
+  {{#if items}}
+    {{#each items as |item|}}
+      {{ignored-user-list-item item=item onRemoveItem=(action "removeItem")}}
```
jjaffeux: onRemoveItem feels too generic

app/assets/javascripts/discourse/templates/components/ignored-user-list.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="ignored-list">
+  {{#if items}}
```
jjaffeux: each already has this behavior:

```
{{#each unknownItems as |item|}}
{{else}}
    will print this message
{{/each}}
```

jjaffeux:
I didn't put it each time, but generally speaking for a feature with a clear usage and wording, if we can avoid "item" I think it's better.

khalilovcmded:
> I didn't put it each time, but generally speaking for a feature with a clear usage and wording, if we can avoid "item" I think it's better.

I find it hard to agree. I personally adopted this convention through looking & working on other Ember projects that seem to agree on this pattern. For example, **Ghost's** Ember project, you will see that mostly children components to some an array models are usually postfixed with `item`:

File_Finder

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
khalilovcmded: Much better naming!

app/assets/javascripts/discourse/components/ignored-user-list.js.es6
khalilovcmded: I am not sure you can do that with the `showModal`

khalilovcmded:
One further iteration, I feel this is better compartmentalisation, the `save changes` button is actually only used for muted users now:

Profile_-_tarek_-_Discourse

app/assets/javascripts/discourse/controllers/ignore-duration-with-username.js.es6
khalilovcmded: I think it is less to do with `protection` from calling an undefined and more to `standari`ing` the way I access the component's properties, though I think it is okay to change it here 👍

jjaffeux:
To be clear I was not talking of files naming which I agree is fine. It's more in the actions and variables naming.

jjaffeux:
@venarius both tests failures seems legit and related to the commit

jjaffeux:
@ZogStriP looks good to me 👍 

👏 @khalilovcmded

PR #7292, opened 2019-03-31
UX: add awaiting-approval class to login modal

majakomel:
when local logins are disabled and account is awaiting approval


app/assets/javascripts/discourse/templates/mobile/modal/login.hbs
```diff
@@ -1,5 +1,5 @@
 {{#login-modal screenX=lastX screenY=lastY loginName=loginName loginPassword=loginPassword secondFactorToken=secondFactorToken action=(action "login")}}
-  {{#d-modal-body title="login.title" class="login-modal"}}
+  {{#d-modal-body title="login.title" class=(concat "login-modal" " " awaitingApprovalClass)}}
```
jjaffeux: cant we do a computed property handling all that logic once?

jjaffeux:
LGTM 👍

PR #7294, opened 2019-04-01
FIX: Relevance search will now consider document length in ranking.

tgxworld:
The default ranking options ranks by the number of matches which is
highly problematic when posts are stuffed with a keyword. The ranking
will now be divided by the document length which is a much fairer way to
rank.


tgxworld:
Merged in https://github.com/discourse/discourse/commit/e87ca594014733d171df0112149f49180d898678

PR #7301, opened 2019-04-01
PERF: Do not create staged users for most rejected incoming emails

davidtaylorhq:
Previously we would create users, then destroy them at the end if the post was rejected. Now we do not create users unless needed.


lib/email/receiver.rb
```diff
@@ -93,6 +94,12 @@ def ensure_valid_address_lists
       end
     end
 
+    def ensure_valid_date
+      if @mail.date.nil?
+        raise InvalidPost, "No post creation date found. Is the e-mail missing a Date: header?"
```
ZogStriP: This probably should be i18n'ed

lib/email/receiver.rb
```diff
@@ -144,12 +151,15 @@ def process_internal
 
       # Lets create a staged user if there isn't one yet. We will try to
```
ZogStriP: I think those comments are not relevant anymore?

lib/email/receiver.rb
```diff
@@ -144,12 +151,15 @@ def process_internal
...
       # delete staged users in process!() if something bad happens.
-      if user.nil?
-        user = find_or_create_user!(@from_email, @from_display_name)
-        log_and_validate_user(user)
-      end
 
       if post = find_related_post
+        # Most of the time, it is impossible to **reply** without a reply key, so exit early
+        if !user.present? && (sent_to_mailinglist_mirror? || !SiteSetting.find_related_post_with_key)
```
ZogStriP: Why not do this?

``` ruby
if user.blank?
  if sent_to_mailinglist_mirror? || !SiteSetting.find_related_post_with_key
    user = stage_from_user
  else
    raise BadDestinationAddress
  end
end
```

lib/email/receiver.rb
```diff
@@ -171,6 +181,9 @@ def process_internal
 
         raise first_exception if first_exception
 
+        # We don't stage new users for emails to reply addresses, exit if user is nil
+        raise BadDestinationAddress unless user.present?
```
ZogStriP:
```suggestion
        raise BadDestinationAddress if user.blank?
```

lib/email/receiver.rb
```diff
@@ -645,6 +663,9 @@ def process_destination(destination, user, body, elided, hidden_reason_id)
                      skip_validations: user.staged?)
 
       when :reply
+        # We don't stage new users for emails to reply addresses, exit if user is nil
+        raise BadDestinationAddress unless user.present?
```
ZogStriP:
```suggestion
        raise BadDestinationAddress if user.blank?
```

lib/email/receiver.rb
```diff
@@ -1136,9 +1153,17 @@ def send_subscription_mail(action, user)
       Email::Sender.new(message, :subscription).send
     end
 
+    def stage_from_user
+      if @from_user.nil?
```
ZogStriP: You could do something like

``` ruby
def stage_from_user
  @from_user ||= begin
    user = find_or_create_user!(@from_email, @from_display_name)
    log_and_validate_user(user)
    user
  end
end
```

lib/email/receiver.rb
```diff
@@ -142,14 +149,16 @@ def process_internal
...
-        user = find_or_create_user!(@from_email, @from_display_name)
-        log_and_validate_user(user)
-      end
-
       if post = find_related_post
+        # Most of the time, it is impossible to **reply** without a reply key, so exit early
+        if user.blank?
+          if sent_to_mailinglist_mirror? || !SiteSetting.find_related_post_with_key
+            user = stage_from_user
+          else !user.present?
```
ZogStriP: Did you mean?

```suggestion
          elsif user.blank?
```

PR #7302, opened 2019-04-02
PERF: Delete search data of posts from trashed topics periodically.

tgxworld:
This keeps both the index and table smaller.

### DIFF
--- a/app/jobs/scheduled/reindex_search.rb
+++ b/app/jobs/scheduled/reindex_search.rb
@@ -66,6 +66,18 @@ module Jobs
         .joins("LEFT JOIN posts p ON p.id = post_search_data.post_id")
         .where("p.raw = ''")
         .delete_all
+
+      DB.exec(<<~SQL, deleted_at: 1.week.ago)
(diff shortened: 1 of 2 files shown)


app/jobs/scheduled/reindex_search.rb
```diff
@@ -66,6 +66,18 @@ def clean_post_search_data
...
         .delete_all
+
+      DB.exec(<<~SQL)
+        DELETE FROM post_search_data
+        WHERE post_id IN (
+          SELECT post_id
+          FROM post_search_data
+          LEFT JOIN posts ON post_search_data.post_id = posts.id
+          LEFT JOIN topics ON posts.topic_id = topics.id
+          WHERE topics.id IS NOT NULL
```
eviltrout: Why not use an `INNER JOIN` if you require a topic?

tgxworld: Ah good catch :+1: I've been very used to just doing a LEFT JOIN and checking for `NOT NULL`.

PR #7309, opened 2019-04-02
FIX: clean up user export csv upload records in scheduled job

techAPJ:
https://meta.discourse.org/t/user-data-exports-do-not-expire-after-48-hours/113048/4?u=techapj

### DIFF
--- a/app/models/user_export.rb
+++ b/app/models/user_export.rb
@@ -1,12 +1,9 @@
 class UserExport < ActiveRecord::Base
   belongs_to :user
+  belongs_to :upload, dependent: :destroy
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
-      file_name = "#{user_export.file_name}-#{user_export.id}.csv.gz"
-      file_path = "#{UserExport.base_directory}/#{file_name}"
-
-      File.delete(file_path) if File.exist?(file_path)
       user_export.destroy!
     end
   end
--- a/spec/models/user_export_spec.rb
+++ b/spec/models/user_export_spec.rb
@@ -5,15 +5,19 @@ RSpec.describe UserExport do
 
   describe '.remove_old_exports' do
     it 'should remove the right records' do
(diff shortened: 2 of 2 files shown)


app/models/user_export.rb
```diff
@@ -3,10 +3,8 @@ class UserExport < ActiveRecord::Base
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
-      file_name = "#{user_export.file_name}-#{user_export.id}.csv.gz"
-      file_path = "#{UserExport.base_directory}/#{file_name}"
-
-      File.delete(file_path) if File.exist?(file_path)
+      upload = Upload.find_by(id: user_export.upload_id)
+      upload.destroy! if upload.present?
       user_export.destroy!
```
tgxworld: Perhaps we can set a `belongs_to :upload, dependent: :destroy`  on `UserExport`?

app/models/user_export.rb
```diff
@@ -3,10 +3,8 @@ class UserExport < ActiveRecord::Base
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
-      file_name = "#{user_export.file_name}-#{user_export.id}.csv.gz"
-      file_path = "#{UserExport.base_directory}/#{file_name}"
```
tgxworld: I think we can nuke `UserExport.base_directory` from our code base as well. We do need to think about how to clean up that folder which is left around.

app/models/user_export.rb
techAPJ: Good idea! 👍

app/models/user_export.rb
techAPJ: We still need `UserExport.base_directory` for creating temporary files. Relevant code [here](UserExport.base_directory).

tgxworld: Hmm I feel like tmp files should go into a tmp directory so that we don't affect the uploads directory.

techAPJ: Okay, I'll get that done in separate commit.

PR #7311, opened 2019-04-02
FIX: prevent anonymous users from changing their email/username/name

majakomel:
https://meta.discourse.org/t/suspect-account-anonymous-account/112813

### DIFF
--- a/spec/components/guardian_spec.rb
+++ b/spec/components/guardian_spec.rb
@@ -9,6 +9,7 @@ describe Guardian do
   let(:user) { Fabricate(:user) }
   let(:moderator) { Fabricate(:moderator) }
   let(:admin) { Fabricate(:admin) }
+  let(:anonymous_user) { Fabricate(:anonymous) }
   let(:trust_level_1) { build(:user, trust_level: 1) }
   let(:trust_level_2) { build(:user, trust_level: 2) }
   let(:trust_level_3) { build(:user, trust_level: 3) }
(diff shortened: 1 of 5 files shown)


spec/components/guardian_spec.rb
```diff
@@ -2403,6 +2404,20 @@
       it "is true for admins" do
         expect(Guardian.new(admin).can_edit_username?(user)).to be_truthy
       end
+
+      it "is true for admins when changing anonymous username" do
+        expect(Guardian.new(admin).can_edit_username?(anonymous_user)).to be_truthy
```
tgxworld: Should admin be able to edit usernames for anon users as well?

test/javascripts/controllers/preferences-account-test.js.es6
```diff
@@ -21,6 +22,10 @@ QUnit.skip("updating of associated accounts", function(assert) {
 
   assert.equal(controller.get("canUpdateAssociatedAccounts"), false);
 
+  controller.set("model.is_anonymous", false);
+
+  assert.equal(controller.get("canUpdateAssociatedAccounts"), false);
```
tgxworld: If the mode is not anonymous, should we be able to update the associated accounts?

majakomel: Non-anonymous users are able to update associated accounts if other conditions are met https://github.com/discourse/discourse/pull/7311/files#diff-4aba2c210bc3e273a9f57c8348785fd4R109

majakomel:
@SamSaffron, should admin still be able to edit email/username/name of anonymous users after the change?

I removed password and associated account from preferences and was thinking that it would make sense to disable editing of about/location/website/avatar as well to prevent anonymous account from being too personalized?

test/javascripts/controllers/preferences-account-test.js.es6
tgxworld: ah icic. There are other conditions 👍

ZogStriP:
I agree we should prevent any kind of customization for Anonymous users but still allow Admins to change anything they want.

@SamSaffron what do you think?

SamSaffron:
I think admins are fine to make changes here if they wish. I am merging this.

PR #7316, opened 2019-04-03
FEATURE: Hide quoted replies for ignored users

khalilovcmded:
## Why?

## UI

## Current approach

1. **Hiding a quote** is happening by returning a list of `ignored usernames` (for the current User) from the backend, and on the front-end, we are using the `ignored usernames` value to do a `find & replace` in the already cooked post HTML for the Quoted text if it matches any of the `ignored usernames` array.

Cons:

- It doesn't feel natural to hack around HTML on the front-end this way, it is unreliable as well, since we are `$($aside).find('.title').text().trim() ....` to extract the Quoted text's username, and then comparing it with **ignored usernames**.

2. **Avoiding the N+1** is important, we are returning `ignored usernames` in the `TopicViewSerializer` for each Topic load, scoped to the current User. I have wrapped this around a read-through cache to ensure it is only called once within an 1 hour of expiry time.

Cons:

- The read-through cache itself isn't terrible, but it can be fragile because we need to make sure cache invalidation works, in this case it is _manual_. Other considerations, Redis round-trip, and how efficient is the expiry period.

## Different approach

1. **Hiding a quote** method can be improved by changing the way we are cooking the post and be more explicit by adding a `data-user-id` attribute, which makes it a lot easier and more reliable for us to hide the Quoted text on the front-end. 

Cons:

- It can be time consuming since we are probably going to have to write some migration to re-bake existing quoted posts?

2. **The ignored usernames N+1** can be avoided by tweaking the way we load the current User's connected data. The relationship is: `User -> IgnoredUsers`, that means, on the front-end, we can potentially be side-loading the `IgnoredUsers` models one time only and let Ember keep the current User state stored temporarily.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
@@ -192,7 +192,7 @@ export default class PostCooked {
       : this.attrs.topicUrl;
   }
 
-  _updateQuoteElements($html, $aside, desc) {
+  _updateQuoteElements($aside, desc) {
     let navLink = "";
     const quoteTitle = I18n.t("post.follow_quote");
     let postNumber = $aside.data("post");
@@ -227,7 +227,7 @@ export default class PostCooked {
     $quotes.each((i, e) => {
       const $aside = $(e);
       if ($aside.data("post")) {
-        this._updateQuoteElements($html, $aside, "chevron-down");
+        this._updateQuoteElements($aside, "chevron-down");
         const $title = $(".title", $aside);
 
         // Unless it's a full quote, allow click to expand
--- a/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-cooked.js.es6
@@ -192,7 +192,7 @@ export default class PostCooked {
       : this.attrs.topicUrl;
   }
 
-  _updateQuoteElements($aside, desc) {
+  _updateQuoteElements($html, $aside, desc) {
     let navLink = "";
     const quoteTitle = I18n.t("post.follow_quote");
     let postNumber = $aside.data("post");
@@ -212,6 +212,9 @@ export default class PostCooked {
       expandContract = iconHTML(desc, { title: "post.expand_collapse" });
       $(".title", $aside).css("cursor", "pointer");
     }
+    if(this.attrs.topic.ignored_usernames.includes($($aside).find('.title').text().trim().slice(0, -1))) {
+      $($aside).find("p").replaceWith("<i>Hidden content</i>");
+    }
     $(".quote-controls", $aside).html(expandContract + navLink);
   }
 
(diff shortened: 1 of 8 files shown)


app/serializers/topic_view_serializer.rb
```diff
@@ -213,6 +214,10 @@ def unpinned
     PinnedCheck.unpinned?(object.topic, object.topic_user)
   end
 
+  def ignored_usernames
+    IgnoredUser.where(user: topic.user).joins(:ignored_user).pluck(:username)
```
khalilovcmded: This is N+1 but I have checked against when is `TopicViewSerializer` called, and it seems to be that it called in the `TopicsController` primarily, which loads once.

I would love some thoughts if this is okay or we want to avoid it completely, and if we want to avoid what other possible ways to achieve this?

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -212,6 +212,9 @@ export default class PostCooked {
       expandContract = iconHTML(desc, { title: "post.expand_collapse" });
       $(".title", $aside).css("cursor", "pointer");
     }
+    if(this.attrs.topic.ignored_usernames.includes($($aside).find('.title').text().trim().slice(0, -1))) {
+      $($aside).find("p").replaceWith("<i>Hidden content</i>");
```
khalilovcmded: - This doesn't seem not to be the cleanest approach
- I can use the `I18n.t()` for the content

app/serializers/topic_view_serializer.rb
ZogStriP: We definitely want to avoid these time 💣 if possible. Usually, to avoid N+1 we preload records. Maybe look at how we preload custom fields?

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
ZogStriP: If your 🕷 sense tells you it's not the cleanest approach, you should definitely trust it 😉

ZogStriP: But since we only have the username to work with on the client-side, I can't think of a better solution that doesn't require changing the way we cook quotes (ie. by adding a `data-user-id` attribute`).

khalilovcmded: Yes, that's what I thought as well. The way the posts are cooked doesn't have a reliable identifier `data-*` to use, and therefore, I fall back to the `.title text` approach.

ZogStriP:
> - **The ignored usernames N+1** can be avoided by tweaking the way we load the current User's connected data. The relationship is: `User -> IgnoredUsers`, that means, on the front-end, we can potentially be side-loading the `IgnoredUsers` models one time only and let Ember keep the current User state stored temporarily.

That's exactly what we want to do. You also need to ensure it's kept up to date on client-side when it's updated on the server-side.

PR #7319, opened 2019-04-03
Show link to the original topic when it was deleted

romanrizzi:
otopic

### Changes
- i18n keys were moved into core since they're being used and not on the Akismet plugin.
- Added `removed_topic_id` to the serializer when the topic was deleted.

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
+++ b/app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
@@ -7,6 +7,9 @@
   {{else if (has-block)}}
     {{yield}}
   {{else}}
-    <span class="title-text">{{i18n "topic.deleted"}}</span>
+    <span class="title-text">
+      {{i18n "review.topics.deleted"}}
+      <a href={{get-url (concat '/t/' reviewable.removed_topic_id)}}>{{i18n "review.topics.original"}}</a>
+    </span>
   {{/if}}
 </div>
--- a/app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
+++ b/app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
@@ -9,7 +9,7 @@
   {{else}}
     <span class="title-text">
       {{i18n "review.topics.deleted"}}
-      <a href={{get-url (concat '/t/' reviewable.removed_topic_id)}}>{{i18n "review.topics.original"}}</a>
+      {{link-to (i18n "review.topics.original") "topic" "-" reviewable.removed_topic_id}}
     </span>
   {{/if}}
 </div>
--- a/app/serializers/reviewable_serializer.rb
+++ b/app/serializers/reviewable_serializer.rb
@@ -73,19 +73,19 @@ class ReviewableSerializer < ApplicationSerializer
   end
 
   def attributes
-    data = super
+    super.tap do |data|
+      data[:removed_topic_id] = object.topic_id unless object.topic
 
-    if object.target.present?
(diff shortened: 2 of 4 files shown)


ZogStriP:
Why isn't `[Topic deleted]` the link to the original topic?

romanrizzi:
Good question :thinking: 

It seems more clear to me but I'm open to change it.

app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
```diff
@@ -6,6 +6,9 @@
   {{else if (has-block)}}
     {{yield}}
   {{else}}
-    <span class="title-text">{{i18n "topic.deleted"}}</span>
+    <span class="title-text">
+      {{i18n "review.topics.deleted"}}
+      <a href='/t/{{reviewable.removed_topic_id}}'>{{i18n "review.topics.original"}}</a>
```
eviltrout: This is not safe for subfolder based installations. You could use the `{{get-url}}` helper.

romanrizzi: Done :ok_hand:

app/serializers/reviewable_serializer.rb
```diff
@@ -72,19 +72,19 @@ def self.payload_attributes(*attributes)
   end
 
   def attributes
-    data = super
+    super.tap do |data|
+      data[:removed_topic_id] = object.topic_id unless object.topic
```
eviltrout: We should have a test that confirms that the serializer includes this attribute.

spec/serializers/reviewable_serializer_spec.rb
```diff
@@ -15,12 +15,20 @@
     expect(json[:category_id]).to eq(reviewable.category_id)
     expect(json[:can_edit]).to eq(true)
     expect(json[:version]).to eq(0)
+    expect(json[:removed_topic_id]).to be_nil
+  end
+
+  it 'Includes the removed topic id when the topis was deleted' do
```
jjaffeux:
```suggestion
  it 'Includes the removed topic id when the topic was deleted' do
```

spec/serializers/reviewable_serializer_spec.rb
```diff
@@ -15,12 +15,20 @@
...
+    reviewable.topic.trash!(admin)
+
+    json = described_class.new(reviewable.reload, scope: Guardian.new(admin), root: nil).as_json
+
+    expect(json[:removed_topic_id]).to eq reviewable.topic_id
```
jjaffeux:
```suggestion
    expect(json[:removed_topic_id]).to eq(reviewable.topic_id)
```

app/assets/javascripts/discourse/templates/components/reviewable-topic-link.hbs
```diff
@@ -6,6 +6,9 @@
   {{else if (has-block)}}
     {{yield}}
   {{else}}
-    <span class="title-text">{{i18n "topic.deleted"}}</span>
+    <span class="title-text">
+      {{i18n "review.topics.deleted"}}
+      <a href={{get-url (concat '/t/' reviewable.removed_topic_id)}}>{{i18n "review.topics.original"}}</a>
```
jjaffeux: @eviltrout I was wondering is there any cons/drawbacks to do this instead:

```
  {{#link-to "topic" "-" reviewable.removed_topic_id}}
   {{i18n "review.topics.original"}}
  {{/link-to}}
```

SamSaffron:
@eviltrout / @jjaffeux  are we good to merge this?

jjaffeux:
Yes LGTM :+1:

PR #7320, opened 2019-04-04
FIX: Hide post replies content for ignored users

khalilovcmded:
## Why?

- [x] Collapse & hide posts of ignored users
- [x] Hide content of first post by ignored user
- [x] Hide content of quoted text by an ignored user
- [x] Hide content of replies by an ignored user
- [x] Hide any post sent in real-time by ignored user

### First post hidden

Is_Fujiten_Snow_Resort_a_good_place_to_be_in_Winter__-_Discourse

### Post replies hidden

Is_Fujiten_Snow_Resort_a_good_place_to_be_in_Winter__-_Discourse

### Quoted replies hidden

Is_Fujiten_Snow_Resort_a_good_place_to_be_in_Winter__-_Discourse

### Replying in real-time by an ignored user

GIF


app/serializers/basic_post_serializer.rb
```diff
@@ -45,7 +45,7 @@ def cooked
 
   def ignored
     return false unless SiteSetting.ignore_user_enabled?
-    object.is_first_post? &&
```
ZogStriP: This definitely needs a test 😉

khalilovcmded: Added ✅

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -241,6 +242,12 @@ export default class PostCooked {
       }
     });
   }
+
+  _applyIgnoredState($html) {
+    if (this.attrs.ignored) {
+      $($html).addClass("post-ignored");
```
ZogStriP: Do you need to call jQuery when `$html` is already a jQuery element?

```suggestion
      $html.addClass("post-ignored");
```

app/serializers/basic_post_serializer.rb
```diff
@@ -45,7 +45,7 @@ def cooked
...
+    (object.is_first_post? || object.reply_to_post_number.present?) &&
```
ZogStriP: Since this is now potentially called for more than 1 posts per topic, can we somehow prevent the N+1?

khalilovcmded: Good point. Let me try and see if we can just do it all front-end.

khalilovcmded: Found a way out by delegating hiding the content work to the front-end. ✅

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -241,6 +255,24 @@ export default class PostCooked {
...
+    const { currentUser } = this.decoratorHelper.widget;
+    if (currentUser) {
+      const ignoredUsers = currentUser.ignored_users;
+      if (
+        ignoredUsers &&
+        ignoredUsers.length > 0 &&
+        ignoredUsers.includes(this.attrs.username)
+      ) {
+        return $(
+          `<div class='cooked post-ignored'>${I18n.t("post.ignored")}</div>`
```
khalilovcmded: This might not be correct. I am fixing it.

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -241,6 +255,24 @@ export default class PostCooked {
       }
     });
   }
+
+  _computeCooked() {
```
khalilovcmded: The logic here isn't complete yet.

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -241,6 +255,24 @@ export default class PostCooked {
...
+    const { currentUser } = this.decoratorHelper.widget;
+    if (currentUser) {
+      const ignoredUsers = currentUser.ignored_users;
```
ZogStriP: Can we extract the list of ignored users only once?

app/assets/javascripts/discourse/controllers/ignore-duration.js.es6
```diff
@@ -17,6 +18,8 @@ export default Ember.Controller.extend(ModalFunctionality, {
       this.get("model")
         .updateNotificationLevel("ignore", this.get("ignoredUntil"))
         .then(() => {
+          const currentUser = PreloadStore.get("currentUser");
+          currentUser.ignored_users.push(this.get("model.username"));
```
khalilovcmded: I could probably avoid this logic entirely by exposing some method of refreshing the preloaded `currentUser`.
The technique we use the [findDetails](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/user.js.es6#L469-L517) here sounds similar to what I want to achieve since it also `refreshes` the `preloaded-store`.

app/assets/javascripts/select-kit/components/user-notifications-dropdown.js.es6
```diff
@@ -50,6 +51,8 @@ export default DropdownSelectBox.extend({
   changeToNormal() {
     this.get("updateNotificationLevel")("normal")
       .then(() => {
+        const currentUser = PreloadStore.get("currentUser");
+        currentUser.ignored_users.removeObject(this.get("user.username"));
```
khalilovcmded: Same concern as [above](https://github.com/discourse/discourse/pull/7320/files#r272278965) ☝️

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -17,6 +17,10 @@ export default class PostCooked {
     this.expanding = false;
     this._highlighted = false;
     this.decoratorHelper = decoratorHelper;
+    this.currentUser = decoratorHelper.widget.currentUser;
+    if (this.currentUser) {
```
ZogStriP: I don't think this test is needed since you're only setting ignoredUsers when there is a currentUser?

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -212,6 +216,18 @@ export default class PostCooked {
       expandContract = iconHTML(desc, { title: "post.expand_collapse" });
       $(".title", $aside).css("cursor", "pointer");
     }
+    if (this.currentUser) {
```
ZogStriP: Same here, I don't think that test is needed?

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -619,6 +619,15 @@ const User = RestModel.extend({
     return ajax(`${userPath(this.get("username"))}/notification_level.json`, {
       type: "PUT",
       data: { notification_level: level, expiring_at: expiringAt }
+    }).then(() => {
+      const currentUser = Discourse.User.current();
```
ZogStriP: Hmm, isn't this.currentUser available? We try to avoid using globals as possible

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -241,6 +257,22 @@ export default class PostCooked {
       }
     });
   }
+
+  _computeCooked() {
+    if (this.currentUser) {
```
ZogStriP: Same here, that test might not be needed?

app/serializers/basic_post_serializer.rb
```diff
@@ -36,21 +35,11 @@ def cooked
...
         I18n.t('flagging.user_must_edit')
       end
-    elsif ignored
-      I18n.t('ignored.hidden_content')
     else
       object.filter_quotes(@parent_post)
     end
   end
 
-  def ignored
```
ZogStriP: Was there any tests for this that needs to be removed?

khalilovcmded: No, I wrote tests in another PR, but that is already closed.

app/assets/javascripts/discourse/models/user.js.es6
khalilovcmded: 1. There is no `currentUser` property (we can access the `this` of the object which is the current model's context)
2. The `this` in the context of the User model doesn't actually refer to the correct version of the `currentUser`, the correct version of the currentUser is stored in the preloaded store and fetched via `Discourse.User.current()` ... in this same `user.js.es6` model file, you will find other parts using the `Discourse.User.current()` because that's the real version
3. You will also notice within the `user.js.es6` file, sometimes we do double updates to the `Discourse.User.current()` and the this (the model user's context)

So I think this is fine.

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -17,6 +17,8 @@ export default class PostCooked {
     this.expanding = false;
     this._highlighted = false;
     this.decoratorHelper = decoratorHelper;
+    this.currentUser = decoratorHelper.widget.currentUser;
+    this.ignoredUsers = this.currentUser.ignored_users;
```
ZogStriP: Is `decoratorHelper.widget.currentUser` **always** available?

PR #7322, opened 2019-04-04
FIX: skip some checks for CJK locale in TextSentinel

fantasticfears:
continuation of #7045


SamSaffron:
change looks good to me but rubocop wants a few lines cleaned up.

PR #7323, opened 2019-04-04
FIX: don't trigger notifications when changing category/tags of unlisted topics

majakomel:
https://meta.discourse.org/t/any-way-to-not-send-emails-when-a-topic-category-is-changed/112845/13

### DIFF
--- a/app/jobs/regular/notify_category_change.rb
+++ b/app/jobs/regular/notify_category_change.rb
@@ -5,7 +5,7 @@ module Jobs
     def execute(args)
       post = Post.find_by(id: args[:post_id])
 
-      if post&.topic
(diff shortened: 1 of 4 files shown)


app/jobs/regular/notify_category_change.rb
```diff
@@ -5,7 +5,7 @@ class NotifyCategoryChange < Jobs::Base
     def execute(args)
       post = Post.find_by(id: args[:post_id])
 
-      if post&.topic
+      if post&.topic.visible?
```
SamSaffron: I think we need ` if post&.topic&.visible?` here to remain safe

PR #7325, opened 2019-04-05
add header cloak after all panels are added

angusmcleod:
See further: https://meta.discourse.org/t/addheaderpanel-api-method-is-effectively-broken/113499


featheredtoast:
Thanks @angusmcleod this is a good catch!

PR #7327, opened 2019-04-05
FEATURE: implements generics filters for reports

jjaffeux:
category,group and file extension now use it.

### DIFF
--- a/spec/models/report_spec.rb
+++ b/spec/models/report_spec.rb
@@ -20,7 +20,6 @@ describe Report do
   end
 
   shared_examples 'category filtering on subcategories' do
-
     it 'returns the filtered data' do
       expect(report.total).to eq(1)
     end
@@ -753,12 +752,12 @@ describe Report do
       end
 
       context "with category filtering" do
-        let(:report) { Report.find('flags', category_id: c1.id) }
+        let(:report) { Report.find('flags', filters: { category: c1.id }) }
 
         include_examples 'category filtering'
 
         context "on subcategories" do
-          let(:report) { Report.find('flags', category_id: c0.id) }
+          let(:report) { Report.find('flags', filters: { category: c0.id }) }
 
           include_examples 'category filtering on subcategories'
         end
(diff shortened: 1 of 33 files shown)


jjaffeux:
This PR allows custom filters to be added very simply:

Screenshot 2019-04-05 at 18 29 28

To make it possible you have to add support for in your report:

```
report.add_filter("filter-name")

# make use of you filter-name later in the report
filter_name = report.filter.dig("filter-name")

data = User.where(something: filter_name)
```

And creates a `report-filters/filter-name` component

It will give you out of the box: query params support (in the filters query param), UI support and caching. 

Meaning you can now link to : `/admin/reports/signups?filters=filter-name:xxx`

eviltrout:
The PR is quite big, but one that raises my eyebrow is that the filters param seems to have a "micro format". These things can be somewhat of a trap. For example, what if the filter value need to contain a colon?

I can think of two potentially more robust solutions:

1. Use the Rails-style `filters[name]=value` format. This will pass automatically to rails as `"filters" => { "name" => "value" }` for easy parsing.

2. You could use `JSON.stringify` and put the parameters in a JSON object instead. That will making parsing/encoding it quite easy.

jjaffeux:
@eviltrout fully agreed, I tried to implement 1) before going with this solution, but discovered it was not working out of the box in Ember and would have to go for a lot of parsing. Given my solution could be changed later if it was not working for a specific case I decided to not solve an issue I didn’t have.

But I didn’t think of 2) and it seems both well working and easy to do, so I implemented it in my last commit.

spec/models/report_spec.rb
```diff
@@ -753,12 +753,12 @@
       end
 
       context "with category filtering" do
-        let(:report) { Report.find('flags', category_id: c1.id) }
+        let(:report) { Report.find('flags', filters: "category:#{c1.id}") }
```
eviltrout: These aren't JSON?

test/javascripts/acceptance/dashboard-test.js.es6
```diff
@@ -86,3 +98,17 @@ QUnit.test("reports tab", async assert => {
     "filter is case insensitive"
   );
 });
+
+QUnit.test("report filters", async assert => {
+  await visit(
+    "/admin/reports/signups?end_date=2018-07-16&filters=group:88&start_date=2018-06-16"
```
eviltrout: Ditto here, I thought it was moved to JSON?

spec/models/report_spec.rb
jjaffeux: Was working on it, yes it's changed

test/javascripts/acceptance/dashboard-test.js.es6
jjaffeux: Working on it, I discovered an issue with caching, will ping you when ready for review.

PR #7328, opened 2019-04-05
FEATURE: Add disable dashboard stats site setting

sau226:
Some Discourse instances may want to disable the weekly dashboard PM called "New advice on your site dashboard". This PR adds a site setting (default off/false) that will cause the PM to not be sent if enabled by an admin.

See this link for some more context: https://meta.discourse.org/t/disable-new-advice-on-your-site-dashboard-inbox-spam/110981

### DIFF
--- a/app/jobs/scheduled/dashboard_stats.rb
+++ b/app/jobs/scheduled/dashboard_stats.rb
@@ -8,7 +8,7 @@ module Jobs
 
     def execute(args)
       return if SiteSetting.disable_dashboard_stats == true
-      
+
       problems_started_at = AdminDashboardData.problems_started_at
       if problems_started_at && problems_started_at < 2.days.ago
         # If there have been problems reported on the dashboard for a while,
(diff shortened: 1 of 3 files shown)


SamSaffron:
Sorry, but I am a pretty strong reject on this. 

https://meta.discourse.org/t/disable-new-advice-on-your-site-dashboard-inbox-spam/110981/7?u=sam

cube-drone:
Boo! BOO! I'm tired of being hassled every week by your software! 
I'm not going to install a CDN for 8 people and I'm honestly frustrated that you want me to!

culturerich:
Our forum is third-party managed, with updates every few days, yet I still have an almost constant notice (usually there's a plugin update, rarely is it security critical). I think we need some way of distinguishing more or less important admin notices, otherwise we're into all-the-machines-in-the-hospital-bleeping territory, which research shows is NOT good for anyone: although I appreciate nobody is dying because of these notices!

jtheletter:
How do we opt out of these messages?

PR #7330, opened 2019-04-06
add image to user's crawler page

mrfinch:
use category-colour instead of static colour in post layout's breadcrumbs

Related to https://meta.discourse.org/t/improving-discourse-static-html-archive/112497

Let me know if you have any suggestions on what we can add to this page. Any other information we can show to the unauthenticated user? An image was already used in meta information for crawler so I have added in UI too.

Screenshot from 2019-04-06 10-54-40

### DIFF
--- a/app/assets/stylesheets/desktop/topic-post.scss
+++ b/app/assets/stylesheets/desktop/topic-post.scss
@@ -900,9 +900,6 @@ span.highlighted {
 }
 
 #breadcrumbs {
-  .badge-category-bg {
-    background-color: $secondary-high;
-  }
   .category-title {
     color: $primary;
   }
--- a/app/views/topics/show.html.erb
+++ b/app/views/topics/show.html.erb
@@ -11,7 +11,7 @@
         itemref="breadcrumb-<%=(i+1)%>"
       <%-end%>>
       <a href="<%=c[:url]%>" itemprop="url" class='badge-wrapper bullet'>
-        <span class="badge-category-bg"></span>
+        <span class="badge-category-bg" style='background-color: #<%= c[:color] %>'></span>
         <span itemprop="title" class='category-title'><%=c[:name]%></span>
       </a>
     </div>
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -1,4 +1,7 @@
-<h2><%= @user.username %></h2>
+<div class='user-crawler'>
+	<img src='<%= get_absolute_image_url(@user.small_avatar_url) %>' alt='Image of <%= @user.username %>' title='<%= @user.username %>' />
+	<h2 class='username'><%= @user.username %></h2>
+</div>
 
 <% unless @restrict_fields %>
 <p><%= raw @user.user_profile.bio_processed %></p>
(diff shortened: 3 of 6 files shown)


app/views/users/show.html.erb
```diff
@@ -1,4 +1,7 @@
-<h2><%= @user.username %></h2>
+<div class='user-crawler'>
+	<img src='<%= get_absolute_image_url(@user.small_avatar_url) %>' alt='Image of <%= @user.username %>' title='<%= @user.username %>' />
```
ZogStriP: The alt should be localized

mrfinch: localization and me :bowing_man:, thanks @ZogStriP, I have fixed it now

config/locales/server.en.yml
```diff
@@ -2225,6 +2225,7 @@ en:
       fixed_primary_email: "Fixed primary email for staged user"
       same_ip_address: "Same IP address (%{ip_address}) as other users"
       inactive_user: "Inactive user"
+    image_alt: "Image of %{username}"
```
SamSaffron: we don't need to carry this localization. Just have the alt be the username

app/views/users/show.html.erb
```diff
@@ -1,4 +1,7 @@
-<h2><%= @user.username %></h2>
+<div class='user-crawler'>
+	<img src='<%= get_absolute_image_url(@user.small_avatar_url) %>' alt='<%= t('user.image_alt', username: @user.username) %>' title='<%= @user.username %>' />
```
SamSaffron: we don't want an absolute image url here, we want a CDN url

mrfinch: do we have any helper function to get a CDN URL? @SamSaffron

mrfinch: used `UrlHelper.local_cdn_url`, let me know if there is an issue with it.

app/assets/stylesheets/desktop/topic-post.scss
```diff
@@ -900,9 +900,6 @@ span.highlighted {
 }
 
 #breadcrumbs {
-  .badge-category-bg {
-    background-color: $secondary-high;
-  }
```
SamSaffron: why is this being removed @awesomerobot  you ok with this?

app/views/topics/show.html.erb
```diff
@@ -11,7 +11,7 @@
         itemref="breadcrumb-<%=(i+1)%>"
       <%-end%>>
       <a href="<%=c[:url]%>" itemprop="url" class='badge-wrapper bullet'>
-        <span class="badge-category-bg"></span>
+        <span class="badge-category-bg" style='background-color: #<%= c[:color] %>'></span>
```
SamSaffron: how is this going to work? where is the foreground? 

My call is not to even style this cause its just too annoying.

SamSaffron:
I am super sorry but I am closing this, there are two changes being proposed here 

1. Adding image to user page in crawler view
2. Fiddle with styling of categories in crawler view 

We need 2 distinct PRs here.

PR #7333, opened 2019-04-08
REFACTOR: Pass down `currentUser` to `post-cooked` widget

khalilovcmded:
## Why?

Instead of evaluating the `currentUser` from the `DecoratorHelper`, we are passing it down from the `post` widget to the `post-cooked` widget.


eviltrout:
This is certainly cleaner! :+1:

PR #7335, opened 2019-04-08
UX: moves local-dates from popupMenu to toolbar

jjaffeux:


jjaffeux:
Screenshot 2019-04-08 at 14 59 01

PR #7336, opened 2019-04-08
 FIX: Remove quoted post text and block realtime replies from ignored users

khalilovcmded:
## Why?

We are doing two things here:

- [x] We want to completely get rid of any Quoted text from an ignored user (previously we put the text as : `Hidden content`)
- [x] We want to disable showing the `[ ] replying...` on the topic when an ignored user is replying to a topic

## UI

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post.js.es6
@@ -355,7 +355,7 @@ createWidget("post-contents", {
   },
 
   html(attrs, state) {
-    let result = [new PostCooked(attrs, new DecoratorHelper(this))];
+    let result = [new PostCooked(attrs, new DecoratorHelper(this), this.currentUser)];
     result = result.concat(applyDecorators(this, "after-cooked", attrs, state));
 
     if (attrs.cooked_hidden) {
--- a/app/assets/javascripts/discourse/widgets/post.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post.js.es6
@@ -355,7 +355,9 @@ createWidget("post-contents", {
   },
 
   html(attrs, state) {
-    let result = [new PostCooked(attrs, new DecoratorHelper(this), this.currentUser)];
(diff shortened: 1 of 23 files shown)


app/assets/javascripts/discourse/widgets/post.js.es6
```diff
@@ -657,14 +659,6 @@ export default createWidget("post", {
     } else {
       classNames.push("regular");
     }
-    if (
```
khalilovcmded: We don't need this anymore since we are doing all the work in `cooked-post`.

plugins/discourse-presence/assets/javascripts/discourse/components/topic-presence-display.js.es6
```diff
@@ -48,6 +48,15 @@ export default Ember.Component.extend({
 
   @computed("presenceUsers", "currentUser.id")
   users(users, currentUserId) {
+    const currentUser = this.get("currentUser");
```
jjaffeux: Feel like this could be done like this:

```
@computed("presenceUsers", "currentUser.{id,ignored_users}")
users(users, currentUser) {
    const ignoredUsers = currentUser.ignored_users;
    if (
      ignoredUsers &&
      ignoredUsers.length > 0 &&
      ignoredUsers.includes(currentUser.username)
    ) {
      return [];
    }

    return (users || []).filter(user => user.id !== currentUser.id);
```

jjaffeux:
```
let ignored_users = null;
console.log(ignored_users || []); // []

ignored_users = undefined;
console.log(ignored_users || []); // []

ignored_users = ["bob"];
console.log(ignored_users || []); // ["bob"]
```

khalilovcmded: Actually, my logic isn't correct, I will fix it :)

PR #7339, opened 2019-04-08
DEV: Expose a way to extend a method that returns a list

romanrizzi:
This is the third time I find myself needing to use this pattern for a plugin (I'm going to need it again for `discourse-sift`) and thought it was a good idea to expose a method in the plugin instance API.

It's going to be used for:
 - Registering a `ReviewableAkismetPost` type
 - Registering a `ReviewableSiftPost` type
 - Extend `UserHistory#user_history_actions` with `discourse-akismet` custom actions.
 - Extend `UserHistory#user_history_actions` with `discourse-sift` custom actions.

### DIFF
--- a/lib/plugin/instance.rb
+++ b/lib/plugin/instance.rb
@@ -605,11 +605,15 @@ class Plugin::Instance
   end
 
   def register_reviewable_type(reviewable_type_class)
-    types = Reviewable.types
-    types << reviewable_type_class.name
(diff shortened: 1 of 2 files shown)


romanrizzi:
I think it's better to keep it generic. There's no real benefit on defining facade methods (I leave the reviewable one for compatibility, we can remove it later)

PR #7340, opened 2019-04-08
FEATURE: Refresh MaxmindDb during assets:precompile.

nbianca:


lib/discourse_ip_info.rb
```diff
@@ -16,6 +18,33 @@ def open_db(path)
     @cache = LruRedux::ThreadSafeCache.new(2000)
   end
 
+  def self.mmdb_path(name)
+    File.join(Rails.root, 'vendor', 'data', "#{name}.mmdb")
+  end
+
+  def self.mmdb_download(name)
+    uri = URI("http://geolite.maxmind.com/download/geoip/database/#{name}.tar.gz")
```
ZogStriP:
```suggestion
    uri = URI("https://geolite.maxmind.com/download/geoip/database/#{name}.tar.gz")
```

config/site_settings.yml
```diff
@@ -1481,6 +1481,9 @@ developer:
   enable_safe_mode:
     default: true
     client: true
+  refresh_maxmind_db_during_precompile_days:
+    default: 2
```
ZogStriP: We should probably add a range for that setting. Maybe 1-30 ?

SamSaffron: Maybe 0-120, if set to 0 it does not update.

lib/discourse_ip_info.rb
```diff
@@ -2,6 +2,8 @@
 
 require 'maxminddb'
 require 'resolv'
+require 'rubygems/package'
```
SamSaffron: I think these requires are too high, just require them in the methods that use them cause the vast majority of time we don't need to pull in either.

SamSaffron:
looks great to me, once you address the minor comments feel free to merge !

lib/tasks/assets.rake
```diff
@@ -3,7 +3,7 @@ task 'assets:precompile:before' do
   require 'uglifier'
   require 'open3'
 
-  unless %w{profile production}.include? Rails.env
+  unless %w{profile production development}.include? Rails.env
```
ZogStriP: Why that change?

PR #7345, opened 2019-04-09
FIX: Keep highlighted text for quoted replies by ignored users

khalilovcmded:
## Why?

We hide/remove any ignored user quoted text in any of the posts for a certain topic. Users can **expand** the ignored user quoted text, and as a result, the **highlighting** logic wasn't working. This PR is to fix this behaviour.

#### Before
GIF 1

#### After

GIF 2 with hightlights

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -2243,7 +2243,7 @@ en:
       quote_reply: "Quote"
       edit_reason: "Reason: "
       post_number: "post {{number}}"
-      ignored: "Hidden content"
+      ignored: "Ignored content"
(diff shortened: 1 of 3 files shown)


config/locales/client.en.yml
```diff
@@ -2243,15 +2243,15 @@ en:
       quote_reply: "Quote"
       edit_reason: "Reason: "
       post_number: "post {{number}}"
-      ignored: "Hidden content"
+      ignored: "Ignored content"
```
khalilovcmded: Minor change suggested by Jeff.

PR #7347, opened 2019-04-09
Footer navigation for PWAs and DiscourseHub app

pmusaraj:
This PR adds a new footer bar at the bottom of the app. This will be visible to iOS users in the DiscourseHub app or in a PWA. 

Screenshot:
image

Actions available in the footer nav: 
- back / forward (disabled if actions are unavailable)
- iOS app only: share current page
- iOS app only: dismiss (i.e. close webview)

The footer nav bar is auto-hidden when: 
- user scroll direction is down
- composer is open


SamSaffron:
Awesome! Feel free to merge once you address the open issue with zindex for expanded topic progress

PR #7348, opened 2019-04-09
WIP: Fixed deleted posts layout

venarius:
### Before:
Screenshot from 2019-04-09 21-50-35

### After:
Screenshot from 2019-04-09 22-52-20

Resolves: https://meta.discourse.org/t/deleted-posts-view-needs-some-love/113868

### DIFF
--- a/app/assets/javascripts/discourse/components/user-stream.js.es6
+++ b/app/assets/javascripts/discourse/components/user-stream.js.es6
@@ -9,11 +9,11 @@ import { getOwner } from "discourse-common/lib/get-owner";
 
 export default Ember.Component.extend(LoadMore, {
   _initialize: function() {
-    const stream = this.get("stream");
-    if (stream && stream.get("filter")) {
+    const filter = this.get("stream.filter");
+    if (filter) {
       this.set("classNames", [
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/components/user-stream.js.es6
```diff
@@ -8,6 +8,16 @@ import { popupAjaxError } from "discourse/lib/ajax-error";
 import { getOwner } from "discourse-common/lib/get-owner";
 
 export default Ember.Component.extend(LoadMore, {
+  _initialize: function() {
+    const stream = this.get("stream");
```
ZogStriP: Can you do this instead?

```suggestion
    const filter = this.get("stream.filter");
```

venarius: Oh, didn't know that was possible. Thanks! ❤️ Any idea on the failing tests though? Been on it for some hours now and still can't get them running correctly.

venarius: cc @ZogStriP finally managed to fix them :) Also added a test

PR #7349, opened 2019-04-10
FEATURE: Remove ignore feature SiteSetting and enable ignore by default

khalilovcmded:
## Why?

We are enabling the Ignore feature by default.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -529,9 +529,6 @@ users:
     default: false
     client: true
   log_personal_messages_views: false
-  ignore_user_enabled:
-    default: false
-    client: true
   ignored_users_count_message_threshold:
     default: 5
     client: true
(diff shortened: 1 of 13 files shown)


config/site_settings.yml
```diff
@@ -529,9 +529,6 @@ users:
     default: false
     client: true
   log_personal_messages_views: false
-  ignore_user_enabled:
```
ZogStriP: We need a migration to remove that value form the database (`site_settings` table).

khalilovcmded:
@ZogStriP oh, that's a good point! Just now I realised that these rows will always be persisted.

db/migrate/20190410102915_remove_ignore_user_enabled_site_setting.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveIgnoreUserEnabledSiteSetting < ActiveRecord::Migration[5.2]
```
khalilovcmded: Will that work? @ZogStriP

ZogStriP: Exactly 👍

PR #7350, opened 2019-04-10
UX: new layout for groups page, allow changes to automatic group bio

majakomel:
https://meta.discourse.org/t/give-groups-and-tags-some-love/113783


app/assets/javascripts/discourse/components/groups-form-profile-fields.js.es6
```diff
@@ -21,6 +21,11 @@ export default Ember.Component.extend({
     }
   },
 
+  @computed("model.automatic")
```
ZogStriP: This can be shortened as

``` javascript
canEdit: Ember.computed.not("model.automatic"),
```

PR #7353, opened 2019-04-10
FEATURE: Allow users to tone down digest emails

romanrizzi:


app/views/email/unsubscribe.html.erb
```diff
@@ -52,15 +52,26 @@
         </p>
       <% end %>
 
-      <% if !@topic %>
-        <% unless SiteSetting.disable_digest_emails %>
+      <% unless @topic || SiteSetting.disable_digest_emails %>
```
SamSaffron: unless X || y is very hard to reason about, instead

`if @digest_unsubscribe` or something like that? then you don't reach to site settings from the view.

SamSaffron:
feel free to merge once you address my minor comment

PR #7355, opened 2019-04-11
FIX: correctly retrieve 'login required' setting value on wizard

techAPJ:
FEATURE: extract 'invite only' setting in a separate checkbox control

Screenshot from 2019-04-11 10-25-47

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -4261,10 +4261,12 @@ en:
             choices:
               open:
                 label: "Public"
-                description: "Anyone can access this community and sign up for an account"
+                description: "Anyone can access this community"
               restricted:
                 label: "Private"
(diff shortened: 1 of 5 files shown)


config/locales/server.en.yml
```diff
@@ -4261,10 +4261,12 @@ en:
...
               open:
                 label: "Public"
-                description: "Anyone can access this community and sign up for an account"
+                description: "Anyone can access this community"
               restricted:
                 label: "Private"
-                description: "Only people I have invited or approved can access this community"
+                description: "Only logged in users can access this community"
+          invite_only:
+              placeholder: "Account can be created by invitation only, sign ups are disabled."
```
```suggestion
              placeholder: "People must be explicitly invited. Public registration is disabled."
```

PR #7356, opened 2019-04-11
UX: Switched composer draft saving to animations

venarius:
This adds animations for the composer drafts instead of plain text.

### Before:
ezgif com-gif-maker (1)

### After:
ezgif com-gif-maker

Addition: https://meta.discourse.org/t/mobile-editor-preview-button-and-toolbar/113942/7

Also fixed a random "double declaration" error in `app/assets/javascripts/discourse/components/invite-panel.js.es6`


SamSaffron:
I am totally fine with this as long as its just for mobile, on desktop the
words are important cause there are a few subtle different states.

venarius:
States other than saving and saved will still be displayed as text. I also wrote this into the meta topic.

coding-horror:
Very nice! 😍 does this work @awesomerobot ?

PR #7359, opened 2019-04-11
Remove support for disable_jump_reply user setting

davidtaylorhq:
https://meta.discourse.org/t/new-posts-count-in-browser-tabs-broken/112864/20?u=david


SamSaffron:
Merging this in for now, we can ship a component later

PR #7363, opened 2019-04-11
DEV: Add utf8-cleaner gem

pmusaraj:
Removes invalid UTF-8 characters from environment, preventing errors and warnings in logs like: 

- Discourse::InvalidParameters (string contains null byte)
- Failed to handle exception in exception app middleware : Invalid query parameters: Invalid encoding for parameter: 1����%2527%2522

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -215,3 +215,4 @@ end
 gem 'webpush', require: false
 gem 'colored2', require: false
 gem 'maxminddb'
+gem 'utf8-cleaner'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -443,6 +443,8 @@ GEM
       kgio (~> 2.6)
       raindrops (~> 0.7)
     uniform_notifier (1.12.1)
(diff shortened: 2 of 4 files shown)


SamSaffron:
I am mixed on this, not only does this add a dependency for somewhat trivial code this is adding middleware now to our rack stack per: 

https://github.com/singlebrook/utf8-cleaner/blob/master/lib/utf8-cleaner/railtie.rb

The middleware is performing a bunch of cleanup work which adds a constant cost to every request.

https://github.com/singlebrook/utf8-cleaner/blob/master/lib/utf8-cleaner/middleware.rb

I am still pro simply rejecting people who send us junk utf-8, I think it is the right thing to do.

coding-horror:
Dependencies are guilty until proven innocent, so I say ditch it.

pmusaraj:
Ok, closing, only benefit here would be less noise in logs but it's not worth the performance cost.

PR #7366, opened 2019-04-11
UX: Do not close login modal and search menu on outside mouse up event.

nbianca:


SamSaffron:
@nbianca  looks like prettier is complaining

@eviltrout  can does this look ok to you?

PR #7367, opened 2019-04-11
UX: Move the admin dashboard trend icons inside the tooltip span

curiousdannii:
Just a small change to move the admin dashboard trend icons inside the `<span>` which has the tooltip.


SamSaffron:
thanks @curiousdannii !

jjaffeux:
Hmm thought I had answered 🤔but yes totally fine, thx.

PR #7369, opened 2019-04-12
UX: Changed error draft status to icon

venarius:
This changes the error draft text to an icon which can be hovered to get the error message.

Screenshot 2019-04-12 at 08 37 31 (2)

cc @awesomerobot it currently doesn't have any animations since it's unclear on if we should keep them (2fea29a). If we are going to keep the scaling animation I will also enhance the error icon 🙂

### DIFF
--- a/app/assets/stylesheets/common/base/compose.scss
+++ b/app/assets/stylesheets/common/base/compose.scss
@@ -307,6 +307,11 @@
         margin-right: 5px;
       }
     }
+    #draft-status .d-icon-user-edit {
+      color: $danger;
+      font-size: 20px;
+      vertical-align: -5.5px;
+    }
   }
 
   .composer-bottom-right {
--- a/app/assets/javascripts/discourse/templates/composer.hbs
+++ b/app/assets/javascripts/discourse/templates/composer.hbs
@@ -164,7 +164,9 @@
                 {{/if}}
                 {{#if model.draftSaving}}<div class="spinner small"></div>{{/if}}
                 {{#if model.draftSaved}}{{d-icon 'check' class='save-animation'}}{{/if}}
-                {{model.draftStatus}}
+                {{#if model.draftStatus}}
+                  <span title="{{model.draftStatus}}">{{d-icon 'times'}}</span>
+                {{/if}}
               </div>
             </div>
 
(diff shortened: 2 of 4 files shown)


SamSaffron:
Trouble though is that `x` means close, not failed, we need a better icon here for failure cc @awesomerobot  ?

venarius:
Would look like this then

edit-conflict

coding-horror:
That seems fine

coding-horror:
I suggest we merge this if it looks OK @ZogStriP

ZogStriP:
Nice 👍

app/assets/stylesheets/common/base/compose.scss
```diff
@@ -307,6 +307,11 @@
         margin-right: 5px;
       }
     }
+    #draft-status .d-icon-user-edit {
+      color: $danger;
+      font-size: 20px;
```
jjaffeux: we should use our $font-size macro

app/assets/stylesheets/common/base/compose.scss
```diff
@@ -307,6 +307,11 @@
...
+      vertical-align: -5.5px;
```
jjaffeux: Not a big fan of this kind of magic css value, can’t we do anything better?

app/assets/stylesheets/common/base/compose.scss
venarius: Not really making any sense here since this is only affecting the size of the icon and no text.

app/assets/stylesheets/common/base/compose.scss
venarius: Not as far as I know with my css knowledge

app/assets/stylesheets/common/base/compose.scss
jjaffeux: Yes I kind of agree, but it helps to keep every thing in sync if people change some value.

What it basically means is: this should be bigger than this.

venarius: Sure, yeah. But I don't really get the relation of making text smaller by design and then also making this icon smaller so that no one can see it. I think it should have a fixed size.

Just wanted to say my opinion. If you still want it changed, I can of course do that.

jjaffeux: I don’t have strong feelings on this, we already do this at various places (and we also have fixed sizes have various places). I do think it's better, but let it as it is, fine :p

PR #7371, opened 2019-04-12
UX: Render wizard previews for high-DPI displays

davidtaylorhq:
Sets up a canvas element of twice the required dimensions, scales the coordinate system by 2x, then shrinks the display in css.

Before:
Screenshot 2019-04-12 at 15 29 02

After:
Screenshot 2019-04-12 at 15 28 53

### DIFF
--- a/app/assets/javascripts/wizard/lib/preview.js.es6
+++ b/app/assets/javascripts/wizard/lib/preview.js.es6
@@ -17,24 +17,27 @@ function canvasFor(image, w, h) {
   w = Math.ceil(w);
   h = Math.ceil(h);
 
+  const scale = window.devicePixelRatio;
+
   const can = document.createElement("canvas");
-  can.width = w * 2;
-  can.height = h * 2;
+  can.width = w * scale;
+  can.height = h * scale;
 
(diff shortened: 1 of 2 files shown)


coding-horror:
ohhh nice! Much better! Thank you David! 😍

app/assets/javascripts/wizard/lib/preview.js.es6
```diff
@@ -18,10 +18,11 @@ function canvasFor(image, w, h) {
   h = Math.ceil(h);
 
   const can = document.createElement("canvas");
-  can.width = w;
-  can.height = h;
+  can.width = w * 2;
```
eviltrout: Shouldn't this be `window.devicePixelRatio` ? It would be 2 on a hi dpi display, 1 otherwise.

eviltrout: (same for all the other hard coded 2s below!)

davidtaylorhq: Oh, I didn't know that API existed 😍

SamSaffron:
awesome change!

PR #7372, opened 2019-04-12
FEATURE: Automatically generate optimized site metadata icons

davidtaylorhq:
This change automatically resizes icons for various purposes. Admins can now upload `logo` and `logo_small`, and everything else will be auto-generated. Specific icons can still be uploaded separately if required.

## Core

- Adds an SiteIconManager module which manages automatic resizing and fallback

- Icons are looked up in the OptimizedImage table at runtime, and then cached in Redis. If the resized version is missing for some reason, then most icons will fall back to the original files. Some icons (e.g. PWA Manifest) will return `nil` (because an incorrectly sized icon is worse than a missing icon). 

- `SiteSetting.site_large_icon_url` will return the optimized version, including any fallback. `SiteSetting.large_icon` continues to return the upload object. This means that (almost) no changes are required in core/plugins to support this new system.

- Icons are resized whenever a relevant site setting is changed, and during post-deploy migrations

## Wizard
- Allows `requiresRefresh` wizard steps to reload data via AJAX instead of a full page reload

- Add placeholders to the **icons** step of the wizard, which automatically update from the "Square Logo"
- Various copy updates to support the changes

- Remove the "upload-time" resizing for `large_icon`. This is no longer required.

Screenshot 2019-04-26 at 14 43 45 Screenshot 2019-04-26 at 14 43 50

## Site Settings UX
- Move logo/icon settings under a new "Branding" tab

- Various copy changes to support the changes

- Adds placeholder support to the `image-uploader` component

- Automatically reloads site settings after saving. This allows setting placeholders to change based on changes to other settings

- Upload site settings will be assigned a placeholder if SiteIconManager `responds_to?` an icon of the same name

Screenshot 2019-04-26 at 14 40 10

## Dashboard Warnings

- Remove PWA icon and PWA title warnings. Both are now handled automatically.

## Bonus
- Updated the sketch logos to use @awesomerobot's new high-res designs

Screenshot 2019-04-26 at 14 40 10

Screenshot 2019-04-26 at 14 40 10


lib/site_icon_manager.rb
```diff
@@ -0,0 +1,75 @@
...
+  def self.get_set_cache(key)
+    @cache[key] ||= yield
+  end
+
+  def self.resolve_original(info)
+    info[:settings].each do |setting_name|
+      value = SiteSetting.send(setting_name)
+      return value if value
+    end
+    return Upload.find(SKETCH_LOGO_ID)  if info[:fallback_to_sketch]
```
jjaffeux:
```suggestion
    return Upload.find(SKETCH_LOGO_ID) if info[:fallback_to_sketch]
```

app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
```diff
@@ -17,7 +17,9 @@ export default Ember.Controller.extend({
 
     if ((!filter || 0 === filter.length) && !this.get("onlyOverridden")) {
       this.set("visibleSiteSettings", this.get("allSiteSettings"));
-      this.transitionToRoute("adminSiteSettings");
+      if (this.get("categoryNameKey") === "all_results") {
```
ZogStriP: What is this fixing?

app/models/admin_dashboard_data.rb
```diff
@@ -172,15 +172,6 @@ def github_config_check
     end
   end
 
-  def pwa_config_check
```
ZogStriP: 😍

app/assets/javascripts/admin/controllers/admin-site-settings.js.es6
davidtaylorhq: I'm now reloading all the site settings asynchronously after each "save". This `transitionToRoute` was causing a change in route every time you saved a setting. Now it only transitions if necessary.

spec/components/site_icon_manager_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:upload) do
+    UploadCreator.new(file_from_fixtures("smallest.png"), 'logo.png').create_for(Discourse.system_user.id)
+  end
+
+  it "works correctly" do
+    SiteSetting.logo = nil
+    SiteSetting.logo_small = nil
+
+    # Falls back to sketch for some icons
+    expect(SiteIconManager.favicon.upload_id).to eq(-6)
```
tgxworld:
```suggestion
    expect(SiteIconManager.favicon.upload_id).to eq(SiteIconManager::SKETCH_LOGO_ID)
```

PR #7374, opened 2019-04-12
FIX: Fixed category reordering using arrow icons

venarius:
This fixes not being able to reorder categories with childs using the arrow icons.

### Before
Peek 2019-04-12 20-13

### After
Peek 2019-04-12 20-14

Resolves: https://meta.discourse.org/t/reorder-categories-behaving-strangely/104190/13


SamSaffron:
Thanks heaps for the fix!

PR #7375, opened 2019-04-12
FEATURE: Add setting to strip whitespaces from incoming emails.

udan11:
Some email clients add leading whitespaces which get are transformed in code blocks when processed.


SamSaffron:
I support this change, it is optional opt in, very low risk. 

I wonder if we can come up with an automatic heuristic that is safe enough for default on? (so we would have 3 modes, auto, on, off), auto would detect that you started doing weird stuff and then turn itself on for the rest of the email... something like that. Not sure just brainstorming.

PR #7376, opened 2019-04-13
FIX: properly log webhook errors in UI on rescue

techAPJ:
Currently when the site is unreachable we are not showing the failing webhook status in Admin UX, nor the error is logged in database. Instead we are generating noise in Rails error logs.

After this change there will be clear indication that the webhook is failing and the error will be logged in database.

Screen Shot 2019-04-13 at 17 17 59

Screen Shot 2019-04-13 at 17 17 15

### DIFF
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -113,8 +113,12 @@ module Jobs
           duration: ((Time.zone.now - now) * 1000).to_i
         )
       rescue => e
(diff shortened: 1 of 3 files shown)


SamSaffron:
not a fan of using I am a teapot here, lets just have status code -1 here or if we can then nil. 

once you sort it out feel free to merge

techAPJ:
Thanks for the review @SamSaffron. Updated per your feedback.

PR #7377, opened 2019-04-14
FEATURE: Let staff add custom post notices.

udan11:


app/assets/javascripts/discourse/controllers/add-post-notice.js.es6
```diff
@@ -0,0 +1,38 @@
+import ModalFunctionality from "discourse/mixins/modal-functionality";
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Controller.extend(ModalFunctionality, {
+  notice: null,
+  saving: false,
```
eviltrout: `resolve` should be added as a data member here so that the shape of the JS object needn't change.

app/assets/javascripts/discourse/controllers/add-post-notice.js.es6
```diff
@@ -0,0 +1,38 @@
...
+      const post = this.get("post");
+      const notice = this.get("notice");
+      const resolve = this.get("resolve");
+
+      post.updatePostField("notice", notice).then(() => {
+        post.setProperties({
+          notice_type: "custom",
+          notice_args: notice
+        });
+        resolve();
```
eviltrout: What if the `updatePostField` promise fails? `resolve` will never be called. It should probably be in a `finally` method.

Either that or we should be passing in the `reject` method and calling that from `catch`.

Additionally, if the modal is closed, we'll want to call `resolve` too so we aren't leaking a promise that way.

app/controllers/posts_controller.rb
```diff
@@ -476,6 +476,22 @@ def locked
     render_json_dump(locked: post.locked?)
   end
 
+  def notice
+    raise Discourse::NotFound unless guardian.is_staff?
+
+    post = find_post_from_params
+
+    if params[:notice].present?
+      post.custom_fields["notice_type"] = Post.notices[:custom]
```
SamSaffron: I worry about corruption here long term, can you add unique indexes? 

```
create unique index idx on post_custom_fields(post_id) where name = 'notice_type'
create unique index idx on post_custom_fields(post_id) where name = 'notice_args'
```

SamSaffron:
quick confirmation did you test for N+1?

coding-horror:
where are we at on this @udan11

udan11:
I have added the partial index and checked again that there are no N+1.

As soon as Travis check passes, I will merge it.

PR #7378, opened 2019-04-14
FEATURE: Add support for Unicode usernames

gschlager:
* Adds two site settings to enable Unicode usernames (default off) and whitelist only certain Unicode characters (e.g. `[äöüßÄÖÜẞ]` or `\p{Greek}`). By default Discourse allows letters ([Ll][Ll] / [Lm][Lm] / [Lo][Lo] / [Lt] / [Lu][Lu]), marks ([Mc][Mc], [Me][Me], [Mn][Mn]) and numbers ([Nd][Nd], [Nl][Nl], but _not_ [No][No]). The whitelist can restrict those characters, but it's not possible to allow additional characters. It's not possible to forbid ASCII letters and numbers.
image

* The Letter Avatar Service was updated to v4 and enabling `unicode_usernames` is only possible when `external_system_avatars_enabled` is true, because the internal avatar generator doesn't support Unicode.

* Detects mentions of Unicode usernames on the client and server when `unicode_usernames` is enabled. It works best with browsers that support for Unicode property escapes (currently Chrome and Safari). Others (including our MiniRacer) will use a transpiled version with lots of codepoint ranges. It works quite well, but isn't as perfect as the regex with Unicode property escapes.

* The `TextPostProcessRuler` from the Markdown engine concatenated regular expressions, but ignored their flags. This changes it so that the flags are merged as well. That might lead to unexpected behavior if a matcher works differently with the `u` flag (Unicode). I tested the ones shipped with core and they seem fine.

* There's a new method `User.normalize_username` which converts Unicode strings into the NFC normal form before downcasing the string. This method is used in a lot of places that previously used only `.downcase`. Usernames in the `users` table are always stored in the NFC form and critical methods for looking up existing users always normalize usernames before doing the comparison.

* The `UsernameValidator` and `UserNameSuggester` use the same regular expressions and respect the `unicode_usernames` and `unicode_username_character_whitelist` site settings. Length checks (min/max) and username truncation work with grapheme clusters instead of codepoints, but the max length of 60 codepoints (column limit) is still ensured.

[Ll]: https://www.compart.com/en/unicode/category/Ll
[Lm]: https://www.compart.com/en/unicode/category/Lm
[Lo]: https://www.compart.com/en/unicode/category/Lo
[Lt]: https://www.compart.com/en/unicode/category/Lt
[Lu]: https://www.compart.com/en/unicode/category/Lu
[Mc]: https://www.compart.com/en/unicode/category/Mc
[Me]: https://www.compart.com/en/unicode/category/Me
[Mn]: https://www.compart.com/en/unicode/category/Mn
[Nd]: https://www.compart.com/en/unicode/category/Nd
[Nl]: https://www.compart.com/en/unicode/category/Nl
[No]: https://www.compart.com/en/unicode/category/No


app/assets/javascripts/pretty-text/engines/discourse-markdown/category-hashtag.js.es6
```diff
@@ -44,7 +44,7 @@ function addHashtag(buffer, matches, state) {
 export function setup(helper) {
   helper.registerPlugin(md => {
     const rule = {
-      matcher: /#([\u00C0-\u1FFF\u2C00-\uD7FF\w-:]{1,101})/,
+      matcher: /#([\u00C0-\u1FFF\u2C00-\uD7FF\w:-]{1,101})/,
```
gschlager: The old regex creates a `SyntaxError` when it is used with the `u` flag.

app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
```diff
@@ -20,14 +20,40 @@ function addMention(buffer, matches, state) {
...
+        .source;
+      return new RegExp(
+        `@((?:[_]|${alnum})(?:[._-]|${alnum}){0,58}(?:${alnum}))|@(?:([_]|${alnum}))`
+      );
+    }
+  } else {
+    return /@(\w[\w.-]{0,58}[^\W_])|@(\w)/;
```
gschlager: The old regex didn't detect `@username_` as a valid mention of `@username` even though usernames can't end with an underscore.

app/jobs/regular/update_username.rb
```diff
@@ -5,16 +5,27 @@ class UpdateUsername < Jobs::Base
...
+        (?:
+          (?<![\p{Alnum}\p{M}`])     # make sure there is no preceding letter, number or backtick
+        )
+        @#{@old_username}
+        (?:
+          (?![\p{Alnum}\p{M}_\-.`])  # make sure there is no trailing letter, number, underscore, dash, dot or backtick
+          |                          # or
+          (?=[-_.](?:\s|$))          # or there is an underscore, dash or dot followed by a whitespace or end of line
+        )
+      /ix
```
gschlager: I converted this regex into free-spacing mode and added comments, because mentally parsing a complex expression isn't fun. 😃 And now it supports Unicode.

app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
```diff
@@ -20,14 +20,40 @@ function addMention(buffer, matches, state) {
...
     };
 
     md.core.textPostProcess.ruler.push("mentions", rule);
   });
 }
+
+function mentionRegex(unicodeUsernames) {
+  if (unicodeUsernames) {
+    try {
+      // Create the regex from a string, because Babel doesn't understand
```
eviltrout: Would upgrading babel help here? Or is this always true?

app/jobs/regular/update_username.rb
eviltrout: It looks really nice!

lib/user_name_suggester.rb
```diff
@@ -39,28 +41,51 @@ def self.fix_username(name)
...
   end
 
-  def self.remove_unallowed_trailing_characters(name)
-    name.gsub!(/[^A-Za-z0-9]+$/, "")
+  def self.rightsize_username(name)
+    name = truncate(name, User.username_length.end)
+    name.gsub!(UsernameValidator::INVALID_TRAILING_CHAR_PATTERN, '')
+
+    missing_char_count = User.username_length.begin - name.grapheme_clusters.size
+    name << '1' * missing_char_count if missing_char_count > 0
```
eviltrout: `1` is an interesting padding character. Probably fine.

app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
gschlager: Maybe. I didn't test if a newer version of Babel uses a version of [babel-plugin-transform-unicode-regex](https://babeljs.io/docs/en/next/babel-plugin-transform-unicode-regex.html) that doesn't have that problem anymore. We certainly don't want to use [babel-plugin-proposal-unicode-property-regex](https://babeljs.io/docs/en/next/babel-plugin-proposal-unicode-property-regex.html) because transpiling Unicode property escapes doesn't create a perfect result. I've already added such a transpiled regex as fallback for browsers that do not support this feature yet.

lib/user_name_suggester.rb
gschlager: I think "1" has been used as a padding character for a very long time. I can replace it with something fancier in a follow-up commit if you have a good suggestion.

gschlager:
I still need to figure out why there are build errors when the whole test suite is running. Somehow setting `before { SiteSetting.unicode_usernames = true }` seems to be ignored and randomly results in username validation errors. 🤔 Please do not merge until this is fixed.

app/models/site_setting.rb
```diff
@@ -121,6 +121,11 @@ def self.attachment_filename_blacklist_regex
     @attachment_filename_blacklist_regex ||= Regexp.union(SiteSetting.attachment_filename_blacklist.split("|"))
   end
 
+  def self.unicode_username_character_whitelist_regex
+    @unicode_username_whitelist_regex = SiteSetting.unicode_username_character_whitelist.present? \
```
gschlager: I wanted to memoize the regex, but that broke my specs. Looks like saving a site setting does not remove memoized values. I guess `SiteSetting.attachment_content_type_blacklist_regex` and `SiteSetting.attachment_filename_blacklist_regex` will have the same problem.

app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
SamSaffron: One question here, markdown.it has some stuff here, https://github.com/markdown-it/uc.micro#readme not sure if it is usable or not.

config/locales/server.en.yml
```diff
@@ -2202,6 +2208,7 @@ en:
     username:
       short: "must be at least %{min} characters"
       long: "must be no more than %{max} characters"
+      too_many_codepoints: "is too long"
```
SamSaffron: I am not sure the key here is a bit confusing, as a programmer I understand it, as a translator, this is confusing

lib/user_name_suggester.rb
```diff
@@ -39,28 +41,51 @@ def self.fix_username(name)
...
+
+    while name.length > UsernameValidator::MAX_CHARS
+      clusters.pop
+      name = clusters.join
+    end
+
+    name
   end
 
+  private_class_method :parse_name_from_email
```
SamSaffron: very mixed on this pattern, I would just leave it public really.

test/javascripts/lib/pretty-text-test.js.es6
```diff
@@ -555,6 +555,48 @@ QUnit.test("Mentions", assert => {
...
+    "it detects mentions of valid Unicode usernames"
+  );
+
+  assert.cookedOptions(
+    "@.狮子 @-狮子 @狮子. @狮子_ @狮子-",
+    { siteSettings: { unicode_usernames: true } },
+    '<p>@.狮子 @-狮子 <span class="mention">@狮子</span>. <span class="mention">@狮子</span>_ <span class="mention">@狮子</span>-</p>',
+    "it does not detect mentions of invalid Unicode usernames"
+  );
 });
```
SamSaffron: be sure to also add at least one test into `pretty_text_spec.rb` that way you cleanly integration test the mini racer pipeline

SamSaffron:
This is a very important change and I love how much care you have take here, once you address all the comments feel free to merge.

app/assets/javascripts/pretty-text/engines/discourse-markdown/mentions.js.es6
gschlager: Unfortunately those regular expression aren't usable in our case.

SamSaffron:
@gschlager feel free to merge first thing today when you start your day.

PR #7379, opened 2019-04-15
FIX: Bulk invite should skip invite and add existing users to groups.

tgxworld:
https://meta.discourse.org/t/inviting-users-with-groups-field-failed-when-there-is-an-existing-user-in-the-list/109990


tgxworld:
Thanks for the review @eviltrout :+1:

PR #7380, opened 2019-04-15
UX: Show user card on load

venarius:
This shows the user card even before it finished loading.

### Before
Peek 2019-04-15 19-31

### After
Peek 2019-04-15 20-10

&nbsp;
&nbsp;
This is a first mockup I made for this 🙂 I fiddled a lot around with different animations and even slower opacity fading of the modal but everything felt bad to me. That's why I kept it that way.

Also as far as I know the avatar can't be shown during load, because all the modal knows from the user is his username.

For: https://meta.discourse.org/t/user-more-info-cache-results-and-render-optimization/17578

### DIFF
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -152,7 +152,7 @@ export default Ember.Component.extend(
 
     _showCallback(username, $target) {
       this._positionCard($target);
-      this.setProperties({visible: true, loading: true});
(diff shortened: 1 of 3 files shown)


venarius:
Changed style of animation and use the Cache:

Peek 2019-04-19 00-08

coding-horror:
oooh super nice 🤩

coding-horror:
If this looks OK we should merge it @ZogStriP

ZogStriP:
Cool, let's try it 👍

PR #7381, opened 2019-04-16
FEATURE: Improve dice rolling for discourse-narrative-bot

neemiasvf:
https://meta.discourse.org/t/improve-dice-rolling-for-discourse-narrative-bot/115177

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -215,3 +215,4 @@ end
 gem 'webpush', require: false
 gem 'colored2', require: false
 gem 'maxminddb'
+gem 'games_dice', '~> 0.3.12'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,6 +149,8 @@ GEM
     ffi (1.10.0)
     flamegraph (0.9.5)
     fspath (3.1.0)
+    games_dice (0.3.12)
+      parslet (>= 1.5.0)
     gc_tracer (1.5.1)
     git (1.5.0)
     globalid (0.4.1)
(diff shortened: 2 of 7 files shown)


SamSaffron:
I am definitely not comfortable pulling in a whole gem dependency here for this feature. Not against making it a bit better as long as we don't carry too much extra code.

coding-horror:
Yeah a whole new dependency for this is _no bueno_

neemiasvf:
Then I guess I'll have to publish a plugin. The "problem" is *way more* complex than it seems and I'm not interested in reimplementing everything… :/

SamSaffron:
Sorry about it @neemiasvf  I am just super protective about adding more dependencies to Discourse if you can think of a hook on narrative bot that makes implementing your plugin easier feel free to submit a patch

neemiasvf:
That's okay @SamSaffron. I understand how dependencies can be a headache.

ZogStriP:
> The "problem" is _way more_ complex than it seems and I'm not interested in reimplementing everything… :/

Is it really though? I guess you could just start with the most used notations?

neemiasvf:
It kind of is. Not the actual calculation of the result itself, but the
parser. Plus, the gem is beautifully written. In fact, I could integrate
the gem implementation to the core code. I'm not up to implementing a new
solution to the same problem, but I could recycle the same code. Would that
be acceptable? :)

RvdE:
I'd be very interested in the enhanced dice rolling as well. I've converted from vBulletin 4 a couple of days ago where I had a dice roller plugin that could do math as well. A plugin would be neat as well.

PR #7382, opened 2019-04-16
UX: Allow the contact url setting to be an email

curiousdannii:
A small change to fix what happens if you set the contact url setting to an email address (no longer makes a link to domain.com/emailaddress).


curiousdannii:
I tried getting Prettier to run in VSCode, but couldn't get it to identify any problems.

gschlager:
Thanks for the PR. I'm closing it for now as I think there's a better solution for the problem. https://meta.discourse.org/t/allow-the-contact-url-setting-to-be-an-email/112879/3

We can discuss this further on Meta if you think there's still a need for your solution.

PR #7385, opened 2019-04-16
 REFACTOR: Quick refactor of the webhook event emitter job

khalilovcmded:
## Why?

While looking at some log errors for the bug crushing game, I noticed some Webhook related error.

```
Webhook event failed: connect_write timeout reached
```

There is more of why this could be happening, I can explain more in a separate PR with some ideas. I thought as a first step I would just do a quick refactor of this job to pave the way for a cleaner set of changes / experiments to this job.

### DIFF
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -81,7 +81,7 @@ module Jobs
       if SiteSetting.retry_web_hook_events?
         @retry_count += 1
         return if @retry_count > MAX_RETRY_COUNT
-        delay = RETRY_BACKOFF ** (@retry_count - 1)
+        delay = RETRY_BACKOFF**(@retry_count - 1)
         Jobs.enqueue_in(delay.minutes, :emit_web_hook_event, arguments)
       end
     end
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -25,13 +25,8 @@ module Jobs
     private
 
     def validate_arguments!
-      %i{
-        web_hook_id
-        event_type
-      }.each do |key|
-        raise Discourse::InvalidParameters.new(key) unless arguments[key].present?
-      end
-
+      validate_argument!(:web_hook_id)
+      validate_argument!(:event_type)
       raise Discourse::InvalidParameters.new(:web_hook_id) if web_hook.blank?
     end
 
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -125,35 +125,35 @@ module Jobs
     end
 
     def event_type
-      @event_type ||= arguments[:event_type].to_s
+      arguments[:event_type]
     end
 
     def event_name
-      @event_name ||= arguments[:event_name].to_s
+      arguments[:event_name]
     end
 
     def retry_count
-      @retry_count ||= @arguments[:retry_count]
+      arguments[:retry_count]
     end
 
     def webhook_id
-      @webhook_id ||= @arguments[:web_hook_id]
+      arguments[:web_hook_id]
     end
 
     def group_id
-      @group_id ||= @arguments[:group_id]
+      arguments[:group_id]
     end
 
     def category_id
-      @category_id ||= @arguments[:category_id]
+      arguments[:category_id]
     end
 
     def tag_ids
-      @tag_ids ||= @arguments[:tag_ids]
+      arguments[:tag_ids]
     end
 
     def payload
-      @payload ||= @arguments[:payload]
+      arguments[:payload]
     end
 
     def parsed_payload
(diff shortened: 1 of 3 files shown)


app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -7,62 +7,43 @@ class EmitWebHookEvent < Jobs::Base
...
+      send_webhook!
     end
 
     private
 
-    def guardian
-      Guardian.new(Discourse.system_user)
-    end
+    def validate_arguments!
+      %i{
```
eviltrout: Since there are only two arguments, it would be simpler to do one line per validation than all this ceremony to loop through them.

eviltrout: Additionally, why not use the `validate_argument` helper you created?

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -7,62 +7,43 @@ class EmitWebHookEvent < Jobs::Base
...
-      else
-        body[event_type] = args[:payload]
-      end
+    def validate_argument!(key)
+      raise Discourse::InvalidParameters.new(key) unless arguments[key].present?
+    end
 
-      new_body = Plugin::Filter.apply(:after_build_web_hook_body, self, body)
-      MultiJson.dump(new_body)
+    def memoize_arguments(args)
```
eviltrout: What is the point of the duplication here? Are we modifying the arguments? If so that seems dangerous.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -71,71 +52,164 @@ def web_hook_request(args, web_hook)
...
+    def tag_webhook_invalid?
+      web_hook.tag_ids.present? && (tag_ids.blank? ||
+        (web_hook.tag_ids & tag_ids).blank?)
+    end
+
+    def arguments
+      @arguments
+    end
+
+    def event_type
```
eviltrout: I don't see the advantage of memoizing these values. Hash accesses are really fast. 

If it's to avoid calling `to_s` repeatedly, I would suggest adding @event_type = arguments[:event_type].to_s in the main `execute` method. However most of them are simply numbers and don't seem to gain anything by being memoized.

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -71,71 +52,164 @@ def web_hook_request(args, web_hook)
...
       end
+
+      headers
     end
+
+    def build_webhook_body
+      body = {}
+
+      if ping_event?(event_type)
+        body[:ping] = 'OK'
```
eviltrout: Are we mixing string and symbol keys in the same body? We should pick one type and stick with it.

app/jobs/regular/emit_web_hook_event.rb
khalilovcmded: > I don't see the advantage of memoizing these values. Hash accesses are really fast.

Makes sense, we don't think need memoization except for [this](https://github.com/discourse/discourse/pull/7385/files#diff-0e98636b29e3b908201e344e101066baR164).

> If it's to avoid calling to_s repeatedly, ....

The reason I wrap them into methods is that we are using them across different methods in this job, it feels better in terms of readability. At the first look at the previous version of the file, arguments are scattered all over the place.

app/jobs/regular/emit_web_hook_event.rb
khalilovcmded: Yes, the `.dup` is because there is one argument we are modifying which is the: `retry_count` ... Thinking more about it, we don't need to do that, and I can just allocate an instance attribute for it.

app/jobs/regular/emit_web_hook_event.rb
khalilovcmded: Makes sense 👍

app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -71,71 +48,164 @@ def web_hook_request(args, web_hook)
...
+
+    def arguments
+      @arguments
+    end
+
+    def event_type
+      arguments[:event_type]
+    end
+
+    def event_name
```
eviltrout: Without getting too bike-sheddy I still don't see the advantage here of having methods for hash accessors. For `webhook` or `parsed_payload` it makes sense but otherwise it seems like a bunch of extra lines of code for a very minor amount of syntactic sugar. It won't stop me from approving the PR, but I'm not sure this is the style we want to go with in the long run.

khalilovcmded: I am not strongly opinionated on it either, to me, it is about improving readability.

Some example differences:

```ruby
web_hook.tag_ids.present? && (arguments[:tag_ids].blank? || (web_hook.tag_ids & arguments[:tag_ids]).blank?)
```
 
vs

```ruby
web_hook.tag_ids.present? && (tag_ids.blank? || (web_hook.tag_ids & tag_ids).blank?)
```

This pattern appears few times in this class. The later appears to me more readable.

khalilovcmded: Removed wrapping methods 👍

PR #7388, opened 2019-04-17
FEATURE: Create new helper method 'Discourse.stats'

vinothkannans:
New helper method added to store missing S3 uploads count.


eviltrout:
@vinothkannans It's a bit weird to me to use a PluginStore for a core feature. What's the reasoning for this change?

PR #7389, opened 2019-04-17
 FIX: Force read the last collapsed or hidden post(s) in a topic

khalilovcmded:
## Why?

Aiming to fix [this bug](https://meta.discourse.org/t/will-discourse-ever-gain-ignore-user-functionality/1525/79).

**TL;DR:** If a new post (or few posts) are the last ones in a topic, and are collapsed / hidden, a User will always see the unread counter for this topic (since the posts are never marked as read/seen).
Discourse

## How?

The way it works is by trying to guesstimate if there is gap of more than 1 post between the **last post number** and the **highest seen number** (highest seen number is decided ONLY based on visible posts). 

## UI

1. You will notice I will have 2 unread posts (these are hidden/collapsed posts)

Discourse

2. When I go to the topic, you will NOT see any of the hidden/collapsed posts

3. Then when I go back to the topic list view, you will see that the unread/unseen marker is gone

Discourse

### DIFF
--- a/app/assets/javascripts/discourse/components/scrolling-post-stream.js.es6
+++ b/app/assets/javascripts/discourse/components/scrolling-post-stream.js.es6
@@ -254,6 +254,7 @@ export default MountWidget.extend({
 
     this._previouslyNearby = newPrev;
     this.screenTrack.setOnscreen(onscreenPostNumbers);
+    this.screenTrack.setPosts(posts);
   },
 
   _scrollTriggered() {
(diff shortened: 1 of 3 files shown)


SamSaffron:
this feels very very hacky to me. 

Instead it should be calling `screenTrack.setIgnoredPosts(ignoredPostNumbers);` Then I guess it cant track 1ms for every ignored postNumber. 

It is a far more complex fix cause it runs deeper. But is far easier to comprehend.

SamSaffron:
closing for now in favor of a different fix

PR #7392, opened 2019-04-17
FIX: Disable webhooks on 410 and 404 HTTP responses

khalilovcmded:
## Why?

To design a more deterministic Webhooks retry mechanism, we need to consider:

- Clear understanding of what the sender interpret as a success or failure from the receiver.
- Suitable (or configurable) retry interval
- Suitable (or configurable) maximum number of retries
- Final action to be performed if the Webhook event can’t be delivered.

I think we can improve (1) and (4) by:

- Retry does not make much sense for failures 3xx and 4xx
- On 3xx and 4xx failures, we don’t need to retry.
- After N number of 3xx and 4xx failures, we disable the Webhook.
- On 2xx and 5xx, a retry is worth doing (as we do now)
- On 410 which is GONE, we disable the Webhook immediately – Zapier, and Intercom are two examples of platforms that do so.

-----

This PR addresses HTTP responses with **404 and 410** only.

### DIFF
--- a/app/jobs/regular/emit_web_hook_event.rb
+++ b/app/jobs/regular/emit_web_hook_event.rb
@@ -41,27 +41,21 @@ module Jobs
 
     def send_webhook!
       uri = URI(web_hook.payload_url.strip)
-
-      conn = Excon.new(
-        uri.to_s,
-        ssl_verify_peer: web_hook.verify_certificate,
-        retry_limit: 0
-      )
+      conn = Excon.new(uri.to_s, ssl_verify_peer: web_hook.verify_certificate, retry_limit: 0)
 
       web_hook_body = build_webhook_body
       web_hook_event = create_webhook_event(web_hook_body)
       web_hook_headers = build_webhook_headers(uri, web_hook_body, web_hook_event)
-
-      response = nil
+      web_hook_response = nil
 
       begin
         now = Time.zone.now
-        response = conn.post(headers: web_hook_headers, body: web_hook_body)
+        web_hook_response = conn.post(headers: web_hook_headers, body: web_hook_body)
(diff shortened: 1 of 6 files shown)


app/jobs/regular/emit_web_hook_event.rb
```diff
@@ -74,7 +68,20 @@ def send_webhook!
...
+    end
+
+    def process_webhook_response(web_hook_response)
+      return if web_hook_response&.status.blank?
+
+      case web_hook_response.status
+      when 200..299
+        return
+      when 404, 410
+        web_hook.update_attribute(:active, false) if @retry_count >= MAX_RETRY_COUNT
```
ZogStriP: I think this is worth a log in in order to keep track of why that webhook was deactivated.

khalilovcmded: Yes, that's actually what I was hoping I get feedback on, so thank you ⭐️ 

What type or kind of log do can I have here? Are we talking about `Rails.logger.info` or we have some other type of logging within Discourse?

ZogStriP: [StaffActionLog](https://github.com/discourse/discourse/blob/master/app/services/staff_action_logger.rb) is what I would recommend.

app/services/staff_action_logger.rb
```diff
@@ -599,6 +599,19 @@ def log_web_hook(web_hook, action, opts = {})
     ))
   end
 
+  def log_web_hook_deactivate(web_hook, response_http_status, opts = {})
```
khalilovcmded: @ZogStriP Does this match what you were thinking of?

PR #7397, opened 2019-04-18
DEV: Test query volume policing

danielwaterworth:
There are tests that do >1,900 queries and I think this is far too many.

As things stand, it's very easy for the number of queries per test to creep upwards over time. The goal of this PR is make this creep visible.

It works by counting the number of queries that are executed per test. When this exceeds the threshold, it raises an exception.

There are two ways to deal with this exception (asides from reducing the number of queries). You can tag the test with `allow_many_queries`, for example:

    it "should ...", allow_many_queries: true do
      ...
    end

Or, you can increase the threshold. Either way, it creates a place in the codebase where a discussion around test performance can happen during code review.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -240,6 +240,32 @@ RSpec.configure do |config|
     end
   end
 
+  # Let's make tests that do too many queries more visible.
+  max_queries = 2000
+  num_queries = 0
+
+  config.before(:suite) do
+    ActiveSupport::Notifications
+      .subscribe('sql.active_record') do |name, start, finish, id, event|
+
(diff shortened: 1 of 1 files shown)


danielwaterworth:
Here's the distribution of the number of queries per test (number of queries along x, number of tests along y).

test_query_distribution

SamSaffron:
I am not against doing something here, but it needs to be default off, use ENV to turn that on. Also keep in mind your counting here will miss out on mini_sql, you want to use the method profiler to catch everything.

danielwaterworth:
I'll rethink this.

PR #7399, opened 2019-04-18
FIX: Add unique index on group_requests(group_id, user_id).

nbianca:


eviltrout:
This is out of scope for this PR but I wonder if we considered using the new review queue to review group requests?

app/controllers/groups_controller.rb
```diff
@@ -408,12 +408,13 @@ def remove_member
...
-    unless current_user.staff?
-      RateLimiter.new(current_user, "request_group_membership", 1, 1.day).performed!
-    end
-
     group = find_group(:id)
-    group_name = group.name
+
+    begin
+      GroupRequest.create(group: group, user: current_user, reason: params[:reason])
+    rescue
```
eviltrout: You should add a specific exception you want to catch here. Otherwise the error will occur on things like typos, nil exceptions, etc. Additionally, it should be `create!` if you want to raise an exception on validation error.

SamSaffron:
Nice change!

PR #7400, opened 2019-04-18
FEATURE: generic theme component bbcode wrapper

jjaffeux:
Usage:

```
[wrap=name foo=bar]
hello world
[/wrap]
```

### DIFF
--- a/app/assets/javascripts/pretty-text/engines/discourse-markdown/d-wrap.js.es6
+++ b/app/assets/javascripts/pretty-text/engines/discourse-markdown/d-wrap.js.es6
@@ -4,8 +4,9 @@ const WRAP_CLASS = "d-wrap";
 
 function parseAttributes(tagInfo) {
   const attributes = tagInfo.attrs._default || "";
+
   return (
-    parseBBCodeTag("[wrap wrap=" + attributes + "]", 0, attributes.length + 12)
+    parseBBCodeTag(`[wrap wrap=${attributes}]`, 0, attributes.length + 12)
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/pretty-text/engines/discourse-markdown/d-wrap.js.es6
```diff
@@ -0,0 +1,47 @@
...
+  token = new state.Token("div_open", "div", 1);
+  token.attrs = [["class", "d-wrap"]];
+
+  const attributes = parsed.attrs || {};
+  const content = attributes.content;
+  delete attributes.content;
+
+  Object.keys(attributes).forEach(tag => {
+    const value = attributes[tag];
+    token.attrs.push([`data-${tag}`, state.md.utils.escapeHtml(value)]);
```
eviltrout: I assume this can't be used for injection right? The parser is smart enough to remove quotes and such?

I would feel better if the attribute name was limited to a regexp, say `[a-z0-9\-]` ?

jjaffeux: I think the spec is rather safe on this and browsers would as you assume get rid of any malicious attempt.

For extra safety I'm also escaping attribute and added two specs to prevent basic regression on this area.

jjaffeux:
(test failure unrelated)

SamSaffron:
hold tight ... discussing this on dev

PR #7401, opened 2019-04-18
DEV: Replace site_setting_saved DiscourseEvent with site_setting_changed

davidtaylorhq:
site_setting_saved is confusing for a few reasons:
- It is attached to the after_save of the ActiveRecord model. This is confusing because it only works 'properly' with the db_provider
- It passes the activerecord model as a parameter, which is confusing because you get access to the 'database' version of the setting, rather than the ruby setting. For example, booleans appear as 'y' or 'n' strings.
- When the event is called, the local process cache has not yet been updated. So if you call SiteSetting.setting_name inside the event handler, you will receive the old site setting value

I have deprecated that event, and added a new site_setting_changed event. It passes three parameters:
- Setting name (symbol)
- Old value (in ruby format)
- New value (in ruby format)

It is triggered after the setting has been persisted, and the local process cache has been updated.

This commit also includes a test case which describes the confusing behavior. This can be removed once site_setting_saved is removed.

### DIFF
--- a/config/initializers/014-track-setting-changes.rb
+++ b/config/initializers/014-track-setting-changes.rb
@@ -1,6 +1,4 @@
 DiscourseEvent.on(:site_setting_changed) do |name, old_value, new_value|
-  next if old_value == new_value
-
   # Enabling `must_approve_users` on an existing site is odd, so we assume that the
   # existing users are approved.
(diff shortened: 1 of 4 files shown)


davidtaylorhq:
:+1: sounds sensible. Updated

PR #7402, opened 2019-04-18
DEV: add site setting type group_list for a list of groups

nlalonde:
I want to add a new setting to the ad plugin which takes a list of groups, so I'm adding this new setting type.

It would be nice to have a js test for this since there are no core settings that are using it, but didn't have success writing one. Something like this?

```
import componentTest from "helpers/component-test";
moduleForComponent("group-list", { integration: true });

componentTest("site-setting-group-list", {
  template: '{{site-settings/group-list value=values}}',

  beforeEach() {
    this.setProperties({
      ???
    });
  },

  async test(assert) {
    
  }
});

```

### DIFF
--- a/app/assets/javascripts/admin/components/site-settings/group-list.js.es6
+++ b/app/assets/javascripts/admin/components/site-settings/group-list.js.es6
@@ -1,0 +1,8 @@
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  @computed()
+  groupChoices() {
+    return this.site.get('groups').map(g => g.name);
+  }
+});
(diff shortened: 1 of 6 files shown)


jjaffeux:
@neil I suggest the following test

[group-list..patch.txt](https://github.com/discourse/discourse/files/3122195/group-list.patch.txt)

the group-list.js.es6 modification is only a prettier fix for quotes

nlalonde:
Committed with the test. Thanks @jjaffeux !

https://github.com/discourse/discourse/commit/1ba72ccab8698c156e502452a036bc7be6e7857b

PR #7404, opened 2019-04-19
UX: User card loading

venarius:
As a follow-up to #7380

Since the caching functionality I added only worked for the current user and not for the user who got clicked on, it had to be removed. As a compensation I also now added a avatar placeholder.

Peek 2019-04-19 12-09

cc @ZogStriP I also opened a PR in discourse-cakeday for the tests https://github.com/discourse/discourse-cakeday/pull/26

### DIFF
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -152,7 +152,7 @@ export default Ember.Component.extend(
 
     _showCallback(username, $target) {
       this._positionCard($target);
-      this.setProperties({visible: true, loading: true});
+      this.setProperties({ visible: true, loading: true });
 
       const args = { stats: false };
       args.include_post_count_for = this.get("topic.id");
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -152,8 +152,7 @@ export default Ember.Component.extend(
 
     _showCallback(username, $target) {
       this._positionCard($target);
-      const currentUser = Discourse.__container__.lookup("current-user:main");
-      this.setProperties({ visible: true, loading: true, user: currentUser });
+      this.setProperties({ visible: true, loading: true });
 
       const args = { stats: false };
       args.include_post_count_for = this.get("topic.id");
@@ -165,7 +164,7 @@ export default Ember.Component.extend(
               user.topic_post_count[args.include_post_count_for]
             );
           }
-          this.setProperties({ user, loading: false });
+          this.setProperties({ user });
         })
         .catch(() => this._close())
         .finally(() => this.set("loading", null));
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -53,6 +53,7 @@ export default Ember.Component.extend(
     showCheckEmail: Ember.computed.and("user.staged", "canCheckEmails"),
 
     user: null,
+    loading: false,
 
     // If inside a topic
     topicPostCount: null,
@@ -150,6 +151,9 @@ export default Ember.Component.extend(
     },
 
     _showCallback(username, $target) {
+      this._positionCard($target);
+      this.setProperties({visible: true, loading: true});
+
       const args = { stats: false };
       args.include_post_count_for = this.get("topic.id");
       User.findByUsername(username, args)
@@ -160,8 +164,7 @@ export default Ember.Component.extend(
               user.topic_post_count[args.include_post_count_for]
             );
           }
-          this._positionCard($target);
-          this.setProperties({ user, visible: true });
+          this.setProperties({ user, loading: false });
         })
         .catch(() => this._close())
         .finally(() => this.set("loading", null));
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -152,7 +152,8 @@ export default Ember.Component.extend(
 
     _showCallback(username, $target) {
       this._positionCard($target);
-      this.setProperties({ visible: true, loading: true });
+      const currentUser = Discourse.__container__.lookup("current-user:main");
+      this.setProperties({ visible: true, loading: true, user: currentUser });
 
       const args = { stats: false };
       args.include_post_count_for = this.get("topic.id");
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -53,7 +53,6 @@ export default Ember.Component.extend(
     showCheckEmail: Ember.computed.and("user.staged", "canCheckEmails"),
 
     user: null,
-    loading: false,
 
     // If inside a topic
     topicPostCount: null,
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/common/components/user-card.scss
```diff
@@ -2,6 +2,34 @@ $card_width: 580px;
...
+  }
+}
+.animated-placeholder {
+  animation-duration: 1.25s;
+  animation-fill-mode: forwards;
+  animation-iteration-count: infinite;
+  animation-name: placeHolderShimmer;
+  animation-timing-function: linear;
+  background: darkgray;
+  background: linear-gradient(to right, #eeeeee 10%, #dddddd 18%, #eeeeee 33%);
```
ZogStriP: Can these colors be somehow extracted from the current theme colors?

app/assets/stylesheets/common/components/user-card.scss
```diff
@@ -2,6 +2,34 @@ $card_width: 580px;
...
+    background-position: 468px 0;
+  }
+}
+.animated-placeholder {
+  animation-duration: 1.25s;
+  animation-fill-mode: forwards;
+  animation-iteration-count: infinite;
+  animation-name: placeHolderShimmer;
+  animation-timing-function: linear;
+  background: darkgray;
```
ZogStriP: Can this color be extracted from the current theme colors?

ZogStriP:
Looks better but the cakeday acceptance test is still broken ;)

venarius:
Hmm, working for me. Can you increase the timeout in the test and check if it still happens, @ZogStriP ?

venarius:
Also, did you test with my branch of discourse-cakeday, which contains the fix? I also opened a PR for it in the plugins repo.

app/assets/stylesheets/common/components/user-card.scss
```diff
@@ -2,6 +2,44 @@ $card_width: 580px;
...
+    to right,
+    $primary-low 10%,
+    $primary-low-mid 18%,
+    $primary-low 33%
+  );
+  height: 20px;
+  position: relative;
+}
+.card-avatar-placeholder {
+  width: 120px;
```
udan11:
```suggestion
  width: $avatar_width;
```

app/assets/stylesheets/common/components/user-card.scss
```diff
@@ -2,6 +2,44 @@ $card_width: 580px;
...
+  height: 120px;
```
udan11:
```suggestion
  height: $avatar_width;
```

app/assets/javascripts/discourse/components/user-card-contents.js.es6
```diff
@@ -53,6 +53,7 @@ export default Ember.Component.extend(
     showCheckEmail: Ember.computed.and("user.staged", "canCheckEmails"),
 
     user: null,
+    loading: false,
```
udan11: Minor nit-pick: this line is not actually useful since `loading` is already defined in `CardContentsBase` mixin.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -92,6 +96,21 @@
       tagName=""}}
     </div>
 
+    {{#if loading}}
```
udan11: I am wondering, wouldn't it be easier to insert at line 3 something like:

```handlebars
{{#if loading}}
  <div class="card-row first-row">
    <div class="user-card-avatar">
      <div class="card-avatar-placeholder animated-placeholder"></div>
    </div>
  </div>

  <div class="card-row second-row">
    <div class="animated-placeholder"></div>
  </div>
  <div class="card-row third-row">
    <div class="animated-placeholder"></div>
  </div>
  <div class="card-row fourth-row">
    <div class="animated-placeholder"></div>
  </div>
  <div class="card-row sixth-row">
    <div class="animated-placeholder"></div>
  </div>
{{else}}
  {{!-- current code --}}
{{/if}}
```

This way you will not need the changes to discourse-cakeday and who knows how many other private plugins.

Also, I believe this will fix other plugin connectors that might break because are not in a `{{if loading}}...{{/if}}` block.

app/assets/javascripts/discourse/components/user-card-contents.js.es6
venarius: Thanks, didn't know that :)

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
venarius: Pushed this approach because I think it is more clean than multiple `#if loading` in between lines but that still didn't fix the error with `discourse-cakeday` because its test instantly tries to find a class on the card as soon as it appears and then gets nothing because the user hasn't finished loading yet.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -1,223 +1,263 @@
...
+      <div class="card-row first-row">
+        <div class="user-card-avatar">
           {{#if user.profile_hidden}}
-            <span class="name-username-wrapper">
-              {{if nameFirst user.name (format-username username)}}
-            </span>
+            <span class="card-huge-avatar">{{bound-avatar user "huge"}}</span>
           {{else}}
-            <a href="{{user.path}}" {{action "showUser" user}} class='user-profile-link'>
+            {{#if loading}}
```
udan11: It does not make sense anymore to keep this check here because we already checked for this on line 3 and it is ensured loading will always be falsey.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -1,223 +1,263 @@
...
-      name="user-card-additional-controls"
-      args=(hash user=user close=(action "close"))
-      tagName=""}}
-    </div>
 
-    {{#if user.profile_hidden}}
-      <div class="card-row second-row">
-        <div class='profile-hidden'>
-          <span>{{i18n "user.profile_hidden"}}</span>
+      {{#if loading}}
```
udan11: The same as above.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -1,223 +1,263 @@
...
+                  <span title={{user.website}}>{{user.website_name}}</span>
+                {{/if}}
+              </span>
+            {{/if}}
+            {{plugin-outlet name="user-card-location-and-website" args=(hash user=user)}}
+          </div>
+        </div>
+      {{/if}}
+
+      {{#unless loading}}
```
udan11: The same as above.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
udan11: Ah, I understand. Had another look at the way `discourse-cakeday` tests this and I see that the check is wrapped under a `andThen`. It should work, which makes me think that the promise chain is broken at some point (which makes `andThen` proceed before the card is actually shown).

venarius: No the chain is correct but when card is shown it only shows the loading animation until the user is loaded but the `find()` already takes place. It worked before because the card only showed after the user has been fully loaded.

venarius:
Thanks for the great input, @udan11 👍

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
venarius: Ouh yeah, forgot to remove that ._.

SamSaffron:
@udan11 if you are happy with the change can you merge it in first thing today?

PR #7405, opened 2019-04-19
DEV: User email test optimization

danielwaterworth:
This PR removes unnecessary fabrications of user emails. It wasn't quite the speed boost I was hoping for, but it does reduce the number of fabricated objects by ~7,000 and speeds up the tests by ~20s.

This PR makes the default `:user` fabricator create a user with a single email address. When you need a user with a secondary email, you can use the old `:user` fabricator which has been renamed to `:user_with_secondary_email`.

### DIFF
--- a/spec/fabricators/user_fabricator.rb
+++ b/spec/fabricators/user_fabricator.rb
@@ -1,7 +1,7 @@
 Fabricator(:user_stat) do
 end
 
-Fabricator(:user_single_email, class_name: :user) do
+Fabricator(:user, class_name: :user) do
   name 'Bruce Wayne'
   username { sequence(:username) { |i| "bruce#{i}" } }
   email { sequence(:email) { |i| "bruce#{i}@wayne.com" } }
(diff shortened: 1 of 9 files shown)


tgxworld:
Thank you for the change here. We should default `Fabricactor(:user)` to one that only contains a single primary email IMO. Any spec that requires  a secondary email can fabricate the user separately.

SamSaffron:
Agree with @tgxworld here ... we should change the default.

danielwaterworth:
I'm glad we agree. That is the effect of this PR.

danielwaterworth:
I was expecting more of the tests to depend on fabricated users have multiple email addresses (which is why I did it this way with diffs affecting all of the specs). Since this wasn't the case, I can make this PR much easier to review by just changing the fabricators and fixing up any broken tests afterwards.

SamSaffron:
cool, yes, please do so!

danielwaterworth:
Only three tests relied on having a user with multiple email addresses.

tgxworld:
Thank you for this @danielwaterworth  :+1:

danielwaterworth:
@tgxworld, thank you for merging it :smile:

PR #7407, opened 2019-04-19
FEATURE: treats @username as username in search-menu

jjaffeux:


jjaffeux:
@SamSaffron this is the simplest change I could come up with while keeping (I think) all the existing behaviour.

A note though, if you type "@sa" and your query limit is 3 chars, it will go under query limit, which is kind of annoying, but I think it's fine as it's 1) kind of edge case 2) it will still show a message "Your search term is too short."

SamSaffron:
I think this is not going to work right, if the context is search within topic it should find all my posts. 

can you link me to the meta topic about this, I am very mixed on this change.

jjaffeux:
> I think this is not going to work right, if the context is search within topic it should find all my posts.
> 
> can you link me to the meta topic about this, I am very mixed on this change.

https://meta.discourse.org/t/search-username-with-and-without-should-both-show-user/110343/2?u=joffreyjaffeux

jjaffeux:
Also I'm not sure I follow the comment, might be missing a bit here, as it won’t be changing behaviour if there's a search context:

Screenshot 2019-04-23 at 10 28 27
Screenshot 2019-04-23 at 10 28 21
Screenshot 2019-04-23 at 10 28 42
Screenshot 2019-04-23 at 10 28 48

PR #7409, opened 2019-04-19
FEATURE: adds an updateComponent hook to plugin-connector

jjaffeux:
This commit allows plugin authors to hook into args mutations. setupComponent is called a the beginning and then you had no way to hook into this.


jjaffeux:
@eviltrout could you have a look at this please?

This is needed for https://github.com/discourse/discourse-cakeday/pull/27

See https://github.com/discourse/discourse/pull/7404 and especially https://github.com/discourse/discourse-cakeday/pull/26/files#diff-5a99943ffc61d3adce817e9d063dbd96R230 for more context

eviltrout:
Why'd you close this? Did you change your mind or are you going to resubmit?

PR #7413, opened 2019-04-22
FIX: remove like_count and <hr> tag from post crawler layout

mrfinch:


eviltrout:
Is there a related discussion somewhere for context on this change? I understand the link change, but why remove the hr?

mrfinch:
> Is there a related discussion somewhere for context on this change? I understand the link change, but why remove the hr?

It is because there are 2 horizontal lines shown in the view. We don't need both so I'm removing one. You can have a look at it here: https://d11a6trkgmumsb.cloudfront.net/original/3X/c/4/c4fc4db55a8afc3aae6046b5a43495c643f1fa65.png

I'm doing this as a part of https://meta.discourse.org/t/improving-discourse-static-html-archive/112497/6

SamSaffron:
I am still not following whey we want the layout to say "0 likes"?

mrfinch:
> I am still not following whey we want the layout to say "0 likes"?

I don't understand. Do you want to remove likes itself or have a problem with how we are presenting it? This PR will show the number of likes post has if it is greater than zero otherwise nothing.

SamSaffron:
I see oops, I read that backwards 

I am not sure if I would keep the span for the 0 case, but yeah this is ok. thanks!

PR #7415, opened 2019-04-23
FIX: Hide ignoring users in preference for users with TL less than a member

khalilovcmded:


app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -6,6 +6,12 @@ import User from "discourse/models/user";
 export default Ember.Controller.extend(PreferencesTabController, {
   saveAttrNames: ["muted_usernames", "ignored_usernames"],
   ignoredUsernames: Ember.computed.alias("model.ignored_usernames"),
+  userIsMemberOrAbove: Ember.computed.gte("model.trust_level", 2),
+  userIsStaffOrModerator: Ember.computed.or("model.staff", "model.moderator"),
```
ZogStriP: Hmm, usually the condition is `staff = moderator || admin`. So checking only for staff should be ok?

khalilovcmded: Ah! Right 👍

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -6,6 +6,9 @@ import User from "discourse/models/user";
 export default Ember.Controller.extend(PreferencesTabController, {
   saveAttrNames: ["muted_usernames", "ignored_usernames"],
   ignoredUsernames: Ember.computed.alias("model.ignored_usernames"),
+  userIsMemberOrAbove: Ember.computed.gte("model.trust_level", 2),
+  userIsStaff: Ember.computed.bool("model.staff"),
```
ZogStriP: I don't think this line is needed anymore?

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -6,6 +6,9 @@ import User from "discourse/models/user";
...
+  ignoredEnabled: Ember.computed.or("userIsMemberOrAbove", "userIsStaff"),
```
ZogStriP: You should be able to remove the preceding line by doing this, right?

```suggestion
  ignoredEnabled: Ember.computed.or("userIsMemberOrAbove", "model.staff"),
```

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
khalilovcmded: Yes, I was hoping to make it clearer, but sure 👍

app/assets/javascripts/discourse/controllers/preferences/users.js.es6
```diff
@@ -6,6 +6,8 @@ import User from "discourse/models/user";
 export default Ember.Controller.extend(PreferencesTabController, {
   saveAttrNames: ["muted_usernames", "ignored_usernames"],
   ignoredUsernames: Ember.computed.alias("model.ignored_usernames"),
+  userIsMemberOrAbove: Ember.computed.gte("model.trust_level", 2),
+  ignoredEnabled: Ember.computed.or("userIsMemberOrAbove", "model.staff"),
```
eviltrout: You are duplicating the logic here from the server side, which is always dangerous as they can easily go out of sync. Instead you should pass `can_ignore_users` via a serializer from the server side to the client and use that.

khalilovcmded: I initially thought of doing so, then I thought that perhaps the front-end should be responsible for this because it is logic to show/hide a component (it is UI specific).

It may be better be to keep it in the backend because of the `go out of sync` problem, but I disagree that it is `dangerous` as it is, the side effect of a bug showing the ignore in preferences isn't going to affect how the feature work, because we always have the `guardian` on the backend checking the ignore requests to be valid or not.

eviltrout: Sorry "dangerous" was too severe a word. It is more of an "easy bug" waiting to happen in the future if we change the logic. For other code like this we pass it through in a serializer, and we should be consistent with this feature.

khalilovcmded: > For other code like this we pass it through in a serializer, and we should be consistent with this feature.

That makes sense 👍 I will cut another PR to refactor it. Thanks for the feedback 🙏

PR #7417, opened 2019-04-23
show read more from category link in crawler view of topic

mrfinch:
I tested it on my local but link gets rendered at the text. Is it due to some set-up issue on local or am I doing something wrong here
Screenshot from 2019-04-23 19-49-19

This is a part of https://meta.discourse.org/t/improving-discourse-static-html-archive/112497/6, it will improve the user's interactivity on the static website when the user scrolls at the end of a topic.

Also, should I change `{{}}` to `%{}` in all locales file or just `en` one is fine?

### DIFF
--- a/app/views/topics/show.html.erb
+++ b/app/views/topics/show.html.erb
@@ -98,6 +98,12 @@
   </div>
 <% end %>
 
+<% if @topic_view.topic && @topic_view.topic.category %>
+  <div class="clearfix topic-body">
+    <%= t('js.topic.read_more_in_category', catLink: "<a href=\"#{@topic_view.topic.category.url}\">#{CGI.escapeHTML(@topic_view.topic.category.name)}</a>", latestLink: "<a href=\"/latest\">#{t("js.topic.view_latest_topics")}</a>") %>
+  </div>
+<% end %>
+
 <% end %>
 
 <% content_for :head do %>
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1893,7 +1893,7 @@ en:
       options: "Topic Options"
       show_links: "show links within this topic"
       toggle_information: "toggle topic details"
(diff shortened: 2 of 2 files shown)


SamSaffron:
Sorry but we can not have this inconsistency in the translation. 

Either we use `{{}}` everywhere on the client.en file or we use `%{}` everywhere we can not mix and match.

mrfinch:
> Sorry but we can not have this inconsistency in the translation.
> 
> Either we use `{{}}` everywhere on the client.en file or we use `%{}` everywhere we can not mix and match.

Ok, so should I change all translation files to use `%{}` instead then so that ruby substitutions work fine?

Also, any idea why `<a>` is not rendering properly in my local?

gschlager:
@SamSaffron The `client.en.yml` already contains a mixture of `{{var}}` and `%{var}` because we either use a string on the server and client (the server requires `%{var}`) or because it's simply hard to select the right format when you see both being used in the same file. The client supports both formats.

I'd like us to switch to `%{var}` in every file and enforce it with a spec. That makes it easier for devs to stick to one format that works everywhere. But this needs to happen without us invalidating existing translations on Transifex. I'm currently testing a process to make this less painless based on a script I wrote a few months ago because I have a couple of similar changes lined up.

I think we should close this PR for now. I'm against duplicating the string in `server.en.yml` and losing an important translation like `js.topic.read_more_in_category` due to a format change for interpolation keys is not an option. We can come back to this at a later time.

SamSaffron:
@gschlager  yes lets close this for now, do a topic on dev to track the consistifying of all strings, cause we can't have a mishmash

gschlager:
> Also, any idea why <a> is not rendering properly in my local?

For future reference: By default HTML is escaped unless the translation key ends with `_html`. It should work when you use `<%= raw t('js.topic.read_more_in_category', ...) %>`

See https://guides.rubyonrails.org/i18n.html#using-safe-html-translations for details.

PR #7419, opened 2019-04-23
UX: consistent email color customization 

hnb-ku:
### summary

Right now, the emails a Discourse site sends have their colors set based on a couple of things. Some elements have their colors set based on the site's active theme, other elements have their colors set based on three site settings. This PR aims to fix the inconsistency here by making email colors independant of the site's active theme. With the changes in this PR all customizable colors in emails will be based on 5 email color site settings.  

#### 1. Email elements with colors set based on the active theme

* the email header (where the logo is) gets the Discourse header background - or the `header_background` color in the active color palette 

* the email header also gets the Discourse header-text-color - or the `header_primary` color in the palette  - for text color - this only comes into effect if the site uses a text-logo

* the email footer links get the tertiary color from the active color palette. Well at least two of them do for now

The "your email settings" link is supposed to be the same color, but it's currently not since we don't set its color to `@anchor_color` like we do with the two other links in email footers

https://github.com/discourse/discourse/blob/master/app/views/user_notifications/digest.html.erb#L421-L423

This covers the elements that have their colors set based on the active theme's colors. 

#### 2. Email elements with colors set based on site settings

There are also other elements in emails that get their colors set based on three site settings like I mentioned earlier.

The relevant site settings are:

`email_accent_bg_color`
`email accent fg color`
`email_link_color`

##### a. The `email_accent_bg_color` site setting

controls the background color of this area 

the background color of the "read more" buttons in digest emails

the background color of the "visit topic" buttons in reply notification emails

and the background color of the "visit topic" buttons in mailing list emails

##### b. The `email_accent_fg_color` site setting

controls the color of the text added here

the color of the text in the "read more" buttons in digest emails

the text color of the "visit topic" buttons in reply notification emails

and the text color of the "visit topic" buttons in mailing list emails

##### c. The `email_link_color` site setting 

sets the colors of all the links in emails except for the footer links discussed earlier. So it sets the colors of links here

and the colors of all links in the reply notification emails

as well as link colors in mailing list emails

#### What will change?

Once the changes in this PR are merged, there will be three changes to how colors in emails work

1. The email header background and text colors will no longer be set based on the theme colors. Instead, they will be set based on the two settings added in this PR. The `email_header_bg_color` and the `email_header_fg_color` This is a breaking change and will require follow up on our side<sup> Footnote 1</sup>
2. The email footer links will all be the same color
3. Changing one of the email color settings will result in changes in all relevant items. In other words, you only need to make changes in one place instead of two separate places like you do now.

##### Footnotes
1. Once these changes are merged, I will write an announcement topic on Meta explaining the changes. The only colors that will actually change are the email header background and text colors - since those won't be based on the theme anymore. 

2. While trying to fix the relevant spec, I noticed that the styles added here 

    https://github.com/hnb-ku/discourse/blob/master/lib/email/styles.rb#L164

    do not apply when I run the spec, this resulted in the spec always failing. So, I removed those styles 
    and instead opted for adding them to the actual tags. 

3. While it's possible to make all the email colors be based on the theme colors and remove all email color settings, I opted for site settings because emails are a different medium and it makes sense to separate their customizations. A set of colors might look good in a theme, but might not be ideal in an email. Site settings is a much less complicated way of doing it.


app/helpers/user_notifications_helper.rb
```diff
@@ -26,7 +26,7 @@ def logo_url
   end
 
   def html_site_link(color)
-    "<a href='#{Discourse.base_url}' style='color: ##{color}'>#{@site_name}</a>"
```
ZogStriP: Does that mean that `color` is **always** a string with the `#` prefix?

ZogStriP:
> 1\. The email header background and text colors will no longer be set based on the theme colors. Instead, they will be set based on the two settings added in this PR. The `email_header_bg_color` and the `email_header_fg_color` This is a breaking change and will require follow up on our side Footnote 1

We could write a migration that would set the new site settings to the default theme values.



> 3\. While it's possible to make all the email colors be based on the theme colors and remove all email color settings, I opted for site settings because emails are a different medium and it makes sense to separate their customizations. A set of colors might look good in a theme, but might not be ideal in an email. Site settings is a much less complicated way of doing it.

The caveat of using site settings for these values is that there can only be 1 email style. Whereas if we use the theme values, we can for example have a light and a dark styling of the emails.

hnb-ku:
> We could write a migration that would set the new site settings to the default theme values.

Yes, that would be a much better way of doing it. 

> The caveat of using site settings for these values is that there can only be 1 email style. Whereas if we use the theme values, we can for example have a light and a dark styling of the emails.

When I tested this, I noticed that the values always come from the sites default theme. They don't change when the user selects an optional theme in their preferences.

hnb-ku:
Just as a small follow up to my reply above, also note that we currently don't really set email colors based on the theme anyway except for 

the header background color
the header text color - if the site uses a text logo
2 out 3 of the footer links

Everything else is currently set via the existing site settings.

ZogStriP:
> When I tested this, I noticed that the values always come from the sites default theme. They don't change when the user selects an optional theme in their preferences.

Ok. We can do that part later.

But I still feels that setting these colors in /admin/customize/colors to be much more natural than using site settings. 

We could perhaps change the way the palette is displayed in order to make 2 groups (one for the UI, the other for the emails).

tgxworld:
@hnb-ku Any updates on this?

hnb-ku:
> @hnb-ku Any updates on this? We might want to close this PR if it isn't ready

You're right, let's close this for now and revisit a bit later.

PR #7420, opened 2019-04-23
use description_text instead of description to show correct sanitized…

mrfinch:
… description

This is to show the proper description of sub-categories at category page. Example: https://meta.discourse.org/c/plugin/?_escaped_fragment_ 
I wanted to use html_safe so that links can be included on that page but that comes with XSS risk. Is there any standard safe way to do it here for sub-categories description?


SamSaffron:
This change looks good to me.

PR #7421, opened 2019-04-24
DEV: Clear DistributedCaches between tests

danielwaterworth:
These caches are potentially able to leak information between tests and so should be cleared before each one.

### DIFF
--- a/lib/distributed_cache.rb
+++ b/lib/distributed_cache.rb
@@ -3,6 +3,46 @@
 require 'message_bus/distributed_cache'
 
 class DistributedCache < MessageBus::DistributedCache
+  module Mixin
+    module Test
+      module ClassMethods
+        def caches
+          @caches ||= {}
+        end
+
+        def clear_caches!
+          caches.clear
+        end
+      end
+
+      def self.included(base)
+        base.extend ClassMethods
+      end
+
+      def distributed_cache(name, key, **opts)
+        define_singleton_method(name) do
+          DistributedCache::Mixin.caches[object_id] ||= {}
+          DistributedCache::Mixin.caches[object_id][name] ||= DistributedCache.new(key, **opts)
+        end
+      end
+    end
+
+    module Real
+      def distributed_cache(name, key, **opts)
+        define_singleton_method(name) do
+          @__distributed_caches ||= {}
+          @__distributed_caches[name] ||= DistributedCache.new(key, **opts)
+        end
+      end
+    end
+
+    if Rails.env.test?
+      include Test
+    else
+      include Real
+    end
+  end
+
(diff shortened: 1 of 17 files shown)


SamSaffron:
maybe, I am uneasy though to have all this new redis traffic for every test. also this tracking would have to be optional.

danielwaterworth:
@SamSaffron, How would you feel about mocking the caches during the tests so that they don't interact with redis at all?

SamSaffron:
We have some precedence for "special implementation" if you look at what SiteSettings do in tests.

I would prefer if we just swapped message bus in the context of testing to use an "in memory" transport for distributed cache, and then hit that up for clearing or something along those lines. 

So rather than "stub" implementation, special fast implementation that is really easy to clear is more the direction I am thinking.

lib/distributed_cache.rb
```diff
@@ -3,6 +3,26 @@
...
+      def caches
+        @caches ||= {}
+      end
+
+      def get(key, opts = {})
+        caches[key] ||= new(key, opts)
+      end
+
+      def clear_caches!
+        caches.each_value do |cache|
```
SamSaffron: what is the perf impact of adding this to all tests?

danielwaterworth: ~40 seconds.

danielwaterworth: which is about ~5% for me.

SamSaffron:
Way too expensive, we need this to be pretty much free

lib/distributed_cache.rb
eviltrout: Would `@caches = {}` not do it?

danielwaterworth: @eviltrout, unfortunately not, references to these caches are held in instances variables on classes themselves.

eviltrout: Maybe we could DRY up the DistributedCache memoizing. We could create a mixin `HasDistributedCache` that classes would add, then they'd use a class method `distributed_cache :last_read_only` which creates the memoized cache.

In test mode we'd make it something we could easily clean up. Otherwise we'd memoize the instance.

danielwaterworth: That would work, I'll do that.

danielwaterworth:
@SamSaffron, this new version is comparative in performance or perhaps even slightly faster than without the changes.

danielwaterworth:
The tests that failed here are all ones that erroneously assumed that they weren't the first to run. They fail on master if you run them on their own.

danielwaterworth:
So, it turns out the performance hit here is mainly down to starting each test with a clear cache, causing it to go to the DB.

PR #7422, opened 2019-04-24
FIX: Do not replace text of collapsed ignore posts

khalilovcmded:
## Why?

Right now, for collapsed posts of an ignored user, if the posts are expanded, we replace the text with `Ignored content`. We don't think this should be the case.

GIF

### DIFF
--- a/app/assets/javascripts/discourse/widgets/embedded-post.js.es6
+++ b/app/assets/javascripts/discourse/widgets/embedded-post.js.es6
@@ -29,6 +29,7 @@ export default createWidget("embedded-post", {
   buildKey: attrs => `embedded-post-${attrs.id}`,
 
   html(attrs, state) {
+    attrs.embeddedPost = true;
     return [
       h("div.reply", { attributes: { "data-post-id": attrs.id } }, [
         h("div.row", [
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/embedded-post.js.es6
```diff
@@ -29,6 +29,7 @@ export default createWidget("embedded-post", {
   buildKey: attrs => `embedded-post-${attrs.id}`,
 
   html(attrs, state) {
+    attrs.embeddedPost = true;
```
khalilovcmded: I am adding this `attr` because we want to know in `post-cooked` whether this is an `embedded reply to a post` or `a normal post`.

khalilovcmded: Now thinking about this again, couldn't I just look at the `passed widget` type / class name? Maybe.

khalilovcmded: Would love to get @eviltrout eyes on this 🙇

eviltrout: This is an OK way to do it. The only slightly awry thing is you are changing attrs inside a widget that’s not using them. It would be arguably better to extend attrs when passing it to post cooked and include the attribute there so that it’s only available where it’s used... but really this is fine!

PR #7423, opened 2019-04-24
SECURITY: jquery CVE-2019-11358

jjaffeux:


jjaffeux:
We are applying this a patch for now. I will upgrade to 3.4 after next beta.

PR #7424, opened 2019-04-24
FIX: Do not move hidden post actions.

udan11:
Hidden (staff-only) post actions are whisper posts with no content, that
are later transformed by the client into post actions (discourse-assign
uses this).

### DIFF
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -224,6 +224,7 @@ class PostMover
     @posts ||= begin
       Post.where(topic: @original_topic, id: post_ids)
         .where.not(post_type: Post.types[:small_action])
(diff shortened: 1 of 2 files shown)


app/models/post_mover.rb
```diff
@@ -224,6 +224,7 @@ def posts
     @posts ||= begin
       Post.where(topic: @original_topic, id: post_ids)
         .where.not(post_type: Post.types[:small_action])
+        .where.not(raw: '')
```
eviltrout: This is quite a code smell :(

Again like the other PR I approved it’s not your fault, but we really should figure out a cleaner way to do this in the future.

PR #7425, opened 2019-04-24
FIX: Make select all and below skip small actions.

udan11:


SamSaffron:
I worry though, does this not end up confusing stuff? 

Post
Small action
Post 
Small action
Post

If we now move this we end up having out of context small actions in the original. 

I wonder if we should be explicit about what small actions are skipped (we definitely want whispers,  probably want PM invites) I guess this is just a very tricky problem.

udan11:
This PR is meant to solve an inconsistency rather than fix the user experience (server excluded small actions and empty posts, but client did not).

You are right about small actions going out of context, but I cannot see any good solution to this problem. First of all, there would be the problem of what small actions are going out of context (type and position; all small actions between selected posts, but also the ones immediately before and after). The second problem would be figuring out what to do with those? Delete them, move them or maybe just change their styling?

PR #7427, opened 2019-04-24
UX: Add transliteration support for the Greek language

jtsagata:
The transliteration file is taken from rails-i18n gem


gschlager:
Thanks!

PR #7433, opened 2019-04-24
DEV: Export the default behaviour of a topic-list-item 

romanrizzi:
This separates the default behavior of a `topic-list-item` from the required code to render it alongside a raw template

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-list-item.js.es6
@@ -29,10 +29,152 @@ export function navigateToTopic(topic, href) {
   return false;
 }
 
+export const ListItemDefaults = {
+  tagName: "tr",
+
+  @computed
+  newDotText() {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -29,175 +29,177 @@ export function navigateToTopic(topic, href) {
...
+  }.on("didInsertElement")
+};
 
-    highlight(opts = { isLastViewedTopic: false }) {
-      const $topic = this.$();
-      $topic
-        .addClass("highlighted")
-        .attr("data-islastviewedtopic", opts.isLastViewedTopic);
+export default Ember.Component.extend(
+  bufferedRender(
```
eviltrout: I don't think `$.extend` should be necessary here. You can include multiple mixins in an ember class:

`export default Ember.Component.extend(ListItemDefaults, bufferedRender( ... buffered specific methods ... ), { ... extra instance methods ... })`

PR #7436, opened 2019-04-25
FIX: Only do plain text table detection if outside code

venarius:
Plain text table detection now only happens outside of code fence blocks.

Resolves: https://meta.discourse.org/t/plain-text-table-detection-should-respect-code-fences/115894

### DIFF
--- a/app/assets/javascripts/discourse/components/d-editor.js.es6
+++ b/app/assets/javascripts/discourse/components/d-editor.js.es6
@@ -802,15 +802,6 @@ export default Ember.Component.extend({
     let html = clipboard.getData("text/html");
     let handled = false;
 
-    if (plainText) {
-      plainText = plainText.trim().replace(/\r/g, "");
-      const table = this._extractTable(plainText);
-      if (table) {
-        this.appEvents.trigger("composer:insert-text", table);
-        handled = true;
-      }
-    }
-
     const { pre, lineVal } = this._getSelected(null, { lineVal: true });
     const isInlinePasting = pre.match(/[^\n]$/);
 
@@ -837,6 +828,13 @@ export default Ember.Component.extend({
 
         this.appEvents.trigger("composer:insert-text", markdown);
         handled = true;
+      } else if (plainText) {
+        plainText = plainText.trim().replace(/\r/g, "");
+        const table = this._extractTable(plainText);
+        if (table) {
+          this.appEvents.trigger("composer:insert-text", table);
+          handled = true;
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


ZogStriP:
Nice, is there a way this can be tested?

test/javascripts/lib/pretty-text-test.js.es6
```diff
@@ -1387,6 +1387,13 @@ QUnit.test("enable/disable features", assert => {
   );
 });
 
+QUnit.test("table inside code block", assert => {
+  assert.cooked(
```
ZogStriP: Is that testing your change? Isn't the fix related to _pasting_ a table in a code block?

venarius: I think *real* pasting can't be tested with qunit, no?

venarius: Also the problem here is that is has also worked before. We're just not wasting resources anymore now.

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -837,6 +828,13 @@ export default Ember.Component.extend({
 
         this.appEvents.trigger("composer:insert-text", markdown);
         handled = true;
+      } else if (plainText) {
```
vinothkannans: This is not intended behavior. I pushed the commit in https://github.com/discourse/discourse/commit/dc2ee78f7a667cd0ded8e198282def7ea51797bb.

vinothkannans:
Done as per commit https://github.com/discourse/discourse/commit/dc2ee78f7a667cd0ded8e198282def7ea51797bb.

app/assets/javascripts/discourse/components/d-editor.js.es6
venarius: @vinothkannans can you tell me why this isn't intended behavior? It's always better if I know what exactly I did wrong.

vinothkannans: This `else if` condition will run only if ["HTML pasting"](https://github.com/venarius/discourse/blob/044a15afcb94e9a359fd95cb3ef82672cdf6c515/app/assets/javascripts/discourse/components/d-editor.js.es6#L823-L830) not happen. The intended behavior is table detection should happen even before HTML pasting. Only if pasting is not inside code block and HTML pasting enabled.

venarius: Oh okay, got it! Thanks!

PR #7437, opened 2019-04-25
FIX: prevents exception when search q params is a hash

jjaffeux:


jjaffeux:
An exception was happening for this kind of request for example:

`/search?q[aaa]=1`

tgxworld:
Is the hash `q` params generated from the client side code? I think we should throw an error if `q` params is provided as a hash.

jjaffeux:
> Is the hash `q` params generated from the client side code? I think we should throw an error if `q` params is provided as a hash.

no I dont think so, I just think sometimes people try things 🤷‍♂️ but I can raise InvalidParameters if you prefer

tgxworld:
I think we should raise `InvalidParameters` here since there is no need for us to support the hash version of `q`

jjaffeux:
> I think we should raise `InvalidParameters` here since there is no need for us to support the hash version of `q`

ok will amend to do this then

jjaffeux:
@tgxworld updated

tgxworld:
@jjaffeux Yay thank you for fixing this :+1:

PR #7439, opened 2019-04-25
add no_subcategories query param in url only if its true

mrfinch:
@coding-horror mentioned to me that the “next link” from the latest is built like so https://meta.discourse.org/latest?no_definitions=true&no_subcategories=false&page=2 which is super nasty. So, his suggestion was to remove queryParams to https://meta.discourse.org/latest?page=2

This is related to it. I haven't removed `no_definitions` since we use it as a filter. It is used as below:
```   
 if options[:no_definitions]
      result = result.where('COALESCE(categories.topic_id, 0) <> topics.id')
 end
```
This adds `no_subcategories` param only if it's true otherwise it will be assumed as false anyways.

### DIFF
--- a/app/controllers/list_controller.rb
+++ b/app/controllers/list_controller.rb
@@ -382,7 +382,9 @@ class ListController < ApplicationController
 
     # hacky columns get special handling
     options[:topic_ids] = param_to_integer_list(:topic_ids)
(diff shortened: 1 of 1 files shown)


tgxworld:
Thanks looks OK to me. Thank you for contributing :+1:

PR #7442, opened 2019-04-25
FIX: Fixed uploadPlaceholder regex failing on not escaped characters

venarius:
This allows files with characters that may break the regex to be uploaded.

Resolves: https://meta.discourse.org/t/cannot-upload-file-with-in-name/115985

### DIFF
--- a/test/javascripts/acceptance/composer-test.js.es6
+++ b/test/javascripts/acceptance/composer-test.js.es6
@@ -101,81 +101,98 @@ QUnit.test("Tests the Composer controls", async assert => {
   assert.ok(!exists(".bootbox.modal"), "the confirmation can be cancelled");
 });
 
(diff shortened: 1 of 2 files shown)


test/javascripts/acceptance/composer-test.js.es6
```diff
@@ -101,81 +101,98 @@ QUnit.test("Tests the Composer controls", async assert => {
   assert.ok(!exists(".bootbox.modal"), "the confirmation can be cancelled");
 });
 
-// Temporarily remove to see if this is breaking the test suite
```
gschlager: 7 months later... 😁
@SamSaffron `QUnit.skip` would have been the better choice. 😉

SamSaffron: :)

SamSaffron:

PR #7443, opened 2019-04-25
FEATURE: Show diff of local changes before updating remote theme

pmusaraj:
image

When updating a remote theme, this checks whether there are any local changes, and if so, it displays a diff of the changes to the user, as in screenshot above.  

Changes to SCSS, HTML (JS) and uploaded assets will trigger the warning. Deleted theme assets will not trigger it.

### DIFF
--- a/app/assets/javascripts/admin/models/theme.js.es6
+++ b/app/assets/javascripts/admin/models/theme.js.es6
@@ -1,6 +1,9 @@
 import RestModel from "discourse/models/rest";
 import { default as computed } from "ember-addons/ember-computed-decorators";
 import { popupAjaxError } from "discourse/lib/ajax-error";
+import { ajax } from "discourse/lib/ajax";
+import { escapeExpression } from "discourse/lib/utilities";
+import highlightSyntax from "discourse/lib/highlight-syntax";
 
 const THEME_UPLOAD_VAR = 2;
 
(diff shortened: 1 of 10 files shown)


davidtaylorhq:
This is very cool! Just a couple of things to check (having been bitten by these things myself in the past)

- How does this work if you have some empty files in the repository? Discourse will ignore them when importing, which means the export will be missing them, and that might result in an unintended diff.

- Similarly, what happens if the file has been emptied in the Discourse UI, but still has content in the remote repository?

- What happens if assets have filenames which are different from their variable name in `about.json`? Discourse always exports themes with asset filename matching the variable name, but imports work with arbitrary filenames.

If those things are causing problems, it might be better for us to fix the root causes, rather than patching over it with the diff. Ideally imports and exports should exactly match up exactly.

pmusaraj:
Thanks David, all great points, and they exposed a weakness that I hadn't noticed yesterday, whereby local changes were being compared to the current version of the repo (they now are compared to the last pulled version). 

**empty files in the repo**: this does trigger an unintended diff, and I'm not sure how to fix it. We want the diff to catch files emptied or deleted via the UI. Maybe we can just live with this? 
**file has been emptied in the Discourse UI**: this now works correctly (deleted file in UI is shown in diff modal)
**assets have filenames which are different from their variable name**: yes, this causes an unintended diff. the output is kind of useful: 
```
diff --git a/assets/ppu.png b/assets/ptest2.png
similarity index 100%
rename from assets/ppu.png
rename to assets/ptest2.png
```
I don't think we store the original filename anywhere, but as you note, it might be good to fix this in general so imports/exports match filenames.

SamSaffron:
This looks great to me ... feel free to merge when you are online

PR #7447, opened 2019-04-26
NEW: Added unlisted topics option to advanced search

venarius:
This adds an options to search for unlisted topics in the advanced search.

Screenshot from 2019-04-26 19-43-24

Resolves: https://meta.discourse.org/t/how-administrator-or-moderator-can-search-unlisted-topics/102834

### DIFF
--- a/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
+++ b/app/assets/javascripts/discourse/components/search-advanced-options.js.es6
@@ -28,6 +28,17 @@ const REGEXP_POST_TIME_WHEN = /^(before|after)/gi;
 
 const IN_OPTIONS_MAPPING = { images: "with" };
 
+const DEFAULT_STATUS_OPTIONS = [
+  { name: I18n.t("search.advanced.statuses.open"), value: "open" },
+  { name: I18n.t("search.advanced.statuses.closed"), value: "closed" },
+  { name: I18n.t("search.advanced.statuses.archived"), value: "archived" },
+  { name: I18n.t("search.advanced.statuses.noreplies"), value: "noreplies" },
+  {
+    name: I18n.t("search.advanced.statuses.single_user"),
+    value: "single_user"
+  }
+];
+
 export default Ember.Component.extend({
   classNames: ["search-advanced-options"],
 
@@ -45,17 +56,6 @@ export default Ember.Component.extend({
     { name: I18n.t("search.advanced.filters.unpinned"), value: "unpinned" },
     { name: I18n.t("search.advanced.filters.wiki"), value: "wiki" },
     { name: I18n.t("search.advanced.filters.images"), value: "images" }
-  ],
-
-  statusOptions: [
-    { name: I18n.t("search.advanced.statuses.open"), value: "open" },
-    { name: I18n.t("search.advanced.statuses.closed"), value: "closed" },
-    { name: I18n.t("search.advanced.statuses.archived"), value: "archived" },
-    { name: I18n.t("search.advanced.statuses.noreplies"), value: "noreplies" },
-    {
-      name: I18n.t("search.advanced.statuses.single_user"),
-      value: "single_user"
-    }
   ],
 
   postTimeOptions: [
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/components/search-advanced-options.js.es6
```diff
@@ -106,6 +106,15 @@ export default Ember.Component.extend({
         ? this.inOptionsForUsers.concat(this.inOptionsForAll)
         : this.inOptionsForAll
     });
+
+    if (this.currentUser.get("staff") && this.statusOptions.length === 5) {
```
venarius: I needed to add `this.statusOptions.length === 5` here because it seemed that `init` got called about 3-4 times every time I reloaded the site, which resulted in 3-4 "are unlisted" entries in the menu. Any way to achieve that on a more cleaner way than this *dirty* hack?

lib/search.rb
```diff
@@ -737,7 +741,7 @@ def posts_query(limit, opts = nil)
 
     is_topic_search = @search_context.present? && @search_context.is_a?(Topic)
 
-    posts = posts.where("topics.visible") unless is_topic_search
+    posts = posts.where("topics.visible") unless is_topic_search || @guardian.is_staff?
```
SamSaffron: unless with an || makes my brain tingle 

```suggestion
    posts = posts.where("topics.visible") if !is_topic_search && @guardian.is_staff?
```

lib/search.rb
```diff
@@ -280,6 +280,10 @@ def self.advanced_filters
     posts.where("topics.participant_count = 1")
   end
 
+  advanced_filter(/^status:unlisted$/) do |posts|
```
SamSaffron: I think for non staff "state:unlisted" should probably just return nothing? Mixed on this, technically an error is probably the best thing to do but that is too complex, I guess we can go with this behavior for now

spec/components/search_spec.rb
```diff
@@ -1129,6 +1129,24 @@ def search
...
       end
 
+      it "can search for unlisted topics as staff" do
+        topic1 = Fabricate(:topic, visible: false)
+        Fabricate(:post, raw: 'Testing post', topic: topic1)
+        topic2 = Fabricate(:topic)
+        Fabricate(:post, raw: 'Testing post', topic: topic2)
+
+        results = Search.execute('Testing post status:unlisted', guardian: Guardian.new(Fabricate(:moderator)))
+        expect(results.posts.length).to eq(1)
```
tgxworld: We need to assert for the post id here since asserting for the length only doesn't tell us if the post we expect is returned

spec/components/search_spec.rb
```diff
@@ -1129,6 +1129,24 @@ def search
...
+      end
+
+      it "unlisted topics can't be found using search for non-staff" do
+        topic = Fabricate(:topic, visible: false)
+        Fabricate(:post, raw: 'Testing post', topic: topic)
+
+        results = Search.execute('Testing post', guardian: Guardian.new(Fabricate(:user)))
+        expect(results.posts.length).to eq(0)
```
tgxworld: Ditto https://github.com/discourse/discourse/pull/7447/files#r279227566

tgxworld:
@venarius Can you rename the commit to use `FEATURE:` instead of `NEW:` Thanks!

lib/search.rb
tgxworld: Yea I think we should just return and do nothing here. Less work to do on the DB side.

SamSaffron: In that case we should return `posts.where("1=0")` if user is not staff.

lib/search.rb
venarius: I think it should also be `!guardian.is_staff?` here so I will commit that.

SamSaffron:
Looks good to me! thanks!

vinothkannans:
@venarius Sorry. I reverted this PR since it failed the build.

venarius:
Okay, I will have a look. Thanks for telling!

geckolinux:
Hi, I would still find it very useful to be able to search for ***Unlisted*** as well as ***Deleted posts***.

PR #7450, opened 2019-04-27
FEATURE: adds support for dates in polls

jjaffeux:


plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -10,7 +10,13 @@ import round from "discourse/lib/round";
 import { relativeAge } from "discourse/lib/formatter";
 
 function optionHtml(option) {
-  return new RawHtml({ html: `<span>${option.html}</span>` });
+  const $node = $(`<span>${option.html}</span>`);
+
+  $node.find(".discourse-local-date").each((index, elem) => {
```
tgxworld:
```suggestion
  $node.find(".discourse-local-date").each((_index, elem) => {
```
I'm surprised prettier doesn't catch this xD

PR #7451, opened 2019-04-29
PERF: Call web hook events in a new thread

vinothkannans:


SamSaffron:
Sorry but I am rejecting this, we can't be creating threads like this its just too risky.

SamSaffron:
oops clicked wrong button (did a revert)

PR #7452, opened 2019-04-29
DEV: Only install danger on travis.

tgxworld:


tgxworld:
Let's see if the build on Travis passes first.

PR #7453, opened 2019-04-29
Fix category logo upload error. Add tags.

imMMX:
1. Due to the `category` object has no attribute called `logo_url`. Fix it with set category.uploaded_logo_id equal upload.id.
2. Some origin posts have tags. To fix that has to add a map value `tags`.


ZogStriP:
Thanks @imMMX 👍

imMMX:
you're welcome.😃

PR #7454, opened 2019-04-29
 FIX: Properly associate user_profiles background urls via upload id.

tgxworld:


tgxworld:
@gschlager  @davidtaylorhq  I would like to extra pairs of eyes on this PR since it involves the migration of  data from one column to another. It'll be great if I can get your reviews on this PR. Thank you in advance.

db/migrate/20190426011148_add_upload_foreign_keys_to_user_profiles.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class AddUploadForeignKeysToUserProfiles < ActiveRecord::Migration[5.2]
+  def up
+    %i{profile_background card_background}.each do |column|
+      Migration::ColumnDropper.mark_readonly(:user_profiles, column)
+    end
+
+    add_column :user_profiles, :profile_background_upload_id, :integer, null: true
+    add_column :user_profiles, :card_background_upload_id, :integer, null: true
+    add_foreign_key :user_profiles, :uploads, column: :profile_background_upload_id
```
davidtaylorhq: The default behavior of `add_foreign_key` is to use `on_delete: :restrict`. That makes it impossible to delete the `upload` row until the foreign key is removed. I think maybe we would prefer `on_delete: :nullify`. What do you think?

spec/serializers/user_serializer_spec.rb
```diff
@@ -69,23 +66,15 @@
...
-        expect(json[:card_background]).to eq 'http://card.com'
-      end
-    end
-
-    context "with filled out profile background" do
+    context "with filled out backgrounds" do
       before do
-        user.user_profile.profile_background = 'http://background.com'
+        user.user_profile.upload_card_background(upload)
+        user.user_profile.upload_profile_background(upload)
```
gschlager: It might be worth using two different uploads in order to make sure that the URLs aren't mixed up.

db/migrate/20190426011148_add_upload_foreign_keys_to_user_profiles.rb
gschlager: I guess it depends. The default behavior is perfectly fine if you want to make sure that you don't accidentally delete the upload. I think it's fine as it is.

spec/services/user_updater_spec.rb
```diff
@@ -145,20 +145,25 @@
...
+        digest_after_minutes: "45",
+        new_topic_duration_minutes: 100,
+        auto_track_topics_after_msecs: 101,
+        notification_level_when_replying: 3,
+        email_in_reply_to: false,
+        date_of_birth: date_of_birth,
+        theme_ids: [theme.id],
+        allow_private_messages: false,
+        card_background_upload_url: upload.url,
+        profile_background_upload_url: upload.url
```
gschlager: Same thing here. I'd use 2 different uploads to prevent mixups.

app/models/user_profile.rb
```diff
@@ -150,21 +149,25 @@ def website_domain_validator
...
+#  profile_background_upload_id :integer
+#  card_background_upload_id    :integer
 #
 # Indexes
 #
-#  index_user_profiles_on_bio_cooked_version  (bio_cooked_version)
-#  index_user_profiles_on_card_background     (card_background)
-#  index_user_profiles_on_profile_background  (profile_background)
+#  index_user_profiles_on_bio_cooked_version            (bio_cooked_version)
+#  index_user_profiles_on_card_background               (card_background)
```
SamSaffron: This index is going no? and the column?

SamSaffron:
As long as you take maintaining any weird with the new foreign key constraint I guess I am ok, go ahead and merge.

db/migrate/20190426011148_add_upload_foreign_keys_to_user_profiles.rb
davidtaylorhq: I don't feel too strongly either way, would be happy for it to stay as `restrict`. However, it might be annoying if you can't `Upload.find(123).destroy` any more 🤔

tgxworld: Yea it should be restrict here. We shouldn't be able to delete the upload if it is still tied to a profile or card background.

app/models/user_profile.rb
tgxworld: @SamSaffron We need to drop the columns in a post migration but I'm playing it safe first by not dropping the columns since we're migrating data.

tgxworld:
> And I love that you added specs for the migrations!

@gschlager  I actually wrote specs because I wasn't sure if the queries I wrote for my migration was going to be correct or not :grin: I'm glad I did actually because it helped me to catch a few edge cases I would have missed have I not written the specs. 

Thanks for reviewing this @gschlager @davidtaylorhq and @SamSaffron  :+1:

PR #7458, opened 2019-04-29
DEV: Spec clarification example

danielwaterworth:
@tgxworld, I thought it would be useful to show you what I mean w.r.t lexical ordered lets and test clarity. Here's a short example. I find it much easier to read these tests after these changes than before. Each commit is small and targetted so that you can follow steps to get to the end result.

### DIFF
--- a/spec/models/topic_posters_summary_spec.rb
+++ b/spec/models/topic_posters_summary_spec.rb
@@ -39,12 +39,6 @@ describe TopicPostersSummary do
 
       let!(:summary) { described_class.new(topic).summary }
 
-      before do
-        topic.last_poster = topic_creator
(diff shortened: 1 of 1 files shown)


tgxworld:
I agree with the refactoring done here but I don't think the initial test was that confusing in anyway. I think the problem here is that clarity of tests is subjective. The original author may think that his test was clear while others might not. After thinking about this for a couple of days, I really don't mind whatever new way `fab!` will require us to write out tests. As long as it makes the test faster, I'm happy to adopt whatever order it requires us to write the `fab!`. Thank you for taking the time to illustrate your point with this example, I greatly appreciate it :+1:

PR #7459, opened 2019-04-29
FIX: Mark invited admins as 'approved'

davidtaylorhq:
This prevents invited admins appearing as no-op reviewables in the queue when `invite_only` or `require_approval` is enabled.


ZogStriP:
I think that's worth a test 😉

PR #7461, opened 2019-04-29
FIX: cache_critical_dns was never caching pg replica

Supermathie:
* it's DISCOURSE_DB_REPLICA_HOST not DISCOURSE_DB_BACKUP_HOST


tgxworld:
:+1:

PR #7462, opened 2019-04-29
PERF: Move `where` clause up to speed up CalculateAvgTime daily job

xfalcox:
Cuts down affected posts earlier in the query, so the generated plan
deals with less rows, and runs faster.

https://meta.discourse.org/t/post-calculate-avg-time-taking-up-a-long-time/49750/13?u=falco


SamSaffron:
Sure, lets try this, thanks!

PR #7464, opened 2019-04-30
FEATURE: English locale with international date formats

gschlager:
This changes the date and time formats of the English (`en`) locale into international formats. The existing formats are now in the `en_US` locale which is also the new default locale.

`en_US` uses `en` as fallback locale. This PR also introduces a couple of changes that will allow other language dialects to use an existing locale as fallback.

It also adds the new `title_remove_extraneous_space` site setting that is default off except for `en` and `en_US`.

The `default locale` site setting, the locale of users and translation overrides get migrated from `en` to `en_US`.

cf. https://meta.discourse.org/t/british-english-locale/64604/13


Dangerfile
```diff
@@ -13,8 +13,8 @@ This PR doesn't match our required code formatting standards, as enforced by pre
   })
 end
 
-locales_changes = git.modified_files.grep(/config\/locales/)
-has_non_en_locales_changes = locales_changes.grep_v(/config\/locales\/(client|server)\.en\.yml/).any?
+locales_changes = git.modified_files.grep(%r{config/locales})
+has_non_en_locales_changes = locales_changes.grep_v(%r{config/locales/(?:client|server)\.(?:en|en_US)\.yml}).any?
```
gschlager: We don't want `en_US` to be on Transifex, so it should get the same protection as `en`.

lib/js_locale_helper.rb
```diff
@@ -155,7 +154,8 @@ def self.output_locale(locale)
   end
 
   MOMENT_LOCALE_MAPPING ||= {
-    "hy" => "hy-am"
+    "hy" => "hy-am",
+    "en" => "en-gb"
```
gschlager: Reuse the British moment.js locale for English.

PR #7476, opened 2019-05-03
FIX: Do not send duplicate alerts for the same post

pmusaraj:
This addresses a small part of duplicate alerts on the same post ([initial discussion on meta](https://meta.discourse.org/t/new-ios-mobile-app-beta-available-for-testing/115912/33?u=pmusaraj)).

Case scenario

- userA replies to a topic owned by userB
- userB receives an alert (push notification)
- userA revises her post and adds a mention of userB or quotes userB

Currently, userB receives a second alert upon being mentioned or quoted. 
This PR suppresses the second alert on the same post.

### DIFF
--- a/app/services/post_alerter.rb
+++ b/app/services/post_alerter.rb
@@ -319,32 +319,44 @@ class PostAlerter
       ).exists?
     end
 
-    # Don't notify the same user about the same type of notification on the same post
-    existing_notification = user.notifications
+    existing_notifications = user.notifications
       .order("notifications.id DESC")
-      .find_by(
+      .where(
         topic_id: post.topic_id,
-        post_number: post.post_number,
-        notification_type: type
-      )
+        post_number: post.post_number
+      ).limit(10)
 
-    return if existing_notification && !should_notify_previous?(user, existing_notification, opts)
+    existing_notification_of_same_type = existing_notifications.find { |n| n.notification_type == type }
+
+    # existing_notification = user.notifications
+    #   .order("notifications.id DESC")
+    #   .find_by(
+    #     topic_id: post.topic_id,
+    #     post_number: post.post_number
+    #   )
+
+    # # Don't notify the same user about the same type of notification on the same post
+    # existing_notification_of_same_type = existing_notification if existing_notification&.notification_type == type
+
+    puts existing_notification_of_same_type.inspect
+
+    return if existing_notification_of_same_type && !should_notify_previous?(user, existing_notification_of_same_type, opts)
 
     notification_data = {}
 
     if is_liked
-      if existing_notification &&
-        existing_notification.created_at > 1.day.ago &&
+      if existing_notification_of_same_type &&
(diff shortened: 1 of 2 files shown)


app/services/post_alerter.rb
```diff
@@ -356,6 +356,15 @@ def create_notification(user, type, post, opts = {})
       end
     end
 
+    # Don't alert same user about the same post
+    # i.e. skip duplicate alerts on a revised post (when editing post triggers a new type of notification like a mention or a quote)
+    existing_alert = user.notifications
```
eviltrout: It seems to me you could query this once, and use it for the other check too, since it's a subset of this one. (You could use ruby code to check for the type in the resulting array.)

app/services/post_alerter.rb
```diff
@@ -319,32 +319,44 @@ def create_notification(user, type, post, opts = {})
...
         topic_id: post.topic_id,
-        post_number: post.post_number,
-        notification_type: type
-      )
+        post_number: post.post_number
+      ).limit(10)
+
+    existing_notification_of_same_type = existing_notifications.find { |n| n.notification_type == type }
+
+    # existing_notification = user.notifications
```
ZogStriP: Why all these comments?

app/services/post_alerter.rb
```diff
@@ -319,32 +319,44 @@ def create_notification(user, type, post, opts = {})
...
+    #   .order("notifications.id DESC")
+    #   .find_by(
+    #     topic_id: post.topic_id,
+    #     post_number: post.post_number
+    #   )
+
+    # # Don't notify the same user about the same type of notification on the same post
+    # existing_notification_of_same_type = existing_notification if existing_notification&.notification_type == type
+
+    puts existing_notification_of_same_type.inspect
```
ZogStriP: Probably want to remove that?

app/services/post_alerter.rb
pmusaraj: Oh my, I didn't clean up properly when I pushed this last Friday.

app/services/post_alerter.rb
```diff
@@ -319,32 +319,33 @@ def create_notification(user, type, post, opts = {})
...
+      .where(
         topic_id: post.topic_id,
-        post_number: post.post_number,
-        notification_type: type
-      )
+        post_number: post.post_number
+      ).limit(10)
+
+    # Don't notify the same user about the same type of notification on the same post
+    existing_notification_of_same_type = existing_notifications.find { |n| n.notification_type == type }
```
ZogStriP: I don't understand how filtering in Ruby is any different than filtering in SQL?

pmusaraj: It's one query instead of two, so it should be more performant?

ZogStriP: My bad, I didn't see you were reusing this at the end of the method.

pmusaraj:
@eviltrout is this good to merge?

PR #7478, opened 2019-05-03
FIX: Use all flag types in moderators activity report

scossar:
The Flags Reviewed section of the moderators activity report is ignoring the `:notify_moderators` flag type. Changing `PostActionType.flag_types_without_custom` to `PostActionType.flag_types` fixes that.

### DIFF
--- a/app/models/reports/moderators_activity.rb
+++ b/app/models/reports/moderators_activity.rb
@@ -69,7 +69,7 @@ Report.add_report("moderators_activity") do |report|
       SELECT agreed_by_id,
       disagreed_by_id
       FROM post_actions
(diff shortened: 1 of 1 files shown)


eviltrout:
There is potentially a good reason to do this - notify moderator is something like a PM and doesn't necessarily mean something is wrong. For example, maybe there's a typo or invalid category?

Is there a relevant discussion for this?

ZogStriP:
> Is there a relevant discussion for this?

Any news @scossar?

scossar:
>Is there a relevant discussion for this?

It was brought up by one of our customers in a private PM. They wanted to know why their agreed flags were not being included in the report.

ZogStriP:
Let's wait for the rule of 3 to hit then ;)

PR #7479, opened 2019-05-06
UX: Switch mobile header back for 5s when notification comes in.

tgxworld:


tgxworld:
@hnb-ku Can you review this for me? Thank you!

PR #7480, opened 2019-05-06
REFACTOR: Set input and textarea font size to 16px on iOS

pmusaraj:
This fixes intermittent zoom-on-input issues on iOS by: 

- adding an `ios-device` class to the html element
- adding a new font size for iOS inputs in SCSS variables and applying that font size to text inputs and textareas for mobile iOS only
- cleaning up previous JS `touchstart` events that disable zooming


SamSaffron:
I like this change a lot, feel free to merge when ready

PR #7481, opened 2019-05-06
FEATURE: use full screen login by default for social login methods

techAPJ:
https://review.discourse.org/t/fix-use-full-screen-login-for-new-topic-route-7467/3170/3?u=techapj

### DIFF
--- a/app/assets/javascripts/discourse/models/login-method.js.es6
+++ b/app/assets/javascripts/discourse/models/login-method.js.es6
@@ -24,7 +24,7 @@ const LoginMethod = Ember.Object.extend({
     );
   },
 
-  doLogin(reconnect = false) {
+  doLogin({ reconnect = false, fullScreenLogin = true } = {}) {
     const name = this.get("name");
     const customLogin = this.get("customLogin");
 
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/discourse/models/login-method.js.es6
```diff
@@ -90,15 +90,6 @@ export function findAll(siteSettings, capabilities, isMobileDevice) {
     methods.pushObject(LoginMethod.create(provider));
   });
 
-  // On Mobile, Android or iOS always go with full screen
```
SamSaffron: yeah we still need to make this decision somewhere, but it can be in the click handler instead.

techAPJ: good catch, let me verify that we don't change existing behaviour here.

techAPJ: ... and done!

app/assets/javascripts/discourse/models/login-method.js.es6
```diff
@@ -79,7 +79,7 @@ const LoginMethod = Ember.Object.extend({
 
 let methods;
 
-export function findAll(siteSettings, capabilities, isMobileDevice) {
```
techAPJ: We do not need the extra parameters anymore. This simplifies the code at all the places where `findAll` is being used.

SamSaffron:
The change looks great to me, feel free to merge. 

Once merged test a variety of sites on both mobile and desktop. We don't want to regress here and the changes run quite deep. Great changes though!

PR #7483, opened 2019-05-06
FEATURE: Use Beacon API for tracking clicks.

nbianca:


SamSaffron:
I really like this and want to try it out! Thanks @nbianca !!!

PR #7484, opened 2019-05-06
Revert "FIX: When mutating a string to build a diff. Dup it first"

romanrizzi:
Reverts discourse/discourse#7482

This change should have been applied to the `frozen_string_literal` branch.

### DIFF
--- a/lib/discourse_diff.rb
+++ b/lib/discourse_diff.rb
@@ -176,25 +176,24 @@ class DiscourseDiff
   end
 
   def add_class_or_wrap_in_tags(html_or_text, klass)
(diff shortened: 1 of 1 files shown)


SamSaffron:
this fix itself does look a bit awkward, this file was by a mile the most complex one to convert to froze_string_literal: true, so much mutating of inputs going on.

romanrizzi:
That is the only method in that class that mutates a string. Since `String#insert` is really handy, I thought using `dup` was the best option here.

I also applied this commit against `frozen_string_literal` but looks like it dissapeared from the history after a rebase. If this commit hits that branch, revisions will no longer on dev.

PR #7485, opened 2019-05-06
Time variable used to be created with users timezone

nahrivera7:
When I ran this test I had a problem because my TimeZone is UTC -5, so I was creating search_log on the correct day but search_log2 in the next day. The test expected to have two logs on the same day but I only had one. I changed the method to create a new time variable because it has to be UTC in order to pass the test.

Another user also had the same problem:
https://meta.discourse.org/t/beginners-guide-to-install-discourse-on-macos-for-development/15772/208


eviltrout:
This seems reasonable to me, thanks!

PR #7487, opened 2019-05-07
UX: Fix like and flag button-count alignment on mobile

hnb-ku:
The markup for the like button changed and this led to some styles not applying which results in alignment issues like this

This PR fixes these issues


SamSaffron:
excellent, this is looking great!

PR #7488, opened 2019-05-07
DEV: Prefabricate more things in topic_view_spec.rb

danielwaterworth:
This reduces run time from 16s to 11s on my computer.


eviltrout:
:+1:

PR #7489, opened 2019-05-07
DEV: Prefabricate more things in guardian_spec.rb

danielwaterworth:
This reduces runtime from 13s to 12s on my computer.


eviltrout:
Thanks again!

PR #7491, opened 2019-05-07
 DEV: Prefabricate more things in post_mover.rb

danielwaterworth:
These changes reduce the runtime from 18s to 14s on my computer.


eviltrout:
Looks great thanks!

PR #7494, opened 2019-05-07
FEATURE: Preserve notifications levels when splitting topics.

nbianca:


app/models/post_mover.rb
```diff
@@ -245,12 +245,20 @@ def update_last_post_stats
...
-    TopicUser.change(
-      destination_topic.user,
-      destination_topic.id,
-      notification_level: TopicUser.notification_levels[:watching],
-      notifications_reason_id: TopicUser.notification_reasons[:created_topic]
-    )
+    if @original_topic.archetype == Archetype.private_message
+      notification_levels = TopicUser.where(topic_id: @original_topic.id, user_id: posts.pluck(:user_id)).pluck(:user_id, :notification_level).to_h
+    else
+      notification_levels = posts.pluck(:user_id).map { |user_id| [user_id, TopicUser.notification_levels[:watching]] }.to_h
```
SamSaffron: minor but I think this makes more sense `posts.pluck('distinct user_id')`

SamSaffron: I don't think this is the proper intent, we don't want to force watching here:

1. If you "became the OP" on the new topic you should watch
2. Otherwise, look in old topic_user and carry across the tracking state (you may actually end up carrying across a mute for example)

nbianca: Using `distinct` here is tricky because there is also the ORDER BY clause.

app/models/post_mover.rb
```diff
@@ -245,12 +245,24 @@ def update_last_post_stats
...
+    else
+      notification_levels = [[@destination_topic.user_id, TopicUser.notification_levels[:watching]]]
+    end
+
+    notification_levels.each do |user_id, notification_level|
+      TopicUser.change(
+        user_id,
+        @destination_topic.id,
+        notification_level: notification_level,
+        notifications_reason_id: TopicUser.notification_reasons[:created_topic]
```
SamSaffron: This is real close, the reason is only "created_topic" for the new OP, other users need a different reason.

SamSaffron:
Cool I think this looks good!

PR #7495, opened 2019-05-07
FIX: lighten instead of adding transparency to consolidated pageviews

jjaffeux:


ZogStriP:
I'm only slightly concerned that the Dark mode is supposed to be "dull" and these colors might be too light? (haven't tested TBH)

jjaffeux:
> I'm slightly concerned that the Dark mode is supposed to be "dull" and these colors might be too bright? (haven't tested TBH)

I think it's fine like this.

I don't think the goal of dark mode is to be dull, it's mostly to not have a gigantic white square lightning your face  :p and keeping some kind of color coherence meanwhile.

If I lower it I'm more concerned that it will be hard to distinguish them.

jjaffeux:
I'm merging this, we can adjust later if needed

PR #7496, opened 2019-05-07
FIX: simpler w function check

jjaffeux:


vendor/assets/javascripts/i18n-patches.js
```diff
@@ -13,7 +13,7 @@
       return oldI18ntoHumanSize.apply(this, [number, options]);
     };
 
-    if (Ember.typeOf(String.prototype.w) === "function") {
+    if ("w" in String.prototype) {
```
ZogStriP: Is that syntax supported in all the browsers we support?

jjaffeux: Yes it is

PR #7501, opened 2019-05-07
PERF: Skip compressing locales for faster rebuilds

pmusaraj:
Several changes to asset precompilation to improve rebuild speeds: 

1. Use uglifier without the `-c (--compress)` option

This results in slightly larger file sizes (300k vs 296k for application.js for example) but it has a significant improvement in compilation speed (60s faster in my tests). Disabling compression is also mentioned in https://github.com/mishoo/UglifyJS2#uglify-fast-minify-mode as a way to speed up Uglify by 3-4 times. 

2. Skipping uglifier and brotli for non-default locales

Non-default locales take up a very large chunk of precompilation, skipping uglifier and brotli for them saves 1-2 minutes per build. These files are still gzipped, so their size now averages 120k per file (vs 80k with uglifier and brotli). 

### Performance tests

`time ./launcher rebuild app` on a fresh site on a basic Digital Ocean instance, no plugins
Benchmark (tests-passed): **7m25s**
This PR: **4m20s**

`time ./launcher bootstrap app` on an existing Discourse instance with a few plugins
Benchmark (current tests-passed): **12m0s**
This PR: **9m44s**

----

The PR also updates the Sprockets JS compression configuration to enable harmony support in the Uglifier gem.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -12,7 +12,8 @@ Discourse::Application.configure do
   # Disable Rails's static asset server (Apache or nginx will already do this)
   config.public_file_server.enabled = GlobalSetting.serve_static_assets || false
 
-  config.assets.js_compressor = :uglifier
+  require 'uglifier'
+  config.assets.js_compressor = Uglifier.new(harmony: true)
 
   # stuff should be pre-compiled
   config.assets.compile = false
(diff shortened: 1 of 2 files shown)


coding-horror:
@SamSaffron does this look good for merge?

SamSaffron:
Only piece I am concerned about here is giving up on compress, cause that tiny bit of saving does mount up over millions of requests. I don't want to really give it up quite yet, I will merge this in and turn compress back on. Let's watch this for a few weeks and then decide if we want to further optimise this.

PR #7503, opened 2019-05-08
FEATURE: Send notification when member was accepted to group.

nbianca:


app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -10,13 +10,14 @@ import {
...
 import { setTransientHeader } from "discourse/lib/ajax";
-import { userPath } from "discourse/lib/url";
+import { groupPath, userPath } from "discourse/lib/url";
 import { iconNode } from "discourse-common/lib/icon-library";
 
 const LIKED_TYPE = 5;
 const INVITED_TYPE = 8;
 const GROUP_SUMMARY_TYPE = 16;
 export const LIKED_CONSOLIDATED_TYPE = 19;
+const GROUP_INVITE = 21;
```
tgxworld: @jjaffeux showed me a really good pattern to use that will allow us to not have to hardcode server side values on the client side. Can you try to adopt this pattern https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/concerns/category-search-priorities.js.es6.erb? 

Note that adding the `.erb` extension prevents the editor from identifying the file as `js` or `es6` so I usually put the `erb` files under `concerns`.

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -111,6 +116,9 @@ createWidget("notification-item", {
       const count = data.inbox_count;
       const group_name = data.group_name;
       return I18n.t(scope, { count, group_name });
+    } else if (notificationType === GROUP_INVITE) {
+      const group_name = data.group_name;
```
tgxworld:
```suggestion
      const groupName = data.group_name;
```

app/models/group.rb
```diff
@@ -506,9 +506,20 @@ def usernames
 
   PUBLISH_CATEGORIES_LIMIT = 10
 
-  def add(user)
+  def add(user, opts)
```
tgxworld:
```suggestion
  def add(user, opts = {})
```

app/models/group.rb
```diff
@@ -506,9 +506,20 @@ def usernames
...
     self.users.push(user) unless self.users.include?(user)
 
+    if opts && opts[:notify]
```
tgxworld:
```suggestion
    if opts[:notify]
```

tgxworld:
@nbianca We'll need to add some client and server side specs for the changes here.

app/assets/javascripts/discourse/components/concerns/notification_types.js.es6.erb
```diff
@@ -0,0 +1,3 @@
+<% Notification.types.each do |key, value| %>
```
tgxworld: minor one but the filename should be `notification-types.js.es6.erb`

app/models/group.rb
```diff
@@ -506,9 +506,20 @@ def usernames
 
   PUBLISH_CATEGORIES_LIMIT = 10
 
-  def add(user)
+  def add(user, opts = {})
```
ZogStriP: I would not rush to adding an `opts` here and just add a new `notify` parameter with a default value.

```suggestion
  def add(user, notify: false)
```

PR #7508, opened 2019-05-08
DEV: Skip post validation by default when fabricating posts

danielwaterworth:
This speeds up the tests by around 10%


SamSaffron:
wow, what a nice change! :+1:

SamSaffron:
I wonder, should we get rid of "post_with_validation" and just use `create_post` if we need a post, cause that goes through post creator which is the best bet if you want an integration test on posts? Just brainstorming not attached to anything here.

PR #7510, opened 2019-05-08
FIX: Do not show self edits in Post Edits report.

nbianca:


app/models/reports/post_edits.rb
```diff
@@ -83,9 +83,12 @@
     sql += <<~SQL
     JOIN topics t
     ON t.id = p.topic_id
-    WHERE t.category_id = ? OR t.category_id IN (SELECT id FROM categories WHERE categories.parent_category_id = ?)
+    WHERE p.user_id != editor_id AND t.category_id = ? OR t.category_id IN (SELECT id FROM categories WHERE categories.parent_category_id = ?)
     SQL
+  else
+    sql += "WHERE p.user_id != editor_id"
```
tgxworld: Is it possible to add the `WHERE` clause directly to the `sql` variable instead of having to specify it twice?

nbianca: That conditional JOIN makes things really weird here. I cannot think of a better way of doing this.

tgxworld: O right. Ah feel free to merge then :)

PR #7514, opened 2019-05-09
REFACTOR: login-controller

jjaffeux:


app/assets/javascripts/discourse/controllers/login.js.es6
```diff
@@ -114,57 +106,62 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
                 showLoginButtons: false,
                 backupEnabled: result.backup_enabled,
                 showSecondFactor: true
               });
 
-              Ember.run.next(() => {
-                $("#second-factor input").focus();
-              });
+              Ember.run.schedule("afterRender", () =>
+                document.getElementById("second-factor input").focus()
```
eviltrout: This seems odd to me. Did you want a querySelector here? Because input is not part of the id.

jjaffeux: yes indeed

PR #7516, opened 2019-05-09
REFACTOR: composer controller

jjaffeux:


app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -719,15 +715,15 @@ export default Ember.Controller.extend({
         this.close();
 
         const currentUser = Discourse.User.current();
```
eviltrout: While you're in there, is this not available as `this.currentUser`?

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -1103,7 +1095,7 @@ export default Ember.Controller.extend({
 
   @computed("model.action")
   canEdit(action) {
-    return action === "edit" && Discourse.User.current().get("can_edit");
+    return action === "edit" && Discourse.User.current().can_edit;
```
eviltrout: Again I wonder why `this.currentUser` is not available here.

app/assets/javascripts/discourse/controllers/composer.js.es6
jjaffeux: Oh yes good catch, I will add this to my list of things to think of when refactoring a file :+1:

PR #7517, opened 2019-05-09
FEATURE: Add attachment support to XenForo importer

elepedus:
If `ATTACHMENT_DIR` is provided, importer will scan each imported post
for `[GALLERY]` and `[ATTACH]` tags, attempt to import the referenced files
as Discourse uploads and replace the tags with Discourse markup.

References to files which cannot be imported are stripped.

NOTE: This only imports attachments which are referenced in imported
posts. Any XenForo media or files which are not referenced in any post
using `[ATTACH]` or `[GALLERY]` tags will not be imported. The goal is to
ensure that we don't have posts with missing images and unsightly
markup, NOT to ensure that all attachments are migrated.

### DIFF
--- a/script/import_scripts/xenforo.rb
+++ b/script/import_scripts/xenforo.rb
@@ -9,10 +9,10 @@ class ImportScripts::XenForo < ImportScripts::Base
   XENFORO_DB = "xenforo_db"
   TABLE_PREFIX = "xf_"
   BATCH_SIZE = 1000
+  ATTACHMENT_DIR = '/tmp/attachments'
 
   def initialize
     super
-
     @client = Mysql2::Client.new(
       host: "localhost",
       username: "root",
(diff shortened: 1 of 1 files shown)


elepedus:
I just found some issues with my implementation, and I'm working on fixes. Should I close this pull request and create a new one with the fixed version?

script/import_scripts/xenforo.rb
```diff
@@ -305,9 +305,88 @@ def process_xenforo_post(raw, import_id)
     s.gsub!(/\[color=[#a-z0-9]+\]/i, "")
     s.gsub!(/\[\/color\]/i, "")
 
+    if ATTACHMENT_DIR
```
SamSaffron: at a minimum here we should check if the directory exists I think

PR #7518, opened 2019-05-09
FIX: Allow access to local uploads even when store is external

davidtaylorhq:
This is necessary to allow access to local uploads while migrating to external storage


SamSaffron:
I think this is in the "its complicated" bucket, if we are allowing stuff through then we are going to have to do a "proxy file with a hijack" if the upload happens to be on s3 after all, no?  

I guess if we are explicit and intentional about this being "is file in transient state" it could work. 

But why not just keep the local store until every file is moved to s3 and only flip the site setting then? Not sure.

PR #7519, opened 2019-05-09
DEV: Allow setup of DB with `rake db:schema:load`.

tgxworld:


tgxworld:
Not sure what is wrong with Travis... but this is tricky so I'm going to differ for now.

joallard:
That would've been pretty cool, considering I'm getting errors from one bad migration in the bunch

PR #7521, opened 2019-05-10
FEATURE: opt-in guidance on topics for users without access

majakomel:
https://meta.discourse.org/t/show-group-opt-in-guidance-on-topics-you-have-no-access-to/110195


app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -905,6 +905,43 @@ export default Ember.Controller.extend(bufferedProperty("model"), {
...
+                }
+              );
+            })
+            .catch(popupAjaxError);
+        } else {
+          return ajax(`/groups/${groupId}/members`, {
+            type: "PUT",
+            data: { user_id: this.get("currentUser.id") }
+          })
+            .then(() => {
```
jjaffeux:
```
.then() => window.location.reload())
```

app/assets/javascripts/discourse/components/topic-join-group-notice.js.es6
```diff
@@ -0,0 +1,22 @@
+import { default as computed } from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  classNames: ["topic-notice"],
+
+  @computed(
```
jjaffeux: I think with the removal of getters we can write something cleaner here:

```
@computed("model.group.{full_name,name,allow_membership_requests}")
accessViaGroupText(group) {
  const name = group.full_name || group.name;
  const allowRequest = group.allow_membership_requests;
  return I18n.t(`topic.${allowRequest ? "group_request" : "group_join"}`, {
      name
    });
}
```

app/controllers/topics_controller.rb
```diff
@@ -84,6 +84,12 @@ def show
       end
     end
 
+    topic = Topic.where(id: (params[:id] || params[:topic_id])).first
```
SamSaffron: the topic is being looked up twice here, I think all the guardian checks etc should happen after we got the topic view and know the user has no access, so this only triggers on exceptional cases

spec/models/topic_spec.rb
```diff
@@ -2392,4 +2392,31 @@ def expect_the_right_notification_to_be_created(inviter, invitee)
...
   end
+
+  describe "#access_topic_via_group" do
+    let(:open_group) { Fabricate(:group, public_admission: true) }
+    let(:request_group) do
+      g = Fabricate(:group)
+      g.add_owner(user)
+      g.allow_membership_requests = true
+      g.save
+      g
```
tgxworld:
```suggestion
      Fabricate(:group, allow_membership_requests: true).tap { |g| g.add_owner(user) }
```

spec/models/topic_spec.rb
```diff
@@ -2392,4 +2392,31 @@ def expect_the_right_notification_to_be_created(inviter, invitee)
...
+    end
+    let(:category) { Fabricate(:category) }
+    let(:topic) { Fabricate(:topic, category: category) }
+
+    it "returns a group that is open or accepts membership requests and has access to the topic" do
+      expect(topic.access_topic_via_group).to eq(nil)
+
+      category.set_permissions(request_group => :full)
+      category.save
```
tgxworld:
```suggestion
      category.save!
```

spec/models/topic_spec.rb
```diff
@@ -2392,4 +2392,31 @@ def expect_the_right_notification_to_be_created(inviter, invitee)
...
+
+      expect(topic.access_topic_via_group).to eq(request_group)
+
+      category.set_permissions(request_group => :full, open_group => :full)
+      category.save
```
tgxworld:
```suggestion
      category.save!
```

spec/requests/topics_controller_spec.rb
```diff
@@ -1245,13 +1246,20 @@ def topic_user_post_timings_count(user, topic)
         allowed_user.save
         c
       end
+      let(:accessible_category) do
+        c = Fabricate(:category)
+        c.set_permissions(accessible_group => :full)
+        c.save
+        c
```
tgxworld:
```suggestion
        Fabricate(:category).tap { |c| c.set_permissions(accessible_group => :full) }
```

app/controllers/topics_controller.rb
```diff
@@ -937,6 +943,25 @@ def perform_show_response
     end
   end
 
+  def perform_hidden_topic_show_response(topic)
+    if request.head?
```
tgxworld: Hmm what is the reason the app needs to handle this?

app/models/topic.rb
```diff
@@ -1377,6 +1377,15 @@ def update_category_topic_count_by(num)
     end
   end
 
+  def access_topic_via_group
+    Group
+      .joins(:category_groups)
+      .where("category_groups.category_id = ?", self.category_id)
+      .where("groups.public_admission = true OR groups.allow_membership_requests = true")
```
tgxworld:
```suggestion
      .where("groups.public_admission OR groups.allow_membership_requests")
```

lib/guardian/topic_guardian.rb
```diff
@@ -147,6 +147,10 @@ def can_see_topic_if_not_deleted?(topic)
     can_see_topic?(topic, false)
   end
 
+  def can_get_access_to_topic?(topic)
+    return true if topic&.access_topic_via_group.present? && authenticated?
```
tgxworld:
```suggestion
    topic&.access_topic_via_group.present? && authenticated?
```

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -905,6 +905,39 @@ export default Ember.Controller.extend(bufferedProperty("model"), {
...
 
+    joinGroup() {
+      const groupId = this.get("model.group.id");
+      if (groupId) {
+        if (this.get("model.group.allow_membership_requests")) {
+          const groupName = this.get("model.group.name");
+          return ajax(`/groups/${groupName}/request_membership`, {
+            type: "POST",
+            data: {
+              reason: I18n.t("topic.group_request_reason", {
```
eviltrout: It's a little odd to me to have the client insert the reason like this. I would assume as reason parameter would be designed for the user to type in a custom reason.

I often think of a 3rd party consumer of our API. Would we expect someone using the API gem to create this reason the same way?

Perhaps we instead pass the topic_id and no reason, and have the server build the string?

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -905,6 +905,39 @@ export default Ember.Controller.extend(bufferedProperty("model"), {
...
+                topic_url: this.get("model.url")
+              })
+            }
+          })
+            .then(() => {
+              bootbox.alert(
+                I18n.t("topic.group_request_sent", {
+                  group_name: this.get("model.group.full_name")
+                }),
+                () => window.history.back()
```
eviltrout: Whenever possible I try to avoid refreshing the entire page, since it's a better/faster user experience. Could you instead route to a URL?

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -905,6 +905,39 @@ export default Ember.Controller.extend(bufferedProperty("model"), {
...
+              );
+            })
+            .catch(popupAjaxError);
+        } else {
+          return ajax(`/groups/${groupId}/members`, {
+            type: "PUT",
+            data: { user_id: this.get("currentUser.id") }
+          })
+            .then(() => window.location.reload())
```
eviltrout: Also here, I would prefer avoiding a full page reload.

app/assets/javascripts/discourse/routes/topic-from-params.js.es6
```diff
@@ -36,6 +36,11 @@ export default Discourse.Route.extend({
         // TODO we are seeing errors where closest post is null and this is exploding
         // we need better handling and logging for this condition.
 
+        // there are no closestPost for hidden topics
+        if (topic.topic_hidden) {
```
eviltrout: `topic.topic_hidden` is a little repetitive. Could we name it `topic.hidden` or perhaps `topic.view_hidden`?

app/assets/javascripts/discourse/templates/components/topic-join-group-notice.hbs
```diff
@@ -0,0 +1,6 @@
+{{accessViaGroupText}}
+<br>
```
eviltrout: I prefer to avoid `<br>` tags for formatting. Instead, I'd suggest to wrap the text in a span or div and add a bottom margin.

app/controllers/topics_controller.rb
```diff
@@ -939,6 +942,20 @@ def perform_show_response
...
 
+  def perform_hidden_topic_show_response(topic)
+    respond_to do |format|
+      format.html do
+        @topic_view = nil
+        render :show
+      end
+
+      format.json do
+        render_json_dump(HiddenTopicViewSerializer.new(topic, scope: guardian, root: nil))
```
eviltrout: Is there a particular reason you didn't use `render_serialized` here? It's somewhat simpler.

config/locales/client.en.yml
```diff
@@ -1913,6 +1913,10 @@ en:
       toggle_information: "toggle topic details"
       read_more_in_category: "Want to read more? Browse other topics in {{catLink}} or {{latestLink}}."
       read_more: "Want to read more? {{catLink}} or {{latestLink}}."
+      group_request: "You can not see this topic, but you can request membership to the {{name}} group to see it"
```
```suggestion
      group_request: "You need to request membership to the {{name}} group to see this topic"
```

config/locales/client.en.yml
```diff
@@ -1913,6 +1913,10 @@ en:
...
+      group_join: "You can not see this topic, but you can join the {{name}} group to see it"
```
```suggestion
      group_join: "You need join the {{name}} group to see this topic"
```

config/locales/client.en.yml
```diff
@@ -1913,6 +1913,10 @@ en:
...
+      group_request_reason: "I would like to join the group to be able to access this [topic]({{topic_url}})"
```
```suggestion
      group_request_reason: "I would like to join the group so I may access [this topic]({{topic_url}})"
```

config/locales/client.en.yml
```diff
@@ -1913,6 +1913,10 @@ en:
...
+      group_request_sent: "Your group membership request has been sent, you will be informed when it's accepted."
```
```suggestion
      group_request_sent: "Your group membership request has been sent. You will be informed when it's accepted."
```

tgxworld:
@majakomel  Sorry it looks like there are some changes to the master branch. Can you rebase this? Thank you :)

SamSaffron:
I think we need one more rebase here, its a bit too big to be considered for 2.3, pushing it back to 2.4

jjaffeux:
https://github.com/discourse/discourse/pull/7521

PR #7524, opened 2019-05-10
DEV: Clear caches between tests

danielwaterworth:
This PR clears all `DistributedCache`s before every test and makes the caches use the memory backend during testing. This makes it possible to prefabricate themes.

I ran the tests before and after these changes, the time before was 11m 52s, the time afterwards was 11m 38s.

### DIFF
--- a/lib/distributed_cache.rb
+++ b/lib/distributed_cache.rb
@@ -3,6 +3,36 @@
 require 'message_bus/distributed_cache'
 
 class DistributedCache < MessageBus::DistributedCache
+  class << self
+    module Test
+      def caches
+        @caches ||= {}
+      end
+
+      def get(key, **opts)
+        caches[key] ||= new(key, **opts)
+      end
+
+      def clear_caches!
+        caches.each_value do |cache|
+          cache.hash.clear
+        end
+      end
+    end
+
+    module Real
+      def get(key, **opts)
+        new(key, **opts)
+      end
+    end
+
+    if Rails.env.test?
+      include Test
+    else
+      include Real
+    end
+  end
+
   def initialize(key, manager: nil, namespace: true)
     super(
       key,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -563,7 +563,7 @@ class ApplicationController < ActionController::Base
   end
 
   def self.banner_json_cache
-    @banner_json_cache ||= DistributedCache.new("banner_json")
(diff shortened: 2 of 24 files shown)


lib/distributed_cache.rb
```diff
@@ -3,6 +3,36 @@
...
+      def caches
+        @caches ||= {}
+      end
+
+      def get(key, **opts)
+        caches[key] ||= new(key, **opts)
+      end
+
+      def clear_caches!
+        caches.each_value do |cache|
```
SamSaffron: hmmm `@caches = {}` ?

SamSaffron:
I am fine with this, it generally makes message bus more predictable in test cause the redis dependency is removed. 

@tgxworld  you good with this?

lib/distributed_cache.rb
danielwaterworth: In this version, the classes still have instance variables that refer to caches.

danielwaterworth: And there's one cache that's [not a class instance variable too](https://github.com/discourse/discourse/blob/master/lib/active_record/connection_adapters/postgresql_fallback_adapter.rb#L17), which is why I prefer it this way if possible.

tgxworld:
I was wondering if we can make this change without changing the API to `.get`. Plugins might be using the cache and it is a pain to have to support two different APIs.

SamSaffron:
The more I think about this I worry a bit about this level of fakedom. 

Maybe as a first step here just leave what we have and swap it to using memory transport. 

Maybe we introduce a custom memory transport that does basic tracking, then we can keep `.new` and all the base implementation and simply ask transport to clear all the caches that published messages during the test? 

My general concern here is that this is not longer really a distributed cache we are simply using a hash here with sugar API.

danielwaterworth:
> My general concern here is that this is not longer really a distributed cache we are simply using a hash here with sugar API.

Perhaps the main takeaway from this PR is that you _can_ just replace the DistributedCaches with a thin wrapper around hashes and the test suite still passes, which means, the tests are never testing with stale or missing data.

SamSaffron:
I am closing this, I did merge in backend memory change cause I think it is a better long term move. 

But I am uncomfortable only half solving the asynchronicity problem. What I want is `backend: :memory, threaded: false` this will help all around. 

> perhaps the main takeaway from this PR is that you can just replace the DistributedCaches

There is that, except that if the implementation of DistributedCache breaks in message_bus, the tests will catch it now, and they will not post this change.

PR #7526, opened 2019-05-10
FEATURE: Multi-file javascript support for themes

davidtaylorhq:
You can now add javascript files under `/js/*` in a theme, and they will be loaded as if they were included in core, or a plugin. If you give something the same name as a core/plugin file, it will be overridden. Support file extensions are `.js.es6`, `.hbs` and `.raw.hbs`.

### DIFF
--- a/spec/models/theme_field_spec.rb
+++ b/spec/models/theme_field_spec.rb
@@ -172,7 +172,7 @@ HTML
     expect(hbs_field.reload.value_baked).to include('Ember.TEMPLATES["discovery"]')
     expect(raw_hbs_field.reload.value_baked).to include('Discourse.RAW_TEMPLATES["discourse/templates/discovery"]')
     expect(unknown_field.reload.value_baked).to eq("")
-    expect(unknown_field.reload.error).to eq("Unrecognized file extension: blah")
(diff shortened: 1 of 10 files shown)


spec/models/theme_field_spec.rb
```diff
@@ -147,6 +147,40 @@
...
+        };
+        console.log('hello');
+      });
+    JS
+    expect(js_field.reload.value_baked).to eq(expected_js.strip)
+
+    expect(hbs_field.reload.value_baked).to include('Ember.TEMPLATES["discovery"]')
+    expect(raw_hbs_field.reload.value_baked).to include('Discourse.RAW_TEMPLATES["discourse/templates/discovery"]')
+    expect(unknown_field.reload.value_baked).to eq("")
+    expect(unknown_field.reload.error).to eq("Unrecognized file extension: blah")
```
OsamaSayegh: Very minor nitpick: it might be worth using `I18n.t` for the assertion instead of hardcoding the string here. A small copy edit shouldn't break the build, right?

SamSaffron:
Feel free to merge as soon as you start your day! Also be sure to update various meta topics to explain new functionality :confetti_ball:

PR #7530, opened 2019-05-10
FIX: Refresh automatic groups after inviting moderators.

nbianca:


tgxworld:
Are automatic groups being altered in the controller? If so, I don't think that should be the case.

spec/requests/invites_controller_spec.rb
```diff
@@ -312,6 +312,17 @@
             expect(Jobs::SendSystemMessage.jobs.size).to eq(1)
           end
 
+          it 'refreshes automatic groups if staff' do
+            topic.user.grant_admin!
+            invite.update(moderator: true)
```
tgxworld:
```suggestion
            invite.update!(moderator: true)
```

spec/requests/invites_controller_spec.rb
```diff
@@ -312,6 +312,17 @@
...
+
+            put "/invites/show/#{invite.invite_key}.json"
+            expect(response.status).to eq(200)
+            expect(JSON.parse(response.body)["success"]).to eq(true)
```
tgxworld: I think asserting for a `200` response status here is sufficient.

spec/requests/invites_controller_spec.rb
```diff
@@ -312,6 +312,17 @@
...
+
+            expect(invite.reload.user.groups.to_a).to include(Group.find_by(name: "moderators"), Group.find_by(name: "staff"))
```
tgxworld:
```suggestion
            expect(invite.reload.user.groups.pluck(:name)).to contain_exactly("moderators", "staff")
```

tgxworld:
Looks good to me. Thank you @nbianca

PR #7537, opened 2019-05-13
Update client.pl_PL.yml

huczas:


ZogStriP:
Thanks but all translations have to be done on [transifex](https://transifex.com/discourse/discourse-org/) otherwise they'd be overwritten next time we pull them.

PR #7538, opened 2019-05-13
Support multi-group user search

romanrizzi:


app/controllers/users_controller.rb
```diff
@@ -843,15 +843,15 @@ def search_users
     topic_id = topic_id.to_i if topic_id
     topic_allowed_users = params[:topic_allowed_users] || false
 
-    if params[:group].present?
-      @group = Group.find_by(name: params[:group])
+    if params[:groups].present?
```
eviltrout: This looks like a breaking API change to me. We need to ensure that if people are using the old `group` parameter that it continues to work.

romanrizzi: Yes, it's a breaking change. Looks like only `discourse-assign` is using this option though (searched in core and the `all-the-plugins` repo).

app/controllers/users_controller.rb
```diff
@@ -843,15 +843,17 @@ def search_users
     topic_id = topic_id.to_i if topic_id
     topic_allowed_users = params[:topic_allowed_users] || false
 
-    if params[:group].present?
-      @group = Group.find_by(name: params[:group])
+    group_names = params[:groups] || []
```
tgxworld: I think we need to update the request specs to reflect the change here.

romanrizzi: Done 👍

spec/requests/users_controller_spec.rb
```diff
@@ -3021,6 +3021,60 @@ def post_user
...
+          expect(users_found).to be_empty
+        end
+
+        def users_found
+          JSON.parse(response.body)['users'].map { |u| u['username'] }
+        end
+      end
+    end
+
+    context 'searching by group name' do
```
tgxworld: Is this complete?

spec/requests/users_controller_spec.rb
```diff
@@ -3021,6 +3021,60 @@ def post_user
...
+            group: exclusive_group.name,
+            term: user.username
+          }
+
+          expect(users_found).to be_empty
+        end
+
+        it 'returns results if the user is member of one of the groups' do
+          exclusive_group.add(user)
+          exclusive_group.save
```
tgxworld:
```suggestion
          exclusive_group.save!
```

spec/requests/users_controller_spec.rb
```diff
@@ -3021,6 +3021,60 @@ def post_user
           expect(JSON.parse(response.body)).not_to have_key(:groups)
         end
       end
+
+      describe 'when searching by group name' do
+        fab!(:exclusive_group) { Fabricate(:group) }
+
+        it 'return results if the user is a group member' do
+          exclusive_group.add(user)
+          exclusive_group.save
```
tgxworld:
```suggestion
          exclusive_group.save!
```

spec/models/user_search_spec.rb
```diff
@@ -51,10 +51,22 @@ def search_for(*args)
...
     expect(results.count).to eq(1)
   end
 
+  it 'allows filtering by multiple groups' do
+    group_1 = Fabricate(:group)
+    sam = Fabricate(:user, username: 'sam')
+    group_2 = Fabricate(:group)
+    samantha = Fabricate(:user, username: 'samantha')
+    group_1.add(sam)
+    group_2.add(samantha)
```
tgxworld: Hmm in `users_controller_spec` we're saving the group after adding a user but we're not doing so here. Might want to ensure the consistency here :grin:

app/models/user_search.rb
```diff
@@ -13,19 +13,19 @@ def initialize(term, opts = {})
     @searching_user = opts[:searching_user]
     @include_staged_users = opts[:include_staged_users] || false
     @limit = opts[:limit] || 20
-    @group = opts[:group]
+    @groups = opts[:groups]
     @guardian = Guardian.new(@searching_user)
-    @guardian.ensure_can_see_group!(@group) if @group
+    @groups.each { |group| @guardian.ensure_can_see_group!(group) } if @groups
```
tgxworld: This is going to hit us with multiple queries. Can you see if there is a possibility of introducing a `can_see_groups` method that does this more efficiently seeing that we've already loaded all the groups into memory.

romanrizzi:
Thanks @tgxworld for the thorough review! Definitely not my best piece of work 🤣

tgxworld:
@romanrizzi No worries :) That is the purpose of reviews :grin: The changes look good to me :+1:

PR #7539, opened 2019-05-14
FEATURE: Implement Onebox for posts including polls.

nbianca:


plugins/poll/assets/stylesheets/common/poll.scss
```diff
@@ -120,6 +120,30 @@ div.poll {
   }
 }
 
+aside.quote {
```
ZogStriP: Is this still needed?

plugins/poll/plugin.rb
```diff
@@ -435,6 +437,16 @@ def voters
     end
   end
 
+  on(:reduce_excerpt) do |fragment|
+    fragment.css("div.poll").each(&:remove)
+  end
+
+  on(:reduce_onebox) do |onebox, post|
```
ZogStriP: I don't think we need to handle oneboxes differently. We should use the excerpt, and if there's a poll in the excerpt, just replace it with a link to the post.

lib/pretty_text.rb
```diff
@@ -340,10 +340,14 @@ def self.extract_links(html)
   def self.excerpt(html, max_length, options = {})
     # TODO: properly fix this HACK in ExcerptParser without introducing XSS
     doc = Nokogiri::HTML.fragment(html)
+    DiscourseEvent.trigger(:before_excerpt, doc, options)
     strip_image_wrapping(doc)
     html = doc.to_html
 
-    ExcerptParser.get_excerpt(html, max_length, options)
+    excerpt = ExcerptParser.get_excerpt(html, max_length, options)
+    DiscourseEvent.trigger(:after_excerpt, excerpt, options)
```
ZogStriP: I don't think we need 2 new events to handle this case.

I would rename `before_excerpt` to `reduce_excerpt` to be consistent with `reduce_cooked`.

And in the `reduce_excerpt` event, in the poll plugin, I would just replace all occurences of a poll with a link to the post.

Then the `ExcerptParser` will take care of doing the "cutting" where it see fits.

plugins/poll/spec/lib/pretty_text_spec.rb
```diff
@@ -119,4 +119,22 @@ def n(html)
...
+      * World
+      [/poll]
+    EOF
+
+    onebox = Oneboxer.onebox_raw(post.full_url, user_id: Fabricate(:user).id)
+    doc = Nokogiri::HTML(onebox[:preview])
+
+    expect(onebox[:preview]).to include("A post with a poll")
+    expect(onebox[:preview]).to include("<a href='#{post.url}'>poll</a>")
+    expect(doc.css("div.poll")).not_to eq(nil)
```
ZogStriP: What is this testing? Shouldn't `doc.css("div.poll")` be empty?

app/models/user_profile.rb
```diff
@@ -24,6 +24,7 @@ class UserProfile < ActiveRecord::Base
   BAKED_VERSION = 1
 
   def bio_excerpt(length = 350, opts = {})
+    return nil unless bio_cooked.present?
```
ZogStriP:
```suggestion
    return nil if bio_cooked.blank?
```

PR #7543, opened 2019-05-15
FIX: Show membership requests link just for group owners.

nbianca:


app/assets/javascripts/discourse/widgets/post.js.es6
```diff
@@ -365,6 +365,23 @@ createWidget("post-contents", {
...
+      attrs.membershipRequest &&
+      this.currentUser.owned_groups &&
+      this.currentUser.owned_groups.includes(attrs.membershipRequest)
+    ) {
+      result.push(
+        h("div.membership-request", [
+          h("hr"),
+          h(
+            "a",
+            { attributes: { href: `/g/${attrs.membershipRequest}/requests` } },
```
eviltrout: I don't think this link is subfolder safe?

app/assets/javascripts/discourse/widgets/post.js.es6
```diff
@@ -365,6 +365,27 @@ createWidget("post-contents", {
     ];
     result = result.concat(applyDecorators(this, "after-cooked", attrs, state));
 
+    if (
+      attrs.membershipRequest &&
+      this.currentUser.owned_groups &&
+      this.currentUser.owned_groups.includes(attrs.membershipRequest)
```
tgxworld: Are we able to figure out if we should show the membership request link in the `TopicView` serializer instead? The worry I have here is that the membership request links are only for PMs but we're adding additional fields to the current user payload for every page load.

app/serializers/current_user_serializer.rb
```diff
@@ -43,6 +43,7 @@ class CurrentUserSerializer < BasicUserSerializer
              :top_category_ids,
              :hide_profile_and_presence,
              :groups,
+             :owned_groups,
```
SamSaffron: I am with @tgxworld  here I think I would like one less query by default, so we need to feed this information in on demand, vast majority of the time the client does not need to keep track of which groups are owned by current user vs not. (note when sending them down we should use ids anyway)

SamSaffron:
I would like this feature, but it feels a bit to big to fit into 2.3 for now. pushing it back.

eviltrout:
Bouncing this for @ZogStriP who I believe was the last one we were waiting on a review from.

app/controllers/groups_controller.rb
```diff
@@ -443,20 +443,12 @@ def request_membership
...
-      </a>
-    EOF
-
     post = PostCreator.new(current_user,
       title: I18n.t('groups.request_membership_pm.title', group_name: group.name),
-      raw: raw,
+      raw: params[:reason],
       archetype: Archetype.private_message,
       target_usernames: usernames.join(','),
+      custom_fields: { membership_request: group.name },
```
ZogStriP: What if the name of the group changes in the meantime?

app/serializers/basic_post_serializer.rb
```diff
@@ -38,7 +38,27 @@ def cooked
         I18n.t('flagging.user_must_edit')
       end
     else
-      object.filter_quotes(@parent_post)
+      cooked = object.filter_quotes(@parent_post)
+
+      if scope&.user
+        group = Group
```
ZogStriP: If you have the `id` of the group, then this could become

```ruby
group_id = object.custom_fields["membership_request"]

if scope&.user && GroupUser.where(user: scope.user, group_id: group_id, owner: true).exists?
  group = Group.find(group_id)
  cooked << <<~HTML
    <hr />
    <a href="#{Discourse.base_uri}/g/#{group.name}/requests">
      #{I18n.t('groups.request_membership_pm.handle')}
    </a>
  HTML
end
```

app/controllers/groups_controller.rb
ZogStriP: Also `membership_request` might not be the best name for that custom field? What about `requested_group_id`?

PR #7545, opened 2019-05-15
Refactor keyboard shortcuts modal to fix RTL locales

OsamaSayegh:
This PR does 2 things:

* It extracts the HTML parts that represents the keys from the the translation file and replaces them with a variable `%{shortcut}`
* Wraps the HTML parts in a `<span dir="ltr">` so that the order of the keys is maintained correctly in RTL locales.

Merging this means all existing translation of all locales will be invalidated. So I've written script that removes the HTML and replaces it with `%{shortcut}` for all locales, and then I'll use @gschlager's script to push the updated translation to Transifex.

Don't merge now since there are some improvements that @gschlager would like to make to his script first, but I've opened this PR in the meantime for review.


app/assets/javascripts/discourse/controllers/keyboard-shortcuts-help.js.es6
```diff
@@ -1,7 +1,123 @@
...
 
+const KEY = "keyboard_shortcuts_help";
+
+const SHIFT = I18n.t("shortcut_modifier_key.shift_lowercase");
+const ALT = I18n.t("shortcut_modifier_key.alt_lowercase");
+const CTRL = I18n.t("shortcut_modifier_key.ctrl_lowercase");
+const ENTER = I18n.t("shortcut_modifier_key.enter");
+const COMMA = I18n.t("number.format.delimiter") + " ";
+
+function trans(key, { sc = [], sc2 = [], deli = COMMA }) {
```
jjaffeux: what is the meaning of "trans" here ?

app/assets/javascripts/discourse/controllers/keyboard-shortcuts-help.js.es6
```diff
@@ -1,7 +1,123 @@
...
+  const context = {};
+  const mapped = [sc, sc2]
+    .map(shortcut => {
```
jjaffeux: oneliner

```
.map(shortcut => shortcut.map(k => `<b>${k}</b>`).join(deli))
```

app/assets/javascripts/discourse/controllers/keyboard-shortcuts-help.js.es6
OsamaSayegh: translate. I'm going to rename it to `translate` since `trans` seems to be ambiguous. If you have a better name for it let me know.

jjaffeux: That's maybe too verbose given how much it's used but maybe something like `handleTextDirection` ?

OsamaSayegh: I like this suggestion! Will make it so, thanks!

config/locales/client.en.yml
```diff
@@ -2792,59 +2797,59 @@ en:
...
         title: "Navigation"
-        jump: "<b>#</b> Go to post #"
-        back: "<b>u</b> Back"
-        up_down: "<b>k</b>/<b>j</b> Move selection &uarr; &darr;"
-        open: "<b>o</b> or <b>Enter</b> Open selected topic"
-        next_prev: "<b>shift</b>+<b>j</b>/<b>shift</b>+<b>k</b> Next/previous section"
+        jump: "%{shortcut} Go to post #"
+        back: "%{shortcut} Back"
+        up_down: "%{shortcut}/%{shortcut_2} Move selection &uarr; &darr;"
+        open: "%{shortcut} or %{shortcut_2} Open selected topic"
```
gschlager: This doesn't feel right to me. As a translator I could quite easily do a stupid thing and break the HTML.

`"%{shortcut_2} or %{shortcut} Open selected topic"` would result in `Enter</span> or <span dir="ltr">o Open selected topic`

Translations should only have one `%{shortcut}` interpolation key. That way translators can't break the HTML and we don't have to update translations when we decide to add or remove a shortcut.

Instead there should be a translation for shortcut alternatives: `shortcut_alternatives: %{shortcut1} or %{shortcut2}`
I wonder if we should split the next/previous and up/down shortcuts into two entries, otherwise we will need a translation for that as well if we want to stick with the "/" instead of "or".

OsamaSayegh: Wow, this is a very good point that I completely missed, thanks for raising it.

Just to make sure I understand correctly, the output of `I18n.t("shortcut_alternatives", ....)` should be wrapped in a `<span>` and then the whole thing will replace the `%{shortcut}` variable in `"%{shortcut} Open selected topic"`. Is this right?

Regarding splitting next/previous etc.: I think we should keep them the way they are right now; having an extra translation key for the `/` is not a huge problem, right?

gschlager: > the output of `I18n.t("shortcut_alternatives", ....)` should be wrapped in a `<span>` and then the whole thing will replace the `%{shortcut}` variable in `"%{shortcut} Open selected topic"`. Is this right?

Yes, exactly.

> having an extra translation key for the `/` is not a huge problem, right?

Correct. That's not a problem.

app/assets/javascripts/discourse/controllers/keyboard-shortcuts-help.js.es6
OsamaSayegh: I deviated a little bit here and went with different names: `buildShortcut()`, `buildHTML()` and `wrapInSpan()`.

`buildHTML()` generates for example `<kbd>g</kbd>, <kbd>h</kbd>` and calls `wrapInSpan()` on that and returns it.
`buildShortcut()` uses `buildHTML()` output as the `%{shortcut}` variable for the the given translation key. Let me know if there is still ambiguous code.

config/locales/client.en.yml
OsamaSayegh: Ok I've made those changes. Let me know if there is anything I missed. Thanks!

config/locales/client.en.yml
```diff
@@ -1464,6 +1464,11 @@ en:
       shift: "Shift"
       ctrl: "Ctrl"
       alt: "Alt"
+      enter: "Enter"
+      shift_lowercase: "shift"
```
gschlager: Do we really need the lowercase versions? AFAIK only Apple uses lowercase keys on their keyboards.

https://codekeyboards.com/galleria/8.jpg

OsamaSayegh: My keyboard (Razer) uses lowercase on the modifier keys 😛 

But I don't have any feelings on this; I simply added the lowercase versions because that's what the keyboard shortcuts modal uses at the moment. If it makes you uneasy I'm happy to remove it.

gschlager: Yeah, I strongly feel we shouldn't create duplicate translations for such a minor reason. I'd go with the existing translations and remove the lowercase versions.

PR #7548, opened 2019-05-15
Feature/add attachment support to xenforo importer

elepedus:
Re-submitting having refactored slightly to avoid logic errors & having added the suggested Dir.exist? check.


SamSaffron:
Thanks @elepedus !

PR #7551, opened 2019-05-15
FEATURE: Support private uploads in S3

pmusaraj:


SamSaffron:
@davidtaylorhq  can you review?

davidtaylorhq:
Overall I like the design here. The redirects are a really clean way to provide security without making the code too complex. I think we could even apply the same technique to images later. I installed the branch on my sandbox and had a play. Feedback:


- When you toggle the setting, it seems like existing uploads are private... But they aren't! You get the JS modal when clicking a private link, but if you right click and copy the link,  it works!
  - Simple option, but still not great: update the copy to make it clear you need to run the rake task, and prevent the JS popup if files aren't actually private
  - Awesome option: detect the site setting change, automtically adjust S3 permissions, and automatically queue the necessary posts for rebake

- Running rake uploads:s3_set_private_uploads_acl should rebake the posts (unless we go for 'awesome' mode above where the rake task isn't required)

- There is no way to make files public again using the rake task. Again, if we go for 'awesome' mode above then the rake task won't be required at all

- This is using the `/uploads/` route to provide the redirects. Works ok in development, but in production we apply cache headers in NGINX, so browsers end up caching the short-lived S3 token for a whole year: https://github.com/discourse/discourse/blob/master/config/nginx.sample.conf#L155-L156

- It only rewrites urls that match the `a.attachment` selector. People often copy/paste links around the forum, or remove the `style` attribute from the link in the composer. I think we should simply check all `a` tags and replace the links if they are secure attachments.

pmusaraj:
Thanks, David!

> Awesome option: detect the site setting change, automtically adjust S3 permissions, and automatically queue the necessary posts for rebake

💯 This would be very nice, will work on it. 

> This is using the /uploads/ route to provide the redirects. Works ok in development, but in production we apply cache headers in NGINX...

The cached headers don't seem to be an issue on a DO instance for me. Maybe they only apply to images? I think you are right, it might be best to use a separate route, maybe `/private-upload/`? It would make access rules easier to follow and having `private-upload` in the URL would be less magic, easier to troubleshoot. And I _think_ we would run into caching issues for private images, eventually. 

> It only rewrites urls that match the a.attachment selector

I'm hesitant about this. Extending to all links seems like overkill, maybe we can extend to any links to AWS URLs?

davidtaylorhq:
> The cached headers don't seem to be an issue on a DO instance for me

I was using a standard install on digital ocean as well. NGINX sets a 1y immutable header, which the browser then respects. I think maybe the browser will disable caching if dev tools are open? But I think you should still be able to see the headers being set in the dev tools. If you still can't repro let me know and I'll see if there's anything unusual on my end.

> I'm hesitant about this. Extending to all links seems like overkill, maybe we can extend to any links to AWS URLs?

Sure, that would be fine. There are other places where we [already iterate over every href](https://github.com/discourse/discourse/blob/fd5c5e326ffd59f3ddf28a4915e8350b4c5e60ec/lib/cooked_post_processor.rb#L600-L612) when cooking a post. Maybe the logic could be moved in there? Or maybe into `UrlHelper.cook_url`?

pmusaraj:
I can reproduce the caching issue in Firefox/Mac but not Safari/Mac. Interesting.

pmusaraj:
Closing this in favour of a much simpler PR.

PR #7552, opened 2019-05-15
bump ruby version in .ruby-version

orlando:
When Ruby version was updated in the docs in this commit https://github.com/discourse/discourse/pull/7100, this file was left out. It should have the recommended Ruby version.


ZogStriP:
Thanks 👍

PR #7556, opened 2019-05-16
DEV: attempts to make d-editor tests more resilient

jjaffeux:


test/javascripts/components/d-editor-test.js.es6
```diff
@@ -805,16 +804,22 @@ composerTestCase("replace-text event for composer", async function(assert) {
       assert,
       textarea
     ) {
-      const focusEvent = $.Event("focus");
-      const $input = $('textarea.d-editor-input');
-      $input.trigger(focusEvent);
-
       this.set("value", BEFORE);
+
+      await textarea.focus();
```
jjaffeux: await probably shouldnt be here

PR #7566, opened 2019-05-17
DEV: Debundle plugin javascript assets and don't load if disabled

vinothkannans:


.gitignore
```diff
@@ -127,3 +127,6 @@ vendor/bundle/*
 
 # Vagrant
 .vagrant
+
+# ignore auto-genarated plugin js assets
```
ZogStriP: Teeny-weeny typo

```suggestion
# ignore auto-generated plugin js assets
```

lib/discourse.rb
```diff
@@ -204,6 +204,22 @@ def self.unofficial_plugins
     plugins.find_all { |p| !p.metadata.official? }
   end
 
+  def self.find_plugins(args)
+    plugins.find_all do |plugin|
+      next if args[:include_official] == false && plugin.metadata.official?
+      next if args[:include_unofficial] == false && !plugin.metadata.official?
+      next if args[:include_disabled] != true && !plugin.enabled?
```
ZogStriP: Using equality is slightly easier to read than inequality

```suggestion
      next if args[:include_disabled] == false && !plugin.enabled?
```

lib/discourse.rb
```diff
@@ -204,6 +204,22 @@ def self.unofficial_plugins
     plugins.find_all { |p| !p.metadata.official? }
   end
 
+  def self.find_plugins(args)
+    plugins.find_all do |plugin|
```
ZogStriP: I tend to prefer `select` over `find_all` because it's shorter and I don't confuse it with `all?`

```suggestion
    plugins.select do |plugin|
```

spec/requests/list_controller_spec.rb
```diff
@@ -609,20 +609,4 @@
       expect(topic_titles).to include(topic_in_sub_category.title)
     end
   end
-
-  describe "safe mode" do
```
ZogStriP: Is it possible to keep a test to ensure safe mode is still loading the proper javascript assets?

vinothkannans: I spent many hours to keep existing or create new spec tests for this PR. But it looks it's not possible since all the dynamic file creations are happening only on initial app booting time.

PR #7567, opened 2019-05-17
FIX: prevents events on toolbar when in form

jjaffeux:
If a button is not of type button, pressing enter inside an `<input>` inside a `<form>` without the action attribute will trigger the first available `<button>` as most browsers default the type of an unspecified button to submit.

This commit also prevents d-editor-modal to be filled when it's hidden.


jjaffeux:
Source: https://github.com/emberjs/ember.js/issues/16760#issuecomment-398511804

PR #7571, opened 2019-05-20
FEATURE: add attachments to outgoing emails

majakomel:
https://meta.discourse.org/t/mailing-list-replacement-email-attachments/17710


lib/email/sender.rb
```diff
@@ -239,6 +242,28 @@ def self.host_for(base_url)
 
     private
 
+    def add_attachments
+      attachment_regex = /<a class="attachment" href="([^"]+)">([^<]*)<\/a>/
```
ZogStriP: Instead of checking the content of the message, I think it might be better to use the `post_uploads` table.

lib/email/sender.rb
```diff
@@ -239,6 +242,28 @@ def self.host_for(base_url)
...
+      attachments = @message.parts[0].body.raw_source.scan(attachment_regex)
+      if attachments.present?
+        upload_urls = attachments.collect(&:first)
+        uploads = Upload.where(url: upload_urls)
+        uploads.each do |upload|
+          if upload.url.start_with?("/uploads/default/")
```
ZogStriP: Does that work in subfolders/multisites?

lib/email/sender.rb
```diff
@@ -239,6 +241,25 @@ def self.host_for(base_url)
 
     private
 
+    def add_attachments(post)
+      uploads = post.uploads
+      return unless uploads.present?
```
ZogStriP: Technically, this is not needed, since `.each` won't iterate over an empty array.

You coud rewrite this

``` ruby
uploads = post.uploads
return unless uploads.present?
uploads.each do |upload|
```

with this one line

``` ruby
post.uploads.each do |upload
```

lib/email/sender.rb
```diff
@@ -239,6 +241,25 @@ def self.host_for(base_url)
...
+          file = Discourse.store.path_for(upload)
+        else
+          file = FileHelper.download(
+            "https:#{upload.url}",
+            max_file_size: SiteSetting.max_image_size_kb.kilobytes,
+            tmp_file_name: upload.original_filename,
+            follow_redirect: true
+          )
+        end
+        @message.attachments[upload.original_filename] = File.read(file)
```
ZogStriP: What happens if there's any error getting/downloading a file?

spec/components/email/sender_spec.rb
```diff
@@ -351,6 +351,35 @@
...
+
+    it "adds attachments to the email" do
+      post.link_post_uploads
+
+      message = Mail::Message.new to: 'disc@ourse.org', body: 'some content'
+      message.header['X-Discourse-Post-Id']  = post.id
+      message.header['X-Discourse-Topic-Id'] = post.topic_id
+      message.stubs(:deliver_now)
+
+      email_sender = Email::Sender.new(message, :valid_type)
```
ZogStriP: Since you don't use `email_sender`, you could just write this one line

``` ruby
Email::Sender.new(message, :valid_type).send
```

lib/email/sender.rb
```diff
@@ -239,6 +241,29 @@ def self.host_for(base_url)
...
 
+    def add_attachments(post)
+      post.uploads.each do |upload|
+        next if FileHelper.is_supported_image?(upload.original_filename)
+        begin
+          if Discourse.store.internal?
+            file = Discourse.store.path_for(upload)
+          else
+            file = FileHelper.download(
+              "https:#{upload.url}",
```
tgxworld: Are we certain that it'll always be `https`? We might want to check `SiteSetting.force_https` here.

spec/components/email/sender_spec.rb
```diff
@@ -351,6 +351,34 @@
...
+      SiteSetting.authorized_extensions = 'pdf'
+    end
+
+    it "adds attachments to the email" do
+      post.link_post_uploads
+
+      message = Mail::Message.new to: 'disc@ourse.org', body: 'some content'
+      message.header['X-Discourse-Post-Id']  = post.id
+      message.header['X-Discourse-Topic-Id'] = post.topic_id
+      message.stubs(:deliver_now)
```
tgxworld: We're tying to avoid the use of `stubs` in our code base moving forward. Do have a read through https://guides.rubyonrails.org/testing.html#testing-your-mailers to see the helpers that Rails provide.

lib/email/sender.rb
```diff
@@ -239,6 +241,29 @@ def self.host_for(base_url)
...
+              max_file_size: SiteSetting.max_image_size_kb.kilobytes,
```
tgxworld: Do we need to do that check here for local uploads as well?

lib/email/sender.rb
```diff
@@ -239,6 +241,33 @@ def self.host_for(base_url)
...
+              max_file_size: SiteSetting.max_attachment_size_kb.kilobytes,
+              tmp_file_name: upload.original_filename,
+              follow_redirect: true
+            )
+          end
+          email_size += File.size(file)
+          break if email_size > max_email_size
+          @message.attachments[upload.original_filename] = File.read(file)
+        rescue => e
+          Rails.logger.warn("Could not attach file #{upload.original_filename}: #{e.class} - #{e.message}\n#{e.backtrace.join("\n")}")
```
tgxworld: I think the `Post#id` will also be helpful here.

lib/email/sender.rb
```diff
@@ -239,6 +241,33 @@ def self.host_for(base_url)
 
     private
 
+    def add_attachments(post)
+      email_size = 0
+      max_email_size = 20.megabytes
```
tgxworld: We should probably move this into a site setting with a reasonable `max` value.

lib/email/sender.rb
```diff
@@ -239,6 +241,33 @@ def self.host_for(base_url)
...
+      post.uploads.each do |upload|
+        next if FileHelper.is_supported_image?(upload.original_filename)
```
tgxworld: Let's also test this as well to make sure images are never added as attachments.

lib/email/sender.rb
tgxworld: Actually I think this can be a multiple of `SiteSetting.max_attachment_size_kb`?

SamSaffron:
It will be nice to have this change but I would like to hold off till 2.4. 

Other stuff to consider is cases where the SMTP server gets angry cause attachments are too big, so we need some configurable server side limit to "max size of attachments in emails" (then the feature can be totally disabled by setting it to 0)

In fact the more I think about it, its probably better to have this default disabled.

gschlager:
@ZogStriP Do you want to review before I squash and merge?

My changes are:
* Site setting (default off) to restrict the max email size
* Trust filesize in uploads table and skip downloading file if it doesn't fit
* Doesn't abort after the first file that doesn't fit into the email. Instead it keeps going and adds every upload until the max email size is reached.
* Uses `FileStore` to get path of upload and download remote file. Should make this less brittle than `FileHelper.download`
* Logging improvements
* Added and improved specs

config/site_settings.yml
```diff
@@ -1021,6 +1021,9 @@ email:
   enable_forwarded_emails: false
   always_show_trimmed_content: false
   private_email: false
+  email_total_attachment_size_limit_kb:
+    default: 0
+    max: 51200
```
ZogStriP: We should probably have a `min: 0` as well?

gschlager: That's actually the default for `min`. 😉 
https://github.com/discourse/discourse/blob/a61ff16740146a0f089fb50ca757c47d8134266f/lib/validators/integer_setting_validator.rb#L6

PR #7572, opened 2019-05-20
DEV: Update test-prof

danielwaterworth:


SamSaffron:
does this mean we can not disable prefabrication anymore?

danielwaterworth:
Good catch. You're right.

danielwaterworth:
I added the option back in and noticed that the test suite didn't succeed with PREFABRICATION=0, so I'll filed a separate PR to fix that too: https://github.com/discourse/discourse/pull/7574

PR #7574, opened 2019-05-20
FIX: Fixed test suite with PREFABRICATION=0

danielwaterworth:
The issue here was that, with prefabrication, bumped_at was being
persisted and then loaded and the DB was storing it with less precision
than the object state.


SamSaffron:
I think we have a time matcher for this issue exactly

PR #7581, opened 2019-05-21
FIX: Update tags on change.

nbianca:
With this change, tags are automatically updated in the topic header when they change. Also, it updates when the topic is assigned / unassigned.

@jjaffeux, can you please review this one? I find this problem weird because:

* adding an unused argument that directly references the tag array works
* `discourse-tags` is still an unbound helper, but it works?
* using a bound helper did not help either (i.e. `htmlHelper`)

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/topic-category.hbs
+++ b/app/assets/javascripts/discourse/templates/components/topic-category.hbs
@@ -7,7 +7,7 @@
 <div class="topic-header-extra">
   {{#if siteSettings.tagging_enabled}}
     <div class="list-tags">
-      {{discourse-tags topic mode="list"}}
+      {{discourse-tags topic mode="list" tags=topic.tags}}
     </div>
   {{/if}}
   {{#if siteSettings.topic_featured_link_enabled}}


jjaffeux:
@eviltrout yes something is fishy here, or I don't understand our unbound helpers. AFAIK, unbound helpers are not supposed to re-render when property change, right?

That's at least how Ember doc defines it: https://api.emberjs.com/ember/3.10/classes/Ember.Templates.helpers/methods/unbound?anchor=unbound

Now if I define this kind of code:

```
// some hbs file
{{test-unbound testInfo something=testSomething}}
{{d-button action=(action "changeTestInfo")}}
{{d-button action=(action "changeTestSomething")}}

// some component file
  testInfo: false,
  testSomething: false,

  actions: {
    changeTestInfo() {
      this.toggleProperty("testInfo");
    },

    changeTestSomething() {
      this.toggleProperty("testSomething");
    }
  }

// helper file
import { registerUnbound } from "discourse-common/lib/helpers";

export default registerUnbound("test-unbound", function(name, params) {
  return new Handlebars.SafeString(`<span>${name}${params.something}</span>`);
});
```

They will render at each change of testInfo or testSomething, doesn't look unbound at all to me.

So it explains why even if not used the tags params, is triggering a re-render. ./cc @nbianca

eviltrout:
@jjaffeux and @nbianca I think this changed in an Ember version and we did not notice. By default in older versions of Ember helpers used to be unbound. At some point they removed this distinction (or we updated our internal API.)

Is it a problem for this PR?

We'll probably want to create a TODO for removing the confusing term `unbound`

jjaffeux:
dont think this is an issue for the PR, it’s just as other things probably relying on a wrong behavior.

Envoyé de mon iPhone

tgxworld:
@nbianca If an acceptance test here is not too complex, I think we should add one for this behavior here. Otherwise, it looks like the PR is ready to merge.

PR #7584, opened 2019-05-22
FIX: add title attribute for notifications

majakomel:
https://meta.discourse.org/t/what-does-the-notification-mean/109852/9

### DIFF
--- a/app/assets/javascripts/discourse/widgets/notification-item.js.es6
+++ b/app/assets/javascripts/discourse/widgets/notification-item.js.es6
@@ -148,6 +148,7 @@ createWidget("notification-item", {
     let infoKey = notName === "custom" ? data.message : notName;
     let text = emojiUnescape(this.text(notificationType, notName));
     let icon = iconNode(`notification.${infoKey}`);
+    let title = I18n.t(`notifications.titles.${notName ? notName : "custom"}`);
 
     // We can use a `<p>` tag here once other languages have fixed their HTML
     // translations.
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -148,6 +148,7 @@ createWidget("notification-item", {
     let infoKey = notName === "custom" ? data.message : notName;
     let text = emojiUnescape(this.text(notificationType, notName));
     let icon = iconNode(`notification.${infoKey}`);
+    let title = I18n.t(`notifications.titles.${notName}`);
```
eviltrout: I think it might be nice to handle the case when the key is not present. Could we use an empty string as the default perhaps?

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -148,6 +148,7 @@ createWidget("notification-item", {
     let infoKey = notName === "custom" ? data.message : notName;
     let text = emojiUnescape(this.text(notificationType, notName));
     let icon = iconNode(`notification.${infoKey}`);
+    let title = I18n.t(`notifications.titles.${notName ? notName : "custom"}`);
```
tgxworld: I don't think we need the blank string to be translated. Something like the following should do

```
let title;

if (notName) {
  title = I18n.t(`notification.titles.${notName}`)
} else {
  title =  "";
}
```

tgxworld:
@majakomel One more change requested otherwise it looks good to me. Feel free to merge once the changes have been addressed.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/c9a34aa10c230f39387d4c519366221d96d8795e

PR #7588, opened 2019-05-22
FIX: trigger user_updated webhook when avatar changes

majakomel:
https://meta.discourse.org/t/user-profile-avatar-icon-not-triggering-webhook/117089

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -124,6 +124,8 @@ class User < ActiveRecord::Base
   after_create :ensure_in_trust_level_group
   after_create :set_default_categories_preferences
 
+  after_update :trigger_user_updated_event, if: :saved_change_to_uploaded_avatar_id?
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
As a pattern does it make sense to do all the hooking here vs deeper in service object or user after save? @tgxworld  / @vinothkannans  what pattern have you been following here?

vinothkannans:
I think it should be in the user after save. Currently, except for user model everywhere else we're triggering the `_updated` in `after_commit` callback.

tgxworld:
@majakomel Shouldn't we only trigger an event if `User#uploaded_avatar_id` changes? 

https://github.com/discourse/discourse/blob/648af78298f7ed15385504a252c0e1541044962a/app/models/user.rb#L1504

tgxworld:
LGTM :+1: Thanks @majakomel

PR #7589, opened 2019-05-23
FEATURE: detect theme errors and catch them

OsamaSayegh:


app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -660,5 +660,32 @@ export function postRNWebviewMessage(prop, value) {
...
+  const currentUser = api.getCurrentUser();
+  if (!currentUser || !currentUser.get("admin")) {
+    return;
+  }
+  const message = I18n.t("themes.broken_theme_alert", {
+    theme: name
+  });
+  const html = `<div class="broken-theme-alert"><b>${message}</b></div>`;
+  Ember.$("body").prepend(html);
+}
```
jjaffeux: This could be replaced with vanilla js:

```
const alertDiv = document.createElement("div");
alertDiv.classList.add("broken-theme-alert")
alertDiv.innerText = "message"
document.body.prepend(alertDiv)
```

Note I removed the `<b>` because it doesn't have much interest to me. You should probably set font-weight on `.broken-theme-alert`

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -660,5 +660,32 @@ export function postRNWebviewMessage(prop, value) {
...
+    type: "POST",
+    cache: false
+  });
+}
+// this function is used in lib/theme_javascript_compiler.rb
+export function rescueThemeError(name, error, api) {
+  console.error(`"${name}" error:`, error);
+  reportToLogster(name, error);
+  const currentUser = api.getCurrentUser();
+  if (!currentUser || !currentUser.get("admin")) {
```
jjaffeux:
```suggestion
  if (!currentUser || !currentUser.admin) {
```

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -660,5 +660,32 @@ export function postRNWebviewMessage(prop, value) {
   }
 }
 
+function reportToLogster(name, error) {
+  const data = {
+    message: `${name} theme/component is throwing errors`,
+    stacktrace: error.stack
+  };
+
+  Ember.$.ajax("/logs" + "/report_js_error", {
```
jjaffeux: Why not using our ajax lib?

Also why not ""/logs/report_js_error"" instead of `"/logs" + "/report_js_error"`

app/assets/stylesheets/common/base/discourse.scss
```diff
@@ -732,3 +732,13 @@ table {
     color: $danger;
   }
 }
+
+.broken-theme-alert {
+  font-size: $base-font-size;
+  padding: 5px 0;
+  background: $danger;
+  width: 100%;
```
jjaffeux: is this necessary ?

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -660,5 +660,32 @@ export function postRNWebviewMessage(prop, value) {
...
+    data,
+    type: "POST",
+    cache: false
+  });
+}
+// this function is used in lib/theme_javascript_compiler.rb
+export function rescueThemeError(name, error, api) {
+  console.error(`"${name}" error:`, error);
+  reportToLogster(name, error);
+  const currentUser = api.getCurrentUser();
```
jjaffeux: I would add a line break before and after this block. Doesn't have any direct link with console logging and the alert warning.

OsamaSayegh:
@jjaffeux thanks for the review! I've made the changes you suggested 👍

app/assets/javascripts/discourse/lib/utilities.js.es6
OsamaSayegh: It seems MiniRacer doesn't like importing `ajax` from the utilities file.

```
MiniRacer::RuntimeError: Error: Could not find module `discourse/lib/ajax` imported from `discourse/lib/utilities`
```

https://travis-ci.org/discourse/discourse/jobs/536272772#L7870

Works perfectly fine locally though.

config/locales/client.en.yml
```diff
@@ -187,6 +187,7 @@ en:
 
     themes:
       default_description: "Default"
+      broken_theme_alert: "⚠️ Theme/component %{theme} has errors which may prevent your site from working correctly! You can disable it at %{path}."
```
gschlager: After this PR gets merged, please watch the [Timeline at Transifex](https://www.transifex.com/discourse/discourse-org/settings/timeline/) carefully for "Autofetch failed" errors. A year ago an Emoji stopped Transifex from doing auto updates. It will be interesting to see if they fixed the bug in the meantime. I highly doubt that it will work, so be prepared to show the warning sign in another way.

OsamaSayegh: Thanks @gschlager for the heads up. I did notice there is another string with an emoji here:

https://github.com/discourse/discourse/blob/1fbe078ae074f29395d79db0856d1991f1675ab5/config/locales/client.en.yml#L182

This line was last changed 2 years ago, so maybe the problem is specific to a set of emojis? This makes me wonder if emojis have to "translatable"?

gschlager: Yes, I think it depends on the Emoji. Transifex expects some of them to be represented as Unicode Surrogates which isn't supported by Ruby. You might not even need to include this Emoji in the translation as long as it still works with RTL. There's definitely no need to translate a warning sign.

OsamaSayegh: Ok I've moved the emoji from `client.en.yml` to the JS file and tested with an RTL locale and it looked alright 👍 

Will merge this PR once the build passes.

PR #7592, opened 2019-05-23
DEV: global s/this.get\("(\w+)"\)/this.$1

jjaffeux:


eviltrout:
This is a fairly risky change. We should only merge when we have some good time to debug!

jjaffeux:
> This is a fairly risky change. We should only merge when we have some good time to debug!

yeah will try to merge this on Monday, so I prepared it in advance.

PR #7595, opened 2019-05-24
FIX: delete system generated message when user_export record is deleted

techAPJ:
FIX: system generated message for user export should be closed by default

### DIFF
--- a/app/jobs/onceoff/clean_up_user_export_topics.rb
+++ b/app/jobs/onceoff/clean_up_user_export_topics.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Jobs
+  class CleanUpUserExportTopics < Jobs::Onceoff
+    def execute_onceoff(args)
+      translated_keys = []
+      I18n.available_locales.each do |l|
+        translated_keys << I18n.with_locale(l.to_sym) { I18n.t("system_messages.csv_export_succeeded.subject_template") }
+      end
+      translated_keys = translated_keys.uniq
+
+      slugs = []
+      translated_keys.each do |k|
+        slugs << "%-#{Slug.for(k.gsub('[%{export_title}]', ''))}"
+      end
+      # "[%{export_title}] 資料匯出已完成" gets converted to "%-topic", do not match that slug.
+      slugs = slugs.reject { |s| s == "%-topic" }
+
+      topics = Topic.with_deleted.where(<<~SQL, slugs, 2.days.ago)
+        slug LIKE ANY(ARRAY[?]) AND
+        archetype = 'private_message' AND
+        subtype = 'system_message' AND
+        posts_count = 1 AND
+        created_at < ? AND
+        user_id = -1
+      SQL
+
+      topics.each do |t|
+        Topic.transaction do
+          t.posts.first.destroy!
+          t.destroy!
+        end
+      end
+    end
+  end
+end
--- a/app/jobs/regular/export_csv_file.rb
+++ b/app/jobs/regular/export_csv_file.rb
@@ -87,7 +87,12 @@ module Jobs
       end
 
     ensure
-      notify_user(download_link, file_name, file_size, export_title)
+      post = notify_user(download_link, file_name, file_size, export_title)
+      if user_export.present? && post.present?
+        topic = post.topic
+        user_export.update_columns(topic_id: topic.id)
+        topic.update_status('closed', true, Discourse.system_user)
+      end
     end
 
     def user_archive_export
@@ -384,8 +389,9 @@ module Jobs
     end
 
     def notify_user(download_link, file_name, file_size, export_title)
+      post = nil
       if @current_user
-        if download_link.present?
+        post = if download_link.present?
           SystemMessage.create_from_system_user(
             @current_user,
             :csv_export_succeeded,
@@ -398,6 +404,7 @@ module Jobs
           SystemMessage.create_from_system_user(@current_user, :csv_export_failed)
         end
       end
+      post
     end
   end
 end
--- a/app/models/user_export.rb
+++ b/app/models/user_export.rb
@@ -3,6 +3,7 @@
 class UserExport < ActiveRecord::Base
   belongs_to :user
   belongs_to :upload, dependent: :destroy
+  belongs_to :topic, dependent: :destroy
 
   around_destroy :ignore_missing_post_uploads
 
@@ -14,7 +15,14 @@ class UserExport < ActiveRecord::Base
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
-      user_export.destroy!
+      UserExport.transaction do
+        begin
+          Post.where(topic_id: user_export.topic_id).find_each { |p| p.destroy! }
+          user_export.destroy!
+        rescue => e
+          Rails.logger.warn("Failed to remove user_export record with id #{user_export.id}: #{e.message}\n#{e.backtrace.join("\n")}")
+        end
+      end
     end
   end
 
@@ -34,4 +42,5 @@ end
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #  upload_id  :integer
+#  topic_id   :integer
 #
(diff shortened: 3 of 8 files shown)


db/migrate/20190523093215_add_topic_id_to_user_exports.rb
```diff
@@ -0,0 +1,5 @@
+class AddTopicIdToUserExports < ActiveRecord::Migration[5.2]
+  def change
+    add_column :user_exports, :topic_id, :integer
```
tgxworld: Does this need to be `null: false`? It seems like the links for user export are only exposed via a PM so I would think that we need to enforce that `UserExport` are always tied to a topic.

tgxworld:
I think we should do a clean up all the invalid PMs that have been created thus far.

db/migrate/20190523093215_add_topic_id_to_user_exports.rb
techAPJ: I was worried about existing rows in `user_exports` table. I will do something like `change_column_null :table_name, :column_name, false.

techAPJ:
@tgxworld regarding cleaning up existing invalid PMs @SamSaffron [already added](https://github.com/discourse/discourse/commit/954293655f9c60689a52101405e6ccb1175484c9#diff-a0e0257135a571aa3b9da75d1d9d6892) a rake task for it. Do we want to use that or create a new onceoff job for it?

tgxworld:
We need to clean this up across instances automatically so I think the rake task isn't going to work for self hosters. Note that the rake task is also incomplete because it only takes into consideration the `en` locale.

db/migrate/20190523093215_add_topic_id_to_user_exports.rb
techAPJ: Actually `null: false` might not be possible because the topic is created _after_ the UserExport record is created and then we update the `topic_id` column, same as what we do for `upload_id`.

tgxworld: Hmm that is tricky. Is there a reason we send a system message for the link instead of an email like we do for admins to download backups?

techAPJ: No specific reason. I guess this was how we envisioned this feature to work in initial spec i.e. downloading archive without needing to leave Discourse UX.

techAPJ: Note that we also send system message (with backup logs) when taking a backup and link Admin to admin/backups page to download that file. This is done for additional security purpose I believe.

tgxworld: I guess we just need to create the topic and upload before creating the `UserExport`?

techAPJ: That's not trivial to do as upload contains file name which includes `UserExport` id and topic has file download link which can't be generated without creating user_export record and upload first.

app/jobs/onceoff/clean_up_user_export_topics.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Jobs
+  class CleanUpUserExportTopics < Jobs::Onceoff
+    def execute_onceoff(args)
+      slug = "%-#{Slug.for(I18n.t('system_messages.csv_export_succeeded.subject_template').gsub('[%{export_title}]', ''))}"
```
tgxworld: I think we can't take the server's locale. If I am not wrong, the user export should be generated based on the `user's` locale?

app/jobs/regular/export_csv_file.rb
```diff
@@ -87,7 +87,12 @@ def execute(args)
       end
 
     ensure
-      notify_user(download_link, file_name, file_size, export_title)
+      post = notify_user(download_link, file_name, file_size, export_title)
+      if user_export.present? && post.present?
+        topic = post.topic
+        user_export.update_columns(topic_id: topic.id)
+        topic.update_status('closed', true, Discourse.system_user)
```
tgxworld: I'm not sure about this but are closed PMs deleted eventually?

db/migrate/20190523093215_add_topic_id_to_user_exports.rb
tgxworld: Ah icic. I guess we can just leave it as nullable then.

app/jobs/onceoff/clean_up_user_export_topics.rb
techAPJ: I verified locally that user export messages are generated based on server locale, not user's.

spec/jobs/clean_up_user_export_topics_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Jobs::CleanUpUserExportTopics do
+  fab!(:user) { Fabricate(:user) }
+
+  it 'should delete ancient user export system messages' do
+    post = SystemMessage.create_from_system_user(
+            user,
```
tgxworld: Minor but indentation looks off here?

spec/jobs/clean_up_user_export_topics_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    I18n.locale = :fr
+    post = SystemMessage.create_from_system_user(
+            user,
+            :csv_export_succeeded,
+            download_link: "http://example.com/download",
+            file_name: "xyz.gz",
+            file_size: "55",
+            export_title: "user_archive"
+          )
+    post.topic.update_attribute(:created_at, 5.days.ago)
```
tgxworld:
```suggestion
    post.topic.update!(:created_at, 5.days.ago)
```

tgxworld:
```suggestion
    post.topic.update!(created_at: 5.days.ago)
```

app/jobs/regular/export_csv_file.rb
techAPJ: No. I am closing the PM here per Sam's request. For hard deleting the PM I added `belongs_to :topic, dependent: :destroy` in `user_export.rb`.

app/jobs/onceoff/clean_up_user_export_topics.rb
tgxworld: What if the server's locale has been changed? If we don't want to loop through all the available locales, we can look through the site settings history. 

Also, it is a bug that the system message isn't generated using the user's locale.

app/jobs/regular/export_csv_file.rb
tgxworld: I think destroying a topic doesn't destroy it posts.

techAPJ: Fixed.

app/models/user_export.rb
```diff
@@ -14,6 +15,7 @@ def ignore_missing_post_uploads
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
+      Post.where(topic_id: user_export.topic_id).find_each { |p| p.destroy! }
```
tgxworld: We probably want to run this in a transaction. 

Also, what happens if an error is raised while destroying a user_export? Do we want to raise the error and stop the clean up midway or should we rescue errors and log?

app/jobs/onceoff/clean_up_user_export_topics.rb
techAPJ: We will now loop through all available locales.

app/models/user_export.rb
techAPJ: Wrapped in transaction and rescued + logged errors.

app/models/user_export.rb
```diff
@@ -14,7 +15,14 @@ def ignore_missing_post_uploads
 
   def self.remove_old_exports
     UserExport.where('created_at < ?', 2.days.ago).find_each do |user_export|
-      user_export.destroy!
+      UserExport.transaction do
+        begin
+          Post.where(topic_id: user_export.topic_id).find_each { |p| p.destroy! }
```
tgxworld: Hmm I think we should be careful here. If `user_export.topic_id` is `nil`, the query will find all posts that has `topic_id IS NULL`. That is though impossible since `Post#topic_id` is not `nullable`. Perhaps an `if` check to ensure that we don't run an extra query unnecessarily?

tgxworld:
@techAPJ  One more comment otherwise the changes look good to me. Feel free to merge after.

app/models/user_export.rb
techAPJ: Agreed, done.

PR #7596, opened 2019-05-24
FEATURE: Make staff action logs page infinite.

nbianca:


app/assets/javascripts/admin/controllers/admin-logs-staff-action-logs.js.es6
```diff
@@ -1,47 +1,75 @@
...
+  default as computed,
+  observes,
+  on
+} from "ember-addons/ember-computed-decorators";
 
 export default Ember.Controller.extend({
   loading: false,
   filters: null,
   userHistoryActions: [],
+  model: [],
```
eviltrout: This is dangerous to do in Ember because it means each instance of the class points at the same array instance. It is recommended to use `null` and initialize the model another way (say in the route.)

eviltrout: This is not recommended to do in Ember because all instances point at the same object (`[]`). Instead you should use `null` and initialize the value in another place such as the route.

app/assets/javascripts/admin/controllers/admin-logs-staff-action-logs.js.es6
```diff
@@ -1,47 +1,75 @@
...
+  nextPage: 0,
+  lastPage: null,
 
   filtersExists: Ember.computed.gt("filterCount", 0),
+  showTable: Ember.computed.gt("model.length", 0),
 
-  filterActionIdChanged: function() {
-    const filterActionId = this.get("filterActionId");
-    if (filterActionId) {
+  @observes("filterActionId")
```
eviltrout: Instead of using an observer here, is there a way you can trigger an event from the control when the value changes? It's always better to avoid observers if we can.

spec/requests/admin/staff_action_logs_controller_spec.rb
```diff
@@ -31,6 +31,26 @@
       )
     end
 
+    it 'generates logs with pages' do
+      1.upto(300).each do |idx|
```
eviltrout: This is a lot of records to create. Is there any way to test it with fewer? I'm just worried about a slow test here.

tgxworld:
Instead of keeping the page number on the client side, are we able to reuse the code here? 

https://github.com/discourse/discourse/blob/04abf181940dc55f760eafc5b0569183efaeb267/app/assets/javascripts/discourse/models/result-set.js.es6#L17-L38

nbianca:
The `ResultSet` changes is more involved than I thought. Can we merge this one as it is and I will followup with the implementation for `ResultSet`?

SamSaffron:
This is pretty surgical and does not add any new strings, I am assuming the risk and merging this in. ResultSet can wait on 2.4.

tgxworld:
@SamSaffron  Shouldn't  we wait for the `ResultSet` implementation that is used everywhere else instead of introducing a totally new pattern?

SamSaffron:
I am ok with this temporary solution for now until result set is in, I want this to land in 2.3, its quite common to want to dig through these logs and the way it is now is very problematic.  

There is no way we can land the result set solution in 2.4

PR #7597, opened 2019-05-24
FIX: Replace details content with instruction.

nbianca:


plugins/discourse-details/plugin.rb
```diff
@@ -31,8 +31,14 @@
...
 
-  on(:reduce_cooked) do |fragment|
-    fragment.css("details.elided").each(&:remove)
+  on(:reduce_cooked) do |fragment, post|
+    fragment.css("details").each do |el|
+      text = fragment.css("summary").text
+      link = fragment.document.create_element("a")
+      link["href"] = post.url if post
+      link.content = I18n.t("details.excerpt_details")
+      el.replace text + " " + link.to_html
```
riking: is this treating the text content of the summary as HTML? possible injection point

PR #7598, opened 2019-05-24
UX: Change category box link to href

awesomerobot:
previously using data-url... not sure if there's a good reason? but it's a simple change


tgxworld:
@SamSaffron can you review this? This reverts your change in https://github.com/discourse/discourse/commit/cc7ecb8a06f0dcb67b9eca8e6818cfef11d8dc59

SamSaffron:
Oh I remember this one @awesomerobot 

I think it was done for cases where we had urls inside the box and some browsers did not like nesting of `<a>` in another element at all.

SamSaffron:
@awesomerobot what should we do here?

PR #7599, opened 2019-05-24
FIX: Consistently handle category param

oblakeerickson:
See https://meta.discourse.org/t/api-post-to-posts-json-inconsistent-between-users/118571
for more info.

This commit removes a 5 year old temporary fix that is no longer needed.

https://github.com/discourse/discourse/commit/bc1824a6ed28bc9d5b75a561a00acde17fed6b7b#diff-d8c648926664f849aec050757bfcb6f9R72

The web interface uses category_id when creating a topic so I think we
should unify on category_id when using the api.

### DIFF
--- a/lib/seed_data/topics.rb
+++ b/lib/seed_data/topics.rb
@@ -132,7 +132,7 @@ module SeedData
         title: title,
         raw: raw,
         skip_validations: true,
-        category: category&.name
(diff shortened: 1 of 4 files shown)


tgxworld:
Looks good to me :+1:

PR #7601, opened 2019-05-25
FEATURE: Allow import of hidden site settings

rrooding:
**Updated**: Hidden site settings are now also imported using the rake site_settings:import task

**Old**:
~~The site_settings:import rake task is updated to accept an optional import_hidden parameter. The default behaviour stays the same.~~

~~When passing in the import_hidden parameter, the rake task will also import hidden site settings.~~

### DIFF
--- a/app/services/site_settings_task.rb
+++ b/app/services/site_settings_task.rb
@@ -11,8 +11,8 @@ class SiteSettingsTask
     h
   end
 
-  def self.import(yml, import_hidden: false)
-    h = SiteSettingsTask.export_to_hash(include_defaults: true, include_hidden: import_hidden)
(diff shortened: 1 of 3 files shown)


SamSaffron:
I am not sure about this, I can not think of a legitimate reason to ignore hidden site sittings in the file, I think we should just change default behavior, even if the default export excludes them

rrooding:
That was my first thought as well, but I figured there would be a reason for it. I can update the PR to import hidden site settings by default.

SamSaffron:
I think this is ok, very low risk cause you need to explicitly work hard to get hidden site settings into your yaml.

PR #7603, opened 2019-05-27
FEATURE: Support `[description|attachment](upload://<short-sha>)` in MD.

tgxworld:


tgxworld:
Will need to fix the tests tomorrow. We're applying callbacks to pretender incorrectly.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -187,6 +188,26 @@ function setupImageDimensions(md) {
   md.renderer.rules.image = renderImage;
 }
 
+function renderAttachment(tokens, idx, options, env, slf) {
+  const linkOpenToken = tokens[idx];
+  const linkTextToken = tokens[idx + 1];
+  const split = linkTextToken.content.split("|");
```
jjaffeux: Got to be careful on this:

```
"".split("|") // [""]
```

A safer way is:
```
"".split("|").filter(x => x) // []
```

eviltrout:
LGTM, nice :+1:

pmusaraj:
Hmm, let's make sure this works with S3 uploads before merging, it doesn't work on my local S3 set up.

tgxworld:
do you mean the short-url route doesn’t work with S3?

pmusaraj:
Yes, the short-url route gives me a 404 for an upload on S3. The controller isn't finding the upload. But also, `send_file` won't work with a remote URL, I think we'll need to do a redirect here for S3 uploads. (I ran into something similar while working on https://github.com/discourse/discourse/pull/7551/files#diff-60328924b5c773b42aacabc0b5698af4L89).

tgxworld:
O good catch :) let me update this PR and the existing tests to make sure S3 uploads work.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
tgxworld: Thanks for the tip. :+1: I'm checking for the length of `split` in the following code so I don't need an empty array.

spec/components/pretty_text_spec.rb
```diff
@@ -1290,17 +1296,25 @@ def strip_image_wrapping(html)
         </li>
         </ul>
         <p><img src="#{upload.url}" alt="upload"></p>
```
SamSaffron: I think it is worth explicitly configuring a CDN here so it is clear that the links that come back are not CDN-ified.

tgxworld: It is impossible for `Upload#url` to return a CDN-ified URL here.

SamSaffron: Sorry I mean set the global setting so the site has a CDN this confirms that when cooking you never cdnify things.

tgxworld: OK done :+1:

SamSaffron:
@pmusaraj if this is looking good to you, please merge it in when you start your day, I had a look at the tests here and its a :+1:  from me! You need to build on this so it makes sense that you merge it.

tgxworld:
test failure is not related

pmusaraj:
This looks awesome! 

One tiny thing: do we need to amend the nginx config to skip caching headers for `uploads/short-url` requests?

PR #7611, opened 2019-05-27
FEATURE: Mute topics tagged with both muted and unmuted tags.

udan11:


lib/topic_query.rb
```diff
@@ -892,13 +892,22 @@ def remove_muted_tags(list, user, opts = nil)
           list # if viewing the topic list for a muted tag, show all the topics
         else
           muted_tag_ids = TagUser.lookup(user, :muted).pluck(:tag_id)
```
SamSaffron: the vast majority of the time this will return `[]` so no need to compose anything extra to the relation.

lib/topic_query.rb
```diff
@@ -892,13 +892,22 @@ def remove_muted_tags(list, user, opts = nil)
...
-          list = list.where("
-            EXISTS (
-              SELECT 1
-                FROM topic_tags tt
-               WHERE tt.tag_id NOT IN (:tag_ids)
-                 AND tt.topic_id = topics.id
-            ) OR NOT EXISTS (SELECT 1 FROM topic_tags tt WHERE tt.topic_id = topics.id)", tag_ids: muted_tag_ids)
+          if SiteSetting.mute_other_present_tags
+            list = list.where("
+              NOT EXISTS(
```
SamSaffron: A bit concerned about `not exists` performance at scale, have you tested the query against a site like meta? you can do it in data explorer.

spec/components/topic_query_spec.rb
```diff
@@ -258,11 +259,27 @@
...
+      expect(topic_ids).to include(tagged_topic.id)
+      expect(topic_ids).to include(untagged_topic.id)
+
+      SiteSetting.mute_other_present_tags = false
+
+      topic_ids = topic_query.list_latest.topics.map(&:id)
+      expect(topic_ids).not_to include(muted_topic.id)
+      expect(topic_ids).to include(muted_tagged_topic.id)
+      expect(topic_ids).to include(tagged_topic.id)
+      expect(topic_ids).to include(untagged_topic.id)
```
tgxworld: There is a `contain_exactly` matcher in RSpec that you can use to remove the duplication here.

```
expect(topic_ids).to contain_exactly(muted_tag_topic.id, tagged_topic.id, untagged_topic.id)
```

lib/topic_query.rb
udan11: The query times are very similar.

lib/topic_query.rb
udan11: I refactored the whole method, it should be better now. :)

PR #7614, opened 2019-05-27
FEATURE: Send notification when member was accepted to group.

nbianca:


app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -58,7 +60,7 @@ createWidget("notification-item", {
       return postUrl(attrs.slug, topicId, attrs.post_number);
     }
 
-    if (attrs.notification_type === INVITED_TYPE) {
+    if (attrs.notification_type === INVITEE_ACCEPTED_TYPE) {
```
ZogStriP: This needs a test ;)

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -72,6 +74,10 @@ createWidget("notification-item", {
       );
     }
 
+    if (attrs.notification_type === GROUP_INVITE_TYPE) {
```
ZogStriP: This also needs a test ;)

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -112,10 +118,13 @@ createWidget("notification-item", {
...
-    if (notificationType === GROUP_SUMMARY_TYPE) {
+    if (notificationType === GROUP_MESSAGE_SUMMARY_TYPE) {
       const count = data.inbox_count;
-      const group_name = data.group_name;
-      return I18n.t(scope, { count, group_name });
+      const groupName = data.group_name;
+      return I18n.t(scope, { count, group_name: groupName });
+    } else if (notificationType === GROUP_INVITE_TYPE) {
+      const groupName = data.group_name;
+      return I18n.t(scope, { group_name: groupName });
```
ZogStriP: This also needs a test ;)

app/models/group.rb
```diff
@@ -506,9 +506,20 @@ def usernames
 
   PUBLISH_CATEGORIES_LIMIT = 10
 
-  def add(user)
+  def add(user, notify: false)
     self.users.push(user) unless self.users.include?(user)
 
+    if notify
+      Notification.create!(
```
ZogStriP: This also needs a test ;)

tgxworld:
Can we find out what broke due to https://github.com/discourse/discourse/pull/7503? We should improve our test case to account for whatever was broken due to the previous PR.

nbianca:
The problem with the previous pull request was that I refactored some things that I expected to be named the same on the client and server side, but they were not.

* On server-side, we had `:invitee_accepted`, but on the client side there was `INVITED_TYPE`.
* On server-side, we had `:group_message_summary`, but on the client side there was `GROUP_SUMMARY_TYPE`.

I will be adding the tests to ensure it will not happen again, but to be honest, it does very little testing.

ZogStriP:
> I expected to be named the same on the client and server side, but they were not

I think we should fix that inconsistency as well.

spec/models/group_spec.rb
```diff
@@ -725,6 +725,10 @@ def can_view?(user, group)
         .and change { user.title }.from('AAAA').to('BBBB')
     end
 
+    it "can send a notification to the user" do
+      expect { group.add(user, notify: true) }.to change { Notification.count }.by(1)
```
tgxworld: We don't want to just check for the count. The attributes of the notification are important as well.

test/javascripts/fixtures/site-fixtures.js.es6
```diff
@@ -19,7 +17,15 @@ export default {
         moved_post: 10,
         linked: 11,
         granted_badge: 12,
-        liked_consolidated: LIKED_CONSOLIDATED_TYPE
```
tgxworld: Is there a reason for us to not import the constant so as to avoid duplication?

nbianca: I preferred to do it this way for consistency. For me, it's fine either way. In my opinion, it doesn't matter so much since it's a test fixture.

tgxworld: Usually we don't want to hard code "magic" numbers in multiple places. If the number changes, we end up having to change it in multiple places and we might not even know which are the places that should change.

ZogStriP:
This is now conflicting 👊

app/models/group.rb
```diff
@@ -521,9 +521,20 @@ def usernames
...
+        notification_type: Notification.types[:group_invite],
```
SamSaffron: I would actually call this `:membership_request_accepted` ... `group_invite` is a bit confusing as a name for this thing.

config/locales/client.en.yml
```diff
@@ -1704,6 +1704,7 @@ en:
       granted_badge: "Earned '{{description}}'"
       topic_reminder: "<span>{{username}}</span> {{description}}"
       watching_first_post: "<span>New Topic</span> {{description}}"
+      group_invite: "Invited to '{{group_name}}'"
```
SamSaffron: Is this not `Membership accepted in {{group_name}}` ?

SamSaffron:
overall this looks good sans a few minor changes, can you rebase?

PR #7616, opened 2019-05-28
FEATURE: Allow running message_bus in a different redis instance

xfalcox:
Adds `DISCOURSE_MESSAGE_BUS_REDIS_ENABLED` env var, that when set
to true, will allow Discourse to connect to a different redis
instance for MessageBus needs.

When enabled you can configure the same env vars user for redis,
but prefixed by `MESSAGE_BUS`, eg:

`DISCOURSE_MESSAGE_BUS_REDIS_HOST`

### DIFF
--- a/app/models/global_setting.rb
+++ b/app/models/global_setting.rb
@@ -136,6 +136,7 @@ class GlobalSetting
   # For testing purposes
   def self.reset_redis_config!
     @config = nil
+    @message_bus_config = nil
   end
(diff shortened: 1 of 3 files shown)


SamSaffron:
Yeah I think this is all we need. It is practically impossible to test short of spinning up a full environment since all this stuff is initialized on boot. I am merging this in... good luck!

PR #7619, opened 2019-05-28
WIP memory optim 2

jjaffeux:


SamSaffron:
This seems like it is doing good stuff to me, shall we merge?

jjaffeux:
> This seems like it is doing good stuff to me, shall we merge?

That's my feeling too, and my local branch is doing more things... but weirdly it has 0 impact, and its even worse... I'm missing something here. Finishing picker and resuming this, but yes I'm sure this is the right path

PR #7640, opened 2019-05-29
DEV: Fix store.path_for for both local and external storage

pmusaraj:


tgxworld:
@pmusaraj ah was my revert incomplete? 🙀

pmusaraj:
Not at all, this restores what you initially did and fixes the `path_for` calls expecting `nil` (hopefully, everywhere).

pmusaraj:
From what I can tell, there are `path_for` calls expecting a local path left in backup-uploads-to-s3 and [downsize_uploads.rb](https://github.com/discourse/discourse/blob/master/script/downsize_uploads.rb).

davidtaylorhq:
I think this is confusing naming. IMO `path_for` should only ever return a local filesystem **path**, not a URL. If it should always return a URL, then it should be renamed `url_for`.

Right now, `Store.path_for` is the only way we have to determine whether a specific upload is on S3/local. If we lose this, then we will need to re-structure the S3 migration procedure so that we are never 'in limbo' between two stores.

pmusaraj:
@david I think you're right about the confusing naming. 

I do think we should use something other than `Store.path_for` to determine whether a specific upload is on S3 or local. Would `Upload.local?` do the same thing, for example? (Now that I mention this, I did add a base store `local?` method in the PR, which is redundant.)

davidtaylorhq:
`Upload.local?` would be a site-level thing. With our current processes, we need something which can be different per-upload. Right now the application can cope with some uploads on S3, and some on local. If we do `.local?` at a site level then that won't work any more.

Conceptually I do prefer only allowing local OR S3, but it will require a lot of process changes in addition to this PR.

pmusaraj:
Closing, I think we won't take this path.

PR #7641, opened 2019-05-29
Fake reads option on PostCreator

danielwaterworth:
When creating a post, you can now pretend that a user has read `post_number` if they have already read `post_number - 1`.


danielwaterworth:
There's already a function to do exactly this on PostTiming, so closing this PR

PR #7642, opened 2019-05-29
DEV: Make groups/new accessible by plugins

romanrizzi:
- Expose a new plugin outlet. 
- Pass group model to the group-member-dropdown so it can be accessed by plugins
- Extend group permitted params from a plugin

### DIFF
--- a/app/assets/javascripts/discourse/templates/group-index.hbs
+++ b/app/assets/javascripts/discourse/templates/group-index.hbs
@@ -64,7 +64,8 @@
                     removeMember=(action "removeMember")
                     makeOwner=(action "makeOwner")
                     removeOwner=(action "removeOwner")
-                    member=m}}
+                    member=m
+                    group=model}}
               {{/if}}
             </td>
           </tr>
--- a/app/assets/javascripts/discourse/templates/group-index.hbs
+++ b/app/assets/javascripts/discourse/templates/group-index.hbs
@@ -67,6 +67,7 @@
                     member=m
                     group=model}}
               {{/if}}
+              {{!-- group parameter is used by plugins --}}
             </td>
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/templates/group-index.hbs
```diff
@@ -64,7 +64,8 @@
                     removeMember=(action "removeMember")
                     makeOwner=(action "makeOwner")
                     removeOwner=(action "removeOwner")
-                    member=m}}
+                    member=m
+                    group=model}}
```
romanrizzi: Not sure if there's a better way to do this. group=model is only useful from a plugin's perspective 🤔

jjaffeux: maybe as a metadata hash ?

romanrizzi: Maybe I should rename it to `pluginMetadata`?

davidtaylorhq: I must be missing something, what's wrong with it being called `group`? That seems like a simple and descriptive variable name to me

romanrizzi: My intention here is to emphasize this is needed for plugins and avoid it being deleted because someone thought is not being used.

davidtaylorhq: Ah, I see. Maybe just a comment next to the line? `{{!-- group parameter is used by plugins --}}`. If it's renamed then I like the idea of `pluginMetadata` - that's super clear :+1:

romanrizzi:
Specs added!

PR #7645, opened 2019-05-30
DEV: Class that converts MD with old attachment links to new MD.

tgxworld:


app/services/inline_uploads.rb
```diff
@@ -0,0 +1,68 @@
...
+class InlineUploads
+  def self.process(markdown)
+    cooked_fragment = Nokogiri::HTML::fragment(PrettyText.cook(markdown.dup))
+    link_occurences = []
+
+    cooked_fragment.traverse do |node|
+      next unless node.children.count == 1 && node.children[0].children.blank?
+
+      if seen_link = upload_fragment_hint(node)
+        if actual_link = node.attributes["href"]&.value
```
SamSaffron: you need to look at IMG SRC as well.

app/services/inline_uploads.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require_dependency "pretty_text"
+
+class InlineUploads
+  def self.process(markdown)
```
SamSaffron: maybe support: `InlineUploads(markdown, on_missing: lambda{ |url| 77  }`

every missing upload gets upload_id 77, this is super flexible cause we can inject how to behave with unknown url from the outside

SamSaffron: so we can use a google search to try and find the image and then create upload, finally giving it an id. or search on disk ... the or in the case of broken image inline a broken image

app/services/inline_uploads.rb
```diff
@@ -0,0 +1,105 @@
...
+      else
+        regexps << /(#{SiteSetting.Upload.s3_base_url}\/original\/(\dX\/(?:[a-f0-9]\/)*[a-f0-9]{40}[a-z0-9\.]*))/
+        regexps << /(#{SiteSetting.Upload.s3_cdn_url}\/original\/(\dX\/(?:[a-f0-9]\/)*[a-f0-9]{40}[a-z0-9\.]*))/
+      end
+    else
+      regexps << /(\/uploads\/#{db}\/original\/(\dX\/(?:[a-f0-9]\/)*[a-f0-9]{40}[a-z0-9\.]*))/
+    end
+
+    regexps.each do |regexp|
+      node.to_s.scan(regexp).each do |matched|
```
ZogStriP: Might want to do the `node.to_s` outside the loop?

app/services/inline_uploads.rb
```diff
@@ -0,0 +1,105 @@
...
+        next unless (link && is_valid)
+
+        if link.include?(seen_link)
+          begin
+            uri = URI(link)
+          rescue URI::Error
+          end
+
+          if !Discourse.store.external?
+            next if uri.host && uri.host != Discourse.current_hostname
```
ZogStriP: What happens if `uri` is `nil` when the `URI::Error` exception is thrown?

app/services/inline_uploads.rb
```diff
@@ -0,0 +1,105 @@
...
+    link_occurences = []
+
+    cooked_fragment.traverse do |node|
+      if node.name == "img"
+        # Do nothing
+      elsif !(node.children.count == 1 && node.children[0].children.blank?)
+        next
+      end
+
+      if seen_link = matched_uploads(node).first
```
ZogStriP: Shouldn't we call `matched_uploads(node)` on `a` and `img` nodes only?

tgxworld: Nope we need to walk all the nodes because there are cases like code blocks which may include a link.

tgxworld:
Merging this in without review because the changes here are too big.

PR #7646, opened 2019-05-30
 DEV: debundle plugin css assets and don't load if disabled

vinothkannans:


SamSaffron:
I worry we are sitting on this for so long, is this ready to merge?

vinothkannans:
@SamSaffron it is ready to merge.

SamSaffron:
This looks quite good to me. Maybe @OsamaSayegh can also have a look at this and review cause he is super familiar with this system. 

The only minor thing I am seeing here (which is not new) is the name `asset_name` is quite confusing imo, maybe we should call it `plugin_directory_name` to be explicit or something else?

vinothkannans:
The variable `asset_name` is now renamed to `plugin_directory_name`. And I added a minor fix too. cc @OsamaSayegh

PR #7650, opened 2019-05-30
Added commit approved notification type

danielwaterworth:


jjaffeux:
@danielwaterworth please make sure you follow https://meta.discourse.org/t/github-checkin-prefix-convention/19392

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -118,6 +119,16 @@ createWidget("notification-item", {
       return I18n.t(scope, { count, group_name });
     }
 
+    if (notificationType === COMMIT_APPROVED_TYPE) {
+      const num_approved_commits = data.num_approved_commits;
```
eviltrout: Our variables in JS should be camelCase

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -118,6 +119,16 @@ createWidget("notification-item", {
...
+
+      if (num_approved_commits == 1) {
+        return I18n.t(scope + '.one');
```
jjaffeux:
```suggestion
        return I18n.t(`${scope}.one`);
```

app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -118,6 +119,16 @@ createWidget("notification-item", {
...
+      } else {
+        return I18n.t(scope + '.other', { num_approved_commits });
```
jjaffeux:
```suggestion
        return I18n.t(`${scope}.other`, { num_approved_commits });
```

app/models/notification.rb
```diff
@@ -63,7 +63,8 @@ def self.types
                         watching_first_post: 17,
                         topic_reminder: 18,
                         liked_consolidated: 19,
-                        post_approved: 20
+                        post_approved: 20,
+                        commit_approved: 21
```
SamSaffron: To me this is probably the scope of the entire change. We just want the number reserved in core, the rest of the functionality / strings and so on can live in the plugin imo.

danielwaterworth: Ok, I'll work out how to monkey-patch the other changes in.

danielwaterworth:
@jjaffeux, @eviltrout, Thanks for the feedback.

app/models/notification.rb
SamSaffron: Prefer a less monkey patchy way, can you expose an API for it from the JS side?

danielwaterworth: Ok

SamSaffron:
Daniel closing this in favor of the far simpler core change of simply keeping the number in core and the rest in the plugin.

PR #7656, opened 2019-05-31
UX: Merge settings related to muted tags.

udan11:


config/site_settings.yml
```diff
@@ -2059,9 +2059,13 @@ tags:
...
   remove_muted_tags_from_latest:
-    default: false
-  mute_other_present_tags:
-    default: true
+    client: true
+    type: enum
+    default: always
+    choices:
+      - never
+      - when used alone or with other muted tags
```
eviltrout: This name is cumbersome. I think we should define an enum with translations, so that the id itself can be one word (or two word with snake case). It keeps the code simpler and translators can change it.

spec/components/topic_query_spec.rb
```diff
@@ -263,13 +263,21 @@
       topic_ids = topic_query.list_new.topics.map(&:id)
       expect(topic_ids).to contain_exactly(tagged_topic.id, untagged_topic.id)
 
-      SiteSetting.mute_other_present_tags = false
+      SiteSetting.remove_muted_tags_from_latest = 'when used alone or with other muted tags'
```
eviltrout: Shouldn't this be only_muted?

PR #7663, opened 2019-05-31
FIX: Periodically ensure consistency of categories.

nbianca:


SamSaffron:
Cool, feel free to merge this in once there is a trivial test for ensure_consistency, you can just fabricate a category and then nuke the topic. 

I think the behavior here needs a bit of adjusting cause. 

```
c=Fabricate(:category)
c.topic.destroy!
... 
Ensure consistency should fix this up. 
```

So you probably want to left join into topics.

app/models/category.rb
```diff
@@ -664,6 +664,14 @@ def permissions_compatibility_validator
     end
   end
 
+  def self.ensure_consistency!
+    Category.all.each do |category|
```
ZogStriP: Even though we aren't supposed to have too much categories, it's always best to use "lazy" loading when iterating over an entire table.

```suggestion
    Category.find_each do |category|
```

ZogStriP: Like @SamSaffron said, you should also filter out categories which already have a topic using the `LEFT JOIN`.

PR #7667, opened 2019-05-31
FIX: use crawler layout when saving url in Wayback Machine

majakomel:
https://meta.discourse.org/t/make-discourse-play-nice-with-the-wayback-machine/34579/45


SamSaffron:
Looks good to me! Overall I am not the hugest fan of shared examples, I find them hard to follow, in this case it is simple enough so I am on the fence. I guess I would have probably gone with a trivial extra integration test, then you don't need to do all that expensive extra integration testing just for wayback machine.

PR #7668, opened 2019-05-31
Cleanup crawler styles, improve schema.org markup 

awesomerobot:
* Added some more spacing to various links for easier tapping
* Added clearer divisions between posts on mobile 
* Updated schema.org markup for likes and reply count
* Updated schema.org markup for tags in topics – previously we used https://schema.org/ItemList, but https://schema.org/keywords seems more appropriate to me... also resolved an error we had with ItemList
* Now that https://schema.org/ItemList is freed up (can only be used once per page) I updated the list of links under OP to use it   
* Added schema for crawler nav
* Validated schema via https://search.google.com/structured-data/testing-tool/u/0/ - we still get one error with itemlist (cancel	
All values provided for url must have the same domain.), but after some googling it looks like an error with the validator

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -63,7 +63,7 @@ en:
   topics: "Topics"
   posts: "posts"
   loading: "Loading"
-  powered_by_html: 'Powered by <a href="https://www.discourse.org">Discourse</a>, best viewed with JavaScript enabled'
+  powered_by_html: 'Powered by <a rel="nofollow" href="https://www.discourse.org">Discourse</a>, best viewed with JavaScript enabled'
   log_in: "Log In"
   submit: "Submit"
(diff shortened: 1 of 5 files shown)


config/locales/server.en.yml
```diff
@@ -63,7 +63,7 @@ en:
   topics: "Topics"
   posts: "posts"
   loading: "Loading"
-  powered_by_html: 'Powered by <a href="https://www.discourse.org">Discourse</a>, best viewed with JavaScript enabled'
+  powered_by_html: 'Powered by <a rel="nofollow" href="https://www.discourse.org">Discourse</a>, best viewed with JavaScript enabled'
```
SamSaffron: do we want to give up our pagerank here? mixed on doing this.

SamSaffron:
I am loving you did this cleanup. Merging it in, but putting a followup on the nofollow

PR #7672, opened 2019-06-01
FIX: Do not raise an exception if no primary email exists.

udan11:


SamSaffron:
I am a bit uneasy here, in that I would love to know how we have users in our systems without emails.

coding-horror:
Yeah this is not good

tgxworld:
Closing this as a user record should never be missing a primary email record.

PR #7673, opened 2019-06-01
DEV: Pass editor to plugin hook handling composer uploads.

udan11:
Passing the `composer-editor` component instance makes it easier for plugins to access jQuery-File-Upload widget (i.e. `editor.$().fileupload(...)`).


SamSaffron:
This is good, thanks!

PR #7676, opened 2019-06-03
DEV: Stop compiling themes during DB migration. Recompile on cdn change.

davidtaylorhq:
This is an improved implementation for bc8b7b13

### DIFF
--- a/db/migrate/20190603134013_change_theme_field_compiler_version.rb
+++ b/db/migrate/20190603134013_change_theme_field_compiler_version.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeThemeFieldCompilerVersion < ActiveRecord::Migration[5.2]
+  def change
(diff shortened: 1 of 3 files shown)


db/migrate/20190603134013_change_theme_field_compiler_version.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeThemeFieldCompilerVersion < ActiveRecord::Migration[5.2]
+  def change
+    change_column(:theme_fields, :compiler_version, :string, limit: 50)
```
eviltrout: It doesn't matter that you added this limit, but did you know in postgres there is no performance/space advantage to setting a limit on varchar?

https://stackoverflow.com/questions/1067061/does-a-varchar-fields-declared-size-have-any-impact-in-postgresql

PR #7677, opened 2019-06-03
FEATURE: Support private attachments when using S3 storage

pmusaraj:


app/jobs/regular/update_private_uploads_acl.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Jobs
+  class UpdatePrivateUploadsAcl < Jobs::Base
+
```
eviltrout: It might be smart to add a comment here saying this job only executes when the site setting changes. When I first saw it I freaked out a little bit about how it's running on every upload.

lib/file_store/s3_store.rb
```diff
@@ -105,6 +104,23 @@ def path_for(upload)
       FileStore::LocalStore.new.path_for(upload) if url && url[/^\/[^\/]/]
     end
 
+    def url_for(upload)
+      if upload.private?
+        if Rails.configuration.multisite
```
eviltrout: This code is repeated below, so it would likely be better as a `get_upload_key(upload)` function that could be re-used.

SamSaffron:
I really need this merged! @tgxworld can you review and merge in, if any small amendments are needed feel free to add followup commits.

config/initializers/014-track-setting-changes.rb
```diff
@@ -34,6 +34,8 @@
 
   Jobs.enqueue(:update_s3_inventory) if [:s3_inventory, :s3_upload_bucket].include?(name)
 
+  Jobs.enqueue(:update_private_uploads_acl) if [:prevent_anons_from_downloading_files].include?(name)
```
SamSaffron:
```suggestion
  Jobs.enqueue(:update_private_uploads_acl) if name == :prevent_anons_from_downloading_files
```
:) 

One interesting general problem we have is if this setting is shadowed by global this thing will never trigger.

SamSaffron:
This looks safe, well tested to me. The big caveat is that old sites are going to need @tgxworld's remapper which is fine cause we are going to make that happen. 

Merging this in and enabling it on dev.

PR #7678, opened 2019-06-03
DEV: introduces prettier for core hbs files

jjaffeux:


jjaffeux:
cc @eviltrout @ZogStriP @romanrizzi

jjaffeux:
@eviltrout https://github.com/discourse/discourse/pull/7678/files#diff-c8c45f49ff4abc8b46ee342667f0d839R53 as discussed

I will try to see if we can at least have this fixed

ZogStriP:
Yeah, that one is really bad 😲

romanrizzi:
The only downside of the PR is that removes a lot of blank lines that were introduced to improve readability

jjaffeux:
> The only downside of the PR is that removes a lot of blank lines that were introduced to improve readability

That doesn't bother me much, it was not consistent at all

SamSaffron:
I also support the "not thinking" movement. It does touch lots of stuff though so lets hold off till after 2.3

jjaffeux:
Update the PR, few things getting better, only one last remaining thing is truly bohering me.

PR #7679, opened 2019-06-03
FIX: use hijack for processing bulk invites

techAPJ:
FIX: do not store bulk invite CSV file on server

### DIFF
--- a/app/controllers/invites_controller.rb
+++ b/app/controllers/invites_controller.rb
@@ -172,24 +172,31 @@ class InvitesController < ApplicationController
   def upload_csv
     guardian.ensure_can_bulk_invite_to_forum!(current_user)
 
-    file = params[:file] || params[:files].first
-    name = params[:name] || File.basename(file.original_filename, ".*")
-    extension = File.extname(file.original_filename)
+    hijack do
+      begin
+        file = params[:file] || params[:files].first
 
-    begin
-      data = if extension.downcase == ".csv"
-        path = Invite.create_csv(file, name)
-        Jobs.enqueue(:bulk_invite, filename: "#{name}#{extension}", current_user_id: current_user.id)
-        { url: path }
-      else
-        failed_json.merge(errors: [I18n.t("bulk_invite.file_should_be_csv")])
+        if File.read(file.tempfile).scan(/\n/).count.to_i > 50000
+          return render json: failed_json.merge(errors: [I18n.t("bulk_invite.max_rows")]), status: 422
+        end
+
+        invites = []
+        CSV.foreach(file.tempfile) do |row|
+          invite_hash = { email: row[0], groups: row[1], topic_id: row[2] }
+          if invite_hash[:email].present?
+            invites.push(invite_hash)
+          end
+        end
+        if invites.present?
+          Jobs.enqueue(:bulk_invite, invites: invites, current_user_id: current_user.id)
+          render json: success_json
+        else
+          render json: failed_json.merge(errors: [I18n.t("bulk_invite.error")]), status: 422
+        end
+      rescue
+        render json: failed_json.merge(errors: [I18n.t("bulk_invite.error")]), status: 422
       end
-    rescue
-      failed_json.merge(errors: [I18n.t("bulk_invite.error")])
     end
-    MessageBus.publish("/uploads/csv", data.as_json, user_ids: [current_user.id])
-
-    render json: success_json
   end
 
   def fetch_username
--- a/app/jobs/regular/bulk_invite.rb
+++ b/app/jobs/regular/bulk_invite.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'csv'
 require_dependency 'system_message'
 
 module Jobs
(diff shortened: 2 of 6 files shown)


SamSaffron:
This looks pretty urgent to merge, @majakomel can you review first thing today and maybe merge in if it looks good (and only needs minor improvements)

tgxworld:
@SamSaffron  @techAPJ  I don't think we want to be using hijack here. Hijack is based off an in-memory queue so the bulk invite can magically disappear if the app happens to restart when the user submits the request.

tgxworld:
I just read through the code, why do we need to use hijack here when we're already processing the bulk of the work in Sidekiq?

techAPJ:
@tgxworld don't we want to read and process the CSV file in a Hijack block? I am seeing the [similar pattern](https://github.com/discourse/discourse/blob/master/app/controllers/tags_controller.rb#L129-L150) in Tags controller where we bulk uploads tags via a CSV file.

SamSaffron:
mixed on absolute need of hijack, but yeah if this CSV is say 100,000 rows
long how long is it going to take to queue the jobs, if less than 5 seconds
I think you can bypass.

tgxworld:
We need to think about what happens if the server is restarted while the job in the hijack queue is running. I don’t like the ideal where things can randomly disappear.

SamSaffron:
same exactly as restarting a unicorn worker mid request... nothing
special happens, stuff half happens by design.

SamSaffron:
Just to clarify here... hijack runs in the same web worker as the original. The purpose of hijack it to make a unicorn worker ... slightly.. multithreaded. It basically gives it one more thread. 

If we have a slow request like an upload or long process of file, its nice to free up the unicorn worker to serve other requests while we are working through something slow. 

In this case it may not be needed, I am not sure, depends how long it takes to work through a large file. But the difference between hijack vs not hijack is just difference between allowing the unicorn worker to serve one more request or not. 

Hence I am fine to merge this in with hijack.

tgxworld:
With the standard docker install and even on our hosting, we don't drop unicorn requests on restart. The same cannot be said for the hijack queue.

tgxworld:
> but yeah if this CSV is say 100,000 rows

Do we also need to add a guard for this? Some one can easily submit a CSV with a whole ton of rows and it'll end up blocking either the unicorn or the hijack queue. We do have a guard in nginx for max upload size so there is some safety.

SamSaffron:
a reboot will drop a request there are plenty cases where it can happen. we can not design processes that can not recover if they are "half done" this is the exact reason we run post creator in transactions and so on. 

Odds are low a half invite will happen, I think we should totally amend our restart process to at least process the current request in the hijack queue (up to say 3 seconds)

That said ... yeah .. what happens if a giant invite csv fails half way, can it recover cleanly?

tgxworld:
> That said ... yeah .. what happens if a giant invite csv fails half way, can it recover cleanly?

It can't with this change since we no longer write the uploaded file to disk. @techAPJ Is there more context for this change?

SamSaffron:
> It can't with this change since we no longer write the uploaded file to disk. @techAPJ Is there more context for this change?

Why not, I just invited Sam... how about I don't invite him again? 

Context is that we don't want to shuffle this file to s3.

tgxworld:
That isn't what the code is doing... If we're reading the file halfway and it fails, there isn't a way to say read the file again. The request fails and the uploaded file is lost somewhere in tmp files. 

Also, this change pushes a much larger payload into Redis for serialization.

tgxworld:
Just thinking out loud but can't we process the file on the client side instead and chunk the rows into smaller requests?

SamSaffron:
One second... previously if you restarted discourse just as an invite was going on, at best sidekiq would re-queue the job and repeat half the invites. 

The only material change here is that we are serializing a bunch more data to redis, so as long as we have some guards here, max rows, max chars per row... I think this is fine.

tgxworld:
> The only material change here is that we are serializing a bunch more data to redis

I don't think that is the only change here. Previously as long as an admin completes the file upload, we can be fairly sure that the job has been en-queued and it'll run at some point. With the hijack queue and reading the file in memory the same cannot be said. We can encounter an error while reading through the CSV file and the admin will have to re-upload the file. If the hijack queue is stuck or if the app server is restarted before the job in the hijack queue has been processed, the bulk invite will not be processed and no indication is provided to the user.

SamSaffron:
That’s the thing if stuff is stuck the admin will see a stuck upload with
an hour glass, same as admin sees today when uploading on a 28.8 k baud
modem

tgxworld:
I don't understand, doesn't hijack return a response first? So on the client side it'll show as the upload has completed.

SamSaffron:
Ahhh you are misunderstanding how hijack works, you only get the 200 back
when absolutely everything is done

tgxworld:
Ah icic, it that is the case this makes sense. We’ll just need some guards here to ensure it doesn’t block the hijack queue for too Long.

techAPJ:
>  lets have some upper sane limit on CSV size... max 50,000 rows or something, if you want to invite more ... make more files ...

Makes sense. I'll add a check for this.

PR #7682, opened 2019-06-03
FIX: fix select badge as user title on badge page

majakomel:


SamSaffron:
Looks good to me, just be sure to double check on meta

PR #7687, opened 2019-06-04
FIX: add support for custom/plugin notification title attribute

majakomel:


app/assets/javascripts/discourse/widgets/notification-item.js.es6
```diff
@@ -160,7 +160,12 @@ createWidget("notification-item", {
     let title;
 
     if (notificationName) {
-      title = I18n.t(`notifications.titles.${notificationName}`);
+      title =
+        notificationName === "custom"
```
ZogStriP: Nested ternaries are hard to understand. Can you re-write them using standard `if`/`else` for an easier reading?

PR #7691, opened 2019-06-04
FIX: Prevent deadlock

danielwaterworth:
Before the locking here was added, replying to a post and liking a post
at roughly the same time could cause a deadlock.

Liking a post grabs an update lock on the post and then on the topic (to
update like counts).

We now lock the replied to post before getting the topic lock so that we
can update the replied to post later without causing a deadlock.

### DIFF
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -272,8 +272,18 @@ class PostCreator
   def self.set_reply_info(post)
     return unless post.reply_to_post_number.present?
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
This sounds good to me... just to clarify ... `.lock` wacks an extra ` FOR UPDATE` to a select statement.

danielwaterworth:
Yes, exactly.

PR #7695, opened 2019-06-04
REFACTOR: safari-hacks lib

jjaffeux:


jjaffeux:
@pmusaraj this one is kind of tricky, could you test/review it please?

pmusaraj:
@jjaffeux I tested this in the iOS simulator and on my iPhone, and after focusing on the input, I can't minimize the composer, and I can't toggle the toolbar either. 

Also, I see that initially this was added as a workaround for an iOS 8 bug. Maybe it has since been fixed? According to https://stackoverflow.com/a/44506266 it has. (A quick test disabling `positioningWorkaround` brought up no bug for me, but it was a very quick test.)

jjaffeux:
cool will dig this tomorrow. We are definitely supporting only ios10+ so we should investigate if we can drop all (or parts) of this. :+1:

SamSaffron:
cool... super duper scared about this change so I am adding a 2.4 tag on it so we can look at it then. 

also I wonder if we new viewport apis mean we can just redo this hack in a far simpler way and just revisit for ios 13? https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API

jjaffeux:
Closing this for now, as I will probably go from scratch

PR #7696, opened 2019-06-04
DEV: simpler spec for wayback machine crawler layout

majakomel:
follow-up on 42809f4d

### DIFF
--- a/spec/requests/topics_controller_spec.rb
+++ b/spec/requests/topics_controller_spec.rb
@@ -2677,7 +2677,6 @@ RSpec.describe TopicsController do
   end
 
   describe "crawler" do
-
     context "when not a crawler" do
       it "renders with the application layout" do
         get topic.url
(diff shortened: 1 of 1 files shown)


spec/requests/topics_controller_spec.rb
```diff
@@ -2711,40 +2710,44 @@ def invite_group(topic, expected_status)
...
-          body = response.body
+        expect(response.headers['Last-Modified']).to eq(page1_time.httpdate)
 
-          expect(response.headers['Last-Modified']).to eq(page2_time.httpdate)
+        get topic.url + "?page=2", env: { "HTTP_USER_AGENT" => user_agent}
+        body = response.body
+
+        expect(response.headers['Last-Modified']).to eq(page2_time.httpdate)
+
+        expect(body).to include('<link rel="prev" href="' + topic.relative_url)
```
tgxworld: We have https://github.com/kucaahbe/rspec-html-matchers available in the test env so you might want to have a look at that.

PR #7705, opened 2019-06-05
fix failing spec following 6d8eb9c144

jjaffeux:


gschlager:
https://github.com/discourse/discourse/commit/a77d8bee68219c8710bbe81e3860012c1a0e4b64 should fix it

PR #7707, opened 2019-06-05
FIX: On tag change notify only users watching the tag.

nbianca:


app/jobs/regular/notify_tag_change.rb
```diff
@@ -9,7 +9,6 @@ def execute(args)
 
       if post&.topic&.visible?
         post_alerter = PostAlerter.new
-        post_alerter.notify_post_users(post, User.where(id: args[:notified_user_ids]))
```
SamSaffron: hmm so why are we passing the job `notified_user_ids` if we no longer need it?

lib/post_revisor.rb
```diff
@@ -95,7 +95,7 @@ def self.track_topic_field(field, &block)
         DB.after_commit do
           post = tc.topic.ordered_posts.first
           notified_user_ids = [post.user_id, post.last_editor_id].uniq
```
SamSaffron: also this line :)

SamSaffron:
hold off on merging responded on dev

PR #7708, opened 2019-06-05
DEV: refactors select-kit helper to prevent it to leak into global state

jjaffeux:


SamSaffron:
I am fine for test only changes like this to be merged at will once you feel comfortable with it.

jjaffeux:
> I am fine for test only changes like this to be merged at will once you feel comfortable with it.

To be clear about these commits they are also memory related (but yes only during tests), as all these old helpers are defined on Window

PR #7711, opened 2019-06-05
FIX: use user template for showing hidden profile instead

majakomel:
Fixes issue with back button on hidden profiles https://meta.discourse.org/t/going-back-to-previous-internal-page-after-404/104067


SamSaffron:
I think @eviltrout fixed this separately, should we close this?

eviltrout:
I confirmed the bug listed on meta is fixed. I'm going to close for now and we can revisit if something is still broken.

PR #7713, opened 2019-06-06
FEATURE: Update pull hotlinked images to use `Upload#short_url`.

tgxworld:


spec/jobs/pull_hotlinked_images_spec.rb
```diff
@@ -139,12 +206,20 @@
         <a href='#{url}'><img src='#{large_image_url}'></a>
         BODY
 
-        Jobs::ProcessPost.new.execute(post_id: post.id)
-        Jobs::PullHotlinkedImages.new.execute(post_id: post.id)
-        Jobs::ProcessPost.new.execute(post_id: post.id)
-        Jobs::PullHotlinkedImages.new.execute(post_id: post.id)
+        2.times do
```
eviltrout: Question - if it's the same amount of lines as repeating them, what is the advantage here?

tgxworld: There are less characters to type and also less lines to update when the expected arguments change.

PR #7720, opened 2019-06-06
DEV: test badge titles presence on badge page

majakomel:
follow-up on 2863e7c1


SamSaffron:
looks like travis is having trouble with this.

PR #7721, opened 2019-06-06
FEATURE: in:tagged and in:untagged advanced search filters

joshmoore:
Similar to in:solved or in:unsolved, the filters check for an
existence of the topic_id in the topic_tags table:

 * `in:tagged` means that *at least one tag* exists for a topic
 * `in:untagged` means that *no tag* exists for a topic

Screen Shot 2019-06-28 at 08 49 43
(_Running from `boot_dev --init` with this branch_)

see: https://meta.discourse.org/t/how-to-search-filter-untagged-topics/119641/2


SamSaffron:
Totally support this change, but we got to add some tests here.

joshmoore:
Thanks, @SamSaffron. I had starting taking a look into that. Is it primarily or even just https://github.com/discourse/discourse/blob/master/test/javascripts/acceptance/search-full-test.js.es6 that needs adding to?

SamSaffron:
oh only the ruby part @joshmoore  have a look at search_spec no need to add js tests for this.

joshmoore:
Before:

```
$ RAILS_ENV=test ./bin/docker/bundle exec rspec spec/components/search_spec.rb

Randomized with seed 39106
.........................................................................................

Finished in 19.1 seconds (files took 2.88 seconds to load)
89 examples, 0 failures

Randomized with seed 39106
```

After:

```
$ RAILS_ENV=test ./bin/docker/bundle exec rspec spec/components/search_spec.rb

Randomized with seed 36721
...........................................................................................

Finished in 23.04 seconds (files took 3.09 seconds to load)
91 examples, 0 failures

Randomized with seed 36721
```

joshmoore:
It's less clear to me if a3c8f503e0b3ec2d7f6923fa5e759104ddf74ec2 can be accepted as is. Feel free to either force push it away (or to have me do so).

SamSaffron:
This is definitely getting there, but I would like the advanced section de-cluttered a bit:

image

We already have a natural spot for people to pick: 

```
[any tag]
[untagged]
[tag name1]
....
```

I guess on click then select kit should add those 2 options on top? @jjaffeux  any specific recs here? 

@joshmoore  as it is a "visibly changing stuff" PR be sure to always add a screenshot or two to the comments cause it helps people figure out what happened to the UX here. 

I am pushing this off to 2.4 (this basically means I can not merge it in for 2 weeks, but we will definitely get it merged very early here in 2.4)

Thanks heaps for this work @joshmoore !

joshmoore:
> I guess on click then select kit should add those 2 options on top?

Hmmm... would definitely need more input to know the preference here. One option would be to have the "Tagged" section not usable if  "in:untagged" is chosen. The same could hold for "in:tagged" but it's less technically wrong.

Alternatively, "Any tag" and "No tag" could be added beside "All the above tags".

Finally, another option would be to just list this in some as yet non-extant help section. I started adding a section to discourse_api_docs/swagger.yml but that won't help the general user.

> be sure to always add a screenshot
:+1:

> "I am pushing this off to 2.4

If I were to split off the first two commits into a separate PR, could they be considered earlier?

SamSaffron:
@joshmoore I definitely think it is worth splitting this into 2 PRs now. One for the server support another for the client. 

I feel on the server side this change is straightforward and I am happy to merge it today. The client side needs a bunch of refinements imo  

cc @jjaffeux

jjaffeux:
Yes split it please, it's unclear to me reading the PR what is the actual UI output. And there's almost no js code so we should just merge server side first.

joshmoore:
Thanks, @SamSaffron and @jjaffeux. I've added a screenshot for the UI side of things. I'll create a slimmer server-side PR now.

joshmoore:
Please see: https://github.com/discourse/discourse/pull/7822

jjaffeux:
Thx.

I'm unsure. What we have:

- Should have at least a tag
- Should have no tags
- Should have these tags
- should have ALL these tags

I think the UI we have ATM is correct but we should make sure it doesn't end up in weird states, eg: one menu has tags, the other has `are untagged`.

One alternative is to remove options from dropdown, and to have a radio list under "Tagged":

Tagged:
[ potential list of tags]
( ) Any tag
( ) All the above tags
( ) No tags
( ) None of the above tags (worth to add ???)

@awesomerobot what do you think ?

joshmoore:
> and to have a radio list under "Tagged":

In terms of correctness, I agree. I hesitated because of the difficulty of getting all the across to the user in the space that's currently allocated.

Since the right hand options really just end up filling in the search bar, I found it straight-forward enough to deal with the query:

```
in:tagged and un:tagged
```

manually. i.e. "no results are found, duh, that doesn't make sense". Of course, the more options you add in, the trickier it gets.

SamSaffron:
I still feel everything "tags" belongs in the tags combo, really uneasy mixing this about in 2 places.

ZogStriP:
Any ideas here @awesomerobot?

SamSaffron:
@joshmoore can you do another PR that only includes the changes to search.rb (plus spec) then at least we can have the base functionality merged and later followup on the UX work?

SamSaffron:
Oh looks like we had a PR ages ago that merged the server part of this. 

Let's close this for now and revisit it later in case we want to add UI.

PR #7722, opened 2019-06-06
FIX: Permit new tags when allow_global_tags true.

udan11:


SamSaffron:
This looks right to me, thanks!

PR #7723, opened 2019-06-06
UX: Hide "Create Tag" option if user cannot create tag.

udan11:


app/serializers/basic_category_serializer.rb
```diff
@@ -65,4 +68,28 @@ def custom_fields
   def include_custom_fields?
     custom_fields.present?
   end
+
+  def include_allowed_tags?
+    SiteSetting.tagging_enabled
+  end
+
+  def allowed_tags
+    object.tags.pluck(:name)
```
SamSaffron: Is this adding an N+1 to the topic list? very concerned about a change here.

ZogStriP:
Any news about this @udan11?

SamSaffron:
I approve of the change but am uneasy with the implementation. 

So we really need to carry 3 flavors of category serialization? I worry here that we can introduce code drift this way even with the mixin.

udan11:
@SamSaffron, I could use the `CategorySerializer`, which would prevent future code drift at the cost of having bigger payloads.

PR #7724, opened 2019-06-06
FEATURE: Trigger Discourse events from authenticators.

udan11:


SamSaffron:
This feels a bit fragile to me people can forget about this. 

Maybe add a wrapper object to do the revoke and always call via it? 

https://github.com/discourse/discourse/blob/a0474a07745596a107384b9feb0d5bf1d408605a/app/controllers/users_controller.rb#L1166-L1166

SamSaffron:
Yes I prefer this, I know there is an element of fragility here if we add a new route or do stuff from rake tasks, but am willing to assume it, cause the actual surface area here is much smaller like this!

PR #7725, opened 2019-06-07
REFACTOR: Broke "notification-item" widget into multiple classes

danielwaterworth:


tgxworld:
This is something not related to the PR but we prefer commit messages to be written in present tense. [This answer on SO](https://stackoverflow.com/a/3580764/2531506) provides a good background about the rational behind it.

SamSaffron:
I very much support this cleanup! It makes it much more flexible for plugins. @eviltrout  can you review? 

Note I think it is too large and risky to merge this week, we are going to have to wait for 2.4 for now

app/assets/javascripts/discourse/widgets/widget.js.es6
```diff
@@ -270,18 +196,26 @@ export default class Widget {
     }
   }
 
-  attach(widgetName, attrs, opts) {
-    let WidgetClass = _registry[widgetName];
+  attach(widgetNames, attrs, opts) {
```
eviltrout: I'm not sure I agree with this change. 

There are only two places in the commit that adds multiple widgets, and both are only adding two. It would be fewer lines of code to call `this.attach` twice than building an array and calling it this way, not to mention that `this.attach` has to do less code (checking if input is array or not, then looping through it.)

app/assets/javascripts/discourse/widgets/widget.js.es6
```diff
@@ -45,90 +43,14 @@ export function changeSetting(widgetName, settingName, newValue) {
...
-    const afterContents = applyDecorators(this, "after", attrs, state) || [];
-    contents = beforeContents.concat(contents).concat(afterContents);
-  }
-
-  return h(this.tagName || "div", properties, contents);
-}
-
-export function createWidget(name, opts) {
-  const result = class CustomWidget extends Widget {};
+export function createWidgetFrom(base, name, opts) {
```
eviltrout: This is a nice change :+1:

app/assets/javascripts/discourse/widgets/user-notifications-large.js.es6
```diff
@@ -12,8 +12,16 @@ createWidget("large-notification-item", {
   },
 
   html(attrs) {
+    const notificationName =
+      this.site.notificationLookup[attrs.notification_type];
+
+    const widgetNames = [
+      `${notificationName.replace(/_/g, '-')}-notification-item`,
```
eviltrout: Could use `Ember.string.dasherize` here.

app/assets/javascripts/discourse/widgets/widget.js.es6
danielwaterworth: This isn't adding multiple widgets, it's adding one widget using the first type that exists. So, for example, since `liked-notification-item` exists, it uses that, whereas there is no `watching-first-post-notification-item`, so the second option, `default-notification-item`, is used instead.

I could still change this back and throw a more specialized error and then, in those two places, do:

```javascript
try {
  this.attach(specialized_name, attrs);
} catch (e) {
  if (e isinstanceof WidgetNotFoundError) {
    this.attach("default-notification-item", attrs);
  } else {
    throw e;
  }
}
```

But I was trying to avoid that.

eviltrout: Oh that was not clear to me based on the `widgetNames` parameter! I would expect that to include them all, not the first one it finds.

Instead I'd recommend an option that can be used if the widget is not found. Something like `defaultWidgetName: 'default-notification-item'`

app/assets/javascripts/discourse/widgets/widget.js.es6
```diff
@@ -270,18 +196,31 @@ export default class Widget {
     }
   }
 
-  attach(widgetName, attrs, opts) {
-    let WidgetClass = _registry[widgetName];
+  attach(widgetName, attrs, opts, otherOpts) {
+    const widgetNamesToTry = [widgetName];
```
eviltrout: Rather than looping over an array of two, why not do `WidgetClass = _registry[widgetName] || _registry[opts.fallbackWidgetName]`? A lot less code.

app/assets/javascripts/discourse/widgets/widget.js.es6
```diff
@@ -291,7 +230,7 @@ export default class Widget {
       result.dirtyKeys = this.dirtyKeys;
       return result;
     } else {
-      throw new Error(`Couldn't find ${widgetName} factory`);
+      throw new Error(`Couldn't find ${widgetNames} factory`);
```
eviltrout: widgetNames is no longer defined.

app/assets/javascripts/discourse/widgets/widget.js.es6
danielwaterworth: The current version looks up the widget name in the register. Is that actually necessary?

danielwaterworth: I'm sorry, I wasn't clear at all. There are two registries. I'm investigating why looking up the widget name in `this.register` is necessary.

eviltrout: I am not sure if that is used anymore. I suspect it is leftover code and can be removed. Worth looking into!

danielwaterworth: There are several different widgets that don't use `createWidget` and so aren't registered in the widget registry. As things stand, this is necessary, but it wouldn't be difficult to fix. However, changing this might also break plugins.

eviltrout: Can you give me an example of one that doesn't use createWidget? I forget how that API works.

danielwaterworth: Have a look at `discourse/widgets/category-link.js.es6`, it's used in `discourse/widgets/search-menu-results.js.es6: 89 & 155`.

eviltrout: Oh right that makes sense and would indeed break backwards compatibility.

What I'd suggest then is creating a helper method that returns the class or null. Then you can call that an extra time if a backup is provided.

We could also remove the register by having an initializer find all widgets and put them in the registry when the app boots (or perhaps the first time a widget is rendered.)

SamSaffron:
@eviltrout can you merge this today?

eviltrout:
Done, let's try this out :)

PR #7727, opened 2019-06-07
HACK: Added commit approved notification type for discourse-code-review

danielwaterworth:


romanrizzi:
You can achieve the same outcome without touching core using this method:

https://github.com/discourse/discourse/blob/master/lib/plugin/instance.rb#L190

danielwaterworth:
@romanrizzi, the ID needs to be reserved so that other plugins don't create conflicting notifications.

SamSaffron:
I don't really see this as HACK, this is best practice ... :)

danielwaterworth:
@sam, I see this as a hack because a plugin shouldn't have to modify core. Ideally, I think getting rid of the notification type ids and just using the names would be better.

PR #7736, opened 2019-06-10
Add base62 sha1 to cooked data attribute

tgxworld:


SamSaffron:
Yeah we need this change! thanks!

PR #7737, opened 2019-06-10
PERF: Use already loaded post when quoting or opening draft.

nbianca:


app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -254,47 +254,52 @@ export default Ember.Controller.extend(bufferedProperty("model"), {
...
-          const quotedText = Quote.build(post, buffer);
-
-          // If we can't create a post, delegate to reply as new topic
-          if (!viewOpen && !this.get("model.details.can_create_post")) {
-            this.send("replyAsNewTopic", post, quotedText);
-            return;
-          }
+      const posts = this.get("model.postStream.posts");
+      const loadedPost = posts && posts.findBy("id", postId);
+      const promise = loadedPost
```
eviltrout: Question: did you consider putting this logic inside `loadPost`? The stream knows if the post is loaded and could return it there, keeping the code here the same.

nbianca: I did consider this, but there are two reasons why I did not want to do it:

1. It is used in one more place and I did not want to risk breaking anything there, given that we are so close to a release.

2. It would be different than the other functions called `load...Post...`, which are all loading data from server.

eviltrout: Thanks that sounds reasonable!

PR #7738, opened 2019-06-10
FIX: Add 'deleted' to the list of status filters

romanrizzi:


romanrizzi:
@eviltrout Looks like the `deleted` state wasn't being shown. I added it to the list.

PR #7739, opened 2019-06-10
FIX: Mark ignored posts as 'read', if last visible post is read

davidtaylorhq:


SamSaffron:
I am comfortable merging this, but before we go ahead. 

Is there any ways gaps could have a bit more information in the object model that is passed in? Eg: I am an ignore type gap? Is there any way to have the server send this down without adding any extra queries?

davidtaylorhq:
> Is there any ways gaps could have a bit more information in the object model that is passed in?

Maybe... but

> without adding any extra queries?

No, I don't think so. Right now we do something like in a single query:

```
SELECT * from posts
WHERE {{not-ignored-by-current-user}}
AND {{not-deleted}}
AND {{in-summary-if-enabled}}
AND {{match-username-filter-if-enabled}}
```
Then to find the gaps we do
```
SELECT id from posts
```
and compare the list of integer IDs in Ruby. 

We have to bear in mind that sometimes "gaps" can be hundreds/thousands of posts long, so we need to be careful not to regress performance on such a critical route. 

I think long-term (after this release) we should refactor ignored posts to make it JS only. That will remove a lot of these edge cases.

SamSaffron:
Makes sense, merge away then and deploy in a few places

PR #7745, opened 2019-06-11
FIX: clean up cooked group bio when bio is removed

majakomel:
https://meta.discourse.org/t/group-bio-cant-be-deleted/119777

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -209,8 +209,10 @@ class Group < ActiveRecord::Base
   end
 
   def cook_bio
-    if !self.bio_raw.blank?
+    if self.bio_raw.present?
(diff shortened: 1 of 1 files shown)


tgxworld:
@majakomel We can clear the `bio_cooked` in a callback on save instead of doing it at run time all the time.

SamSaffron:
@tgxworld I am confused `cook_bio` is called from `before_save`

merging this, looks good to me

tgxworld:
O I thought this was the serializer, my bad. @SamSaffron Is there a way you determine when a test is required vs when it isn't?

SamSaffron:
These days on local I just break code around where I made the change and
run the full test suite, then at least I can tell the code is covered

In this case we actually had a faulty spec that this fix corrected thus
broke the build

PR #7756, opened 2019-06-12
FEATURE: SKIP_DB_AND_REDIS env to enable bootstrapping without a DB

SamSaffron:
Sometimes we would like to create a base image without any DB access, this
assists in creating custom base images with custom plugins that already
includes `public/assets`

Following this change set you can run:

```
SPROCKETS_CONCURRENT=1 DONT_PRECOMPILE_CSS=1 SKIP_DB_AND_REDIS=1 RAILS_ENV=production bin/rake assets:precompile
```

Then it is straight forward to create a base image without needing a DB or
Redis.

### DIFF
--- a/app/models/emoji.rb
+++ b/app/models/emoji.rb
@@ -89,10 +89,12 @@ class Emoji
   def self.load_custom
     result = []
 
-    CustomEmoji.includes(:upload).order(:name).each do |emoji|
-      result << Emoji.new.tap do |e|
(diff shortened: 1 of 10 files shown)


SamSaffron:
Don not merge quite yet, I want to sit on this for a day.

SamSaffron:
OK having spent time thinking about this and reorganizing I think it is safe to merge.

PR #7757, opened 2019-06-12
DEV: improve teardown of initialisers to cleanly remove messagebus

jjaffeux:


test/javascripts/helpers/qunit-helpers.js.es6
```diff
@@ -128,6 +128,16 @@ export function acceptance(name, options) {
       resetPostCookedDecorators();
       resetOneboxCache();
       resetCustomPostMessageCallbacks();
+
+      Discourse._runInitializer("instanceInitializers", function(
+        initializerName,
+        initializer
+      ) {
+        if (initializer && initializer.teardown) {
+          initializer.teardown(Discourse.__container__);
```
eviltrout: Is this working? Because I did a test locally and found that Discourse.__container__ here was not the same container that the initializer was run with.

jjaffeux:
@eviltrout given your messagebus callbacks fix I have reduced the PR to only pass container, as I will probably need it:

TO test it I created a fake initializer:
```
export default {
  name: "foo",

  after: "message-bus",

  initialize(container) {
    console.log(container, Ember.guidFor(container));
  },

  teardown(container) {
    console.log(container, Ember.guidFor(container));
  }
};
```

It will print the same object guid, so it seems to me it's working correctly.

eviltrout:
@jjaffeux something weird is going on then, because I ran a similar test on my codebase and got different guids for each:

```
nit ember1673
discourse/initializers/message-bus:85 remove ember2073
discourse/initializers/message-bus:27 init ember2321
discourse/initializers/message-bus:85 remove ember2617
discourse/initializers/message-bus:27 init ember2867
discourse/initializers/message-bus:85 remove ember3167
```

The only difference is I added this to the message-bus initializer rather than creating a new one.

jjaffeux:
@eviltrout Ok when I run only ONE acceptance module it works, but when I run on multiple modules I have your behavior, so somehow, this using a detached one or something like this.

eviltrout:
@jjaffeux I'll investigate a little more today. Maybe I can figure something out.

eviltrout:
I followed up with: https://github.com/discourse/discourse/pull/7769

PR #7758, opened 2019-06-12
DEV: removes dead code

jjaffeux:


jjaffeux:
@eviltrout doesnt look used anywhere and is leaking in global context

jjaffeux:
> redirectedTo

yes but it's DiscourseURL.redirectedTo not _DiscourseURL.redirectedTo we are directly importing `import DiscourseURL from "discourse/lib/url";`

PR #7759, opened 2019-06-12
Fix minor typos in server.en.yml

AhmadFCheema:
Additional possible typos:

1. While defining image sizes, sometimes the "×" character is used (512 × 512) and at other times the normal "x" (512 x 512). Should these be changed to one type?

2. Also, in the "Terms of Service" section, every full stop within a sentence is followed by 2 spaces. Is there a reason for this peculiarity or should they be replaced with 1 space?

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -354,7 +354,7 @@ en:
   groups:
     success:
       bulk_add:
-        one: "%{count} user have been added to the group."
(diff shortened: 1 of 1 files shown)


eviltrout:
Thanks these are all decent changes. As for your questions, I think the raised x character looks nicer, and we should remove the two spaces.

PR #7760, opened 2019-06-12
Remove extra trailing blank spaces from server.en.yml

AhmadFCheema:
As noted in [this comment](https://github.com/discourse/discourse/pull/7759#issuecomment-501267443).

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -2936,7 +2936,7 @@ en:
 
         This is an automated message from %{site_name} to let you know you that your account has been placed on hold due to a large number of community flags.
 
-        As a precautionary measure, your new account has been silenced from creating new replies or topics until a staff member can review your account.  We apologize for the inconvenience.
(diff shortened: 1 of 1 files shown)


coding-horror:
I support this but I think it will break translations @SamSaffron ? So maybe this has to wait until 2.4 beta?

eviltrout:
Not I merged some earlier translations today because I missed the note about holding off for 2.4. @SamSaffron should I revert those?

AhmadFCheema:
If it's relevant, the change affects just three translation strings and there too the translators would have to just re-save the previously translated ones.

SamSaffron:
Definitely can not happen for 2.3, got to wait till next week to merge.

gschlager:
We need to ensure that we don't lose existing translations on Transifex. Some of these minor changes affect quite important strings. I'm assigning this to myself.

gschlager:
Thanks! Merged in https://github.com/discourse/discourse/commit/bfbd97d3b52ac72434102ab80e4d6bfa33b7e7ad

PR #7764, opened 2019-06-12
replace subfolder on cdn url conversion between general cdn and s3

featheredtoast:
When both a cdn URL and an s3 cdn URL defined, subfolder paths were leaking
through to the s3 cdn URL. If we are replacing the cdn url with the s3_cdn url,
we also need to make sure that the subpath is removed as well, as it appears in
the original cdn url.

The test should give a fairly good gist of the situations - in subfolder
situations where s3_cdn and a cdn is defined:
`asset_path` returns the asset with a subfolder, in the form `{cdn_url}/{subfolder}/{asset_path}`

Currently this is being replaced to `{s3_cdn_url}/{subfolder}/{asset_path}`
I am proposing we change this to: `{s3_cdn_url}/{asset_path}` as it seems like
for s3_cdn urls we should not be carrying around app subfolder pathing anywhere
we are looking up s3 paths.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # frozen_string_literal: true
 require 'current_user'
 require 'canonical_url'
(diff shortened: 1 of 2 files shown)


eviltrout:
@SamSaffron can you review this? It would help with one of our big customers.

PR #7767, opened 2019-06-13
UX: Rename "Keep Post" to "Keep Post Hidden" when hidden

eviltrout:
This is based on this feedback:
https://meta.discourse.org/t/category-group-review-moderation/116478/19

When a post is hidden this makes the operation much more clear.

image

@SamSaffron I made this a PR because I think translations are frozen now and this introduces a new one?

### DIFF
--- a/app/models/reviewable.rb
+++ b/app/models/reviewable.rb
@@ -41,6 +41,11 @@ class Reviewable < ActiveRecord::Base
     Jobs.enqueue(:notify_reviewable, reviewable_id: self.id) if pending?
   end
 
+  # Can be used if several actions are equivalent
+  def self.action_aliases
+    {}
+  end
+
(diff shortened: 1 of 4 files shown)


SamSaffron:
I am so mixed here, your call. 

I feel this change heavily improves usability of this feature, kind of thinking it makes sense to sneak it in, but yeah we are officially frozen on string.

eviltrout:
@SamSaffron Let's wait since it's quite similar to what we had before this release, and the existing behavior does work.

SamSaffron:
looks good to merge to me pending a rebase

SamSaffron:
@eviltrout looks like we need a rebase here.

eviltrout:
Merged now thanks @SamSaffron

PR #7768, opened 2019-06-13
FIX: Show poll voters in Oneboxed posts.

nbianca:


SamSaffron:
Before / after screenshot?

nbianca:
It looks pretty much the way as before, but now it is interactive and properly shows the results.
Before
Screenshot 2019-06-14 at 12 09 15

After
Screenshot 2019-06-14 at 12 10 28

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -161,7 +161,13 @@ export default class PostCooked {
 
       ajax(`/posts/by_number/${topicId}/${postId}`)
         .then(result => {
+          const post = this.decoratorHelper.getModel();
+          const quotedPosts = post.quoted || {};
+          quotedPosts[result.id] = result;
+          post.set("quoted", quotedPosts);
+
           const div = $("<div class='expanded-quote'></div>");
+          div.data("post-id", result.id);
```
jjaffeux: all of this can be easily replaced by vanilla js:

```
const div = document.createElement("div");
div.classList.add("expanded-quote");
div.setAttribute("data-post-id", result.id);
div.innerHTML =  result.cooked;
```

You will have to $(div) in forEach loop though, so it's only a partial step for now.

plugins/poll/assets/javascripts/initializers/extend-for-poll.js.es6
```diff
@@ -55,37 +55,44 @@ function initializePolls(api) {
...
     api.preventCloak(post.id);
-    const votes = post.get("polls_votes") || {};
-
     post.pollsChanged();
 
-    const polls = post.get("pollsObject");
-    if (!polls) {
-      return;
-    }
+    const polls = post.get("pollsObject") || {};
```
jjaffeux:
```suggestion
    const polls = post.pollsObject || {};
```

plugins/poll/assets/javascripts/initializers/extend-for-poll.js.es6
```diff
@@ -55,37 +55,44 @@ function initializePolls(api) {
...
+    const votes = post.get("polls_votes") || {};
```
jjaffeux:
```suggestion
    const votes = post.polls_votes || {};
```

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
nbianca: The `div` object is also used in multiple places below which make it hard to use vanilla JavaScript.

SamSaffron:
@jjaffeux  / @nbianca  is this good to merge?

app/assets/javascripts/discourse/widgets/post-cooked.js.es6
```diff
@@ -165,7 +165,13 @@ export default class PostCooked {
 
       ajax(`/posts/by_number/${topicId}/${postId}`)
         .then(result => {
+          const post = this.decoratorHelper.getModel();
+          const quotedPosts = post.quoted || {};
+          quotedPosts[result.id] = result;
+          post.set("quoted", quotedPosts);
```
SamSaffron: Do we need to use #set here @jjaffeux ? or does post.quoted = quotedPosts work?

SamSaffron:
@nbianca feel free to merge once we figure out if the .set is needed.

PR #7770, opened 2019-06-14
FEATURE: Add plugin outlets to login/create-account modals

tshenry:
This will add basic plugin outlets just above the modal-body of the login and create-account modals.


SamSaffron:
seems safe to me

PR #7773, opened 2019-06-16
FIX: keep read position when opening a modal on topic page

majakomel:


app/assets/javascripts/discourse/lib/show-modal.js.es6
```diff
@@ -12,7 +12,7 @@ export default function(name, opts) {
   //
   // Also, very little value in showing main outlet and header on iOS
```
majakomel: This hack was added because of an iOS bug that was resolved in 11.3 update. It fixed an issue with focus input being displaced (https://meta.discourse.org/t/mobile-safari-login-woes/73099) but it causes scroll position being lost when opening a modal on topic page (https://meta.discourse.org/t/losing-scroll-position-in-a-topic-after-flagging/90183).

app/assets/javascripts/discourse/lib/show-modal.js.es6
```diff
@@ -12,7 +12,7 @@ export default function(name, opts) {
...
   // anyway, so just hide it.
-  if (isAppleDevice()) {
+  if (opts.checkIfAppleDevice && isAppleDevice()) {
```
SamSaffron: I am confused then, do we need to even keep this "isAppleDevice" check anymore? Why not simply drop it?

majakomel: No need to keep it, I just wasn't sure how long we need to carry this kind of fixes for old iOS versions.

SamSaffron:
OK I am pulling this in, I want to see how it goes, not sure if we need to backport it though.

PR #7775, opened 2019-06-17
UX: show like icon next to count on archived posts

techAPJ:
https://meta.discourse.org/t/missing-icon-for-likes-on-archived-topics/120452/

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-menu.js.es6
@@ -61,8 +61,14 @@ function likeCount(attrs) {
         ? "post.has_likes_title_only_you"
         : "post.has_likes_title_you"
       : "post.has_likes_title";
-    const icon = attrs.yours ? "d-liked" : "";
+    let icon = attrs.yours ? "d-liked" : "";
+    let addContainer = attrs.yours;
     const additionalClass = attrs.yours ? "my-likes" : "regular-likes";
+
+    if (!attrs.showLike) {
+      icon = attrs.yours ? "d-liked" : "d-unliked";
+      addContainer = true;
+    }
 
     return {
       action: "toggleWhoLiked",
@@ -71,7 +77,7 @@ function likeCount(attrs) {
       contents: count,
       icon,
       iconRight: true,
-      addContainer: attrs.yours,
+      addContainer,
       titleOptions: { count: attrs.liked ? count - 1 : count }
     };
   }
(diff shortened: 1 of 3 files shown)


SamSaffron:
not following the CSS change, why is it needed?

techAPJ:
CSS changes are required (on both desktop and mobile) because the `regular-likes` class is currently not styled to have proper color and margins as it was not being used.

Before CSS change:

Screen Shot 2019-06-18 at 11 55 34

After CSS change:


Screen Shot 2019-06-18 at 11 56 01

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -61,17 +61,23 @@ function likeCount(attrs) {
         ? "post.has_likes_title_only_you"
         : "post.has_likes_title_you"
       : "post.has_likes_title";
-    const icon = attrs.yours ? "d-liked" : "";
+    let icon = attrs.yours ? "d-liked" : "";
+    let addContainer = attrs.yours;
     const additionalClass = attrs.yours ? "my-likes" : "regular-likes";
 
+    if (!attrs.showLike) {
+      icon = attrs.yours ? "d-liked" : "d-unliked";
```
eviltrout: I'm a little confused as to why we only add `d-unliked` when showing likes is disabled?

techAPJ: Currently when showing likes is disabled we are displaying just the like count without any icon and that is confusing. With this change we show ❤️ icon next to count to indicate that the count displayed is for likes.

For example, here is how the Like count looked on archived topics for anons before the change:

Screen Shot 2019-06-18 at 22 17 44

... after the change:

Screen Shot 2019-06-18 at 22 19 00

eviltrout: Okay I think I get it. It's just a little hard to follow currently. What if you *can* show likes but it's not yours? There's no icon?

techAPJ: When like is enabled (`showLike` is true) the like icon is shown via this codepath:

https://github.com/discourse/discourse/blob/7a9e85b1d7a33b2876248b3e2f566606b87a6b00/app/assets/javascripts/discourse/widgets/post-menu.js.es6#L82-L113

techAPJ:
@eviltrout Does this PR looks good for merge?

PR #7776, opened 2019-06-17
Make onbox headers RTL compliant

Ddorda:
Makes onebox title link to be compliant with RTL (Right To Left) languages.
The issue is also displayed here: https://meta.discourse.org/t/rtl-onebox-suffers-from-overlapping/120518

### DIFF
--- a/app/assets/stylesheets/common/base/onebox.scss
+++ b/app/assets/stylesheets/common/base/onebox.scss
@@ -90,6 +90,7 @@ a.loading-onebox {
     background: image-url("favicons/#{$image}.png") no-repeat;
     background-size: 16px 16px;
     padding-left: 20px;
+    padding-right: 20px;
   }
 }
 


Ddorda:
> Thanks for contributing this pull request! Could you please sign our CLA so we can review it? http://www.discourse.org/cla

Done.

SamSaffron:
@hnb-ku does this look good? if so please merge!

SamSaffron:
@OsamaSayegh  can you review and test on local (make sure this does not regress English) and merge?

OsamaSayegh:
Hmmm I can't repro the issue reported in the Meta topic:

It looks like this **without** this fix:
image

With this fix it looks like this:

image

Am I missing something?

Ddorda:
Yes, you're not in RTL so you're not supposed to experience the fix :)
The bug is described here: https://meta.discourse.org/t/rtl-onebox-suffers-from-overlapping/120518

OsamaSayegh:
Ok I had to enable the `support_mixed_text_direction` site setting to repro the problem. The root cause of the problem is the R2 gem flips the `padding-left` rule to `padding-right` when you switch to an RTL locale. I think an ideal fix would be to tell the gem to exclude this particular rule, but that doesn't seem to be possible.

This fix **doesn't** regress LTR locales, but it causes a small spacing problem (see my second screenshot above notice how the github link is misaligned) when you're in RTL AND don't have `support_mixed_text_direction` enabled.
We don't have a lot of freedom here because the favicon itself is not a DOM element, it's just a CSS background rule on the onebox header, so I think we can live with that for now @SamSaffron?

SamSaffron:
Something here is still not sitting right for me we only want this padding if `rtl` is set ... can we somehow target rtl here? In theory this adds some padding into titles that we do not want.

ZogStriP:
@Ddorda can you fix this PR?

ZogStriP:
I'm gonna close this PR for now. @Ddorda feel free to re-open.

PR #7779, opened 2019-06-19
Update server.zh_CN.yml

jasl:
`开发` means `development`, the raw sentence is `Is your community open to everyone`, `open` should be translated to `开放` or `公开`


keonjeo:
hi JiangJun, You are so careful.  lol

gschlager:

PR #7780, opened 2019-06-19
PERF: Use Oj for serializing JSON.

nbianca:


SamSaffron:
```
  1) PostCreator when a post is created when plugin is enabled should schedule a chat notification job
     Failure/Error:
       expect(job['at'])
         .to eq((Time.zone.now + SiteSetting.chat_integration_delay_seconds.seconds).to_f)
     
       expected: 1560934756.1195285
            got: 1560934756.119529
     
       (compared using ==)
     # ./plugins/discourse-chat-integration/spec/jobs/regular/notify_chats_spec.rb:29:in `block (4 levels) in <main>'
```

Plugin failure is legit here, it should be comparing time with the match_time thing.

SamSaffron:
@nbianca  can you merge this in first thing today? (you will need to fix the chat integration plugin spec)

PR #7781, opened 2019-06-19
FIX: Ensure topic exists before making a banner.

nbianca:


SamSaffron:
seems safe, thanks, maybe add a mega cheap spec to guardian that explains this does not blow up nil.

PR #7782, opened 2019-06-19
FEATURE: apply a small penalty to closed topics when searching

pmusaraj:


SamSaffron:
I am ok with this, no choice really but to fudge numbers ... may make sense though to pull all our magic numbers to the top of the file in consts.

PR #7783, opened 2019-06-19
feat(field): allow custom fields to have a FA icon

warlof:
I would beeing able to put a font awesome icon near my custom fields (ie: Steam ID, Battle Tag, etc...).

To do so, I think about adding an extra column to the table which will store the font awesome class.
If an icon has been set, it will be displayed on both admin list and user profile.

I'm not quite familiar with RoR environment, so I hope it's done properly.
At least, it seams to work on by dev environment ^^'

### DIFF
--- a/app/assets/javascripts/admin/components/admin-user-field-item.js.es6
+++ b/app/assets/javascripts/admin/components/admin-user-field-item.js.es6
@@ -75,7 +75,8 @@ export default Ember.Component.extend(bufferedProperty("userField"), {
         "required",
         "show_on_profile",
         "show_on_user_card",
-        "options"
+        "options",
+        "icon"
       );
 
       this.userField
--- a/app/assets/javascripts/admin/templates/components/admin-user-field-item.hbs
+++ b/app/assets/javascripts/admin/templates/components/admin-user-field-item.hbs
@@ -11,6 +11,10 @@
     {{input value=buffered.description class="user-field-desc" maxlength="255"}}
   {{/admin-form-row}}
 
+  {{#admin-form-row label="admin.user_fields.icon"}}
+    {{input value=buffered.icon class="user-field-icon" maxlength="255"}}
+  {{/admin-form-row}}
+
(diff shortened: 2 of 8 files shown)


SamSaffron:
Can you post some screenshots of how this would look?

warlof:
Hi @SamSaffron ,

Here you go :

User profile view with a custom field and its icon
2019-06-23_12h36_42

Admin backend with user fields list and related icons
2019-06-23_12h37_28

Admin backend with user field form
2019-06-23_12h37_37

SamSaffron:
there are a couple of things here

- I think, we need to tell the icon system to build assets for this thing no? the particular `fab-` may not be in our payload. 

- We need a bit more margin there on the user page. 

@coding-horror  how do you feel about this? Do you think it should be a core feature or should people simply be using theme components to do this. I am a bit mixed here. 

cc @pmusaraj

pmusaraj:
Yes, this needs two things to ensure icons are included when building assets: 

- expire the SvgSprite cache when saving a custom field
- add any icon names defined in these custom fields to `SvgSprite.all_icons`(see `group_icons` in svg_sprite.rb for an example)

I also think the icon would be best displayed to the left of the field label in the profile view.

coding-horror:
I trust whatever Penar thinks is best here

SamSaffron:
I have thought about this for a bit, I would prefer not to carry this for now. Do you need a hand figuring out how to achieve the exact same thing in a theme component @warlof  ? 

We need to deal with signup, decide on positioning and adding icons is not trivial. I am closing for now. But if you are stuck creating a theme component to achieve this exact use case please open a topic on meta.

warlof:
Hi @SamSaffron ,

I'll probably do as I have absolutely not clue of the way to do it since id/class seems to be dynamic :p

SamSaffron:
Oh a PR to give you more stable classnames is reasonable imo. you could
always use nth selectors today if you had to.

PR #7784, opened 2019-06-19
FEATURE: admin/user exports are compressed using the zip format

romanrizzi:


eviltrout:
We always prefer not to add a dependency if possible. So my questions for this PR are:

- Is there any way to do this without adding a gem?
- If not, how does this gem perform compared to the gzip approach? Are there other zip libraries and how did we choose this one?

SamSaffron:
I am pretty sure there is a missing translation change here ... cause our email says "hi this is how you decompress a gzip file" 

> Is there any way to do this without adding a gem?

Only way would be to ship `zip` in our base image and shell to it.  I thought about this but its a complex coordinated effort. I am certain that the gem will make these things about 10x slower, but there are so few user exports going on that this is at best a rounding problem. I do worry about pathological cases so I want this tested on simulated "enormous" data dump an enormous site would generate.

coding-horror:
Now you are seeing why we didn't do this in 2014, turns out "make a zip
file" is stupidly annoying on Linux..

eviltrout:
In that case I just want to be sure we are using the library in the best way possible and that it's not a footgun for performance.

romanrizzi:
> I am pretty sure there is a missing translation change here ... cause our email says "hi this is how you decompress a gzip file"

Yeah, this was still work in progress. I should have opened as a draft PR.

I'll do some synthetic benchmarks soon to see how it affects performance.

romanrizzi:
#### Some updates:

- Theme exports use `.zip` instead of `.gz`.
- Theme imports work with both `.zip`and `.gz`.
- We point to the `7-zip` page because it supports packing/unpacking zips and it's open source.

#### Benchmarks

I downloaded a CSV dataset (109 MB) from the internet and compared both compression strategies.

Running the following:

```ruby
require 'benchmark'
require 'zip'

input_path = '/Users/romeer/projects/discourse/bench/craigslistVehiclesFull.csv'

n = 1

Benchmark.bm do |benchmark|
  benchmark.report(".gz") do
    n.times do |n|
      system('gzip', '-5', '--force', '--keep', input_path)
    end
  end

  benchmark.report(".zip") do
    n.times do |n|
      compressed_file_path = "#{input_path}_#{n}.zip"

      Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
        zipfile.add('craigslistVehiclesFull.csv', input_path)

        zipfile.close
      end
    end
  end
end
```

With N=1:

```
      user     system      total        real
.gz  0.000113   0.000636   8.351566 (  8.368490)
.zip 11.463222   0.453336  11.916558 ( 11.934653)
```

Zip compression is ~42% slower.

With N=5:
```
      user     system      total        real
.gz  0.000453   0.002339  40.535668 ( 40.613858)
.zip 57.206713   2.180608  59.387321 ( 59.276257)
```
Zip comression is ~46% slower.

eviltrout:
@romanrizzi have you considered using streams? In other words have the CSV output a stream that is passed to ZIP for writing rather than writing an entire file then compressing that entire file?

How does that affect benchmarks?

romanrizzi:
I'm not sure if the benchmark will change since the library is already reading the file as a stream. On the other hand, saving the CSV to disk just to compress it seems like an unnecessary step.

I'll update the code and the benchmarks to see if there's an improvement. Keep in mind that this approach will only work for the CSV exporter.

romanrizzi:
```ruby
require 'benchmark'
require 'zip'

input_path = '/Users/romeer/projects/discourse/bench/craigslistVehiclesFull.csv'

n = 1

iter = File.foreach(input_path)

Benchmark.bm do |benchmark|
  benchmark.report(".zip") do
  n.times do |n|
  compressed_file_path = "#{input_path}.zip"

  Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
  zipfile.get_output_stream("craigslistVehiclesFull.csv") do |f|
  iter.each { |l| f << l }
  end
  end
  end
  end
end
```

```
  user system total real
.zip 13.809884 1.375531 15.185415 ( 15.179317)
```

I also tried using `File#readlines` instead of `foreach` to simulate a scenario where I have all the CSV contents loaded into memory:

```
  user system total real
.zip 12.640760 1.203374 13.844134 ( 13.903592)
```

For some reason, both versions are slower than saving to disk and then zipping 🤔.

SamSaffron:
> File#readlines 

That will still stream the input file. What you need to do to simulate is to read the file in entirety into memory (with File.read) and use a memory stream.

romanrizzi:
I see your point 😁 

Tried this approach:

```ruby
input_path = '/Users/romeer/projects/discourse/bench/craigslistVehiclesFull.csv'

stream = StringIO.new('')
File.open(input_path) do |f|
  stream << f.read
end

Benchmark.bm do |benchmark|
  benchmark.report(".zip  + read") do
    compressed_file_path = "#{input_path}.zip"

    Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
      zipfile.get_output_stream("craigslistVehiclesFull.csv") do |f|
        f << stream.read
      end
    end
  end

  benchmark.report(".zip  + readlines") do
    compressed_file_path = "#{input_path}.zip"

    Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
      zipfile.get_output_stream("craigslistVehiclesFull.csv") do |f|
        stream.readlines.each { |l| f << l }
      end
    end
  end
end
```

Awesome results:
```
       user     system      total        real
.zip  + read  0.000910   0.001124   0.002034 (  0.002606)
.zip  + readlines  0.000422   0.000552   0.000974 (  0.000975)
```

We could definitely skip saving a CSV file into disk and stream the export query results directly into the zip

romanrizzi:
Tried the suggested approach:

```ruby
input_path = '/Users/romeer/projects/discourse/bench/craigslistVehiclesFull.csv'

stream = StringIO.new('')
File.open(input_path) do |f|
  stream.write f.read
end

Benchmark.bm do |benchmark|
  stream.rewind

  benchmark.report(".zip  + read") do
    compressed_file_path = "#{input_path}_1.zip"

    Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
      zipfile.get_output_stream("craigslistVehiclesFull.csv") do |f|
        f.puts stream.read
      end
    end
  end

  stream.rewind
  
  benchmark.report(".zip  + readlines") do
    compressed_file_path = "#{input_path}_2.zip"

    Zip::File.open(compressed_file_path, Zip::File::CREATE) do |zipfile|
      zipfile.get_output_stream("craigslistVehiclesFull.csv") do |f|
        stream.readlines.each { |l| f.puts l }
      end
    end
  end
end
```

```
       user     system      total        real
.zip  + read 11.306406   0.637996  11.944402 ( 11.962858)
.zip  + readlines 13.245832   1.053418  14.299250 ( 14.404505)
```


No notable difference with my first benchmark (keep in mind we're compressing a 500MB CSV).  However, we could directly skip having to save a CSV first, directly streaming the export query results into the zip file.

romanrizzi:
I did the changes to skip saving the CSV to disk before compressing and tested that everything worked.

config/locales/client.en.yml
```diff
@@ -3514,7 +3514,7 @@ en:
           delete_upload_confirm: "Delete this upload? (Theme CSS may stop working!)"
           import_web_tip: "Repository containing theme"
           import_web_advanced: "Advanced..."
-          import_file_tip: ".tar.gz or .dcstyle.json file containing theme"
+          import_file_tip: ".tar.gz, .tar.zip, or .dcstyle.json file containing theme"
```
davidtaylorhq: `.tar.zip` seems unusual... does this also work with regular `.zip` files created using the windows/mac UI?

PR #7786, opened 2019-06-20
FIX: do not allow creation of topic if there is no category available for posting

techAPJ:
If posting in 'uncategorized' category is disabled and all other categories have security settings such that it requires group membership for posting then a regular user (who doesn't belong to any group permitted for creating topic) sees <kbd>+ New Topic</kbd> button but when the composer opens they aren't allowed to choose any category, rendering the button useless and confusing.

This commit adds a check in guardian verifying that there is at least one category available for creating a topic.

### DIFF
--- a/lib/guardian/topic_guardian.rb
+++ b/lib/guardian/topic_guardian.rb
@@ -39,7 +39,8 @@ module TopicGuardian
     is_staff? ||
     (user &&
       user.trust_level >= SiteSetting.min_trust_to_create_topic.to_i &&
-      can_create_post?(parent))
+      can_create_post?(parent) &&
+      Category.topic_create_allowed(self).limit(1).count == 1)
   end
 
   def can_create_topic_on_category?(category)
--- a/spec/components/guardian_spec.rb
+++ b/spec/components/guardian_spec.rb
@@ -1668,6 +1668,40 @@ describe Guardian do
     end
   end
 
+  context "can_create_topic?" do
+    it 'returns true for staff user' do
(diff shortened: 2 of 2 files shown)


lib/guardian/topic_guardian.rb
```diff
@@ -36,10 +36,16 @@ def can_publish_topic?(topic, category)
...
-      can_create_post?(parent))
+    return true if is_staff?
+
+    is_allowed = user && user.trust_level >= SiteSetting.min_trust_to_create_topic.to_i && can_create_post?(parent)
+    return false unless is_allowed
+
+    # if uncategorized topics are allowed then there is at least one category available
+    return true if SiteSetting.allow_uncategorized_topics
+
+    allowed_categories = Category.topic_create_allowed(self).pluck(:id)
```
SamSaffron: no need to pluck all the category ids here.

`Category.topic_create_allowed(self).limit(2).count == 2`

techAPJ: Done, thanks for the tip! 👍

SamSaffron:
Wait... I am re-reading this: 

 `Category.topic_create_allowed(self)` why is it counting uncategorized if you are not allowed to post there due to site setting? `topic_create_allowed` for the user should not return it imo.

techAPJ:
@SamSaffron That might actually be a bug (edge case). I'll update [scoped_to_permissions](https://github.com/discourse/discourse/blob/a0474a07745596a107384b9feb0d5bf1d408605a/app/models/category.rb#L138-L162) to not include 'uncategorized' if posting in uncategorized is disabled.

PR #7787, opened 2019-06-21
FIX: CategoryUser#batch_set

danielwaterworth:
Users were able to watch, track and mute each category simultaneously.

### DIFF
--- a/app/models/category_user.rb
+++ b/app/models/category_user.rb
@@ -184,6 +184,6 @@ end
 #
 # Indexes
 #
-#  idx_category_users_u1  (user_id,category_id,notification_level) UNIQUE
-#  idx_category_users_u2  (category_id,user_id,notification_level) UNIQUE
+#  idx_category_users_category_id_user_id  (category_id,user_id) UNIQUE
+#  idx_category_users_user_id_category_id  (user_id,category_id) UNIQUE
 #
--- a/app/models/category_user.rb
+++ b/app/models/category_user.rb
@@ -10,10 +10,6 @@ class CategoryUser < ActiveRecord::Base
     self.where(user: user, notification_level: notification_levels[level])
   end
 
-  def self.lookup_by_category(user, category)
-    self.where(user: user, category: category)
-  end
-
   def self.notification_levels
     NotificationLevels.all
   end
--- a/app/models/category_user.rb
+++ b/app/models/category_user.rb
@@ -22,11 +22,16 @@ class CategoryUser < ActiveRecord::Base
     level_num = notification_levels[level]
     category_ids = Category.where(id: category_ids).pluck(:id)
 
+    changed = false
+
     # Update pre-existing category users
-    changed =
-      CategoryUser
-        .where(user_id: user.id, category_id: category_ids)
-        .update_all(notification_level: level_num) > 0
+    unless category_ids.empty?
+      changed ||=
+        CategoryUser
+          .where(user_id: user.id, category_id: category_ids)
+          .where.not(notification_level: level_num)
+          .update_all(notification_level: level_num) > 0
+    end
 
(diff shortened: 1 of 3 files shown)


app/models/category_user.rb
```diff
@@ -23,20 +19,31 @@ def self.watching_levels
...
+      CategoryUser
+        .where(user_id: user.id, category_id: category_ids)
+        .update_all(notification_level: level_num) > 0
+
+    # Remove extraneous category users
+    changed ||=
+      CategoryUser
+        .where(user_id: user.id, notification_level: level_num)
+        .where.not(category_id: category_ids)
+        .delete_all() > 0
```
eviltrout:
```suggestion
        .delete_all > 0
```

app/models/category_user.rb
```diff
@@ -23,20 +19,31 @@ def self.watching_levels
...
+
+    # Create new category users
+    new_category_ids =
```
eviltrout: It would probably be better to use SQL here. You could write one insert statement for all the records, and because of the unique index do a `ON CONFLICT DO NOTHING`.

It's faster, but also safer under concurrency.

spec/models/category_user_spec.rb
```diff
@@ -120,7 +140,6 @@ def regular
     end
 
     it "unwatches categories that have been changed" do
-      user = Fabricate(:user)
```
eviltrout: Nice to remove all these fabricates :+1:

app/models/category_user.rb
danielwaterworth: @eviltrout, I've done this. I used the `connection.exec_query` instead of `DB.exec` because I didn't want to use a different db connection as that would have complicated the tests.

Rails 6 has bulk inserts, [see here](https://blog.bigbinary.com/2019/04/15/bulk-insert-support-in-rails-6.html), so I hope this code doesn't last too long.

SamSaffron:
DB.exec uses the exact same connection, no need to use anything else

danielwaterworth:
@SamSaffron, Good to know, will fix.

app/models/category_user.rb
eviltrout: This is marked as resolved yet I'm still seeing `delete_all()` - is this github being weird?

app/models/category_user.rb
```diff
@@ -23,21 +19,43 @@ def self.watching_levels
...
+        .where(user_id: user.id, notification_level: level_num)
+        .where.not(category_id: category_ids)
+        .delete_all() > 0
+
+    # Create new category users
+    unless category_ids.empty?
+      # TODO: rewrite this when we upgrade to rails 6
+      changed ||=
+        DB.exec(
+          "
```
eviltrout: FYI, I think `<<~SQL` (multiline heredoc strings in ruby) work really well with `DB.exec`. This string is fine, but it's something to consider.

db/migrate/20190621095105_remove_notification_level_from_category_user_indexes.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class RemoveNotificationLevelFromCategoryUserIndexes < ActiveRecord::Migration[5.2]
+  def up
+    execute <<SQL
+DELETE FROM category_users cu USING category_users cu1
```
eviltrout: This is another place where `<<~` might be good.

app/models/category_user.rb
danielwaterworth: Oh, I see what happened. I committed the change using github's new commit suggestion feature, but then I didn't pull before rebasing.

SamSaffron:
User merger spec just failed on travis, merging and will sort it out.

PR #7788, opened 2019-06-21
FIX: Don't use DistributedCache to store redis readonly state

davidtaylorhq:
This can cause unbound CPU usage in some cases, and excessive logging in other cases. This commit moves redis readonly information into the local process, but maintains the DistributedCache for postgres readonly state.


eviltrout:
Tagging @SamSaffron for this one

SamSaffron:
@tgxworld I think you are the best person to review this.

PR #7789, opened 2019-06-21
Add endpoint to individually update a theme setting

pmusaraj:
Adds a new endpoint at `admin/themes/ID/setting` which accepts a POST request with name/value parameters to update an individual setting in a theme. In the admin UI, this endpoint is now used and it fixes an issue whereby updating one setting resets unsaved changes to other theme settings. 

The PR also adds a new staff action type for `change theme setting` for easier logging of staff changes to theme settings, example: 

image

### DIFF
--- a/app/assets/javascripts/admin/components/theme-setting-editor.js.es6
+++ b/app/assets/javascripts/admin/components/theme-setting-editor.js.es6
@@ -7,7 +7,7 @@ export default Ember.Component.extend(BufferedContent, SettingComponent, {
   layoutName: "admin/templates/components/site-setting",
   _save() {
     return ajax(`/admin/themes/${this.model.id}/setting`, {
-      type: "POST",
+      type: "PUT",
       data: {
         name: this.setting.setting,
         value: this.get("buffered.value")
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/admin/components/theme-setting-editor.js.es6
```diff
@@ -1,12 +1,17 @@
...
 
 export default Ember.Component.extend(BufferedContent, SettingComponent, {
   layoutName: "admin/templates/components/site-setting",
   _save() {
-    return this.model.saveSettings(
-      this.get("setting.setting"),
-      this.get("buffered.value")
-    );
+    return ajax(`/admin/themes/${this.model.id}/setting`, {
+      type: "POST",
```
eviltrout: In Rails the convention is to use `PUT` for an update and `POST` for a create. In this case it seems more like an update than a create so I think it should use `PUT`.

app/controllers/admin/themes_controller.rb
```diff
@@ -261,6 +261,22 @@ def diff_local_changes
...
+
+    setting_name = params[:name].to_sym
+    new_value = params[:value] || nil
+
+    previous_value = @theme.included_settings[setting_name]
+    @theme.update_setting(setting_name, new_value)
+    @theme.save
+
+    log_theme_setting_change(setting_name, previous_value, new_value)
+    render json: {}, status: :ok
```
eviltrout: Could we return the serialized setting here? It would be more consistent with other RESTful routes.

pmusaraj:
Thanks @eviltrout I went ahead and merged after addressing your comments.

PR #7793, opened 2019-06-22
FIX: do not include uncategorized_category_id in `topic_create_allowed` if posting in uncategorized is disabled

techAPJ:
https://github.com/discourse/discourse/pull/7786#issuecomment-504262406

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -108,8 +108,15 @@ class Category < ActiveRecord::Base
   }
 
   TOPIC_CREATION_PERMISSIONS ||= [:full]
-  POST_CREATION_PERMISSIONS  ||= [:create_post, :full]
-  scope :topic_create_allowed, -> (guardian) { scoped_to_permissions(guardian, TOPIC_CREATION_PERMISSIONS) }
+  scope :topic_create_allowed, -> (guardian) do
+    category_ids = scoped_to_permissions(guardian, TOPIC_CREATION_PERMISSIONS)
+    if !guardian.is_staff? && !SiteSetting.allow_uncategorized_topics
+      category_ids.delete(SiteSetting.uncategorized_category_id)
+    end
+    category_ids
+  end
+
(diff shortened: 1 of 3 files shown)


app/models/category.rb
```diff
@@ -108,8 +108,15 @@ class Category < ActiveRecord::Base
   }
 
   TOPIC_CREATION_PERMISSIONS ||= [:full]
-  POST_CREATION_PERMISSIONS  ||= [:create_post, :full]
-  scope :topic_create_allowed, -> (guardian) { scoped_to_permissions(guardian, TOPIC_CREATION_PERMISSIONS) }
+  scope :topic_create_allowed, -> (guardian) do
+    category_ids = scoped_to_permissions(guardian, TOPIC_CREATION_PERMISSIONS)
+    if !guardian.is_staff? && !SiteSetting.allow_uncategorized_topics
+      category_ids.delete(SiteSetting.uncategorized_category_id)
```
SamSaffron: I am confused ... shouldn't we be fixing `scoped_to_permissions` here?

techAPJ: I didn't added this fix in `scoped_to_permissions` method because it's also being used in `post_create_allowed` scope. Pushing this fix in `scoped_to_permissions` would mean that normal users wouldn't be able to reply in a topic that is uncategorized and later on `allow_uncategorized_topics` is enabled.

SamSaffron: I am mixed on allowing this shaningans, if mods disable uncategorized stuff should probably get quite brutal for end users on uncategorized topics anyway ... I made a slight improvement on this commit and pushed it.

SamSaffron:
ok I amended this in core, needed these two

https://review.discourse.org/t/fix-do-not-include-uncategorized-category-id-in-topic-create-allowed-if-posting-in-uncategorized-is-disabled/4425

and 

https://review.discourse.org/t/dev-correct-edge-case-introduced-in-333b5a19/4426

PR #7800, opened 2019-06-24
FIX: generated class names for tags with colon

rrooding:
In the topic list, the tag name is automatically converted to a css class name. We are using tags that contain a colon in the tag name like `threat:high`. This was converted incorrectly into the css class name `tag-threat:high`. This fixes the class name generation to be `tag-threat-high`. See attached screenshots.

before:
before
after:
after

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-list-item.js.es6
@@ -51,7 +51,9 @@ export const ListItemDefaults = {
     }
 
     if (topic.get("tags")) {
-      topic.get("tags").forEach(tagName => classes.push("tag-" + tagName));
+      topic.get("tags").forEach(tagName => {
+        classes.push("tag-" + tagName.replace(/:/g, '-'));
+      });
     }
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am very very very on the fence here... I think we should simply disallow these odd tag names, simply ban `:` in a tag name ... if we are going down the route of requiring slugification of tags we are going to need to be wider than this.

rrooding:
There is some inconsistency in how special characters in tags are handled currently:
Screenshot 2019-06-26 at 11 11 25

We are using tags with colons for this purpose: https://www.us-cert.gov/tlp

Of course we could replace it with a dash or something else, but even if the useable characters in tags will be more restricted, I think there is something to say for colons as a sort of grouping.

Another solution would be to be able to prepend the group name.

SamSaffron:
Wait... isn't this just a matter of targetting here? 

https://mathiasbynens.be/demo/crazy-class

This should work no? 

```
.tag-threat\:high {
   
}
```

rrooding:
I just tested the escaping in the CSS and that works! So for now it seems that there is no issue anymore.

PR #7801, opened 2019-06-25
SECURITY: XSS in Ember User Routes.

tgxworld:
A carefully crafted user path can lead to us hitting the uploads
endpoint in nginx. If the upload contains a carefully crafted payload,
custom HTML can be injected leading to XSS.

### DIFF
--- a/app/assets/javascripts/discourse/routes/user.js.es6
+++ b/app/assets/javascripts/discourse/routes/user.js.es6
@@ -1,3 +1,5 @@
+import { USERNAME_ROUTE_REGEXP } from "discourse/routes/concerns/username-route-regexp";
+
 export default Discourse.Route.extend({
   titleToken() {
     const username = this.modelFor("user").username;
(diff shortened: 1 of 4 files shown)


lib/route_format.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 module RouteFormat
+  USERNAME_REGEXP = '[%\w.\-]+?'
```
ZogStriP: We might want to add a comment saying this regular expression is being used in JS as well (thus preventing us from using Ruby-specifics classes)

SamSaffron:
What happens with Unicode usernames?

tgxworld:
The scope of this PR isn’t to tackle Unicode usernames though.

app/assets/javascripts/discourse/routes/user.js.es6
```diff
@@ -39,7 +41,11 @@ export default Discourse.Route.extend({
       return this.currentUser;
     }
 
-    return Discourse.User.create({ username: params.username });
+    if (params.username.match(USERNAME_ROUTE_REGEXP)) {
```
gschlager: Does `/u/%E7%8B%AE%E5%AD%90/summary` (for the user `狮子`) still work? This will probably break Unicode usernames unless `params.username` contains the URL encoded username.

gschlager:
Closing in favor of #7805

PR #7802, opened 2019-06-25
 FIX: provides an emoji helper to replace codes by images

jjaffeux:


app/helpers/emoji_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module EmojiHelper
+  def emoji_codes_to_img(str)
+    if str
+      str = str.gsub(/:([\w\-+]*(?::t\d)?):/) do
+        code = $1
+        "<img src=\"#{Emoji.url_for(code)}\" title=\"#{code}\" class=\"emoji\" alt=\"#{code}\">"
```
eviltrout: To avoid injection, perhaps we shouldn't render anything if the `url_for(code)` doesn't find an emoji?

That way there is no harm injecting the `code` into the string like that.

jjaffeux: In this case this is supposed to be escaped, but I do agree we should prevent it from the source

SamSaffron: I am with Robin that we should support

```
EmojiHelper.emoji_codes_to_img("a test :bogus: a test") => "a test :bogus: a test"
```

If we don't find a URL just gsub with yourself. 

Besides that this is good to merge imo.

app/models/emoji.rb
```diff
@@ -48,19 +43,30 @@ def self.exists?(name)
   end
 
   def self.[](name)
-    Emoji.custom.detect { |e| e.name == name }
+    name = name.delete_prefix(':').delete_suffix(':')
```
ZogStriP: I wonder if this might be faster 
```suggestion
    name = name[1...-1]
```

app/models/emoji.rb
```diff
@@ -48,19 +43,30 @@ def self.exists?(name)
...
+    is_toned = name.match?(/.+:t[1-6]/)
+    normalized_name = name.gsub(/(.+):t[1-6]/, '\1')
+
+    Emoji.all.detect do |e|
+      e.name == normalized_name &&
+      (!is_toned || is_toned && e.tonable)
```
ZogStriP: Depending on the precedence of the `||` and the `&&` operators, this might not do what you want.

It's always better to use parentheses when mixing them.

```suggestion
      (!is_toned || (is_toned && e.tonable))
```

app/models/emoji.rb
```diff
@@ -48,19 +43,30 @@ def self.exists?(name)
...
   end
 
   def self.create_from_db_item(emoji)
     name = emoji["name"]
     filename = emoji['filename'] || name
+    is_tonable = Emoji.tonable_emojis.include?(name)
+
     Emoji.new.tap do |e|
       e.name = name
+      e.tonable = is_tonable
```
ZogStriP: `is_tonable` feels unnecessary here.

```suggestion
      e.tonable = Emoji.tonable_emojis.include?(name)
```

app/helpers/emoji_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module EmojiHelper
+  def emoji_codes_to_img(str)
```
ZogStriP: Small suggestion, but I find it better when we do the short-circuits at the beginning of the method.
Plus, it reduces the nesting level 👍 

``` ruby
def emoji_codes_to_img(str)
  return if str.blank?

  raw str.gsub(/foobar/) do |name|
    # ...
  end
end
```

PR #7803, opened 2019-06-25
FIX: Don't check for second factor when switching to anonymous account

rrooding:
This is a fix for the issue discussed in https://meta.discourse.org/t/enforce-second-factor-disables-use-of-anonymous-posting/121151


SamSaffron:
I think this looks correct to me @featheredtoast can you review and merge?

PR #7805, opened 2019-06-26
SECURITY: XSS in Ember User Routes

gschlager:
Alternative to https://github.com/discourse/discourse/pull/7801


app/controllers/uploads_controller.rb
```diff
@@ -69,7 +69,7 @@ def lookup_urls
   end
 
   def show
-    return render_404 if !RailsMultisite::ConnectionManagement.has_db?(params[:site])
+    return render_404 if !RailsMultisite::ConnectionManagement.has_db?(params[:site]) || request.xhr?
```
davidtaylorhq: I think it would be good to add an explanation comment here, because it's not obvious why `||request.xhr?` is included

app/controllers/uploads_controller.rb
```diff
@@ -88,6 +88,8 @@ def show
   end
 
   def show_short
+    return render_404 if request.xhr?
```
davidtaylorhq: Ditto

PR #7812, opened 2019-06-26
FEATURE: allow disabling theme components

OsamaSayegh:


app/models/theme.rb
```diff
@@ -156,8 +156,10 @@ def self.transform_ids(ids, extend: true)
 
       all_ids = [parent, *components]
 
-      disabled_ids = Theme.where(id: all_ids).includes(:remote_theme)
-        .reject(&:enabled?).pluck(:id)
+      disabled_ids = Theme.where(id: all_ids)
+        .includes(:remote_theme)
+        .select { |t| !t.supported? || t.disabled? }
```
SamSaffron: This select is a bit confusing to me ... can't it use a `.where` here?

app/models/theme.rb
```diff
@@ -506,6 +525,7 @@ def generate_metadata_hash
 #  color_scheme_id  :integer
 #  remote_theme_id  :integer
 #  component        :boolean          default(FALSE), not null
+#  disabled         :boolean          default(FALSE), not null
```
SamSaffron: Maybe we go with `enabled` here ... 

Reasoning about `where enabled` is way simpler than thinking about double negatives, `where not disabled`

spec/requests/admin/themes_controller_spec.rb
```diff
@@ -322,6 +322,54 @@
...
+      json = JSON.parse(response.body)
+      expect(json["theme"]["disabled"]).to eq(true)
+      expect(UserHistory.where(
+        context: child.id.to_s,
+        action: UserHistory.actions[:disable_theme_component]
+      ).size).to eq(1)
+      expect(json["theme"]["disabled_by"]["id"]).to eq(admin.id)
+    end
+
+    it 'enabling a component doesn\'t create a staff action log' do
```
SamSaffron: minor ... but one of those places where `"` should win over `'`

SamSaffron:
overall very happy with this change, just need to tune a few of my comments also get a tiny bit of UX feedback from @awesomerobot

app/models/theme.rb
OsamaSayegh: `.supported?` isn't a column it's a method that checks if the theme version is compatible with discourse's, so we need to load up the themes and filter out the ones that aren't compatible or disabled. Does that make sense?

SamSaffron: yes, makes sense... what I got confused here about is the way it does a `.pluck` here afterwards. 

```
 pry(main)> Topic.limit(10).select{|t| t.created_at < 1.year.ago}.class
  Topic Load (1.4ms)  SELECT  "topics".* FROM "topics" WHERE ("topics"."deleted_at" IS NULL) LIMIT 10
=> Array

 [Topic.new].pluck(:id)
=> [nil]
```

I guess active support likes monkey patching.

app/models/user_history.rb
```diff
@@ -95,7 +95,8 @@ def self.actions
       embeddable_host_update: 74,
       embeddable_host_destroy: 75,
       web_hook_deactivate: 76,
-      change_theme_setting: 77
+      change_theme_setting: 77,
+      disable_theme_component: 78
```
SamSaffron: hmmm so we have a record for "disable" we should also have one for "enable"

SamSaffron:
Overall happy with this change pending my 1 piece of feedback, can you post a current state of UI in a screenshot here?

OsamaSayegh:
Ok @SamSaffron I've added a staff log for enabling components.

Screenshots:

SamSaffron:
Merging this :confetti_ball:  awesome work @OsamaSayegh

PR #7814, opened 2019-06-26
FEATURE: Add new group visibility option for "logged on users"

pmusaraj:
In some cases it is useful to only allow logged on users to be able to view group membership. That's what this PR does. It also sets all automatic groups (except`everyone`) to be visible to "logged on users", previously they were marked as public but suppressed in the group page for non-staff.  

Admin UI screenshot: 
image

### DIFF
--- a/db/migrate/20190705173948_renumber_group_visibility_levels.rb
+++ b/db/migrate/20190705173948_renumber_group_visibility_levels.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class RenumberGroupVisibilityLevels < ActiveRecord::Migration[5.2]
+  def up
+    execute "UPDATE groups SET visibility_level = 4 WHERE visibility_level = 3"
+    execute "UPDATE groups SET visibility_level = 3 WHERE visibility_level = 2"
+    execute "UPDATE groups SET visibility_level = 2 WHERE visibility_level = 1"
+  end
+
+  def down
+    execute "UPDATE groups SET visibility_level = 0 WHERE visibility_level = 1"
+    execute "UPDATE groups SET visibility_level = 1 WHERE visibility_level = 2"
+    execute "UPDATE groups SET visibility_level = 2 WHERE visibility_level = 3"
+    execute "UPDATE groups SET visibility_level = 3 WHERE visibility_level = 4"
+  end
+end
--- a/spec/components/guardian_spec.rb
+++ b/spec/components/guardian_spec.rb
@@ -3027,7 +3027,7 @@ describe Guardian do
   end
 
   describe(:can_see_group) do
-    it 'Correctly handles owner visibile groups' do
+    it 'Correctly handles owner visible groups' do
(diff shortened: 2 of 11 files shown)


SamSaffron:
This looks good to me, go ahead and merge... a tiny bit mixed about id (4) belonging to this thing, cause the numbers are kind of expanding in scope... but re-numbering this thing is going to be complicated. 

Can you first perhaps see how complex it will be to re-number?

pmusaraj:
It shouldn't be too painful to renumber. Is this below the preferred numbering? 

public: 0,
logged_on_users: 1,
members: 2,
staff: 3,
owners: 4

SamSaffron:
Yeah I think that numbering is right ... technically "staff" kind of means "moderators" cause "admins" unconditionally get to see groups. I think "owners" can be 4 as long as "mods" can not see it when set... otherwise it needs to be 3.

pmusaraj:
I have added a migration for the renumbering, and made two more changes.

First, I set the default visibility for the "everyone" group to be "staff" in `db/fixtures/002_groups.rb`. This way it's more consistent with the `Group.refresh_automatic_group!` method that gets called by the ensure_consistency job, see https://github.com/discourse/discourse/blob/master/app/models/group.rb#L320 

Second change: I modified the language in the UI, adding a field description with "Admins can see all groups.": 

image

and removed the "admins" from the label of each option in the dropdown: 

image

db/migrate/20190705173948_renumber_group_visibility_levels.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class RenumberGroupVisibilityLevels < ActiveRecord::Migration[5.2]
+  def up
+    execute "UPDATE groups SET visibility_level = 4 WHERE visibility_level = 3"
```
eviltrout: Super minor but this could have been `SET visibility_level = visibility_level + 1` right? And below the opposite? (`-1`).
No need to change it just an observation :)

spec/components/guardian_spec.rb
```diff
@@ -3144,9 +3167,10 @@
       expect(Guardian.new(moderator).can_see_groups?([group])).to eq(true)
       expect(Guardian.new(owner).can_see_groups?([group])).to eq(true)
       expect(Guardian.new.can_see_groups?([group])).to eq(false)
+      expect(Guardian.new(another_user).can_see_groups?([group])).to eq(false)
     end
 
-    it 'correctly handles member visibile groups' do
```
eviltrout: Oh my this typo is in so many places. Just goes to show how much code is copied and pasted :)

db/migrate/20190705173948_renumber_group_visibility_levels.rb
pmusaraj: Yeah, that would have been better (I would need to modify it so that 0-level groups shouldn't be bumped to level 1 on `up`).

PR #7815, opened 2019-06-27
FIX: don't insert TranslationOverride records if their values match original

OsamaSayegh:


spec/models/translation_override_spec.rb
```diff
@@ -47,6 +47,23 @@
...
+    I18n.backend.store_translations(:en, keyfortests: "some text")
+
+    TranslationOverride.upsert!(:en, 'keyfortests', 'some text')
+    expect(TranslationOverride.where(translation_key: 'keyfortests').count).to eq(0)
+  end
+
+  it "destroys existing record if new value matches original" do
+    I18n.backend.store_translations(:en, keyfortests: "some text")
+
+    TranslationOverride.upsert!(:en, 'keyfortests', 'some text 2', bypass_reload: true)
```
gschlager: Is the `bypass_reload` parameter only used by this test? Could you try rewriting the test so that this parameter isn't needed?

OsamaSayegh: @gschlager I could use the key of a real translation for this test. Does that sound good to you?

SamSaffron: Yeah reusing a translation is fine here, it will be clear how to fix it if needed.

OsamaSayegh: OK, done 👍

app/models/translation_override.rb
```diff
@@ -20,6 +20,12 @@ class TranslationOverride < ActiveRecord::Base
   def self.upsert!(locale, key, value)
     params = { locale: locale, translation_key: key }
 
+    original_text = I18n.overrides_disabled { I18n.t(key, locale: locale) }
+    if original_text == value
+      revert!(locale, [key])
```
SamSaffron: nice ... yeah this makes sense.

db/migrate/20190627040131_remove_superfluous_translation_override_records.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class RemoveSuperfluousTranslationOverrideRecords < ActiveRecord::Migration[5.2]
+  def up
+    I18n.overrides_disabled do
```
SamSaffron: I am very uneasy with this migration cause it loads up the entire localization system, in general I prefer migration to simple only run sql statements. 

Can we live with this superfluous data for now? End users I assume will have a simple way to correct if needed.

SamSaffron:
What specific issue is this fixing, is this just about reducing the amount of data in the db? AFAIK there is a "revert" button in the UX, is it not clearing stuff up properly?

OsamaSayegh:
@SamSaffron I think we can live without the migration. I don't think there are more than 100 superfluous records in any site, which isn't too many?

The revert button does clear stuff up properly, but the logic for it to show up is:

https://github.com/discourse/discourse/blob/427cf295d90080cae6cea148db87b1c2cffb9875/app/serializers/site_text_serializer.rb#L14-L20

So if a translation string has a record with the same exact value as the default translation, the button won't show up. I can change the logic so the button shows up if there is a record in the db. I'd need to be careful not to cause N queries, but I think it's doable. Does that sound good?

SamSaffron:
Yeah I love that fix best, it simplifies a ton here, just show the button unconditionally if a record exists in the DB. A lot less needs to change then including the magical delete on upsert.

SamSaffron:
This looks safe to me... merging it in.

PR #7818, opened 2019-06-27
FIX: ensures discovery params are kept in history before reset

jjaffeux:


jjaffeux:
@eviltrout since https://github.com/discourse/discourse/commit/02f6db4e59e03f49739be7e228641613dab0ae26

opening a new tab with: `/latest?order=views`, clicking on a topic and then trying to hit back button will not work.

My understanding is that we are reseting params before storing them in history making it impossible to go back to there after. I'm unsure this is the correct fix though.

PR #7820, opened 2019-06-27
PERF: Use Oj for serializing JSON.

nbianca:
Another try at using Oj to optimize JSON serialization.


SamSaffron:
Awesome! Let's try this out!

PR #7822, opened 2019-06-28
FEATURE: in:tagged search (srv side)

joshmoore:
Similar to in:solved or in:unsolved, the filters check for an
existence of the topic_id in the topic_tags table:

 * `in:tagged` means that *at least one tag* exists for a topic
 * `in:untagged` means that *no tag* exists for a topic

see:
 * https://meta.discourse.org/t/how-to-search-filter-untagged-topics/119641/2
 * UI-related PR: https://github.com/discourse/discourse/pull/7721

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -267,15 +267,12 @@ class Search
 
   advanced_filter(/^in:tagged$/) do |posts|
     posts
-      .joins("JOIN topic_tags ON
-        topic_tags.topic_id = posts.topic_id")
+      .where('EXISTS (SELECT 1 FROM topic_tags WHERE topic_tags.topic_id = posts.topic_id)')
   end
 
   advanced_filter(/^in:untagged$/) do |posts|
     posts
-      .joins("LEFT JOIN topic_tags ON
-        topic_tags.topic_id = posts.topic_id")
-      .where("topic_tags.id IS NULL")
+      .where('NOT EXISTS (SELECT 1 FROM topic_tags WHERE topic_tags.topic_id = posts.topic_id)')
   end
 
   advanced_filter(/^status:open$/) do |posts|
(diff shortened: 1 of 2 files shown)


lib/search.rb
```diff
@@ -265,6 +265,19 @@ def self.advanced_filters
     @advanced_filters
   end
 
+  advanced_filter(/^in:tagged$/) do |posts|
+    posts
+      .joins("JOIN topic_tags ON
```
SamSaffron: wait I think this expands the result set ... don't we want?

.where('EXISTS (SELECT 1 FROM topic_tags WHERE topic_tags.topic_id = posts.topic_id)')

joshmoore: I'm unfamiliar with the result set semantics and was mostly just copying what I saw for similar tags (in:seen, in:unseen). Happy to try with a subselect both here and for in:untagged _a la_
```
.where(`NOT EXISTS...`)
```

joshmoore: Tested locally in the UI & pushed.

lib/search.rb
```diff
@@ -265,6 +265,16 @@ def self.advanced_filters
...
   end
 
+  advanced_filter(/^in:tagged$/) do |posts|
+    posts
+      .where('EXISTS (SELECT 1 FROM topic_tags WHERE topic_tags.topic_id = posts.topic_id)')
+  end
+
+  advanced_filter(/^in:untagged$/) do |posts|
+    posts
+      .where('NOT EXISTS (SELECT 1 FROM topic_tags WHERE topic_tags.topic_id = posts.topic_id)')
```
SamSaffron: sorry ... last one... leave this as a left join, it is certainly the fastest way of doing it and will not expand the result set like the opposite does... merging it right after that .

joshmoore: :smile: no worries. Done, tested, & pushed. Let me know if you want me to collapse these down.

SamSaffron:
:heart: thanks! and sorry about the huge delay here!

joshmoore:
yw & np

PR #7826, opened 2019-06-28
DEV: Replace Overcommit with Lefthook

Arkweid:
Overcommit uses prebuilt hooks and require global installation.
To avoid this issues replace it with [Lefthook](https://github.com/Arkweid/lefthook).
Lefthook will be installed with npm packages. New contributors
will have fully consistent git hooks.

Goals:
- **Strict workflow**. Avoid situations when newcomers pushed PR without checks with rubocop or eslint. And members should [recommend](https://github.com/discourse/discourse/pull/7192#issuecomment-473973502) to install overcommit.
- **Flexibility**. You free to combine basic operations.
- **Сlarification**. Overcommit provide prebuilt hooks. So its hard to understend what is going on without inspecting overcommit repository.
- **20% Faster**. Well, it not a big issue in development environment, but a like faster instruments :)
- **Twice speed up for CI lints**. With Lefthook we can run lints on CI in paralel mode. You can check [example here](https://travis-ci.org/discourse/discourse/jobs/551862476#L956).

Example output for this commit:
PR example

Example output for CI lints:
PR example

Don't forget to remove preinstalled hooks from you repo `overcommit --uninstall` if you want to test it action.

So what do you think about. Is it worth?


ZogStriP:
I 😍 that it also runs prettier.

gschlager:
@Arkweid How does Lefthook handle Security? I like that [overcommit uses signature checking](https://github.com/sds/overcommit#security) to prevent arbitrary commands from running on my system without me knowing about it.

Arkweid:
@gschlager thank you for this question. Lefthook uses signatures to keep hooks up to date. And does not use them for security purposes.
My opinion is that this feature does not bring any security improvements.
When you clone the project like discourse, you should to run several commands `yarn install` and `bundle install`. And you hope none of 300+ npm dependencies don't do something terrible. Same for gems. With installation process it runs Make file(for native extensions) and who know what is going inside. You again hope it's ok.

Instruments like Overcommit hide hooks implementation inside. So when developers see changes in `.overcommit.yml` they must sign these changes, and again they hope it's ok.

Example:
```yml
PreCommit:
  YamlSyntax:
    enabled: true
```
To understand what exactly going on here you should look at the code in overcommit repository. And between versions, this code can be changed. I hope the new code will not try to break something. So, I see only one benefit from the _sign option_ it: "Hey, `.overcommit.yml` changed. Do you trust it?".

For my opinion, the only way to achieve security - install project into Docker container.
___
But if you think that this is still an important feature, then I can add it. Just let me know.

ZogStriP:
Let's try this 👍

PR #7829, opened 2019-07-01
FEATURE: support query params when redirecting to internal link on login

techAPJ:
https://meta.discourse.org/t/redirect-after-login-loses-query-params/121178

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -101,8 +101,7 @@ class StaticController < ApplicationController
            (uri.host.blank? || uri.host == forum_uri.host) &&
            uri.path !~ /\./
 
-          destination = uri.path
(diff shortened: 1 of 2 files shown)


app/controllers/static_controller.rb
```diff
@@ -101,8 +101,7 @@ def enter
            (uri.host.blank? || uri.host == forum_uri.host) &&
            uri.path !~ /\./
 
-          destination = uri.path
-          destination = "#{uri.path}?#{uri.query}" if uri.path =~ /new-topic/ || uri.path =~ /new-message/ || uri.path =~ /user-api-key/
+          destination = "#{uri.path}#{uri.query ? "?#{uri.query}" : ""}"
```
eviltrout: This does seem slightly less secure than before. However, I can't think of a way `uri.query` could be exploited so it's probably fine.

PR #7838, opened 2019-07-02
FIX: Show 'Export' button for all tabular reports.

nbianca:


app/assets/javascripts/admin/components/admin-report.js.es6
```diff
@@ -55,7 +55,7 @@ export default Ember.Component.extend({
   showTitle: true,
   showFilteringUI: false,
   showDatesOptions: Ember.computed.alias("model.dates_filtering"),
-  showExport: Ember.computed.not("model.isTable"),
+  showExport: Ember.computed.alias("model.isTable"),
```
eviltrout: A grep shows me that `showExport` is only used in one template. I think it would be nicer if we removed this computed property and put `model.isTable` in the template itself.

PR #7840, opened 2019-07-02
FIX: Use title attribute for notification items.

nbianca:
`title` function was also used by the `Widget` class to build the attributes of `li` element. The `title` attribute should only appear on the `a` element.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/custom-notification-item.js.es6
+++ b/app/assets/javascripts/discourse/widgets/custom-notification-item.js.es6
@@ -4,7 +4,7 @@ import { formatUsername } from "discourse/lib/utilities";
 import { iconNode } from "discourse-common/lib/icon-library";
 
 createWidgetFrom(DefaultNotificationItem, "custom-notification-item", {
-  title(notificationName, data) {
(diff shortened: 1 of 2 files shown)


eviltrout:
Looks okay to me, @danielwaterworth what do you think?

danielwaterworth:
@nbianca, There are a few specializations of the `default-notification-item` widget, (see `./app/assets/javascripts/discourse/widgets/*-notification-item.js.es6`) and some of them override the `title` method. Could you change them too? We'll need to fix discourse-code-review too, (see [this file](https://github.com/discourse/discourse-code-review/blob/master/assets/javascripts/discourse/widgets/code-review-commit-approved-notification-item.js.es6)).

nbianca:
@danielwaterworth Looks like only `custom-notification-item.js.es6` overrides the `title` method which is already included in this PR. I will have a look at discourse-code-review.

danielwaterworth:
@nbianca, I'm sorry, I thought there were more than that. Thanks!

eviltrout:
Thanks, merging!

PR #7842, opened 2019-07-02
FIX: Set a minimum reading time per post.

nbianca:


jjaffeux:
@nbianca feels like the 4 here should be variable so we can give it a meaningful name, feels too much out of nowhere otherwise

app/assets/javascripts/discourse/widgets/toggle-topic-summary.js.es6
```diff
@@ -1,15 +1,20 @@
...
     if (attrs.topicSummaryEnabled) {
       return I18n.t("summary.enabled_description");
     }
 
     if (attrs.topicWordCount) {
-      const readingTime = Math.floor(
-        attrs.topicWordCount / this.siteSettings.read_time_word_count
+      const readingTime = Math.ceil(
+        Math.max(
+          attrs.topicWordCount / this.siteSettings.read_time_word_count,
```
ZogStriP: Since `read_time_word_count` might be `0`, you need to add some protection for that case

lib/topic_view.rb
```diff
@@ -208,7 +209,13 @@ def summary(opts = {})
 
   def read_time
     return nil if @post_number > 1 # only show for topic URLs
-    (@topic.word_count / SiteSetting.read_time_word_count).floor if @topic.word_count
+
+    if @topic.word_count && SiteSetting.read_time_word_count > 0
```
ZogStriP: Can you add a test for that?

spec/components/topic_view_spec.rb
```diff
@@ -694,4 +694,23 @@ def topic_view_near(post, show_deleted = false)
       expect(topic_view.last_post_id).to eq(p3.id)
     end
   end
+
+  describe '#read_time' do
+    let!(:post) { Fabricate(:post, topic: topic) }
+
+    before do
+      post.revise(Discourse.system_user, raw: "![image|100x100](upload://upload.png)")
```
ZogStriP: Why do you need to revise it, can't you pass the raw directly?

spec/components/topic_view_spec.rb
```diff
@@ -694,4 +694,23 @@ def topic_view_near(post, show_deleted = false)
...
+      topic_view.topic.reload
+    end
+
+    it 'should return the right read time' do
+      expect(topic_view.read_time).to eq(1)
```
ZogStriP: This test is testing default values and is bound to fail whenever we change them.

spec/components/topic_view_spec.rb
```diff
@@ -694,4 +694,23 @@ def topic_view_near(post, show_deleted = false)
...
+
+      SiteSetting.read_time_word_count = 1000000000
```
ZogStriP: Not sure we need to test this very (very) unlikely used value.

spec/components/topic_view_spec.rb
nbianca: It's a hack to make it update the word_count of the topic.

lib/topic_view.rb
ZogStriP: If you have a topic with 3 posts: OP with some text and 2 replies with only some images.

How does this take into account the replies into the reading time?

nbianca: It depends on the post itself and on the `SiteSetting.read_time_word_count` value.

If the post is short, then the result will be 1 minute of reading (4 seconds per post). If the post is long (> SiteSetting.read_time_word_count words), then the result will be dominated by that value (`word_count / SiteSetting.read_time_word_count` minutes).

ZogStriP: Nevermind, I thought posts with only an image would have a `word_count == 0` but it turns out they have some words (ie, chunks from the URL).

eg. a post with this `raw` content

``` text
![image|277x128](upload://lYlShwE97fhTvChFLbM3zXyUPnj.png) 
``` 

will have the following "words": image, 277x128, upload, lYlShwE97fhTvChFLbM3zXyUPnj and	png.

PR #7845, opened 2019-07-02
FIX: Ensure lightbox image download has correct content disposition in S3

pmusaraj:
Fixes an inconsistency where clicking on the "Download" button of a lightboxed image opens the image in the browser when the site is configured to use S3 (instead of downloading the image directly, which is the default behaviour for local uploads). 

To fix, we need to use a presigned URL in S3 and pass it a content-disposition parameter ("attachment; filename="something.png") to encore a download as well as use the original filename.

See https://meta.discourse.org/t/viewing-images-at-full-size/121359 for more discussion.

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -383,11 +383,7 @@ class CookedPostProcessor
     img.add_next_sibling(a)
 
     if upload
-      if Discourse.store.internal?
-        a["data-download-href"] = Discourse.store.download_url(upload)
-      else
-        a["data-download-href"] = "#{upload.short_path}?dl=1"
-      end
+      a["data-download-href"] = Discourse.store.download_url(upload)
     end
 
     a.add_child(img)
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -382,8 +382,12 @@ class CookedPostProcessor
     a = create_link_node("lightbox", img["src"])
     img.add_next_sibling(a)
 
-    if upload && Discourse.store.internal?
-      a["data-download-href"] = Discourse.store.download_url(upload)
+    if upload
+      if Discourse.store.internal?
+        a["data-download-href"] = Discourse.store.download_url(upload)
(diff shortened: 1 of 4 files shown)


lib/file_store/s3_store.rb
```diff
@@ -104,10 +104,15 @@ def path_for(upload)
       FileStore::LocalStore.new.path_for(upload) if url && url[/^\/[^\/]/]
     end
 
-    def url_for(upload)
-      if upload.private?
+    def url_for(upload, params = {})
```
eviltrout: I am not a fan of passing the web parameters directly through to service objects / models. In this case it seems the `s3_store` class is even comparing against a string.

It would be better if we added a keyword argument here: `def url_for(upload, force_download: false)` and then have the controller set that value to `params[:dl] == "1"`.

pmusaraj: Fixed. I initially opted for passing the parameters to avoid nesting `if` statements in the controller.

eviltrout: You could avoid it in this case with `force_downloaded: params[:dl] == "1"` though?

lib/cooked_post_processor.rb
```diff
@@ -382,8 +382,12 @@ def add_lightbox!(img, original_width, original_height, upload, cropped: false)
     a = create_link_node("lightbox", img["src"])
     img.add_next_sibling(a)
 
-    if upload && Discourse.store.internal?
-      a["data-download-href"] = Discourse.store.download_url(upload)
+    if upload
+      if Discourse.store.internal?
+        a["data-download-href"] = Discourse.store.download_url(upload)
```
SamSaffron: wait ... why can't we trust "download_url" to always be correct regardless of store? "dl=1" to me feels very much like a "store" concern not something cooked_post_processor.

pmusaraj: You're right of course, I've made the change, it's cleaner to have this in the store.

PR #7848, opened 2019-07-03
FEATURE: Rake themes installer

rrooding:
This makes it possible to install themes and theme components from the command line. It accepts a YAML file with the themes you want to install.

```
$ bundle exec rake themes:install < config/themes.yml
  example1: is already installed
  example2: installed from https://github.com/example/theme.git
  doesntexist: Error cloning git repository, access is denied or repository is not found

Results:
 Installed: 1
 Updated:   0
 Skipped:   1
 Errors:    1
```

### DIFF
--- a/spec/models/theme_spec.rb
+++ b/spec/models/theme_spec.rb
@@ -59,8 +59,6 @@ describe Theme do
 
   end
 
-  it { is_expected.to belong_to(:remote_theme).dependent(:destroy) }
-
   it "can automatically disable for mismatching version" do
     expect(theme.enabled?).to eq(true)
     theme.create_remote_theme!(remote_url: "", minimum_discourse_version: "99.99.99")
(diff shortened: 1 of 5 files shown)


rrooding:
The failures don't seem related to the PR but I can't restart the tests in Travis.

The failing tests are both Qunit tests for something this PR doesn't touch

spec/models/theme_spec.rb
```diff
@@ -59,6 +59,8 @@
 
   end
 
+  it { is_expected.to belong_to(:remote_theme).dependent(:destroy) }
```
eviltrout: Generally we find such tests to be of little value (since they are 1:1 with code) and do not include them.

rrooding: Removed 👍

eviltrout:
Thanks, merged!

PR #7852, opened 2019-07-03
FEATURE: opt-in guidance on topics for users without access

jjaffeux:
Screenshot 2019-07-03 at 17 31 47

https://github.com/discourse/discourse/pull/7521

### DIFF
--- a/app/assets/javascripts/discourse/routes/topic-from-params.js.es6
+++ b/app/assets/javascripts/discourse/routes/topic-from-params.js.es6
@@ -84,9 +84,11 @@ export default Discourse.Route.extend({
   },
 
   actions: {
-    willTransition(transition) {
-      const url = transition.router.generate(transition.from.name);
-      this.controllerFor("topic").set("previousURL", url);
+    willTransition() {
+      this.controllerFor("topic").set(
+        "previousURL",
+        document.location.pathname
+      );
     }
   }
 });
--- a/app/assets/javascripts/discourse/routes/topic-from-params.js.es6
+++ b/app/assets/javascripts/discourse/routes/topic-from-params.js.es6
@@ -35,6 +35,11 @@ export default Discourse.Route.extend({
         // TODO we are seeing errors where closest post is null and this is exploding
         // we need better handling and logging for this condition.
 
+        // there are no closestPost for hidden topics
+        if (topic.view_hidden) {
+          return;
+        }
+
         // The post we requested might not exist. Let's find the closest post
         const closestPost = postStream.closestPostForPostNumber(
           params.nearPost || 1
@@ -76,5 +81,12 @@ export default Discourse.Route.extend({
           console.log("Could not view topic", e);
         }
       });
+  },
+
+  actions: {
+    willTransition(transition) {
+      const url = transition.router.generate(transition.from.name);
+      this.controllerFor("topic").set("previousURL", url);
+    }
   }
 });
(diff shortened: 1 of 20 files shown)


eviltrout:
Is there a related discussion for this? Some context would be nice.

app/assets/javascripts/discourse/routes/topic-from-params.js.es6
```diff
@@ -76,5 +81,12 @@ export default Discourse.Route.extend({
           console.log("Could not view topic", e);
         }
       });
+  },
+
+  actions: {
+    willTransition(transition) {
+      const url = transition.router.generate(transition.from.name);
```
eviltrout: Could we get the URL from `document.location` instead since it's not private APIs?

jjaffeux:
> Is there a related discussion for this? Some context would be nice.



> Is there a related discussion for this? Some context would be nice.

This is basically https://github.com/discourse/discourse/pull/7521 but I took ownership of it

app/controllers/groups_controller.rb
```diff
@@ -420,12 +420,19 @@ def remove_member
   end
 
   def request_membership
-    params.require(:reason)
+    params.require(:reason) if !params[:topic_id].present?
```
eviltrout:
```suggestion
    params.require(:reason) if params[:topic_id].blank?
```

app/controllers/groups_controller.rb
```diff
@@ -420,12 +420,19 @@ def remove_member
...
 
     group = find_group(:id)
 
+    if params[:topic_id].present?
+      topic_url = Topic.find(params[:topic_id].to_i).url
```
eviltrout: This will raise an exception if that topic_id does not exist won't it?

config/locales/client.en.yml
```diff
@@ -1967,6 +1967,9 @@ en:
       toggle_information: "toggle topic details"
       read_more_in_category: "Want to read more? Browse other topics in {{catLink}} or {{latestLink}}."
       read_more: "Want to read more? {{catLink}} or {{latestLink}}."
+      group_request: "You need to request membership to the {{name}} group to see this topic"
+      group_join: "You need join the {{name}} group to see this topic"
+      group_request_sent: "Your group membership request has been sent, you will be informed when it's accepted."
```
```suggestion
      group_request_sent: "Your group membership request has been sent. You will be informed when it's accepted."
```

config/locales/server.en.yml
```diff
@@ -381,6 +381,7 @@ en:
     request_membership_pm:
       title: "Membership Request for @%{group_name}"
       handle: "handle membership request"
+    view_hidden_topic_request_reason: "I would like to join the group so I may access [this topic](%{topic_url})"
```
eviltrout: Should it say the group name here? Makes it extra clear to the approver.

app/controllers/groups_controller.rb
jjaffeux: [1] pry(main)> {}[:topic_id].present?
=> false

jjaffeux: I guess I would write this way:

```ruby
if params[:topic_id] && topic = Topic.find(params[:topic_id].to_i)
  reason = I18n.t("groups.view_hidden_topic_request_reason", topic_url: topic.url)
end
```

eviltrout: Yes the `present?` check is fine, I was more worried if the id was `10123012031203` and no topic could be returned.

PR #7855, opened 2019-07-03
FIX: Respect the full_screen_login parameter from plugin auth providers

davidtaylorhq:
This behavior was regressed in 427979e7e51bb9bc858eb6ad2456ce0075fdc643

@techAPJ I'm not sure if this was intentional or not. Please can you review?


techAPJ:
This was not intentional. Your change looks good. 👍

PR #7856, opened 2019-07-04
FIX: Prevent emoji-picker from not showing

SantosGuillamot:
If an external plugin inserts an element with class "emoji-picker", something probable (and already happening) if they extend EmojiPicker, it could cause troubles as css is added depending on the emoji-picker height. Just by adding a class of a parent <div> as could be d-editor, we prevent this from happening.


jjaffeux:
Seems reasonable to me, thx.

SantosGuillamot:
Thank you for the quick fix! ☺️

PR #7859, opened 2019-07-04
FIX: ensures routing to / with query string works

jjaffeux:
Note this commit also adds polyfills for String.prototype.startsWith and String.prototype.endsWith


jjaffeux:
@eviltrout this is fixing https://meta.discourse.org/t/state-links-on-my-preferences-categories-not-working/121805

I could go without startsWith (and endsWith) but I think it makes sense to have this globally available so I use this opportunity to polyfill it. Let me know if you prefer me to:

1- do not use the polyfill
2- add the polyfill in a different commit
3- leave it unchanged 🎉

eviltrout:
@jjaffeux I would prefer to avoid the polyfill since we can do `path.substring(0, 2) == "/?` which is not significantly worse code.

PR #7860, opened 2019-07-04
DEV: group_list site settings should store IDs instead of group names

romanrizzi:
#### Why do we need this?

Identifying groups by name is not reliable enough when you want to use defaults since they could have been renamed. IDs are a better alternative because they are unique.

After some discussion [here](https://github.com/discourse/discourse-assign/pull/37), we decided that we should change core to simplify the plugin logic and use defaults safely.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -575,9 +575,6 @@ groups:
   enable_category_group_review:
     client: true
     default: false
-  group_list_changed:
-    hidden: true
-    default: true
 
 posting:
   min_post_length:
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -575,6 +575,9 @@ groups:
   enable_category_group_review:
     client: true
     default: false
+  group_list_changed:
+    hidden: true
+    default: true
 
 posting:
   min_post_length:
--- a/db/migrate/20190708153331_migrate_group_list_site_settings.rb
+++ b/db/migrate/20190708153331_migrate_group_list_site_settings.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class MigrateGroupListSiteSettings < ActiveRecord::Migration[5.2]
+  def up
+    migrate_value(:name, :id)
+  end
+
+  def down
+    migrate_value(:id, :name)
+  end
+
+  def migrate_value(from, to)
+    cast_type = from == :id ? '::int[]' : ''
+    DB.exec <<~SQL
+      UPDATE site_settings
+      SET value = array_to_string(
+        (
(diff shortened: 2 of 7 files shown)


eviltrout:
My first question is often: is there a context for this change? It always helps to tie it back to a post on one of our forums in case someone in the future wants to know why we did this.

Secondly, as you noticed in your plugin follow ups this is a breaking change. Did you consider adding a new site setting type, for example, `group_id`.

In that case we can use group ids for new settings without breaking old ones in plugins.

nlalonde:
I'm guessing the reason for this is because group names are visible in client code.

romanrizzi:
I updated the plugin description with some context. Only `discourse-adplugin` and `discourse-assign` use this setting type. Maintaining two different setting types will be confusing since we should always use IDs to reference groups.

config/site_settings.yml
```diff
@@ -575,6 +575,9 @@ groups:
   enable_category_group_review:
     client: true
     default: false
+  group_list_changed:
```
SamSaffron: I support something here, but using a site setting for feature detection is a bit overboard imo.

Technically what happened here is that this API went up: 

https://github.com/discourse/discourse/blob/b9a2a4d904ee28ff458fe34577e433ca719b2174/app/assets/javascripts/discourse/lib/plugin-api.js.es6#L47-L48

So we are now at 0.8.32

So bump that up, and figure out a way of letting the server know about this as well, something reasonably crude can work here like splitting off PLUGIN_API_VERSION into a dedicated JS file and then having `Discourse.client_plugin_api` read the file.

This makes it far cleaner to make decisions in plugins.

romanrizzi: The server will always return IDs. The plugin API version will be used by plugins to assume which value they are dealing with (IDs or names), and also decide if they have to migrate.

All new code using this type of site setting must assume IDs are going to be stored.

db/migrate/20190708153331_migrate_group_list_site_settings.rb
```diff
@@ -0,0 +1,21 @@
...
+    migrate_value(:name, :id)
+  end
+
+  def down
+    migrate_value(:id, :name)
+  end
+
+  def migrate_value(from, to)
+    ActiveRecord::Base.transaction do
+      SiteSetting.where(data_type: SiteSettings::TypeSupervisor.types[:group_list]).each do |setting|
```
eviltrout: You should only use SQL in migrations for queries. We learned this the hard way when in the future the objects change and the migration no longer runs.

romanrizzi: Done 👌

PR #7862, opened 2019-07-06
FIX: Show same username or name for post notices.

nbianca:


app/assets/javascripts/discourse/widgets/post.js.es6
```diff
@@ -453,9 +453,12 @@ createWidget("post-notice", {
 
   html(attrs) {
     const user =
-      this.siteSettings.prioritize_username_in_ux || !attrs.name
-        ? attrs.username
-        : attrs.name;
+      this.siteSettings.display_name_on_posts &&
```
ZogStriP: Can we extract this logic? It's being used in 4 places at least.

PR #7863, opened 2019-07-06
FIX: Ensure suppressed categories do not produce any featured topics.

nbianca:


app/models/category_featured_topic.rb
```diff
@@ -49,7 +49,8 @@ def self.feature_topics_for(c, existing = nil)
       per_page: c.num_featured_topics,
       except_topic_ids: [c.topic_id],
       visible: true,
-      no_definitions: true
+      no_definitions: true,
+      exclude_category_ids: Category.where(suppress_from_latest: true).pluck(:id)
```
ZogStriP: Is that generating an N+1?

app/models/category_featured_topic.rb
```diff
@@ -38,6 +38,16 @@ def self.feature_topics(batched: false, batch_size: nil)
...
   end
 
+  @@exclude_category_ids = DistributedCache.new('excluded_category_ids_from_featured')
+
+  def self.cached_exclude_category_ids
+    @@exclude_category_ids['ids'] ||= Category.where(suppress_from_latest: true).pluck(:id)
+  end
+
+  def self.clear_exclude_category_ids
+    @@exclude_category_ids['ids'] = nil
```
ZogStriP:
```suggestion
    @@exclude_category_ids.clear
```

app/models/category_featured_topic.rb
```diff
@@ -38,6 +38,16 @@ def self.feature_topics(batched: false, batch_size: nil)
     end
   end
 
+  @@exclude_category_ids = DistributedCache.new('excluded_category_ids_from_featured')
+
+  def self.cached_exclude_category_ids
+    @@exclude_category_ids['ids'] ||= Category.where(suppress_from_latest: true).pluck(:id)
```
ZogStriP:
```suggestion
    @@exclude_category_ids = Category.where(suppress_from_latest: true).pluck(:id)
```

PR #7866, opened 2019-07-08
FIX: Do not show bootbox if post has no replies.

nbianca:


ZogStriP:
Can you also add a test?

PR #7867, opened 2019-07-08
FEATURE: show login and signup button on no-ember layout

techAPJ:
This PR adds signup and login button to no-ember layout pages for anonymous users.

Feature requested here:

- https://meta.discourse.org/t/how-can-users-sign-up-for-or-log-into-the-forum-when-they-encounter-page-not-found/120945/5?u=techapj
- https://meta.discourse.org/t/missing-log-in-button-on-error-pages/110659/5?u=techapj

Demo:

Screen Shot 2019-07-08 at 16 32 51

### DIFF
--- a/app/assets/stylesheets/common/base/header.scss
+++ b/app/assets/stylesheets/common/base/header.scss
@@ -15,6 +15,23 @@
       display: flex;
       align-items: center;
       height: 100%;
+
+      .header-row {
+        width: 100%;
+
+        .logo-wrapper {
+          float: left;
+        }
+        .auth-buttons {
+          float: right;
+          margin-top: 0.4em;
+
+          .login-button,
+          .signup-button {
+            padding: 8px 14px;
+          }
+        }
+      }
     }
   }
 
--- a/app/views/application/_header.html.erb
+++ b/app/views/application/_header.html.erb
@@ -2,7 +2,7 @@
   <header class="d-header">
     <div class="wrap">
       <div class="contents">
-        <div class="row">
+        <div class="header-row">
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/common/base/header.scss
```diff
@@ -15,6 +15,23 @@
       display: flex;
       align-items: center;
       height: 100%;
+
+      .header-row {
+        width: 100%;
+
+        .logo-wrapper {
```
eviltrout: One thing to consider here is using flexbox. It's not required for this PR but I find it's much easier to wrap your head around than floats.

app/views/application/_header.html.erb
```diff
@@ -12,6 +12,12 @@
               <%- end %>
             </a>
           </div>
+          <% unless current_user %>
+            <div class='auth-buttons'>
+              <a href="<%= path "/signup"%>" class='btn btn-primary btn-small signup-button'><%= I18n.t('sign_up') %></a>
```
eviltrout: Shouldn't there be some logic here to ensure that users can sign up? For example the front end app does this:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/controllers/application.js.es6#L9

app/assets/stylesheets/common/base/header.scss
techAPJ: Noted. Will start using that.

app/views/application/_header.html.erb
techAPJ: Good catch, done!

PR #7868, opened 2019-07-08
DEV: Upgrades to Ember 3.10

jjaffeux:
Co-Authored-By: majakomel <maja.komel@gmail.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,7 @@ gem 'redis-namespace'
 
 gem 'active_model_serializers', '~> 0.8.3'
 
-gem 'onebox', '1.9.0'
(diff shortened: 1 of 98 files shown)


eviltrout:
A commit this large is basically impossible to approve with a high level of confidence, so I think our best approach should be to merge it following a beta release. I'm sure there will be regressions but we can fix them quickly as follow up commits.

jjaffeux:
> A commit this large is basically impossible to approve with a high level of confidence, so I think our best approach should be to merge it following a beta release. I'm sure there will be regressions but we can fix them quickly as follow up commits.

I will first do the usual "push on meta"

PR #7869, opened 2019-07-09
UX: Discard selected post if it is not in viewport.

udan11:
This way, users can combine keyboard shortcuts with mouse scrolling.


eviltrout:
Can you explain this a little more?

Is the bug here that if a post is selected with the keyboard and you scroll down, hitting a key will move the selection down from above and not in the current viewport?

udan11:
Right now, if the user uses J/K to navigate and then scrolls using the mouse, when they press J/K again it will just jump to where they left off (the next / prev post).

After these changes are applied, a user can use J/K, scroll with the mouse and then use J/K again.

This has been suggested by a user on Meta: https://meta.discourse.org/t/pressing-j-k-in-topic-breaks-if-user-scrolls-with-mouse/116816/19

PR #7870, opened 2019-07-09
Make Discourse work while offline

xfalcox:


app/assets/javascripts/service-worker.js.erb
```diff
@@ -1,117 +1,25 @@
 'use strict';
 
+importScripts("<%= ::UrlHelper.absolute("/javascripts/workbox/workbox-sw.js") %>");
```
eviltrout: A couple questions here:

1. Do we need all the included files below? It seems like quite a few and I suspect not all of them are used by this include. It's hard to tell though.

2. Can this be installed via `yarn`? We've moving other dependencies to that approach.

xfalcox: > Do we need all the included files below? It seems like quite a few and I suspect not all of them are used by this include. It's hard to tell though.

No, currently we don't use all of those files.

What I did was I got the WorkboxJS release, and extracted all the contents to `public/javascripts/workbox/`, kinda like we do with Ace.

In the code we import the main file `/javascripts/workbox/workbox-sw.js`, and that will lazily import the necessary modules, based in what we actually use in our script.

Currently it is:

image


> 2) Can this be installed via yarn? We've moving other dependencies to that approach.

We indeed are, but I found no published package in NPM for the released files.

Looks like most people use it by either:

- Hotlinking from the Google CDN

- Using their custom webpack integration.

And their current releases cover only those uses. So I went with the old school, download the release, and unpack in the folder removing all previous files.

pmusaraj: @xfalcox I had initially looked at this and given up, but I just had a second look and I think we can add these via `yarn`. Each module is a separate package in npm/yarn. For example: https://yarnpkg.com/en/package/workbox-core. When added via `yarn add workbox-core` it has the build files in `node_modules/workbox-core/build/`. 

So we would need to select the modules used, and modify our javascript rake task to copy the selected files to `public/` or `vendor/`. 

Let me know if I can help further.

xfalcox: AWESOME tip @pmusaraj ! 

Just force pushed with all being handled by yarn now!

xfalcox:
@eviltrout let's merge after next beta?

eviltrout:
Sounds good to me! The ember upgrade is going in at the same time so we'll have to take that into account when triaging issues.

PR #7872, opened 2019-07-10
FIX: ensures routing with hash doesn't stuck history

jjaffeux:
Original issue: https://meta.discourse.org/t/hash-anchor-in-url-prevents-further-url-updates/122068/4

Basically when the path has a hash, state would be null, and nothing would happen.

### DIFF
--- a/app/assets/javascripts/discourse/lib/discourse-location.js.es6
+++ b/app/assets/javascripts/discourse/lib/discourse-location.js.es6
@@ -101,7 +101,7 @@ const DiscourseLocation = Ember.Object.extend({
     const state = this.getState();
     path = this.formatURL(path);
 
-    if (state && state.path !== path) {
+    if (!state || (state && state.path !== path)) {
       this.replaceState(path);
     }
   },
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/lib/discourse-location.js.es6
```diff
@@ -101,7 +101,7 @@ const DiscourseLocation = Ember.Object.extend({
     const state = this.getState();
     path = this.formatURL(path);
 
-    if (state && state.path !== path) {
+    if (!state || (state && state.path !== path)) {
```
ZogStriP: Shouldn't this be?

```suggestion
    if (!state || state.path !== path) {
```

jjaffeux: yes kinda... I think it’s less prone to error in the future to have two clear conditions, instead of relying on the first condition to validate the second...

But I can see why you wouldn’t agreee 😅

jjaffeux: Up to you, both are fine to me

eviltrout: I think @ZogStriP 's is more clear, let's go with that.

PR #7873, opened 2019-07-10
FIX: Heartbeat check per sidekiq process

OsamaSayegh:


lib/demon/sidekiq.rb
```diff
@@ -3,6 +3,49 @@
...
+
+  def self.heartbeat_queues_list_key
+    @@heartbeat_queues_list_key ||= "#{RANDOM_HEX}_heartbeat_queues_list"
+  end
+
+  def self.queues_last_heartbeat_hash_key
+    @@queues_last_heartbeat_hash_key ||= "#{RANDOM_HEX}_queues_last_heartbeat_hash"
+  end
+
+  def self.set_queue_heartbeat(name)
```
eviltrout: I think we should rename this method, perhaps `trigger_heartbeat`. When I think `set_xyz` it sounds to me like replacing the member `xyz` with something else.

config/unicorn.conf.rb
```diff
@@ -144,25 +144,35 @@ def check_sidekiq_heartbeat
...
-              restart = true
+            if !restarted
+              Demon::Sidekiq.heartbeat_queues.each do |queue|
+                last_heartbeat = Demon::Sidekiq.get_queue_last_heartbeat(queue)
+
+                if last_heartbeat < Time.new.to_i - @sidekiq_heartbeat_interval
+                  pid = queue.split("_").last.to_i
+                  STDERR.puts "Sidekiq heartbeat test for worker #{pid} failed, restarting"
+                  Rails.logger.warn "Sidekiq heartbeat test for worker #{pid} failed, restarting"
+                  Demon::Sidekiq.demons.values.each do |demon|
```
ZogStriP: Since you're only looking for one item in the list, you could do something like this

```ruby
if demon = Demon::Sidekiq.demons.values.find { |d| d.pid == pid }
  demon.stop
  demon.start
  restarted = true
end
```

lib/demon/sidekiq.rb
```diff
@@ -3,6 +3,49 @@
...
+    $redis.hset(queues_last_heartbeat_hash_key, name, Time.new.to_i.to_s)
+  end
+
+  def self.get_queue_last_heartbeat(name)
+    $redis.hget(queues_last_heartbeat_hash_key, name)&.to_i || 0
```
ZogStriP: You don't need the `&.` or the `|| 0` when using `to_i` since `nil.to_i == 0` 😉 

```suggestion
    $redis.hget(queues_last_heartbeat_hash_key, name).to_i
```

lib/demon/sidekiq.rb
```diff
@@ -3,6 +3,57 @@
...
+
+  def self.clear_heartbeat_queues!
+    $redis.del(heartbeat_queues_list_key)
+    $redis.del(queues_last_heartbeat_hash_key)
+  end
+
+  def self.before_start
+    # cleans up heartbeat queues from previous boot up
+    Sidekiq::Queue.all.each do |queue|
+      next if queue.name !~ /^[a-f0-9]{32}_\d+$/
```
ZogStriP:
```suggestion
      next if queue.name !~ /^\h{32}_\d+$/
```

lib/demon/sidekiq.rb
```diff
@@ -3,6 +3,57 @@
...
+      queue.clear if queue.size == 0
```
ZogStriP: Shouldn't that be?

```suggestion
      queue.clear if queue.size > 0
```

ZogStriP: Or just

```suggestion
      queue.clear
```

lib/demon/sidekiq.rb
```diff
@@ -3,6 +3,57 @@
...
+  end
+
+  def self.clear_heartbeat_queues!
+    $redis.del(heartbeat_queues_list_key)
+    $redis.del(queues_last_heartbeat_hash_key)
+  end
+
+  def self.before_start
+    # cleans up heartbeat queues from previous boot up
+    Sidekiq::Queue.all.each do |queue|
```
ZogStriP: Maybe this could be refactored like this?

``` ruby
Sidekiq::Queue.all.each { |queue| queue.clear if queue.name[/^\h{32}_\d+$/] }
```

lib/demon/sidekiq.rb
OsamaSayegh: Yeah I think we can just drop this condition altogether because the queue name already matches the name of the special heartbeat queues.

OsamaSayegh:
@eviltrout the way I tested this change is I made sidekiq boot up without any queues locally by passing an empty array to `cli.parse` here:
https://github.com/discourse/discourse/blob/fb2df0b43bf9588f56a2f23a22dc61424771d6a4/lib/demon/sidekiq.rb#L52

that way sidekiq couldn't process the jobs we enqueue here:

https://github.com/discourse/discourse/blob/fb2df0b43bf9588f56a2f23a22dc61424771d6a4/app/jobs/scheduled/heartbeat.rb#L6-L12

Then I lowered the heartbeat check interval to 5 seconds (so that I don't have to wait 30 minutes) and made the `Heartbeat` job run every 2 seconds. After that I ran `UNICORN_SIDEKIQS=2 ./bin/unicorn` and shortly after that the `check_sidekiq_heartbeat` method ran and the 2 sidekiq processes that were spawned restarted. Does this sound like correct/enough testing?

-----

There is still one problem that I'd like address before merging this: the problem is that every time the master unicorn boots up, 2 unique keys are created in redis (`heartbeat_queues_list_key` and `queues_last_heartbeat_hash_key`). This means that if there are 3 app servers and you hit deploy, 6 redis keys are created per deploy. Over time this would accumulate lots of redis keys that aren't used at all. I'd like to clean up keys as they become unneeded.

One way to do that would to prefix the 2 keys with the server `hostname`, and then in the `before_start` method (I just added it) we can iterate over all redis keys and delete the ones that aren't needed. The prefix is needed to ensure no server deletes other servers' keys. That way each time master unicorn boots up, it'd clean up the keys of the previous boot up. Does that sound good? Is there a better approach to this?

eviltrout:
> Does that sound good? Is there a better approach to this?

Keeping track of what keys you've set in the past can be messy, and it wouldn't account for the situation where we retire a host and its previous keys would exist forever.

Instead I recommend giving the keys an EXPIRY, which you can bump whenever the heartbeat runs. If the heartbeat runs every 3 minutes for example, you could set it as expiring in 60 minutes or something just to be very careful.

After one hour of not being used, the keys will vanish.

config/unicorn.conf.rb
```diff
@@ -144,25 +144,33 @@ def check_sidekiq_heartbeat
...
+              restarted = true
             end
 
-            if last_heartbeat < Time.new.to_i - @sidekiq_heartbeat_interval
-              STDERR.puts "Sidekiq heartbeat test failed, restarting"
-              Rails.logger.warn "Sidekiq heartbeat test failed, restarting"
-
-              restart = true
+            if !restarted
+              Demon::Sidekiq.heartbeat_queues.each do |queue|
```
SamSaffron: hmmm is this checking all heartbeat queues? Should it not just check heartbeats for child sidekiq processes it has?

SamSaffron:
Did you test with unicorn_sidekiqs set to 2 to see how it properly spawns and monitors? What are the blockers for a merge here, overall this is a fantastic change!

OsamaSayegh:
> Did you test with unicorn_sidekiqs set to 2 to see how it properly spawns and monitors? What are the blockers for a merge here, overall this is a fantastic change!

Yes I did test with `UNICORN_SIDEKIQS`, I posted details of how I tested this change above https://github.com/discourse/discourse/pull/7873#issuecomment-510970132.

config/unicorn.conf.rb
OsamaSayegh: `Demon::Sidekiq.heartbeat_queues` returns all sidekiq processes that are spawned by unicorn master and we check their heartbeat one by one here. Is this not what we supposed to do?

SamSaffron: Oh I see... so each process will give a different list here ...

A confusing thing here is that if you call `Demon::Sidekiq.heartbeat_queues` you will get one list .. and if you call this on a forked child process you will get an empty list. Maybe we call this `Demon::Sidekiq.child_process_heartbeat_queues` ?

SamSaffron:
I am really liking this change ... all I seem to have here is superficial comments, I would like you to merge in the beginning of your day, then watch on meta... then deploy another bigger cluster and watch on it as well. Critical that this gets merged when you have a few hours to double check we are not getting a swamp of "restarting child process cause it died"

SamSaffron:
@OsamaSayegh can you have a look to see if you can merge it this week?

OsamaSayegh:
@SamSaffron I've amended this PR so now it keeps track of the special queues using a constant `QUEUE_IDS` rather than a redis list. Let me know what you think.

SamSaffron:
I think this looks good, can you merge it monday?

OsamaSayegh:
Sure, will do

PR #7874, opened 2019-07-10
Revert "FEATURE: admin/user exports are compressed using the zip format"

romanrizzi:
Reverts discourse/discourse#7784


eviltrout:
If you're reverting I don't think you need a PR unless the revert is doing something odd. Just go ahead!

romanrizzi:
Just using the Github UI 😛. Will revert directly.

PR #7875, opened 2019-07-10
FEATURE: send max 200 emails every minute for bulk invites

techAPJ:
DEV: deprecate `invite.via_email` in favor of `invite.emailed_status`

This commit adds a new column `emailed_status` in `invites` table for
 tracking email sending status.
 0 - not required
 1 - pending
 2 - bulk pending
 3 - sending
 4 - sent

For normal email invites, invite record is created with emailed_status
 set to 'pending'.

When bulk invites are sent invite record is created with emailed_status
 set to 'bulk pending'.

For invites that generates link, invite record is created with
 emailed_status set to 'not required'.

When invite email is in queue emailed_status is updated to 'sending'

Once the email is sent via `InviteEmail` job the invite emailed_status
 is updated to 'sent'.

### DIFF
--- a/app/jobs/regular/bulk_invite.rb
+++ b/app/jobs/regular/bulk_invite.rb
@@ -23,8 +23,13 @@ module Jobs
       @current_user = User.find_by(id: args[:current_user_id])
       raise Discourse::InvalidParameters.new(:current_user_id) unless @current_user
       @guardian = Guardian.new(@current_user)
+      @total_invites = invites.length
 
       process_invites(invites)
+
+      if @total_invites > Invite::BULK_INVITE_EMAIL_LIMIT
+        Jobs.enqueue(:process_bulk_invite_emails)
+      end
     ensure
       notify_user
     end
@@ -104,7 +109,15 @@ module Jobs
             end
           end
         else
-          Invite.invite_by_email(email, @current_user, topic, groups.map(&:id))
+          if @total_invites > Invite::BULK_INVITE_EMAIL_LIMIT
+            invite = Invite.create_invite_by_email(email, @current_user,
+              topic: topic,
+              group_ids: groups.map(&:id),
+              emailed_status: Invite.emailed_status_types[:bulk_pending]
+            )
+          else
+            Invite.invite_by_email(email, @current_user, topic, groups.map(&:id))
+          end
         end
       rescue => e
         save_log "Error inviting '#{email}' -- #{Rails::Html::FullSanitizer.new.sanitize(e.message)}"
--- a/app/jobs/regular/invite_email.rb
+++ b/app/jobs/regular/invite_email.rb
@@ -15,8 +15,10 @@ module Jobs
 
       message = InviteMailer.send_invite(invite)
       Email::Sender.new(message, :invite).send
+
+      if invite.emailed_status != Invite.emailed_status_types[:not_required]
+        invite.update_column(:emailed_status, Invite.emailed_status_types[:sent])
+      end
     end
-
   end
-
 end
--- a/app/jobs/regular/process_bulk_invite_emails.rb
+++ b/app/jobs/regular/process_bulk_invite_emails.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require_dependency 'email/sender'
+
+module Jobs
+
+  class ProcessBulkInviteEmails < Jobs::Base
+
+    def execute(args)
+      pending_invite_ids = Invite.where(emailed_status: Invite.emailed_status_types[:bulk_pending]).limit(Invite::BULK_INVITE_EMAIL_LIMIT).pluck(:id)
+
+      if pending_invite_ids.length > 0
+        Invite.where(id: pending_invite_ids).update_all(emailed_status: Invite.emailed_status_types[:sending])
+        pending_invite_ids.each do |invite_id|
+          Jobs.enqueue(:invite_email, invite_id: invite_id)
+        end
+        Jobs.enqueue_in(1.minute, :process_bulk_invite_emails)
+      end
+    end
+  end
+end
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -3,9 +3,15 @@
 require_dependency 'rate_limiter'
 
 class Invite < ActiveRecord::Base
+  self.ignored_columns = %w{
+    via_email
+  }
+
   class UserExists < StandardError; end
   include RateLimiter::OnCreateRecord
   include Trashable
+
+  BULK_INVITE_EMAIL_LIMIT = 200
 
   rate_limit :limit_invites_per_day
 
@@ -30,6 +36,10 @@ class Invite < ActiveRecord::Base
 
   validate :user_doesnt_already_exist
   attr_accessor :email_already_exists
+
+  def self.emailed_status_types
+    @emailed_status_types ||= Enum.new(not_required: 0, pending: 1, bulk_pending: 2, sending: 3, sent: 4)
+  end
 
   def user_doesnt_already_exist
     @email_already_exists = false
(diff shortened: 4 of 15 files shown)


app/jobs/regular/bulk_invite.rb
```diff
@@ -104,7 +109,16 @@ def send_invite(invite)
             end
           end
         else
-          Invite.invite_by_email(email, @current_user, topic, groups.map(&:id))
+          if @total_invites > Invite::BULK_INVITE_EMAIL_LIMIT
+            invite = Invite.create_invite_by_email(email, @current_user,
+              topic: topic,
+              group_ids: groups.map(&:id),
+              send_email: false,
```
eviltrout: Do we still need `send_email` now that we have email status?

app/jobs/regular/process_bulk_invite_emails.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require_dependency 'email/sender'
+
+module Jobs
+
+  class ProcessBulkInviteEmails < Jobs::Base
+
+    def execute(args)
+      bulk_invite_email_key = Invite::BULK_INVITE_EMAIL_KEY
```
eviltrout: Looks like this variable is unused? Probably the constant too.

app/jobs/regular/process_bulk_invite_emails.rb
```diff
@@ -0,0 +1,23 @@
...
+      pending_invite_ids = Invite.where(emailed_status: Invite.emailed_status_types[:pending]).limit(Invite::BULK_INVITE_EMAIL_LIMIT).pluck(:id)
+
+      if pending_invite_ids.length > 0
+        Invite.where(id: pending_invite_ids).update_all(emailed_status: Invite.emailed_status_types[:sending])
+        pending_invite_ids.each do |invite_id|
+          Jobs.enqueue(:invite_email, invite_id: invite_id)
+        end
+        Invite.where(id: pending_invite_ids).update_all(via_email: true, emailed_status: Invite.emailed_status_types[:sent])
```
eviltrout: `Jobs.enqueue` should be sending the emails asynchronously so you can't update immediately afterwards and know that the emails were all sent. Instead the status should be updated by the `:invite_email` job.

app/jobs/regular/process_bulk_invite_emails.rb
techAPJ: Good catch, fixed!

app/jobs/regular/bulk_invite.rb
techAPJ: Yes `send_email` is required here as this argument decides if email should be sent or not, as per [this code](https://github.com/discourse/discourse/blob/e7fe7010b8dcea495a46089c9fd5ec215f3de46e/app/models/invite.rb#L146).

app/jobs/regular/process_bulk_invite_emails.rb
techAPJ: Understood, will update PR to do so.

techAPJ: PR updated.

app/jobs/regular/bulk_invite.rb
eviltrout: Understood, but if we are storing the status of the invite, including a new status for "not required", we could use that instead of a `send_email` parameter. Code that previously passed `send_mail` as true/false could instead use set the correct status.

app/jobs/regular/process_bulk_invite_emails.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def execute(args)
+      pending_invite_ids = Invite.where(emailed_status: Invite.emailed_status_types[:pending]).limit(Invite::BULK_INVITE_EMAIL_LIMIT).pluck(:id)
+
+      if pending_invite_ids.length > 0
+        Invite.where(id: pending_invite_ids).update_all(emailed_status: Invite.emailed_status_types[:sending])
+        pending_invite_ids.each do |invite_id|
+          Jobs.enqueue(:invite_email, invite_id: invite_id)
+        end
+        Jobs.enqueue_in(1.minute, :process_bulk_invite_emails)
```
eviltrout: I wonder if `1.minute` is enough for 200 emails to be sent?

app/jobs/regular/bulk_invite.rb
techAPJ: I see. Will update the PR to deprecate `invite.via_email` in favor of `invite.emailed_status`

app/jobs/regular/process_bulk_invite_emails.rb
techAPJ: That is what Sam recommended initially. Maybe we should send max 100 invites in a minute?

techAPJ:
@eviltrout PR is updated to deprecate `invite.via_email` in favor of `invite.emailed_status`.

app/jobs/regular/invite_email.rb
```diff
@@ -15,8 +15,11 @@ def execute(args)
 
       message = InviteMailer.send_invite(invite)
       Email::Sender.new(message, :invite).send
-    end
 
+      if invite.emailed_status == Invite.emailed_status_types[:sending]
```
eviltrout: Is there a good reason to only update the status to `sent` if it was sending? I feel like if the email was sent it could always be updated.

app/models/invite.rb
```diff
@@ -66,7 +76,7 @@ def self.invite_by_email(email, invited_by, topic = nil, group_ids = nil, custom
       topic: topic,
       group_ids: group_ids,
       custom_message: custom_message,
-      send_email: true
+      emailed_status: emailed_status_types[:sending]
```
eviltrout: It seems odd to create an invite with a status of `sending`. It would only seem appropriate to be `sending` to me if the email is queued for sending. At this point it makes more sense to be `pending.`

app/jobs/regular/invite_email.rb
techAPJ: The only exception I think where we should not update the status to `sent` is when the status is `not required`. That is because when the invite link is generated we set invite `emailed_status` to 'not required'. Now the inviter has invite link with him. If inviter sends same invite again via email this time the `emailed_status` will be updated to 'sent' and that invite link (generated via invite link feature) can be used to create an active account (as per [this code](https://github.com/discourse/discourse/pull/7875/files#diff-2a355fe6078d7a0ee0e440532e4032e2R63)) without verifying the email which can be classified as security issue.

I will change to code to update status to `sent` if the original status was `pending` or `sending`.

app/models/invite.rb
techAPJ: `pending` status is reserved for bulk invites. If I set the status to `pending` here then invite email will not be sent instantaneously and since we are not queuing the `ProcessBulkInviteEmails` job here the email will be sent only after a bulk invite process is triggered.

Here is how I am using the `emailed_status` column for different types of invites (copied from PR description):

> For normal email invites, invite record is created with emailed_status
> set to 'sending'.
> 
> When bulk invites are sent invite record is created with emailed_status
> set to 'pending'.
> 
> For invites that generates link, invite record is created with
> emailed_status set to 'not required'.
> 
> Once the email is sent via InviteEmail job the invite emailed_status
> is updated to 'sent'.

eviltrout: Logically it confuses me that if you want an email to be sent you would assign `sending` as the status because at that point it is not "sending".

I would suggest this:

Normal invites are created with "pending"

Bulk Invites are created with "bulk_pending"

When an email is being send it's "sending"

When an email has been sent it's "sent"

techAPJ: Understood, will make it so.

techAPJ: PR is updated.

app/jobs/regular/process_bulk_invite_emails.rb
eviltrout: Let's stick with this for now.

techAPJ:
Thanks for the review, @eviltrout. ❤️

PR #7884, opened 2019-07-11
FEATURE: Allow customization of robots.txt

OsamaSayegh:
This allows admins to customize/override the content of the robots.txt
file at /admin/customize/robots. That page is not linked to anywhere in
the UI -- admins have to manually type the URL to access that page.

Meta topic: https://meta.discourse.org/t/needing-to-edit-robots-txt-file-where-is-it/93879?u=osama

Screenshots:

image

image

@eviltrout does it make sense to prepend a comment to robots.txt that says something along the lines of "this robots.txt file has been customized at /admin/customize/robots" **if** the file is customized? It might help with figuring out why certain things are in the file and how to remove/change them?

### DIFF
--- a/app/controllers/robots_txt_controller.rb
+++ b/app/controllers/robots_txt_controller.rb
@@ -33,8 +33,12 @@ class RobotsTxtController < ApplicationController
   }
 
   def index
+    if SiteSetting.overridden_robots_txt.present?
+      render plain: SiteSetting.overridden_robots_txt
+      return
+    end
     if SiteSetting.allow_index_in_robots_txt?
-      @robots_info = fetch_robots_info
+      @robots_info = self.class.fetch_robots_info
       render :index, content_type: 'text/plain'
     else
       render :no_index, content_type: 'text/plain'
@@ -46,12 +50,13 @@ class RobotsTxtController < ApplicationController
   # JSON that can be used by a script to create a robots.txt that works well with your
   # existing site.
   def builder
-    render json: fetch_robots_info
+    result = self.class.fetch_robots_info
+    overridden = SiteSetting.overridden_robots_txt
+    result[:overridden] = overridden if overridden.present?
+    render json: result
   end
 
-protected
-
-  def fetch_robots_info
+  def self.fetch_robots_info
     deny_paths = DISALLOWED_PATHS.map { |p| Discourse.base_uri + p }
     deny_all = [ "#{Discourse.base_uri}/" ]
 
@@ -87,5 +92,4 @@ protected
 
     result
   end
-
 end
--- a/app/controllers/robots_txt_controller.rb
+++ b/app/controllers/robots_txt_controller.rb
@@ -4,6 +4,8 @@ class RobotsTxtController < ApplicationController
   layout false
   skip_before_action :preload_json, :check_xhr, :redirect_to_login_if_required
 
+  OVERRIDDEN_HEADER = "# This robots.txt file has been customized at /admin/customize/robots\n"
+
   # NOTE: order is important!
   DISALLOWED_PATHS ||= %w{
     /auth/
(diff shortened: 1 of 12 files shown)


coding-horror:
how does this interact with the handful of site settings that modify `robots.txt`, such as `whitelisted crawler user agents`, `slow down crawler user agents`, and `allow index in robots txt` ?

OsamaSayegh:
@coding-horror this allows overriding the whole robots.txt file. So if an admin goes to this page and makes some changes, then that's what's going to be served. Changes to the site settings that you mentioned won't apply if there is an overriding copy in the database (admins can remove the overriding copy and restore the default robots.txt). Is this how you think it should work?

coding-horror:
It's fine, we just need to make that clear somewhere, also is there a
"restore" button to revert robots.txt to default?

OsamaSayegh:
Yes, there is a "revert changes" button. I will add a copy to the page to make that clear.

app/controllers/robots_txt_controller.rb
```diff
@@ -46,12 +50,13 @@ def index
...
+    result = self.class.fetch_robots_info
+    overridden = SiteSetting.overridden_robots_txt
+    result[:overridden] = overridden if overridden.present?
+    render json: result
   end
 
-protected
-
-  def fetch_robots_info
+  def self.fetch_robots_info
```
eviltrout: This method now does not return the robots info, it returns the default robots info (other code is supposed to know about this and return an override if it exists.) I would suggest renaming this to `fetch_default_robots_info`

spec/requests/admin/robots_txt_controller_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  describe "non-admin users" do
+    before { sign_in(user) }
+
+    it "can't see #show" do
+      get "/admin/customize/robots.json"
+      expect(response.status).to eq(404)
+    end
+
+    it "can't see #update" do
```
eviltrout:
```suggestion
    it "can't perform #update" do
```

spec/requests/admin/robots_txt_controller_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      put "/admin/customize/robots.json", params: { content: "adasdasd" }
+      expect(response.status).to eq(404)
+      expect(SiteSetting.overridden_robots_txt).to eq("")
+    end
+  end
+
+  describe "#show" do
+    before { sign_in(admin) }
+
+    it "returns default content is there are no overrides" do
```
eviltrout:
```suggestion
    it "returns default content if there are no overrides" do
```

eviltrout:
> does it make sense to prepend a comment to robots.txt that says something along the lines of "this robots.txt file has been customized at /admin/customize/robots" if the file is customized?

The problem is that information is only really useful to staff. You could put in logic that only includes that message if the request is via a staff account, but some people will use those with API keys to generate the robots.txt for the root of a subfolder site.

app/controllers/robots_txt_controller.rb
```diff
@@ -46,12 +50,13 @@ def index
   # JSON that can be used by a script to create a robots.txt that works well with your
   # existing site.
   def builder
-    render json: fetch_robots_info
+    result = self.class.fetch_robots_info
```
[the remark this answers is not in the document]
OsamaSayegh: I'm sorry I'm not sure I understand this point correctly. Do you mean if there is an overridden version, we should try to parse it and send it in JSON format like we do for the default version like this?

```
{
   "header":"# See http://www.robotstxt.org/robotstxt.html ...",
   "agents":[
      {
         "name":"*",
         "disallow":[
            "/auth/",
            "/assets/browser-update*.js",
            ...
         ]
      },
      {
         "name":"mauibot",
         "disallow":[
            "/"
         ]
      },
     ...
   ]
}
```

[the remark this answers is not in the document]
eviltrout: Oooh I see the format is much more complicated than I thought. You can disregard my comment, sorry. Let's change the API the way you suggested.

PR #7886, opened 2019-07-12
add limit to custom field ordering

angusmcleod:
While not likely, it is possible to end up with two rows in a custom field table with the same related id.

See further: https://meta.discourse.org/t/custom-field-casting-affected-by-recent-update/122746

For example: https://discourse.angusmcleod.com.au/t/events-code-error-500/1254/4

The data will look something like this.

``topic_custom_fields``

| id  | topic_id | name | value |
| ------------- | ------------- | ------------- | ------------- |
| 9180 | 62 | event_start | 1563100201 |
| 9179  | 62  | event_start | 1563100201 |

In most respects this eventuality can be handled in a plugin, however there is one place where this causes a cardinality issue that can't be handled in a plugin: ordering by the custom field in a ``TopicQuery``.

This PR adds a LIMIT to the custom field ordering statement in ``TopicQuery`` to handle that eventuality.

### DIFF
--- a/lib/topic_query.rb
+++ b/lib/topic_query.rb
@@ -631,7 +631,7 @@ class TopicQuery
 
     if sort_column.start_with?('custom_fields')
       field = sort_column.split('.')[1]
-      return result.order("(SELECT CASE WHEN EXISTS (SELECT true FROM topic_custom_fields tcf WHERE tcf.topic_id::integer = topics.id::integer AND tcf.name = '#{field}') THEN (SELECT value::integer FROM topic_custom_fields tcf WHERE tcf.topic_id::integer = topics.id::integer AND tcf.name = '#{field}') ELSE 0 END) #{sort_dir}")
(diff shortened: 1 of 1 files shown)


ZogStriP:
I'm sorry, but adding this `LIMIT 1` would just hide bad data. I'd rather it blow up so that one can properly fix the data than have it silently fail.

angusmcleod:
@ZogStriP I understand where you're coming from, but I'm not sure adding a data migration in a plugin is a good solution. If I follow that approach I'll need to add a migration whenever I use custom_field ordering in ``TopicQuery``, to guard against it.

Moreover it would still cause my users' sites to crash whenever the issue rears its head again (i.e. in-between migrations). And the handling of custom field storage may change in the future, invalidating the assumptions the migration is written on. 

I think the solution for me for now is to move ordering inside the plugin itself.

PR #7888, opened 2019-07-12
FEATURE: Add support for secure media

pmusaraj:
This PR introduces a new secure media setting. When enabled, it prevent unathorized access to media uploads (files of type image, video and audio). When the `login_required` setting is enabled, then all media uploads will be protected from unauthorized (anonymous) access. When `login_required`is disabled, only media in private messages will be protected from unauthorized access. 

A few notes: 

- the `prevent_anons_from_downloading_files` setting no longer applies to audio and video uploads
- the `secure_media` setting can only be enabled if S3 uploads are already enabled and configured
- upload records have a new column, `secure`, which is a boolean `true/false` of the upload's secure status
- when creating a public post with an upload that has already been uploaded and is marked as secure, the post creator will raise an error
- when enabling or disabling the setting on a site with existing uploads, the rake task `uploads:ensure_correct_acl` should be used to update all uploads' secure status and their ACL on S3


app/controllers/uploads_controller.rb
```diff
@@ -111,6 +111,17 @@ def show_short
...
   end
 
+  def show_secure
+    # do not serve uploads requested via XHR to prevent XSS
+    return xhr_not_allowed if request.xhr?
+
+    if Discourse.store.secure_images_enabled?
+      redirect_to Discourse.store.signed_url_for_path("#{params[:path]}.#{params[:extension]}")
+    else
+      return render_404
```
eviltrout:
```suggestion
      render_404
```

app/models/upload.rb
```diff
@@ -147,7 +147,12 @@ def local?
 
   def private?
     return false if self.for_theme || self.for_site_setting
-    SiteSetting.prevent_anons_from_downloading_files && !FileHelper.is_supported_image?(self.original_filename)
+
+    if FileHelper.is_supported_image?(self.original_filename)
+      return Discourse.store.secure_images_enabled?
```
eviltrout:
```suggestion
      Discourse.store.secure_images_enabled?
```

app/models/upload.rb
```diff
@@ -147,7 +147,12 @@ def local?
...
+    else
+      return SiteSetting.prevent_anons_from_downloading_files
```
eviltrout:
```suggestion
      SiteSetting.prevent_anons_from_downloading_files
```

config/site_settings.yml
```diff
@@ -280,7 +280,7 @@ basic:
     default: true
   enable_badge_sql:
     client: true
-    default: false
+    default: true
```
eviltrout: Was this enabled by accident? It doesn't seem relevant for this PR.

lib/file_store/s3_store.rb
```diff
@@ -91,6 +92,10 @@ def external?
       true
     end
 
+    def secure_images_enabled?
+      SiteSetting.secure_images && SiteSetting.login_required
```
eviltrout:
```suggestion
      SiteSetting.secure_images? && SiteSetting.login_required?
```

lib/site_settings/validations.rb
```diff
@@ -55,6 +55,11 @@ def validate_enable_s3_uploads(new_val)
     validate_error :s3_upload_bucket_is_required if new_val == "t" && SiteSetting.s3_upload_bucket.blank?
   end
 
+  def validate_secure_images(new_val)
+    validate_error :secure_images_requirements if new_val == "t" && SiteSetting.login_required.blank?
```
eviltrout: These two lines needn't be `blank?` because the site setting is typed as a boolean and will never be `nil` right? Could be `SiteSetting.login_required?` etc

config/site_settings.yml
pmusaraj: Yes, thanks for catching it!

lib/site_settings/validations.rb
pmusaraj: I switched them to `!SiteSetting.login_required`, but initially I used `blank?` because that was the convention for other validations on the same file. I'm guess they are equivalent.

eviltrout: `.blank?` is better for strings or things that might be objects or `nil`. In this case it's a boolean so I think it's better with `!SiteSetting.login_required?`

db/migrate/20190716173854_add_secure_to_uploads.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddSecureToUploads < ActiveRecord::Migration[5.2]
+  def up
+    add_column :uploads, :secure, :boolean, default: false, null: false
+
+    Upload.find_each do |upload|
```
eviltrout: We need to stick to manual SQL in migrations, because the fields in objects change over time and it's too easy to break old migrations who count on a particular table structure.

SamSaffron:
Looks like this is conflicting, any chance we can aim to get this into a mergable state in the next week or so? iPad fixes though are higher priority for sure.

lib/tasks/uploads.rake
```diff
@@ -907,6 +907,28 @@ task "uploads:recover" => :environment do
   end
 end
 
+task "uploads:ensure_correct_acl" => :environment do
```
martin-brennan: This rake task is my only real concern here as it could take a veeeery long time for people with lots and lots of uploads. I think each of these grouped calls should probably be in a transaction as well:

```
Upload.transaction do
  upload.update_secure_status
  upload.posts.each { |post| post.rebake! }
end
```

To speed it up I guess the uploads could be grouped and the code to update the status + rebake the post could be split into separate threads per group? Another nicety could be a counter of how many uploads there are and what progress we are up to, as it's hard to tell how far along this task is ATM. So far I have no real idea of the scale of data of real-world Discourse instances so I'm not sure if this is a valid worry.

Would be the intention here be to run this rake task manually on a server after the user has changed the secure media setting? If so should we add a message near the secure media setting to let them know that?

martin-brennan:
A (hopefully) helpful overview of how this works code-wise based on my review. This just helped my brain connect the dots.

1. File is uploaded to a topic and an upload is created
2. The reply button is pressed and `PostCreator` is called
3. Within this `update_uploads_secure_status` is called. This goes through all the uploads in the post and calls `update_secure_status` on them
4. The upload is marked as secure IF:
  * Secure media setting enabled
  * Media is supported
  * login_required OR first post in topic has secure media, determined by calling `with_secure_media?` on post
  * `with_secure_media?` returns true IF secure media setting enabled, and the post is a private message OR `login_required`
5. If the upload is secure, we then update the ACL to private by calling `update_upload_ACL`

When a post is cooked, we do the following:

1. If `with_secure_media?` is true, replace all supported media URLS in the post with the `secure-media-uploads` `UploadController` URL
2. This URL 404s if the secure media setting is not enabled, otherwise it presigns the media URL.

We strip secure media URLs when emailing post excerpts using `PrettyText`.

lib/tasks/uploads.rake
pmusaraj: Thanks @martin-brennan, I think you're right to raise a performance concern here. 

The intention is for this to be run manually, either when a relevant site setting has been changed (secure media, login required, etc.) or if a site has been moved, restored, and so on. Making the task faster and more verbose to the user running it (i.e. adding a progress status indicator) would be awesome!

Regarding the message near the secure media setting, I would wait and see if users raise this as an issue on meta often enough to warrant adding a note about the task in the UI.

app/mailers/user_notifications.rb
```diff
@@ -349,13 +349,23 @@ def user_locale(user)
...
+    result = +"#{post.with_secure_media? ? strip_secure_urls(post.raw) : post.raw}\n\n"
     if add_posted_by
       result << "#{I18n.t('user_notifications.posted_by', username: post.username, post_date: post.created_at.strftime("%m/%d/%Y"))}\n\n"
     end
     result
   end
 
+  def strip_secure_urls(raw)
+    raw.each_line do |line|
+      if (line.start_with?(Discourse.store.s3_upload_host) && FileHelper.is_supported_media?(line))
```
ZogStriP: Are upload URLS always at the beginning of a line?

db/migrate/20190716173854_add_secure_to_uploads.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AddSecureToUploads < ActiveRecord::Migration[5.2]
+  def up
+    add_column :uploads, :secure, :boolean, default: false, null: false
+
+    if SiteSetting.prevent_anons_from_downloading_files
```
ZogStriP: We should not use models in migrations in case we ever rename of model, this would break the migrations.

Also, that loop should ideally be replace by 1 SQL query if possible.

app/controllers/uploads_controller.rb
```diff
@@ -110,6 +110,17 @@ def show_short
     end
   end
 
+  def show_secure
+    # do not serve uploads requested via XHR to prevent XSS
+    return xhr_not_allowed if request.xhr?
+
+    if SiteSetting.secure_media?
+      redirect_to Discourse.store.signed_url_for_path("#{params[:path]}.#{params[:extension]}")
```
ZogStriP: Could this be a DoS vector? Should this be rate limited and/or cached?

db/migrate/20190716173854_add_secure_to_uploads.rb
pmusaraj: I don't know how to get the upload's `for_theme` and  `for_site_setting` attributes without loading the model. We don't want to mark these uploads as secure, they might be font files, for example.

app/mailers/user_notifications.rb
pmusaraj: Thanks, this should be addressed now.

db/migrate/20190716173854_add_secure_to_uploads.rb
ZogStriP: Except that any of [these `for_<something>` `attr_accessor`](https://github.com/discourse/discourse/blob/836b3f4d82dfc610574db93cfd1d86e4064d4062/app/models/upload.rb#L21-L26) are **not** stored in the database.

They're only used when creating an upload to do some different validation/processing depending on what the upload is for.

If you want to know if an upload is associated to a `theme` or a `site_setting`, you'll have to `join` some tables. Have a look at the [`clean_up_uploads`](https://github.com/discourse/discourse/blob/master/app/jobs/scheduled/clean_up_uploads.rb) job.

app/mailers/user_notifications.rb
```diff
@@ -349,13 +349,29 @@ def user_locale(user)
...
+  def strip_secure_urls(raw)
+    urls = Set.new
+    raw.scan(URI.regexp(%w{http https})) { urls << $& }
+
+    urls.each do |url|
+      begin
+        if (url.start_with?(Discourse.store.s3_upload_host) && FileHelper.is_supported_media?(url))
+          raw = raw.sub(url, "<p class='secure-media-notice'>#{I18n.t("emails.secure_media_placeholder")}</p>")
+        end
+      rescue URI::InvalidURIError
```
ZogStriP: What could be throwing this exception?

pmusaraj: Any bad URL, for example `http:error]`. See also https://github.com/discourse/discourse/blob/master/spec/lib/search_spec.rb#L46

ZogStriP: I meant both `start_with?` and `FileHelper.is_supported_media?` are working on a `string` and, as far as I know, are not trying to convert the `string` into an `URI` object.

pmusaraj: D'oh... 🤦‍♂ Cleaned up now.

app/controllers/uploads_controller.rb
pmusaraj: I don't think so because the `show_secure` method requires a logged in user.

PR #7890, opened 2019-07-14
FIX: apply defaults constraints to routes format

OsamaSayegh:
Note I intentionally didn't fix the indentation so that it's clear what actually changed. Will fix it before merging.


spec/requests/users_controller_spec.rb
```diff
@@ -2372,20 +2372,20 @@ def post_user
 
   describe '#confirm_admin' do
     it "fails without a valid token" do
-      get "/u/confirm-admin/invalid-token.josn"
```
eviltrout: WAT

PR #7893, opened 2019-07-15
FIX: backups taken by pg_dump >= 11 are nonportable

Supermathie:


Supermathie:
travis will probably fail due to
```
  1) GroupsController#index should return the right response
```
that's broken at the moment and unrelated to this PR

PR #7895, opened 2019-07-16
FIX: Fallback to gzip compression if brotli isn't supported

OsamaSayegh:


spec/requests/static_controller_spec.rb
```diff
@@ -120,6 +120,28 @@
     end
   end
 
+  context '#cdn_asset' do
```
ZogStriP: How is this relevant to the changes made in `application_helper.rb`?

Looks like this is testing brotli assets when the other changes are about falling back to gzip?

OsamaSayegh: Oh I just noticed we were missing a test for `#cdn_asset` so I added one. (also gzip and brotlit assets go through the same exact code path in this endpoint so I didn't duplicate this test for gzip)

ZogStriP: I see, next time, it would be better to have that in a different commit ;)

PR #7896, opened 2019-07-16
UX: Disable system edit notifications by default

davidtaylorhq:
Enabling this setting prevents notifications when the system downloads hotlinked images. This stops an onslaught of notifications when old posts are rebaked. It does not affect regular edit notifications


ZogStriP:
I agree, it's a better default 👍

PR #7899, opened 2019-07-17
Watched words improvements

OsamaSayegh:
This PR has 3 main changes:

1. It adds a button that allows downloading all words of an action (e.g. block, flag etc.) in a CSV file which will have each word on a line by itself.

2. It adds a button that clears all words of an action.

3. The error message that shows up when a post contains blocked words is now improved to show all blocked words contained in the post rather than just the first one.

Screenshot:

image

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
@@ -37,6 +37,11 @@ export default Ember.Controller.extend({
     return a ? a.words.length : 0;
   },
 
+  @computed("actionNameKey")
+  downloadLink(type) {
+    return `/admin/logs/watched_words/action/${type}/download`;
+  },
+
   actions: {
     recordAdded(arg) {
       const a = this.findAction(this.actionNameKey);
--- a/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
@@ -87,12 +87,9 @@ export default Ember.Controller.extend({
         I18n.t("yes_value"),
         result => {
           if (result) {
-            ajax(
-              `/admin/logs/watched_words/action/${actionKey}.json`,
-              {
-                method: "DELETE"
-              }
-            ).then(() => {
+            ajax(`/admin/logs/watched_words/action/${actionKey}.json`, {
+              method: "DELETE"
+            }).then(() => {
               const action = this.findAction(actionKey);
               if (action) {
                 action.setProperties({
--- a/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
@@ -1,5 +1,6 @@
 import computed from "ember-addons/ember-computed-decorators";
 import WatchedWord from "admin/models/watched-word";
+import { ajax } from "discourse/lib/ajax";
 
 export default Ember.Controller.extend({
   actionNameKey: null,
@@ -17,13 +18,13 @@ export default Ember.Controller.extend({
   },
 
   @computed("actionNameKey", "adminWatchedWords.model")
-  filteredContent(actionNameKey) {
-    if (!actionNameKey) {
-      return [];
-    }
+  currentAction(actionName) {
+    return this.findAction(actionName);
+  },
 
-    const a = this.findAction(actionNameKey);
-    return a ? a.words : [];
+  @computed("currentAction.words.[]", "adminWatchedWords.model")
+  filteredContent(words) {
+    return words || [];
   },
 
   @computed("actionNameKey")
@@ -31,10 +32,9 @@ export default Ember.Controller.extend({
     return I18n.t("admin.watched_words.action_descriptions." + actionNameKey);
   },
 
-  @computed("actionNameKey", "adminWatchedWords.model")
-  wordCount(actionNameKey) {
-    const a = this.findAction(actionNameKey);
-    return a ? a.words.length : 0;
+  @computed("currentAction.count")
+  wordCount(count) {
+    return count || 0;
   },
 
   @computed("actionNameKey")
@@ -67,10 +67,9 @@ export default Ember.Controller.extend({
     },
 
     recordRemoved(arg) {
-      const a = this.findAction(this.actionNameKey);
-      if (a) {
-        a.words.removeObject(arg);
-        a.decrementProperty("count");
+      if (this.currentAction) {
+        this.currentAction.words.removeObject(arg);
+        this.currentAction.decrementProperty("count");
       }
     },
 
@@ -78,6 +77,33 @@ export default Ember.Controller.extend({
       WatchedWord.findAll().then(data => {
         this.set("adminWatchedWords.model", data);
       });
+    },
+
+    clearAll() {
+      const actionKey = this.actionNameKey;
+      bootbox.confirm(
+        I18n.t(`admin.watched_words.clear_all_confirm_${actionKey}`),
+        I18n.t("no_value"),
+        I18n.t("yes_value"),
+        result => {
+          if (result) {
+            ajax(
+              `/admin/logs/watched_words/action/${actionKey}/clear_all.json`,
+              {
+                method: "DELETE"
+              }
+            ).then(() => {
+              const action = this.findAction(actionKey);
+              if (action) {
+                action.setProperties({
+                  words: [],
+                  count: 0
+                });
+              }
+            });
+          }
+        }
+      );
     }
   }
 });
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/admin/controllers/admin-watched-words-action.js.es6
```diff
@@ -17,24 +18,28 @@ export default Ember.Controller.extend({
...
-  @computed("actionNameKey", "adminWatchedWords.model")
-  wordCount(actionNameKey) {
-    const a = this.findAction(actionNameKey);
-    return a ? a.words.length : 0;
+  @computed("currentAction.count")
+  wordCount(count) {
+    return count || 0;
+  },
+
+  @computed("actionNameKey")
```
jjaffeux: you can use fmt() helper for this

app/assets/javascripts/admin/templates/watched-words-action.hbs
```diff
@@ -26,3 +33,11 @@
     {{i18n 'admin.watched_words.word_count' count=wordCount}}
   {{/if}}
 </div>
+
+<div class="clear-all-row">
+  {{d-button
+    class="btn btn-danger clear-all"
```
jjaffeux: doesnt need btn class

app/assets/javascripts/admin/templates/watched-words-action.hbs
```diff
@@ -26,3 +33,11 @@
...
+    label="admin.watched_words.clear_all"
+    icon="trash"
```
jjaffeux: I think it should be trash-alt now no?

app/assets/stylesheets/common/admin/staff_logs.scss
```diff
@@ -395,13 +413,18 @@ table.screened-ip-addresses {
 }
 
 .watched-words-uploader {
-  margin-left: auto;
+  margin-left: 7px;
```
jjaffeux: dont really like having a "7px", can't we got with 5 or 10 ?

davidtaylorhq:
Tried this out locally and it works great. I found one issue (which might have been broken already):

- Click upload
- Select file
- Words are uploaded, confirmation appears ✅ 
- Click upload again
- Select different file
- Nothing happens 😢 

I have to reload the page before it will let me upload another CSV.

app/assets/javascripts/admin/templates/watched-words-action.hbs
```diff
@@ -9,8 +9,15 @@
   filteredContent=filteredContent
   regularExpressions=adminWatchedWords.regularExpressions}}
 
+<div class="download">
+  <a class="btn btn-default download-link" target="_blank" href={{downloadLink}}>
```
eviltrout: `{{d-button}}` should now accept an `href` parameter which is preferred.

OsamaSayegh:
@davidtaylorhq I can't reproduce this 😢 Did you see any errors in the console?

config/routes.rb
```diff
@@ -177,6 +177,8 @@
       resources :watched_words, only: [:index, :create, :update, :destroy] do
         collection do
           get "action/:id" => "watched_words#index"
+          get "action/:id/download" => "watched_words#download"
+          delete "action/:id/clear_all" => "watched_words#clear_all"
```
ZogStriP: You don't need to create a new endpoint for clearing out all the watched words.

Using the `DELETE` verb on the "collection" is enough.

```suggestion
          delete "action/:id" => "watched_words#clear_all"
```

app/assets/stylesheets/common/admin/staff_logs.scss
```diff
@@ -365,14 +365,32 @@ table.screened-ip-addresses {
   float: left;
 }
 
+.admin-watched-words {
+  .clear-all-row {
+    display: flex;
+    margin-top: 10px;
+
+    .btn.clear-all {
+      margin-left: auto;
```
eviltrout: Could you use `justify-content: flex-end` to allow flexbox to align this item?

eviltrout: Actually there are a few `floats` below - if we could remove them and use flex that would be ideal.

config/routes.rb
```diff
@@ -177,6 +177,8 @@
       resources :watched_words, only: [:index, :create, :update, :destroy] do
         collection do
           get "action/:id" => "watched_words#index"
+          get "action/:id/download" => "watched_words#download"
```
ZogStriP: Since we have an hardcoded limit of 1000 watched words per action, they all are serialized to the client.

I don't think we need to call the server to generate a CSV when we have all the information already available in the browser.

config/routes.rb
eviltrout: We haven't been great at enforcing it but we prefer urls to use dashes instead of underscores. The ruby method can be `clear_all` but the URL would ideally be `clear-all`

davidtaylorhq:
> I can't reproduce this

Ah, it only happens if both files have the same filename

config/routes.rb
davidtaylorhq: We need to make sure to escape/quote the CSV correctly, in case the watched words contain commas. In Ruby we can do that easily with [CSV](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html). I'm not sure it's so easy in Javascript 🤔

app/services/word_watcher.rb
```diff
@@ -54,13 +57,29 @@ def should_flag?
...
-  def word_matches_for_action?(action)
-    r = self.class.word_matcher_regexp(action)
-    r ? r.match(@raw) : false
+  def word_matches_for_action?(action, all_matches: false)
+    regexp = self.class.word_matcher_regexp(action)
+    if regexp
+      match = regexp.match(@raw)
+      return match if !all_matches || !match
+      matches = []
+      regexps = self.class.get_cached_words(action).map do |w|
```
eviltrout: This could have a bad performance impact. On a site with 1000 watched words, it looks like each post will loop through each word, make a regexp and run it against the raw?

Is there a reason we can't use the `word_matcher_regexp` to find all the matches?

OsamaSayegh: I'm worried about performance too, but `word_matcher_regexp` only finds the first blocked word in raw and not all of the blocked words. For example if raw was `word1 word0` and we had `word0` and `word1` as blocked words, `word_matcher_regexp` would just give us `word1`, whereas we want both `word1` and `word0`.

I tried to come up with a single regexp that combines all blocked words and gives all matches, everything I tried didn't work, and I'm not sure if that's even possible with regular expressions?

eviltrout: Can you not call `string.scan(regexp)` to return a list of all the matches in one go?

If not, `match` does take an index of the last match. Your loop would be much more efficient if you only searched from the last match onwards.

OsamaSayegh: `String#scan` seems to do exactly what we want, thanks!!!! I'll change the code to use `scan` when I'm back at my computer in a couple of hours.

OsamaSayegh: Ok I've update the PR to use `scan`, let me know what you think. (the second commit below has the changes for this)

config/routes.rb
OsamaSayegh: Actually I'm not sure why we require a CSV file here. When you upload a file, it's treated as a regular txt file and each line is treated as one word even though we require a CSV file:

https://github.com/discourse/discourse/blob/514aaacdf486613bda49f8c729867926fbceb174/app/controllers/admin/watched_words_controller.rb#L31-L33

Do you think there would be problems if we dropped the CSV requirement and switched to txt?

davidtaylorhq: In most cases, I think it would be fine. However, some people will be using other applications to generate the list. For example,  in excel I can add two words to a list and end up with this CSV:

```
"word,with,comma"
"word
with
newline"
```
Anyway, as long as we're consistent with imports and exports I don't think it matters. We should import/export `.csv` **or** import/export `.txt`

app/services/word_watcher.rb
```diff
@@ -55,12 +65,34 @@ def should_flag?
...
-    r = self.class.word_matcher_regexp(action)
-    r ? r.match(@raw) : false
-  end
+  def word_matches_for_action?(action, all_matches: false)
+    regexp = self.class.word_matcher_regexp(action)
+    if regexp
+      match = regexp.match(@raw)
+      return match if !all_matches || !match
 
+      matches = []
```
eviltrout: It would be better to make this a `Set.new` which ensures it's unique as you insert.

config/routes.rb
OsamaSayegh: OK I went with importing/exporting txt files.

@ZogStriP I tried to do the downloading on the client side, it worked well on chrom but didn't work on IE11 and Edge didn't work.

app/services/word_watcher.rb
OsamaSayegh: Done ✅

davidtaylorhq:
@OsamaSayegh is this ready to merge now?

OsamaSayegh:
Yeah I think so, merging it in...

PR #7900, opened 2019-07-17
Add default Spanish quotation mark style and TLD

iapellaniz:
In Spanish it's recommended to use latin (« »), then English (“ ”) quotation marks.

Source: http://lema.rae.es/dpd/srv/search?id=SSTAZ5sDyD6h59vijX

Also adds .es TLD to be automatically recognized by Markdown by default.


iapellaniz:
Forgot prefix. Will make another PR later.

PR #7902, opened 2019-07-17
FIX: S3 uploads were missing a cache-control header

xfalcox:


coding-horror:
Errr do we want this on EVERY SINGLE upload? That seems unwise?

lib/tasks/s3.rake
```diff
@@ -140,6 +140,50 @@ task 's3:correct_acl' => :environment do
...
+  puts "ensuring cache-control is set on every upload and optimized image"
+
+  i = 0
+
+  base_url = Discourse.store.absolute_base_url
+
+  acl = SiteSetting.prevent_anons_from_downloading_files ? 'private' : 'public-read'
+  cache_control = 'max-age=31556952, public, immutable'
+
+  objects = Upload.pluck(:id, :url).map { |array| array << :upload }
```
eviltrout: Are we going to need a find_each for this? I don't want us to run out of memory when there are many many uploads / images.

xfalcox: The `pluck` usage makes it be an array of [int, string, int] tuples, instead of the full blow objects, which is the same approach used by the existing rake task at

https://github.com/discourse/discourse/blob/master/lib/tasks/s3.rake#L109-L141
(this task is basically a copy of that one using different S3 API calls)

If you feel strongly about this I can try and extract the common bits into a method and process it on batches.

eviltrout: Let's try living with it for now, and revisit as batches later.

PR #7903, opened 2019-07-18
UX: adds theme body tab contents to no-ember pages

hnb-ku:
Context: https://meta.discourse.org/t/customization-in-body-tag-wont-appear-in-some-pages/105167

If a theme contains some html in the `</body>` tab, this html is not loaded on no-ember pages like `/404`

This is not consistent with fact that the contents of the theme's `footer` tab are actually loaded on those pages.

**Before** 

Normal pages:

no-ember01

No-ember pages 

no-ember02

Notice how the body tab contents are not loaded.

After:

Normal pages (no change)

no-ember01

No-ember pages

no-ember03

The body tab contents are loaded on the page.

Here's a sample theme that contains the markup added in the screenshots for testing:

[discourse-light.tar.gz](https://github.com/discourse/discourse/files/3406449/discourse-light.tar.gz)

### DIFF
--- a/app/views/layouts/no_ember.html.erb
+++ b/app/views/layouts/no_ember.html.erb
@@ -22,6 +22,7 @@
     </div>
   </section>
   <%= theme_lookup("footer") %>
+  <%= raw theme_lookup("body_tag") %>
   <%= build_plugin_html 'no-client:footer' %>
   <%= build_plugin_html 'server:before-body-close' %>
 </body>
--- a/app/views/layouts/crawler.html.erb
+++ b/app/views/layouts/crawler.html.erb
@@ -45,6 +45,7 @@
       </nav>
       <p class='powered-by-link'><%= t 'powered_by_html' %></p>
     </footer>
+    <%= theme_lookup("footer") %>
     <%= theme_lookup("body_tag") %>
   </body>
   <%= yield :after_body %>


ZogStriP:
Is this the only page where it's missing?

hnb-ku:
> Is this the only page where it's missing?

I believe so @ZogStriP since the other pages where we want theme html to show use either `application.html.erb` or `crawler.html.erb` and loaded on in both of those. 

However, the footer tab is missing from the `crawler.html.erb` would you like me to include that in this PR or add it in a separate PR?

ZogStriP:
The `crawler` template should also have that as well. Fine to add in that PR.

app/views/layouts/no_ember.html.erb
```diff
@@ -22,6 +22,7 @@
     </div>
   </section>
   <%= theme_lookup("footer") %>
+  <%= raw theme_lookup("body_tag") %>
```
ZogStriP: Do you need the `raw` here? If you need, why isn't it needed for the `footer` and in the `crawler` template?

```suggestion
  <%= theme_lookup("body_tag") %>
```

hnb-ku:
@ZogStriP I'm closing this now in favour of a separate PR that implements that changes discusses above.

hnb-ku:
@ZogStriP this is continued in #7955

PR #7907, opened 2019-07-18
FEATURE: Allow choice of category when making a PM public

eviltrout:
Previously it would default to uncategorized, which was not ideal on
some forums. This gives the staff member more choice about what they'd
like to do.

image

### DIFF
--- a/app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
@@ -1,0 +1,26 @@
+import { popupAjaxError } from "discourse/lib/ajax-error";
+
+export default Ember.Controller.extend({
+  publicCategoryId: null,
+  saving: true,
+
+  onShow() {
+    this.publicCategoryId = null;
+    this.saving = false;
+  },
+
+  actions: {
+    makePublic() {
+      let topic = this.model;
+      topic
+        .convertTopic("public", { categoryId: this.publicCategoryId })
+        .then(() => {
+          topic.set("archetype", "regular");
+          topic.set("category_id", this.publicCategoryId);
+          this.appEvents.trigger("header:show-topic", topic);
+          this.send("closeModal");
+        })
+        .catch(popupAjaxError);
+    }
+  }
+});
--- a/app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
+++ b/app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
@@ -1,12 +1,12 @@
 import { popupAjaxError } from "discourse/lib/ajax-error";
+import ModalFunctionality from "discourse/mixins/modal-functionality";
 
-export default Ember.Controller.extend({
+export default Ember.Controller.extend(ModalFunctionality, {
   publicCategoryId: null,
   saving: true,
 
   onShow() {
-    this.publicCategoryId = null;
-    this.saving = false;
+    this.setProperties({ publicCategoryId: null, saving: false });
   },
 
   actions: {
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
```diff
@@ -0,0 +1,26 @@
...
+    this.saving = false;
+  },
+
+  actions: {
+    makePublic() {
+      let topic = this.model;
+      topic
+        .convertTopic("public", { categoryId: this.publicCategoryId })
+        .then(() => {
+          topic.set("archetype", "regular");
```
jjaffeux: we can use setProperties here

app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
```diff
@@ -0,0 +1,26 @@
...
+          topic.set("category_id", this.publicCategoryId);
+          this.appEvents.trigger("header:show-topic", topic);
+          this.send("closeModal");
```
jjaffeux: shouldn't this controller be extending ModalFunctionality because of this?

app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
```diff
@@ -0,0 +1,26 @@
...
+  saving: true,
+
+  onShow() {
+    this.publicCategoryId = null;
+    this.saving = false;
+  },
+
+  actions: {
+    makePublic() {
+      let topic = this.model;
```
jjaffeux: I would say this variable here is not very useful, maybe to make it explicit it's a topic?

app/controllers/topics_controller.rb
```diff
@@ -821,7 +821,7 @@ def convert_topic
     guardian.ensure_can_convert_topic!(topic)
 
     if params[:type] == "public"
-      converted_topic = topic.convert_to_public_topic(current_user)
+      converted_topic = topic.convert_to_public_topic(current_user, params[:category_id])
```
ZogStriP: Are we ensuring the current user is able to post in that category somewhere down the line?

app/models/topic.rb
```diff
@@ -1307,8 +1307,8 @@ def self.with_no_response_total(opts = {})
     builder.query_single.first.to_i
   end
 
-  def convert_to_public_topic(user)
-    public_topic = TopicConverter.new(self, user).convert_to_public_topic
+  def convert_to_public_topic(user, opts = nil)
```
ZogStriP: Is it an opts object? Looks like it's only supporting one optional parameter: category_id.

eviltrout: Good catch, fixed.

app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
eviltrout: Yes, and the variable is used 3 times, so I felt it was slightly better than `this.model` three times.

app/assets/javascripts/discourse/controllers/convert-to-public-topic.js.es6
eviltrout: Weird, it worked without it! But I'll add it to be safe.

app/controllers/topics_controller.rb
eviltrout: No but only staff members can convert topics from PMs to public, and staff members can [change any category](https://github.com/discourse/discourse/blob/master/lib/guardian/topic_guardian.rb#L57).

PR #7910, opened 2019-07-18
FEATURE: Allow viewing of raw emails for reviewable queued posts

eviltrout:
If a post arrives via email but must be reviewed, we now show an
icon that can be clicked to view the raw contents of the email.

This is useful if Discourse's email parser is acting odd and the user
reviewing the post wants to know what the original contents were before
approving/rejecting the post.

### DIFF
--- a/app/assets/stylesheets/common/base/reviewables.scss
+++ b/app/assets/stylesheets/common/base/reviewables.scss
@@ -366,7 +366,7 @@
 .reviewable-item {
   .show-raw-email {
     color: $primary-medium;
-    font-size: 0.8em;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/common/base/reviewables.scss
```diff
@@ -364,6 +364,10 @@
 }
 
 .reviewable-item {
+  .show-raw-email {
+    color: $primary-medium;
+    font-size: 0.8em;
```
jjaffeux: should be using $font-size helper

PR #7913, opened 2019-07-18
FEATURE: Add search operator to see all direct messages from a user

OsamaSayegh:
The new search operator is `in:personal-direct`. This operator returns all messages you have access to that has exactly 2 participants and no groups. If you combine this operator with `@username` then you'll get all messages between you and @username.

Also this adds a link at the bottom of the "Related messages" section that shows you all PMs (using the new search operator) with the user you're interacting.

image

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -265,6 +265,24 @@ class Search
     @advanced_filters
   end
 
+  advanced_filter(/^in:personal-direct$/) do |posts|
+    if @guardian.user
+      posts
+        .joins("JOIN topic_allowed_users tu ON posts.topic_id = tu.topic_id")
+        .joins("LEFT JOIN topic_allowed_groups tg ON posts.topic_id = tg.topic_id")
+        .where(<<~SQL, user_id: @guardian.user.id)
+          tg.id IS NULL
+          AND tu.user_id = :user_id
+          AND posts.topic_id IN (
+            SELECT topic_id
+            FROM topic_allowed_users
+            GROUP BY topic_id
+            HAVING COUNT(*) = 2
+          )
+        SQL
+    end
+  end
+
   advanced_filter(/^in:tagged$/) do |posts|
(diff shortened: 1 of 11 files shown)


coding-horror:
I like this but I think "all messages" should be the hyperlink target
rather than @username please

lib/search.rb
```diff
@@ -265,6 +265,24 @@ def self.advanced_filters
...
+    if @guardian.user
+      posts
+        .joins("JOIN topic_allowed_users tu ON posts.topic_id = tu.topic_id")
+        .joins("LEFT JOIN topic_allowed_groups tg ON posts.topic_id = tg.topic_id")
+        .where(<<~SQL, user_id: @guardian.user.id)
+          tg.id IS NULL
+          AND tu.user_id = :user_id
+          AND posts.topic_id IN (
+            SELECT topic_id
+            FROM topic_allowed_users
```
eviltrout: For performance reasons you'll want to add the user and topics filters in this correlated subquery. Otherwise it will query all rows in the database!

spec/components/search_spec.rb
```diff
@@ -236,6 +236,54 @@
...
+      let(:group) do
+        group = Fabricate(:group, has_messages: true)
+        group.add(current)
+        group.add(participant)
+        group
+      end
+
+      def create_pm(users:, group: nil)
+        pm = Fabricate(:private_message_post, user: users.first).topic
+        pm.topic_allowed_users.where.not(user_id: users.first.id).destroy_all
```
eviltrout: It's better to construct a new object that has the data you want rather than using an existing one and deleting the stuff you don't.

lib/search.rb
OsamaSayegh: I've updated the PR and I think the query should be more efficient now. Let me know if we can do better. Thanks for the review!

PR #7914, opened 2019-07-19
UX: new {{categories-drop}} component to unify parent/sub category

jjaffeux:
Screenshot 2019-07-19 at 12 26 07
Screenshot 2019-07-19 at 12 26 45
Screenshot 2019-07-19 at 12 26 01

### DIFF
--- a/app/assets/javascripts/select-kit/components/categories-drop.js.es6
+++ b/app/assets/javascripts/select-kit/components/categories-drop.js.es6
@@ -1,0 +1,208 @@
+import ComboBoxComponent from "select-kit/components/combo-box";
+import DiscourseURL from "discourse/lib/url";
+import { default as computed } from "ember-addons/ember-computed-decorators";
+import Category from "discourse/models/category";
+import { categoryBadgeHTML } from "discourse/helpers/category-link";
+
+export default ComboBoxComponent.extend({
+  pluginApiIdentifiers: ["categories-drop"],
+  classNames: "categories-drop",
+  classNameBindings: ["categoryStyle"],
+  categoryStyle: Ember.computed.reads("siteSettings.category_style"),
+  verticalOffset: 3,
+  headerComponent: "category-drop/category-drop-header",
+  allowAutoSelectFirst: false,
+  tagName: "li",
+  fullWidthOnMobile: true,
+  caretDownIcon: "caret-right",
+  caretUpIcon: "caret-down",
+  subCategory: false,
+  isAsync: true,
+  rowComponent: "category-row",
+
+  init() {
+    this._super(...arguments);
+
+    this.rowComponentOptions.setProperties({
+      hideParentCategory: false,
+      allowUncategorized: true,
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/select-kit/components/categories-drop.js.es6
```diff
@@ -0,0 +1,208 @@
...
+import DiscourseURL from "discourse/lib/url";
+import { default as computed } from "ember-addons/ember-computed-decorators";
+import Category from "discourse/models/category";
+import { categoryBadgeHTML } from "discourse/helpers/category-link";
+
+export default ComboBoxComponent.extend({
+  pluginApiIdentifiers: ["categories-drop"],
+  classNames: "categories-drop",
+  classNameBindings: ["categoryStyle"],
+  categoryStyle: Ember.computed.alias("siteSettings.category_style"),
```
eviltrout: `reads`?

app/assets/javascripts/select-kit/components/categories-drop.js.es6
```diff
@@ -0,0 +1,208 @@
...
+        name: I18n.t("categories.all"),
+        __sk_row_type: "noopRow",
+        id: "all-categories"
+      });
+    }
+
+    if (subCategory && (hasSelection || !noSubcategories)) {
+      shortcuts.push({
+        name: this.noCategoriesLabel,
+        __sk_row_type: "noopRow",
```
eviltrout: This is an unusual API. Is it something meant to be private but must be used?

app/assets/javascripts/select-kit/components/categories-drop.js.es6
```diff
@@ -0,0 +1,208 @@
...
+      this.forceValue(category.id);
+    }
+  },
+
+  @computed("content")
+  filterable(content) {
+    const contentLength = (content && content.length) || 0;
+    return (
+      contentLength >= 15 ||
+      (this.isAsync && contentLength < Discourse.Category.list().length)
```
eviltrout: Can we import `Category` and call `Category.list()` instead?

app/assets/javascripts/select-kit/components/categories-drop.js.es6
jjaffeux: oops

app/assets/javascripts/select-kit/components/categories-drop.js.es6
jjaffeux: This is mostly a shortcoming of initial implementation and I had to come up with a solution. It will eventually disappear, but it gives me much flexibility for now, and also makes it very easy to track where are these special row cases so I can fix them all at once, when I have a better solution.

jjaffeux:
I'm working on various select-kit changes, and putting this on hold

SamSaffron:
Should we close this till you are ready to have it merged?

PR #7915, opened 2019-07-20
BUG: send featuredLink as featured_link to backend to update correct …

mrfinch:
…value

https://meta.discourse.org/t/editing-a-topic-link-does-not-change-its-featured-link/123007

### DIFF
--- a/app/assets/javascripts/discourse/models/composer.js.es6
+++ b/app/assets/javascripts/discourse/models/composer.js.es6
@@ -710,7 +710,7 @@ const Composer = RestModel.extend({
         Object.keys(_edit_topic_serializer)
       );
       // frontend should have featuredLink but backend needs featured_link
-      if (typeof topicProps.featuredLink !== "undefined") {
+      if (topicProps.featuredLink) {
         topicProps.featured_link = topicProps.featuredLink;
         delete topicProps.featuredLink;
       }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -709,6 +709,11 @@ const Composer = RestModel.extend({
       const topicProps = this.getProperties(
         Object.keys(_edit_topic_serializer)
       );
+      // frontend should have featuredLink but backend needs featured_link
```
mrfinch: this is similar to how we do it for categoryId and category_id. I didn't want to change the variable in front-end to `featured_link`. This also re-uses `_edit_topic_serializer` so that it is easy to add new serialized variables and send them to the backend. I could create topicProps from scratch like ```const topicProps = { title: 'XYZ', .... }``` but it will also mean when adding a new variable in serializer, it needs to be updated here too.

app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -709,6 +709,11 @@ const Composer = RestModel.extend({
...
+      if (typeof topicProps.featuredLink !== "undefined") {
```
eviltrout: Any reason why you didn't do `if (topicProps.featuredLink)` here?

mrfinch: No particular reason, I have changed it to `if (topicProps.featuredLink)` :bowing_man:

eviltrout:
Thanks, looks good now!

PR #7916, opened 2019-07-20
FEATURE: list category moderators on the about page

OsamaSayegh:
image

### DIFF
--- a/app/assets/javascripts/discourse/helpers/category-link.js.es6
+++ b/app/assets/javascripts/discourse/helpers/category-link.js.es6
@@ -19,7 +19,7 @@ function categoryStripe(color, classes) {
 /**
   Generates category badge HTML
 
-  @param {Object|number} category The category to generate the badge for, or the category ID.
+  @param {Object} category The category to generate the badge for.
   @param {Object} opts
     @param {String}  [opts.url] The url that we want the category badge to link to.
     @param {Boolean} [opts.allowUncategorized] If false, returns an empty string for the uncategorized category.
@@ -28,10 +28,6 @@ function categoryStripe(color, classes) {
 **/
 export function categoryBadgeHTML(category, opts) {
   opts = opts || {};
-
-  if (typeof category === "number") {
-    category = Discourse.Site.currentProp("categories").findBy("id", category);
-  }
 
   if (
     !category ||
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/discourse/helpers/category-link.js.es6
```diff
@@ -29,6 +29,10 @@ function categoryStripe(color, classes) {
 export function categoryBadgeHTML(category, opts) {
   opts = opts || {};
 
+  if (typeof category === "number") {
```
eviltrout: I find this kind of API "surprising" and I think in the long term it's much easier to pick one rather than the other. I suggest only accepting the category object, and having the route wire it up for you. I should note if you use the Store for doing this query, category_id should automatically be converted to objects for you.

OsamaSayegh: Ok done, does it look good now?

app/models/about.rb
```diff
@@ -66,4 +80,20 @@ def stats
...
+    return [] if category_ids.blank?
+    results = DB.query(<<~SQL, category_ids: category_ids)
+      SELECT c.id category_id, array_agg(gu.user_id) user_ids
+      FROM categories c
+      JOIN group_users gu
+      ON gu.group_id = reviewable_by_group_id
+      WHERE c.id IN (:category_ids)
+      GROUP BY c.id
+    SQL
+    moderators = User.where(id: results.map(&:user_ids).flatten).map { |u| [u.id, u] }.to_h
```
tgxworld:
```suggestion
    moderators = {}
    User.where(id: results.map(&:user_ids).flatten).each do |user|
      moderators[user.id] = user
    end
    moderators
```

This reduces multiple array allocations

app/assets/javascripts/discourse/routes/about.js.es6
```diff
@@ -16,6 +16,17 @@ export default Discourse.Route.extend({
       });
       result.about.admins = activeAdmins;
       result.about.moderators = activeModerators;
+
+      const { category_moderators: categoryModerators } = result.about;
+      if (categoryModerators && categoryModerators.length) {
+        categoryModerators.forEach((obj, index) => {
+          const category = Discourse.Site.currentProp("categories").findBy(
```
eviltrout: Is there no `this.site` available in the route?

eviltrout: Also you could import `Category` and do `Category.findById()`

PR #7918, opened 2019-07-21
Unify under Ruby 2.6.3

larouxn:
While we're already testing with Ruby `2.6.3`, there are a couple instances of `<2.6.3`.
https://github.com/discourse/discourse/blob/eb26bee046c264602933f43d29c734a7e266928e/.travis.yml#L36-L37

Was wondering if we'd like to unify everything under Ruby `2.6.3`. Note, there are some mentions of "at least Ruby 2.5" but I think those make sense as is.


eviltrout:
I don't think we're ready to make this call yet, sorry.

PR #7923, opened 2019-07-23
bug: keep query params present in auth_redirect

mrfinch:
https://meta.discourse.org/t/user-api-keys-payload-and-existing-query-string-leads-to-a-double-question-mark/123617


eviltrout:
This looks good, thanks!

PR #7931, opened 2019-07-24
DEV: prevents staff computed property to be overridden

jjaffeux:
As we create user from json and staff is already set in the json answer, this is triggering https://deprecations.emberjs.com/v3.x/#toc_computed-property-override and is doing twice the work.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -199,11 +199,7 @@ export default Ember.Controller.extend({
     );
   },
 
-  @computed
-  isStaffUser() {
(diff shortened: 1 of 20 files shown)


jjaffeux:
> I don't see `staff` returned in the `UserSerializer`? https://github.com/discourse/discourse/blob/master/app/serializers/user_serializer.rb Am I missing a mixin or something?

Ok I had kind of misunderstood the issue, so "staff" is only added to current_user json response. So we can do:

- the PR I just updated (which is what ember deprecation guide is suggesting)
- or just do a `delete userJson.staff` to be sure we are never overwriting the model computed property

eviltrout:
I don't mind either, I'll leave it to you to choose!

PR #7932, opened 2019-07-24
UX: Add expanded/collapsed class to post-controls

awesomerobot:
This adds a .collapsed/.expanded class to the post-controls. 

The initial reason for this is so I can treat the solved button differently based on the space available on mobile.


eviltrout:
Looks good!

PR #7934, opened 2019-07-24
FEATURE: customization of html emails

nlalonde:
Add a way to customize html emails via a template and css. Not covered is a way to customize anything not already covered by translation overrides (Admin > Customize > Email Templates).

Some outstanding items:

- [x] Need to test translation overrides for some emails like "user_replied". (It works)
- [x] Trying to customize html template of all emails might clash with summary emails because there's so much styling there, and it's hard-coded instead of applied from Email::Styles.


app/mailers/user_notifications.rb
```diff
@@ -12,6 +12,7 @@ class UserNotifications < ActionMailer::Base
   include ApplicationHelper
   helper :application, :email
   default charset: 'UTF-8'
+  layout 'email_template'
```
nlalonde: Hmm not sure this is actually doing anything.

app/assets/javascripts/admin/components/email-styles-editor.js.es6
```diff
@@ -0,0 +1,45 @@
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  editorId: Ember.computed.alias("fieldName"),
```
eviltrout: This should likely be `reads`

app/assets/javascripts/admin/models/email-style.js.es6
```diff
@@ -0,0 +1,27 @@
+import { ajax } from "discourse/lib/ajax";
+import RestModel from "discourse/models/rest";
+
+const EmailStyle = RestModel.extend({
```
eviltrout: You could export default `RestModel.extend` here and not put it in a constant.

app/assets/javascripts/admin/models/email-style.js.es6
```diff
@@ -0,0 +1,27 @@
...
+  changed: false,
+
+  setField(fieldName, value) {
+    this.set(`${fieldName}`, value);
+    this.set("changed", true);
+  },
+
+  saveChanges() {
```
eviltrout: Since you're using a `RestModel` you probably don't need to implement this method, and could install call `model.update(...)` instead of `model.saveChanges()`

app/assets/javascripts/admin/routes/admin-customize-email-style-edit.js.es6
```diff
@@ -0,0 +1,39 @@
...
+      fieldName: params.field_name
+    };
+  },
+
+  setupController(controller, model) {
+    controller.setProperties({
+      fieldName: model.fieldName,
+      model: model.model
+    });
+    this.set("shouldAlertUnsavedChanges", true);
```
eviltrout: This variable is only used in the route and never observed so I'd recommend making it a `this._shouldAlertUnsavedChanges = true` - no need for get/set and it's obvious it's private.

app/assets/javascripts/admin/routes/admin-customize-email-style.js.es6
```diff
@@ -0,0 +1,14 @@
+import EmailStyle from "admin/models/email-style";
+import { ajax } from "discourse/lib/ajax";
+
+export default Ember.Route.extend({
+  model() {
+    return ajax("/admin/customize/email_style.json").then(json => {
```
eviltrout: Similar to my comment above you should be able to say `this.store.find('emailStyle')` to do this automatically.

app/assets/javascripts/admin/templates/customize-email-style.hbs
```diff
@@ -0,0 +1,7 @@
+<div class='row'>
+  <h2>{{i18n 'admin.customize.email_style.heading'}}</h2>
+
+  <!-- Some explanation or something here? -->
```
eviltrout: Seems like a deccent idea to add some text here.

app/services/email_style_updater.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class EmailStyleUpdater
+
+  attr_reader :errors
```
eviltrout: This might be a good fit for the `HasErrors` mixin which includes a way to add errors easily, and the serializer knows how to render it for display in a popup.

app/views/email/default_template.html
```diff
@@ -0,0 +1,24 @@
...
+
+<!--[if mso]>
+<style type="text/css">
+body, table, td, th, h1, h2, h3 {font-family: Helvetica, Arial, sans-serif !important;}
+</style>
+<![endif]-->
+
+%{email_content}
+
+<!-- prevent Gmail on iOS font size manipulation -->
```
eviltrout: I know you didn't add this but :scream:

Gemfile
```diff
@@ -78,6 +78,7 @@ gem 'discourse_image_optim', require: 'image_optim'
 gem 'multi_json'
 gem 'mustache'
 gem 'nokogiri'
+gem 'css_parser'
```
eviltrout: Did we have a discussion about this gem anywhere that I might have missed? We are very stingy about adding new things and must do due-diligence to make sure they meet our standards.

app/assets/javascripts/admin/models/email-style.js.es6
```diff
@@ -0,0 +1,27 @@
+import { ajax } from "discourse/lib/ajax";
+import RestModel from "discourse/models/rest";
+
+const EmailStyle = RestModel.extend({
+  changed: false,
+
+  setField(fieldName, value) {
+    this.set(`${fieldName}`, value);
```
jjaffeux: could probably be:

```
this.setProperties({`${fieldName}`: value, "changed": true});
```

app/views/email/default_template.html
nlalonde: Yeah there were a lot of discoveries about html rendering in email clients while working on the summary email...

Gemfile
nlalonde: Nope, and I'd be happy to use another solution to do the same. Was hoping nokogiri could do this, but it doesn't.

app/assets/javascripts/admin/routes/admin-customize-email-style.js.es6
nlalonde: I tried this, but RestModel doesn't know about singular resources, so I shouldn't make this a RestModel.

jjaffeux:
@nlalonde I worked on the beginning of a patch which should do multple things @eviltrout mentionned:

[customize-emails.patch.zip](https://github.com/discourse/discourse/files/3434792/customize-emails.patch.zip)

Hope it helps

app/assets/javascripts/admin/routes/admin-customize-email-style.js.es6
eviltrout: Ah right, everything needs an ID.

Gemfile
eviltrout: Okay I've created a document on our developer forum about the dependency. If we add the dependency it should be `require: false` and only required when we intend to use it, since it's probably a rarely used code path.

nlalonde:
The UI is currently broken because rest model is trying to load `/email_styles`, but it should be `/admin/customize/email_style`. I'm trying to fix. I commented yesterday about rest model not understanding singular resources, and that still is a problem after Joffrey's changes.

PR #7935, opened 2019-07-24
FEATURE: add setting to show content of forwarded emails in topics

LeoMcA:
https://meta.discourse.org/t/enable-forwarded-emails-doesnt-behave-like-conventional-email-forwarding/111501

@ZogStriP I imagine this is yours to review

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -811,6 +811,39 @@ module Email
       end
     end
 
+    def forwarded_email_create_topic(destination, user, raw, title, date=nil)
+      case destination[:type]
+      when :group
+        group = destination[:obj]
+        topic_user = block_given? ? yield : user
+        create_topic(user: topic_user,
+                     raw: raw,
+                     title: title,
+                     archetype: Archetype.private_message,
+                     target_usernames: [user.username],
+                     target_group_names: [group.name],
+                     is_group_message: true,
+                     skip_validations: true,
+                     created_at: date)
+
+      when :category
+        category = destination[:obj]
+
+        return false if user.staged? && !category.email_in_allow_strangers
+        return false if !user.has_trust_level?(SiteSetting.email_in_min_trust)
+
+        topic_user = block_given? ? yield : user
+        create_topic(user: topic_user,
+                     raw: raw,
+                     title: title,
+                     category: category.id,
+                     skip_validations: topic_user.staged?,
+                     created_at: date)
+      else
+        false
+      end
+    end
+
     def forwarded_email_create_replies(destination, user)
       embedded = Mail.new(embedded_email_raw)
       email, display_name = parse_from_field(embedded)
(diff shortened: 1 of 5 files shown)


lib/email/receiver.rb
```diff
@@ -861,6 +872,42 @@ def process_forwarded_email(destination, user)
       true
     end
 
+    def forwarded_email_quote_forwarded(destination, user)
+      embedded = embedded_email_raw
+      raw = @before_embedded +
+        "\n\n[quote=\"Forwarded message\"]\n" +
```
eviltrout: This should be translated.

LeoMcA: I wasn't so sure about keeping this in my commit, it was a "nice hack" I discovered while messing about with ways of formatting the forwarded message.

Localising to the default locale of the server shouldn't be hard (I'll add a commit doing that), but it feels like I'll be opening a whole can of worms trying to localise this locally client side. Happy to remove this and have it as an unnamed quote.

lib/email/receiver.rb
```diff
@@ -861,6 +872,45 @@ def process_forwarded_email(destination, user)
...
 
+    def forwarded_email_quote_forwarded(destination, user)
+      embedded = embedded_email_raw
+      raw = <<~EOF
+        #{@before_embedded}
+
+        [quote="#{I18n.t('emails.forwarded_message')}"]
+        #{PlainTextToMarkdown.new(embedded).to_markdown}
+        [/quote]
+        EOF
```
ZogStriP: We usually indent the end of the squiggly heredoc comment at the same level as the variable

```suggestion
      EOF
```

lib/email/receiver.rb
```diff
@@ -861,6 +872,45 @@ def process_forwarded_email(destination, user)
...
+      title = subject
+
+      case destination[:type]
```
ZogStriP: Any way we could avoid duplicated that logic between `forwarded_email_create_replies` and `forwarded_email_quote_forwarded`?

lib/email/receiver.rb
eviltrout: I'd be fine with an unnamed quote. It's less burden on translators. If we find it confusing in the future we can always add it back.

lib/email/receiver.rb
LeoMcA: Yep, I had to use a block for creating the embedded user at the correct point in the logic, but it seems pretty clean to me.

lib/email/receiver.rb
```diff
@@ -801,44 +801,62 @@ def embedded_email_raw
...
+    def forwarded_email_create_replies(destination, user)
+      embedded = Mail.new(embedded_email_raw)
+      email, display_name = parse_from_field(embedded)
+
+      return false if email.blank? || !email["@"]
+
+      raw = try_to_encode(embedded.decoded, "UTF-8").presence || embedded.to_s
+      title = embedded.subject.presence || subject
+
+      post = forwarded_email_create_topic(destination, user, raw, title, embedded.date) do
```
SamSaffron: Is there a reason to pass in the param via a block vs simply adding a `topic_user: nil` param?

LeoMcA: I did this to emulate the current logic around when a staged user is created:

```
      when :category
        category = destination[:obj]

        return false if user.staged? && !category.email_in_allow_strangers
        return false if !user.has_trust_level?(SiteSetting.email_in_min_trust)
```

originally the next line was:

```
embedded_user = find_or_create_user(email, display_name)
```

now the next line is:
```
topic_user = block_given? ? yield : user
```

If we pass the user in with a param, then we create the user prior to checking those conditions, and so I might be unnecessarily creating a staged user. I did try just creating the user and passing it as a param, but it caused a spec to fail.

SamSaffron: I see ... do we need this level of flexibility here? Maybe just pass email and display_name in optionally? something about this block is not sitting right for me... I think we should probably change the other occurrence as well but this is out of scope for this pr. If we must must have this flexibility I think passing in a lambda is a bit clearer intent wise.

LeoMcA: > Maybe just pass email and display_name in optionally?

You know, I hadn't actually thought of this, and jumped straight for the complicated solution. Whoops! Will update the PR.

LeoMcA: Actually I opted for a lambda because passing in email and display_name creates hell like this:

```
topic_user = embedded_user ? find_or_create_user(embedded_user[:email], embedded_user[:name]) : user
```

Splitting things over multiple lines doesn't make it much clearer, far less than what we get with a lambda (imo):

```
topic_user = embedded_user&.call || user
```

lib/email/receiver.rb
```diff
@@ -801,44 +801,60 @@ def embedded_email_raw
...
+      when "create_replies"
+        forwarded_email_create_replies(destination, user)
+      when "quote"
+        forwarded_email_quote_forwarded(destination, user)
+      else
+        false
+      end
+    end
 
+    def forwarded_email_create_topic(destination, user, raw, title, date = nil, embedded_user = nil)
```
SamSaffron: I say given we have so many params ... let's name them all: 

`forwarded_email_create_topic(destination: , user: , raw:, title:, date: nil, embedded_user: nil)`

It makes the callers way clearer. I think once we reach about 3 params we should consider strongly naming params.

LeoMcA: I had though that there were too many arguments, and had considered passing them in as a hash, but hadn't realised this was possible (and so simple!)

I've added a commit doing this

ZogStriP:
LGTM :+1:

PR #7938, opened 2019-07-24
FIX: Frozen string error in `TopicEmbed.import`

xrav3nz:
When `SiteSetting.embed_truncate` is enabled (by default), the truncated
string is mutable and does not raise an error. :heavy_check_mark: 

However, when the setting is disabled, the `contents` string is frozen
and immutable, and will raise a `FrozenError`. :boom:

### DIFF
--- a/app/models/topic_embed.rb
+++ b/app/models/topic_embed.rb
@@ -36,8 +36,8 @@ class TopicEmbed < ActiveRecord::Base
     if SiteSetting.embed_truncate
       contents = first_paragraph_from(contents)
     end
-    contents ||= +''
-    contents << imported_from_html(url)
+    contents ||= ''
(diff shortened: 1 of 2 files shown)


app/models/topic_embed.rb
```diff
@@ -36,8 +36,8 @@ def self.import(user, url, title, contents)
     if SiteSetting.embed_truncate
       contents = first_paragraph_from(contents)
     end
-    contents ||= +''
-    contents << imported_from_html(url)
+    contents ||= ''
+    contents = +contents << imported_from_html(url)
```
xrav3nz: Not sure if this is the best way to fix this. Open to suggestions, thanks!

eviltrout:
Thanks!

PR #7944, opened 2019-07-25
Clarify how to start a tutorial

nedbat:
https://meta.discourse.org/t/does-the-advanced-user-tutorial-still-exist/123661/14

### DIFF
--- a/plugins/discourse-narrative-bot/config/locales/server.en.yml
+++ b/plugins/discourse-narrative-bot/config/locales/server.en.yml
@@ -117,8 +117,8 @@ en:
         tracks: |-
           I currently know how to do the following things:
 
-          `@%{discobot_username} %{reset_trigger} %{default_track}`
-          > Starts one of the following interactive tutorials: %{tracks}.
+          `@%{discobot_username} %{reset_trigger} {name-of-tutorial}`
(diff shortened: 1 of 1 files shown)


plugins/discourse-narrative-bot/config/locales/server.en.yml
```diff
@@ -117,8 +117,8 @@ en:
         tracks: |-
           I currently know how to do the following things:
 
-          `@%{discobot_username} %{reset_trigger} %{default_track}`
-          > Starts one of the following interactive tutorials: %{tracks}.
+          `@%{discobot_username} %{reset_trigger} {name-of-tutorial}`
```
tgxworld: This was meant to provide an example of the command so I don't think a placeholder will work well here.

nedbat: Even with the explanation text? "`{name-of-tutorial}` can be one of: `tutorial, advanced tutorial`"

eviltrout: @tgxworld this text was specifically suggested by @coding-horror: https://meta.discourse.org/t/does-the-advanced-user-tutorial-still-exist/123661/10?u=eviltrout

tgxworld: Ah icic. Merging in then.

PR #7945, opened 2019-07-25
DEV: Add ENV flag to restrict backups

OsamaSayegh:


app/controllers/admin/backups_controller.rb
```diff
@@ -32,6 +32,9 @@ def status
   end
 
   def create
+    if BackupRestore.backups_disallowed?
```
eviltrout: Trivial but this could be one line if you like `raise Discourse::InvalidAccess if BackupRestore.backups_disallowed?`

lib/backup_restore/backup_restore.rb
```diff
@@ -60,11 +60,30 @@ def self.can_rollback?
...
+  def self.strict_backup_creation?
+    ENV["DISCOURSE_STRICT_BACKUP_CREATION"] == "1" && Discourse.store.external?
+  end
+
+  def self.disallow_backups!
+    $redis.set(backup_disallowed_key, 1)
+  end
+
+  def self.backups_disallowed?
+    $redis.get(backup_disallowed_key).present?
```
eviltrout: Another trivial change but redis has an `exists` method which is slightly simpler.

lib/backup_restore/backuper.rb
```diff
@@ -452,6 +456,13 @@ def save_log(message, timestamp)
       @logs << "[#{timestamp}] #{message}"
     end
 
+    def before_run
+      if @with_uploads && BackupRestore.strict_backup_creation?
+        Discourse.enable_readonly_mode(Discourse::USER_READONLY_MODE_KEY)
+        SiteSetting.include_s3_uploads_in_backups = true
```
eviltrout: Is this site setting the only way to trigger this? I would prefer if it could be passed as an option rather than setting a global setting like this.

OsamaSayegh:
I think this PR can be closed because Neil fixed the underlying problem this PR was meant to fix?

PR #7946, opened 2019-07-25
FIX: Better error when SSO fails due to blank secret

OsamaSayegh:


spec/requests/session_controller_spec.rb
```diff
@@ -816,6 +816,16 @@ def sso_for_ip_specs
         expect(response.status).to eq(500)
       end
 
+      it "it fails with a nice error message if secret is blank" do
```
eviltrout:
```suggestion
      it "fails with a nice error message if secret is blank" do
```

PR #7947, opened 2019-07-26
UX: Show a warning when admin clicks save without adding group when changing category permissions

OsamaSayegh:
https://meta.discourse.org/t/forgotten-new-permissions-in-the-category-security-tab/122481?u=osama


app/assets/javascripts/discourse/components/edit-category-security.js.es6
```diff
@@ -13,6 +13,15 @@ export default buildCategoryPanel("security", {
       }
     },
 
+    onDropdownChange() {
+      Ember.run.later(() => {
```
OsamaSayegh: I'm not comfortable with this hack, but I had to do it because `selectedGroup` is mutated after `onSelect` is triggered, and I need the new value of  `selectedGroup`. I could use an observer here, but wanted to check first if there is a better way.

eviltrout: Does `Ember.run.scheduleOnce('afterRender', () => ... )` work? I really am against having delays like this in our codebase. cc @jjaffeux

app/assets/javascripts/discourse/templates/modal/edit-category.hbs
```diff
@@ -12,7 +12,7 @@
 
   {{#d-modal-body}}
     {{#each panels as |tab|}}
-      {{component tab selectedTab=selectedTab category=model}}
+      {{component tab selectedTab=selectedTab category=model modal=this}}
```
eviltrout: You shouldn't pass the component/controller down into another component like this. I would suggest instead passing an action that can be called by the component.

app/assets/stylesheets/common/base/modal.scss
```diff
@@ -350,13 +350,28 @@
...
+  .pending-permission-change-alert {
+    margin-left: auto;
+    max-width: 250px;
+    background: $primary-very-high;
+    color: $secondary;
+    margin-top: 10px;
+    padding: 5px 10px;
     position: relative;
-    top: 0.1em;
+    .arrow-div {
```
eviltrout: Am I crazy or did we have a tooltip CSS/library already @jjaffeux  that we could use here?

app/assets/javascripts/discourse/components/edit-category-security.js.es6
OsamaSayegh: That was the first thing I tried (and I retried again just to make sure), but sadly it didn't work.
I know you don't like observers, but for now I've switched to an observer because I think it's much better than a delay.

app/assets/javascripts/discourse/templates/modal/edit-category.hbs
OsamaSayegh: Ok I had to remove the panels loop because the rest of the tabs don't need the action and the extra attributes. Let me know what you think.

OsamaSayegh: Never mind this^^, I've simplified the code which also removed the need for an observer or delay. Let me what you think.

app/assets/stylesheets/common/base/modal.scss
OsamaSayegh: I can see we have `d-popover`, but from my understanding this is used when you want to show a tooltip on hover over an element, but here we want to alert to show up when you click save and we need to keep it shown.

OsamaSayegh:
Thanks @eviltrout! I've renamed them to `validators` and will merge this PR shortly (unless there are other changes you want me to make).

PR #7953, opened 2019-07-29
FIX: DistributedMutex

danielwaterworth:
DistributedMutex would unlock when it shouldn't, when:

 * The mutex was held for too long,
 * When there was an error getting the mutex,

### DIFF
--- a/spec/components/distributed_mutex_spec.rb
+++ b/spec/components/distributed_mutex_spec.rb
@@ -98,4 +98,47 @@ describe DistributedMutex do
     end
   end
 
+  context "executions" do
+    it "should not allow critical sections to overlap" do
+      connections = (0...3).map { DiscourseRedis.new }
+
+      scenario =
+        Concurrency::Scenario.new do |execution|
+          locked = false
+
+          $redis.del('mutex_key')
+
+          connections.each do |connection|
+            connection.unwatch
+          end
+
+          3.times do |i|
+            execution.spawn do
+              begin
+                redis =
+                  Concurrency::RedisWrapper.new(
+                    connections[i],
+                    execution
+                  )
+
+                2.times do
+                  DistributedMutex.synchronize('mutex_key', redis: redis) do
+                    raise "already locked #{execution.path}" if locked
+                    locked = true
+
+                    execution.yield
+
+                    raise "already unlocked #{execution.path}" unless locked
+                    locked = false
+                  end
+                end
+              rescue Redis::ConnectionError
+              end
+            end
+          end
+        end
+
+      scenario.run(runs: 10)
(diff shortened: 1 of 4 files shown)


SamSaffron:
I am ok with this, but how hard is it for you to engineer a test that would fail previously for dist mutex... I think sadly this is one of the rare cases where you would need a mock.

danielwaterworth:
@SamSaffron, I'll give it a go

spec/components/distributed_mutex_spec.rb
```diff
@@ -98,4 +98,47 @@
...
+                    locked = false
+                  end
+                end
+              rescue Redis::ConnectionError
+              end
+            end
+          end
+        end
+
+      scenario.run(runs: 100)
```
SamSaffron: how much time is the spec adding to our test suite? maybe this is a test we only run once in a while? 

The test reads pretty nicely which is nice, very creative use of fibers, interesting way of avoiding full mocking.

danielwaterworth: It takes around half a second to run. It's only exploring 100 paths through the state space, which is by no means exhaustive, but it's enough to find problems with the old implementation. It's certainly helpful to up this number during development. I've tested 10s of thousands of paths with the new implementation.

I started writing a test with a mocked redis, but I couldn't think of a way to make it so that it would fail before the fix and succeed afterwards, but wouldn't be contrived.

I wish I could take credit for the idea, but I didn't come up with it. It's fairly well known. For example, there's a great implementation for the flow programming language (which FoundationDB is written in).

> Simulation is able to conduct a deterministic simulation of an entire FoundationDB cluster within a single-threaded process.

[source](https://apple.github.io/foundationdb/testing.html)

It's also not the first time I've used this technique. I did something similar for [my rust raft library](https://gitlab.com/danielwaterworth/raft-logic), but since I didn't want to use co-routines there, I had to structure the whole library as the controlee instead of the controller.

I'm glad you see the benefits. My mind is spinning with all the possibilities for finding concurrency bugs using this thing.

SamSaffron:
OK I say merge at will but maybe take it down to 10 paths, we will eventually get a heisenfail if something breaks here given the random nature of your walk through. 

I know it is kind of arbitrary drawing the line saying its ok to add 50ms to the test suite time for the vs 500ms, but yeah anyones guess on the best number here is as good as anyone else. We certainly don't want to pay 10 seconds for this, I feel half a second is a tad high.

danielwaterworth:
OK, I will bring down the number of runs.

It won't become a heisentest though, because it seeds the PRNG so that it always explores the same collection of paths.

PR #7955, opened 2019-07-30
DEV: consistent theme lookup in application / crawler / no-ember views

hnb-ku:
history: https://meta.discourse.org/t/customization-in-body-tag-wont-appear-in-some-pages/105167

This PR ensures that theme fields are consistently injected in the different relevant views. Those views are 

1. `application`
2. `crawler`
3. `no-ember`

The application view currently contains all the required theme fields.
1. `head_tag`
2. `header`
3. `footer` (this is actually added in the [application.hbs](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/templates/application.hbs#L29) template and not in the view file)
4. `body_tag`

The crawler view contains three fields
1. `head_tag` 
2. `header`
3. `body_tag` 

_the `footer` field is missing from this view_

The no-ember view contains three fields
1. `head_tag` 
2. `header`
3. `footer` 

_the `body_tag`  field is missing in this view_

This PR adds those missing fields. 

Before (no-ember view)

image

after (no-ember view)

image

Before (crawler view)

theme001

after (crawler view)

theme002

<hr> 

This PR also removes the `raw` helper that was used in some places where we lookup theme html, JS, and translations. 

The reason is that it wasn't used consistently (used in some places but not others) and because it's actually not needed.  

We already use `html_safe` in `theme_lookup` / `theme_js_lookup` and `theme_translations_lookup` to mark them as safe. So the `raw` helper is not really doing anything here.

https://github.com/discourse/discourse/blob/eff1c19e3be473b609c7f585337fd9dd1af968d1/app/helpers/application_helper.rb#L432-L445

I tested to confirm that removing the `raw` helpers wouldn't cause any issues in all the theme fields we lookup. 

This is a simple theme that contains the markup used in the screenshots above.

[discourse_light .tar.gz](https://github.com/discourse/discourse/files/3446014/discourse_light.tar.gz)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -34,9 +34,9 @@
     <%- end %>
 
     <%- unless customization_disabled? %>
-      <%= raw theme_translations_lookup %>
-      <%= raw theme_js_lookup %>
-      <%= raw theme_lookup("head_tag") %>
+      <%= theme_translations_lookup %>
+      <%= theme_js_lookup %>
+      <%= theme_lookup("head_tag") %>
     <%- end %>
 
     <%= render_google_tag_manager_head_code %>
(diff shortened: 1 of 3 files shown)


eviltrout:
Looks good to me! :+1:

PR #7956, opened 2019-07-30
UX: Make profile views consistent with other elements

tshenry:
This is just a little consistency fix. It makes it so that a hidden profile won't show an empty "Views" element.

Screen Shot 2019-07-30 at 2 32 42 AM


eviltrout:
Looks good to me, thanks!

PR #7957, opened 2019-07-31
FIX: Gravatar uploads being dependent on authorized_extensions.

tgxworld:


lib/file_helper.rb
```diff
@@ -124,6 +124,10 @@ def self.memoize(*args)
     (@memoized ||= {})[args] ||= yield
   end
 
+  def self.supported_gravatar_extensions
+    @@supported_gravatar_images ||= Set.new(%w{jpg jpeg png})
```
SamSaffron: This is close ... but you are missing `gif` here. I am fine to punt on svg here ... maybe add a comment saying we explicitly know svg could be a thing but gravatar does not support it.

SamSaffron:
merge at will once comment is addressed.

tgxworld:
@SamSaffron  @eviltrout  THanks for reviewing :+1:

PR #7958, opened 2019-07-31
REFACTOR: Rename SiteSetting.disable_edit_notifications to disable_system_edit_notifications

rishabhnambiar:
The older name could cause some confusion because the setting does not disable all edit notifications, only system ones.


rishabhnambiar:
One question: Like it was done in [c34a6ba](https://github.com/discourse/discourse/commit/c34a6ba674ecf78eb4d561b9f633cc497e973b74#diff-0d741fb5cc6333293e95cf2f23930f55), do I need to add an entry for this change in [deprecated_settings.rb#L6](https://github.com/discourse/discourse/blob/master/lib/site_settings/deprecated_settings.rb#L6)? If yes, would adding this line to the bottom of that list do the trick? _(not sure what `override` does)_

`['disable_edit_notifications', 'disable_system_edit_notifications', true, '2.4']`

gschlager:
I couldn't find any plugin that uses `disable_edit_notifications`, but I guess it's still a good idea to deprecate the old setting.

`['disable_edit_notifications', 'disable_system_edit_notifications', true, '2.4']` is the correct thing to do. When `override` is `true` calling of the old site setting will be forwarded to the new site setting. When it's `false` it will still use the old setting. The latter is useful when the old and new setting behave differently (e.g. one accepts a URL and the other an `Upload` object).

PR #7960, opened 2019-08-01
REFACTOR: Use less queries when moving posts

gschlager:
Previously moving required 4 additional DML statements for each moved post in order to move dependent data. This reduces it to 1 additional statement per post and does all the work after all posts have been moved.

This is a preparation for moving additional data like read state which would have required 3 more queries per post.

### DIFF
--- a/app/models/post_mover.rb
+++ b/app/models/post_mover.rb
@@ -63,6 +63,7 @@ class PostMover
 
     moving_all_posts = (@original_topic.posts.pluck(:id).sort == @post_ids.sort)
 
+    create_temp_table
     move_each_post
     notify_users_that_posts_have_moved
     update_statistics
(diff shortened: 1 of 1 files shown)


app/models/post_mover.rb
```diff
@@ -162,25 +175,50 @@ def move(post)
...
+      INSERT INTO moved_posts(old_topic_id, old_post_id, old_post_number, new_topic_id, new_topic_title, new_post_id, new_post_number)
+      VALUES (#{metadata[:old_topic_id]}, #{metadata[:old_post_id]}, #{metadata[:old_post_number]}, #{metadata[:new_topic_id]}, '#{metadata[:new_topic_title]}', #{metadata[:new_post_id]}, #{metadata[:new_post_number]})
+    SQL
+  end
+
+  def move_incoming_emails
+    DB.exec <<~SQL
+      UPDATE incoming_emails ie
+      SET topic_id = mp.new_topic_id,
+          post_id= mp.new_post_id
```
eviltrout:
```suggestion
          post_id = mp.new_post_id
```

PR #7961, opened 2019-08-01
FIX: Better detection of Apple browsers

xfalcox:
Introduces isIpadOS to our capabilities sniffer, which is currently
compatible with latest iPad beta OS. Older iPad versions will match our
isIOS detection.

Using this new feature, fixes the service worker being blocked in Google
Chrome, Firefox and other browsers while running on Mac OS, where they
work just fine. This enables Desktop PWA install in Mac OS.


SamSaffron:
I can support this ... Mac Chrome should get the service worker. Safari Mac clearly not :) 

Feel free to merge and keep an eye out

PR #7962, opened 2019-08-02
DEV: default to test env when running d/rspec

xrav3nz:
`RAILS_ENV` currently defaults to `developement` when running the `rspec` docker wrapper.

This changes the default to `test` so the tests are running with right setups.


SamSaffron:
oops! good catch

PR #7964, opened 2019-08-02
Poll migration is not idempotent

eduardopoleo:
The  poll migration script (for 2.3.0) is not idempotent due to database constrains on the
poll related objects, namely:
```
polls: index_polls_on_post_id_and_name (post_id,name) UNIQUE
poll_options: index_poll_options_on_poll_id_and_digest  (poll_id,digest) UNIQUE
poll_votes:  index_poll_votes_on_poll_id_and_poll_option_id_and_user_id  (poll_id,poll_option_id,user_id) UNIQUE
```
This change skips a particular poll migration if it's already found on
the db. We also wrap the poll migration in a transaction so that if the
script happens to fail mid-way while migrating a poll we rollback the
poll and all it's related records, thus preventing undesirable partial
states.


SamSaffron:
This is very confusing to me... all migrations run in a transaction so we should not need to layer more transactions here.

eduardopoleo:
Just updated. Thanks for looking into this.

ZogStriP:
Thanks 👍

PR #7966, opened 2019-08-05
FIX: Rename deprecated "refresh" icon to "sync"

xrav3nz:
I picked this up in the console:

> Deprecation notice: Please replace all occurrences of "refresh" with "sync". FontAwesome 4.7 icon names are now deprecated and will be removed in the next release.

Fixed this in the exception page, and tested that it's working as intended. I also did a quick grep through the codebase, this seems to be the only occurrence of "refresh".

cc/ @pmusaraj Thanks!


SamSaffron:
nice, looks good to me!

SamSaffron:
confirmed we ship it here: 

https://github.com/discourse/discourse/blob/67f5ad5ac0383c80647aaebc40cc771829b12b67/lib/svg_sprite/svg_sprite.rb#L167-L167

PR #7967, opened 2019-08-05
FIX: Use multiple multisite databases in parallel tests

danielwaterworth:
These new databases can be managed using the `parallel:*` commands.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -153,6 +153,52 @@ else
   end
 end
 
+module MultiSiteSetup
+  def self.with_temp_file(contents)
+    file = Tempfile.new
+
+    begin
+      file.write(contents)
+      file.rewind
+
+      yield file
+    ensure
+      file.close
+      file.unlink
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -153,6 +153,54 @@ def fab!(name, &blk)
...
+
+    config = {
+      "second" => {
+        "adapter" => 'postgresql',
+        "database" => database.join('_'),
+        "host_names" => ['test2.localhost']
+      }
+    }
+
+    puts YAML.dump(config)
```
eviltrout: Left this in unintentionally?

SamSaffron:
I am really not a fan of multiple multisite dbs, we have like 3 tests that run in multisite this is a lot of change for a feature we almost never use.

SamSaffron:
I would like to close this off this week, its been going for quite a while, I think simply running all the multisite tests in a dedicated single process is the simplest solution here.

PR #7972, opened 2019-08-05
FIX: korean typo

HanguChoi:


gschlager:

HanguChoi:
OK. I'll updates on there. Thank you.

PR #7978, opened 2019-08-06
FIX: allows replacement of digits and symbols emojis

jjaffeux:


app/assets/javascripts/pretty-text/emoji.js.es6
```diff
@@ -21,11 +21,23 @@ export function extendedEmojiList() {
...
-const unicodeRegexp = new RegExp(
-  Object.keys(replacements)
+export function buildReplacementsList(emojiReplacements) {
+  return Object.keys(emojiReplacements)
     .sort()
     .reverse()
-    .join("|") + "|\\B:[^\\s:]+(?::t\\d)?:?\\B",
+    .map(replacement => {
+      // "*️⃣".replace(/*️⃣/, "foo")
+      // would fail if not escaped
```
eviltrout: What's not clear to me is why is the one with the asterisk a special case when the other ones in the test are not handled this way?

PR #7979, opened 2019-08-06
FIX: Use unescaped title as combo-box id

romanrizzi:
User title options are escaped using `Ember.Handlebars.Utils.escapeExpression` for security reasons. We don't want to send the escaped version to the back-end.


darix:
Will this be backported to stable?

PR #7980, opened 2019-08-07
FIX: Notify on Reviewable update.

udan11:
If a post is flagged after an action was already performed on it, it
will update the previous Reviable instance and not create a new one.
The notification logic was implemented in the :create callback which was
completely skipped in this case.

### DIFF
--- a/spec/models/reviewable_spec.rb
+++ b/spec/models/reviewable_spec.rb
@@ -237,12 +237,25 @@ RSpec.describe Reviewable, type: :model do
 
   context "message bus notifications" do
     fab!(:moderator) { Fabricate(:moderator) }
+    let(:post) { Fabricate(:post) }
 
     it "triggers a notification on create" do
       reviewable = Fabricate(:reviewable_queued_post)
       job = Jobs::NotifyReviewable.jobs.last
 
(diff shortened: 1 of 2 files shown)


spec/models/reviewable_spec.rb
```diff
@@ -237,12 +237,25 @@
...
       Jobs.expects(:enqueue).with(:notify_reviewable, has_key(:reviewable_id))
       Fabricate(:reviewable_queued_post)
     end
 
+    it "triggers a notification on update" do
+      Jobs.stubs(:enqueue)
+      reviewable = PostActionCreator.spam(moderator, post).reviewable
+      reviewable.perform(moderator, :disagree)
+
+      Jobs.expects(:enqueue).with(:notify_reviewable, has_key(:reviewable_id))
```
tgxworld: We want to avoid using stubs and mocks in our test suite as much as possible. In the test environment, Sidekiq jobs are pushed into an array that can be accessed via the class. Example: `Jobs::NotifyReviewable.jobs`. This is a good read as well https://github.com/mperham/sidekiq/wiki/Testing. I've cleaned up the existing mocks in https://github.com/discourse/discourse/commit/32b4bdaccd0e72e79e9a3972f40e7f9828d48761 so you can use that as a reference :)

PR #7985, opened 2019-08-08
FIX: Ensure permissions are correctly enforced

nbianca:


ZogStriP:
Looks like the spec is failing?

SamSaffron:
I don't think this change is right. 

If only owners are allowed visibility then ONLY owners+admins are allowed to see it, not moderators. This is an expansion of mod privs.

PR #7987, opened 2019-08-08
FEATURE: Allow themes to override color transformation variables

davidtaylorhq:
Theme developers can now add any of the transformed color variables to their color scheme in about.json. For example

```
  "color_schemes": {
    "Light": {
      "primary": "333333",
      "secondary": "ffffff",
      "primary-low": "ff0000"
    }
  },
```

would override the primary-low variable when compiling SCSS for the color scheme. The primary-low variable will also be visible in administrator color palette UI.


app/models/remote_theme.rb
```diff
@@ -203,24 +203,32 @@ def update_theme_color_schemes(theme, schemes)
...
+        color_scheme_color.hex = override || color[:hex]
+        theme.notify_color_change(color_scheme_color)
+      end
+
+      # Update advanced colors
+      ColorScheme.color_transformation_variables.each do |variable_name|
+        override = normalize_override(colors[variable_name])
+        color_scheme_color = scheme.color_scheme_colors.to_a.find { |c| c.name == variable_name }
+        if override
+          color_scheme_color = color_scheme_color || scheme.color_scheme_colors.build(name: variable_name)
```
ZogStriP:
```suggestion
          color_scheme_color ||= scheme.color_scheme_colors.build(name: variable_name)
```

PR #7988, opened 2019-08-08
FEATURE: Add a preview to the poll builder

CvX:
Closes https://meta.discourse.org/t/show-preview-in-poll-builder/117323/6

poll builder preview

Shout out to @pmusaraj for the recent poll [builder tweaks](https://github.com/discourse/discourse/commit/78500fb77017ae1defd9d1a77f26164493847624). Saved me some CSS tweaking! 😃

I've hardcoded both column widths, which isn't ideal, but modal's `max-width` means that it expands horizontally as you type (until it reaches that max width). As far as I know, there is no way to force a flex element to fill all the available space up to the `max-width`. I'm open to suggestions. 😉 

Oh, and the `poll-ui-builder.hbs` diff looks messy only because I've wrapped all the poll options in a `<div class="options">`.

### DIFF
--- a/app/assets/stylesheets/common/base/modal.scss
+++ b/app/assets/stylesheets/common/base/modal.scss
@@ -142,7 +142,7 @@
   overflow: auto;
 
   &:not(.history-modal) {
-    .modal-body:not(.reorder-categories) {
+    .modal-body:not(.reorder-categories):not(.poll-ui-builder) {
       max-height: none !important;
     }
   }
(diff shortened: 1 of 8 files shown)


eviltrout:
(confirmed CLA was signed)

eviltrout:
This looks quite good to me. Do you have any feedback @jjaffeux ?

jjaffeux:
Also minor but one .poll-ui-builder-form is in a poll.scss file and ther other in poll-ui-builder.scss file, should we move all in poll.scss or have another poll-ui-builder.scss for mobile?.

CvX:
Both are good points @jjaffeux, I've updated the PR. 🙂

CvX:
Tests failed due to an unrelated flaky test. Forcing rebuild.

eviltrout:
Thanks, merging it in.

PR #7989, opened 2019-08-08
FEATURE: Publish read state on group messages.

romanrizzi:


app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -35,6 +35,34 @@ export const ListItemDefaults = {
...
   "data-topic-id": Ember.computed.alias("topic.id"),
 
+  init() {
+    this._super(...arguments);
+
+    if (typeof this.get("topic.read_by_group_member") !== "undefined") {
+      this.messageBus.subscribe(
+        `/private-messages/group-read/${this.get("topic.id")}`,
+        data => {
+          const node = document.querySelectorAll(
```
jjaffeux: Might be wrong but I think 

```
document.querySelector(`.indicator-topic-${data.topic_id}`)
```

would do the same

app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -35,6 +35,34 @@ export const ListItemDefaults = {
   attributeBindings: ["data-topic-id"],
   "data-topic-id": Ember.computed.alias("topic.id"),
 
+  init() {
+    this._super(...arguments);
+
+    if (typeof this.get("topic.read_by_group_member") !== "undefined") {
```
jjaffeux: Is there any reason

```
if (this.get("topic.read_by_group_member")) {}
```

wouldn't work ?

romanrizzi: We want to be subscribed even if  `topic.read_by_group_member` is false, that's why we explicitly check for undefined.

app/assets/javascripts/discourse/components/topic-list-item.js.es6
romanrizzi: Thanks, @jjaffeux <3

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -52,6 +52,38 @@ export function buildButton(name, widget) {
...
+registerButton("read-count", attrs => {
+  if (attrs.showReadIndicator) {
+    const count = attrs.readCount;
+    if (count > 0) {
+      return {
+        action: "toggleWhoRead",
+        title: "post.controls.read_indicator",
+        className: "button-count read-indicator",
+        contents: count,
+        icon: "",
```
eviltrout: I think if you don't want an icon you should be able to leave this attribute out? If that doesn't work we should adjust the button code to handle `undefined` instead of an empty string

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -52,6 +52,38 @@ export function buildButton(name, widget) {
...
+    }
+  }
+});
+
+registerButton("read", attrs => {
+  const disabled = attrs.readCount === 0;
+  if (attrs.showReadIndicator) {
+    return {
+      action: "toggleWhoRead",
+      className: "",
```
eviltrout: Same comment as above but for `className` here - we should be able to omit it.

app/serializers/listable_topic_serializer.rb
```diff
@@ -121,6 +122,16 @@ def unpinned
     PinnedCheck.unpinned?(object, object.user_data)
   end
 
+  def read_by_group_member
+    return false unless object.respond_to?(:minimum_unread_count)
```
eviltrout: In what case would the object not have this method?

romanrizzi: I Should have added a comment here. This attribute is dynamically calculated [here](https://github.com/discourse/discourse/pull/7989/files?file-filters%5B%5D=.rb#diff-2c209e01ac5c122f000094f5c285e082R1063).

app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -35,6 +35,38 @@ export const ListItemDefaults = {
   attributeBindings: ["data-topic-id"],
   "data-topic-id": Ember.computed.alias("topic.id"),
 
+  didInsertElement() {
+    if (typeof this.get("topic.read_by_group_member") !== "undefined") {
```
eviltrout: You should have a `this._super(...arguments)` in here and `willDestroyElements` just to be safe.

app/serializers/listable_topic_serializer.rb
eviltrout: I see, that should be OK although traditionally `respond_to?` is slow. We probably won't notice it.

PR #7990, opened 2019-08-09
UX: Link users to `..preferences/notifications` directly on `new` & `unread`

rishabhnambiar:
## /new

preffs

## /unread

dasf

---
Right now, clicking these takes users to `../preferences`. Then, they have to click around
to find where the notifications settings are. Taking them to `..preferences/notifications` could be a tiny but useful change, imo.

If this change is welcome but we don't want to add fiddly urls in `client.en.yml` like I've done, would it be better to change the definition of `userPrefsUrl` [here](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/controllers/discovery/topics.js.es6#L168) or maybe define a new one instead?


ZogStriP:
> If this change is welcome but we don't want to add fiddly urls in `client.en.yml` like I've done, would it be better to change the definition of `userPrefsUrl` [here](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/controllers/discovery/topics.js.es6#L168) or maybe define a new one instead?

Yes, we might want something like `userPrefsNotificationsUrl` so that we can link directly to the notifications tab.

rishabhnambiar:
Done 👌 

I've replaced the variable instead of adding a new one because it wasn't being used anywhere else.

SamSaffron:
I have a strong feeling this will bite us back with translations. @OsamaSayegh had to deal with a similar issue with another one. 

It would be nice if we could just change it all like this, but next time we update from transifex it will override. 

How do we feed this into transifex?

rishabhnambiar:
> How do we feed this into transifex?

I'm not really sure. I'll do some digging later but considering that this is a very very small change, we can close this for now 👍

PR #7991, opened 2019-08-09
UX: Improve error handling for common OmniAuth exceptions

davidtaylorhq:
This displays more useful messages for the most common issues we see:
- CSRF (when the user switches browser)
- Invalid IAT (when the server clock is wrong)
- OAuth::Unauthorized for OAuth1 providers (e.g. Twitter), when the credentials are incorrect

This commit also stops earlier for disabled authenticators. Now we stop at the request phase, rather than the callback phase.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -91,7 +91,7 @@ class Users::OmniauthCallbacksController < ApplicationController
   end
 
   def failure
-    error_key = params[:message].to_s.gsub(/[^a-z0-9_-]/, "") || "generic"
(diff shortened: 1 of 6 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -91,7 +91,8 @@ def complete
   end
 
   def failure
-    flash[:error] = I18n.t("login.omniauth_error")
+    error_key = params[:message].to_s.gsub(/[^a-z0-9_-]/, "") || "generic"
```
ZogStriP:
```suggestion
    error_key = params[:message].to_s.gsub(/[^\w-]/, "") || "generic"
```

PR #7995, opened 2019-08-11
make docker dev install posix compliant

spirobel:
we cant expect /bin/bash to be around on every system. Depending on
/bin/sh instead makes the docker install for development work on nixos
aswell.


eviltrout:
I'm not sure all these scripts are compatible with regular `sh` - @SamSaffron wrote most of them and could chip in.

SamSaffron:
Oops... looks like we need to close this, sorry, the changes are not bin/sh compliant :(

PR #7996, opened 2019-08-11
REFACTOR: user-menu-links widget for extensibility

xrav3nz:
> CONTEXT: [Quick access to bookmarks and messages on user menu](https://meta.discourse.org/t/quick-access-to-bookmarks-and-messages-on-user-menu/32787/11)

This extracts a few helper methods in `UserMenuLinks` widget, so that link definitions and rendering can be easily extended by plugins and/or themes. (Only bookmarks and messages are extracted as I only need the two, but I am happy to extract the rest if this is the preferred design)

Let me know if you have other APIs in mind! Thanks. 💪


eviltrout:
Thanks!

PR #7997, opened 2019-08-12
FIX: RIP swipe-in menus on Android

featheredtoast:
This removes the side swipe-in menus for android.


jjaffeux:
@featheredtoast sorry I didn't follow any conversation on this, do you have some context please?

davidtaylorhq:
Latest Android uses swipe-in gestures for navigation, so we can't use them for Discourse menus

https://www.androidcentral.com/android-q-getting-rid-back-button-side-swipe-gesture

PR #7998, opened 2019-08-12
DEV: select-kit 2

jjaffeux:
This will soon be fully forbidden, allowInternalMutation=false is only a temporary option while I work my way through all the cases we have.


SamSaffron:
Any news on this one?

jjaffeux:
also @SamSaffron I'm rebasing very frequently so there's not risk on this side at least

PR #7999, opened 2019-08-12
FIX: Do not create a double like notification.

udan11:
When a user liked, unliked and liked again the same post, the poster
would receive a notification such as "X and X liked ...". This happened
because PostActionNotifier.post_action_created was called twice.


lib/post_action_creator.rb
```diff
@@ -205,7 +205,6 @@ def create_post_action
       post_action.recover!
       action_attrs.each { |attr, val| post_action.public_send("#{attr}=", val) }
       post_action.save
-      PostActionNotifier.post_action_created(post_action)
```
udan11: See `#perform` and line 91.

PR #8002, opened 2019-08-14
DEV: Mount plugin symlinks to dev docker container

xrav3nz:
When developing using docker, in order to support symlinks in the
`plugins/` directory, this reads the symlinks' values and mounts them to
the dev docker container.

Tested in macOS 10.14.5 and Ubuntu 18.04.

Open to suggestions, thanks!


SamSaffron:
This looks good to me, fancy, I like how you docker running part. :+1:

PR #8003, opened 2019-08-14
FIX: Make DistributedMutex threadsafe

danielwaterworth:
Since we use the same redis connection in multiple threads, a rogue
transaction in another thread can trample the connection state (watched
keys) that we need to acquire and release the lock properly.

This is fixed by preventing other threads from using the connection when
we are performing these actions.


SamSaffron:
Is this good to merge? I know how complex it is to test.

PR #8005, opened 2019-08-14
FIX: Race-condition in fallback handlers

danielwaterworth:
Calling `verify_master` in multiple threads simultaneously would cause
multiple threads to be spawned.


ZogStriP:
Ok, let's try it 👍

PR #8006, opened 2019-08-14
FIX: ensures reports can't modify records

jjaffeux:
We do this in discourse-data-explorer and I think it totally makes sense to have the same guard in our reports to prevent any mistake and future usage.


SamSaffron:
Looks safe to me, the thing to watch for is pgbouncer but since you are in a transaction already we are good!

PR #8008, opened 2019-08-14
FEATURE: Embed topics list on remote sites via Javascript API.

eviltrout:
This adds support for a `<d-topics-list>` tag you can embed in your site
that will be rendered as a list of discourse topics. Any attributes on
the tag will be passed as filters. For example:

`<d-topics-list discourse-url="URL" category="1234">` will filter to category 1234.

To use this feature, enable the `embed topics list` site setting. Then
on the site you want to embed, include the following javascript:

`<script
src="http://URL/javascripts/embed-topics.js"></script>`

Where `URL` is your discourse forum's URL.

Then include the `<d-topics-list discourse-url="URL">` tag in your HTML document and it will
be replaced with the list of topics.

### DIFF
--- a/app/assets/javascripts/embed-application.js.no-module.es6
+++ b/app/assets/javascripts/embed-application.js.no-module.es6
@@ -24,17 +24,20 @@
 
   window.onload = function() {
     // get state info from data attribute
-    var header = document.querySelector("header");
+    var embedState = document.querySelector("[data-embed-state]");
     var state = "unknown";
-    if (header) {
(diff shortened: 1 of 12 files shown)


eviltrout:
I could use some reviews on this so I'm tagging a few random people!

SamSaffron:
I think you want anon cache here, this can be quite a chunk of anon traffic if used in certain ways.

eviltrout:
Good call @SamSaffron, I've added the caching.

PR #8010, opened 2019-08-15
UX: Consistent placement of category-title-before plugin outlet

tshenry:
The following changes are included in this PR:

- Provide more consistent placement of the `category-title-before` plugin outlet.
- Remove unnecessary `inline-block` styling on the mobile category list item heading

Note that I'm leaving out the "Boxes with subcategories" category style for now. I feel like the boxes in this style are too small to facilitate extra icons. Also, I'm not sure what we'd want to do about the listed subcategories. Currently, they are just standard links that don't even include the lock icon if they are a restricted category.

@pmusaraj if you could take a look at this and make sure there isn't anything else you think should be included, that would be great! I'll be creating a PR for the Category Icons component that adjusts for these changes (assuming they are accepted!)

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
+++ b/app/assets/javascripts/discourse/templates/components/categories-boxes-with-topics.hbs
@@ -11,10 +11,10 @@
           {{/if}}
 
           <h3>
+            {{category-title-before category=c}}
             {{#if c.read_restricted}}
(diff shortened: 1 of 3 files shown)


pmusaraj:
This looks good to me @tshenry, I would only change the name of the PR to "UX: Consistent placement of category-title-before plugin outlet" since this only affects plugins/components using that outlet (not default Discourse category list layout).

PR #8011, opened 2019-08-15
FEATURE: favorites emojis will also show in composer autocomplete

jjaffeux:


app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -21,6 +21,7 @@ import { wantsNewWindow } from "discourse/lib/intercept-click";
 import { translations } from "pretty-text/emoji/data";
 import { emojiSearch, isSkinTonableEmoji } from "pretty-text/emoji";
 import { emojiUrlFor } from "discourse/lib/text";
+import EmojisStore from "discourse/lib/emojis-store";
```
eviltrout: I wonder if this would be better as an ember service that's injected? It seems to be used in components only and that would save a little code.

app/assets/javascripts/discourse/lib/emojis-store.js.es6
```diff
@@ -0,0 +1,50 @@
+import KeyValueStore from "discourse/lib/key-value-store";
+
+const EMOJI_USAGE = "emojiUsage";
+const EMOJI_SELECTED_DIVERSITY = "emojiSelectedDiversity";
+const TRACKED_EMOJIS = 15;
+const STORE_NAMESPACE = "discourse_emojis_";
+
+export default class EmojiStore {
```
eviltrout: I notice this is `EmojiStore` and elsewhere it's EmojisStore (with an s). It looks like in English it can be either emojis or emoji but we should standardize on one. I think our codebase uses "emoji" much more for the plural format so we should probably stick with that.

app/assets/javascripts/discourse/lib/emojis-store.js.es6
```diff
@@ -0,0 +1,50 @@
...
+  constructor() {
+    this.store = new KeyValueStore(STORE_NAMESPACE);
+
+    this._setup();
```
eviltrout: Since `_setup` is only called here I would probably inline the function.

PR #8013, opened 2019-08-15
FEATURE: Incorporate PWA install prompt into Discourse UI

xfalcox:
This is mainly done so Discourse forums stop nagging people to install
on the very first visits to a website.

We will prevent the native install "mini-info" bar from ever appearing,
capture the event that pops with it, and delay it until the user meets
our criteria, which currently is trust_level 1.

If the event happens and the user meets our criteria we show a Discourse
alert banner proposing the install to the user. Dismissal of the banner
is recorded so the user ins't bothered anymore on the same device.

### DIFF
--- a/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
+++ b/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
@@ -24,14 +24,19 @@ export default Ember.Component.extend({
     window.addEventListener("beforeinstallprompt", this._promptEventHandler);
   },
 
+  @on("willDestroyElement")
+  _unregisterListener() {
+    window.removeEventListener("beforeinstallprompt", this._promptEventHandler);
+  },
+
   @computed
   bannerDismissed: {
     set(value) {
-      localStorage.setItem(USER_DISMISSED_PROMPT_KEY, value);
-      return localStorage.getItem(USER_DISMISSED_PROMPT_KEY);
+      this.keyValueStore.set({key: USER_DISMISSED_PROMPT_KEY, value: value});
+      return this.keyValueStore.get(USER_DISMISSED_PROMPT_KEY);
     },
     get() {
-      return localStorage.getItem(USER_DISMISSED_PROMPT_KEY);
+      return this.keyValueStore.get(USER_DISMISSED_PROMPT_KEY);
     }
   },
 
@@ -49,13 +54,11 @@ export default Ember.Component.extend({
     turnOn() {
       this.set("bannerDismissed", true);
       this.deferredInstallPromptEvent.prompt();
-      window.removeEventListener(
-        "beforeinstallprompt",
-        this._promptEventHandler
-      );
+      this._unregisterListener();
     },
     dismiss() {
       this.set("bannerDismissed", true);
+      this._unregisterListener();
     }
   }
 });
--- a/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
+++ b/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
@@ -32,7 +32,7 @@ export default Ember.Component.extend({
   @computed
   bannerDismissed: {
     set(value) {
-      this.keyValueStore.set({key: USER_DISMISSED_PROMPT_KEY, value: value});
+      this.keyValueStore.set({ key: USER_DISMISSED_PROMPT_KEY, value: value });
       return this.keyValueStore.get(USER_DISMISSED_PROMPT_KEY);
     },
     get() {
--- a/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
+++ b/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
@@ -15,7 +15,7 @@ export default Ember.Component.extend({
     this.set("deferredInstallPromptEvent", event);
   },
 
-  @on("init")
+  @on("didInsertElement")
   _registerListener() {
     this._promptEventHandler = Ember.run.bind(
       this,
@@ -54,11 +54,9 @@ export default Ember.Component.extend({
     turnOn() {
       this.set("bannerDismissed", true);
       this.deferredInstallPromptEvent.prompt();
-      this._unregisterListener();
     },
     dismiss() {
       this.set("bannerDismissed", true);
-      this._unregisterListener();
     }
   }
 });
--- a/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
+++ b/app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
@@ -3,38 +3,42 @@ import {
   on
 } from "ember-addons/ember-computed-decorators";
 
-const userDismissedPromptKey = "dismissed-pwa-install-banner";
+const USER_DISMISSED_PROMPT_KEY = "dismissed-pwa-install-banner";
 
 export default Ember.Component.extend({
   deferredInstallPromptEvent: null,
 
+  _handleInstallPromptEvent(event) {
+    // Prevent Chrome 76+ from automatically showing the prompt
+    event.preventDefault();
+    // Stash the event so it can be triggered later
+    this.set("deferredInstallPromptEvent", event);
+  },
+
   @on("init")
   _registerListener() {
-    var that = this;
-    window.addEventListener("beforeinstallprompt", e => {
-      // Prevent Chrome 76+ from automatically showing the prompt
-      e.preventDefault();
-      // Stash the event so it can be triggered later
-      that.set("deferredInstallPromptEvent", e);
-    });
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
...
+} from "ember-addons/ember-computed-decorators";
+
+const userDismissedPromptKey = "dismissed-pwa-install-banner";
+
+export default Ember.Component.extend({
+  deferredInstallPromptEvent: null,
+
+  @on("init")
+  _registerListener() {
+    var that = this;
```
jjaffeux: why do you need that here ?

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
...
+    },
+    get() {
+      return localStorage.getItem(userDismissedPromptKey);
+    }
+  },
+
+  @computed("deferredInstallPromptEvent", "bannerDismissed")
+  showPWAInstallBanner() {
+    return (
+      this.currentUser && // User must be logged in
```
jjaffeux: can probably be replaced by:

```
this.get("currentUser.trust_level") > 0
```

app/assets/javascripts/discourse/templates/components/pwa-install-banner.hbs
```diff
@@ -0,0 +1,10 @@
+{{#if showPWAInstallBanner}}
+  <div class="row">
+    <div class="pwa-install-banner alert alert-info">
+      <div class="close" {{action "dismiss"}}>
+        {{d-icon 'times'}}
+      </div>
+      {{i18n 'pwa.install_banner.question'}} <a {{action "turnon"}}>{{i18n 'pwa.install_banner.link' title=siteSettings.title}}</a>.
```
jjaffeux: Can we wrap `{{i18n 'pwa.install_banner.question'}}` in a span please? I don't like text with no direct html parent.

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
...
+
+const userDismissedPromptKey = "dismissed-pwa-install-banner";
+
+export default Ember.Component.extend({
+  deferredInstallPromptEvent: null,
+
+  @on("init")
+  _registerListener() {
+    var that = this;
+    window.addEventListener("beforeinstallprompt", e => {
```
jjaffeux: Is this event listener being removed somewhere ? I'm afraid it's leaking.

Also you are setting a property to an even, which probably holds a reference to a DOM object and is probably going to leak too. We should ensure this is unset on destruction of the component.

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
...
+    get() {
+      return localStorage.getItem(userDismissedPromptKey);
+    }
+  },
+
+  @computed("deferredInstallPromptEvent", "bannerDismissed")
+  showPWAInstallBanner() {
+    return (
+      this.currentUser && // User must be logged in
+      this.currentUser.trust_level > 0 && // Be at least trust_level 1
```
jjaffeux: I think this comment is not necessary. `this.currentUser.trust_level > 0` seems very clear that we are checking we are above trust_level 1

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
jjaffeux: This comment is also not necessary IMO

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
+import {
+  default as computed,
+  on
+} from "ember-addons/ember-computed-decorators";
+
+const userDismissedPromptKey = "dismissed-pwa-install-banner";
```
jjaffeux: Usually this kind of const should be UPPERCASE: USER_DISMISSED_PROMPT_KEY

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,53 @@
...
+      this.currentUser && // User must be logged in
+      this.currentUser.trust_level > 0 && // Be at least trust_level 1
+      this.deferredInstallPromptEvent && // Pass the browser engagement checks
+      !window.matchMedia("(display-mode: standalone)").matches && // Not be in the installed PWA already
+      !this.bannerDismissed // Have not a previously dismissed install banner
+    );
+  },
+
+  actions: {
+    turnon() {
```
jjaffeux: Should be `turnOn` no ?

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
xfalcox: Just so I can set on line 18. Is there a better way?

jjaffeux:
```
  window.addEventListener("beforeinstallprompt", e => {
      // Prevent Chrome 76+ from automatically showing the prompt
      e.preventDefault();
      // Stash the event so it can be triggered later
      this.set("deferredInstallPromptEvent", e);
    });
  }
```

Should just work as your are using fat arrow. It doesn't?

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,61 @@
...
+      this,
+      this._handleInstallPromptEvent
+    );
+    window.addEventListener("beforeinstallprompt", this._promptEventHandler);
+  },
+
+  @computed
+  bannerDismissed: {
+    set(value) {
+      localStorage.setItem(USER_DISMISSED_PROMPT_KEY, value);
```
eviltrout: It would be better I think to use the `KeyValueStore` for this. It has some extra checks about if localStorage works.

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,61 @@
...
+  deferredInstallPromptEvent: null,
+
+  _handleInstallPromptEvent(event) {
+    // Prevent Chrome 76+ from automatically showing the prompt
+    event.preventDefault();
+    // Stash the event so it can be triggered later
+    this.set("deferredInstallPromptEvent", event);
+  },
+
+  @on("init")
```
eviltrout: In general you'll want to attach events on `didInsertElement`, because it's after the component is inserted into the DOM.

You should also tear the event down in a `willDestroyElement` otherwise this event will leak in test mode where we tear up/down repeatedly.

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,64 @@
...
```
jjaffeux: I agree it would be better in didInsertElement

Basic rule of thumb I use:
- any internal object state not directly impacting the DOM, setup in init
- anything with DOM, setup in didInsertElement

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,64 @@
...
+      !window.matchMedia("(display-mode: standalone)").matches && // Not be in the installed PWA already
+      !this.bannerDismissed // Have not a previously dismissed install banner
+    );
+  },
+
+  actions: {
+    turnOn() {
+      this.set("bannerDismissed", true);
+      this.deferredInstallPromptEvent.prompt();
+      this._unregisterListener();
```
jjaffeux: is there a risk this would be triggered again ? If not I would remove both calls and let willDestroy take care of this.

app/assets/javascripts/discourse/components/pwa-install-banner.js.es6
```diff
@@ -0,0 +1,62 @@
...
+
+  @on("willDestroyElement")
+  _unregisterListener() {
+    window.removeEventListener("beforeinstallprompt", this._promptEventHandler);
+  },
+
+  @computed
+  bannerDismissed: {
+    set(value) {
+      this.keyValueStore.set({ key: USER_DISMISSED_PROMPT_KEY, value: value });
```
jjaffeux:
```suggestion
      this.keyValueStore.set({ key: USER_DISMISSED_PROMPT_KEY, value });
```

PR #8014, opened 2019-08-16
FEATURE: Add site setting to show more detailed #404 errors.

udan11:
If the setting is turned on, then the user will receive information
about the subject: if it was deleted or requieres some special access to
a group (only if the group is public). Otherwise, the user will receive
a generic #404 error message. For now, this change affects only the
topics and categories controller.

This commit also tries to refactor some of the code related to error
handling. To make error pages more consistent (design-wise), the actual
error page will be rendered server-side.

---

### Preview

* Topic in a secret category (with public group)

topic in secret category

* Deleted topic in a public category (with public group)

a deleted topic

* Deleted topic in a secret category

deleted topic in secret category

* Topic in a secret category (with private group)

topic in super secret category

* Secret category (with a public group)

category page


app/models/category.rb
```diff
@@ -265,6 +265,15 @@ def description_text
     end
   end
 
+  def access_category_via_group
+    Group
+      .joins(:category_groups)
+      .where("category_groups.category_id = ?", self.id)
+      .where("groups.public_admission OR groups.allow_membership_requests")
+      .order(:allow_membership_requests)
```
ZogStriP: Sorting by a boolean is a little odd.

app/controllers/application_controller.rb
```diff
@@ -784,9 +790,12 @@ def build_not_found_page(status = 404, layout = false)
     end
 
     @container_class = "wrap not-found-container"
-    @slug = (params[:slug].presence || params[:id].presence || "").tr('-', '')
+    @slug = (params[:slug].presence || params[:id].presence || "").tr('-', ' ')
```
ZogStriP: Why replace the dash (`-`) by space (` `) in the slug?

app/assets/javascripts/discourse/models/post-stream.js.es6
```diff
@@ -1054,31 +1054,10 @@ export default RestModel.extend({
...
-        topic.set("message", I18n.t("topic.invalid_access.description"));
-      } else {
-        topic.set("message", I18n.t("topic.invalid_access.login_required"));
-      }
-      return;
-    }
-
-    // Otherwise supply a generic error message
-    topic.set("message", I18n.t("topic.server_error.description"));
+    topic.set("notFoundHtml", result.jqXHR.responseJSON.extras.html);
```
ZogStriP: Are we 100% sure the response always has the `extras.html` attributes?

SamSaffron:
Since this is opt in there is not too much risk here, the big thing to watch for is that we don't want perf of standard 404s to get any slower after this change (if site setting is not enabled) 

Can we get this rebased and ready to merge addressing @ZogStriP's concerns?

app/controllers/application_controller.rb
udan11: Let's take [an example](https://meta.discourse.org/t/discourse-2-4-0-beta5-release-notes/12998810).

The search input field has all the words in the slug concatenated.

app/models/category.rb
udan11: It is done this way to favor groups with public admission over groups with membership requests.

PR #8015, opened 2019-08-17
FIX: uses simpler patternj for custom href on extra nav items

jjaffeux:
The main advantage of this solution is that it will be called on each re-render whereas the previous is not once href has been set.

Example API:
```
      api.addNavigationBarItem({
        name: "foo",
        displayName: "Foo",
        customHref: function(category, args) {
          const router = api.container.lookup("service:router");
          const queryParams = { bar: "1" };
          return router.urlFor(router.currentRouteName, category, {
            queryParams
          });
        }
      });
```


SamSaffron:
Cool, I say we try this out!

PR #8019, opened 2019-08-19
FIX: Improve Onebox detection

nbianca:
Follow-up to 7c83d2eeb261ac676a8320e6a704752c56fd242e.

### DIFF
--- a/app/models/post_analyzer.rb
+++ b/app/models/post_analyzer.rb
@@ -9,10 +9,11 @@ class PostAnalyzer
     @raw = raw
     @topic_id = topic_id
     @onebox_urls = []
+    @found_oneboxes = false
   end
 
   def found_oneboxes?
-    @onebox_urls.present?
+    @found_oneboxes
   end
 
   def has_oneboxes?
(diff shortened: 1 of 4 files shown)


ZogStriP:
Looks good, but the build is broken. Can you have a look?

app/models/post_analyzer.rb
```diff
@@ -36,7 +37,9 @@ def cook(raw, opts = {})
     result = Oneboxer.apply(cooked) do |url|
       @onebox_urls << url
```
ZogStriP: Should we keep track of `onebox_urls` if `Oneboxer.cached_onebox` might fail later?

app/models/post_analyzer.rb
```diff
@@ -126,7 +129,7 @@ def raw_links
 
   # How many links are present in the post
   def link_count
-    raw_links.size
+    raw_links.size + @onebox_urls.size
```
ZogStriP: Why that change?

nbianca: Some links were either in the `raw_links` array or in the `@onebox_urls` array, so `link_count` was returning only a partial count of the links present in the post.

app/models/post_analyzer.rb
nbianca: `@onebox_urls` keeps all links that should be Oneboxed whether they succeed or not. It is used later to determine the number of links.

PR #8020, opened 2019-08-20
FEATURE: Protect against replay attacks when using TLS 1.3 0-RTT

xfalcox:


SamSaffron:
Do we need to wait till we mark the new image in launcher or is this compatible with the current on. 

If it is compatible I guess merge away

PR #8023, opened 2019-08-20
DEV: Remove code deprecated by the new Reviewable API

romanrizzi:
`FlagQuery` is the only deprecation left, we'll have to wait for a little more since `discourse-assign` still depends on it.


SamSaffron:
Oh I am so happy to see code removed!!! 

https://www.folklore.org/StoryView.py?story=Negative_2000_Lines_Of_Code.txt

Feel free to merge first thing whey you start your day, and watch closely!

PR #8025, opened 2019-08-21
FEATURE: Publish read state on group messages. (Originally introduced in #7989)

romanrizzi:


SamSaffron:
read_by_member does not sound right to me. 

We should be storing the post_number that we are up to. I get the appeal of topic_allowed_groups here, but I think another table makes more sense here.

Non fancy implementation is

TopicGroup (to match TopicUser naming wise)

`topic_id, group_id, last_read_post_number`

Fancy implementation would maybe have a vector of user -> post_number in a column for top N. But this is way too complex for first iteration.

romanrizzi:
@SamSaffron I just made the changes to pre-calculate `last_read_post_number` in the new `topic_groups` table. 

> Fancy implementation would maybe have a vector of user -> post_number in a column for top N. But this is way too complex for first iteration.

Why this approach would be better?

SamSaffron:
Can you upload a few screenshots? Overall the approach here looks right and code looks nice and well tested.

I would wait till monday to merge, but first thing on monday when you start your day maybe merge it? Then gather feedback from community team during your day.

db/migrate/20190820192341_create_topic_group_table.rb
```diff
@@ -0,0 +1,14 @@
...
+class CreateTopicGroupTable < ActiveRecord::Migration[5.2]
+  def change
+    create_table :topic_groups do |t|
+      t.integer :group_id, null: false
+      t.integer :topic_id, null: false
+      t.integer :last_read_post_number, null: false, default: 0
+      t.timestamps null: false
+    end
+
+    add_index :topic_groups, %i[group_id topic_id], unique: true
```
SamSaffron: hmmm so this is starting blank? Why not populate it in the migration?

SamSaffron:
I am mixed here, but it feel like there should be a site setting here to disable this? @coding-horror ? Should this be default on? (read indicator)

db/migrate/20190820192341_create_topic_group_table.rb
romanrizzi: Since the feature is disabled by default, the data we insert could become outdated quickly.

PR #8027, opened 2019-08-21
FEATURE: Add hidden attribute to PostWordpressSerializer

scossar:
Adds the `hidden` attribute to the `PostWordpressSerializer` so that hidden posts can be handled on WordPress.

Related topic on meta: https://meta.discourse.org/t/flagged-and-hidden-posts-whisper-authors-showing-up-in-wordpress/126409


scossar:
This is an easy change to make on Discourse and handle with WordPress, but there is no reason to send hidden posts to WordPress. Instead of adding the `hidden` attribute, we could update the `TopicViewWordpressSerializer` to filter out hidden posts.

PR #8032, opened 2019-08-22
Filter out hidden posts from TopicViewWordpressSerializer

scossar:
We never need to send hidden posts to WordPress.


ZogStriP:
It's even better to filter these out at the request level ;)

Since it's a small change, I've pushed it in 2ff5592941ca567a19a792a2a4c4bef71d858277

PR #8035, opened 2019-08-23
FEATURE: Adds a pop up that shows a more detailed score for reviewables

eviltrout:
If you click a (?) icon beside the reviewable status a pop up will
appear with expanded informatio that explains how the reviewable got its
score, and how it compares to system thresholds.

image


eviltrout:
Thanks for the feedback @jjaffeux, I made both changes.

PR #8036, opened 2019-08-24
DEV: Trigger event when Middleware::RequestTracker is registered.

udan11:
This is required to let plugins register middlewares before `Middleware::RequestTracker`. Doing it in the global scope is too early for `Middleware::RequestTracker` to exist and doing it in `after_initialize` is impossible because the middleware stack gets frozen.


SamSaffron:
Hmm ... I think this makes a bit more sense, a lot more flexible:

https://github.com/discourse/discourse/commit/208c6389001f2818be3f2391b8c403316efc4aae

PR #8038, opened 2019-08-25
DEV: adds event hook when add/remove user to group

jjaffeux:
This is needed for a project of mine 🤐 ... and also I think it makes sense to have this hook in core.

### DIFF
--- a/app/jobs/regular/automatic_group_membership.rb
+++ b/app/jobs/regular/automatic_group_membership.rb
@@ -22,7 +22,7 @@ module Jobs
         .where(staged: false)
         .find_each do |user|
         next unless user.email_confirmed?
-        group.add(user)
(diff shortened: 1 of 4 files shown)


SamSaffron:
I am fine to merge this provided you are comfortable with this not triggering when some automatic group membership is triggered. 

Also I think we should probably add a spec here (just tack on to existing specs) cause this will ensure we have this will not break.

jjaffeux:
> I am fine to merge this provided you are comfortable with this not triggering when some automatic group membership is triggered.
> 
> Also I think we should probably add a spec here (just tack on to existing specs) cause this will ensure we have this will not break.

Cool thanks, I did :+1:

From my tests, it seems like it's triggering for automatic group membership too.

PR #8041, opened 2019-08-26
Feature/Fix: Flagged posts user notifications

romanrizzi:
### Changes:

- FIX: User should get notified when a post is deleted 
- FEATURE: Notify posters when restoring flagged posts 

### Description:

The `PostDestroyer` looks for pending reviewables associated with the post (It'll never find one in this case) and tries to act on them. This logic is correct when manually deleting a post that was flagged, but it won't work when calling the destroyer from a reviewable action.

The destroyer now receives the reviewable object to get some context and directly notify the poster without trying to act again on the reviewable.

Also, a new PM will be created when disagreeing with the reviewable.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -2547,7 +2547,23 @@ en:
         The community flagged this post and now it is hidden. **Because this post has been hidden more than once, your post will now remain hidden until it is handled by a staff member.**
 
         For additional guidance, please refer to our [community guidelines](%{base_url}/guidelines).
+    
+    flags_disagreed:
+      title: "Flagged port restored by staff"
+      subject_template: "Flagged post restored by staff"
+      text_body_template: |
+        Hello,
 
+        This is an automated message from %{site_name} to let you know that [your post](%{base_url}%{url}) was restored.
+
+        This post was flagged by the community and a staff member opted to restore it.
+
+        [details="Click to expand restored post"]
+        ``` markdown
+        %{flagged_post_raw_content}
+        ```
+        [/details]
+  
     flags_agreed_and_post_deleted:
       title: "Flagged post removed by staff"
       subject_template: "Flagged post removed by staff"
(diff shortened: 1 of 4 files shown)


config/locales/server.en.yml
```diff
@@ -2547,7 +2547,23 @@ en:
         The community flagged this post and now it is hidden. **Because this post has been hidden more than once, your post will now remain hidden until it is handled by a staff member.**
 
         For additional guidance, please refer to our [community guidelines](%{base_url}/guidelines).
+    
+    flags_disagreed:
+      title: "Flagged port restored by staff"
```
ZogStriP:
```suggestion
      title: "Flagged post restored by staff"
```

spec/models/reviewable_flagged_post_spec.rb
```diff
@@ -273,4 +273,29 @@ def pending_count
...
+    it "notifies the user about the flagged post deletion" do
+      reviewable = Fabricate(:reviewable_flagged_post)
+      reviewable.add_score(
+        moderator, PostActionType.types[:spam],
+        created_at: reviewable.created_at
+      )
+
+      expect do
+        reviewable.perform(moderator, :delete_and_agree)
+      end.to change(Jobs::SendSystemMessage.jobs, :size).by(1)
```
ZogStriP: We usually prefer checking that the job that was queued is the one we're actually testing (example:  d7f6d37a985a3e6233d4285f7d55c9e14905e552).

spec/models/reviewable_flagged_post_spec.rb
```diff
@@ -273,4 +273,29 @@ def pending_count
...
+  end
+
+  describe "#perform_disagree" do
+    it "notifies the user about the flagged post being restored" do
+      reviewable = Fabricate(:reviewable_flagged_post)
+      reviewable.post.hide!(PostActionType.types[:spam])
+
+      expect do
+        reviewable.perform(moderator, :disagree)
+      end.to change(Jobs::SendSystemMessage.jobs, :size).by(1)
```
ZogStriP: Same here ^^

PR #8046, opened 2019-08-27
FIX: Non-default post menu options should be allowed. Plugins need to add values to this list.

romanrizzi:


eviltrout:
We discussed this and it's better to use the plugin api `addPostMenuButton` instead.

PR #8047, opened 2019-08-28
UX: better alignment for tags in the header

hnb-ku:
This PR improves the alignment of tags in the header. 

Before:

before0001

Note how misaligned the tag community is.

After 

after0001

This is essentially the only notable visual change in the PR. 

The PR also includes a few commits that do not cause any visual changes, namely

1. removal of a number of duplicate / unused styles
2. the addition of a class to the `<h1>` tag inside `extra-info-wrapper` to avoid using a tag-name selector in CSS


SamSaffron:
Thanks so much!

PR #8048, opened 2019-08-28
FIX: Try to match advanced tutorial reset first

xfalcox:


eviltrout:
@xfalcox this could use more detail in the description. How does this fix a bug? What was the bug?

PR #8051, opened 2019-08-29
DEV: Split out multisite tests in bin/turbo_rspec

danielwaterworth:
* A new process is started that just runs the multisite tests
 * The other processes are instructed to exclude the multisite tests

### DIFF
--- a/lib/turbo_tests/runner.rb
+++ b/lib/turbo_tests/runner.rb
@@ -44,8 +44,10 @@ module TurboTests
 
       setup_tmp_dir
 
-      tests_in_groups.each_with_index do |tests, process_num|
-        start_subprocess(tests, process_num + 1)
+      start_multisite_subprocess(@files)
+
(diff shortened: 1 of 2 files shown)


danielwaterworth:
Unfortunately, RSpec causes each process to spit out a log line that says, "Run options: exclude {:type=>"multisite"}".

danielwaterworth:
The log line problem has been fixed.

lib/turbo_tests/runner.rb
```diff
@@ -83,49 +85,71 @@ def setup_tmp_dir
...
+        ["--tag", "type:multisite"],
+        tests,
+        "multisite"
+      )
+    end
+
+    def start_regular_subprocess(tests, process_id)
+      start_subprocess(
+        { 'TEST_ENV_NUMBER' => process_id.to_s },
+        ["--tag", "~type:multisite"],
```
ZogStriP: Nice :+1:

PR #8055, opened 2019-08-30
make various email + security related settings shadowed_by_global

LeoMcA:
Being able to set these by env var would be very useful for our deployment pipeline


eviltrout:
I'm going to close this in favor of:

https://github.com/discourse/discourse/pull/8061

PR #8057, opened 2019-08-30
PERF: Apply a lightweight compression to anon_cache

xfalcox:


SamSaffron:
This is done per 
https://github.com/discourse/discourse/commit/ed00f3530621d61a272b237746b8bbb1eaea7998

:confetti_ball:

PR #8058, opened 2019-08-31
FIX: correct docker image name

hawm:
`discourse/dev` is no more using and correct to the image is using


SamSaffron:
confirmed CLA, thanks @hawm  this change looks good!

PR #8059, opened 2019-09-02
Fallback handler fixes

danielwaterworth:
It's important to use different fallback handlers for different redis pairs since redis failures aren't necessarily correlated. I also refactored the FallbackHandler class and added lots of tests.

### DIFF
--- a/lib/discourse_redis.rb
+++ b/lib/discourse_redis.rb
@@ -13,6 +13,9 @@ class DiscourseRedis
     CONNECTION_TYPES = %w{normal pubsub}.each(&:freeze)
 
     def initialize(master_config, slave_config)
+      master_config = master_config.dup.freeze unless master_config.frozen?
+      slave_config = slave_config.dup.freeze unless slave_config.frozen?
+
       @master_config = master_config
       @slave_config = slave_config
     end
@@ -199,6 +202,8 @@ class DiscourseRedis
     end
 
     def handler_for(config)
+      config = config.dup.freeze unless config.frozen?
+
       @mutex.synchronize do
         @fallback_handlers[[config[:host], config[:port]]] ||= begin
           log_prefix = "FallbackHandler #{config[:host]}:#{config[:port]}"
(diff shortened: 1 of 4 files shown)


SamSaffron:
Curious, how much time do these specs add to the test suite?

xfalcox:
Just hit a fallback bug, is this en route to merging?

ZogStriP:
@davidtaylorhq can you take this one over the finish line?

SamSaffron:
I think Kris may be planning to take this when he gets back @ZogStriP

lib/discourse_redis.rb
```diff
@@ -4,139 +4,251 @@
...
+        was_reset =
+          @mutex.synchronize do
+            if @master
+              @master = false
+              trigger(:down)
+              true
+            else
+              false
+            end
+          end
```
lis2: I would extract that code to separate method something like
```
def start_reset
   @mutex.synchronize do
            if @master
              @master = false
              trigger(:down)
              true
            else
              false
            end
          end
end
```
and then you would be able to change later code to be like
```
        if start_reset
          @execution.spawn do
            loop do
```

lib/discourse_redis.rb
```diff
@@ -4,139 +4,251 @@
...
+    include Singleton
+
+    def initialize
+      @mutex = Mutex.new
+      @fallback_handlers = {}
+    end
+
+    # Used by tests
+    def reset
+      @fallback_handlers = {}
```
lis2: I think that you are using that reset method only in one place (in before block) do maybe instead of having that code here, we can use instance_variable_set in spec itself? https://apidock.com/ruby/Object/instance_variable_set

It is uglier, but on the other hand we will not put the code required by specs into the application.

lis2:
@danielwaterworth I was reading the code for half of a day and played with your solution. I run 2 Redis servers on my local, killed master, killed slave, start again Redis in different configurations and it worked very well.

I got 2 small comments in code but nothing serious, so please change them only if you feel that it is better. 

The only thing which bothers me is that I don't see the reason why the previous solution didn't work on meta, so I am not sure if your refactored version will behave better. 

I guess that we should rebase master and try to deploy that to dev and kill master Redis to see how it works in real life.

How do you feel about that?

ZogStriP:
> How do you feel about that?

Feel free to do that 👍

lis2:
I created rebased version with Daniel's commits so I will close this one - https://github.com/discourse/discourse/pull/8771

PR #8060, opened 2019-09-02
Fix typo 'veschoben' in German localization

jondo:
The correct spelling is 'verschoben', as in the adjacent lines.


ZogStriP:
Thanks but local changes have to be done on [transifex](https://www.transifex.com/discourse/discourse-org/), otherwise they would be overwritten the next time we pull them.

jondo:
According to [Transifex](https://www.transifex.com/discourse/discourse-org/translate/#de/core-server-yml?q=translation_text%3Aveschoben) the spelling is already fixed there. How often are Transifex changes merged?

PR #8063, opened 2019-09-02
FEATURE: support data-explorer outlet in group navigation

markvanlan:
This is dependent on [this data-explorer plugin PR](https://github.com/discourse/discourse-data-explorer/pull/36/files) being merged in.

This adds the `Reports` tab to the group navigation, if the data explorer plugin is installed (and if user is a part of the group, and the group has access to at least 1 report).


SamSaffron:
This looks good to me! shall I merge it?

markvanlan:
@SamSaffron that would be great 👍

SamSaffron:
Thanks Mark!

PR #8064, opened 2019-09-02
DEV: Remove FlagQuery class and old code

romanrizzi:
We can remove the `FlagQuery` class after merging https://github.com/discourse/discourse-assign/pull/47.


SamSaffron:
Merge at will, this looks good to me.

PR #8067, opened 2019-09-03
Fix typo in client.en.yml and server.en.yml.

AhmadFCheema:
"Occured" [misspelling](https://en.wiktionary.org/wiki/occured) of "occurred".


SamSaffron:
thanks!

PR #8069, opened 2019-09-03
FIX: Mobile Safari composer improvements

pmusaraj:
This includes two fixes: 

- prevents keyboard from being invoked when textarea is disabled (this happens when a category selection is required when creating a new topic)
- avoids scrolling up when switching focus from title to textarea on new topic creation

See below videos for an illustration of the second fix. 

Before: 
https://monosnap.com/file/3ohWIJB8TcqvUYqnkypuBZVzq1tG2C 

After: 
https://monosnap.com/file/HngAVObO5gy3XUTnYP7mNDITC504zQ

### DIFF
--- a/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
+++ b/app/assets/javascripts/discourse/lib/safari-hacks.js.es6
@@ -129,8 +129,21 @@ function positioningWorkaround($fixedElement) {
     if (fixedElement.style.top === "0px") {
       if (this !== document.activeElement) {
         evt.preventDefault();
+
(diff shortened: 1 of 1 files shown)


SamSaffron:
Sure this seems like a safe hack to add to safari hack mountain. I wonder if we are getting any closer to getting rid of this stuff with ios 13, I recall there are some new apis that let you get viewport in 13, this is now something we can lean on https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API with 13

coding-horror:
Oh my goooddd I can't wait for iOS 13 to land for this reason alone. (New
carplay is also very nice)

PR #8070, opened 2019-09-04
FEATURE: New post editing period for >= tl2 users

rimian:
* Adds a new setting and for tl2 editing posts (30 days same as old value)
* Sets the tl0/tl1 editing period as 1 day


eviltrout:
Is there a relevant meta discussion for this change? I'm looking for some context here.

coding-horror:
See https://meta.discourse.org/t/spam-bots-tricking-discourse-filter-by-editing/101684/20

rimian:
Sorry, I am not sure where to reference things and where to put the context just yet. I'll clear that up before my next PR :)

spec/components/guardian_spec.rb
```diff
@@ -1389,32 +1389,51 @@
...
       end
 
-      context 'post is older than post_edit_time_limit' do
-        let(:old_post) { build(:post, topic: topic, user: topic.user, created_at: 6.minutes.ago) }
+      describe 'post edit time limits' do
 
-        before do
-          SiteSetting.post_edit_time_limit = 5
-        end
+        shared_examples "edit old posts" do
```
SamSaffron: can we avoid shared_examples,  I think it just confuses stuff. makes the specs a bit harder to follow.

spec/requests/posts_controller_spec.rb
```diff
@@ -313,8 +313,28 @@
...
       end
 
-      it 'does not allow to update when edit time limit expired' do
+      it 'does not allow TL0 or TL1 to update when edit time limit expired' do
         SiteSetting.post_edit_time_limit = 5
+        SiteSetting.tl2_post_edit_time_limit = 30
+
+        post = Fabricate(:post, created_at: 10.minutes.ago, user: user)
+
+        user.trust_level = 1
```
SamSaffron: Maybe `update_columns` should do the trick it avoids a bunch of callbacks making the spec a tiny bit faster.

SamSaffron:
Overall the change looks good to me! minor changes to the specs and we should be good to merge.

spec/requests/posts_controller_spec.rb
rimian: Done!

PR #8071, opened 2019-09-04
Clarify example droplet name

jacobherrington:


coding-horror:
thanks, but I think a better version of this is just adding the word "Droplet"

PR #8073, opened 2019-09-05
FEATURE: Quick access panels in user menu

xrav3nz:
> CONTEXT: [Quick access to bookmarks and messages on user menu](https://meta.discourse.org/t/quick-access-to-bookmarks-and-messages-on-user-menu/32787/12?u=xrav3nz)

This feature adds quick access panels for bookmarks and personal
message. It allows uses to browse recent items directly in the user
menu, without being redirected to the full page.

4474a28 shows how the new `QuickAccessPanel` is extracted from the old `UserNotifications`, and I hope this will help with the review a bit.

Thanks and open to suggestions!

(I will send another PR to `discourse-assign` by the end of this week to add its quick access panel.)


app/assets/javascripts/discourse/widgets/quick-access-bookmarks.js.es6
```diff
@@ -0,0 +1,59 @@
...
+      cache: "false",
+      data: {
+        username: this.currentUser.username,
+        filter: UserAction.TYPES.bookmarks,
+        limit: this.estimateItemLimit(),
+        no_results_help_key: "user_activity.no_bookmarks"
+      }
+    }).then(({ user_actions, no_results_help }) => {
+      emptyStatePlaceholderItemText = no_results_help;
+      return (staleItems = user_actions.slice(0, this.estimateItemLimit()));
```
xrav3nz: The `user_actions` endpoint does not support a `limit` parameter, so this truncates the results to ensure they will fit on the screen. I am not sure if it is preferred to add a new `limit` param to the endpoint?

app/assets/javascripts/discourse/widgets/quick-access-messages.js.es6
```diff
@@ -0,0 +1,59 @@
...
+
+  findNewItems() {
+    return this.store
+      .findFiltered("topicList", {
+        filter: `topics/private-messages/${this.currentUser.username_lower}`
+      })
+      .then(({ topic_list }) => {
+        return (staleItems = topic_list.topics
+          .map(toItem)
+          .slice(0, this.estimateItemLimit()));
```
xrav3nz: Same here.

app/assets/javascripts/discourse/widgets/quick-access-bookmarks.js.es6
```diff
@@ -0,0 +1,59 @@
...
+import DiscourseURL from "discourse/lib/url";
+import QuickAccessPanel from "discourse/widgets/quick-access-panel";
+import UserAction from "discourse/models/user-action";
+import { ajax } from "discourse/lib/ajax";
+import { createWidgetFrom } from "discourse/widgets/widget";
+import { postUrl } from "discourse/lib/utilities";
+
+const ICON = "bookmark";
+
+let staleItems = [];
```
eviltrout: My feeling is this data would be better stored in the state of a widget, since its value is populated based on the result of the ajax call. Putting the variable here in the closure makes it a bit leaky in tests.  Same for `emptyPlaceholderItemText` below.

app/assets/javascripts/discourse/widgets/quick-access-messages.js.es6
```diff
@@ -0,0 +1,59 @@
...
+let staleItems = [];
+
+function toItem(message) {
+  const lastReadPostNumber = message.last_read_post_number || 0;
+  const nextUnreadPostNumber = Math.min(
+    lastReadPostNumber + 1,
+    message.highest_post_number
+  );
+
+  return Ember.Object.create({
```
eviltrout: Does this need to be an `Ember.Object`? They are heavier than regular JS objects, and if you don't need to observe or use computed properties on them I recommend the plain JS version.

app/assets/javascripts/discourse/widgets/quick-access-notifications.js.es6
```diff
@@ -0,0 +1,61 @@
...
+
+  showAll() {
+    DiscourseURL.routeTo(`${this.attrs.path}/notifications`);
+  },
+
+  hasUnread() {
+    return this.state.items.filterBy("read", false).length > 0;
+  },
+
+  findStaleItemsInStore_() {
```
eviltrout: To denote a private method we prefix with the underscore, like `_findStatleItemsInStore()`

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
```diff
@@ -0,0 +1,146 @@
...
+      return "";
+    }
+  },
+
+  defaultState() {
+    return { items: [], loading: false, loaded: false };
+  },
+
+  markRead() {
+    this.markReadRequest().then(() => {
```
eviltrout: It's a good habit to always return the promise in case something eventually wants to wait on it.

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
```diff
@@ -0,0 +1,146 @@
...
+      this.refreshNotifications(this.state);
+    });
+  },
+
+  estimateItemLimit() {
+    // Estimate (poorly) the amount of notifications to return.
```
eviltrout: I appreciate the honesty :rofl:

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
```diff
@@ -0,0 +1,146 @@
...
+    const staleItems = this.findStaleItems();
+
+    if (staleItems.length > 0) {
+      state.items = staleItems;
+    } else {
+      state.loading = true;
+    }
+
+    this.findNewItems()
+      .then(items => {
```
eviltrout: Minor but I'd make these `then`/`catch` one liners.

app/assets/javascripts/discourse/widgets/user-menu.js.es6
```diff
@@ -81,42 +105,39 @@ createWidget("user-menu-links", {
...
     return h("ul.menu-links-row", [
       links.map(l => h("li.user", this.linkHtml(l))),
       h("li.glyphs", glyphs.map(l => this.glyphHtml(l)))
     ]);
+  },
+
+  markAsActive(definition) {
+    // Clicking on an active quick access tab icon should redirect the user to
+    // the full page.
+    delete definition.action;
```
eviltrout: This is probably not a perf issue here, but in general you don't want to change the shape of the JS objects. https://mathiasbynens.be/notes/shapes-ics

I would recommend setting the values to `null` instead of `delete`.

test/javascripts/helpers/create-store.js.es6
```diff
@@ -16,6 +17,11 @@ export default function() {
           }
           return this._restAdapter;
         }
+        if (type === "adapter:topicList") {
```
eviltrout: Why is this necessary? Shouldn't the topic list adapter automatically be returned by name?

xrav3nz: In test environment, `store` seems to be stubbed by this `create-store.js.es6` helper. It expects an exact match when `lookup`, and would otherwise [fallback to `adapter:rest`](https://github.com/discourse/discourse/blob/f47ae17d6dc92addd473e02470c78f03fd3eb3ed/app/assets/javascripts/discourse/models/store.js.es6#L258).

app/assets/javascripts/discourse/widgets/user-menu.js.es6
xrav3nz: That was a very informative read, TIL! Thanks!

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
xrav3nz: This is actually a verbatim copy from the previous `NotificationItem` widget, [by ... some very evil fish](https://github.com/discourse/discourse/commit/514c3976f05785f487d275af1e96c43533e7a478#diff-15a87278f0e3eadb2066736f78b119feR20). 😂

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
```diff
@@ -0,0 +1,146 @@
...
+      : [this.emptyStatePlaceholderItem()];
+
+    if (this.hasMore()) {
+      items.push(
+        h(
+          "li.read.last.heading.show-all",
+          this.attach("button", {
+            title: "view_all",
+            icon: "chevron-down",
+            action: "showAll",
```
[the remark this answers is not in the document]
xrav3nz: Good point, thanks! This was carried over from the original `UserNotification` widget. I have changed this to a `link` with `href` so that right-click/cmd-click will work too!

app/assets/javascripts/discourse/widgets/quick-access-bookmarks.js.es6
xrav3nz: Totally! I have moved `emptyStatePlaceholderItemText` to the widget state instead.

However, for stale items, I am under the impression that the widget instance is destroyed when "unmounted"? So,

1. Open quick access panel for notifications, and `staleItems` is saved to state.
2. Switch to quick access panel for bookmarks. The notifications widget along with its state (i.e. `staleItems`) is destroyed.
3. Switch back to quick access for notifications. A new notifications widget is initialized --`staleItems` is empty and the loading circle is shown while fresh items are being fetched.

xrav3nz: After having thought about this a bit more, if we can store `items` as some kind of static variable on the widget, we can drop `staleItems` (and `findStaleItems`) altogether. Because `items === staleItems` when switching back to a quick access panel.

What do you think of:

1. A single object in the base abstract widget's closure. 

    ```js
    const staleItems = {};

    createWidget("quick-access-panel", {
      html() {
        const items = staleItems[this.key] || [];
      }
    });
    ```

2. Emulate static variable with a property on a widget method, like:

    ```js
    createWidget("quick-acces-panel", {
      staleItems() {},

      html() {
        const items = this.staleItems[this.key] || [];
      }
    });
    ```

eviltrout: Oh I see, `staleItems` is meant to remember previously seen things even when the widget is destroyed. In that case you should store your data in `Session`. It is a singleton meant for transient data and is cleared between every test run.

app/assets/javascripts/discourse/widgets/quick-access-bookmarks.js.es6
eviltrout: It  looks like it would be quite simple to add a `limit` parameter to the route, so I would suggest doing that.

app/assets/javascripts/discourse/widgets/quick-access-bookmarks.js.es6
xrav3nz: Ah nice, TIL about the `Session` singleton. Thanks!

app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
```diff
@@ -0,0 +1,143 @@
...
+
+    return [h("ul", items)];
+  },
+
+  getItems() {
+    return Discourse.Session.currentProp(`${this.key}-items`) || [];
+  },
+
+  setItems(newItems) {
+    Discourse.Session.currentProp(`${this.key}-items`, newItems);
```
eviltrout: It would be better to import the `Session` module than use the `Discourse.Session` here.

eviltrout:
Thanks @xrav3nz ! It's been merged.

xrav3nz:
Thanks again for the reviews!

PR #8074, opened 2019-09-06
UX: Use medium format for displaying time in post notices.

udan11:


eviltrout:
This looks good but is there a context/reason for this change?

coding-horror:
The abbreviations weren't necessary for text we use rarely, which is never
really repeated on the screen either (thus leading to more redundancy).

PR #8075, opened 2019-09-06
FIX: Topics with muted tag didn't show up when filtering by category and tag

gschlager:
It also removes the redundant `filter` parameter. Previously URLs looked like this:

```
http://example.com/tags/c/some-category/muted-tag/l/latest.json?filter=tags/c/some-category/muted-tag/l/latest
```

But it looks like the `filter` parameter was only used to find out if topics with a muted tag should be removed or not. But the same thing can be accomplished by using the first tag ID. The following URL looks a lot cleaner.

```
http://example.com/tags/c/some-category/muted-tag/l/latest.json
```

### DIFF
--- a/spec/requests/tags_controller_spec.rb
+++ b/spec/requests/tags_controller_spec.rb
@@ -207,6 +207,11 @@ describe TagsController do
     end
 
     context 'tagging enabled' do
+      def parse_topic_ids
+        JSON.parse(response.body)["topic_list"]["topics"]
+          .map { |topic| topic["id"] }
+      end
+
       it "can filter by tag" do
         get "/tags/#{tag.name}/l/latest.json"
         expect(response.status).to eq(200)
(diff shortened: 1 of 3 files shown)


spec/requests/tags_controller_spec.rb
```diff
@@ -288,17 +287,53 @@
...
+            get "/tags/#{tag.name}/l/latest.json"
+            expect(response.status).to eq(200)
+
+            topic_ids = parse_topic_ids
+            expect(topic_ids).to include(single_tag_topic.id)
+          end
+
+          it "includes topics when filtered by category and muted tag" do
+            category = Fabricate(:category)
+            single_tag_topic.update!(category: category)
```
gschlager: I have no idea why this doesn't work with the prefabricated `category` 🤷‍♂ 
https://github.com/discourse/discourse/blob/820cdc23a3fb472a50bbfd043cd28af5fe515618/spec/requests/tags_controller_spec.rb#L194

nlalonde:
I have no idea. It doesn't even look like a valid route...

PR #8078, opened 2019-09-06
UX/PERF: Updates to the read indicator.

romanrizzi:
### New behavior:

 - When reading someone else’s post, it should show my avatar in there and include me in the count.

 - When reading my post, it should not show my avatar and not include me in the count (because it’s implied that I read that post.)

### Perf changes:

Do not make an HTTP call just to update the count. The new count is already available through message bus.

### DIFF
--- a/app/assets/javascripts/discourse/models/post-stream.js.es6
+++ b/app/assets/javascripts/discourse/models/post-stream.js.es6
@@ -716,6 +716,19 @@ export default RestModel.extend({
     return resolved;
   },
 
+  triggerReadPost(postId, readersCount) {
+    const resolved = Ember.RSVP.Promise.resolve();
+    resolved.then(() => {
+      const post = this.findLoadedPost(postId);
+      if (post && readersCount > post.readers_count) {
+        post.set("readers_count", readersCount);
+        this.storePost(post);
+      }
+    });
+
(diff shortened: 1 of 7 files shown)


coding-horror:
if this looks good @eviltrout  can we merge it?

app/assets/javascripts/discourse/models/post-stream.js.es6
```diff
@@ -716,6 +716,19 @@ export default RestModel.extend({
     return resolved;
   },
 
+  triggerReadPost(postId, readersCount) {
+    const resolved = Ember.RSVP.Promise.resolve();
```
eviltrout: This is a little confusing. Why does it create a resolved promise only to call `then` on it and not do any async work?

If the method doesn't need to do anything async but needs to return a promise, why call `then` at all?

romanrizzi: I did this to be consistent with the current API (returning a promise). You're right about executing the code inside a `then`, I'll execute it outside the promise.

SamSaffron:
Let's try this out, we can always follow it up, but I prefer to merge now to avoid the large amount of HTTP calls when we are all actively reading a message

PR #8079, opened 2019-09-06
FIX: Use Discourse.getURL for /clicks/track so clicks can be tracked

qrush:
Our install of Discourse @chatterbugapp was seeing requests for click tracking go through to our main app (we use a relative URL root) - I dug in a bit and it seems like this file was missing a call to `Discourse.getURL`. 

🙏 🙇

### DIFF
--- a/app/assets/javascripts/discourse/lib/ajax.js.es6
+++ b/app/assets/javascripts/discourse/lib/ajax.js.es6
@@ -159,7 +159,7 @@ export function ajax() {
   if (
     args.type &&
     args.type.toUpperCase() !== "GET" &&
-    url !== "/clicks/track" &&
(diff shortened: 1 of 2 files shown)


SamSaffron:
Oh, absolutely, thanks heaps for the PR @qrush

eviltrout:
nice catch :+1:

qrush:
@eviltrout @SamSaffron thanks! Is this included in 2.4.0.beta4 from today? Wondering how I can roll this out easily on our install.

eviltrout:
@qrush it will be included in our next beta if you are following that channel. if you are on `tests-passed` you should get it with your next update.

Betas are usually cut once a week or so.

PR #8081, opened 2019-09-08
FIX: Migrate post_edit_time_limit setting

rimian:
When tl0 setting is larger than tl2, save new tl2 setting and leave tl0 as default


rimian:
Original context: https://meta.discourse.org/t/spam-bots-tricking-discourse-filter-by-editing/101684

rimian:
Do it with a single insert.

PR #8082, opened 2019-09-09
FIX: Migrate post_edit_time_limit to tl2_post_edit_time_limit

rimian:
Insert new value into `tl2_post_edit_time_limit` if:
* tl2 setting does not exist
* tl0 setting is larger than default tl2
Also:
* Uses current timestamp
* No rollback

This doesn't set the new tl0 value.

### DIFF
--- a/db/migrate/20190908234054_migrate_post_edit_time_limit.rb
+++ b/db/migrate/20190908234054_migrate_post_edit_time_limit.rb
@@ -19,8 +19,6 @@ class MigratePostEditTimeLimit < ActiveRecord::Migration[5.2]
       FROM site_settings
       WHERE
         name = 'post_edit_time_limit'
-      AND
-        CAST(value AS INTEGER) > 43200
     ON CONFLICT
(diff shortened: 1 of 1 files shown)


db/migrate/20190908234054_migrate_post_edit_time_limit.rb
```diff
@@ -0,0 +1,28 @@
...
+        'tl2_post_edit_time_limit',
+        value,
+        data_type,
+        CURRENT_TIMESTAMP,
+        CURRENT_TIMESTAMP
+      FROM site_settings
+      WHERE
+        name = 'post_edit_time_limit'
+      AND
+        CAST(value AS INTEGER) > 43200
```
SamSaffron: I don't think we need this cast, we can just insert unconditionally, that way post upgrade everything will work exactly as it used to if you did any overriding.

SamSaffron:
This looks great, thank Rim

PR #8085, opened 2019-09-11
DEV: Add notification_types table

danielwaterworth:
This table stores the ids of the notfication types so that notification
types can be added by plugins.


SamSaffron:
I am very mixed on this, the current approach of requiring plugins to submit a PR to core solves this. I don't really see anything hugely problematic with this approach. 

It gets confusing if id 22 on one db means something and another something else. Current approach gives us parity across instances.

PR #8087, opened 2019-09-11
FIX: Cast all numerical values in reports

CvX:
The backend can return some numerical values in report as strings.
That results in unexpected order of values when sorting report tables:

invalid value order


app/assets/javascripts/admin/models/report.js.es6
```diff
@@ -374,14 +374,14 @@ const Report = Discourse.Model.extend({
 
   _secondsLabel(value) {
     return {
-      value,
+      value: typeof value === "number" ? value : Number(value),
```
ZogStriP: We might want to extract this into an helper method so that we don't have to repeat it 4 times.

jjaffeux: Also we have this:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/lib/utilities.js.es6#L602

Not sure if we should use this, or check on type.

CvX:
@eviltrout There are performance differences, and it looks like `typeof` + `parseFloat` is the fastest path: https://jsperf.com/number-vs-typeof-vs-parsefloat#results

I can extract that to a helper as @ZogStriP suggested, and then update the mentioned `isNumeric`, e.g.

```js
export function toNumber(input) {
  return typeof input === "number" ? input : parseFloat(input);
}

export function isNumeric(input) {
  return !isNaN(toNumber(input)) && isFinite(input);
}
```

Though `isNumeric` change would be unrelated to this fix and just a micro-optimization.

PR #8094, opened 2019-09-12
FEATURE: allow UploadRecovery to be run on a single post

Supermathie:
the diff makes it look like I made a ton of changes below the method signature, in reality all that happened below

```
def recover_post(post)
```

was a dedent.


SamSaffron:
Cool I support splitting up this code.

PR #8096, opened 2019-09-13
FIX: Ensure page is reloaded when logging out

davidtaylorhq:
To demonstrate the issue:
- Visit https://meta.discourse.org/#somethingHere while logged in
- Click "log out"
- You will be logged out, but the page will not be reloaded

Setting `window.location.pathname = "/"` will not reload the page if there is a hash present. Using `window.location = "/"` gives us the desired behavior.


ZogStriP:
Should we do that everywhere else we're also setting `window.location.pathname` instead of `window.location`?

davidtaylorhq:
> Should we do that everywhere else

Yes, I think so. I've updated the 6 other locations where we were doing the same thing

PR #8099, opened 2019-09-15
[WIP] FEATURE: Webauthn authenticator management with 2FA login (Security Keys)

mjrbrennan:
Users can now add Security Keys as second factor authentication method via the Webauthn API. RFC on Meta: https://meta.discourse.org/t/webauthn-support/126454/10.

Users can register security keys from the Second Factor Authentication preferences screen. They can give each key a name and edit the name & disable a key if required. When they register the key they are prompted to insert and use their physical key:

image
image
image

When the user logs in and they have active security keys, they are shown a prompt to authenticate using a security key. When they click "Authenticate With Security Key" they are prompted to use their physical key. In the case of having multiple keys registered, we just tell the webauthn API all of the allowed credential IDs for the user, and the first key matching any of them is used when logging in.

image
image

If a user does not have their security key with them they can choose to fall back to TOTP based 2FA.

image

The missing case I've found is "What happens when a user has only security keys registered and loses all of them?". This seems to happen as well when I only set up one TOTP code and no backup codes, I cannot proceed in login if I have "lost" my code. What happens normally here?:

image

Users can also use their security key when resetting their password, and when logging in via email:

image

### Implementation Assumptions

* I have used `Discourse.current_hostname` for the Relaying Party ID (https://w3c.github.io/webauthn/#rp-id) which should be the "effective domain" of the relaying party. In localhost this is just `localhost` and I assume in production sites it would be `meta.discourse.org`, `forums.envato.com` etc.
* I've left out certain validation steps that are either optional or unnecessary for an initial implementation, e.g. `clientExtensionResults`, fully verifying the attestation statement using trust anchors (using attestation format of `none` for now), checking `tokenBinding.status`. This does not reduce the security of the implementation.
* I've just used a single supported public key algorithm for now -- `-7` which is ECDSA w/ SHA-256.  https://www.iana.org/assignments/cose/cose.xhtml#algorithms
* I have only tested the implementation with Yubikeys; I do not have an Android phone / macOS / any other physical hardware authentication methods

### Dependencies Added

* cbor (gem) - webauthn credential attestation data is CBOR encoded, and we need to decode it to inspect the attestation
* cose (gem) - the public key for the webauthn credential is encoded in the COSE key format

### Schema Changes

* Added a `UserSecurityKey` model and database table. The information stored for security keys differs enough from the `UserSecondFactor` table that they need to be stored on their own. Also, at some point security keys will be used for first + multi factor auth.
* Added `secure_identifier` column to `User` which is only written when it is needed. This was done to satisfy `webauthn.credentials.create()` by providing a random string to identify a Discourse user. https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity. Using the user ID integer worked OK on an older Yubikey, but did not work at all on the YubiKey 5 NFC, so I think it's safer to go with a random string for this.

### WIP TODO

* The code currently is feature complete, however controller tests for the session controller + user controller as well as a QUnit integration spec is still missing.


SamSaffron:
> base64js.min.js 

IE9 is not a release target for us, can we just use the built in implementation? https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa

Gemfile
```diff
@@ -114,6 +114,8 @@ gem 'execjs', require: false
 gem 'mini_racer'
 gem 'highline', '~> 1.7.0', require: false
 gem 'rack-protection' # security
+gem 'cbor'
```
SamSaffron: can we `require: false` the new gems?

config/locales/client.en.yml
```diff
@@ -1444,6 +1454,11 @@ en:
       second_factor_backup_title: "Two Factor Backup"
       second_factor_backup_description: "Please enter one of your backup codes:"
       second_factor: "<a href>Log in using Authenticator app</a>"
+      security_key_description: "When you have your physical security key prepared press the Authenticate with Security Key button below."
+      security_key_alternative: "<a href>Can't find your security key or want to use another method?</a>"
```
eviltrout: This is a bit odd - why is there an empty link in the translation? It would be better to keep such markup in the handlebars code.

db/migrate/20190904104533_create_user_security_keys.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class CreateUserSecurityKeys < ActiveRecord::Migration[5.2]
+  def up
+    create_table :user_security_keys do |t|
+      t.references :user, null: false, index: true, foreign_key: true
+      t.string :credential_id, null: false
+      t.string :public_key, null: false, index: true
+      t.integer :factor_type, null: false, default: 0, index: true
+      t.boolean :enabled, null: false, default: true
```
eviltrout: There is a scope for `enabled` above - thoughts on adding that as part of an index?

spec/fabricators/user_second_factor_fabricator.rb
```diff
@@ -7,6 +7,13 @@
   method UserSecondFactor.methods[:totp]
 end
 
+Fabricator(:user_second_factor_webauthn, from: :user_second_factor) do
+  user
+  data 'TODO (martin) - fill this in with a good approximation'
```
eviltrout: No time like the present :)

eviltrout:
P.S. Overall this is looking excellent! A ton of good work in here.

spec/fabricators/user_second_factor_fabricator.rb
mjrbrennan: Argh thought I got rid of this one! Removing, as this fabricator is no longer relevant

mjrbrennan:
Thank you both for your feedback so far, I'm making changes now!

config/locales/client.en.yml
mjrbrennan: The above `second_factor` translation has it as well, I thought that it must have been a convention followed. I will change it and the other translation to not contain an empty link

app/assets/javascripts/discourse/controllers/login.js.es6
```diff
@@ -126,15 +135,22 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
 
-              Ember.run.schedule("afterRender", () =>
-                document
-                  .getElementById("second-factor")
-                  .querySelector("input")
-                  .focus()
-              );
+              // only need to focus the 2FA input for TOTP
+              if (!this.showSecurityKey) {
+                Ember.run.schedule("afterRender", () =>
```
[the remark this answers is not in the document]
mjrbrennan: This schedule code was already there. I'm not overly familiar with Ember but what you're saying sounds right + correlates with the docs, will change

app/assets/javascripts/discourse/models/user.js.es6
```diff
@@ -375,6 +375,25 @@ const User = RestModel.extend({
...
+    });
+  },
+
+  registerSecurityKey(credential) {
+    let formData = new FormData();
+    for (const [key, value] of Object.entries(credential)) {
+      formData.set(key, value);
+    };
+    return ajax('/u/register_second_factor_security_key.json', {
+      data: formData,
```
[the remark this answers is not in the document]
mjrbrennan: Have no idea what I was thinking here -- have changed.

db/migrate/20190904104533_create_user_security_keys.rb
mjrbrennan: You're right. Will add this and some model scopes to bring more into line with `UserSecondFactor`

app/assets/javascripts/discourse/controllers/login.js.es6
```diff
@@ -126,15 +135,22 @@ export default Ember.Controller.extend(ModalFunctionality, {
...
+                  document
```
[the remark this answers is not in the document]
mjrbrennan: I didn't realise there were more linters than the ones that ran in the pre-commit hook. I've found lefthook and run prettier to fix the issues in the files I added :)

app/assets/javascripts/discourse/components/security-key-form.js.es6
```diff
@@ -0,0 +1,14 @@
+import { SECOND_FACTOR_METHODS } from "discourse/models/user";
+
+export default Ember.Component.extend({
+  actions: {
+    authenticateSecurityKey() {
+      this.attrs['action']();
```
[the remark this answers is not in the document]
mjrbrennan: Another Ember faux pas! I've corrected this now, thanks Robin

mjrbrennan:
@SamSaffron @eviltrout I've now made all the  changes and am waiting on another Travis build :+1:

db/migrate/20190904104533_create_user_security_keys.rb
```diff
@@ -0,0 +1,30 @@
...
+    add_index :user_security_keys, :last_used
+  end
+
+  def down
+    if column_exists?(:user_second_factors, :user_security_key_id)
+      remove_column(:user_second_factors, :user_security_key_id)
+    end
+
+    if table_exists?(:user_second_factors)
+      drop_table(:user_second_factors)
```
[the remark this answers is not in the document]
mjrbrennan: No I did not -- that whole `down` method is nonsensical, I must have went one way then changed direction with the `up` then did not change the `down` to go with it! Fixed.

mjrbrennan:
Thanks for letting me know about the other routes @featheredtoast; I'll address them one by one, starting with login over email. Will try and get that part up tonight!

mjrbrennan:
@featheredtoast I've now added security key functionality for the password reset, and login over email links, only one left now is admin login. :smile:

mjrbrennan:
@featheredtoast I've now added the admin-login security key functionality, and also changed that route to use the `no_ember` layout because I needed to be able to access ES6 modules, hopefully I've done that the right way, the admin_login.html.erb view was a bit strange compared to the rest of the app!

featheredtoast:
Still reviewing the new routes, but just FYI I'm no longer able to register keys, it looks like your buffer to base64 function is now defined in webauthn.js.es6 rather than utilities.

mjrbrennan:
@featheredtoast sorry about that Jeff, I must have broken it when I did some changes for admin login the other day, I really need to add some integration specs to catch that! It has been fixed now, and as you said I hid the other submit/finish login buttons in all the different paths when a security key is being used

image
image

featheredtoast:
Very cool! I'm really liking the way it's looking now.

Last little use-case niggle I can think of is we might want to have a message if an already-registered key is attempting to be registered again for the same user as allowing duplicate entries could be both confusing and a little dangerous (You think you're removing that key, but if it's registered twice... 💣).

mjrbrennan:
@featheredtoast that sounds like what `excludeCredentials` is used for https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions/excludeCredentials. From the wording it sounds like an exception is thrown if the Yubikey/hardware contains any of the provided credential IDs provided, so we would pull all of the active keys for the user from the database when they are registering a new one. I'll try it out tonight and let you know what I find 👍

featheredtoast:
That would be fantastic! I'm down with merging once that's in.

mjrbrennan:
OK @featheredtoast all done! This is what a user sees in Chrome if they attempt to use the same authenticator to register multiple security keys:

image

image

featheredtoast:
@mjrbrennan I've merged this, as it's in a very good state - thank you so much for your excellent work on this!

We can definitely improve on other nice-to-haves (security key challenges as an alternative to passwords to access 'security' auth sections, and security keys as first-factor/passwordless logins) but I'm very excited to get feedback of the base feature sets here. Thanks again!

PR #8102, opened 2019-09-17
Make INSTALL-cloud agree with discourse-setup

pfaffman:
`discourse-setup` will install Docker. Several people lately have had trouble with a `|` getting turned into a `>` when trying to `curl` the site into `sh`.

Also, the recommended `git clone` command will create `/var/discourse`, so don't tell people to make that directory, as that too can be confusing.


coding-horror:
I don't see a need for this change?

coding-horror:
oh wait LOL I see that `discourse-setup` was changed. my bad.

PR #8103, opened 2019-09-18
FIX: Split migration into two steps in developer guide

lis2:
To temporary mitigate the problem, we need to split rake `db:create` and rake `db:migrate` into two commands. 

It was discussed here: https://meta.discourse.org/t/hello-from-gitpod-installing-on-google-cloud-automated-dev-setup/128783/


eviltrout:
Thanks, seems good!

PR #8105, opened 2019-09-18
FIX: Remove Versions from Active Record warm up

lis2:
This is a fix for the problem mentioned here: https://meta.discourse.org/t/hello-from-gitpod-installing-on-google-cloud-automated-dev-setup/128783/18


eviltrout:
Thanks, looks good.

PR #8107, opened 2019-09-20
FIX: Correctly escape category description text

CvX:
This bug has been introduced in db14e10943aeb87f3a2e06f02ca788986f039077.

### DIFF
--- a/app/assets/javascripts/discourse/helpers/category-link.js.es6
+++ b/app/assets/javascripts/discourse/helpers/category-link.js.es6
@@ -121,7 +121,7 @@ function defaultCategoryLinkRenderer(category, opts) {
     'data-drop-close="true" class="' +
     classNames +
     '"' +
-    (description ? 'title="' + escapeExpression(description) + '" ' : "") +
+    (description ? 'title="' + description + '" ' : "") +
     ">";
 
   let categoryName = escapeExpression(get(category, "name"));
--- a/app/assets/javascripts/discourse/helpers/category-link.js.es6
+++ b/app/assets/javascripts/discourse/helpers/category-link.js.es6
@@ -75,7 +75,7 @@ export function categoryLinkHTML(category, options) {
 registerUnbound("category-link", categoryLinkHTML);
 
 function defaultCategoryLinkRenderer(category, opts) {
-  let description = get(category, "description_text");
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/discourse/helpers/category-link.js.es6
```diff
@@ -121,7 +121,7 @@ function defaultCategoryLinkRenderer(category, opts) {
     'data-drop-close="true" class="' +
     classNames +
     '"' +
-    (description ? 'title="' + escapeExpression(description) + '" ' : "") +
+    (description ? 'title="' + description + '" ' : "") +
```
ZogStriP: Are we 100% sure that `description` is safe?

CvX: The `description` here is returned by `Category#description_text` that [has been updated in this PR](https://github.com/discourse/discourse/pull/8107/files#diff-af69b723432608c351fb1dd5db9e8348R265) to escape unsafe characters. The return value is cached there, but that's a thread-local cache, so there's no need to worry about stale/unsafe data.

ZogStriP: I see, kinda confusing... Would have been better to call that variable "`descriptionText`"

CvX: Agree. I'm changing that.

app/helpers/application_helper.rb
```diff
@@ -437,17 +437,14 @@ def replace_plugin_html(name)
 
   def theme_lookup(name)
     Theme.lookup_field(theme_ids, mobile_view? ? :mobile : :desktop, name)
-      &.html_safe
```
eviltrout: This comment removes the automat `html_safe` calls regardless of what is being looked up, but the commit seems to focus on description only. Does this mean it's possible other fields that were previously escaped would no longer be?

discoursereviewbot: [Régis Hanol posted](https://review.discourse.org/t/fix-correctly-escape-category-description-text-pr-8107/5800/8):

It's actually a duplicate. We're already calling `html_safe` in `Theme.lookup_field`.

CvX:
ping @davidtaylorhq 🙂

PR #8108, opened 2019-09-21
List dots as a valid character in usernames

xfix:
They are already allowed, just an error doesn't list them as allowed.


SamSaffron:
Yes this looks fine to me. Thanks.

PR #8110, opened 2019-09-22
Add topic and category context to post webhook payload

ryan-boder:
Adds additional fields topic_archetype and category_slug to the post
webhook so that handlers have some context about the post event without
having to call back to the API.

Discussed [here](https://meta.discourse.org/t/webhooks-how-best-to-differentiate-a-pm-from-a-public-post/76363/13).

### DIFF
--- a/app/serializers/web_hook_post_serializer.rb
+++ b/app/serializers/web_hook_post_serializer.rb
@@ -2,7 +2,9 @@
 
 class WebHookPostSerializer < PostSerializer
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
This looks good, we have a accrued a bit of tech debt around the structure of the serializer though. kind of hacky how it omits certain attributes.

Kiaira22 (commented):
[]()

PR #8113, opened 2019-09-30
FIX: attempts to improve screen-tracker logic

jjaffeux:
This commit includes multiple fixes to make screen-tracker more reliable, especially in the context of long running sessions.

- do not rely on an observed Discourse.hasFocus
- defines a new appEvent when focus is changed
- stop tracking when tab loses focus


jjaffeux:
Un-optimized code, mostly for experiment, will improve it if it proves useful.

PR #8116, opened 2019-10-01
FIX: gotFocus is not called in discourse-topic

jjaffeux:


SamSaffron:
I think we no longer need this due to other awesome fix.

PR #8118, opened 2019-10-01
Add group name classes to tag page so groups can be hidden with CSS

awesomerobot:


SamSaffron:
will this work with unicode usernames? 

@gschlager  can you review? Maybe there is a general helper we can use here for this type of "make username name CSS friendly thing?

app/assets/javascripts/discourse/components/tag-list.js.es6
```diff
@@ -19,5 +19,16 @@ export default Ember.Component.extend({
   @computed("category.fullSlug")
   categoryClass(slug) {
     return slug && `tag-list-${slug}`;
+  },
+
+  @computed("tagGroupName")
+  tagGroupNameClass(groupName) {
+    if (groupName) {
+      var groupName = groupName
```
jjaffeux:
```suggestion
      const groupName = groupName
```

jjaffeux: Here groupName being in the function params is considered already declared as mutable and doesn't needs a new declaration

PR #8121, opened 2019-10-01
DEV: prevents attempt to redefine THEME_TARGETS constant

jjaffeux:


SamSaffron:
Yeah this is getting annoying.

PR #8123, opened 2019-10-01
DEV: adds discourse:focus-changed app event

jjaffeux:


jjaffeux:
@eviltrout note I couldn't find where is this `notify: false` used, do you know ?

SamSaffron:
This is a walk down memory lane: 

https://github.com/discourse/discourse/blame/b9ccf4d09c62fd87a3be0bd42ecd3ed036dd6129/app/assets/javascripts/discourse.js

We used to put `(*) topic title` for certain cases, something that we totally stopped doing. 

Merging this in!

PR #8127, opened 2019-10-02
FEATURE: topic is validated for blocked words

lis2:
Currently, the topic is only validated for censored words and should be validated for blocked words as well.

Blocked words validation is now used by both `Post` and `Topic`. To avoid code duplication, I extracted blocked words validation code into separate Validator, and use it in both places.

The only downside is that even if the topic contains blocked words validation message is saying "Your post contains a word that's not allowed: tomato" but I think this is descriptive enough.

Screenshot from 2019-10-02 09-51-33


SamSaffron:
This looks good to me!

PR #8130, opened 2019-10-02
DEV: removes dead code

jjaffeux:
Couldn't find where this code is being used. So far it seems that this has been superseed by:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/widgets/post-menu.js.es6#L399

Which I also quickly refactor in this PR

### DIFF
--- a/app/assets/javascripts/discourse/widgets/post-menu.js.es6
+++ b/app/assets/javascripts/discourse/widgets/post-menu.js.es6
@@ -396,8 +396,7 @@ export default createWidget("post-menu", {
   },
 
   menuItems() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -396,8 +396,7 @@ export default createWidget("post-menu", {
   },
 
   menuItems() {
-    let result = this.siteSettings.post_menu.split("|");
-    return result;
+    return this.siteSettings.post_menu.split("|").filter(Boolean);
```
eviltrout: This is a nice trick! :+1:

PR #8134, opened 2019-10-02
FIX: solution for pending migrations for bin/turbo_rspec

lis2:
Currently, if you try to run `./bin/turbo_rspec` you will got that error `There are pending migrations, run rake parallel:migrate`

Reason for that is that command is running in `development` mode which includes plugins migration files in ActiveRecord::Migrator.migrations_paths:
```
["db/migrate",
 "/home/lis2/projects/discourse/plugins/discourse-details/db/migrate",
 "/home/lis2/projects/discourse/plugins/discourse-details/db/post_migrate",
 "/home/lis2/projects/discourse/plugins/discourse-local-dates/db/migrate",
 "/home/lis2/projects/discourse/plugins/discourse-local-dates/db/post_migrate",
...
]
```

A workaround solution would be to run the command with the TEST environment like `RAILS_ENV=test ./bin/turbo_rspec`

I want to propose in this PR to override migration_paths to check only Discourse core migrations.

### DIFF
--- a/lib/turbo_tests/runner.rb
+++ b/lib/turbo_tests/runner.rb
@@ -80,6 +80,8 @@ module TurboTests
           .configurations["test"]
           .merge("database" => "discourse_test_1")
 
+      ActiveRecord::Migrator.migrations_paths = ['db/migrate', 'db/post_migrate']
+
       conn = ActiveRecord::Base.establish_connection(config).connection
       begin
         ActiveRecord::Migration.check_pending!(conn)


SamSaffron:
Maybe this is a bit too harsh, maybe simply re-exec self if you launch turbo_rspec and ENV['RAILS_ENV'] is not `test` ? That way we can make turbo rspec work with plugins eventually if we want. 

@danielwaterworth  ?

danielwaterworth:
This definitely used to work - I guess it was broken by rails 6. It's entirely reasonable for TurboTests::Runner to set RAILS_ENV since the rails environment isn't loaded in the turbo_tests root process. In which case, we shouldn't need to specify that we want the test configuration on line 80.

SamSaffron:
We need to make some sort of call here, @danielwaterworth  if you are uncomfortable with this hack feel free to close, if it is good enough for now merge it in :)

danielwaterworth:
Agreed, let's not let perfect be the enemy of good.

PR #8135, opened 2019-10-02
[FIX] Webauthn security key fixes after real-world usage

mjrbrennan:
Fixes for initial round of QA/real-world usage outlined in https://meta.discourse.org/t/webauthn-support/126454/12

* Fix security key 2FA not working in the mobile login view 

image

* Bring the password confirmation & reset password link UI for the second factor preferences area of the profile into line with other UI elements in the app

image

* Change copy from Disable when removing security keys to Delete

image

* Improve error message copy when a user attempts to register a security key they already have registered ("borrowing" Chrome's excellent copy here)

image

### DIFF
--- a/app/assets/javascripts/discourse/templates/mobile/modal/login.hbs
+++ b/app/assets/javascripts/discourse/templates/mobile/modal/login.hbs
@@ -40,8 +40,21 @@
           secondFactorMethod=secondFactorMethod
           secondFactorToken=secondFactorToken
           class=secondFactorClass
+          backupEnabled=backupEnabled
           isLogin=true}}
-          {{second-factor-input value=secondFactorToken inputId='login-second-factor' secondFactorMethod=secondFactorMethod backupEnabled=backupEnabled}}
+          {{#if showSecurityKey}}
+            {{#security-key-form
+              allowedCredentialIds=securityKeyAllowedCredentialIds
+              challenge=securityKeyChallenge
+              showSecurityKey=showSecurityKey
+              showSecondFactor=showSecondFactor
+              secondFactorMethod=secondFactorMethod
+              otherMethodAllowed=secondFactorRequired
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/discourse/templates/mobile/modal/login.hbs
```diff
@@ -54,11 +67,13 @@
...
-        icon="unlock"
-        label=loginButtonLabel
-        disabled=loginDisabled
-        class="btn btn-large btn-primary"}}
+      {{#unless showSecurityKey }}
+        {{d-button action=(action "login")
+          icon="unlock"
+          label=loginButtonLabel
+          disabled=loginDisabled
+          class="btn btn-large btn-primary"}}
```
eviltrout: I believe `d-button` already has `btn` as a class.

app/assets/javascripts/discourse/templates/preferences-second-factor.hbs
```diff
@@ -142,16 +142,14 @@
...
-
           {{#unless showEnforcedNotice}}
-            {{cancel-link route="preferences.account" args= model.username}}
+            {{cancel-link route="preferences.account" args=model.username}}
+          {{/unless}}
+        </div>
+        <div class="controls" style="margin-top: 5px">
+          {{resetPasswordProgress}}
+          {{#unless resetPasswordLoading}}
+            <a href="" class="instructions" {{action "resetPassword"}}>{{ i18n 'user.second_factor.forgot_password' }}</a>
```
eviltrout:
```suggestion
            <a href class="instructions" {{action "resetPassword"}}>{{i18n 'user.second_factor.forgot_password'}}</a>
```

featheredtoast:
I'm going to merge this in, so mobile login isn't broken; other code suggestions here are pretty trivial. Thanks @mjrbrennan !

PR #8136, opened 2019-10-02
FIX: Update nginx.sample.conf to fix IPv6

max-foss:
Fix of broken IPv6 with the Nginx sample config.


max-foss:
Do not use, I will bring a proper one later. Free to be deleted.

PR #8137, opened 2019-10-02
FIX: Check for category conflicts in SiteSetting validations

CvX:
It was possible to add a category to more than one default group, e.g. "default categories muted" and "default categories watching first post".

The bug was caused by category validations inadvertently comparing strings and numbers.


eviltrout:
I tend to prefer integers being represented by integers. Could we instead change it at the point of comparison?

CvX:
Sure, I could go either way.

Comparing integers makes sense from the general type correctness standpoint, while comparing strings reduces complexity (we have two string inputs and just compare them, without any conversion)

eviltrout:
I feel like long term we'll have fewer bugs if we cast them to integers (although this is hard to prove). It also means we're less likely to introduce things like SQL injection down the line. Comparing ints is also faster.

PR #8138, opened 2019-10-02
DEV: Remove SiteSetting.default_categories_selected

CvX:
Added on Aug 21, 2015 in bef80633b1b58189a732a3959d3a59da44aa55e9
The only usage removed on Aug 26, 2015 in 4ba89eec2799390feea4fae5ae5032013106f092

As far I can tell it isn't used by core or any official plugins.


SamSaffron:
sure lets kill it and see what happens! :+1:

PR #8139, opened 2019-10-02
FIX: first post true if user creates topic first

justindirose:
Reported here: https://meta.discourse.org/t/user-was-banned-for-posting-a-reply-within-3-seconds/128823/12
The problem here is the user could create a topic, then reply and get
silenced on the second time (though technically their first post) for being below the
`min_first_post_typing_time` threshold.

### DIFF
--- a/lib/new_post_manager.rb
+++ b/lib/new_post_manager.rb
@@ -40,7 +40,8 @@ class NewPostManager
 
     !!(
       args[:first_post_checks] &&
-      user.post_count == 0
+      user.post_count == 0 &&
+      user.topic_count == 0
     )
   end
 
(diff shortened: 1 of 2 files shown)


eviltrout:
This looks good @justindirose but you should add a test for it. Let me know if you need help doing so!

justindirose:
Ah yes thanks @eviltrout -- I'm learning when it's best to add tests. I'll check it out!

justindirose:
Test is now added!

eviltrout:
Thanks @justindirose

bjorn:
It seems so strange to me to not count a topic as a post, but I'm sure it makes sense internally (or at least to count them separately). Thanks for fixing up the first-post condition to take into account topics as well! I think it will prevent many false positives with the auto-silencing based on typing speed.

PR #8142, opened 2019-10-02
FIX: EmojiOne is now JoyPixels

jjaffeux:


eviltrout:
Ugh, does this mean we'll have to update all our `emoji_one` references eventually :scream:

PR #8144, opened 2019-10-02
FIX: update user vote count on topic trash/recover

markvanlan:
Goes with this PR for discourse-voting
https://github.com/discourse/discourse-voting/pull/46


pmusaraj:
Merging this, the accompanying discourse-voting PR needs it.

PR #8146, opened 2019-10-02
[FIX] Check webauthn support when registering security keys

mjrbrennan:
When registering security keys check for the presence of `PublicKeyCredential` and if it is missing show an error message saying Webauthn is not supported:

image


SamSaffron:
Cool this looks good to me! :confetti_ball:

PR #8153, opened 2019-10-03
FEATURE: Input for name when creating a new authenticator 

markvanlan:
Added an input to name name a new authenticator. Initially, the value is set to `My Authenticator`, which is the default name currently.

Screenshot from 2019-10-03 13-27-10


pmusaraj:
Would it be too complicated to have the default value be empty and add `My Authenticator` as a placeholder? It does seem useful to force the user to choose a specific name.

pmusaraj:
Related: I also think the Code field needs some help text. Here is what Github has: 

image

markvanlan:
> What happens if they submit an empty string?
It does not submit and displays: `param is missing or the value is empty: name`

markvanlan:
I need to commit the placeholder changes before this gets merged.

eviltrout:
@markvanlan looks like there's a linting failure now. You are good to merge once you sort that out!

PR #8155, opened 2019-10-04
UX: Add "Drafts" to quick access profile tab.

xrav3nz:
> CONTEXT: https://meta.discourse.org/t/please-add-drafts-to-user-menu/129840

I talked to @techAPJ and "stole" this from him 🙈.


SamSaffron:
That is one awkward translation key ... yikes `user_action_groups.15`

Looks good though!

eviltrout:
@SamSaffron oh yes I really dislike those. They're quite old though. Would be a great task to replace them.

PR #8156, opened 2019-10-04
FEATURE: Weighted reviewable user accuracy

markvanlan:
@eviltrout Accuracy is now + or - relative to the 70% accuracy point.

This accuracy value is then multiplied by a logarithmic value based on total number of votes. This gives more weight to users who have lots of votes, but also starts to flat-line a certain point so that accuracy bonuses don't get too high. To flatten the graph out, the log base (in this case 4) can be increased.

I wrote the tests to show how a user' accuracy can swing with low accuracy/low flags, to high accuracy/high flags.


eviltrout:
@markvanlan this looks good, thanks!

PR #8159, opened 2019-10-04
FIX: moves reports locales to client

jjaffeux:


jjaffeux:
@gschlager what would be the best way to move all the other locales from server to client please?

eviltrout:
This is adding a bunch more strings to the front end app. Is fix because the reports are cached with the incorrect locale? 

It might be better to only cache the parts of the reports without titles, or if that's too complicated to use the locale in the cache key.

jjaffeux:
Hm indeed that's a very good point 🤔 Will give it more thought

gschlager:
> This is adding a bunch more strings to the front end app.

It's staff users only, so it shouldn't be that bad, right?
An alternative could be to do the actual translation on the server after loading the data from the cache. I guess it might be possible to iterate all properties in the data which are known to contain translation keys. 🤷‍♂

eviltrout:
It is probably simpler to cache the fragment of the JSON string that changes and not the rest (headers, etc.) 

I do agree that admin only is less risky, but I think this is a relatively easy way to prevent a bunch of extra strings from taking up download time and memory.

PR #8160, opened 2019-10-04
UX: Better composer hyperlink modal

pmusaraj:
This brings the hyperlink modal in line with all other modals (upload, date, etc.) by placing the fixed modal element outside of the composer. 

Before: 

image

After: 

image

### DIFF
--- a/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
+++ b/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
@@ -5,9 +5,6 @@ export default Ember.Controller.extend(ModalFunctionality, {
   linkUrl: "",
   linkText: "",
 
-  @computed()
-  title: () => "composer.link_dialog_title",
-
   onShow() {
     Ember.run.next(() =>
       $(this)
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -0,0 +1,50 @@
+import ModalFunctionality from "discourse/mixins/modal-functionality";
+import { default as computed } from "ember-addons/ember-computed-decorators";
+
+export default Ember.Controller.extend(ModalFunctionality, {
+  linkUrl: "",
+  linkText: "",
+
+  @computed()
+  title: () => "composer.link_dialog_title",
```
CvX: Does it have to be a computed property if the value is static?

pmusaraj: Good spot, no, it doesn't need to be computed at all.

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -0,0 +1,47 @@
+import ModalFunctionality from "discourse/mixins/modal-functionality";
+import { default as computed } from "ember-addons/ember-computed-decorators";
```
CvX: ESLint error, `computed` is no longer used: https://travis-ci.org/discourse/discourse/jobs/593993307#L701-L702

SamSaffron:
I support this cleanup, feel free to merge!

CvX:
@pmusaraj It seems it's no longer possible to submit the form by pressing enter/return. Could this PR be the cause?

PR #8161, opened 2019-10-05
DEV: improves speed of live reload css in core and plugins

jjaffeux:


SamSaffron:
I think this looks fine to me, feel free to merge.

PR #8162, opened 2019-10-05
FIX: Clarify revert text

DT9:
When you open the revision history of a post a button is shown on the bottom right labelled “Revert to this revision”. It is shown under the later revision on the right, but it actually refers to the older revision on the left.

It might be less confusing if it were renamed.


coding-horror:
Hmm no we have other plans here, such as putting the revision number on the
button per discussion on meta.

SamSaffron:
sorry, closing this per @coding-horror

PR #8163, opened 2019-10-06
import_scripts: add fluxbb prefix to missing query

jelly:
Signed-off-by: Jelle van der Waa <jelle@archlinux.org>


SamSaffron:
oh are you considering running arch forums on discourse? I would love that, btw if you ever decide to do so we would be super happy to host the instance for you free of charge, forever, I am an avid arch user and struggle with the current forum a lot, the notification story is so weak I keep needing to head to the forum to see if people replied to me.

svenstaro:
It might just happen. @SamSaffron could you hop into #archlinux-devops with us and give us a hand with a test migration?

PR #8165, opened 2019-10-07
FIX: Store user_accuracy_bonus to clarify explanations

markvanlan:


db/migrate/20191007140446_add_user_accuracy_bonus_to_reviewable_scores.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AddUserAccuracyBonusToReviewableScores < ActiveRecord::Migration[6.0]
+  def up
+    # Add user_accuracy_bonus to reviewable_scores
+    execute <<~SQL
+    ALTER TABLE reviewable_scores
```
eviltrout: The ALTER part should use rails helpers to be consistent with our other migrations.

db/migrate/20191007140446_add_user_accuracy_bonus_to_reviewable_scores.rb
```diff
@@ -0,0 +1,30 @@
...
+    ADD COLUMN user_accuracy_bonus float
+    DEFAULT 0
+    SQL
+
+    # Set user_accuracy_bonus = score - user.trust_level - 1
+    execute <<~SQL
+    UPDATE reviewable_scores
+    SET user_accuracy_bonus = score - (
```
eviltrout: I tend to prefer an UPDATE with a JOIN for this kind of thing rather than a correlated subquery. However, there probably aren't many scores in most databases so maybe it doesn't matter?

db/migrate/20191007140446_add_user_accuracy_bonus_to_reviewable_scores.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class AddUserAccuracyBonusToReviewableScores < ActiveRecord::Migration[6.0]
+  def up
+    # Add user_accuracy_bonus to reviewable_scores
+    add_column :reviewable_scores, :user_accuracy_bonus, :float, default: 0
```
eviltrout: Very minor but I'd add a `null: false` here to make sure the column is not nullable.

PR #8167, opened 2019-10-07
FIX: updates discourse-ember-source gem

jjaffeux:
This is related to fix made to prevent a crash in iOS 9.5


SamSaffron:
seems good to me

PR #8169, opened 2019-10-08
FIX: Prevent from creation of duplicated TopicAllowedUsers

lis2:
Something changed in Rails6 and now it behaves correctly. When we try to build a collection object with the same params it creates duplication which later we try to save.

Ensure that we don't try to create duplicated TopicAllowedUsers

Related to https://meta.discourse.org/t/error-message-topic-allowed-users-is-invalid/130382/5

Screenshot from 2019-10-08 15-42-00

### DIFF
--- a/app/models/topic_converter.rb
+++ b/app/models/topic_converter.rb
@@ -81,6 +81,7 @@ class TopicConverter
       user.user_stat.save!
     end
     @topic.topic_allowed_users.build(user_id: @user.id) unless @topic.topic_allowed_users.where(user_id: @user.id).exists?
+    @topic.topic_allowed_users = @topic.topic_allowed_users.uniq(&:user_id)
     # update topics count
     @topic.user.user_stat.topic_count -= 1
     @topic.user.user_stat.save!
--- a/spec/models/topic_converter_spec.rb
+++ b/spec/models/topic_converter_spec.rb
@@ -114,7 +114,7 @@ describe TopicConverter do
 
     context 'success' do
       it "converts regular topic to private message" do
-        private_message = topic.convert_to_private_message(admin)
+        private_message = topic.convert_to_private_message(post.user)
         expect(private_message).to be_valid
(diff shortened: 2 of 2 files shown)


app/models/topic_converter.rb
```diff
@@ -75,7 +75,6 @@ def update_user_stats
   def add_allowed_users
     @topic.posts.where(deleted_at: nil).each do |p|
       user = User.find(p.user_id)
-      @topic.topic_allowed_users.build(user_id: user.id) unless @topic.topic_allowed_users.where(user_id: user.id).exists?
```
SamSaffron: This is adding `user.id` not `@user.id` this is the user associated with the post. 

So instead, keep track of all the new users in a hash or array and do the addition the end.

```
users = {}
if ...  
     # allows for deleted users
     user = users[p.user_id] ||= User.find_by(id: p.user_id)
     if user
       user.user_stat...
     end
end

finally ensuring al these new users exist in topic allowed users...

Comment (lis2) [reply app/models/topic_converter.rb]: You are right, my bad. What do you think of instead of building a hash of users, remove duplications just before creation?

File: app/models/topic_converter.rb [hunk 2]
```diff
@@ -81,6 +81,7 @@ def add_allowed_users
       user.user_stat.save!
     end
     @topic.topic_allowed_users.build(user_id: @user.id) unless @topic.topic_allowed_users.where(user_id: @user.id).exists?
+    @topic.topic_allowed_users = @topic.topic_allowed_users.uniq(&:user_id)
```
SamSaffron: does this fix it? Can you add a failing test case here?

lis2: @SamSaffron fixed - https://github.com/discourse/discourse/pull/8169/commits/5244b3d0e4784606c92e2fed4e1f2f1a273524bf

SamSaffron:
OK lets give this a shot!

PR #8170, opened 2019-10-08
DEV: support --fail-fast in bin/turbo_rspec

markvanlan:


lib/turbo_tests/runner.rb
```diff
@@ -229,5 +235,12 @@ def handle_messages
       rescue Interrupt
       end
     end
+
+    def fail_fast_met
+      return false if @fail_fast.nil? || @fail_fast < @failure_count
+
+      @threads.each(&:kill)
```
danielwaterworth: With such a short method, this is a borderline nitpick, but I like to see more of a separation between methods that do things and methods that report on the state of things. As a caller, I don't expect a method called `fail_fast_met` to start killing threads.

Other than that, I'd be happy to see this merged.

PR #8174, opened 2019-10-09
FIX: Narrative Bot certificates are ERB templates

lis2:
There are at least two ways of rendering templates outside of the controller. The first one is Rails way enabled with Rails 5 https://evilmartians.com/chronicles/new-feature-in-rails-5-render-views-outside-of-actions
The downside of this method is that all variables need to be passed as params (I could find a way to pass the whole context)

Another way is to use instance_eval described in Erubi documentation
https://github.com/jeremyevans/erubi#usage - it works perfectly fine, however, I didn't feel very confident about using eval unless necessary.

An additional benefit of using `ApplicationController.render` is that if Rails would change the ERB engine in the future, this code should still work.

If you want to test it on your local, you need to be signed in and then that two URLs are generating certificates:
http://localhost:3000/discobot/certificate.svg?date=Oct+07+2019&type=standard&user_id=1
http://localhost:3000/discobot/certificate.svg?date=Oct+07+2019&type=advanced&user_id=1

Dev: https://dev.discourse.org/t/discourse-narrative-bot-should-not-be-storing-giant-strings/17130


SamSaffron:
This looks good to me (checked out on local cause the diff is so big) 

TIL the inline renders define a method on an object and then erase it. 

@eviltrout this opens the door to making this extensible... simply adding an accessor to certificate_generator.rb will do the trick and then simply giving it different filenames.

eviltrout:
Great, there is a customer of ours who wants to customize this and it'll be a great test of it.

PR #8175, opened 2019-10-09
Serialize draft

fzngagan:
This PR adds a function `serializeToDraft` to the composer model to allow saving topic custom fields to draft and setting them back when the draft is reopened.

Here is the post on meta for this.
https://meta.discourse.org/t/including-composer-custom-fields-on-save-draft/126913/2

### DIFF
--- a/app/assets/javascripts/discourse/models/composer.js.es6
+++ b/app/assets/javascripts/discourse/models/composer.js.es6
@@ -58,17 +58,17 @@ const CLOSED = "closed",
     featuredLink: "topic.featured_link"
   },
   _draft_serializer = {
-    reply: 'reply',
-    action: 'action',
-    title: 'title',
-    categoryId: 'categoryId',
-    archetypeId: 'archetypeId',
-    whisper: 'whisper',
-    metaData: 'metaData',
-    composerTime: 'composerTime',
-    typingTime: 'typingTime',
-    postId: 'post.id',
-    usernames: 'targetUsernames',
+    reply: "reply",
+    action: "action",
+    title: "title",
+    categoryId: "categoryId",
+    archetypeId: "archetypeId",
+    whisper: "whisper",
+    metaData: "metaData",
+    composerTime: "composerTime",
+    typingTime: "typingTime",
+    postId: "post.id",
+    usernames: "targetUsernames",
   },
   _add_draft_fields = {},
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -1097,6 +1099,18 @@ Composer.reopenClass({
     return Object.keys(_create_serializer);
   },
 
+  serializeToDraft(fieldName, property){
```
eviltrout: All javascript extensibility should go via the plugin-api so it can be versioned and so we can support backwards compatibility if need be.

fzngagan:
> It would be great if there was a test for this. Did you investigate if it could be tested easily?

We did investigate this but noticed that there's no straightforward way of testing it.

Also similar functions `serializeToTopic` and `serializeOnCreate` don't have tests.

app/assets/javascripts/discourse/models/composer.js.es6
fzngagan: The methods I referred here were the ember side Category model methods `serializeOnCreate` and `serializeToTopic` and thought that this method should be here itself. Is it fine if I create an alias of the same method for `plugin-api` which essentially calls this method? This would solve the versioning and backwards compatibility problems.

eviltrout: Yes creating the alias is fine, and all plugins should access them that way.

fzngagan:
> Thanks, looks good! I was about to merge but I've noticed there are some linting failures.

Sorry about that. It was a weird setting on my IDE which caused that. Its fixed now.

eviltrout:
It's still failing unfortunately. You can see the failures in Travis by clicking Details above.

fzngagan:
Ok @eviltrout now all the checks are passing. :)

eviltrout:
Thanks! Merged now.

fzngagan:
@eviltrout , how about bumping the plugin-api version.

eviltrout:
@fzngagan ah yes, good call. Can you do a follow up PR for that?

fzngagan:
What version do you want me to bump it to?

eviltrout:
0.8.33 I believe is next. Let's go with that!

PR #8176, opened 2019-10-09
FIX: Order UserFields by position, by default

markvanlan:
This solves the issue brought up [in this topic](https://meta.discourse.org/t/onboarding-process-for-new-member/74837/8), and I don't see why we wouldn't always order the UserField by the position column.

The solution for the topic mentioned could be solved by sorting the user fields all over the place, but this works, and might even solve the same issue elsewhere that we are not aware of.


eviltrout:
Wow, nice and simple! Thanks.

PR #8177, opened 2019-10-10
FIX: Show a correct diff when editing consecutive paragraphs

nachocab:
Currently, if two consecutive paragraphs are edited, their contents are not properly diffed:
Screen Shot 2019-10-10 at 09 35 13

The current algorithm performs a pairwise diff whenever a deletion follows an addition or vice versa. This happens here:

https://github.com/discourse/discourse/blob/c5326682d6346c13b495ece9f3ee4e82f1b13a72/lib/discourse_diff.rb#L38-L39

However, in the example above we have two additions followed by two deletions. This leads to:
- an added paragraph: `this is one great paragraph`
- a pairwise comparison: `this is one paragraph` vs. `this is another` 
- a deleted paragraph: `here is yet another`

The ideal approach would be to perform an alignment step before diffing (which is not trivial to implement, see [Gale-Church algorithm](https://en.wikipedia.org/wiki/Gale–Church_alignment_algorithm)), so a good compromise is to use one of the alternative edit sequences to the one returned by ONPDiff. 

In other words, `ONPDiff.paragraph_diff` transforms the edit sequence from `ONPDiff.diff` `Add1 Add2 Delete1 Delete2` into this one `Add1 Delete1 Add2 Delete2`. 

Current behavior using `ONPDiff.diff`:

image

Proposed behavior using `ONPDiff.paragraph_diff`:

image

### DIFF
--- a/lib/discourse_diff.rb
+++ b/lib/discourse_diff.rb
@@ -12,7 +12,7 @@ class DiscourseDiff
     before_markdown = tokenize_line(CGI::escapeHTML(@before))
     after_markdown = tokenize_line(CGI::escapeHTML(@after))
 
-    @block_by_block_diff = ONPDiff.new(before_html, after_html).diff
+    @block_by_block_diff = ONPDiff.new(before_html, after_html).paragraph_diff
     @line_by_line_diff = ONPDiff.new(before_markdown, after_markdown).short_diff
   end
 
--- a/lib/onpdiff.rb
+++ b/lib/onpdiff.rb
@@ -25,6 +25,10 @@ class ONPDiff
     @short_diff ||= build_short_edit_script(compose)
   end
 
+  def paragraph_diff
+    @paragraph_diff ||= build_paragraph_edit_script(diff)
+  end
+
   private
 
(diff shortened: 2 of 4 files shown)


ZogStriP:
Wow 😍

SO MUCH BETTERER 👏

ZogStriP:
Can you benchmark this on very large diffs? Both when there are lots of text but a small number of diffs or a large-ish amount of text but a lot of diffs.

nachocab:
> Can you benchmark this on very large diffs? Both when there are lots of text but a small number of diffs or a large-ish amount of text but a lot of diffs.
Sure, do you need me to use a specific tool or just to sanity check it on my computer?

ZogStriP:
> Sure, do you need me to use a specific tool or just to sanity check it on my computer?

Sanity check is fine as long as I have before/after ;)

> And how big do you want the diffs?

Regarind text size, near the default maximum post lenght.

nachocab:
They were both equally fast (less than a second, probably less than .5s): 

image

PR #8178, opened 2019-10-10
FEATURE: search topics when adding a link in composer

pmusaraj:
When adding a hyperlink, typing something in the URL field that isn't a URL (i.e. doesn't start with `http`) will trigger a search of topics. 

image

When selecting a topic from the results, the topic url and title will be added to the respective fields. 

Related: https://meta.discourse.org/t/search-existing-topics-when-hyperlinking/112922.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
+++ b/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
@@ -172,7 +172,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
         this.selectLink(e.target);
       }
     },
-    search(value) {
+    search() {
       this._debounced = Ember.run.debounce(this, this.triggerSearch, 400);
     }
   }
--- a/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
+++ b/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
@@ -44,6 +44,13 @@ export default Ember.Controller.extend(ModalFunctionality, {
           e.stopPropagation();
         }
         break;
+      case 27:
+        if (this.searchResults.length) {
+          this.set("searchResults", []);
+          e.preventDefault();
+          e.stopPropagation();
+        }
+        break;
     }
   },
 
--- a/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
+++ b/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
@@ -30,22 +30,24 @@ export default Ember.Controller.extend(ModalFunctionality, {
   keyDown(e) {
     switch (e.which) {
       case 40:
-        this.selectRow(e, "down");
+        this.highlightRow(e, "down");
         break;
       case 38:
-        this.selectRow(e, "up");
+        this.highlightRow(e, "up");
         break;
       case 13:
         // override Enter behaviour when a row is selected
         if (this.selectedRow > -1) {
-          this.setUpdated(this.selectedRow);
-          document.querySelector("input.link-text").focus();
+          const selected = document.querySelectorAll(
+            ".internal-link-results .search-link"
+          )[this.selectedRow];
+          this.selectLink(selected);
           e.preventDefault();
           e.stopPropagation();
         }
         break;
       case 27:
-        // override Esc too
+        // Esc should cancel dropdown first
         if (this.searchResults.length) {
           this.set("searchResults", []);
           e.preventDefault();
@@ -61,7 +63,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
     }
   },
 
-  selectRow(e, direction) {
+  highlightRow(e, direction) {
     const index =
       direction === "down" ? this.selectedRow + 1 : this.selectedRow - 1;
 
@@ -72,24 +74,24 @@ export default Ember.Controller.extend(ModalFunctionality, {
       this.set("selectedRow", index);
     } else {
       this.set("selectedRow", -1);
-      document.querySelector("input.link-text").focus();
+      document.querySelector("input.link-url").focus();
     }
 
     e.preventDefault();
   },
 
-  setUpdated(index) {
-    const topic = this.searchResults[index];
-    if (topic) {
-      this.set("linkUrl", Discourse.BaseUrl + topic.url);
-      if (!this.linkText) {
-        this.set("linkText", topic.title);
-      }
-    }
+  selectLink(el) {
     this.setProperties({
+      linkUrl: el.href,
       searchResults: [],
       selectedRow: -1
     });
+
+    if (!this.linkText && el.dataset.title) {
+      this.set("linkText", el.dataset.title);
+    }
+
+    document.querySelector("input.link-text").focus();
   },
 
   triggerSearch() {
@@ -163,8 +165,12 @@ export default Ember.Controller.extend(ModalFunctionality, {
     cancel() {
       this.send("closeModal");
     },
-    select(index) {
-      this.setUpdated(index);
+    linkClick(e) {
+      if (!e.metaKey && !e.ctrlKey) {
+        e.preventDefault();
+        e.stopPropagation();
+        this.selectLink(e.target);
+      }
     },
     search(value) {
       this._debounced = Ember.run.debounce(this, this.triggerSearch, 400);
--- a/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
+++ b/app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
@@ -1,10 +1,10 @@
 import ModalFunctionality from "discourse/mixins/modal-functionality";
 import { searchForTerm } from "discourse/lib/search";
-import { observes } from "ember-addons/ember-computed-decorators";
-
-let searchTopics;
 
 export default Ember.Controller.extend(ModalFunctionality, {
+  _debounced: null,
+  _activeSearch: null,
+
   onShow() {
     this.setProperties({
       linkUrl: "",
@@ -45,6 +45,7 @@ export default Ember.Controller.extend(ModalFunctionality, {
         }
         break;
       case 27:
+        // override Esc too
         if (this.searchResults.length) {
           this.set("searchResults", []);
           e.preventDefault();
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
-  linkUrl: "",
-  linkText: "",
-
   onShow() {
-    Ember.run.next(() =>
-      $(this)
-        .find("input.link-url")
-        .focus()
-    );
+    this.set("linkUrl", "");
```
jjaffeux: should be grouped:

```
this.setProperties({
  linkUrl: "",
  linkText: "",
  searchResults: [],
  selectedRow: -1
});
```

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+    this.set("linkText", "");
+    this.set("searchResults", []);
+    this.set("selectedRow", -1);
+
+    Ember.run.next(() => {
```
jjaffeux: Can you explain why you use next and not afterRender here ?

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+      const element = document.querySelector(".insert-link");
+
+      element.addEventListener("keydown", e => {
```
jjaffeux: this listed should be removed, in the `onClose` hook maybe?

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+              e.preventDefault();
+              e.stopPropagation();
+            }
+            break;
+        }
+      });
+
+      element
+        .closest(".modal-inner-container")
+        .addEventListener("mousedown", e => {
```
jjaffeux: this listener should also be removed

jjaffeux:
Also note, once I release my select-kit update with better async support, this will be done in few lines of code. I do love the idea though!!!

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+    if (topic) {
+      this.set("linkUrl", Discourse.BaseUrl + topic.url);
+      if (!this.linkText) {
+        this.set("linkText", topic.title);
+      }
+    }
+    this.set("selectedRow", -1);
+  },
+
+  @observes("linkUrl")
```
jjaffeux: we really try to not use observers, they are mostly an implementation detail of Ember, do you really need this ?

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+  triggerSearch() {
+    if (
+      this.linkUrl &&
+      this.linkUrl.length > 3 &&
+      !this.linkUrl.startsWith("http")
```
jjaffeux: startsWith is not available under IE11 (https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/startsWith) and we decided to not polyfill it because you can do it very simply with `indexOf`.

app/assets/stylesheets/common/base/modal.scss
```diff
@@ -201,9 +201,35 @@
     }
 
     &.insert-link {
+      overflow-y: visible;
       input {
         min-width: 300px;
       }
+
+      .inputs {
```
jjaffeux: did you test this on mobile ?

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+    ) {
+      const searchTopics = function() {
```
jjaffeux: Seems low probability but on a slow network I think we could have results showing of previous search showing when using backspace and going under 4 chars for linkUrl

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,98 @@
...
+        searchForTerm(this.linkUrl, { typeFilter: "topic" }).then(results => {
+          if (results && results.topics && results.topics.length > 0) {
+            this.set("searchResults", results.topics);
+          } else {
+            this.set("searchResults", []);
+          }
+        });
+      };
+
+      Ember.run.debounce(this, searchTopics, 400);
```
jjaffeux: I know we don't do it often, but a debounce like this one is a potential leak, you should cancel it on close

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
pmusaraj: No reason, that call to next was already there. But replacing it with `afterRender` makes sense.

jjaffeux:
I'm also concerned by the lack of loading indicator, you really don't know if something is happening here.

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
jjaffeux: Also this debounce is 100% non functional, try to add a `console.log("foo")`  in the `searchTopics` and type in the input, it will take 400ms to appear, but  you should have multiple foo showing, as it's just creating a debounce for each char, debounce can be tricky, and it should def not be initialised inside an observer.

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
pmusaraj: I can't figure out how to do this cleanly without an observer. Do you have a specific suggestion?

app/assets/stylesheets/common/base/modal.scss
pmusaraj: Yes, I did, the overflow doesn't apply on mobile. Here's a screenshot: 

IMG_70DF6904735D-1

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
pmusaraj: I don't think this can happen because the results get cleared if the input has less than 4 chars (or if it's a URL).

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
pmusaraj: I couldn't reproduce the debounce being non-functional. It works correctly when I add my `console.log`, it displays it for every search request. Keep in mind that the debounce is only triggered if character count is at least 4 and string starting with something other than http.

pmusaraj: Thanks, done.

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
pmusaraj: Nevermind, you're right about this, it does happen.

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
jjaffeux: Try a higher value than 400 ms you will repro easy

app/assets/javascripts/discourse/controllers/insert-hyperlink.js.es6
```diff
@@ -1,15 +1,130 @@
 import ModalFunctionality from "discourse/mixins/modal-functionality";
+import { searchForTerm } from "discourse/lib/search";
+import { observes } from "ember-addons/ember-computed-decorators";
 
-export default Ember.Controller.extend(ModalFunctionality, {
-  linkUrl: "",
-  linkText: "",
+let searchTopics;
```
jjaffeux: You should make sure this is cleaned after tests, otherwise we carry this forever doing tests.

app/assets/javascripts/discourse/templates/modal/insert-hyperlink.hbs
```diff
@@ -1,6 +1,18 @@
 {{#d-modal-body title="composer.link_dialog_title" class="insert-link"}}
     <div class="inputs">
       {{text-field value=linkUrl placeholderKey="composer.link_url_placeholder" class="link-url"}}
```
jjaffeux: this is the issue here for the observer, you are basically relying on mutation of linkUrl, instead of having an event when the textfield is changed, you could have an event on input and make `linkUrl` readonly, this ways it would make the whole code better.

I realise this is not your code at the beginning, but you are now impacted by this "bad" patter :p

app/assets/javascripts/discourse/templates/modal/insert-hyperlink.hbs
```diff
@@ -1,6 +1,18 @@
...
+      {{#if searchLoading}}
+        {{loading-spinner}}
+      {{/if}}
+      {{#if searchResults}}
+        <div class="internal-link-results">
+          {{#each searchResults as |r index|}}
+            <a class="search-link" href="#" {{action "select" index}}>
```
jjaffeux: I'm unsure here, but wouldn't just `href` alone be enough/better ?

test/javascripts/acceptance/composer-hyperlink-test.js.es6
```diff
@@ -70,4 +70,38 @@ QUnit.test("add a hyperlink to a reply", async assert => {
...
+  assert.equal(
+    find(".internal-link-results").length,
+    1,
+    "shows internal links search dropdown when entering keywords"
+  );
+
+  await keyEvent(".insert-link", "keydown", 40);
+  await keyEvent(".insert-link", "keydown", 13);
+
+  assert.equal(
```
jjaffeux: you can use `exists` helper for this case instead of checking on length

pmusaraj:
@jjaffeux I think I have now addressed the observer/debounce issues.

jjaffeux:
Yep thx!! Great work, LGTM ❤️

PR #8179, opened 2019-10-10
DEV: Split max decompressed setting for themes and backups

romanrizzi:


config/site_settings.yml
```diff
@@ -1201,9 +1201,12 @@ files:
     default: 5
     min: 0
     max: 20
-  decompressed_file_max_size_mb:
+  decompressed_theme_max_file_size_mb:
```
eviltrout: I think we should try to migrate their custom values since we are renaming the setting.

romanrizzi: Done. I also added an error message.

PR #8180, opened 2019-10-10
FEATURE: Option to update category preferences of all users when site setting changed

vinothkannans:
While changing the default categories user preferences site settings now we can update the change to all existing users.


ravenzachary:
Vinoth - thank you for submitting this code. I am looking forward to using it in an upcoming Discourse release.A question for you about hos this code works - if I already have a category set for  "default categories watching first post", after I upgrade Discourse to a new version with this code included, what is the best solution for me to populate that existing category setting into all of the existing users that don't have it already? Remove the existing category and re-add it to refresh the copy to all user accounts?

PR #8189, opened 2019-10-14
FIX: show edited site text overrides for non `_MF` keys

markvanlan:
This also fixes the `expanding-text-area`. 

`this.value` was being updated, but `this.element.value` was not, so the textarea was not actually expanding.

### DIFF
--- a/app/models/translation_override.rb
+++ b/app/models/translation_override.rb
@@ -100,7 +100,7 @@ class TranslationOverride < ActiveRecord::Base
 
   def transform_pluralized_key(key)
     match = key.match(/(.*)\.(zero|two|few|many)$/)
-    match ? match.to_a.second + '.one' : key
+    match ? match.to_a.second + '.other' : key
   end
 end
 
--- a/app/models/translation_override.rb
+++ b/app/models/translation_override.rb
@@ -65,8 +65,10 @@ class TranslationOverride < ActiveRecord::Base
   private
 
   def check_interpolation_keys
+    transformed_key = transform_pluralized_key(translation_key)
+
     original_text = I18n.overrides_disabled do
-      I18n.t(translation_key, locale: :en)
+      I18n.t(transformed_key, locale: :en)
     end
 
     if original_text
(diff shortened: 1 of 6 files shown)


config/locales/client.en.yml
```diff
@@ -1996,9 +1996,11 @@ en:
       unread: "unread"
       new_topics:
         one: "%{count} new topic"
+        few: "%{count} new topics"
```
gschlager: You can't add the `few` key to pluralized strings in English. English uses only `one` and `other`.

spec/models/translation_override_spec.rb
```diff
@@ -96,4 +96,33 @@ def cached_value(guardian, types_name, name_key, attribute)
       include_examples "resets site text"
     end
   end
+
+  describe '#transform_pluralized_key' do
+    let(:translation_override) { TranslationOverride.new }
+    it "transforms key that end with a pluralization to end with 'one'" do
+      transformed_key = 'js.new_topics.one'
+      key = translation_override.send(:transform_pluralized_key, "js.new_topics.few")
```
gschlager: I'm not a fan of testing private methods. The `transform_pluralized_key` method should just be an implementation detail. It would be better to test that `upsert!` works, otherwise it would be quite easy to break the functionality without us noticing.

markvanlan: Yeah fair. I'll do that.

app/models/translation_override.rb
```diff
@@ -96,6 +98,10 @@ def check_interpolation_keys
     end
   end
 
+  def transform_pluralized_key(key)
+    match = key.match(/(.*)\.(zero|two|few|many)$/)
+    match ? match.to_a.second + '.one' : key
```
gschlager: This should use the `other` key. Usually it doesn't matter, but `other` is the correct key to use.  Sorry for nitpicking. 😉

spec/models/translation_override_spec.rb
```diff
@@ -7,6 +7,7 @@
     describe '#value' do
       before do
         I18n.backend.store_translations(I18n.locale, some_key: '%{first} %{second}')
+        I18n.backend.store_translations(:en, something: { one: '%{first} %{second}', other: '%{first} %{second}' })
```
gschlager: I suggest using different values for `one` and `other`. That way your test can ensure that the interpolation keys from `other` are used.

PR #8191, opened 2019-10-14
More encoded slug fixes

xfalcox:
A bunch of extra places to consider when dealing with encoded slugs.

### DIFF
--- a/app/assets/javascripts/discourse/models/category.js.es6
+++ b/app/assets/javascripts/discourse/models/category.js.es6
@@ -243,7 +243,11 @@ Category.reopenClass({
   },
 
   findSingleBySlug(slug) {
-    return Category.list().find(c => Category.slugFor(c) === slug);
+    if (Discourse.SiteSettings.slug_generation_method !== "encoded") {
+      return Category.list().find(c => Category.slugFor(c) === slug);
+    } else {
+      return Category.list().find(c => Category.slugFor(c) === encodeURI(slug));
+    }
   },
 
   findById(id) {
--- a/app/jobs/onceoff/fix_encoded_category_slugs.rb
+++ b/app/jobs/onceoff/fix_encoded_category_slugs.rb
@@ -7,7 +7,7 @@ module Jobs
       return unless SiteSetting.slug_generation_method == 'encoded'
 
       #Make custom categories slugs nil and let the app regenerate with proper encoded ones
-      Category.where('id > 4').each do |c|
+      Category.all.reject { |c| c.seeded? }.each do |c|
         c.slug = nil
         c.save!
       end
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/discourse/models/category.js.es6
```diff
@@ -243,7 +243,11 @@ Category.reopenClass({
   },
 
   findSingleBySlug(slug) {
-    return Category.list().find(c => Category.slugFor(c) === slug);
+    if (Discourse.SiteSettings.slug_generation_method !== "encoded") {
```
eviltrout: Could you use `this.siteSettings`? Should be available in the Category model.

app/jobs/onceoff/fix_encoded_category_slugs.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Jobs
+
+  class FixEncodedCategorySlugs < ::Jobs::Onceoff
+    def execute_onceoff(args)
+      return unless SiteSetting.slug_generation_method == 'encoded'
+
+      #Make custom categories slugs nil and let the app regenerate with proper encoded ones
+      Category.where('id > 4').each do |c|
```
eviltrout: `id > 4`? why that specific hard coded value?

xfalcox: I want to skip the built-in categories (staff, uncategorized, Lounge, Site Feedback are 1,2,3,4). Is there a way to query those equivalent to `User.human?` or `Group.automatic?` ?

app/assets/javascripts/discourse/models/category.js.es6
xfalcox: I could use `Category.list()[0].siteSettings.slug_generation_method` to access it but IMO is ugly as hell, and looks like almost every model is using the global:

image

eviltrout: Ugh, good point. We should make it a service eventually.

app/jobs/onceoff/fix_encoded_category_slugs.rb
eviltrout: It doesn't look like there is. I think at least we should use a constant here. Perhaps add a `seeded?` method to `Category` that is true if the `id` is in that range. Later we could make that an actual column.

gschlager: I agree. It would be a good idea to add `seeded?` and check if the `id` equals one of the following site settings.

https://github.com/discourse/discourse/blob/48c123849228cc8a392e989558fe89734cd02656/config/site_settings.yml#L1823-L1835

xfalcox: Damn this is so much better! Thanks for the suggestion @gschlager !

PR #8192, opened 2019-10-15
FIX: public_file_server.enabled is false in test

lis2:
After a small conversation, we decided that we can set `public_file_server.enabled` to false in the `test` environment to have the same value as `production`.


SamSaffron:
I support this change, anything that brings test closer to prd without impacting perf makes me happy.

PR #8193, opened 2019-10-15
Add unique class to custom field wrappers

awesomerobot:
This makes hiding custom fields with CSS at signup much more reliable.


eviltrout:
@awesomerobot does `Ember.String.dasherize` not work?

eviltrout:
Sounds good.

PR #8194, opened 2019-10-15
FIX: include topic/PM link & unsubscribe link in email footer

techAPJ:
When existing users are invited to a topic or PM they are not seeing topic/PM link & unsubscribe link in email footer. This PR fixes that.

Message:
Screen Shot 2019-10-15 at 11 25 37

Topic:
Screen Shot 2019-10-15 at 11 29 21

### DIFF
--- a/app/mailers/user_notifications.rb
+++ b/app/mailers/user_notifications.rb
@@ -569,7 +569,15 @@ class UserNotifications < ActionMailer::Base
         site_description: SiteSetting.site_description
       )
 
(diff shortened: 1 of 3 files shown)


techAPJ:
> It's another erb that we'll need to figure out how to let users customize.

I see.... I don't think we necessarily need to add erb file to include topic/PM link. I updated the invite templates to add the link instead. https://github.com/discourse/discourse/commit/600233482f3bf499e7655a1596908eee32f8cbbb

PR #8207, opened 2019-10-17
DEV: Remove badge-select-controller

CvX:
1. The `selectableUserBadges` computed property of `badge-title` was being incorrectly overwritten, which triggered computed-property.override deprecation.

2. The `badge-select-controller` mixin contained properties that were used either by `badge-title` component or `badges/show` controller, but none were being used by both. This change moves properties where they belong, and removes the mixin.


app/assets/javascripts/discourse/components/badge-title.js.es6
```diff
@@ -1,12 +1,30 @@
...
+  savingStatus(saving) {
+    return saving ? I18n.t("saving") : I18n.t("save");
+  },
+
+  @computed("selectableUserBadges", "selectedUserBadgeId")
+  selectedUserBadge(selectableUserBadges, selectedUserBadgeId) {
+    selectedUserBadgeId = parseInt(selectedUserBadgeId);
+    let selectedUserBadge = null;
+
+    selectableUserBadges.forEach(function(userBadge) {
```
eviltrout: Should this keep looping after it's already found the id?

It seems to me we could use `return selectableUserBadges.findBy('id', selectedUserBadgeId)`

app/assets/javascripts/discourse/controllers/badges/show.js.es6
```diff
@@ -17,6 +17,19 @@ export default Ember.Controller.extend(BadgeSelectController, {
     return userBadgesAll.filterBy("badge.allow_title", true);
   },
 
+  @computed("filteredList")
+  selectableUserBadges(filteredList) {
+    filteredList = _.uniq(filteredList, false, function(e) {
```
eviltrout:
```suggestion
    filteredList = _.uniq(filteredList, false, e => {
```

eviltrout:
@CvX yes I'm cool to merge this when you are ready.

PR #8208, opened 2019-10-17
UX: emphasises on local-dates being UTC in excerpts

jjaffeux:


eviltrout:
Thanks! Gonna merge it now.

PR #8214, opened 2019-10-18
Support for importing `@ember/component` and `@ember/routing/route`

eviltrout:
This will allow us to take the first step in updating our modules to be consistent with example Ember code and to align with Ember CLI

### DIFF
--- a/app/assets/javascripts/discourse-loader.js
+++ b/app/assets/javascripts/discourse-loader.js
@@ -1,6 +1,15 @@
 var define, requirejs;
 
 (function() {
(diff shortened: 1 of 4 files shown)


jjaffeux:
Given how important is this commit, I wonder if we shouldn't do two commits:
- prettify the file
- the actual fix

we might enjoy having a clean history on this one day

eviltrout:
@jjaffeux  I moved the prettier/MOVED_MODULES thing into master separately. Can you please review again?

PR #8216, opened 2019-10-19
FIX: Add additional HTML5 media extensions to onebox audio and video tags

nachocab:
Modern browsers support more file extensions than the ones we're currently specifying ([source](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#Common_codecs)). 

For example, Ogg files usually have the `.ogg` extension, but can also have `.ogv` (for video) and `.oga` (for audio). 

Including these additional extensions ensures that valid media files will appear with an HTML audio/video player.


eviltrout:
This seems good to me, thank you!

PR #8225, opened 2019-10-21
DEV: Import DiscourseRoute rather than Discourse.Route

markvanlan:


eviltrout:
Thanks, let's bring it in!

PR #8228, opened 2019-10-22
WIP: When a category is muted, subcategory is as well

lis2:

Let's say we got parent category with id 1 and subcategory with id 2. When parent category is muted topics linked to subcategory are not hidden because we are only looking for CategoryUser rows with subcategory id and mute notification level.

We should ensure if both, subcategory and parent category are not muted.


eviltrout:
Oops I missed that it's closed.

PR #8229, opened 2019-10-22
DEV: prepares internet explorer deprecation

jjaffeux:
This core plugin, which could be removed in the future from core, allows us to load IE specific  code only when needed.


plugins/discourse-internet-explorer/plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      previous_discourse_stylesheet_link_tag(name, opts)
+    end
+  end
+
+  # not using patch on preload_script as js is fine and we need this file
+  # to be loaded before other files
+  register_html_builder('server:before-script-load') do |controller|
+    if controller.request.env['HTTP_USER_AGENT'] =~ /MSIE|Trident/
```
davidtaylorhq: I think we need a `plugin.enabled?` check here, or fix the `register_html_builder` function so that it checks whether the plugin is enabled.

davidtaylorhq:
This is changing the app output based on user agent, so I think we need to add an `is_ie` boolean to `AnonymousCache#cache_key`. Otherwise we risk serving IE polyfills to other browsers (not too bad), or missing out the IE polyfills for requests from IE (bad).

plugins/discourse-internet-explorer/plugin.rb
jjaffeux: any reason we don't want this to be added to `register_html_builder` ?

jjaffeux:
> This is changing the app output based on user agent, so I think we need to add an `is_ie` boolean to `AnonymousCache#cache_key`. Otherwise we risk serving IE polyfills to other browsers (not too bad), or missing out the IE polyfills for requests from IE (bad).

oh yes totally forgot about this discussion yesterday

plugins/discourse-internet-explorer/plugin.rb
davidtaylorhq: No reason. I think we should add it! (although it is more work than simply adding plugin.enabled?` in this plugin)

jjaffeux:
@davidtaylorhq ok should be better :+1:

lib/plugin/instance.rb
```diff
@@ -418,7 +418,11 @@ def register_custom_html(hash)
   end
 
   def register_html_builder(name, &block)
-    DiscoursePluginRegistry.register_html_builder(name, &block)
+    reloadable_patch do |plugin|
+      if plugin.enabled?
```
davidtaylorhq: I don't think this will work as you expect. It will only check `plugin.enabled?` when the app boots, and will only check the `default` database in a multisite cluster. The solution is to do something like [this](https://github.com/discourse/discourse/blob/master/app/models/user.rb#L267-L289). Store a reference to the plugin for each html_builder, and then check `plugin.enabled?` at runtime.

There are more details about this problem in [this commit message](https://github.com/discourse/discourse/commit/0d0d78841b02c9747796b6aa6cfe58c0d6eec457#diff-b1e56232326e1a7a433641f2c21261bd).

lib/middleware/anonymous_cache.rb
```diff
@@ -73,7 +91,7 @@ def is_crawler?
       end
 
       def cache_key
-        @cache_key ||= "ANON_CACHE_#{@env["HTTP_ACCEPT"]}_#{@env["HTTP_HOST"]}#{@env["REQUEST_URI"]}|m=#{is_mobile?}|c=#{is_crawler?}|b=#{has_brotli?}|t=#{theme_ids.join(",")}#{GlobalSetting.compress_anon_cache}"
+        @cache_key ||= "ANON_CACHE_#{@env["HTTP_ACCEPT"]}_#{@env["HTTP_HOST"]}#{@env["REQUEST_URI"]}|m=#{is_mobile?}|ie=#{is_ie?}|c=#{is_crawler?}|b=#{has_brotli?}|t=#{theme_ids.join(",")}#{GlobalSetting.compress_anon_cache}"
```
davidtaylorhq: Looks good, but I wonder if we should try and keep all ie-related logic in the plugin 🤔? On the other hand, putting it here is much more obvious, and less likely to cause confusion in future. 

If we keep it like this, I think we should add a comment on the `is_ie?` method explaining why it's there.

discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/dev-prepares-internet-explorer-deprecation-pr-8229/6366/12):

Yes I have been thinking the same... But I was also thinking at this tweet of the other day: I must resist to change everything my code is interacting with 😅

lib/plugin/instance.rb
discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/dev-prepares-internet-explorer-deprecation-pr-8229/6366/13):

This API seems very misleading then...

davidtaylorhq: I agree the whole situation is confusing, but which API is confusing? `register` happens when the app boots, so it is impossible for it to update the site setting live.

`reloadable_patch` is used for auto-reloading in development, it does nothing in production, and has nothing to do with multisite environments.

lib/middleware/anonymous_cache.rb
```diff
@@ -32,6 +33,23 @@ def blocked_crawler?
         CrawlerDetection.is_blocked_crawler?(@env[USER_AGENT])
       end
 
+      def is_ie=(val)
+        @is_ie = val ? :true : :false
```
eviltrout: Why are we using symbols for `true` and `false` here when we could use a boolean?

If it's because you want to memoize it, you can use https://jtway.co/how-to-memoize-false-and-nil-values-fe28b8ede9f8 `defined?`

jjaffeux: @eviltrout I'm using the patterns already established in this library:

https://github.com/discourse/discourse/blob/master/lib/middleware/anonymous_cache.rb#L36

eviltrout: @jjaffeux @SamSaffron has agreed to migrate away from that pattern in favor of `defined?`

jjaffeux:
For context here, @eviltrout has asked me to finish this PR himself.

eviltrout:
I'm closing this while I work on it in a new branch, thanks all!

plugins/discourse-internet-explorer/plugin.rb
eviltrout: Added this here: https://github.com/discourse/discourse/pull/8454

lib/middleware/anonymous_cache.rb
eviltrout: I've decided to make a PR for this: 

https://github.com/discourse/discourse/pull/8455

PR #8231, opened 2019-10-22
PERF: Add index on group to category_groups.

udan11:


SamSaffron:
confirmed, this is good to merge without the group_user index

The category group index is good!

PR #8233, opened 2019-10-22
DEV: Remove RSS feed polling in favor of plugin 

markvanlan:
Forked the wellfed repo into the Discourse organization, which covers all these features. https://github.com/discourse/discourse-wellfed


ZogStriP:
@markvanlan can you fix the conflict and merge the PR?

PR #8235, opened 2019-10-23
FEATURE: Add remembering topic list for group pms

markvanlan:
@eviltrout These routes are all so similar. Do you think there is an abstraction here that wouldn't make things more confusing that it would be worth?

### DIFF
--- a/app/assets/javascripts/discourse/models/topic-list.js.es6
+++ b/app/assets/javascripts/discourse/models/topic-list.js.es6
@@ -189,6 +189,20 @@ TopicList.reopenClass({
   // hide the category when it has no children
   hideUniformCategory(list, category) {
     list.set("hideCategory", !displayCategoryInList(list.site, category));
+  },
+
+  findOrResetCachedBy(filter) {
+    const session = Discourse.Session.current();
+    const lastTopicList = session.get("topicList");
+    if (lastTopicList && lastTopicList.filter === filter) {
+      return lastTopicList;
+    } else {
+      session.setProperties({
+        topicList: null,
+        topicListScrollPosition: null
+      });
+      return false;
+    }
   }
 });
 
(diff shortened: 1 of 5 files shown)


markvanlan:
> I do agree it could be DRYer. I feel like a helper function would handle this nicely?

What do you think about this? Helper inside topic-list

app/assets/javascripts/discourse/models/topic-list.js.es6
```diff
@@ -189,6 +189,20 @@ TopicList.reopenClass({
   // hide the category when it has no children
   hideUniformCategory(list, category) {
     list.set("hideCategory", !displayCategoryInList(list.site, category));
+  },
+
+  findOrResetCachedBy(filter) {
```
eviltrout: One thing that we've started to move to over time is not using class methods on Ember objects anymore. They were useful back in the day when JS had no modules, but now that any file can export a named function it's easier to make it a pure function and export it that way.

Also, it would be nice if you could pass `session` as a param, then use `this.session` in your instance. It should be available that way!

markvanlan: @eviltrout I put the function is a new file for cached topic list. I am really unsure where it should sit.

app/assets/javascripts/discourse/routes/build-private-messages-route.js.es6
```diff
@@ -18,19 +19,10 @@ export default (viewName, path, channel) => {
...
-      if (lastTopicList && lastTopicList.filter === filter) {
-        return lastTopicList;
-      } else {
-        session.setProperties({
-          topicList: null,
-          topicListScrollPosition: null
-        });
-        return this.store.findFiltered("topicList", { filter });
-      }
+      const lastTopicList = findOrResetCachedTopicList(this.session,filter);
```
eviltrout:
```suggestion
      const lastTopicList = findOrResetCachedTopicList(this.session, filter);
```

PR #8236, opened 2019-10-23
FIX: When running the wizard and using a custom theme, fallback to the color_scheme name if the base_scheme_id is nil

romanrizzi:
When re-running the wizard and using a custom theme, the logo step canvas is not working. This is caused by the `base_scheme_id` of the `color_scheme` being nil for some custom themes.

To fix this, this PR proposes to add the custom theme (set as the default) to the list of themes previewed in the `colors` step. Here, I'm using the hibiscus theme as my default:

Screen Shot 2019-10-23 at 18 29 07

When jumping to the logo step, the canvas is using the custom theme colors:

Screen Shot 2019-10-23 at 18 52 03

### DIFF
--- a/lib/wizard/builder.rb
+++ b/lib/wizard/builder.rb
@@ -146,12 +146,10 @@ class Wizard
         default_theme = Theme.find_by(id: SiteSetting.default_theme_id)
         default_theme_override = SiteSetting.exists?(name: "default_theme_id")
 
-        scheme_id =
-          if default_theme_override
-            default_theme&.color_scheme&.base_scheme_id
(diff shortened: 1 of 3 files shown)


lib/wizard/builder.rb
```diff
@@ -160,6 +158,13 @@ def build
           value: scheme_id
         )
 
+        if scheme_id && default_theme_override && base_scheme.nil?
```
eviltrout: I would add a simple comment here saying something like `# fix for the case when base_scheme is nil`

PR #8237, opened 2019-10-24
FIX: Restore for non-multisite is not raising an error on reconnect step

lis2:
That commit introduced a bug to the system: https://github.com/discourse/discourse/commit/f69dacf979317946ed4a3e81e85fcca1b45bedf0

Restore works fine for multisite, however, stopped working for non-multisite.

Reason for that was that `establish_connection` method got a check if the multisite instance is available:
```
    def self.instance
      @instance
    end

    def self.establish_connection(opts)
      @instance.establish_connection(opts) if @instance
    end
```
However, the reload method don't have that check
```
    def self.reload
      @instance = new(instance.config_filename)
    end
```

To solve it, let's ensure we are in a multisite environment before call reload

### DIFF
--- a/lib/backup_restore/restorer.rb
+++ b/lib/backup_restore/restorer.rb
@@ -404,7 +404,7 @@ module BackupRestore
 
     def reconnect_database
       log "Reconnecting to the database..."
-      RailsMultisite::ConnectionManagement::reload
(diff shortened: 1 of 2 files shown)


SamSaffron:
thanks ! looks good!

PR #8239, opened 2019-10-24
FIX: Hide muted subcategories

lis2:

Problem is that subcategories are always visible in `/categories` path even if muted.

Categories/subcategories are loaded in at least two places
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/site.js.es6#L146
https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/routes/discovery-categories.js.es6#L46

I discussed that with @jjaffeux and we thought that maybe it would be a good idea to filter that on frontend level

### DIFF
--- a/app/assets/javascripts/discourse/models/category.js.es6
+++ b/app/assets/javascripts/discourse/models/category.js.es6
@@ -3,6 +3,7 @@ import RestModel from "discourse/models/rest";
 import computed from "ember-addons/ember-computed-decorators";
 import { on } from "ember-addons/ember-computed-decorators";
 import PermissionType from "discourse/models/permission-type";
+import { NotificationLevels } from "discourse/lib/notification-levels";
 
 const Category = RestModel.extend({
   permissions: null,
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/models/category.js.es6
```diff
@@ -44,6 +45,11 @@ const Category = RestModel.extend({
     return { type: "category", id, category: this };
   },
 
+  @computed("notification_level")
+  isMuted(notification_level) {
```
jjaffeux: I will usually change this to `notificationLevel` when I can, we are receiving a "ruby" formatted value, but we are now in js world so I will transform it. I see it as a boundary between two world.

app/assets/javascripts/discourse/templates/components/categories-only.hbs
```diff
@@ -21,11 +21,13 @@
                    {{#if c.subcategories}}
               <div class='subcategories'>
                 {{#each c.subcategories as |s|}}
-                  <span class='subcategory'>
-                    {{category-title-before category=s}}
-                    {{category-link s hideParent="true"}}
-                    {{category-unread category=s}}
-                  </span>
+                  {{#unless s.isMuted}}
```
jjaffeux: that's not your code, but this "s" would be much better as "subcategory"

PR #8243, opened 2019-10-25
FIX: Reload plugin translations in development

udan11:
The default locale is :en_US, which is just a thin layer over :en. In
other words, :en_US has the :en locale as a fallback. When "en.yml" is
edited, only the :en locale is refreshed and :en_US becomes stale.

This commit ensures that there is a dependency on the fallback locales
too.


eviltrout:
OMG thank you!!!

PR #8244, opened 2019-10-25
FEATURE: Site setting/UI to allow users to set their primary group

markvanlan:
This adds the dropdown at the bottom of this screenshot.

Screenshot from 2019-10-25 11-20-02

### DIFF
--- a/app/assets/javascripts/discourse/controllers/preferences/account.js.es6
+++ b/app/assets/javascripts/discourse/controllers/preferences/account.js.es6
@@ -17,7 +17,7 @@ export default Controller.extend(CanCheckEmails, PreferencesTabController, {
   init() {
     this._super(...arguments);
 
-    this.saveAttrNames = ["name", "title"];
+    this.saveAttrNames = ["name", "title", "primary_group_id"];
     this.set("revoking", {});
   },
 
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/discourse/controllers/preferences/account.js.es6
```diff
@@ -54,6 +55,7 @@ export default Controller.extend(CanCheckEmails, PreferencesTabController, {
   },
 
   canSelectTitle: Ember.computed.gt("model.availableTitles.length", 0),
+  canSelectPrimaryGroup: Ember.computed.gt("model.filteredGroups.length", 0),
```
eviltrout: This could check the siteSetting too so the template doesn't need a nested if

lib/guardian.rb
```diff
@@ -307,6 +307,14 @@ def can_grant_title?(user, title = nil)
     user.groups.where(title: title).exists?
   end
 
+  def can_have_primary_group?(user, group_id=nil)
```
eviltrout: I think we should change this method name to something more clear. I suggest `can_use_primary_group` or `can_show_primary_group`.

markvanlan: Use is great. this naming tormented me.

PR #8249, opened 2019-10-26
DEV: simplify logic

nbianca:


nbianca:
I will add a better commit message in the merge commit.

PR #8250, opened 2019-10-26
DEV: Add test

nbianca:


nbianca:
Well, there is not much to say. It adds a test for some work @majakomel did in 2863e7c181843261a0d7b6c9255185a38b7f2680.

eviltrout:
Even that small change to the text has much more context and is appreciated.

PR #8253, opened 2019-10-28
FEAT: Remove attachments and truncate raw field for incoming emails

lis2:
We want to keep incoming emails however we want to limit the size of the raw message. To achieve that attachments are removed and long content is truncated. 

Dev: https://dev.discourse.org/t/introduce-a-sane-limit-to-size-of-incoming-email-storage/19623


app/jobs/scheduled/delete_rejected_emails.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DeleteRejectedEmails < ::Jobs::Scheduled
+    every 1.month
```
SamSaffron: I would just add this to the weekly task. I don't think it needs a dedicate task.

config/site_settings.yml
```diff
@@ -1051,6 +1051,9 @@ email:
     default: 0
     max: 51200
   post_excerpts_in_emails: false
+  raw_email_max_length: 220000
+  raw_rejected_email_max_length: 4000
+  delete_rejected_email_after_days: 180
```
SamSaffron: let's halve this to 90

lib/email/cleaner.rb
```diff
@@ -0,0 +1,37 @@
...
+      @mail.charset = 'UTF-8'
+      @remove_attachments = remove_attachments
+      @truncate = truncate
+      @rejected = rejected
+    end
+
+    def execute
+      @mail.without_attachments! if @remove_attachments
+      truncate! if @truncate
+      remove_null_byte(@mail.to_s)
```
SamSaffron: minor but you are copying this string 1 extra time here. 

```
def remove_null_byte!(message)
   message.gsub!(...)
   message
end
```

PR #8255, opened 2019-10-28
FIX: make notification consent banner usable via keyboard and screenr…

LeoMcA:
…eader

### DIFF
--- a/app/assets/javascripts/discourse/components/d-button.js.es6
+++ b/app/assets/javascripts/discourse/components/d-button.js.es6
@@ -12,7 +12,7 @@ export default Component.extend({
   type: "button",
 
   tagName: "button",
-  classNameBindings: [":btn", "noText", "btnType"],
+  classNameBindings: ["btnLink::btn", "btnLink", "noText", "btnType"],
   attributeBindings: [
     "form",
     "disabled",
@@ -23,6 +23,8 @@ export default Component.extend({
   ],
 
   btnIcon: notEmpty("icon"),
+
+  btnLink: Ember.computed.equal("display", "link"),
 
   @computed("icon", "translatedLabel")
   btnType(icon, translatedLabel) {
--- a/app/assets/javascripts/discourse/templates/components/notification-consent-banner.hbs
+++ b/app/assets/javascripts/discourse/templates/components/notification-consent-banner.hbs
@@ -2,7 +2,7 @@
   <div class="row">
     <div class="consent_banner alert alert-info">
       {{d-button icon="times" action="dismiss" class="btn btn-flat close" title="banner.close"}}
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/discourse/templates/components/notification-consent-banner.hbs
```diff
@@ -1,10 +1,10 @@
 {{#if showNotificationPromptBanner}}
   <div class="row">
     <div class="consent_banner alert alert-info">
-      <div class="close" {{action "dismiss"}}>
+      <button class="btn-link close" aria-label="{{i18n 'banner.close'}}" {{action "dismiss"}}>
         {{d-icon 'times'}}
-      </div>
-      {{i18n 'user.desktop_notifications.consent_prompt'}} <a {{action "turnon"}}>{{i18n 'user.desktop_notifications.enable'}}</a>.
+      </button>
+      {{i18n 'user.desktop_notifications.consent_prompt'}} <button class="btn-link" {{action "turnon"}}>{{i18n 'user.desktop_notifications.enable'}}</button>.
```
eviltrout: Can we use `d-button` for these buttons?

app/assets/javascripts/discourse/components/d-button.js.es6
```diff
@@ -23,6 +23,11 @@ export default Component.extend({
 
   btnIcon: Ember.computed.notEmpty("icon"),
 
+  @computed("class")
```
eviltrout: This is a bit odd to me - is the idea where that you don't want `btn` if `btn-link` is present?

I think checking `class` is awkward for a couple reasons. It is a reserved word and it is confusing to pair it with classNameBinding because you get somewhat of a circular relationship.

Instead it would be better I think to pass type="link" and use that property to remove `btn` and andd `btn-link`

LeoMcA: Yep, that's the idea: using `<a>` tags as buttons is bad for accessibility (you have to explicitly give it a role of button, and even then some browsers won't allow keyboard interaction with it), so it's far better to use an actual `<button>` but then style it as a link. So `btn-link` does that styling and I don't want the `btn` class intefering.

Yeah, I hadn't caught that we have a circular dependency here - I'm amazed ember didn't blow up...

LeoMcA: I decided to use a new `display` attribute rather than using `type` and then having to deal with setting that to `type="button"` in the actual rendered `<button>`

LeoMcA:
@awesomerobot it looks like (https://github.com/discourse/discourse/commit/bacc1144627f73f15650fab1df9c14277b2ea5b8) you've just tacked the issue of all these close icons being non-semantic buttons.

What are your thoughts on this PR as a first step at converting all the remaining `<a ... {{action`s that exist?

LeoMcA:
Merge conflict sorted :)

eviltrout:
Thank you!

PR #8256, opened 2019-10-28
DEV: Import EmberObject rather than global variable

markvanlan:


app/assets/javascripts/discourse/controllers/reorder-categories.js.es6
```diff
@@ -1,4 +1,4 @@
-import ObjectProxy from '@ember/object/proxy';
```
CvX: [The recommended import](https://api.emberjs.com/ember/3.13/classes/ObjectProxy) is the one you did first, so it may be a good idea to revert this change and to add a new entry to [the loader](https://github.com/discourse/discourse/blob/e7ff6809a39bbaafb5d0b88b23ec330f5233ccf4/app/assets/javascripts/discourse-loader.js#L7-L13) 🙂

PR #8258, opened 2019-10-28
FIX: flaky jobs_base_spec.rb

lis2:
I was searching for a reason for randomly failing jobs_base_spec.rb. The reason was that after restorer_spec, the database is not restored to default.
After restorer spec RailsMultisite::ConnectionManagement.all_dbs is rstill eturning array of `['default', 'second']`

Then base job execution is evaluated twice
```
dbs = RailsMultisite::ConnectionManagement.all_dbs
dbs.each do |db|
  execute(opts)
end
```


SamSaffron:
nice catch!

PR #8262, opened 2019-10-29
FIX: Account for empty uploads directory upon backup restore

justindirose:
Worked with @romanrizzi and @danielwaterworth on this one. We weren't factoring in cases where the uploads directory would be completely empty, such as when we're running an import without attachments. 

This commit simply checks for an empty directory and stops the process if this is the case to avoid restore failure.


SamSaffron:
I think this is safe @gschlager , shall we merge?

lib/backup_restore/restorer.rb
```diff
@@ -436,6 +436,7 @@ def extract_uploads
         FileUtils.mkdir_p("uploads")
 
         tmp_uploads_path = Dir.glob(File.join(@tmp_directory, "uploads", "*")).first
+        return unless tmp_uploads_path.present?
```
ZogStriP: When possible, we usually prefer the positive condition instead of the negative counterpart which is not always clear at first sight.

```suggestion
        return if tmp_uploads_path.blank?
```

justindirose: Makes sense. 

https://github.com/discourse/discourse/blob/b39a2c3181024081a5a836d647d40dd1a44fdbf7/lib/backup_restore/restorer.rb#L430

I was following the pattern already used in the block, but happy to change it!

PR #8263, opened 2019-10-30
FIX: Require q param in /tags/filter/search route

martin-brennan:
Fixes this log error: https://dev.discourse.org/logs/show/4e66b724b8133292a3f94e4a59a3caa4

* if not provided this route was causing a 500 error when
  DiscourseTagging.clean_tag was called, because .downcase
  was being called on the param (which was nil)
* this was detected in the logs, and upon further investigation
  all instances were occurring via community.nash.io
* nowhere in the frontend appeared to allow this behaviour; all
  calls to the tag search always provided the q param, so suspected
  direct call to the endpoint. now return a 400 error instead

### DIFF
--- a/spec/requests/tags_controller_spec.rb
+++ b/spec/requests/tags_controller_spec.rb
@@ -428,6 +428,16 @@ describe TagsController do
         json = ::JSON.parse(response.body)
         expect(json["results"].map { |j| j["id"] }).to eq(['тема-в-разработке'])
       end
+
+      context 'when tag query parameter is not provided' do
+        it 'does not cause a 500 error, returns a param required message' do
+          get "/tags/filter/search.json", params: {}
+          expect(response.status).not_to eq(500)
+          expect(response.status).to eq(400)
+          json = ::JSON.parse(response.body)
+          expect(json['errors'].first).to eq('param is missing or the value is empty: q')
+        end
+      end
     end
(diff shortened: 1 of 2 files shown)


spec/requests/tags_controller_spec.rb
```diff
@@ -428,6 +428,16 @@ def parse_topic_ids
         json = ::JSON.parse(response.body)
         expect(json["results"].map { |j| j["id"] }).to eq(['тема-в-разработке'])
       end
+
+      context 'when tag query parameter is not provided' do
+        it 'does not cause a 500 error, returns a param required message' do
+          get "/tags/filter/search.json", params: {}
+          expect(response.status).not_to eq(500)
```
SamSaffron: This line is not really needed.

spec/requests/tags_controller_spec.rb
```diff
@@ -428,6 +428,16 @@ def parse_topic_ids
...
+          expect(response.status).to eq(400)
+          json = ::JSON.parse(response.body)
+          expect(json['errors'].first).to eq('param is missing or the value is empty: q')
```
SamSaffron: this can be a bit fragile

SamSaffron:
looks good to me, improve the 2 little comments and merge away.

PR #8273, opened 2019-10-31
FIX: Hide muted subcategories for mobile

lis2:

Fix was done in that PR https://github.com/discourse/discourse/pull/8239
however I forgot about mobile template


lis2:
Hey @jjaffeux, could you take a look?

PR #8277, opened 2019-10-31
DEV: Adds a new plugin outles to add custom icons on the admin users list

romanrizzi:
Akismet will use this outlet to show if the user was skipped, checked, etc.


romanrizzi:
Done!

PR #8280, opened 2019-10-31
FEATURE: Create New Topic button on embed with params

markvanlan:
This allows embed urls to take the `allow_create` param, and add a `Create New Topic` button, that will open the composer in a new tab with category, and tags passed into the composer.

You will notice that this transforms the `category` param into `category_id` for the link to the composer. This is just how the params are accepted by each endpoint.

Screenshot from 2019-11-01 13-27-41

### DIFF
--- a/app/assets/stylesheets/embed.scss
+++ b/app/assets/stylesheets/embed.scss
@@ -195,8 +195,8 @@ div.lightbox-wrapper {
     height: 16px;
     margin-right: 2px;
     background-image: svg-uri(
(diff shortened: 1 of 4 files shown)


app/views/embed/topics.html.erb
```diff
@@ -1,4 +1,12 @@
 <%- if @list && @list.topics.present? %>
+  <%- if @allow_create %>
+    <%= link_to @create_url, target: "_blank" do |link| %>
+      <button class=<%="new-topic-btn" + (@template == "complete" ? " complete" : "") %>>
+        <span class=<%="new-topic-btn__icon" if @template == "complete" %>></span>
+        <span class=<%="new-topic-btn__text" if @template == "complete" %>>Create New Topic!</span>
```
eviltrout: This needs to be in the server i18n file.

Also the standard for Discourse is to not use an exclamation point

app/assets/stylesheets/embed.scss
```diff
@@ -176,6 +176,20 @@ div.lightbox-wrapper {
   margin-bottom: 20px;
 }
 
+.new-topic-btn {
```
eviltrout: This should match the default Discourse styles for primary buttons (blue background, + icon)

PR #8282, opened 2019-11-01
FIX: Badge and user title interaction fixes

martin-brennan:
Original meta thread: https://meta.discourse.org/t/engagement-level-x-badge-nucommunity/129514
Dev thread: https://dev.discourse.org/t/badge-titles-not-revoked-when-badge-name-is-customized/16978

The issue described here is that `BadgeGranter#revoke_ungranted_titles!` was not taking into account system badges with custom text set via `TranslationOverride`, and thus when a badge was revoked that was being used for a user title, the user still retained that title. There are a few different fixes in this PR:

* When the user title is changed (manually or setting it via using the `badge_title` route in `UsersController`), we now consider `badge_granted_title` to be true on `UserProfile` if the title matches either the badge name OR the custom `TranslationOverride` text for the badge
* Introduces a new `granted_title_badge_id` column to `UserProfile` which is set at the same time as above, so we have a record of which badge granted the user title
* Added a migration to backfill `granted_title_badge_id` for all users who have a title based on a badge name OR based on translation override text for a badge
* When a badge name is changed, or when custom text is added/revoked for a system badge, we find all user profiles using that badge for a title via `granted_title_badge_id` and update the user's title to match the new text
* When a badge is revoked if the user title matches the badge name OR custom text then we revoke the user's title
* Finally, added a `change_title` and a `revoke_title` `UserHistory` action and corresponding staff log, to keep better track of when and where title changes occur

### Notes

* I did intend to split the code into `BulkUserTitleUpdater` into a background job, but I'm now leaning towards saying its probably going to be unlikely that these will be hugely slow queries to run. Thoughts?
* The `BulkUserTitleUpdater` does not add individual `UserHistory` records for each change_title action it carries out, I'm not sure if this is a problem though
* The migration has only been tested on a _very_ small amount of users locally


SamSaffron:
@vinothkannans would you mind reviewing this?

db/migrate/20191101055132_backfill_granted_title_badge_id_for_user_profile.rb
```diff
@@ -0,0 +1,36 @@
...
+class BackfillGrantedTitleBadgeIdForUserProfile < ActiveRecord::Migration[6.0]
+  def up
+    # update all the regular badge derived titles based
+    # on the normal badge name
+    ActiveRecord::Base.connection.execute <<-SQL
+      UPDATE user_profiles
+      SET granted_title_badge_id = b.id
+      FROM users
+      INNER JOIN badges b ON users.title = b.name
+      WHERE users.title = b.name
```
vinothkannans: Why we checking the `users.title = b.name` both in inner join condition and where condition?

db/migrate/20191101055132_backfill_granted_title_badge_id_for_user_profile.rb
```diff
@@ -0,0 +1,36 @@
...
+        AND user_profiles.granted_title_badge_id IS NULL
+        AND users.id = user_profiles.user_id
+        AND users.title IS NOT NULL AND users.title != ''
```
vinothkannans: I think we don't have to check the `users.title` here again, since we already checked it like `users.title = b.name`. If it's empty it won't even satisfy the existing inner join condition.

spec/models/user_spec.rb
```diff
@@ -1997,12 +1997,35 @@ def filter_by(method)
       expect(user.user_profile.reload.badge_granted_title).to eq(false)
 
       badge.update!(allow_title: true)
+      user.badges.reload
```
vinothkannans: We don't need to reload the badges here. Since we only set the title of user.

db/migrate/20191101055132_backfill_granted_title_badge_id_for_user_profile.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+class BackfillGrantedTitleBadgeIdForUserProfile < ActiveRecord::Migration[6.0]
```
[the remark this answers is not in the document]
martin-brennan: No problem, will do that.

martin-brennan:
@vinothkannans thanks for the review. For this last point:

> I think it's okay to put it in a background job. Since only the limited number of users who use a badge's name as title is going to be updated.

Did you mean to say it's okay to NOT put it in a background job? If you meant for me to move it into a background job that's fine and I will do so :)

spec/models/user_spec.rb
martin-brennan: I do need to do that because the spec fails without doing so. The `User#check_if_title_is_badged_granted` method uses ruby `Enumerable#find` to save trip to the DB https://devdocs.io/ruby~2.6/enumerable#method-i-find.

```
badge_matching_title = title.present? && badges.find do |badge|
  badge.allow_title? && (badge.display_name == title || badge.name == title)
end
```

Because the user's `badges` AR association is already loaded it will not query the DB again to fetch the updated `allow_title`. I can alternatively call `user.badges.find { |b| b.id == badge.id }.update!(allow_title: true)` but I think that's much of a muchness really. Or I can change the User model code to fetch the badges from the db instead of using `badges.find`. Is there a downside to reloading that I'm not aware of?

martin-brennan: @vinothkannans

martin-brennan:
@vinothkannans made all the changes except for my comments left about the background job & reloading user badges.

vinothkannans:
> Did you mean to say it's okay to NOT put it in a background job?

No, I said that we can put it in a background job. @martin-brennan

martin-brennan:
> > Did you mean to say it's okay to NOT put it in a background job?
> 
> No, I said that we can put it in a background job. @martin-brennan

No problem, sorry I just read your initial comment incorrectly and confused myself! :sweat_smile:

app/models/user.rb
```diff
@@ -1478,8 +1478,14 @@ def trigger_user_updated_event
...
     if title_changed? && !new_record? && user_profile
-      badge_granted_title = title.present? && badges.where(allow_title: true, name: title).exists?
-      user_profile.update_column(:badge_granted_title, badge_granted_title)
+      badge_matching_title = title.present? && badges.find do |badge|
+        badge.allow_title? && (badge.display_name == title || badge.name == title)
+      end
+      user_has_badge_like_title = badge_matching_title.present?
+      user_profile.update(
+        badge_granted_title: user_has_badge_like_title,
+        granted_title_badge_id: user_has_badge_like_title ? badge_matching_title.id : nil
```
vinothkannans: Super minor but you can do `badge_matching_title&.id` here instead of inline if.

martin-brennan: No problem. We had banned the safe navigation operator at my old job so I need to get myself in the habit of using it

martin-brennan:
@vinothkannans have moved the updater into a job now.

PR #8284, opened 2019-11-01
FEATURE: Overhaul of admin API key system

davidtaylorhq:
- Allow revoking keys without deleting them
- Auto-revoke keys after a period of no use (default 6 months)
- Allow multiple keys per user
- Allow attaching a description to each key, for easier auditing
- Log changes to keys in the staff action log
- Move all key management to one place, and improve the UI


app/assets/javascripts/admin/controllers/admin-api-keys-show.js.es6
```diff
@@ -0,0 +1,57 @@
+import { bufferedProperty } from "discourse/mixins/buffered-content";
+import { default as computed } from "ember-addons/ember-computed-decorators";
+import { popupAjaxError } from "discourse/lib/ajax-error";
+
+export default Ember.Controller.extend(bufferedProperty("model"), {
+  @computed("model.id")
```
eviltrout: This could be `isNew: empty('model.id')`

app/assets/javascripts/admin/templates/user-index.hbs
```diff
@@ -292,33 +292,15 @@
...
-            icon="key"
-            label="admin.api.generate"}}
-        </div>
-      {{/if}}
+      <div class="field">{{i18n "admin.api.active_keys"}}</div>
+      <div class="value">
+        {{model.api_key_count}}
+      </div>
+      <div class="controls">
+        {{#link-to "adminApiKeys" class="btn btn-default"}}
```
eviltrout: `{{d-button}}` supports links and that's the preferred way to do this now.

app/controllers/admin/api_controller.rb
```diff
@@ -1,30 +1,92 @@
...
 
   def index
-    render_serialized(ApiKey.where(hidden: false).to_a, ApiKeySerializer)
+    keys = ApiKey.where(hidden: false)
+
+    # Put active keys first
+    # Sort active keys by created_at, sort revoked keys by revoked_at
+    keys = keys.order(<<~SQL)
+      CASE WHEN revoked_at IS NULL THEN 0 ELSE 1 END,
+      CASE WHEN revoked_at IS NULL THEN created_at ELSE revoked_at END DESC
```
eviltrout: This could be `COALESCE(revoked_at, created_at)` I think

app/models/api_key.rb
```diff
@@ -4,23 +4,44 @@ class ApiKey < ActiveRecord::Base
...
+
+  def self.last_used_epoch
+    SiteSetting.api_key_last_used_epoch.presence
   end
 
+  def self.revoke_unused_keys!
+    return if SiteSetting.revoke_api_keys_days == 0 # Never expire keys
+    to_revoke = active.where("GREATEST(last_used_at, created_at, updated_at, :epoch) < :threshold",
+                  epoch: last_used_epoch,
+                  threshold: Time.zone.now - SiteSetting.revoke_api_keys_days.days
```
eviltrout:
```suggestion
                  threshold: SiteSetting.revoke_api_keys_days.days.ago
```

app/services/staff_action_logger.rb
```diff
@@ -663,6 +663,29 @@ def log_embeddable_host(embeddable_host, action, opts = {})
...
 
+  def log_api_key(api_key, action, opts = {})
+    opts[:changes]&.delete("key") # Do not log the full key
+
+    history_params = params(opts).merge(
+      action: action,
+      subject: api_key.truncated_key
+    )
+
+    if opts[:changes]&.keys == ["revoked_at", "updated_at"]
```
eviltrout: This code block is quite confusing. Is there a more straightforward way to figure out if the key is being revoked/restored than inspecting the changes hash?

db/migrate/20191101113230_add_revoked_at_to_api_key.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class AddRevokedAtToApiKey < ActiveRecord::Migration[5.2]
+  def change
+    add_column :api_keys, :revoked_at, :datetime
+    add_column :api_keys, :description, :text
+
+    execute "INSERT INTO site_settings(name, data_type, value, created_at, updated_at)
```
SamSaffron: technically I am not sure this is correct given this is a reversable migration, maybe just define the up and throw an exception on down?

app/services/staff_action_logger.rb
davidtaylorhq: You're right, this method got unnecessarily complicated. I've removed this logic and created separate functions for logging revoke/restore.

PR #8285, opened 2019-11-01
FIX: Allow theme translations to be accessed in initializers

davidtaylorhq:
Previously theme translations were loaded along with other plugin API scripts. These run after pre-initializers and initializers when the app boots. This commit moves theme translation loading into pre-initializers, so their behaviour matches core translations more closely.


davidtaylorhq:
Note to self: bump plugin API version before merging

PR #8288, opened 2019-11-02
FIX: prevents whitelisted_generic_onebox_spec to fail with zeitwerk

jjaffeux:


spec/components/onebox/engine/whitelisted_generic_onebox_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'oneboxer'
```
danielwaterworth: @jjaffeux, I see that this fixes the issue, but I don't quite understand why it's necessary. Why isn't the lazy loading working and could a similar thing happen outside of tests?

jjaffeux: I'm unsure. I mostly realised that this was not called.

My only VERY wild guess is that `oneboxer` is defined as a module and I have seen this on SO: https://stackoverflow.com/questions/57277351/rails-6-with-zeitwerk-doesnt-load-class-from-module

🤷‍♂ 

maybe we should ask to @lis2 ?

lis2: @danielwaterworth @jjaffeux this is interesting, I took a quick look at what may be causing that. 

We are overriding `WhitelistedGenericOnebox` with
```
  def self .===  
  end
  def self.priority
     Float::INFINITY
  end
```

I checked what is happening when we remove require "oneboxer"
`described_class.priority # 100`

So it falls back https://github.com/discourse/onebox/blob/master/lib/onebox/engine.rb#L95 
which means that Zeitwerk finds this class and is satisfied https://github.com/discourse/onebox/blob/master/lib/onebox/engine/whitelisted_generic_onebox.rb 

It sounds like correct behaviour to me, I guess that I was too aggressive with Zeitwerk PR when I was removing `require_dependency` blocks

PR #8290, opened 2019-11-03
FEATURE: Load translation overrides without JS `eval`

gschlager:
Removes translation overrides from the `PreloadStore` and loads them via `ExtraLocalesController` instead.

* There's no need for using `eval` for MessageFormat strings on the client anymore, because they are already compiled on the server.
* It's multisite aware, so `ExtraLocalesController` caches the hashes for the `overrides` bundle by site.
* The `<script>` tag for the `overrides` bundle is only included when there are client-side overrides.

Other notable changes:
* Bundles are cached for 1 year instead of just 24 hours. When the content of a bundle changes, its hash (and the `v` parameter) change as well, so we should be good.
* I removed the distinction between development and production mode. Bundles are always served with the `v` parameter which makes debugging easier.

### DIFF
--- a/app/assets/javascripts/discourse/initializers/localization.js.es6
+++ b/app/assets/javascripts/discourse/initializers/localization.js.es6
@@ -1,5 +1,3 @@
-import PreloadStore from "preload-store";
-
 export default {
   name: "localization",
   after: "inject-objects",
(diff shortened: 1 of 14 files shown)


gschlager:
PR updated to make it work with multisite and some other smaller fixes and more specs.

PR #8294, opened 2019-11-04
DEV: use native Promise instead of Ember.RSVP.Promise

markvanlan:
This is dependent on adding a polyfill for Promise in IE. Which is not added


SamSaffron:
I am loving this change but we got to polyfill ie for now.

PR #8295, opened 2019-11-05
FEATURE: support to mute all categories by default.

vinothkannans:


lib/topic_query.rb
```diff
@@ -869,7 +869,34 @@ def remove_muted_topics(list, user)
...
+        list = list.references("cu")
+          .where("
+          NOT EXISTS (
+            SELECT 1
+              FROM categories c
+              LEFT OUTER JOIN category_users cu
+              ON c.id = cu.category_id AND cu.user_id = :user_id
+            WHERE c.id = topics.category_id
+              AND c.id <> :category_id
+              AND (cu.notification_level IS NULL OR cu.notification_level = :muted)
```
eviltrout: Small tip here. You can use `COALESCE`:

`AND (COALESCE(cu.notification_level, :muted) = :muted)`

(same for the next line.)

lib/topic_query.rb
```diff
@@ -869,7 +869,34 @@ def remove_muted_topics(list, user)
...
+          )", user_id: user.id,
+              muted: CategoryUser.notification_levels[:muted],
+              tracking: TopicUser.notification_levels[:tracking],
+              category_id: category_id || -1)
+      else
+        category_ids = [
+          SiteSetting.default_categories_watching.split("|"),
+          SiteSetting.default_categories_tracking.split("|"),
+          SiteSetting.default_categories_watching_first_post.split("|")
+        ].flatten.map(&:to_i)
```
eviltrout: Do we need to call `uniq` here? In case there are duplicates there's no point in passing them to the DB.

vinothkannans:
@eviltrout I pushed a fix and added the `COALESCE` in SQL.

config/locales/server.en.yml
```diff
@@ -2066,6 +2066,7 @@ en:
     default_categories_tracking: "List of categories that are tracked by default."
     default_categories_muted: "List of categories that are muted by default."
     default_categories_watching_first_post: "List of categories in which first post in each new topic will be watched by default."
+    mute_all_categories_by_default: "Set the default notification level of all the categories to muted."
```
SamSaffron: I think we need a better description:

"Require users opt-in to tracking and watching categories for them to appear in latest and /categories page. If you wish to amend the defaults for anonymous users set default_categories_tracking and default_categories_watching"

Something like that so it is a bit clearer what it does.

SamSaffron:
overall I am happy with this! cc @tgxworld

config/locales/server.en.yml
vinothkannans: The categories will appear even if they opt-in for "normal" or "watching first post". And the same for `default_categories_watching_first_post` site setting. So I'm changing it like below. Is it looks good?

> Set the default notification level of all the categories to muted. Require users opt-in to categories for them to appear in 'latest' and 'categories' pages. If you wish to amend the defaults for anonymous users set 'default_categories_' settings.

PR #8301, opened 2019-11-06
FEATURE: Implement support for IMAP and SMTP email protocols.

udan11:
Co-authored-by: Joffrey JAFFEUX <j.jaffeux@gmail.com>


app/assets/javascripts/discourse/components/groups-form-email-fields.js.es6
```diff
@@ -0,0 +1 @@
+export default Ember.Component.extend({});
```
eviltrout: Can you import `@ember/component` please? It's our new standard way of doing this.

Also super minor but you can `Component.extend()` without the empty object.

app/assets/javascripts/discourse/routes/group-manage-email.js.es6
```diff
@@ -0,0 +1,7 @@
+export default Discourse.Route.extend({
```
eviltrout: This can also be imported from `discourse/routes/discourse`

app/jobs/regular/group_smtp_email.rb
```diff
@@ -0,0 +1,29 @@
...
+module Jobs
+  class GroupSmtpEmail < ::Jobs::Base
+    sidekiq_options queue: 'critical'
+
+    def execute(args)
+      group = Group.find_by(id: args[:group_id])
+      post = Post.find_by(id: args[:post_id])
+      email = args[:email]
+
+      Rails.logger.debug("[IMAP] Sending email for #{group.name} (#{group.id}) and post #{post.id}.")
```
eviltrout: Did we mean to leave this in? Usually debugging information like this behind some kind of flag or setting.

lib/imap/sync.rb
```diff
@@ -0,0 +1,223 @@
...
+        emails = @provider.emails(@group.imap_mailbox_name, new_uids, ['UID', 'FLAGS', 'LABELS', 'RFC822'])
+        emails.each do |email|
+          # Synchronously process emails because the order of emails matter
+          # (for example replies must be processed after the original email
+          # to have a topic where the reply can be posted).
+          begin
+            receiver = Email::Receiver.new(email['RFC822'],
+              allow_auto_generated: true,
+              import_mode: import_mode,
+              destinations: [{ type: :group, obj: @group }],
```
eviltrout: `obj` here is fairly generic. I would prefer `group:` or `target:` to be more consistent with the rest of our codebase.

lib/imap/providers/generic.rb
```diff
@@ -0,0 +1,112 @@
...
+            if tag = to_tag(name)
+              labels[tag] = name
+            end
+          end
+
+          labels
+        end
+      end
+
+      def open_mailbox(mailbox_name, write = false)
```
eviltrout: I dislike boolean parameters like this, because when you read the code at the calling point it looks like this:

`open_mailbox('my_mailbox', true)` and it's totally unclear what `true` means.

Instead we should use a keyword argument like `open_mailbox('my_mailbox', write: true)` which is self documenting.

app/jobs/scheduled/sync_imap_idle.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    every 30.seconds
+    queue :imap_idle
+
+    def execute(args)
+      return if !SiteSetting.enable_imap || !SiteSetting.enable_imap_idle || !$redis.set(LOCK_KEY, 1, ex: 60.seconds, nx: true)
+
+      @running = true
+      @syncs = {}
+      @syncs_lock = Mutex.new
```
eviltrout: Personally I am not a fan of using threads, as historically they have been a source of hard to track down bugs and are very hard to get right. Having said that, if this was discussed outside of this PR and this was the conclusion that is OK, but I would like someone else to review it who has more experience such as @SamSaffron.

lib/imap/providers/gmail.rb
```diff
@@ -0,0 +1,148 @@
+# frozen_string_literal: true
+
+module Imap
+  module Providers
+    class Gmail < Generic
+      X_GM_LABELS = 'X-GM-LABELS'
+
+      def imap
+        @imap ||= begin
```
eviltrout: Might be able to simplify this: `@imap ||= super.tap { |imap| apply_gmail_patch(imap) }`

lib/imap/providers/gmail.rb
```diff
@@ -0,0 +1,148 @@
...
+        return '\\Important' if tag == 'important'
+        return '\\Starred' if tag == 'starred'
+
+        super(tag)
+      end
+
+      private
+
+      def apply_gmail_patch(imap)
+        class << imap.instance_variable_get('@parser')
```
eviltrout: The logic is a little hard to follow so can you clarify something for me here? Does this patch the `Imap` class, so that every instance of the imap will now be patched for gmail? Or does it only patch the specific instance?

app/jobs/regular/group_smtp_email.rb
udan11: I believe that production instances usually have logging level set to `>= warn` so it is not an issue and is super handy in development.

lib/imap/providers/gmail.rb
udan11: Yes, it opens the metaclass (/ eigenclass / singleton class) of the parser instance to add support for Gmail's IMAP extensions (`X-GM-*` attributes).

lib/imap/sync.rb
udan11: I agree that `obj` is generic, but I believe that is how it is supposed to be, because `type` specifies what `obj` is.

I think I can just pass `@group` and perform a type check with `is_a?`, but will have to do a quick benchmark first because I have a feeling it will be a lot slower.

discoursereviewbot: [Robin Ward posted](https://review.discourse.org/t/feature-implement-support-for-imap-and-smtp-email-protocols-pr-8301/6748/15):

Sure, that's fine with me.

config/routes.rb
```diff
@@ -540,6 +540,7 @@
       get 'mentionable'
       get 'messageable'
       get 'logs' => 'groups#histories'
+      get 'mailboxes'
```
davidtaylorhq, line 543: Is this being used? I don't see a controller action for it 🤔

config/unicorn.conf.rb
```diff
@@ -94,77 +94,131 @@
         STDERR.puts "#{Time.now}: Issuing stop to sidekiq"
         Demon::Sidekiq.stop
       end
+    end
 
-      class ::Unicorn::HttpServer
-        alias :master_sleep_orig :master_sleep
+    puts "Starting up email sync"
```
davidtaylorhq, line 99: Just checking... is it safe to run multiple `Demon::EmailSync` at the same time? It looks like there is one Demon per host, which means we will have multiple instances running against the same database.

lib/demon/email_sync.rb
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  def suppress_stderr
+    false
+  end
+
+  def start_thread(db, group)
+    Thread.new do
+      RailsMultisite::ConnectionManagement.with_connection(db) do
+        begin
```
davidtaylorhq, line 26: Should we check `SiteSetting.imap_enabled` here?

app/serializers/basic_group_serializer.rb
```diff
@@ -55,6 +68,62 @@ def include_automatic_membership_email_domains?
     scope.is_admin?
   end
 
+  def include_smtp_server?
```
ZogStriP, line 71: Can you instead add a method `admin_attributes` which would be similar to [`staff_attributes`](https://github.com/discourse/discourse/blob/1c0c425f38616e801b5b917d3c95706f9a2a8e5c/app/serializers/user_card_serializer.rb#L6-L13) in order to reduce all these lines into

```ruby
admin_attributes smtp_server, smtp_port, ....
```

lib/post_creator.rb
```diff
@@ -186,10 +186,8 @@ def create
         @post.link_post_uploads
         update_uploads_secure_status
         ensure_in_allowed_users if guardian.is_staff?
-        unarchive_message
-        if !@opts[:import_mode]
-          DraftSequence.next!(@user, draft_key)
-        end
+        unarchive_message if !@opts[:import_mode]
```
ZogStriP, line 189: Is this fixing a bug? Might be worth extracting into another (and much smaller) PR.

udan11: This was a problem during the initial synchronization of a mailbox which would create a lot of posts and would sometime reset the archived status of a topic.

PR #8303, opened 2019-11-06
DEV: Allow execute_command to receive a block

davidtaylorhq:
This makes it easy to run multiple commands with the same keyword arguments. The main use is for using `chdir` across multiple commands. The `Dir.chdir` method is not concurrency safe, as it switches the working directory of the entire process.

### DIFF
--- a/lib/discourse.rb
+++ b/lib/discourse.rb
@@ -23,19 +23,51 @@ module Discourse
   end
 
   class Utils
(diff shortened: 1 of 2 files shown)


SamSaffron:
looks great, merge at will, but please add a test that spins say 5 threads and then runs pwd and has chdir making sure this takes care of working directory bleed.

PR #8305, opened 2019-11-06
DEV: Import ember ENV instead of Ember.testing

markvanlan:


app/assets/javascripts/admin/models/api-key.js.es6
```diff
@@ -2,9 +2,10 @@ import AdminUser from "admin/models/admin-user";
 import RestModel from "discourse/models/rest";
 import { ajax } from "discourse/lib/ajax";
 import computed from "ember-addons/ember-computed-decorators";
+import emberComputed from "@ember/object/computed";
```
eviltrout: I think the computed one will have to be a bigger PR discussion

PR #8306, opened 2019-11-06
FIX: downcase SSO external email before checking against Discourse email

scossar:
Automatic group membership based off email domains is failing for sites using SSO when the email address contains uppercase letters. This is because the `email_confirmed?` method is checking the SSO record's `external_email` against the user's normalized Discourse email.

If we don't want to make this change, another option would be to disable automatic group membership based off email domains for site's using SSO.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -908,7 +908,7 @@ class User < ActiveRecord::Base
   def email_confirmed?
     email_tokens.where(email: email, confirmed: true).present? ||
     email_tokens.empty? ||
-    single_sign_on_record&.external_email.downcase == email
+    single_sign_on_record&.external_email&.downcase == email
   end
 
   def activate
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -908,7 +908,7 @@ def username_format_validator
   def email_confirmed?
     email_tokens.where(email: email, confirmed: true).present? ||
     email_tokens.empty? ||
-    single_sign_on_record&.external_email == email
+    single_sign_on_record&.external_email&.downcase == email
```
eviltrout: is `email` already downcased here? I wonder if it should be 

`single_sign_on_record&.external_email&.downcase == email.downcase`

scossar: Yes, `email` is downcased there. My understanding is that user emails are always downcased before being stored to the database. They are downcased by the `strip_downcase_email` `before_validation` filter.

PR #8308, opened 2019-11-07
FEATURE: remove support for 'suppress_from_latest' category setting.

vinothkannans:
We are removing this in favor of the PR https://github.com/discourse/discourse/pull/8295.


app/models/category.rb
```diff
@@ -656,10 +655,6 @@ def clear_url_cache
     @@url_cache.clear
   end
 
-  def clear_featured_cache
-    CategoryFeaturedTopic.clear_exclude_category_ids
-  end
-
   def full_slug(separator = "-")
```
SamSaffron: I think you need something like this on category:   
```
self.ignored_columns = %w{
    suppress_from_latest
  }
```

db/post_migrate/20191107190330_remove_suppress_from_latest_from_category.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class RemoveSuppressFromLatestFromCategory < ActiveRecord::Migration[6.0]
+  DROPPED_COLUMNS ||= {
+    categories: %i{suppress_from_latest}
+  }
+
+  def up
+    muted_category_ids = SiteSetting.default_categories_muted.split("|")
+    suppressed_category_ids = Category.where(suppress_from_latest: true).pluck(:id).map(&:to_s)
```
SamSaffron: we got to use direct SQL here with no access to the object model. Also if you find more than say 10 here do not add it to the default mute.

If you are adding to default mute you need to apply historically with all the logic of skipping if people explicitly watch/track.

SamSaffron:
getting really close here, it does eliminate a fair amount of code!

db/post_migrate/20191107190330_remove_suppress_from_latest_from_category.rb
```diff
@@ -0,0 +1,46 @@
...
+      ids += muted_ids.split("|") if muted_ids.present?
+      ids.uniq!
+
+      # We shouldn't encourage to have more than 10 categories in `default_categories_muted` site setting.
+      if ids.count <= 10
+        DB.exec(<<~SQL, muted: CategoryUser.notification_levels[:muted])
+          INSERT INTO category_users (category_id, user_id, notification_level)
+            SELECT c.id category_id, u.id user_id, :muted
+            FROM users u
+              CROSS JOIN categories c
```
vinothkannans: I hope this `CROSS JOIN` won't hit any performance issues. Anyway, we're not migrating more than 10 categories.

vinothkannans:
@SamSaffron the changes are done.

SamSaffron:
I support this change... looks good to me Vinoth! 

Can you merge this in once you resolve the conflict here?

PR #8312, opened 2019-11-07
DEV: Move computed to discourseComputed 

markvanlan:


app/assets/javascripts/discourse-loader.js
```diff
@@ -3,6 +3,10 @@ var define, requirejs;
 (function() {
   // In future versions of ember we don't need this
   var EMBER_MODULES = {};
+  var ALIASES = {
```
markvanlan: @eviltrout Seems like simple aliases can be handled very easily like this.

PR #8314, opened 2019-11-07
REFACTOR: Attach resize controls to images from the markdown pipeline

romanrizzi:


app/assets/javascripts/pretty-text/engines/discourse-markdown/upload-protocol.js.es6
```diff
@@ -4,7 +4,6 @@ function addImage(uploads, token) {
     for (let i = 0; i < token.attrs.length; i++) {
       if (token.attrs[i][1].indexOf("upload://") === 0) {
         uploads.push([token, i]);
-        break;
```
SamSaffron: hmmm why was this break in there?

SamSaffron:
how do we make sure this rule only applies when cooking text in "preview" mode on the server when we do the cooked post processor?

romanrizzi:
@SamSaffron 

What about conditionally appllying the rule:

```js
helper.registerPlugin(md => {
    if (opts.previewing) {
      md.core.ruler.after("upload-protocol", "resize-controls", rule);
    }
  });
```

app/assets/javascripts/pretty-text/engines/discourse-markdown/upload-protocol.js.es6
romanrizzi: I believe it's not possible for the `upload://` to be present more than once. I think I removed it by accident, will add it again.

PR #8325, opened 2019-11-08
FIX: create_master_key got renamed

pfaffman:
In this commit https://github.com/discourse/discourse/commit/52c5cf33f87990a5828a34f68ff9df833cdc4a15#diff-4ae3b8df24c11943c19750fd06a1528d
@davidtaylorhq


davidtaylorhq:
Good catch - thanks @pfaffman. I think we should also rename this rake task to api:create, since this will always create a new key. It will never "get" an existing key.

PR #8330, opened 2019-11-11
FEATURE: Dismiss new per category

lis2:
Ability to dismiss new topics per category. I recorded a quick demo of how it is working

https://dev.discourse.org/t/add-per-category-dismiss-new/22865

Screencast-from-11-11-19-16_00_28

@vinothkannans could you take a look? Especially on part related to `mute_all_categories_by_default` because I changed it to use join instead

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -848,7 +848,23 @@ class TopicsController < ApplicationController
   end
 
   def reset_new
-    current_user.user_stat.update_column(:new_since, Time.now)
+    if params[:category_id].present?
+      category_ids = [params[:category_id]]
+
+      if params[:include_subcategories] == 'true'
+        category_ids = category_ids.concat(Category.where(parent_category_id: params[:category_id]))
+      end
+
+      category_ids.each do |category_id|
+        current_user
+          .category_users
+          .where(category_id: category_id)
+          .first_or_initialize
+          .update!(last_seen_at: Time.zone.now)
+      end
+    else
+      current_user.user_stat.update_column(:new_since, Time.now)
+    end
     render body: nil
   end
 
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -852,7 +852,7 @@ class TopicsController < ApplicationController
       category_ids = [params[:category_id]]
 
       if params[:include_subcategories] == 'true'
-        category_ids = category_ids.concat(Category.where(parent_category_id: params[:category_id]))
+        category_ids = category_ids.concat(Category.where(parent_category_id: params[:category_id]).pluck(:id))
       end
 
       category_ids.each do |category_id|
--- a/lib/topic_query.rb
+++ b/lib/topic_query.rb
@@ -875,7 +875,7 @@ class TopicQuery
         list = list
           .references("cu")
           .joins("LEFT JOIN category_users ON category_users.category_id = topics.category_id")
-          .where("category_users.user_id IS NULL OR category_users.user_id = :user_id", user_id: user.id)
+          .where("COALESCE(category_users.user_id, :user_id) = :user_id", user_id: user.id)
           .where("topics.category_id = :category_id
                  OR COALESCE(category_users.notification_level, :muted) <> :muted
                  OR tu.notification_level > :regular",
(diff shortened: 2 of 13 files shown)


lib/topic_query.rb
```diff
@@ -871,21 +872,16 @@ def remove_muted_categories(list, user, opts = nil)
...
-              AND (COALESCE(cu.notification_level, :muted) = :muted)
-              AND (COALESCE(tu.notification_level, :regular) <= :regular)
-          )", user_id: user.id,
-              muted: CategoryUser.notification_levels[:muted],
-              regular: TopicUser.notification_levels[:regular],
-              category_id: category_id || -1)
+        list = list
+          .references("cu")
+          .joins("LEFT JOIN category_users ON category_users.category_id = topics.category_id")
+          .where("category_users.user_id IS NULL OR category_users.user_id = :user_id", user_id: user.id)
```
vinothkannans: You can change it like `COALESCE(category_users.user_id, :user_id) = :user_id`.

lib/topic_query.rb
```diff
@@ -897,20 +893,16 @@ def remove_muted_categories(list, user, opts = nil)
...
-             AND cu.category_id <> :category_id
-             AND (tu.notification_level IS NULL OR tu.notification_level < :tracking)
-        )", user_id: user.id,
-            muted: CategoryUser.notification_levels[:muted],
-            tracking: TopicUser.notification_levels[:tracking],
-            category_id: category_id || -1)
+      list = list
+        .references("cu")
+        .joins("LEFT JOIN category_users ON category_users.category_id = topics.category_id")
+        .where("category_users.user_id IS NULL OR category_users.user_id = :user_id", user_id: user.id)
```
vinothkannans: Same here.

lib/topic_query.rb
```diff
@@ -897,20 +893,16 @@ def remove_muted_categories(list, user, opts = nil)
...
+        .where("category_users.notification_level IS NULL
+               OR category_users.notification_level <> :muted
```
vinothkannans: Also here `COALESCE(category_users.notification_level, :regular) = :muted`. Additionally you should new `regular` param for this.

lis2: thank you for your review,
I don't follow last comment
Original code was like that 
```
NOT EXISTS (
          SELECT 1
            FROM category_users cu
           WHERE cu.user_id = :user_id
             AND cu.category_id = topics.category_id
             AND cu.notification_level = :muted
             AND cu.category_id <> :category_id
             AND (tu.notification_level IS NULL OR tu.notification_level < :tracking)
```
so I want notification level different than muted
`category_users.notification_level <> :muted`

vinothkannans: Sorry! I have a typo in my previous comment. I like to change it like `COALESCE(category_users.notification_level, :regular) <> :muted` where the `:regular` is a new param `CategoryUser.notification_levels[:regular]`.

app/controllers/topics_controller.rb
```diff
@@ -848,7 +848,23 @@ def bulk
   end
 
   def reset_new
-    current_user.user_stat.update_column(:new_since, Time.now)
+    if params[:category_id].present?
+      category_ids = [params[:category_id]]
+
+      if params[:include_subcategories] == 'true'
```
ZogStriP: Why isn't `include_subcategories` a boolean?

app/controllers/topics_controller.rb
```diff
@@ -848,7 +848,23 @@ def bulk
...
+        category_ids = category_ids.concat(Category.where(parent_category_id: params[:category_id]).pluck(:id))
+      end
+
+      category_ids.each do |category_id|
```
ZogStriP: Is there a way we could do this in 1 SQL query instead of 1 per category?

SamSaffron: I think the first_or_initialize makes this somewhat tricky. I guess we could work this in 2 batches (existing category users followed by new category user records) however I feel it does make the code a fair bit more complex to reason about.

This should somewhat work even if a category has say 50-100 sub-categories, would start being really really painful if you try dismissing and there 10,000 sub categories. 

I think we can wait and see a bit here.

app/controllers/topics_controller.rb
SamSaffron: I think this is rails shanigans, we do this elsewhere in a few places in the same file https://github.com/discourse/discourse/blob/1c179177e7c84362572481e61b18597f7d2d5050/app/controllers/topics_controller.rb#L221-L221

lib/topic_query.rb
SamSaffron: OR in conditionals worry me a bit cause this is one area where SQL can get slow, but NOT EXISTS can be just as bad or worse.... did you time the full query this on some users with lots of state ?

PR #8333, opened 2019-11-11
FEATURE: hide posts from incoming email based on dmarc verdict

LeoMcA:
Here's my first pass at this.

To provide a brief explanation of what's going on here:
* `AuthenticationResults#parse_header` is a hellscape, blame whoever decided comments could be written in practically all parts of the header
* `AuthenticationResults#verdict` returns `:pass`, `:fail`, or `:gray`. Currently this just calculated based on dmarc, but this could also use spf/dkim/etc if dmarc is missing from a message in future.
* `AuthenticationResults#action` is where the fun happens and we decide what to do with each verdict. Currently we have some sensible defaults, but I see this being changed with a lax/normal/strict setting in future.
* `SiteSetting.email_in_authserv_id` really needs to be set for this to all properly work. [Wikipedia has a good explanation of what this is](https://en.wikipedia.org/wiki/Email_authentication#Authentication-Results). Even without it, we still fail the message if there's a header with a failing dmarc.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1881,6 +1881,7 @@ en:
     log_mail_processing_failures: "Log all email processing failures to <a href='%{base_path}/logs' target='_blank'>/logs</a>"
     email_in: 'Allow users to post new topics via email (requires manual or pop3 polling). Configure the addresses in the "Settings" tab of each category.'
     email_in_min_trust: "The minimum trust level a user needs to have to be allowed to post new topics via email."
+    email_in_authserv_id: "The authentication service identifier used by the configured polling method."
     email_in_spam_header: "The email header to detect spam."
     email_prefix: "The [label] used in the subject of emails. It will default to 'title' if not set."
     email_site_title: "The title of the site used as the sender of emails from the site. Default to 'title' if not set. If your 'title' contains characters that are not allowed in email sender strings, use this setting."
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1881,7 +1881,7 @@ en:
     log_mail_processing_failures: "Log all email processing failures to <a href='%{base_path}/logs' target='_blank'>/logs</a>"
     email_in: 'Allow users to post new topics via email (requires manual or pop3 polling). Configure the addresses in the "Settings" tab of each category.'
     email_in_min_trust: "The minimum trust level a user needs to have to be allowed to post new topics via email."
-    email_in_authserv_id: "The authentication service identifier used by the configured polling method."
+    email_in_authserv_id: "The identifier of the service doing authentication checks on incoming emails. See <a href='https://meta.discourse.org/t/134358'>https://meta.discourse.org/t/134358</a> for instructions on how to configure this."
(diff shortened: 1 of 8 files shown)


lib/email/receiver.rb
```diff
@@ -196,7 +196,13 @@ def process_internal
...
 
     def hidden_reason_id
-      @hidden_reason_id ||= is_spam? ? Post.hidden_reasons[:email_spam_header_found] : nil
+      @hidden_reason_id ||= if is_spam?
+                              Post.hidden_reasons[:email_spam_header_found]
+                            elsif auth_res_action == :hide
+                              Post.hidden_reasons[:email_authentication_result_header]
+                            else
+                              nil
+                            end
```
LeoMcA: I had initially thought that this would *also* flag the post, but through testing realised this wasn't the case. In fact, the hidden reason isn't actually surfaced anywhere in the UI.

I started going down the rabbit hole of flagging this post as well, or enqueuing it instead, but relatively quickly realised it was going to take some time. I think some method of ensuring that these posts appear in a moderator's reviewables is important, but a bit beyond the scope of this PR - as I feel the same thing should happen to incoming emails flagged as spam.

What should we do here?

config/locales/server.en.yml
```diff
@@ -1881,6 +1881,7 @@ en:
     log_mail_processing_failures: "Log all email processing failures to <a href='%{base_path}/logs' target='_blank'>/logs</a>"
     email_in: 'Allow users to post new topics via email (requires manual or pop3 polling). Configure the addresses in the "Settings" tab of each category.'
     email_in_min_trust: "The minimum trust level a user needs to have to be allowed to post new topics via email."
+    email_in_authserv_id: "The authentication service identifier used by the configured polling method."
```
gschlager: I think the setting description could be improved. Without looking at the code I have no idea what that setting is for and how to use it.

lib/email/receiver.rb
gschlager: I agree. It might be a good idea to create reviewables for hidden posts in an additional PR. @eviltrout might have some hints for you on that topic.

> the hidden reason isn't actually surfaced anywhere in the UI

I noticed that as well, because I wanted to point out that there's a missing translation for the new hidden reason. But yeah, they are not surfaced on the UI, so no translation is missing. I wonder if that will change when reviewables are created.

gschlager:
@LeoMcA Any chance you can improve the description of the `email_in_authserv_id` site setting? The rest of the PR looks good to me. Maybe include an example and/or link to Wikipedia or a topic on Meta?

LeoMcA:
@gschlager apologies, this slipped through the cracks of a rather hectic week last week - will get on it now.

LeoMcA:
Updated the setting text (is it any clearer?), but crucially included a link to this shortish help article I've written (currently as a PM, but you're all admins so I assume you can see it): https://meta.discourse.org/t/configuring-authentication-checks-on-incoming-email/134358

I didn't want to "announce" a feature with a help article before it had actually been merged, you can convert it into a proper topic when the time is right :)

PR #8334, opened 2019-11-11
REFACTOR: Remove context from filterMode

danielwaterworth:
The context is already provided via: category, noSubcategories and
tagId. Let's not duplicate that information in the filterMode.


SamSaffron:
I totally support this change. However you are going to need to be careful when applying it, I wonder if any plugins / theme components depend on the old behavior of filterMode.

Also, I noticed we have almost no tests around this area, should we have a few?

PR #8335, opened 2019-11-11
FEATURE: Send a message to new TL2 users and point them to the advanced tutorial

romanrizzi:
Screen Shot 2019-11-11 at 18 09 57

### DIFF
--- a/plugins/discourse-narrative-bot/autoload/jobs/send_advanced_tutorial_message.rb
+++ b/plugins/discourse-narrative-bot/autoload/jobs/send_advanced_tutorial_message.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Jobs
+  class SendAdvancedTutorialMessage < ::Jobs::Base
+    def execute(args)
+      user = User.find_by(id: args[:user_id])
+      raise Discourse::InvalidParameters.new(:user_id) if user.nil?
+
+      discobot_user = User.find(-2)
+
+      PostCreator.create!(
+        discobot_user,
+        title: I18n.t("discourse_narrative_bot.tl2_promotion_message.subject_template"),
+        raw: I18n.t("discourse_narrative_bot.tl2_promotion_message.text_body_template"),
+        archetype: Archetype.private_message,
+        target_usernames: user.username,
+        skip_validations: true
+      )
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


plugins/discourse-narrative-bot/autoload/jobs/send_advanced_tutorial_message.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Jobs
+  class SendAdvancedTutorialMessage < ::Jobs::Base
+    def execute(args)
+      user = User.find_by(id: args[:user_id])
+      raise Discourse::InvalidParameters.new(:user_id) if user.nil?
+
+      discobot_user = User.find(-2)
```
ZogStriP: You can instead use

```suggestion
      discobot_user = DiscourseNarrativeBot::Actions.discobot_user
```

plugins/discourse-narrative-bot/spec/jobs/send_advanced_tutorial_message_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+RSpec.describe Jobs::SendAdvancedTutorialMessage do
+  before do
+    Jobs.run_immediately!
+    SiteSetting.discourse_narrative_bot_enabled = true
+  end
+
+  it 'sends a message to the promoted user' do
+    user = Fabricate(:user)
+    discobot_user = User.find(-2)
```
eviltrout: I'd also update this one to use the constant @ZogStriP  suggested.

romanrizzi:
@eviltrout @ZogStriP I replaced all occurrences of `User.find(-2)` with that method call. 👍

coding-horror:
wait why is this coming from discobot? What if discobot is disabled? I don't think the TL1 "congrats" PM comes from discobot?

romanrizzi:
> I don't think the TL1 "congrats" PM comes from discobot?

Good catch, @coding-horror! I changed it to system user instead of discobot.

> What if discobot is disabled?

If this happens, the message won't be sent since there won't be an advanced tutorial either.

plugins/discourse-narrative-bot/spec/jobs/send_advanced_tutorial_message_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+RSpec.describe Jobs::SendAdvancedTutorialMessage do
+  before do
+    Jobs.run_immediately!
+    SiteSetting.discourse_narrative_bot_enabled = true
+  end
+
+  it 'sends a message to the promoted user' do
+    user = Fabricate(:user)
+    discobot_user = Discourse.system_user
```
ZogStriP: Technically, `system_user` is not `discobot` ;)

PR #8336, opened 2019-11-12
FIX: tag and category watchers regression

lis2:
I made a regression here https://github.com/discourse/discourse/commit/17366d3bcc8ac27980e4fb3b7cfdc1ed86c6ab02#diff-ddeebb36d131f89ca91be9d04c2baefaR10

When the tag is added, people watching specific tag are notified but also people watching specific category.

Therefore, `notify_post_users` should accept options who should be notified.

So when `category` is added to the topic, users watching topic and users watching category are notified.

When `tag` is added to the topic, users watching topic and users watching tag are notified

Finally, when a new post is created, everybody is notified, topic watchers, category watchers, tag watchers.

### DIFF
--- a/app/jobs/regular/notify_category_change.rb
+++ b/app/jobs/regular/notify_category_change.rb
@@ -7,7 +7,7 @@ module Jobs
 
       if post&.topic&.visible?
         post_alerter = PostAlerter.new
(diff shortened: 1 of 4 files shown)


SamSaffron:
Aha, yes this change makes sense! lets try it out!

PR #8337, opened 2019-11-12
DEV: Remove use of `cd` in the app

davidtaylorhq:
`FileUtils.cd` and `Dir.chdir` cause the working directory to change for the entire process. We run sidekiq jobs, hijacked requests and deferred jobs in threads, which can make working directory changes have unintended side-effects. 

- Add a rubocop rule to warn about usage of `Dir.chdir` and `FileUtils.cd`
- Added rubocop:disable for scripts used outside the app
- Refactored code using `cd` to use alternative methods
- Temporarily skipped the rubocop check for `lib/backup_restore`. This will require more complex refactoring, so I will create a separate PR for review

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -133,5 +133,4 @@ DiscourseCops/NoChdir:
   Enabled: true
   Exclude:
     - 'spec/**/*' # Specs are run sequentially, so chdir can be used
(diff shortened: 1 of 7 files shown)


SamSaffron:
Great change, feel free to merge!

PR #8339, opened 2019-11-12
DEV: prevents input/change events to cause a full rerender

jjaffeux:
Code should decide when do something with the event value, and maybe cause a rererender but it shouldn't be automatic. This is currently a gigantic waste of resources.


eviltrout:
The thinking here is it can be hard to always know when something needs to be re-rendered, so just diff the DOM.

If we remove this, and a user edits an input, it might not re-render the widget to handle that edit right?

jjaffeux:
@eviltrout it's an HTML component, so it will update it's client state. For me not doing this is simply refusing DDAU.

When you rerender widget on each update, basically what you say is let event mutate state. With my PR it forces you to do:

```
input(event) {
  this.sendWidgetActions("fillterChanged", event.target.value);
}
```

Which is more efficient, cleaner and flexible.

PR #8343, opened 2019-11-12
DEV: Remove Discourse.User and import instead

markvanlan:


jjaffeux:
Would be nice to remove the `.get("xxx")` too

eviltrout:
I merged this in a different branch. Thanks for the help @markvanlan

PR #8345, opened 2019-11-13
FEATURE: Allow scoping search to tag

martin-brennan:
Meta thread: https://meta.discourse.org/t/automatically-offer-to-scope-search-by-tag-when-browsing-tags/

* When browsing a tag the search can now be scoped to the tag
* This also applies on the full page search when context is provided in the mobile view

image


lib/search.rb
```diff
@@ -791,6 +791,8 @@ def posts_query(limit, opts = nil)
     posts = Post.where(post_type: Topic.visible_post_types(@guardian.user))
       .joins(:post_search_data, :topic)
       .joins("LEFT JOIN categories ON categories.id = topics.category_id")
+      .joins("LEFT JOIN topic_tags ON topic_tags.topic_id = topics.id")
```
SamSaffron: hmmm why are we joining these tables if @search_context is not a Tag?

SamSaffron:
Very tidy PR except for 1 minor comment.

lib/search.rb
martin-brennan: Good point, will do this conditionally along with the category join

martin-brennan: Actually, never mind, the categories join is used for a few other things, so I'll only conditionally join the tag tables

app/controllers/search_controller.rb
```diff
@@ -169,6 +169,8 @@ def lookup_search_context
         context_obj = Category.find_by(id: search_context[:id].to_i)
       elsif 'topic' == search_context[:type]
         context_obj = Topic.find_by(id: search_context[:id].to_i)
+      elsif 'tag' == search_context[:type]
+        context_obj = Tag.where_name(search_context[:id]).first
```
SamSaffron: Minor but API wise this is a bit confusing. Usually when we hand over "ids" you would expect this to be the Tag id. 

Perhaps we populate `search_context[:name]` in this case?

martin-brennan: Hmm yeah good point

spec/components/search_spec.rb
```diff
@@ -792,6 +792,21 @@ def search
...
 
+    it 'can use tag as a search context' do
+      tag = Fabricate(:tag, name: 'important-stuff')
+
+      topic = Fabricate(:topic)
+      topic_no_tag = Fabricate(:topic)
+      Fabricate(:topic_tag, tag: tag, topic: topic)
+
+      post = Fabricate(:post, topic: topic, user: topic.user, raw: 'This is my hello')
+      _another_post = Fabricate(:post, topic: topic_no_tag, user: topic.user)
```
eviltrout: In this case it's a bit cleaner to not assign the variable rather than use the underscore to indicate it's not used.

spec/components/search_spec.rb
```diff
@@ -792,6 +792,21 @@ def search
...
+
+      search = Search.execute('hello', search_context: tag)
+      expect(search.posts.map(&:id).sort).to eq([post.id].sort)
```
eviltrout: You could use https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher rather than sorting yourself.

martin-brennan:
@eviltrout thanks for the review, you're right on both counts, will fix up. That's what happens when you copy another test and don't read over them critically :sweat_smile:

PR #8348, opened 2019-11-13
Slow down the "Screaming Frog SEO Spider"

Supermathie:
* adds 'seo spider' to the default slowdown list
* we're seeing obnoxious crawl rates (20+ requests/sec) hammering 
customer sites


coding-horror:
Just block it by default don’t bother with the slow down

SamSaffron:
Sure, this is now blocked by default.

PR #8350, opened 2019-11-14
Bump prettier to 1.19.1 and lock dependency

martin-brennan:
There were issues that devs were encountering where their local prettier was on one version,  e.g. 1.18.2, and the jenkins version was the latest,  e.g. 1.19.1, which use different default rulesets.

This was causing lefthook to pass locally but fail in jenkins. Now we should have no surprises from  prettier when merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "install-peerdeps": "^1.10.2",
     "lodash-cli": "https://github.com/lodash-archive/lodash-cli.git",
     "pretender": "^1.6",
-    "prettier": "^1.18.2",
+    "prettier": "1.19.1",
     "puppeteer": "1.20",
     "qunit": "2.8.0",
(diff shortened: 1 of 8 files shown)


pmusaraj:
Do we have to pin the version @martin-brennan?

martin-brennan:
@pmusaraj if we don't pin the version I thought we would end up with the same issue, which is that devs will have a local version e.g. 1.18.2 that does not change because of their yarn.lock file, whereas jenkins gets the latest each time. This has already caused broken builds a few times after a master merge.

The only way that devs can fix locally is `yarn upgrade prettier` if we set the dependency to `^1.19.1` but we would have to remember to do that, which is the current issue. I'm not sure if its possible to check whether the locked local version is different from the latest allowable by the package in the pre-commit hook or something.

PR #8351, opened 2019-11-14
FIX: Tracking Topic State know about category_seen_at

lis2:
If category got `last_seen_at` is set TrackingTopicState should know about it and exclude those topics from marking them as new.

@SamSaffron could you take a look on that? Logically it makes sense to me, however, I couldn't replicate bug from dev on my local


SamSaffron:
This looks correct to me, but we need to think through message bus implications as well.

PR #8352, opened 2019-11-14
FEATURE: Add chart option to polls

markvanlan:


ZogStriP:
Can we get screenshots of that awesome feature?

markvanlan:
> Can we get screenshots of that awesome feature?

This is going to get much more complex over the coming days. 

Here is a very early preview:

Screenshot from 2019-11-14 13-16-20@2x

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
```diff
@@ -17,6 +17,16 @@
                     valueAttribute="value"}}
       </div>
 
+      {{#unless isNumber}}
+        <div class="input-group poll-select">
+          <label class="input-group-label">{{i18n 'poll.ui_builder.poll_chart_type.label'}}</label>
+          {{combo-box content=pollChartTypes
+                      value=chartType
+                      allowInitialValueMutation=true
```
jjaffeux: this is going to be deprecated.

The preferred solution is to set `chartType` with an initial value

plugins/poll/assets/javascripts/initializers/extend-for-poll.js.es6
```diff
@@ -88,12 +88,17 @@ function initializePolls(api) {
...
           id: `${pollName}-${post.id}`,
           post,
           poll,
-          vote
-        });
+          vote,
+          groupableUserFields: api.container
+            .lookup("site-settings:main")
+            .poll_groupable_user_fields.split("|")
+            .filter(f => f)
```
jjaffeux:
```suggestion
            .filter(Boolean)
```

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,70 @@
...
+        20: [255, 236, 179, 1],
+        45: [232, 82, 133, 1],
+        65: [106, 27, 154, 1],
+        100: [0, 0, 0, 1]
+      };
+      break;
+  }
+
+  //Get a sorted array of the gradient keys
+  var gradientKeys = Object.keys(gradient);
```
jjaffeux: we don't user var anymore in the codebase

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -323,7 +326,11 @@ createWidget("discourse-poll-container", {
 
     if (attrs.showResults) {
       const type = poll.get("type") === "number" ? "number" : "standard";
-      return this.attach(`discourse-poll-${type}-results`, attrs);
+      const resultsWidget =
+        type === "number" || attrs.poll.get("chart") !== "pie"
```
jjaffeux:
```suggestion
        type === "number" || attrs.poll.chart !== "pie"
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+    if (error) {
+      popupAjaxError(error);
+    } else {
+      bootbox.alert(I18n.t("poll.error_while_fetching_voters"));
+    }
+  });
+}
+
+function transformUserFieldToLabel(fieldName) {
+  let newName = fieldName;
```
jjaffeux: guess you could replace directly on `fieldName` here, that intermediate assignment doesn't seem necessary

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+      const existingSel = document.getElementById(fieldSelectId);
+      if (existingSel) {
+        existingSel.parentNode.removeChild(existingSel);
+      }
+
+      const sel = document.createElement("select");
+      sel.classList.add("poll-group-by-selector");
+      sel.id = fieldSelectId;
+
+      const clearFix = document.createElement("div");
```
jjaffeux: is this clearfix necessary? can you try to use flexbox and not rely on this kind of css tricks?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+
+      parent.prepend(sel);
+      parent.append(clearFix);
+
+      for (var i = 0; i < attrs.groupableUserFields.length; i++) {
+        const opt = document.createElement("option");
+        opt.innerHTML = transformUserFieldToLabel(fields[i]);
+        opt.value = fields[i];
+        sel.appendChild(opt);
+        $(sel).val(attrs.groupedBy);
```
jjaffeux: is jquery needed for this?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+          chartIdx < result.grouped_results.length;
+          chartIdx++
+        ) {
+          const data = result.grouped_results[chartIdx].options.map(
+            o => o.votes
+          );
+          const labels = result.grouped_results[chartIdx].options.map(
+            o => o.html
+          );
+          const chartConfig = pieChartConfig(data, labels, 1.2);
```
jjaffeux: I think last arg here should be an option to make it explicit=

```
const chartConfig = pieChartConfig(data, labels, { aspectRatio: 1.2 });
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+          }
+        }
+      });
+    });
+    return "";
+  },
+
+  click(e) {
+    let select = $(e.target).closest("select");
+    if (select.length !== 0) {
```
jjaffeux: wouldn't that be enough?

```suggestion
    if (select.length) {
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+});
+
+function pieChartConfig(data, labels, aspectRatio) {
+  aspectRatio = aspectRatio || 2.0;
+  return {
+    type: "pie",
+    data: {
+      datasets: [
+        {
+          data: data,
```
jjaffeux:
```suggestion
          data,
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+          backgroundColor: getColors(data.length)
+        }
+      ],
+      labels: labels
```
jjaffeux:
```suggestion
      labels
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+    },
+    options: {
+      responsive: true,
+      aspectRatio: aspectRatio,
```
jjaffeux:
```suggestion
      aspectRatio,
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+      if (attrs.groupableUserFields.length > 0) {
+        btn = this.attach("button", {
+          className: "btn btn-default poll-group-by-toggle",
+          label: "poll.group-results.label",
+          title: "poll.group-results.title",
+          icon: "far-eye",
+          action: "toggleGroupedPieCharts"
+        });
+      }
+      const data = attrs.poll.get("options").map(o => o.votes);
```
jjaffeux:
```suggestion
      const data = attrs.poll.options.map(o => o.votes);
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+      const labels = attrs.poll.get("options").map(o => o.html);
```
jjaffeux:
```suggestion
      const labels = attrs.poll.options.map(o => o.html);
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+    let select = $(e.target).closest("select");
+    if (select.length !== 0) {
+      this.sendWidgetAction("refreshCharts", select[0].value);
+    }
+  }
+});
+
+function clearPieChart(id) {
+  let el = document.querySelector(`#poll-results-chart-${id}`);
+  if (el) el.parentNode.removeChild(el);
```
jjaffeux: That's a nitpick, feel free to ignore, but I like to write it that way:

```suggestion
  el && el.parentNode.removeChild(el);
```

plugins/poll/assets/stylesheets/common/poll.scss
```diff
@@ -119,6 +121,37 @@ div.poll {
...
     }
   }
+
+  .poll-results-chart {
+    height: 300px;
+    overflow: auto;
+  }
+
+  .poll-group-by-toggle {
+    float: left;
```
jjaffeux: would really prefer we try to code with flexbox and not floats

plugins/poll/assets/stylesheets/common/poll.scss
```diff
@@ -119,6 +121,37 @@ div.poll {
...
+    margin-bottom: 10px;
+  }
+
+  .poll-group-by-selector {
+    float: right;
```
jjaffeux: would really prefer we try to code with flexbox and not floats

plugins/poll/assets/stylesheets/common/poll.scss
```diff
@@ -119,6 +121,37 @@ div.poll {
...
+    height: 30px;
+  }
+
+  .clear-both {
```
jjaffeux: same remark

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,70 @@
...
+  gradientKeys.sort(function(a, b) {
+    return +a - +b;
+  });
+
+  //Calculate colors
+  var colors = [];
+  for (let i = 0; i < count; i++) {
+    var gradientIndex = (i + 1) * (100 / (count + 1)); //Find where to get a color from the gradient
```
jjaffeux: no var

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,70 @@
...
+        break;
+      } else if (gradientIndex < +gradientKey) {
+        //It's somewhere between this gradient key and the previous
+        var prevKey = gradientKeys[j - 1];
+        var gradientPartIndex =
+          (gradientIndex - prevKey) / (gradientKey - prevKey); //Calculate where
+        var color = [];
+        for (let k = 0; k < 4; k++) {
+          //Loop through Red, Green, Blue and Alpha and calculate the correct color and opacity
+          color[k] =
```
jjaffeux: why not just push into color ?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+        `poll-results-grouped-pie-charts-${attrs.id}`
+      );
+      const fieldSelectId = `field-select-${attrs.id}`;
+
+      const existingSel = document.getElementById(fieldSelectId);
+      if (existingSel) {
+        existingSel.parentNode.removeChild(existingSel);
+      }
+
+      const sel = document.createElement("select");
```
jjaffeux: `groupBySelect` would probably be a better name, you use `sel` multiple lines later, and without context we don't know what it is

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+        opt.value = fields[i];
+        sel.appendChild(opt);
+        $(sel).val(attrs.groupedBy);
+      }
+      fetchGroupedVoters({
+        post_id: attrs.post.id,
+        poll_name: attrs.poll.name,
+        user_field_name: attrs.groupedBy
+      }).then(result => {
+        for (
```
jjaffeux: I see you use `for` a lot, I didn't look at each of them, but might be worth considering using `forEach`

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+          icon: "far-eye-slash",
+          action: "toggleGroupedPieCharts"
+        })
+      ]);
+      chart = this.attach("discourse-poll-grouped-pies", attrs);
+      clearPieChart(this.attrs.id);
+    } else {
+      if (attrs.groupableUserFields.length > 0) {
+        btn = this.attach("button", {
+          className: "btn btn-default poll-group-by-toggle",
```
jjaffeux: pretty sure btn class is added automatically

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
markvanlan: @jjaffeux the initial value of `chartType` is set in the controller. What do you mean be "this" being deprecated?

jjaffeux: I mean I will remove "allowInitialValueMutation" option from select-kit, this is not a good practice.

So if `chartType` is set, why do you need this? `allowInitialValueMutation` is basically doing:

```
if (!this.chartType) {
  this.set("chartType", pollChartTypes.firstObject);
}
```

markvanlan: Yeah, I had no clue what you were referring to.

All the other combo-boxes have that key value pair, and I copy pasted without thinking much about it. Happy to remove it

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -38,6 +42,11 @@ export default Controller.extend({
     ];
   },
 
+  @computed("chartType", "pollType", "numberPollType")
+  isPie(chartType, pollType, numberPollType) {
+    return pollType !== numberPollType && chartType === "pie";
```
eviltrout: We've been moving away from string constants because we find them error prone. It's okay for the variable to be a string but could you compare it against a constant of some sort?

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
```diff
@@ -17,6 +17,15 @@
                     valueAttribute="value"}}
       </div>
 
+      {{#unless isNumber}}
+        <div class="input-group poll-select">
+          <label class="input-group-label">{{i18n 'poll.ui_builder.poll_chart_type.label'}}</label>
+          {{combo-box content=pollChartTypes
+                      value=chartType
+                      valueAttribute="value"}}
```
eviltrout: I thought `value` is the default valueAttribute? Am I mistaken?

markvanlan:
@jjaffeux I am looking into flexbox and using forEach instead. Anything else look wrong?

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -173,6 +182,8 @@ export default Controller.extend({
     "pollMax",
     "pollStep",
     "autoClose",
+    "chartType",
+
```
markvanlan: remove this space

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,71 @@
...
+        65: [106, 27, 154, 1],
+        100: [0, 0, 0, 1]
+      };
+      break;
+  }
+
+  //Get a sorted array of the gradient keys
+  let gradientKeys = Object.keys(gradient);
+  gradientKeys.sort(function(a, b) {
+    return +a - +b;
```
eviltrout: Curious about why there's a `+a` and `+b` here?

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,71 @@
+export function getColors(count, palette) {
+  palette = palette || "cool";
+  let gradient;
+
+  switch (palette) {
```
eviltrout: One issue here is these colours are not really customizable via theme. That might be OK but it's something to think about.

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,71 @@
...
+    case "cool":
+      gradient = {
+        0: [255, 255, 255, 1],
```
eviltrout: Each gradient has an alpha of 1, so why not omit it?

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,71 @@
...
+
+  //Get a sorted array of the gradient keys
+  let gradientKeys = Object.keys(gradient);
+  gradientKeys.sort(function(a, b) {
+    return +a - +b;
+  });
+
+  //Calculate colors
+  let colors = [];
+  for (let i = 0; i < count; i++) {
```
eviltrout: This seems a little more complicated than it could be. The loop is O(n^3) now, looping through each count, gradient and 4 color values.

The gradients only increase, so there is no need to check all gradientKeys every iteration. Instead you could use a variable to represent the current value and one of the next value. Replace those values when your current index passes into the next value.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
   }
 });
 
+function fetchGroupedResults(data) {
```
eviltrout: This function is only used once so it might be simpler to inline it.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+      popupAjaxError(error);
+    } else {
+      bootbox.alert(I18n.t("poll.error_while_fetching_voters"));
+    }
+  });
+}
+
+function transformUserFieldToLabel(fieldName) {
+  let transformed = fieldName.split("_");
+  transformed[0] =
```
eviltrout: This seems to be doing a `classify` operation on the string? Might as well use Ember's version!

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+  tagName: "div.poll-grouped-pies",
+  buildAttributes(attrs) {
+    return {
+      id: `poll-results-grouped-pie-charts-${attrs.id}`
+    };
+  },
+
+  html(attrs) {
+    const fields = Object.assign({}, attrs.groupableUserFields);
+    later(() => {
```
eviltrout: We will have to be careful to test this for memory leaks and to make sure that if the poll is re-rendered it still works.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+      attrs.groupedBy = attrs.groupedBy || fields[0];
+      const parent = document.getElementById(
+        `poll-results-grouped-pie-charts-${attrs.id}`
```
eviltrout: Isn't this `attrs.id`? No reason to build the string again.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+            label.classList.add("poll-pie-label");
+            label.textContent = result.grouped_results[chartIdx].group;
+
+            const canvas = document.createElement("canvas");
+            canvas.classList.add(`poll-grouped-pie-${attrs.id}`);
+            canvas.id = canvasId;
+
+            container.appendChild(label);
+            container.appendChild(canvas);
+            parent.appendChild(container);
```
eviltrout: I wonder if it would make more sense to have the widget emit the HTML and then only perform the `Chart` in `later()`? Did you try that approach?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+                instance.destroy();
+                // eslint-disable-next-line
+                new Chart(el.getContext("2d"), chartConfig);
+              }
+            });
+          }
+        }
+      });
+    });
+    return "";
```
eviltrout: You should be able to omit this `return`. If a widget returns nothing it shouldn't render anything.

plugins/poll/config/locales/client.en.yml
```diff
@@ -63,6 +63,14 @@ en:
         title: "Back to your votes"
         label: "Hide results"
 
+      group-results:
+        title: "Show votes in grouped by user field"
```
eviltrout:
```suggestion
        title: "Group votes by user field"
```

plugins/poll/config/locales/client.en.yml
```diff
@@ -63,6 +63,14 @@ en:
...
+        label: "Show breakdown"
+
+      ungroup-results:
+        title: "Show all votes combined"
```
eviltrout:
```suggestion
        title: "Combine all votes"
```

plugins/poll/config/locales/server.en.yml
```diff
@@ -20,6 +20,7 @@ en:
     poll_maximum_options: "Maximum number of options allowed in a poll."
     poll_edit_window_mins: "Number of minutes after post creation during which polls can be edited."
     poll_minimum_trust_level_to_create: "Define the minimum trust level needed to create polls."
+    poll_groupable_user_fields: "User field names that poll results can be grouped and viewed by."
```
eviltrout:
```suggestion
    poll_groupable_user_fields: "A set of user field names that can be used to group and filter poll results."
```

plugins/poll/config/settings.yml
```diff
@@ -14,3 +14,8 @@ plugins:
     default: 1
     client: true
     enum: 'TrustLevelSetting'
+  poll_groupable_user_fields:
+    default: ""
```
eviltrout: I believe you can omit the default here.

plugins/poll/db/migrate/20191114160613_add_graph_to_polls.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddGraphToPolls < ActiveRecord::Migration[6.0]
+  def change
+    add_column :polls, :chart, :integer, default: 0, null: false
```
eviltrout: Maybe this should be called `chart_type`?

plugins/poll/plugin.rb
```diff
@@ -230,6 +230,66 @@ def voters(post_id, poll_name, user, opts = {})
...
       end
 
+      def transform_for_user_field_override(custom_user_field)
+        existing_field = UserField.find_by(name: custom_user_field)
+        existing_field ? "user_field_#{existing_field.id}" : custom_user_field
+      end
+
+      def grouped_poll_results(post_id, poll_name, user_field_name, user)
+        post = Post.find_by(id: post_id)
+        raise Discourse::InvalidParameters.new("post_id is invalid") unless post
```
eviltrout: Usually we pass the symbol for the parameter to this exception. Ex: `post_id` (same note below.)

plugins/poll/plugin.rb
```diff
@@ -230,6 +230,66 @@ def voters(post_id, poll_name, user, opts = {})
...
+
+        poll = Poll.find_by(post_id: post_id, name: poll_name)
```
eviltrout: If you `includes(:poll_votes).includes(:poll_options)` and then update the queries below to be `poll.poll_options` and `poll.poll_votes` rails might eager load them and save some queries.

plugins/poll/plugin.rb
```diff
@@ -230,6 +230,66 @@ def voters(post_id, poll_name, user, opts = {})
...
+          poll_options[option.id.to_s] = { html: option.html, digest: option.digest }
+        end
+
+        user_ids = poll_votes.map(&:user_id).uniq
+        user_fields = UserCustomField.where(user_id: user_ids, name: transform_for_user_field_override(user_field_name))
+
+        user_field_map = {}
+        user_fields.each do |f|
+          # Build hash, so we can quickly look up field values for each user.
+          user_field_map[f.user_id.to_s] = f.value
```
eviltrout: Why cast the `user_id` to a string?

plugins/poll/spec/integration/poll_endpoints_spec.rb
```diff
@@ -119,4 +119,71 @@
       end
     end
   end
+
+  describe "#grouped_poll_results" do
+    let(:user1) { Fabricate(:user) }
```
eviltrout: Could you use the `fab!` helpers here? They can improve performance.

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
markvanlan: Other combo-boxes in the same file have it set like this. I followed suite. I can fix it in the other places as well.

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
markvanlan: Are you saying to do  this?
```
return pollType !== "number" && chartType === "pie";
```

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/feature-add-chart-option-to-polls-pr-8352/7018/57):

no it’s id sadly and won’t change I think too much legacy...

plugins/poll/assets/javascripts/discourse/templates/modal/poll-ui-builder.hbs
jjaffeux: Yes I know all of this is not documented ATM, and a for a reason, there were a lot of weird options to handle all the cases we had and all of this will be gone :)

farewell

jjaffeux:
I think the feature is 😍and code is good 👍 but it looks like a lot of code and kinda complex for what it's doing, so I'm waiting for when you think you are done, to pull the PR play with it, and maybe do a deeper review.

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
jjaffeux: No I think he thinks of something like this:

```
    const PIE_TYPE_IDENTIFIER = "pie";

    // code omitted

    return pollType !== numberPollType && chartType === PIE_TYPE_IDENTIFIER;
```

plugins/poll/assets/javascripts/lib/chart-colors.js.es6
```diff
@@ -0,0 +1,71 @@
...
+        45: [232, 82, 133, 1],
+        65: [106, 27, 154, 1],
+        100: [0, 0, 0, 1]
+      };
+      break;
+  }
+
+  //Get a sorted array of the gradient keys
+  let gradientKeys = Object.keys(gradient);
+  gradientKeys.sort(function(a, b) {
```
jjaffeux: might want to use a fat arrow here

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+      groupBySelect.id = fieldSelectId;
+
+      const clearFix = document.createElement("div");
+      clearFix.classList.add("clear-both");
+
+      parent.prepend(groupBySelect);
+      parent.append(clearFix);
+
+      for (var i = 0; i < attrs.groupableUserFields.length; i++) {
+        const opt = document.createElement("option");
```
jjaffeux: opt could maybe use a better name

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+          label: "poll.ungroup-results.label",
+          title: "poll.ungroup-results.title",
+          icon: "far-eye-slash",
+          action: "toggleGroupedPieCharts"
+        })
+      ]);
+      chart = this.attach("discourse-poll-grouped-pies", attrs);
+      clearPieChart(this.attrs.id);
+    } else {
+      if (attrs.groupableUserFields.length > 0) {
```
jjaffeux:
```suggestion
      if (attrs.groupableUserFields.length) {
```

should be enough no ?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+        btn = this.attach("button", {
+          className: "btn-default poll-group-by-toggle",
+          label: "poll.group-results.label",
+          title: "poll.group-results.title",
+          icon: "far-eye",
+          action: "toggleGroupedPieCharts"
+        });
+      }
+      const data = attrs.poll.options.map(o => o.votes);
```
jjaffeux: I think it should work:

```suggestion
      const data = attrs.poll.options.mapBy("votes");
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+      const labels = attrs.poll.options.map(o => o.html);
```
jjaffeux: I think it should work:

```suggestion
      const labels = attrs.poll.options.mapBy("html");
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+      loadScript("/javascripts/Chart.min.js").then(() => {
+        later(() => {
```
jjaffeux: can you explain this later please?

Should more likely be a `schedule("afterRender", () => ` I think

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+        post_id: attrs.post.id,
+        poll_name: attrs.poll.name,
+        user_field_name: attrs.groupedBy
+      }).then(result => {
+        for (
+          let chartIdx = 0;
+          chartIdx < result.grouped_results.length;
+          chartIdx++
+        ) {
+          const data = result.grouped_results[chartIdx].options.map(
```
jjaffeux: mapBy

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+            o => o.votes
+          );
+          const labels = result.grouped_results[chartIdx].options.map(
```
jjaffeux: mapBy

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,211 @@ createWidget("discourse-poll-info", {
...
+    if (error) {
+      popupAjaxError(error);
+    } else {
+      bootbox.alert(I18n.t("poll.error_while_fetching_voters"));
+    }
+  });
+}
+
+function transformUserFieldToLabel(fieldName) {
+  let transformed = fieldName.split("_");
```
jjaffeux: it's safer do do this:

```suggestion
  let transformed = fieldName.split("_").filter(Boolean);
```
Because:

```
"".split("_") // [""]
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
jjaffeux: Yes I don't like this later all, we got to find a better way

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
eviltrout: Yes what @jjaffeux said!

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
eviltrout: You can still use the `split` and `join`, but with `classify` in between.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +422,210 @@ createWidget("discourse-poll-info", {
...
+
+    let contents = [];
+
+    let selectInnerHtml = "";
+    attrs.groupableUserFields.forEach(field => {
+      selectInnerHtml += `<option value='${field}' ${
+        attrs.groupedBy === field ? "selected" : ""
+      } >${transformUserFieldToLabel(field)}</option>`;
+    });
+    contents.push(
```
eviltrout: Is there a reason you are not pushing HTML elements using the `h` helper? `RawHtml` is slower and should be avoided if you can.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -415,6 +423,209 @@ createWidget("discourse-poll-info", {
...
+            { value: field },
+            transformUserFieldToLabel(field)
+          );
+        })
+      )
+    );
+
+    contents.push(h("div.clearfix"));
+
+    later(() => {
```
jjaffeux: I'm sorry but I highly dislike this later + ajax inside html, `html` should be  a pure function more or less.

What you could do is having a `state.groupedPollResultsLoaded` and architecture your html like this:

```
init() {
  this.loadGroupedPollResults().then((results) => {
     this.state.groupedResults = results.grouped_results;
     this.state.isLoaded = true;
   });
},

html() {
  let contents = [];
  contents.push(foo);
  if (this.state.isLoaded) {
      content.push(this.attach("bar"));
  }
  return contents;
}
```

This is not complete code, but I think this should work and be cleaner.

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -2,13 +2,15 @@ import Controller from "@ember/controller";
...
   observes
-} from "ember-addons/ember-computed-decorators";
+} from "discourse-common/utils/decorators";
 import EmberObject from "@ember/object";
 
 export default Controller.extend({
   regularPollType: "regular",
   numberPollType: "number",
   multiplePollType: "multiple",
+  pieTypeIdentifier: "pie",
```
eviltrout: Sorry for the miscommunication here but if these values don't change they needn't be attached to every object. Instead what I'd recommend is creating `const PIE_TYPE = "pie";` outside of the declaration of this class and use that `const` for comparisons.

plugins/poll/assets/javascripts/controllers/poll-ui-builder.js.es6
```diff
@@ -20,6 +22,14 @@ export default Controller.extend({
     this._setupPoll();
   },
 
+  @computed("barTypeIdentifier", "pieTypeIdentifier")
```
eviltrout: Since `PIE_TYPE` won't ever change, you won't need it to be part of the computed property (which is slower). The dependent keys should only be things that can change.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -323,7 +327,11 @@ createWidget("discourse-poll-container", {
 
     if (attrs.showResults) {
       const type = poll.get("type") === "number" ? "number" : "standard";
-      return this.attach(`discourse-poll-${type}-results`, attrs);
+      const resultsWidget =
+        type === "number" || attrs.poll.chart_type !== "pie"
```
eviltrout: These should use the same `const` as the other file. You can `export const` and then import them here to compare.

plugins/poll/assets/stylesheets/common/poll.scss
eviltrout: Bumping this, @jjaffeux is right here.

PR #8353, opened 2019-11-14
UX: Add +n indicator in topic list for PMs with > 5 participants

pmusaraj:
Screenshot: 

image

The fourth avatar is dropped in order to make room for the +n element but also to keep the last avatar visible, since it is an indicator of the last poster.

### DIFF
--- a/app/serializers/topic_list_item_serializer.rb
+++ b/app/serializers/topic_list_item_serializer.rb
@@ -14,7 +14,8 @@ class TopicListItemSerializer < ListableTopicSerializer
              :bookmarked_post_numbers,
              :liked_post_numbers,
              :featured_link,
-             :featured_link_root_domain
+             :featured_link_root_domain,
+             :allowed_user_count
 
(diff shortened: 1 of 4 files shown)


coding-horror:
Oh this is a neat change 😁

app/serializers/topic_list_item_serializer.rb
```diff
@@ -86,4 +87,12 @@ def include_featured_link_root_domain?
     SiteSetting.topic_featured_link_enabled && object.featured_link.present?
   end
 
+  def allowed_user_count
+    return object.allowed_users.count
```
ZogStriP: You don't need the `return` here since it's the last statement of the method.

(Also, weird that rubocop didn't catch it)

pmusaraj: Thanks, I added the Rubocop role for it now in https://github.com/discourse/discourse/commit/be15abcf7f40ca01d8b9dfe57a0c3bd8ed9fe8d9

PR #8355, opened 2019-11-15
Block enforce second factor if enable local logins is disabled

martin-brennan:
https://meta.discourse.org/t/block-enforce-second-factor-if-enable-local-logins-is-disabled/124216

* Disallows enforcing 2FA if local logins are disabled
* Disallows disabling local logins if enforce 2FA is enabled


SamSaffron:
Looks good to me!

PR #8356, opened 2019-11-15
Refactor Group model

udan11:


app/assets/javascripts/discourse/models/group.js.es6
```diff
@@ -1,31 +1,31 @@
...
+import User from "discourse/models/user";
 
 const Group = RestModel.extend({
-  limit: 50,
-  offset: 0,
   user_count: 0,
+  limit: null,
+  offset: null,
+
+  request_count: 0,
```
ZogStriP:
```suggestion
  requestCount: 0,
```

app/assets/javascripts/discourse/mixins/load-more.js.es6
```diff
@@ -21,6 +21,7 @@ export default Mixin.create(Scrolling, {
     const eyeline = new Eyeline(this.eyelineSelector + ":last");
     this.set("eyeline", eyeline);
     eyeline.on("sawBottom", () => this.send("loadMore"));
+    eyeline.update(); // Force update once to consider current position
```
ZogStriP: This should be in another PR/commit as it's not really related to the group refactoring (remember we squash all the commits when we merge a PR)

ZogStriP:
LGTM 👍

app/assets/javascripts/discourse/models/group.js.es6
udan11: This is similar to `user_count` above. We are building the model from the exact response we get from the server.

ZogStriP:
Having some tests would be great here as well ;)

PR #8357, opened 2019-11-15
FIX: Ensure load-more considers current position

udan11:
If user is already at the bottom of the page, the loadMore action will
not be called.


ZogStriP:
Any changes we could have a test for that?

PR #8358, opened 2019-11-15
PERF: Do fewer queries when converting posts.

romanrizzi:
- Avoid using User#find and then fetching the user_stat association just to update a counter - We did this for every post in the topic.
- Avoid having to query the topic_allowed_user table every time we want to add a new topic_allowed_user.
- Use pluck to fetch just what we need.

### DIFF
--- a/app/models/topic_converter.rb
+++ b/app/models/topic_converter.rb
@@ -60,31 +60,30 @@ class TopicConverter
 
   private
 
+  def posters
+    @posters ||= @topic.posts.distinct.pluck(:user_id).to_a
+  end
+
   def update_user_stats
-    @topic.posts.where(deleted_at: nil).each do |p|
-      user = User.find(p.user_id)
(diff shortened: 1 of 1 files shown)


app/models/topic_converter.rb
```diff
@@ -60,31 +60,30 @@ def convert_to_private_message
 
   private
 
+  def posters
+    @posters ||= @topic.posts.where(deleted_at: nil).pluck(:user_id).to_a
```
danielwaterworth: `pluck` gives you an array and you can remove the duplicates by adding a `.distinct`.

romanrizzi: Nice suggestion, thanks!

eviltrout: Isn't `deleted_at: nil` the default? Unless you explicitly say `with_deleted`?

PR #8359, opened 2019-11-15
FEATURE: Normalize the service worker route

featheredtoast:
Update cache headers so they are not immutable outside of the rails app

Add the ability to purge the service worker cache from localhost

Rails -> nginx will pass immutable flags so the file is cached until reloaded.
In most cases, nginx will have its cache flushed on rebuild (new image)

For those needing dynamic re-caching (such as upgrading via the UI),
a rake task for flushing the service worker script is provided
through `assets:flush_sw`

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -456,8 +456,6 @@ module ApplicationHelper
   end
 
   def client_side_setup_data
-    service_worker_url = 'service-worker.js'
-
     setup_data = {
       cdn: Rails.configuration.action_controller.asset_host,
       base_url: Discourse.base_url,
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -456,7 +456,7 @@ def preloaded_json
   end
 
   def client_side_setup_data
-    service_worker_url = Rails.env.development? ? 'service-worker.js' : Rails.application.assets_manifest.assets['service-worker.js']
+    service_worker_url = 'service-worker.js'
```
ZogStriP: Any reasons this is not a constant?

featheredtoast: Alternatively we could just inlined in `setup_data` - does that work?

PR #8363, opened 2019-11-18
FIX: Ensure revisions are made to store edit reasons and no reasons get wiped

martin-brennan:
Meta: https://meta.discourse.org/t/merged-edits-will-clobber-the-first-edit-reason/123475

* Fix an issue where if an edit was made to a post with a reason provided, and then another edit was made with no reason, the original edit reason got wiped out
* We now always make a post revision (even with ninja edits) if an edit reason has been provided and it is different from the current edit reason

### DIFF
--- a/lib/post_revisor.rb
+++ b/lib/post_revisor.rb
@@ -130,8 +130,9 @@ class PostRevisor
     @fields[:user_id] = @fields[:user_id].to_i if @fields.has_key?(:user_id)
     @fields[:category_id] = @fields[:category_id].to_i if @fields.has_key?(:category_id)
 
-    # always reset edit_reason unless provided
-    @fields[:edit_reason] = nil unless @fields[:edit_reason].present?
+    # always reset edit_reason unless provided, do not set to nil else
+    # previous reasons are lost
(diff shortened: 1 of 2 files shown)


lib/post_revisor.rb
```diff
@@ -130,8 +130,9 @@ def revise!(editor, fields, opts = {})
     @fields[:user_id] = @fields[:user_id].to_i if @fields.has_key?(:user_id)
     @fields[:category_id] = @fields[:category_id].to_i if @fields.has_key?(:category_id)
 
-    # always reset edit_reason unless provided
-    @fields[:edit_reason] = nil unless @fields[:edit_reason].present?
+    # always reset edit_reason unless provided, do not set to nil else
+    # previous reasons are lost
+    @fields.delete(:edit_reason) unless @fields[:edit_reason].present?
```
SamSaffron:
```suggestion
    @fields.delete(:edit_reason) if @fields[:edit_reason].blank?
```

SamSaffron:
Nice fix, feel free to merge once minor comment is addressed

PR #8364, opened 2019-11-18
FIX: Automatically recover from bad sprockets cache in development

martin-brennan:
We were having issues in development mode where the JS code had errors due to a bad cache. When starting a server in development mode we now get the git sha of the discourse HEAD and get a git sha of all plugins, and store them in a file. If the sha has changed then we delete `tmp/cache` to refresh the assets cache.


SamSaffron:
This is fine to me... but maybe let's just move this into `bin/unicorn` that way we only trigger this code when you launch your web server.

martin-brennan:
@SamSaffron no problem I just wasn't sure whether the existing stuff needed to stay where it was. Will move to `bin/unicorn`

PR #8368, opened 2019-11-18
DEV: Submit forms via `type=submit` button, not explicit form action

davidtaylorhq:


app/assets/javascripts/discourse/templates/preferences-second-factor.hbs
```diff
@@ -143,7 +143,8 @@
             type="submit"
```
jjaffeux: seems like `type="submit"` was already there ?

davidtaylorhq: You're right, I removed the duplicate. It is still working fine for me

PR #8369, opened 2019-11-18
DEV: Add test for load-more

udan11:
Follow-up to 3650c64bca2ce6d4012598446e10278d8fdda75e.

### DIFF
--- a/app/assets/javascripts/discourse/lib/eyeline.js.es6
+++ b/app/assets/javascripts/discourse/lib/eyeline.js.es6
@@ -1,21 +1,42 @@
 import ENV from "discourse-common/config/environment";
 import { EventTarget } from "rsvp";
 
-//  Track visible elemnts on the screen.
+let _skipUpdate;
+let _rootElement;
+
+export function configureEyeline(opts) {
+  if (opts) {
+    _skipUpdate = opts.skipUpdate;
+    _rootElement = opts.rootElement;
+  } else {
+    _skipUpdate = ENV.environment === "test";
+    _rootElement = null;
+  }
+}
+
+configureEyeline();
+
+//  Track visible elements on the screen.
 const Eyeline = function Eyeline(selector) {
   this.selector = selector;
 };
 
 Eyeline.prototype.update = function() {
-  if (ENV.environment === "test") {
+  if (_skipUpdate) {
     return;
   }
 
-  const docViewTop = $(window).scrollTop(),
-    windowHeight = $(window).height(),
-    docViewBottom = docViewTop + windowHeight,
-    $elements = $(this.selector),
-    bottomOffset = $elements.last().offset();
+  const docViewTop = _rootElement
+    ? $(_rootElement).scrollTop()
+    : $(window).scrollTop();
+  const windowHeight = _rootElement
+    ? $(_rootElement).height()
+    : $(window).height();
+  const docViewBottom = docViewTop + windowHeight;
+  const $elements = $(this.selector);
+  const bottomOffset = _rootElement
+    ? $elements.last().position()
+    : $elements.last().offset();
 
   let atBottom = false;
   if (bottomOffset) {
@@ -25,7 +46,7 @@ Eyeline.prototype.update = function() {
 
   return $elements.each((i, elem) => {
     const $elem = $(elem),
-      elemTop = $elem.offset().top,
+      elemTop = _rootElement ? $elem.position().top : $elem.offset().top,
       elemBottom = elemTop + $elem.height();
 
     let markSeen = false;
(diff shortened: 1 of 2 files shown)


test/javascripts/components/load-more-test.js.es6
```diff
@@ -0,0 +1,28 @@
...
+      <table class='numbers'>
+        <tr><td>1</td></tr>
+        <tr><td>2</td></tr>
+        <tr><td>3</td></tr>
+      </table>
+    {{/load-more}}`,
+
+  beforeEach() {
+    this.set("loadMore", () => this.set("loadedMore", true));
+    _setTestParams(false, true);
```
jjaffeux: I would prefer a hash, you have to dig to know what are these `false` `true` atm, something like:

```
configureEyeLineBehavior({foo: true, bar: false});
```

Would be better I think. I also changed the name, because it's used all the time and not only in tests now, even if you created it for tests, its real goal is to tweak eye line behaviour.

jjaffeux: Also I'm wondering if we couldn't use: `Discourse.rootElement` to have all of this automatically working?

test/javascripts/components/load-more-test.js.es6
```diff
@@ -0,0 +1,28 @@
+import { _setTestParams } from "discourse/lib/eyeline";
+import componentTest from "helpers/component-test";
+
+moduleForComponent("load-more", { integration: true });
+
+componentTest("preview updates with markdown", {
+  template: `
+    {{#load-more selector=".numbers tr" action=(action loadMore)}}
```
jjaffeux: I think this should work for both cases:

```
action=loadMore
action=(action "loadMore")
```

and in this case `action=loadMore` should be enough

test/javascripts/components/load-more-test.js.es6
```diff
@@ -0,0 +1,28 @@
...
+      </table>
+    {{/load-more}}`,
+
+  beforeEach() {
+    this.set("loadMore", () => this.set("loadedMore", true));
+    _setTestParams(false, true);
+  },
+
+  afterEach() {
+    _setTestParams(true);
```
jjaffeux: might be worth to have a:  `defaultEyeLineBehavior` function to call, so we don't have to worry about default eyeline behavior changing in the future, which could lead to hard to track side effects.

test/javascripts/components/load-more-test.js.es6
```diff
@@ -0,0 +1,28 @@
...
+  },
+
+  async test(assert) {
```
jjaffeux: is it `async` though? if we don't use `await` inside?

test/javascripts/components/load-more-test.js.es6
```diff
@@ -0,0 +1,28 @@
+import { _setTestParams } from "discourse/lib/eyeline";
+import componentTest from "helpers/component-test";
+
+moduleForComponent("load-more", { integration: true });
+
+componentTest("preview updates with markdown", {
+  template: `
+    {{#load-more selector=".numbers tr" action=(action loadMore)}}
+      <table class='numbers'>
+        <tr><td>1</td></tr>
```
jjaffeux: do we need all the lines to test the behavior?

test/javascripts/components/load-more-test.js.es6
udan11: This makes sense. I implemented the `configureEyeline` function.

Regarding the other comment, `.position()` and `.offset()` return slightly different results (relative to parent vs document) and I recommend this to be used only during testing.

test/javascripts/components/load-more-test.js.es6
udan11: Thanks, it makes sense!

test/javascripts/components/load-more-test.js.es6
udan11: Calling `configureEyeline` with no parameter will reset its behaviour.

test/javascripts/components/load-more-test.js.es6
udan11: Nope. :)

test/javascripts/components/load-more-test.js.es6
udan11: I significantly simplified the table to keep a single empty `tr`.

app/assets/javascripts/discourse/lib/eyeline.js.es6
```diff
@@ -1,21 +1,42 @@
 import ENV from "discourse-common/config/environment";
 import { EventTarget } from "rsvp";
 
-//  Track visible elemnts on the screen.
+let _skipUpdate;
+let _rootElement;
+
+export function configureEyeline(opts) {
+  if (opts === undefined) {
```
jjaffeux: can't we do:

```
if (opts) {
    _skipUpdate = opts.skipUpdate;
    _rootElement = opts.rootElement;
} else {
      _skipUpdate = ENV.environment === "test";
    _rootElement = null;
}
```

jjaffeux:
LGTM, nice :+1:

PR #8370, opened 2019-11-18
FEATURE: Add export poll button

xfalcox:
This PR aims to make poll results easily exportable to staff in a CSV format, so they can be analyzed in external software.

It also makes the export data easily customizable by allowing users to leverage any data explorer query to generate the report. By default, we use a query that ships with data explorer, but user can change the ID in settings or use `0` to disable this feature.

One potential upgrade is using the recent work that allows arbitrary group to run data explorer and allow all the groups with access to the configured query to also export polls, but that can be added later.

### DIFF
--- a/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
+++ b/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
@@ -720,7 +720,7 @@ export default createWidget("discourse-poll", {
       const blob = new Blob([csvContent],{type: 'text/csv;charset=utf-8;'});
       const url = URL.createObjectURL(blob);
       pom.href = url;
-      pom.setAttribute('download', 'poll.csv');
+      pom.setAttribute("download", "poll.csv");
       pom.click();
     });
   },
--- a/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
+++ b/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
@@ -716,12 +716,13 @@ export default createWidget("discourse-poll", {
         download: 1
       }
     }).then(csvContent => {
-      const pom = document.createElement('a');
+      const downloadLink = document.createElement('a');
       const blob = new Blob([csvContent],{type: 'text/csv;charset=utf-8;'});
       const url = URL.createObjectURL(blob);
-      pom.href = url;
-      pom.setAttribute("download", "poll.csv");
-      pom.click();
+      downloadLink.href = url;
+      downloadLink.setAttribute("download", `poll-export-${attrs.poll.name}-${attrs.post.id}.csv`);
+      downloadLink.click();
+      downloadLink.remove();
     });
   },
 
--- a/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
+++ b/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
@@ -425,6 +425,7 @@ createWidget("discourse-poll-buttons", {
     const closed = attrs.isClosed;
     const staffOnly = poll.results === "staff_only";
     const isStaff = this.currentUser && this.currentUser.staff;
+    const dataExplorerEnabled = this.siteSettings.data_explorer_enabled === true;
     const hideResultsDisabled = !staffOnly && (closed || topicArchived);
 
     if (attrs.isMultiple && !hideResultsDisabled) {
@@ -473,6 +474,19 @@ createWidget("discourse-poll-buttons", {
           })
         );
       }
+    }
+
+    if (isStaff && dataExplorerEnabled && poll.get("voters") > 0) {
+      contents.push(
+        this.attach("button", {
+          className: "btn btn-default toggle-results",
+          label: "poll.export-results.label",
+          title: "poll.export-results.title",
+          icon: "download",
+          disabled: poll.get("voters") === 0,
+          action: "exportResults"
+        })
+      );
     }
 
     if (poll.get("close")) {
@@ -683,6 +697,32 @@ export default createWidget("discourse-poll", {
 
   toggleResults() {
     this.state.showResults = !this.state.showResults;
+  },
+
+  exportResults() {
+    const { attrs } = this;
+
+    // This uses the Data Explorer plugin export as CSV route
+    // There is detection to check if the plugin is enabled before showing the button
+    ajax("/admin/plugins/explorer/queries/-16/run.csv", {
+      type: "POST",
+      data: {
+        params: JSON.stringify({ // needed for data-explorer route compatibility
+          poll_name: attrs.poll.get("name"),
+          post_id: attrs.post.id.toString() // needed for data-explorer route compatibility
+        }),
+        explain: false,
+        limit: 1000000,
+        download: 1
+      }
+    }).then(( csvContent ) => {
+      const pom = document.createElement('a');
+      const blob = new Blob([csvContent],{type: 'text/csv;charset=utf-8;'});
+      const url = URL.createObjectURL(blob);
+      pom.href = url;
+      pom.setAttribute('download', 'poll.csv');
+      pom.click();
+    });
   },
 
   showLogin() {
--- a/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
+++ b/plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
@@ -425,7 +425,7 @@ createWidget("discourse-poll-buttons", {
     const closed = attrs.isClosed;
     const staffOnly = poll.results === "staff_only";
     const isStaff = this.currentUser && this.currentUser.staff;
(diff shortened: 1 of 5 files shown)


plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -475,6 +476,19 @@ createWidget("discourse-poll-buttons", {
...
     }
 
+    if (isStaff && dataExplorerEnabled && poll.get("voters") > 0) {
+      contents.push(
+        this.attach("button", {
+          className: "btn btn-default toggle-results",
+          label: "poll.export-results.label",
+          title: "poll.export-results.title",
+          icon: "download",
+          disabled: poll.get("voters") === 0,
```
jjaffeux:
```suggestion
          disabled: poll.voters === 0,
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +699,32 @@ export default createWidget("discourse-poll", {
...
+  exportResults() {
+    const { attrs } = this;
+
+    // This uses the Data Explorer plugin export as CSV route
+    // There is detection to check if the plugin is enabled before showing the button
+    ajax("/admin/plugins/explorer/queries/-16/run.csv", {
+      type: "POST",
+      data: {
+        params: JSON.stringify({ // needed for data-explorer route compatibility
+          poll_name: attrs.poll.get("name"),
```
jjaffeux:
```suggestion
          poll_name: attrs.poll.name,
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +699,32 @@ export default createWidget("discourse-poll", {
...
+        explain: false,
+        limit: 1000000,
+        download: 1
+      }
+    }).then(( csvContent ) => {
+      const pom = document.createElement('a');
+      const blob = new Blob([csvContent],{type: 'text/csv;charset=utf-8;'});
+      const url = URL.createObjectURL(blob);
+      pom.href = url;
+      pom.setAttribute('download', 'poll.csv');
```
jjaffeux: be sure to configure prettier, this should (and other places of your code) with double quotes.

```suggestion
      pom.setAttribute("download", "poll.csv");
```

jjaffeux: maybe we should have `poll-${topicId}-${postId}.csv` or something like that

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +699,32 @@ export default createWidget("discourse-poll", {
...
+        params: JSON.stringify({ // needed for data-explorer route compatibility
+          poll_name: attrs.poll.get("name"),
+          post_id: attrs.post.id.toString() // needed for data-explorer route compatibility
+        }),
+        explain: false,
+        limit: 1000000,
+        download: 1
+      }
+    }).then(( csvContent ) => {
+      const pom = document.createElement('a');
```
jjaffeux: what does pom means here ?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -425,6 +425,7 @@ createWidget("discourse-poll-buttons", {
     const closed = attrs.isClosed;
     const staffOnly = poll.results === "staff_only";
     const isStaff = this.currentUser && this.currentUser.staff;
+    const dataExplorerEnabled = this.siteSettings.data_explorer_enabled === true;
```
jjaffeux: do we need explicit check on true here ?

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -475,6 +476,19 @@ createWidget("discourse-poll-buttons", {
       }
     }
 
+    if (isStaff && dataExplorerEnabled && poll.get("voters") > 0) {
```
jjaffeux:
```suggestion
    if (isStaff && dataExplorerEnabled && poll.voters > 0) {
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +699,32 @@ export default createWidget("discourse-poll", {
...
+      data: {
+        params: JSON.stringify({ // needed for data-explorer route compatibility
+          poll_name: attrs.poll.get("name"),
+          post_id: attrs.post.id.toString() // needed for data-explorer route compatibility
+        }),
+        explain: false,
+        limit: 1000000,
+        download: 1
+      }
+    }).then(( csvContent ) => {
```
jjaffeux:
```suggestion
    }).then(csvContent => {
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
xfalcox: Yeah, the data explorer can be either missing (plugin not installed) or false (plugin disabled)

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
xfalcox: good idea!

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
xfalcox: renamed, that was just lazy copy & paste :stuck_out_tongue:

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
jjaffeux: so it will work for a if anyways:

if (false) // not called
if (undefined) // not called

xfalcox: cool, fixed

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -475,6 +477,19 @@ createWidget("discourse-poll-buttons", {
       }
     }
 
+    if (isStaff && dataExplorerEnabled && poll.voters > 0 && exportQueryID !== 0) {
```
jjaffeux:
```suggestion
    if (isStaff && dataExplorerEnabled && poll.voters > 0 && exportQueryID) {
```

if (0) // not called

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +700,40 @@ export default createWidget("discourse-poll", {
...
+        params: JSON.stringify({ // needed for data-explorer route compatibility
+          poll_name: attrs.poll.name,
+          post_id: attrs.post.id.toString() // needed for data-explorer route compatibility
+        }),
+        explain: false,
+        limit: 1000000,
+        download: 1
+      }
+    }).then(csvContent => {
+      const downloadLink = document.createElement('a');
```
jjaffeux: still needs prettier, various things are not formatted correctly like these quotes

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +700,40 @@ export default createWidget("discourse-poll", {
...
+      const blob = new Blob([csvContent],{type: 'text/csv;charset=utf-8;'});
+      const url = URL.createObjectURL(blob);
+      downloadLink.href = url;
```
jjaffeux: nitpick, but I guess this could be put in one line, there's no value for that intermediate property here:

```
downloadLink.href = URL.createObjectURL(blob);
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -685,6 +700,40 @@ export default createWidget("discourse-poll", {
...
+      downloadLink.setAttribute("download", `poll-export-${attrs.poll.name}-${attrs.post.id}.csv`);
```
jjaffeux: I was thinking maybe post_number would be better

xfalcox: name and post_id are the primary keys for polls, so this will guarantee unique named files.

jjaffeux: ok all good then :+1:

ZogStriP:
I think we should have at least one "working" integration test.

xfalcox:
> I think we should have at least one "working" integration test.

It uses a data-explorer route, so the test would fail if data-explorer is missing. Wouldn't that be bad?

PR #8371, opened 2019-11-19
FIX: When dismissing category inform via MessageBus

lis2:
When the category is dismissed, `dismiss_new` message is sent to fronted to clean state.

In addition, I noticed that when old dismiss new button is clicked, no message is sent so I decided to kill two birds with one stone.

Screencast-from-19-11-19-02_49_03

### DIFF
--- a/app/assets/javascripts/discourse/models/topic-tracking-state.js.es6
+++ b/app/assets/javascripts/discourse/models/topic-tracking-state.js.es6
@@ -28,7 +28,7 @@ function isUnread(topic) {
 }
 
 function isUnseen(topic) {
-  return !topic.last_seen_at || topic.last_seen_at < topic.created_at;
+  return !topic.is_seen;
 }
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/models/topic-tracking-state.js.es6
```diff
@@ -28,6 +29,10 @@ function isUnread(topic) {
   );
 }
 
+function isUnseen(topic) {
+  return !topic.last_seen_at || topic.last_seen_at < topic.created_at;
```
ZogStriP: When is `topic.last_seen_at` ever `< topic.created_at`?

lis2: @ZogStriP thank you, very good point, I switched code to simply mark each post as read for a specific category.

PR #8372, opened 2019-11-19
FIX: parallel spec system needs needs a dedicated upload folder for each worker.

vinothkannans:


lib/file_store/base_store.rb
```diff
@@ -31,7 +31,16 @@ def remove_file(url, path)
     end
 
     def upload_path
-      File.join("uploads", RailsMultisite::ConnectionManagement.current_db)
+      return File.join("uploads", RailsMultisite::ConnectionManagement.current_db) unless ENV['RAILS_ENV'] == "test" && ENV['TEST_ENV_NUMBER']
```
SamSaffron:
```suggestion
      if  ENV['RAILS_ENV'] != "test" || !ENV['TEST_ENV_NUMBER']
           return File.join("uploads", RailsMultisite::ConnectionManagement.current_db)
      end
```

lib/file_store/base_store.rb
```diff
@@ -31,7 +31,16 @@ def remove_file(url, path)
...
+
+      # Parallel spec system
+      if ENV['TEST_ENV_NUMBER'] == ''
+        n = "1"
+      else
+        n = ENV['TEST_ENV_NUMBER']
+      end
```
lis2: Rails is givine quite nice helper method called `presence` https://apidock.com/rails/Object/presence

so I think that this may become
n = ENV['TEST_ENV_NUMBER'].presence || '1'

danielwaterworth: This isn't a like-for-like change. It differs if `ENV['TEST_ENV_NUMBER']` is `nil`.

vinothkannans: I understand. But it won't affect here. Since we already checking the `nil` possibility [here](https://github.com/discourse/discourse/pull/8372/files#diff-b9d1dc76a883b70545074540b1d77fa9R34-R35).

PR #8377, opened 2019-11-20
FEATURE: improve email change workflow

SamSaffron:
- Show old and new email address during the process
- Ensure correct user is logged on when attempting to make email changes
- Support reloading a page during the email reset process without resubmit
of form
- Clean up tests

internal refactor of single action into 4 distinct actions that are simpler
to reason about.

This also removes the step that logs on an account after you confirm an
email change, since it is no longer needed which leaves us with safer
internals.

This left me no choice but to amend translations cause the old route was
removed.


SamSaffron:
@eviltrout this need some careful review... don't merge quite yet

Very happy with the end result here, but it touches a lot of stuff and I needed to fuss with migrations

gschlager:
> I needed to fuss with translations

I'd suggest adding a DB migration to delete or modify translation overrides when they contain the old `authorize-email` route.

eviltrout:
@SamSaffron in the future you can create a Draft PR. It makes it unmergeable until you mark it as finished.

app/controllers/users_email_controller.rb
```diff
@@ -29,38 +47,142 @@ def update
...
+
+    if @token
+      if type == :old
+        @change_request = @token.user&.email_change_requests.where(old_email_token_id: @token.id).first
+      elsif type == :new
+        @change_request = @token.user&.email_change_requests.where(new_email_token_id: @token.id).first
+      end
+    end
+
+    if (!@token || !@token&.user || !@change_request)
```
eviltrout:
```suggestion
    if (!@token&.user || !@change_request)
```

No need to check for token if you use the `&` operator.

app/views/users_email/show_confirm_new_email.html.erb
```diff
@@ -0,0 +1,54 @@
+<div id="simple-container">
+  <% if @done %>
+    <h2>
+      <%= t 'change_email.confirmed' %>
+    </h2>
+    <p>
+      <a class="btn" href="/"><%= t('change_email.please_continue', site_name: SiteSetting.title) %></a>
```
eviltrout: This path (`/`) is not subfolder aware.

app/views/users_email/show_confirm_new_email.html.erb
```diff
@@ -0,0 +1,54 @@
...
+    </p>
+  <% elsif @error %>
+    <div class='alert alert-error'>
+      <%= @error %>
+    </div>
+  <% else %>
+    <h2><%= t 'change_email.authorizing_new.title' %></h2>
+    <p>
+      <%= t 'change_email.authorizing_new.description' %>
+      <br>
```
eviltrout: Can we use CSS instead of multiple `<br>` tags?

SamSaffron:
This is now merged in. @gschlager I am option to play this by ear regarding the migration, it just feels odd to carry a once off or a migration here for this enormous edge case

@eviltrout tried to address all your comments, thanks for the review!

eviltrout:
@SamSaffron why did you close this and make a new commit? This means that github loses the linking between the PR and the commit. Also I can't see what changed after my suggestions!

SamSaffron:
Apologies @eviltrout I just force pushed here so you can see the full context ... but merging ain't going to fly since it is already merged

was being lazy cause it avoided me needing to force push to my repo.

PR #8379, opened 2019-11-20
FIX: Use correct MIME type for theme exports

udan11:
We use `ThemeStore::ZipExporter` which always exports a `.zip` archive. This fix is for some download managers that will change the file extension based on the `Content-Type`.


udan11:
I could not find any. I searched for `content_type` and `application/`.

PR #8390, opened 2019-11-21
FIX: Remove unneeded  from discourse loader

markvanlan:


jjaffeux:
minor but this should be prefixed REFACTOR or even DEV as it’s not actually fixing anything 😁

markvanlan:
@jjaffeux Oh no, it broke literally everything :)

My message got messed up somehow. It should have said removed unneeded "This"

jjaffeux:
Ok cool then 👍

PR #8391, opened 2019-11-21
FIX: Show quoted images correctly.

udan11:
This commit attempts to fix two issues that affect quoted images.

The first issue is observed while loading. The 'position: absolute' CSS
property makes 'width' and 'height' behave differently. Instead of using
the known image size, this makes it use the computed width and height of
the image, which should be the right size, as shown to the user.

The second issue is caused by 'object-fit: cover' property which trimmed
the left and right sides of wide pictures to make them fit inside the
quote.

### DIFF
--- a/app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
+++ b/app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
@@ -56,16 +56,34 @@ function show(image) {
     copyImg.className = imageData.className;
 
     let inOnebox = false;
+    let inQuote = false;
     for (let element = image; element; element = element.parentElement) {
+      if (element.tagName === "ARTICLE") {
+        break;
+      }
       if (element.classList.contains("onebox")) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/lib/lazy-load-images.js.es6
```diff
@@ -68,6 +68,28 @@ function show(image) {
       copyImg.style.height = `${imageData.height}px`;
     }
 
+    if (imageData.width && imageData.height) {
+      let inQuote = false;
+      for (let element = image; element; element = element.parentElement) {
```
ZogStriP: Does this climb the entire DOM tree up to the root? Maybe we should bail once we reach out of the post body?

udan11: I merged this for-loop with the previous one and made it stop when getting out of the post body.

PR #8392, opened 2019-11-21
FEATURE: New API to apply custom filters to the review queue

romanrizzi:
This PR adds a new API to include custom filters in the review queue. An example can be found [here](https://github.com/discourse/discourse-assign/pull/58).


app/controllers/reviewables_controller.rb
```diff
@@ -55,7 +56,7 @@ def index
       end,
       meta: filters.merge(
         total_rows_reviewables: total_rows, types: meta_types, reviewable_types: Reviewable.types,
-        reviewable_count: Reviewable.list_for(current_user).count
+        reviewable_count: Reviewable.list_for(current_user).count,
```
eviltrout: I think this line was a mistake. No comma required.

PR #8393, opened 2019-11-21
FIX: method names incorrect

angusmcleod:
The placeholder computed property is ``computedPlaceholder`` (see: https://github.com/discourse/discourse/blob/master/app/assets/javascripts/select-kit/templates/components/select-kit/select-kit-filter.hbs#L4). These methods may have been incorrectly named in a "replace all" of ``computed`` with ``discourseComputed``.


jjaffeux:
Yes indeed looks like a find and replace has been too aggressive here :p Thanks

PR #8397, opened 2019-11-22
Changed order of flags_given_received_count in de translation

soerface:
The translation is shown alongside the counter. In English it is translated as "given / received", so the words correspond to the order of the numbers. The order of the German words is swapped.


ZogStriP:
Thanks @soerface but all translations have to be done on [transifex](https://www.transifex.com/discourse/discourse-org/) otherwise they would be overriden the next time we pull them.

PR #8406, opened 2019-11-25
FIX: Don't error when the empty current value in dif

lis2:
If current value is nil we should use `&.` combined with `dig` to protect diff from erroring


SamSaffron:
We should probably add 1 test here to cover this (or amend one test to explode if not for this fix)

lis2:
Test added, I also checked on my local with real-life scenario. It is happening when the theme is deleted and a new value is blank

PR #8408, opened 2019-11-25
FEATURE: Make site texts controller handle pages and locales

nbianca:


app/controllers/admin/site_texts_controller.rb
```diff
@@ -24,7 +24,11 @@ def index
       extras[:recommended] = true
       results = self.class.preferred_keys.map { |k| record_for(k) }
     else
-      results = find_translations(query, overridden)
+      if locale = params[:locale].presence
+        results = I18n.with_locale(locale) { find_translations(query, overridden) }
```
ZogStriP: What happens if `params[:locale]` is some gibberish?

app/controllers/admin/site_texts_controller.rb
```diff
@@ -41,8 +45,13 @@ def index
       end
     end
 
-    extras[:has_more] = true if results.size > 50
-    render_serialized(results[0..49], SiteTextSerializer, root: 'site_texts', rest_serializer: true, extras: extras, overridden_keys: overridden_keys)
+    per_page = 50
+    page = params[:page].to_i
```
ZogStriP: What if `params[:page]` is `-99`?

spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -58,6 +58,28 @@
...
+
+        get "/admin/customize/site_texts.json", params: { q: 'e' }
+        JSON.parse(response.body)['site_texts'].each { |text| texts << text['id'] }
+        expect(texts.size).to eq(100)
+      end
+
+      it 'works with locales' do
+        get "/admin/customize/site_texts.json", params: { q: 'yes_value', locale: 'en' }
+        value = obj['site_texts'].find { |text| text['id'] == 'js.yes_value' }['value']
+        expect(value).to eq('Yes')
```
ZogStriP: This test will fail if/when the translation changes. It's best to call `I18n.t()` with the `en` locale to make this test less fragile.

spec/requests/admin/site_texts_controller_spec.rb
```diff
@@ -58,6 +58,28 @@
...
+
+        get "/admin/customize/site_texts.json", params: { q: 'yes_value', locale: 'de' }
+        value = obj['site_texts'].find { |text| text['id'] == 'js.yes_value' }['value']
+        expect(value).to eq('Ja')
```
ZogStriP: Ditto about the test "fragility"

ZogStriP:
Why is this change needed? Can you add a description explaining why the change and how it solves the problem?

app/controllers/admin/site_texts_controller.rb
```diff
@@ -41,8 +45,15 @@ def index
...
+    first = page * per_page
+    last = first + per_page
+
+    raise Discourse::InvalidParameters.new(:page) if page < 0
```
ZogStriP: If `page < 0`, then you have calculated `first` and `last` for nothing 😉

nbianca: I wanted to keep all computations in one place hoping to make things more clear. 😊

app/controllers/admin/site_texts_controller.rb
```diff
@@ -41,8 +45,16 @@ def index
       end
     end
 
-    extras[:has_more] = true if results.size > 50
-    render_serialized(results[0..49], SiteTextSerializer, root: 'site_texts', rest_serializer: true, extras: extras, overridden_keys: overridden_keys)
+    per_page = 50
```
ZogStriP: This one could also be moved near where it's being used after the `raise` 😉

PR #8409, opened 2019-11-25
FIX: Improve downsize_uploads

CvX:
* Actually remove original large-sized images
* Don't save processed files if their size increased
* Prevent inconsistent state


script/downsize_uploads.rb
```diff
@@ -77,23 +32,72 @@
     new_file = false
   end
 
+  before = upload.filesize
   upload.filesize = File.size(path)
+  delta = upload.filesize - before
```
ZogStriP: You don't actually need that `delta` variable ;)

```ruby
return if before < upload.filesize
```

or 

```ruby
return if upload.filesize > before
```

script/downsize_uploads.rb
```diff
@@ -77,23 +32,72 @@
...
+
+  original_upload.posts.each do |post|
+    post.update!(raw: post.raw.gsub(original_upload.short_url, upload.short_url))
     Jobs.enqueue(:process_post, post_id: post.id, bypass_bump: true, cook: true)
   end
+
+  if new_file
+    Discourse.store.remove_upload(original_upload)
+  else
+    User.where(uploaded_avatar_id: original_upload.id).update_all(uploaded_avatar_id: upload.id)
```
ZogStriP: Are these the only ones where we use an `upload_id`? What about some of the uploads settings? What about the user profiles?

script/downsize_uploads.rb
CvX: Ah, yes, that was a leftover from a *log-everything* version 😉

script/downsize_uploads.rb
CvX: Those were taken from [Upload's after_destroy](https://github.com/discourse/discourse/blob/master/app/models/upload.rb#L33-L37)
That probably should be updated as well.

ZogStriP: Yeah, have a look at [this query](https://github.com/discourse/discourse/blob/c75ebfee579e502f42cb7ea686ced91f31e654db/app/jobs/scheduled/clean_up_uploads.rb#L54-L80) I think it's the most up to date.

script/downsize_uploads.rb
```diff
@@ -77,23 +33,103 @@
...
+
+    User.where(uploaded_avatar_id: original_upload.id).update_all(uploaded_avatar_id: upload.id)
+    UserAvatar.where(gravatar_upload_id: original_upload.id).update_all(gravatar_upload_id: upload.id)
+    UserAvatar.where(custom_upload_id: original_upload.id).update_all(custom_upload_id: upload.id)
+    UserProfile.where(profile_background_upload_id: original_upload.id).update_all(profile_background_upload_id: upload.id)
+    UserProfile.where(card_background_upload_id: original_upload.id).update_all(card_background_upload_id: upload.id)
+    Category.where(uploaded_logo_id: original_upload.id).update_all(uploaded_logo_id: upload.id)
+    Category.where(uploaded_background_id: original_upload.id).update_all(uploaded_background_id: upload.id)
+    CustomEmoji.where(upload_id: original_upload.id).update_all(upload_id: upload.id)
+    ThemeField.where(upload_id: original_upload.id).update_all(upload_id: upload.id)
```
SamSaffron: can you confirm we have indexes on all these tables (especially user avatar / profile)

SamSaffron:
I would love to see this merged, all the fixes sound great!

script/downsize_uploads.rb
CvX: @SamSaffron `User` and `UserAvatar` have the correct indexes. `UserProfile` needed an update (https://github.com/discourse/discourse/pull/8691). `Category`, `CustomEmoji`, and `ThemeField` don't have indexes on those columns - should I add them?

SamSaffron: I think we can live with just user and user  avatar and user profile for now, people are only going to rarely run this script so it is not a huge deal if it takes a bit longer

PR #8410, opened 2019-11-25
FIX: improves recurring by ensuring DST is computed correctly

jjaffeux:


plugins/discourse-local-dates/assets/javascripts/discourse-local-dates.js.no-module.es6
```diff
@@ -199,22 +199,37 @@
...
     const parts = recurring.split(".");
     const count = parseInt(parts[0], 10);
     const type = parts[1];
     const diff = moment().diff(dateTime, type);
     const add = Math.ceil(diff + count);
-    const wasDST = dateTime.isDST();
-    let dateTimeWithRecurrence = dateTime.add(add, type);
-    const isDST = dateTimeWithRecurrence.isDST();
 
+    // we create new moment object object from format
```
eviltrout:
```suggestion
    // we create new moment object from format
```

PR #8411, opened 2019-11-25
FEATURE: Limit the number of active sessions for a user

davidtaylorhq:
If a user has more than 100 active sessions, the oldest sessions will be terminated automatically and the user will be shown a popup warning.

This protects performance when logging in and when loading the list of recently used devices.

Screenshot 2019-11-25 at 17 59 05


coding-horror:
100 seems like a high limit here, who could really log in via 100 different
devices? Maybe 50 is a better default limit?

davidtaylorhq:
I think the place this is most likely to be hit is in corporate situations where people's browser cookies are deleted every time they log out of the computer. In this case, someone might 'create a new session' every day when they arrive at work. As far as Discourse is concerned, these are all active sessions.

Our default session length is 60 days, so I think we should set the limit higher than 60 to avoid 'crying wolf' at people who are forced to log in every day.

coding-horror:
Ok 60 is fine then

app/models/user_auth_token.rb
```diff
@@ -220,6 +222,16 @@ def rotate!(info = nil)
...
   end
+
+  def self.enforce_session_count_limit!(user_id)
+    tokens_to_destroy = where(user_id: user_id).
+      where('rotated_at > ?', SiteSetting.maximum_session_age.hours.ago).
+      order("rotated_at DESC").offset(MAX_SESSION_COUNT)
+
+    count = tokens_to_destroy.delete_all # Returns the number of deleted rows
+
+    count
```
eviltrout: this `count` is redundant. In fact, no `count` variable is required because ruby will return the result from `delete_all` - if you're worried it's not clear the comment is there, plus you have a test!

config/locales/client.en.yml
```diff
@@ -3136,6 +3136,10 @@ en:
     safe_mode:
       enabled: "Safe mode is enabled, to exit safe mode close this browser window"
 
+    too_many_sessions:
+      one: You were logged in to more than %{limit} devices. The oldest %{count} session has been logged out. Review recently used devices in your <a href="%{url}">account preferences</a>.
```
```suggestion
      one: You were logged in to more than %{limit} devices. The oldest session has been logged out. You may review your recently used devices in your <a href="%{url}">account preferences</a>.
```
I don't think we need to write "The oldest 1 session" here.

config/locales/client.en.yml
```diff
@@ -3136,6 +3136,10 @@ en:
...
+      other: You were logged in to more than %{limit} devices. The oldest %{count} sessions have been logged out. Review recently used devices in your <a href="%{url}">account preferences</a>.
```
```suggestion
      other: You were logged in to more than %{limit} devices. The oldest %{count} sessions have been logged out. You may review your recently used devices in your <a href="%{url}">account preferences</a>.
```

spec/components/auth/default_current_user_provider_spec.rb
```diff
@@ -586,6 +586,38 @@ def provider(url, opts = nil)
     expect(UserAuthToken.where(user_id: user.id).count).to eq(2)
   end
 
+  it "logging on user cleans up old sessions" do
```
eviltrout:
```suggestion
  it "cleans up old sessions when a user logs in" do
```

spec/components/auth/default_current_user_provider_spec.rb
```diff
@@ -586,6 +586,38 @@ def provider(url, opts = nil)
...
+    user = Fabricate(:user)
+
+    yesterday = Time.zone.now - 1.day
```
eviltrout:
```suggestion
    yesterday = 1.day.ago
```

spec/components/auth/default_current_user_provider_spec.rb
```diff
@@ -586,6 +586,38 @@ def provider(url, opts = nil)
...
+        created_at: yesterday + i.seconds,
+        updated_at: yesterday + i.seconds,
+        rotated_at: yesterday + i.seconds,
+        prev_auth_token: "abc#{i}",
+        auth_token: "abc#{i}"
+      }
+    end)
+
+    # Has 2 more sessions than allowed
+    expect(UserAuthToken.where(user_id: user.id).count).to eq(UserAuthToken::MAX_SESSION_COUNT + 2)
```
eviltrout: Assuming `insert_all` is tested elsewhere, I don't think you need to test that the Auth Tokens are created successfully by this test.

config/locales/client.en.yml
gschlager: Those translations are kinda problematic as they contain 2 numbers. I recommend using [Message Format](https://meta.discourse.org/t/message-format-support-for-localization/7035), because the alternative of splitting the string into two parts isn't great either.

app/models/user_auth_token.rb
davidtaylorhq: Ah yes, I added the count variable for some puts debugging and forgot about it. Will remove it

config/locales/client.en.yml
davidtaylorhq: I did think about that, but I decided `limit` was a constant equal to 100 (or a similarly high number), so there is never a need for singular/plural on that variable. Do you think we still need to use message format? Do some languages have different grammar even for higher numbers?

CvX:
@davidtaylorhq Is the popup even necessary? We could just silently clear the old sessions. This alert is just a technical FYI. 🤔

config/locales/client.en.yml
discoursereviewbot: [Gerhard Schlager posted](https://review.discourse.org/t/feature-limit-the-number-of-active-sessions-for-a-user-pr-8411/7290/17):

If it is really a constant then there's no need for an interpolation key. But if the `limit` can be changed somehow (site setting?), then it will be an issue for some languages. I doesn't matter how high the number is. Have fun reading http://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html :wink:

davidtaylorhq:
> We could just silently clear the old sessions. This alert is just a technical FYI.

I agree @CvX, there isn't really any 'action' required from the user. It is just us doing housekeeping. I regularly find myself logged out of Google for... unknown reasons, and it's fine.

@SamSaffron the bootbox was in the spec I've been working from, but what do you think about simplifying this PR to remove the notification?

SamSaffron:
> the bootbox was in the spec I've been working from, but what do you think about simplifying this PR to remove the notification?

Yeah I guess simplicity should win here, our behavior is sane here and 60 sessions is clearly too much, no need to notify on this quite yet, lets see if complaints mount.

PR #8414, opened 2019-11-26
FIX: oneboxer.js infinitely retrying failed requests

martin-brennan:
When we encountered a 404 or other issue when oneboxing a URL, we kept retrying it ad infinitum which can cause DDOS issues. The fix in oneboxer.js was simple:

* setFailedCache was used like a variable object, when it was in fact a function

At some point the API must have changed and not all places were updated.

### DIFF
--- a/app/assets/javascripts/pretty-text/oneboxer.js.es6
+++ b/app/assets/javascripts/pretty-text/oneboxer.js.es6
@@ -86,7 +86,7 @@ function loadNext(ajax) {
           removeLoading = false;
           loadingQueue.unshift({ url, refresh, $elem, categoryId, topicId });
         } else {
-          setFailedCache[normalize(url)] = true;
+          setFailedCache(normalize(url), true);
         }
       }
     )


SamSaffron:
hmmm I guess we should cache the failed state for say 5 minutes or so on the server to avoid us being used as a jump box that denies service  but feel free to merge this first.

martin-brennan:
@SamSaffron yeah I was thinking something like this in `OneboxController`, before we even attempt to generate a preview:

```
url_sha = Digest::SHA1.hexdigest(url)
if Rails.cache.read("onebox-url-blacklist-#{url_sha}").present?
  puts "You are a bad onebox URL!"
  return render body: "blacklisted", status: 404
end
```

Then further down on first instance of failed preview:

```
if preview.blank?
  Rails.cache.write("onebox-url-blacklist-#{url_sha}", url, expires_in: 1.hour)
  render body: nil, status: 404
else
  render plain: preview
end
```

SamSaffron:
yes something like that works, you can just use redis directly there which should be fine.

```
$redis.setex ... 
```

and

```
$redis.key
```

martin-brennan:
Any reason why some place we use `Rails.cache.write` and sometimes use `$redis.setex`? Just curious as I have never set directly in redis before.

SamSaffron:
I think we should ask this on dev :) we should make stuff consistent, that is for sure.

I say redis wins here cause if we are going for consistency it provides more features so it wins by default.

coding-horror:
Should this be backported?

PR #8415, opened 2019-11-26
 REFACTOR: poll-pie-chart widget

jjaffeux:
This commit mostly get rid of the later() call and encapsulate all pie chart display logic inside discourse-poll-pie-canvas widget instead of sharing it between discourse-poll-pie-chart and discourse-poll-pie-canvas


plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -542,6 +541,21 @@ function clearPieChart(id) {
...
   tagName: "canvas.poll-results-canvas",
 
+  init(attrs) {
+    loadScript("/javascripts/Chart.min.js").then(() => {
+      const data = attrs.poll.options.mapBy("votes");
+      const labels = attrs.poll.options.mapBy("html");
+      const config = pieChartConfig(data, labels);
+
+
+        const el = document.getElementById(
```
ZogStriP: Indentation is a bit off

PR #8420, opened 2019-11-27
FIX: oneboxer.js infinitely retrying failed requests (#8414)

martin-brennan:
* setFailedCache was used like a variable object, when it was in fact a function

### DIFF
--- a/app/assets/javascripts/pretty-text/oneboxer.js.es6
+++ b/app/assets/javascripts/pretty-text/oneboxer.js.es6
@@ -77,7 +77,7 @@ function loadNext(ajax) {
           removeLoading = false;
           loadingQueue.unshift({ url, refresh, $elem, categoryId, topicId });
         } else {
-          failedCache[normalize(url)] = true;
+          setFailedCache(normalize(url), true);
         }
       }
     )


app/assets/javascripts/pretty-text/oneboxer.js.es6
```diff
@@ -77,7 +77,7 @@ function loadNext(ajax) {
           removeLoading = false;
           loadingQueue.unshift({ url, refresh, $elem, categoryId, topicId });
         } else {
-          failedCache[normalize(url)] = true;
+          setFailedCache(normalize(url), true);
```
techAPJ: `setFailedCache` is not defined here.

techAPJ:
I think this will need more work since we are importing `setFailedCache` in discourse `master` branch, and I am not sure if we have that defined already in `stable` branch.

martin-brennan:
@techAPJ ah...I did not even think to look at whether `setFailedCache` was imported already. We don't even have `app/assets/javascripts/pretty-text/oneboxer.js.es6` in `stable` :(

martin-brennan:
Yeah running `git --no-pager diff origin/stable origin/master app/assets/javascripts/pretty-text/oneboxer.js.es6` there is a pretty huge difference between the two. What would we usually do in this case? Pull the whole file over from master, including creating the oneboxer-cache.js file?

SamSaffron:
My call here is lets just skip the backport, it is too complex.

martin-brennan:
@SamSaffron no problem, closing

PR #8421, opened 2019-11-27
FIX: Cache failed onebox URL request server-side

martin-brennan:
We already cache failed onebox URL requests client-side, we now want to cache this on the server-side for extra protection. failed onebox previews will be cached for 1 hour, and any more requests for that URL will fail with a 404 status.

### DIFF
--- a/app/controllers/onebox_controller.rb
+++ b/app/controllers/onebox_controller.rb
@@ -19,7 +19,7 @@ class OneboxController < ApplicationController
     invalidate = params[:refresh] == 'true'
     url = params[:url]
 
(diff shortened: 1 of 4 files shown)


SamSaffron:
This is great, a "rebake" should force another try, cause at least admins can then work around non resolution / bad internal net config and so on. 

This button I mean: 
image

martin-brennan:
@SamSaffron sure, I will just change this method:

```
def self.invalidate(url)
  Discourse.cache.delete(onebox_cache_key(url))
end
```

to also clear the failed URL cache, as this is what is called already from the rebake process

PR #8423, opened 2019-11-27
FIX: Do not start the login flow when logging out from SSO/Authenticator

davidtaylorhq:
This affects login_required sites which use SSO or have only one authenticator enabled. Previously, logging out would redirect to the homepage, which would then redirect to the identity provider. Now, users will be redirected to the Discourse login page. This avoids the confusing situation where a user appears to remain logged in after clicking logout.

Sites which have explicitly defined a logout_redirect url are not affected by this change.

For context, see https://meta.discourse.org/t/134138/2

### DIFF
--- a/app/assets/javascripts/discourse/lib/logout.js.es6
+++ b/app/assets/javascripts/discourse/lib/logout.js.es6
@@ -1,4 +1,6 @@
 import { isEmpty } from "@ember/utils";
+import { findAll } from "discourse/models/login-method";
+
 export default function logout(siteSettings, keyValueStore) {
   if (!siteSettings || !keyValueStore) {
     const container = Discourse.__container__;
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/lib/logout.js.es6
```diff
@@ -9,9 +11,21 @@ export default function logout(siteSettings, keyValueStore) {
...
   const redirect = siteSettings.logout_redirect;
-  if (isEmpty(redirect)) {
-    window.location = Discourse.getURL("/");
-  } else {
+  if (!isEmpty(redirect)) {
     window.location.href = redirect;
+    return;
   }
+
+  const isSso = siteSettings.enable_sso;
```
ZogStriP: Minor but I think `isSSO` is easier to read. To many s otherwise ;)

app/assets/javascripts/discourse/lib/logout.js.es6
```diff
@@ -9,9 +11,21 @@ export default function logout(siteSettings, keyValueStore) {
...
+  const isOneAuthenticator =
```
ZogStriP: What about `isUniqueAuthenticator` or `onlyOneAuthenticator`?

davidtaylorhq:
Thanks @ZogStriP - I renamed the variables to simply `sso` and `oneAuthenticator`. I think it's easier to read now

PR #8427, opened 2019-11-28
FIX: Optimize quoted images

udan11:
Only images that were part of a lightbox used to be optimized. This
patch ensures that quoted images are also optimized.

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -208,9 +208,7 @@ class CookedPostProcessor
     # minus emojis
     @doc.css("img.emoji") -
     # minus oneboxed images
-    oneboxed_images -
-    # minus images inside quotes
-    @doc.css(".quote img")
+    oneboxed_images
   end
 
   def extract_images_for_post
@@ -348,7 +346,8 @@ class CookedPostProcessor
       end
     end
 
-    add_lightbox!(img, original_width, original_height, upload, cropped: crop)
+    add_lightbox!(img, original_width, original_height, upload, cropped: crop) if img.ancestors('.quote').blank?
+    optimize_image!(img, upload, cropped: crop) if upload
   end
 
   def loading_image(upload)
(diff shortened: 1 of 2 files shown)


lib/cooked_post_processor.rb
```diff
@@ -373,6 +372,38 @@ def each_responsive_ratio
...
+      img["src"] = thumbnail.url
+
+      srcset = +""
+
+      each_responsive_ratio do |ratio|
+        resized_w = (w * ratio).to_i
+        resized_h = (h * ratio).to_i
+
+        if !cropped && upload.width && resized_w > upload.width
+          cooked_url = UrlHelper.cook_url(upload.url, secure: upload.secure?)
```
jjaffeux: feels like the two repeated lines here, could be extracted in a method

lib/cooked_post_processor.rb
```diff
@@ -373,6 +372,38 @@ def each_responsive_ratio
...
+          srcset << ", #{cooked_url} #{ratio.to_s.sub(/\.0$/, "")}x"
+        elsif t = upload.thumbnail(resized_w, resized_h)
+          cooked_url = UrlHelper.cook_url(t.url, secure: upload.secure?)
+          srcset << ", #{cooked_url} #{ratio.to_s.sub(/\.0$/, "")}x"
+        end
+
+        img["srcset"] = "#{UrlHelper.cook_url(img["src"], secure: upload.secure?)}#{srcset}" if srcset.present?
```
jjaffeux: think if here would deserve it's own line, this quite a long line of code and it's easy to not see it

lib/cooked_post_processor.rb
ZogStriP: Not sure we need a whole method for that. Just re-order the logic a tiny bit?

```ruby

upload_url = if !cropped && upload.width && resized_w > upload.width
  upload.url
elseif t = upload.thumbnail(resized_w, resized_h)
  t.url
end

if upload_url.present?
  cooked_url = UrlHelper.cook_url(upload_url, secure: upload.secure?)
  srcset << ", #{cooked_url} #{ratio.to_s.sub(/\.0$/, "")}x"
end

PR #8431, opened 2019-11-28
DEV: s/\$redis/Discourse\.redis

jjaffeux:


eviltrout:
Another thought - can we deprecate `$redis` with an error?

jjaffeux:
> Another thought - can we deprecate `$redis` with an error?

Yew I would want to do this but want to do a check of plugins first

eviltrout:
Sorry I should have said warning! Output something when it's used.

jjaffeux:
@eviltrout I added a rubocop rule which should do this

SamSaffron:
feel free to merge when you are comfortable.

PR #8434, opened 2019-11-28
Support for custom messages and redirects when creating posts

eviltrout:
When a post/topic is created Discourse serializes a `NewPostResult`
object. Normally this contains a status like `created_post` or
errors describing why the post could not be created.

There are times when a plugin might want to take the inputted post
and do something in the background. In this case, the plugin
can return a custom `message` and `route_to` attribute in the
`NewPostResult`.

If present, the message will be displayed in an alert, and when "Ok" is
clicked the user will be routed to the new URL.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js.es6
+++ b/app/assets/javascripts/discourse/controllers/composer.js.es6
@@ -711,14 +711,13 @@ export default Controller.extend({
         }
 
         if (result.responseJson.route_to) {
-          return this.destroyDraft().then(() => {
-            if (result.responseJson.message) {
-              return bootbox.alert(result.responseJson.message, () => {
-                DiscourseURL.routeTo(result.responseJson.route_to);
-              });
-            }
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -709,6 +709,18 @@ export default Controller.extend({
         if (result.responseJson.action === "create_post") {
           this.appEvents.trigger("post:highlight", result.payload.post_number);
         }
+
+        if (result.responseJson.route_to) {
+          return this.destroyDraft().then(() => {
```
ZogStriP: I wonder if we should instead show the message first before destroying the draft?

eviltrout: The message isn't a confirmation, but what I can do is delete the draft in parallel. I've created a follow up commit, thanks!

PR #8435, opened 2019-11-28
Tag synonyms

nlalonde:
Allow tags to have synonyms. The UI for managing a tag's synonyms is at `/tags/:tag_name`. Click the "Info" button in the top right to reveal it. I moved other tag Admin functions to this new section. In the future we can add things like a tag description there too.

https://meta.discourse.org/t/merging-tags-tag-synonyms/81869/12


hnb-ku:
I think it looks very good. kudos for all the tooltips :+1: 

If I can nitpick a little bit, I'd say remove the "delete" tag icon from the synonym . I think it's fine for the admin to have to head to the tag page and delete it from there. 

I'm referring to this

image

Other than that, I think it's ready. I know you want to add more features and controls in that section. So I say go ahead and do it and we can tidy everything up - if needed - after that. 

Very nice work!

nlalonde:
Thanks Joe!

Not sure about removing the delete from here though. Deleting a synonym would become a multi-step process (unlink, search for tag, go to tag page, delete). Synonyms have no objects associated with them, so making it harder to delete them doesn't protect from anything bad happening. A confirmation would be helpful though.

app/assets/javascripts/select-kit/components/single-select.js.es6
```diff
@@ -285,7 +285,11 @@ export default SelectKitComponent.extend({
           this
         );
 
-        this._boundaryActionHandler("onSelect", computedContentItem.value);
+        this._boundaryActionHandler(
```
jjaffeux: This is changed in select-kit 2, I agree this was missing :+1:

PR #8440, opened 2019-11-30
Extend core for discourse-encrypt uploads

udan11:


app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -867,6 +870,19 @@ class PluginApi {
...
+  /**
+   * Registers a function to generate Markdown after a file has been uploaded.
+   *
+   * Example:
+   *
+   * api.addComposerUploadMarkdownGetter(upload => {
+   *   return `_uploaded ${upload.original_filename}_`;
+   * })
+   */
+  addComposerUploadMarkdownGetter(getter) {
```
SamSaffron: Maybe `addComposerUploadMarkdownResolver` ?

Also behavior when chained is a bit tricky, do we even want to allow chaining?

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
...
-          match[4] !== "x"
-        )
-          token.attrs.push(["class", "resizable"]);
+  const token = tokens[idx];
+  const alt = slf.renderInlineAsText(token.children, options, env);
+
+  const split = alt.split("|");
+  const alt_split = [];
+
+  for (let i = 0, match, data; i < split.length; ++i) {
```
jjaffeux: I think it could be interesting to have variables for each used match. eg: `const foo = match[5]`

This would make the whole code block much more readable.

app/assets/javascripts/pretty-text/upload-short-url.js.es6
```diff
@@ -53,8 +53,11 @@ function _loadCachedShortUrls($uploads) {
 
           if (url !== MISSING) {
             $upload.attr("href", url);
-            const content = $upload.text().split("|");
 
+            // Replace "|attachment" with class='attachment'
+            // TODO: This is a part of the cooking process now and should be
+            // removed in the future.
+            const content = $upload.text().split("|");
```
jjaffeux: if there's any chance `$upload.text()` can be `""` we should do:

```suggestion
            const content = $upload.text().split("|").filter(Boolean);
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -190,16 +199,24 @@ function setupImageDimensions(md) {
...
-    linkOpenToken.attrs.unshift(["class", split[1]]);
-    linkTextToken.content = split[0];
+  const linkToken = tokens[idx];
+  const textToken = tokens[idx + 1];
+
+  const split = textToken.content.split("|");
+  const content_split = [];
+
+  for (let i = 0, data; i < split.length; ++i) {
+    if (split[i] === ATTACHMENT_CSS_CLASS) {
```
jjaffeux: probably worth to have split[i] as a variable here as it's reused multiple times.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
...
-        if (
-          options.discourse.previewing &&
-          match[6] !== "x" &&
-          match[4] !== "x"
-        )
-          token.attrs.push(["class", "resizable"]);
+  const token = tokens[idx];
+  const alt = slf.renderInlineAsText(token.children, options, env);
+
+  const split = alt.split("|");
```
jjaffeux: if there's any chance `alt` can be "" we should do:

```suggestion
  const split = alt.split("|").filter(Boolean);
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
   md.renderer.rules.html_raw = renderHoisted;
 }
 
+function extractDataAttribute(str) {
```
ZogStriP: This method should be tested

app/assets/javascripts/discourse/lib/plugin-api.js.es6
udan11: It will use the last "truthy" value returned by a resolver. This way it allows multiple plugins to work correctly as long as an upload is not of interest to multiple plugins (or they could still alter attributes of the `upload` object such as filename).

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
udan11: The problem is that some parts may have multiple meanings. An "accurate" name for `match[5]` would be `widthOrHeightRatioOrPercentage`, which is useless. Later, variables are defined to hold the value (for example for `match[5]` it is `percent`, `wr` or `hr`).

app/assets/javascripts/pretty-text/upload-short-url.js.es6
udan11: `[double pipe ||]()` should generate a link with title 'double pipe ||'. If blank values would be filtered, then the output would be 'double pipe '.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
udan11: Similar to the other place, it does not hold a specific value. It would have a generic name such as `part` or `fragment` which does not really add any value.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
udan11: Alt texts can also contain '|'.

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
udan11: I agree. Added a test. :)

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -190,16 +199,24 @@ function setupImageDimensions(md) {
...
-  ];
-
-  if (isValid && split.length === 2 && split[1] === ATTACHMENT_CSS_CLASS) {
-    linkOpenToken.attrs.unshift(["class", split[1]]);
-    linkTextToken.content = split[0];
+  const linkToken = tokens[idx];
+  const textToken = tokens[idx + 1];
+
+  const split = textToken.content.split("|");
+  const content_split = [];
```
jjaffeux:
```suggestion
  const contentSplit = [];
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -190,16 +199,24 @@ function setupImageDimensions(md) {
...
+
+  for (let i = 0, data; i < split.length; ++i) {
+    if (split[i] === ATTACHMENT_CSS_CLASS) {
+      linkToken.attrs.unshift(["class", split[i]]);
+    } else if ((data = extractDataAttribute(split[i]))) {
+      linkToken.attrs.push(data);
+    } else {
+      content_split.push(split[i]);
```
jjaffeux:
```suggestion
      contentSplit.push(split[i]);
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -190,16 +199,24 @@ function setupImageDimensions(md) {
...
+    }
+  }
+
+  if (content_split.length > 0) {
```
jjaffeux:
```suggestion
  if (contentSplit.length > 0) {
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -190,16 +199,24 @@ function setupImageDimensions(md) {
...
+    textToken.content = content_split.join("|");
```
jjaffeux:
```suggestion
    textToken.content = contentSplit.join("|");
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
...
-          options.discourse.previewing &&
-          match[6] !== "x" &&
-          match[4] !== "x"
-        )
-          token.attrs.push(["class", "resizable"]);
+  const token = tokens[idx];
+  const alt = slf.renderInlineAsText(token.children, options, env);
+
+  const split = alt.split("|");
+  const alt_split = [];
```
jjaffeux:
```suggestion
  const altSplit = [];
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
...
+        token.attrs.push(["class", "resizable"]);
+    } else if ((data = extractDataAttribute(split[i]))) {
+      token.attrs.push(data);
+    } else {
+      alt_split.push(split[i]);
     }
   }
 
-  token.attrs[token.attrIndex("alt")][1] = alt;
+  token.attrs[token.attrIndex("alt")][1] = alt_split.join("|");
```
jjaffeux:
```suggestion
  token.attrs[token.attrIndex("alt")][1] = altSplit.join("|");
```

app/assets/javascripts/pretty-text/engines/discourse-markdown-it.js.es6
```diff
@@ -124,64 +124,73 @@ function setupHoister(md) {
...
+      if (token.attrIndex("height") === -1) {
+        token.attrs.push(["height", height]);
       }
+
+      if (options.discourse.previewing && match[6] !== "x" && match[4] !== "x")
+        token.attrs.push(["class", "resizable"]);
+    } else if ((data = extractDataAttribute(split[i]))) {
+      token.attrs.push(data);
+    } else {
+      alt_split.push(split[i]);
```
jjaffeux:
```suggestion
      altSplit.push(split[i]);
```

SamSaffron:
Overall I am fine with this and would love to see it merged ASAP. 

One thing we should think about is properly documenting our plugin API in tests that way we have some examples of how they are consumed and what they do. It is not trivial to add this but worth it.

PR #8443, opened 2019-12-02
FEATURE: Plugin outlet adjacent to new topic btn

markvanlan:


app/assets/javascripts/discourse/templates/components/d-navigation.hbs
```diff
@@ -12,6 +12,13 @@
   {{category-notifications-button value=category.notification_level category=category}}
 {{/if}}
 
+{{plugin-outlet name="alongside-create-topic-button"
```
eviltrout: I would suggest calling it `before-create-topic-button` but otherwise it's all good.

markvanlan: Agreed, thanks :)

eviltrout:
You can merge this whenever you like!

PR #8446, opened 2019-12-03
FIX: 500 error for missing badge grouping

oblakeerickson:
If a badge grouping happens to have been deleted a 500 error will be
thrown when looking a user's badges.

This fix allows the badge page to still be shown without any errors. The
badge with the missing badge grouping is still displayed.

I'll follow up with a separate pr/commit that will ensure if a badge
grouping is deleted, all badges with that badge_grouping_id will also be
updated.


SamSaffron:
Seems safe to me, yeah a follow up to restrict deletion of groups when badges still belong to the group totally makes sense.

PR #8447, opened 2019-12-03
FEATURE: ability to add all active components to theme

lis2:
Improvement that now you can add active components when editing the theme, and add active themes when editing components. 

When click “add all” only active themes are added. However, the inactive theme can still be added manually (for example if I am preparing a theme before I want to publish it). Also if you click “add all” it will not remove your manual choice.
In addition, when all active themes are on the list, then “add all” button disappears.

Finally, I fixed incorrect indentation of the template

Demo behaviour is presented in gifs below:
component
theme

### DIFF
--- a/app/assets/javascripts/admin/mixins/setting-component.js.es6
+++ b/app/assets/javascripts/admin/mixins/setting-component.js.es6
@@ -102,12 +102,16 @@ export default Mixin.create({
 
   @discourseComputed("buffered.value")
   bufferedValues(bufferedValuesString) {
-    return bufferedValuesString && bufferedValuesString.split("|");
+    return (
+      bufferedValuesString && bufferedValuesString.split("|").filter(Boolean)
+    );
   },
 
   @discourseComputed("setting.defaultValues")
   defaultValues(defaultValuesString) {
-    return defaultValuesString && defaultValuesString.split("|");
+    return (
+      defaultValuesString && defaultValuesString.split("|").filter(Boolean)
+    );
   },
 
   @discourseComputed("defaultValues", "bufferedValues")
--- a/app/assets/javascripts/admin/mixins/setting-component.js.es6
+++ b/app/assets/javascripts/admin/mixins/setting-component.js.es6
@@ -221,7 +221,15 @@ export default Mixin.create({
     },
 
     setDefaultValues() {
-      this.set("buffered.value", this.get("setting.defaultValues"));
+      const buffered_values = this.get("buffered.value").split("|");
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/admin/mixins/setting-component.js.es6
```diff
@@ -100,6 +100,24 @@ export default Mixin.create({
     return settingDefault !== bufferedValue;
   },
 
+  @discourseComputed("buffered.value")
+  bufferedValues(bufferedValuesString) {
+    return bufferedValuesString && bufferedValuesString.split("|");
```
jjaffeux: can `bufferedValuesString`  be a `""` ? If yes we should filter `.filter(Boolean)`  as `"".split("|") // [""]`

app/assets/javascripts/admin/mixins/setting-component.js.es6
```diff
@@ -100,6 +100,24 @@ export default Mixin.create({
...
+  },
+
+  @discourseComputed("setting.defaultValues")
+  defaultValues(defaultValuesString) {
+    return defaultValuesString && defaultValuesString.split("|");
```
jjaffeux: can `defaultValuesString`  be a `""` ? If yes we should filter `.filter(Boolean)`  as `"".split("|") // [""]`

app/assets/javascripts/admin/templates/customize-themes-show.hbs
```diff
@@ -125,24 +125,29 @@
...
 
-         {{#if colorSchemeChanged}}
-            {{d-button action=(action "changeScheme") class="btn-primary submit-edit" icon="check"}}
-            {{d-button action=(action "cancelChangeScheme") class="btn-default cancel-edit" icon="times"}}
-         {{/if}}
+          <div class="desc">{{i18n "admin.customize.theme.color_scheme_select"}}</div>
+        </div>
+        <div class="setting-controls">
+          {{#if colorSchemeChanged}}
+            {{d-button action=(action "changeScheme") class="btn ok submit-edit" icon="check"}}
```
jjaffeux: btn class is not needed here

app/assets/javascripts/admin/templates/customize-themes-show.hbs
```diff
@@ -125,24 +125,29 @@
...
+            {{d-button action=(action "cancelChangeScheme") class="btn cancel cancel-edit" icon="times"}}
```
jjaffeux: btn class is not needed here

PR #8448, opened 2019-12-03
Add timezone to current_user_serializer

martin-brennan:
* the code to get current user already got their
  user_option record as well, so adding the timezone
  column to the attributes returned should not create
  any additional overhead. this timezone will be very
  useful for getting momentjs dates relative to the
  user's timezone


SamSaffron:
This sounds good to me.

PR #8450, opened 2019-12-03
FEATURE: Hide discourse version in page header

rrooding:
Added a new setting to the security tab of Discourse to hide the Discourse version in the 'generator' meta tag.

The default generator tag contains something like: `content="Discourse 2.3.0 - https://github.com/discourse/discourse version 0bcb62fc2d027ee031d2275112da7ac9e26d31bf"`

When you change this setting to true, it will just show `content="Discourse"`

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -285,6 +285,15 @@ module ApplicationHelper
     content_tag(:script, MultiJson.dump(json).html_safe, type: 'application/ld+json'.freeze)
   end
 
+  def discourse_generator_tag
+    if SiteSetting.hide_discourse_version
+      version_content = "Discourse"
+    else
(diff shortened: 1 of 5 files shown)


rrooding:
^ Not sure about the 2 rubocop offences. They are not introduced by this PR?

eviltrout:
The rubocop thing can be ignored, sorry about that.

This PR looks good, but my question is why would you want to hide the Discourse version? Was this discussed anywhere like our meta forums in advance?

rrooding:
A customer of us is using Discourse for a platform that is focussed on security for large companies, and they asked us to remove the version information to disclose as little as possible. Thought it might interest more people :)

SamSaffron:
I sadly can not support a site setting here, if you really need this I recommend a plugin. 

Then inject middleware that strips this out. I know its not the cleanest thing in the world to do but you will not need a fork of the software. 

I feel having this option actually hurts the ecosystem and makes it harder to troubleshoot stuff.

PR #8451, opened 2019-12-03
FIX: Secure Upload URLs in lightbox

CvX:
This fixes the following issues:

* The link element on the lightbox which pops open the lightbox was linking to the S3 URL with a private ACL instead of the secure media URL for the image
* Change to use `@post.with_secure_media?` in `CookedPostProcessor` for URL cooking, as in some cases, like when a post is edited and an upload is added, `upload.secure?` can be false which resulted in `srcset` URLs not being cooked correctly to secure media upload urls.

**NOTE: these notes were modified by @martin-brennan**

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -375,9 +375,10 @@ class CookedPostProcessor
   def optimize_image!(img, upload, cropped: false)
     w, h = img["width"].to_i, img["height"].to_i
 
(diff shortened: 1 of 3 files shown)


SamSaffron:
@martin-brennan  assume you are taking this over?

martin-brennan:
@SamSaffron yes I am, I'm just trying to figure out how these fixes relate to the other issue on that Australia meetup image where none of the `srcset` URLs were secure, and roll it a fix for that into this.

PR #8455, opened 2019-12-03
DEV: Provide API for anonymous cache segments

eviltrout:
This can be used from a plugin that needs to establish something new in
the anonymous cache. For example `is_ie` for an internet explorer
plugin.

### DIFF
--- a/lib/middleware/anonymous_cache.rb
+++ b/lib/middleware/anonymous_cache.rb
@@ -7,10 +7,40 @@ require_dependency "guardian"
 module Middleware
   class AnonymousCache
 
+    def self.cache_key_segments
+      @@cache_key_segments ||= {
+        m: 'key_is_mobile?',
+        c: 'key_is_crawler?',
+        b: 'key_has_brotli?',
+        t: 'key_cache_theme_ids',
+        ca: 'key_compress_anon'
+      }
+    end
+
+    # Compile a string builder method that will be called to create
+    # an anonymous cache key
+    def self.compile_key_builder
+      method = +"def self.__compiled_key_builder(h)\n  \""
+      cache_key_segments.each do |k, v|
+        raise "Invalid key name" unless k =~ /^[a-z]+$/
+        raise "Invalid method name" unless v =~ /^key_[a-z_\?]+$/
+        method << "|#{k}=#\{h.#{v}}"
+      end
+      method << "\"\nend"
+      eval(method)
+      @@compiled = true
+    end
+
+    def self.build_cache_key(helper)
+      compile_key_builder unless defined?(@@compiled)
+      __compiled_key_builder(helper)
+    end
+
     def self.anon_cache(env, duration)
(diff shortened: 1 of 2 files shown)


lib/middleware/anonymous_cache.rb
```diff
@@ -11,6 +11,17 @@ def self.anon_cache(env, duration)
...
     end
 
+    # This gives us an API to insert anonymous cache segments
+    def self.cache_key_segments
+      @@cache_key_segments ||= {
+        m: ->(h) { h.is_mobile? },
+        c: ->(h) { h.is_crawler? },
+        b: ->(h) { h.has_brotli? },
+        t: ->(h) { h.theme_ids.join(',') },
+        ca: ->(_) { GlobalSetting.compress_anon_cache.to_s }
```
eviltrout: Note: previously the `compress_anon_cache` was not separated by a variable and simply appended `true` or `false` to the key, which I assume was a mistake and not deliberate.

SamSaffron: GlobalSetting.compress_anon_cache.to_s is never going to change at runtime, maybe just hardcode it?

One note here is that the new flexible implementation is almost 3x slower than the original per: https://gist.github.com/00a759f5d8bc45726773afb1dbde379a 

I am normally not such a stickler, but this is a code path that every single request goes through so I am a bit sensitive.

Maybe we keep the old method for the base and then allow people to add stuff to "additional_cache_key_segments" if they wish to. As an advantage here they will not be able to break base caching support in a plugin cause they will only be allowed to add stuff on.

davidtaylorhq:
How is this used by plugins? Will they override the `cache_key_segments` function, and append something?

I wonder if it would be better to use the same pattern as other plugin APIs, and provide an easy method in `instance.rb`, which automatically respects whether the plugin is enabled.

eviltrout:
@SamSaffron while I think it's questionable if we'll ever notice the performance difference, I'm never one to back away from a challenge and have added a small method compiler. In my tests it has equivalent or better performance to your string one:

https://gist.github.com/eviltrout/b41224d28a406579fa28ad0ef6b180e0


@davidtaylorhq the latest version has a plugin_api method too, good suggestion!

lib/plugin/instance.rb
```diff
@@ -82,6 +82,12 @@ def initialize(metadata = nil, path = nil)
     @idx = 0
   end
 
+  def register_anonymous_cache_key(key, &block)
+    key_method = "key_#{key}"
+    add_to_class(Middleware::AnonymousCache, key_method, &block)
+    Middleware::AnonymousCache.cache_key_segments[key] = key_method
```
SamSaffron: overall the change looks good to me but I would re-run `compile_key_builder` here to ensure that if for any reason it was ran earlier it will be re-cached.

eviltrout: Sure, good idea!

PR #8457, opened 2019-12-04
FIX: inbound link when the only slug available

lis2:
Problem mentioned in [meta](https://meta.discourse.org/t/inbound-links-dont-show-up-when-topic-id-is-not-included/100551)

When there is an internal link without ID, only slug, we should still try to create reflection link.

### DIFF
--- a/app/models/topic_link.rb
+++ b/app/models/topic_link.rb
@@ -172,6 +172,7 @@ class TopicLink < ActiveRecord::Base
     internal = false
     topic_id = nil
     post_number = nil
+    topic = nil
 
     if upload = Upload.get_from_url(url)
       internal = Discourse.store.internal?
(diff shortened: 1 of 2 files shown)


app/models/topic_link.rb
```diff
@@ -185,9 +186,10 @@ def self.ensure_entry_for(post, link, parsed)
 
       topic_id = route[:topic_id].to_i
       post_number = route[:post_number] || 1
+      topic_slug = route[:id]
 
       # Store the canonical URL
-      topic = Topic.find_by(id: topic_id)
+      topic = Topic.find_by("id = ? OR slug = ?", topic_id, topic_slug)
```
SamSaffron: oh the index is on `topic_slug is not null` so this can trigger a table scan... instead double query is probably correct here

```
topic = Topic.find_by(id: topic_id)
if !topica && topic_slug
    topic = Topic.find_by(slug: topic_slug)
end
```

SamSaffron:
Yeah looks good to me... merge at will.

PR #8459, opened 2019-12-04
FIX: Do not autocomplete categories or emojis in code blocks

udan11:


app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -410,5 +410,46 @@ export function rescueThemeError(name, error, api) {
   document.body.prepend(alertDiv);
 }
 
+const CODE_BLOCKS_RULES = [
+  { rule: "^(    |\\t).*", end: "" },
+  { rule: "`.+?`", end: "`" },
```
ZogStriP: I wonder if this one is any faster?

```suggestion
  { rule: "`[^`]+`", end: "`" },
```

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -410,5 +410,46 @@ export function rescueThemeError(name, error, api) {
...
+  { rule: "^```[^]*?^```", end: "\n```" },
+  { rule: "\\[code\\][^]*?\\[\\/code\\]", end: "\n[/code]" }
+];
+
+export function inCodeBlock(text, pos) {
+  const regexp = new RegExp(
+    CODE_BLOCKS_RULES.map(rules => rules.rule).join("|"),
+    "gm"
+  );
```
SamSaffron: minor... but why map this every time you run the function, just do it once upfront ?

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -410,5 +410,46 @@ export function rescueThemeError(name, error, api) {
...
+  let match;
+  while ((match = regexp.exec(text)) != null) {
+    const begin = match.index;
+    const end = match.index + match[0].length;
+    if (begin <= pos && pos <= end) {
+      return true;
+    }
+  }
+
+  // Try to end block and see if other code blocks are found
```
SamSaffron: hmmm do we want to do this? I get that having auto complete after "test /` @" can be arguably allowed but honestly I am fine to just ban it in this case.

test/javascripts/lib/utilities-test.js.es6
```diff
@@ -186,3 +187,41 @@ QUnit.test("fillMissingDates", assert => {
     "it returns a JSON array with 31 dates"
   );
 });
+
+QUnit.test("inCodeBlock", assert => {
```
SamSaffron: super nice to have this test and nice specification. I wonder though if we should make it a bit more readable?

`test V'X'`... use V to denote that auto complete will open and X to denote it will not, then you zoom through the string and create assertions on the fly. Way easier to reason about the implementation.

app/assets/javascripts/discourse/lib/utilities.js.es6
udan11: They are both super-fast, but your version looks more readable.

app/assets/javascripts/discourse/lib/utilities.js.es6
udan11: You are right. There is also no point in keeping the array of rules anymore. :)

app/assets/javascripts/discourse/lib/utilities.js.es6
udan11: The user can still end the code block before proceed with its contents, so I will just get rid of this.

test/javascripts/lib/utilities-test.js.es6
udan11: The test was built kind of like that. `foo` was always outside of a code block and `bar` was always inside a code block. I replaced it with "0" (not in code block) and "1" (in code block) and am generating the assertions on-the-fly now.

I am not a big fan of generating assertions because then I feel like the test contains too much logic. I do agree though that it is a lot more readable.

test/javascripts/lib/utilities-test.js.es6
```diff
@@ -186,3 +187,25 @@ QUnit.test("fillMissingDates", assert => {
...
+  const phrase = "Lorem ipsum dolor sit amet, consectetur adipiscing elit.";
+  const text = `${phrase}\n\n\`\`\`\n${phrase}\n\`\`\`\n\n${phrase}\n\n\`${phrase}\n${phrase}\n\n${phrase}\n\n[code]\n${phrase}\n[/code]\n\n${phrase}\n\n    ${phrase}\n\n\`${phrase}\`\n\n${phrase}`;
+
+  const start = performance.now();
+  inCodeBlock(text, text.length);
+  const end = performance.now();
+
+  // This runs in 'keyUp' event handler so it should run as fast as
+  // possible. It should take less than 1ms for the test text.
+  assert.ok(end - start < 10);
```
SamSaffron: From my perspective this PR is good to merge, but I am a bit concerned about an assertion for performance, a single GC can through it off so this is ultra likely to trip. I think we should skip this test for now despite the awesome intent.

Maybe we should start thinking about a "performance" front end test suite longer term that handles this and more.

At the integration level this becomes mega interesting, maybe "typing in the composer does not lead to lag" type of integration smoke test. Probably worth discussing with @ZogStriP  and @eviltrout

udan11: I tried to take into account the system noise by ensuring that the limit is 10 times of what I am seeing on my laptop. Something else I could do to consolidate this test is to run the code 10 times and take the minimum execution time. I doubt the GC can impact more than a few of these runs.

Measuring the minimum time makes sense because ultimately the execution time is a sum of the "real" execution time and noise. By taking the minimum, we are also taking the minimum noise, because in theory the execution time should stay constant. In other words, `time = real + noise`, `min(time) = min(real + noise)`, but `min(real) ~= real` so `min(time) = real + min(noise)`.

Anyway, for now, I will remove this test. I agree that it would be helpful to implement a framework for this kind of test and also to write tests for vital components such as the composer or topic view. Chrome kind of does this by `console.warn`ing the user if an event handler takes too much to run.

eviltrout: I definitely think this should be a different class of test like smoke, because I guarantee we will see it fail in places we never expected it to otherwise.

As a future project we could add a "speed regression" suite that lets us know if things get slower.

PR #8461, opened 2019-12-04
FEATURE: Featured topic for user profile & card

markvanlan:


markvanlan:
@awesomerobot Can see what you think of the changes to [this html](https://github.com/discourse/discourse/pull/8461/files#diff-fd7105b5b3bc704b4a8f440904c5f106) and [this css](https://github.com/discourse/discourse/pull/8461/files#diff-0547d872eb008dc55b763ec8a4309e18)?

The user card right now has all its rows numbered (from first-sixth), but all the rows following the first and second are styled no differently. I needed to add another row, and this structure made it so annoying to arbitrarily increase the class names from "third" to "fourth" and so on. 

This change does not alter the user card appearance, until the user "features a topic" on their profile

markvanlan:
@awesomerobot Yeah, all those rows have classes like `.location-and-website`, and I would be shocked if themes would access those elements by row rather than that class.

markvanlan:
@eviltrout 

If you created the topic you are viewing, this `Feature On Profile` button will now appear.

Screenshot from 2019-12-04 14-50-24@2x

Now the `Featured Topic` link appears in the user card.

Screenshot from 2019-12-04 14-50-43@2x

As well as the user's profile (this is the view logged in as someone else. This view looks different if you are viewing yourself, and does not show the link.)

Screenshot from 2019-12-04 14-52-10@2x

And finally, users can see their featured topic under `preferences -> profile`

[Screenshot from 2019-12-04 14-51-23@2x](https://user-images.githubusercontent.com/16214023/70181400-71a1cc80-16a7-11ea-896e-c4e2eef18c67.png)

spec/serializers/web_hook_user_serializer_spec.rb
```diff
@@ -23,18 +23,13 @@
...
-      message = ""
-
-      if difference < 0
-        message << "#{difference * -1} key(s) have been removed from this serializer."
-      else
-        message << "#{difference} key(s) have been added to this serializer."
-      end
-
-      message << "\nPlease verify if those key(s) are required as part of the web hook's payload."
+      message = (difference < 0 ?
```
markvanlan: When this error has failed in the past, the error is because the test itself broke; it did not say the error message it was supposed to. It was the cannot modify frozen string error.

eviltrout: I believe this is unrelated and can be safely ignored.

app/assets/javascripts/discourse/components/topic-footer-buttons.js.es6
```diff
@@ -58,5 +58,15 @@ export default Component.extend({
...
-    archived ? "topic.move_to_inbox.title" : "topic.archive_message.title"
+    archived ? "topic.move_to_inbox.title" : "topic.archive_message.title",
+
+  @discourseComputed("topic.id", "currentUser.featured_topic")
+  topicFeaturedOnProfile(topicId, featuredTopic) {
+    return featuredTopic && featuredTopic.id === topicId;
+  },
+
+  @discourseComputed("topic.user_id", "topic.isPrivateMessage")
+  showToggleFeatureOnProfileButton(user_id, isPm) {
```
eviltrout: We should use `userId` here because it's javascript.

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -679,6 +679,24 @@ export default Controller.extend(bufferedProperty("model"), {
       }
     },
 
+    toggleFeaturedOnProfile() {
+      if (!this.currentUser) return;
+      const toggle = () => {
```
eviltrout: I don't think this needs to be a closure since it's not wrapping any variables in this scope. We'd use less memory by making it a method on the controller like `_performToggleFeatured`

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -444,6 +444,22 @@ const Topic = RestModel.extend({
...
+      type: "PUT",
+      data: { topic_id: this.id }
+    })
+      .then(() => {
+        const newFeaturedTopic = removing ? null : this;
+        user.set("featured_topic", newFeaturedTopic);
+        return;
+      })
+      .catch(popupAjaxError)
+      .finally();
```
eviltrout: This finally doesn't seem to be doing anything? Could be removed.

app/assets/javascripts/discourse/components/topic-footer-buttons.js.es6
eviltrout: Also we should make sure this topic isn't in a restricted group.

app/controllers/users_controller.rb
```diff
@@ -1399,6 +1400,24 @@ def revoke_auth_token
     render json: success_json
   end
 
+  def feature_topic
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
+    topic = Topic.find(params[:topic_id].to_i)
+
```
eviltrout: We're going to need more checks on the topic here. I suggest a `can_feature_topic` guardian method because we'll want to make sure:

1. The topic is 100% public (not a PM, not in a secure group)
2. The user created the topic

Otherwise I could hit this API endpoint and feature a topic that is private!

app/controllers/users_controller.rb
```diff
@@ -1399,6 +1400,24 @@ def revoke_auth_token
...
+    raise Discourse::NotFound unless topic && topic.user_id == user.id
+    user.user_profile.update(featured_topic_id: topic.id)
+
+    render body: nil
```
eviltrout: We prefer to return `success_json` from methods rather than an empty body.

app/models/user_profile.rb
```diff
@@ -145,6 +146,9 @@ def website_domain_validator
     self.errors.add :base, (I18n.t('user.website.domain_not_allowed', domains: allowed_domains.split('|').join(", "))) unless allowed_domains.split('|').include?(domain)
   end
 
+  def self.remove_featured_topic_from_all_profiles(topic)
+    self.where(featured_topic_id: topic.id).update_all(featured_topic_id: nil)
```
eviltrout:
```suggestion
    where(featured_topic_id: topic.id).update_all(featured_topic_id: nil)
```

db/migrate/20191202202212_add_featured_topic_id_to_user_profiles.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddFeaturedTopicIdToUserProfiles < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :user_profiles, :featured_topic, foreign_key: { to_table: 'topics' }
```
eviltrout: We've had some debate internally about using foreign keys. While we might use them in the future for now it's part of a bigger discussion we haven't finished.

It would be better if you used a regular NULLable int here

spec/serializers/web_hook_user_serializer_spec.rb
markvanlan: I added to the user serializer, so this test failed.

This is the error message I got:
```
Failures:
  1) WebHookUserSerializer should only include the required keys
     Failure/Error: message << "#{difference} key(s) have been added to this serializer."
     
     FrozenError:
       can't modify frozen String
     # ./spec/serializers/web_hook_user_serializer_spec.rb:34:in `block (3 levels) in <main>'
     # ./spec/serializers/web_hook_user_serializer_spec.rb:28:in `block (2 levels) in <main>'
```

With this change, the new error message would be (if I didn't fix the error)
```
Failures:

  1) WebHookUserSerializer should only include the required keys
     Failure/Error:
       expect(difference).to eq(0), lambda {
         message = (difference < 0 ?
                   "#{difference * -1} key(s) have been removed from this serializer." :
                   "#{difference} key(s) have been added to this serializer.") +
                   "\nPlease verify if those key(s) are required as part of the web hook's payload."
       }

       1 key(s) have been added to this serializer.
       Please verify if those key(s) are required as part of the web hook's payload.
     # ./spec/serializers/web_hook_user_serializer_spec.rb:28:in `block (2 levels) in <top (required)>'

```

eviltrout: Oh right if you add to to the user serializer this is legit. I personally hate this test that fails every time we do that. You have to account for the new column in this test.

app/assets/javascripts/discourse/components/topic-footer-buttons.js.es6
```diff
@@ -58,5 +58,15 @@ export default Component.extend({
 
   @discourseComputed("topic.message_archived")
   archiveLabel: archived =>
-    archived ? "topic.move_to_inbox.title" : "topic.archive_message.title"
+    archived ? "topic.move_to_inbox.title" : "topic.archive_message.title",
+
+  @discourseComputed("topic.id", "currentUser.featured_topic")
```
jjaffeux: How about:

```
import { propertyEqual } from "discourse/lib/computed";

topicFeaturedOnProfile: propertyEqual("topic.id", ""currentUser.featured_topic.id")
```

app/assets/javascripts/discourse/components/user-card-contents.js.es6
```diff
@@ -30,6 +30,7 @@ export default Component.extend(CardContentsBase, CanCheckEmails, CleansUp, {
   showBadges: setting("enable_badges"),
 
   postStream: alias("topic.postStream"),
+  featuredTopic: alias("user.featured_topic"),
```
jjaffeux: `reads` would probably be enough

app/assets/javascripts/discourse/controllers/topic.js.es6
```diff
@@ -679,6 +679,24 @@ export default Controller.extend(bufferedProperty("model"), {
...
+        this.model
+          .toggleFeaturedOnProfile(this.currentUser)
+          .catch(popupAjaxError);
+      };
+
+      if (
+        this.currentUser.featured_topic &&
+        this.currentUser.featured_topic.id !== this.model.id
```
jjaffeux: Could be simplify by:

```
this.get("currentUser.featured_topic.id") !== this.model.id
```

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -444,6 +444,22 @@ const Topic = RestModel.extend({
     });
   },
 
+  toggleFeaturedOnProfile(user) {
+    const removing = user.featured_topic && this.id === user.featured_topic.id;
```
jjaffeux:
```suggestion
    const removing = user.get("featured_topic.id") === this.id;
```

app/assets/javascripts/discourse/controllers/topic.js.es6
markvanlan: These statements are not equivalent, and does not work for what I need. 

I do not want to entire this branch of the if, when the featured_topic is null. `this.get("currentUser.featured_topic.id") !== this.model.id` evaluates to `true` when the `featured_topic` is null.

app/assets/javascripts/discourse/components/user-card-contents.js.es6
```diff
@@ -50,6 +50,13 @@ export default Component.extend(CardContentsBase, CanCheckEmails, CleansUp, {
   // If inside a topic
   topicPostCount: null,
 
+  @discourseComputed("user.featured_topic")
+  showFeaturedTopic(featuredTopic) {
+    return (
+      featuredTopic && this.siteSettings.allow_featured_topic_on_user_profiles
+    );
+  },
+
```
jjaffeux: What about:

```
import { and } from '@ember/object/computed';

showFeaturedTopic: and("user.featured_topic", "siteSettings.allow_featured_topic_on_user_profiles")
```

app/assets/javascripts/discourse/controllers/user.js.es6
```diff
@@ -100,6 +100,13 @@ export default Controller.extend(CanCheckEmails, {
     return prioritizeNameInUx(name, this.siteSettings);
   },
 
+  @discourseComputed("model.featured_topic")
```
jjaffeux: what about:

```
import { and } from '@ember/object/computed';

showFeaturedTopic: and("model.featured_topic", "siteSettings. allow_featured_topic_on_user_profiles")
```

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -444,6 +444,21 @@ const Topic = RestModel.extend({
...
+  toggleFeaturedOnProfile(user) {
+    const removing = user.get("featured_topic.id") === this.id;
+    const path = removing ? "clear-featured-topic" : "feature-topic";
+    return ajax(`/u/${user.username}/${path}`, {
+      type: "PUT",
+      data: { topic_id: this.id }
+    })
+      .then(() => {
+        const newFeaturedTopic = removing ? null : this;
+        user.set("featured_topic", newFeaturedTopic);
```
jjaffeux: can we make it `featuredTopic` instead?

PR #8462, opened 2019-12-04
REFACTOR: post_destroyer destroy method

markvanlan:


lib/post_destroyer.rb
```diff
@@ -75,10 +76,10 @@ def destroy
     DiscourseEvent.trigger(:post_destroyed, @post, @opts, @user)
     WebHook.enqueue_post_hooks(:post_destroyed, @post, payload)
 
-    if @post.is_first_post? && @post.topic
-      UserActionManager.topic_destroyed(@post.topic)
-      DiscourseEvent.trigger(:topic_destroyed, @post.topic, @user)
-      WebHook.enqueue_topic_hooks(:topic_destroyed, @post.topic, topic_payload)
+    if is_first_post
```
eviltrout: I should note that this could potentially have different outcomes, if the @post was mutated by any of the above dozen lines or so of code when you stored it in a variable. It's unlikely though, but something to keep in mind.

PR #8463, opened 2019-12-04
FIX: bufferedValuesString.split is not a function

lis2:
Fix an error `bufferedValuesString.split is not a function`.
It is happening when bufferedValue is a positive boolean (checkbox value)


lis2:
@eviltrout nice one, thank you, fixed

PR #8466, opened 2019-12-06
Improving bookmarks part 1

martin-brennan:
**Note: All of this functionality is hidden behind a hidden, default false, site setting called `enable_bookmarks_with_reminders`. Also, any feedback on Ember code would be greatly appreciated!**

This is part 1 of the bookmark improvements. The next PR will address the backend logic to send reminder notifications for bookmarked posts to users. This PR adds the following functionality:

* We are adding a new `bookmarks` table and `Bookmark` model to make the bookmarks a first-class citizen and to allow attaching reminders to them.
* Posts now have a new button in their actions menu that has the icon of an actual book

image

* Clicking the button opens the new bookmark modal:

image

* Both name and the reminder type are optional.
* If you close the modal without doing anything, the bookmark is saved with no reminder.
* If you click the Cancel button, no bookmark is saved at all.
* All of the reminder type tiles are dynamic and the times they show will be based on your user timezone set in your profile (this should already be set for you).

image

* If for some reason a user does not have their timezone set they will not be able to set a reminder, but they will still be able to create a bookmark.

image

* A bookmark can be deleted by clicking on the book icon again which will be red if the post is bookmarked.

This PR does NOT do anything to migrate or change existing bookmarks in the form of `PostActions`, the two features live side-by-side here. Also this does nothing to the topic bookmarking.


app/assets/javascripts/discourse/components/tap-tile.js.es6
```diff
@@ -0,0 +1,15 @@
+import Component from "@ember/component";
+import discourseComputed from "discourse-common/utils/decorators";
+
+export default Component.extend({
+  classNames: ["tap-tile"],
+  classNameBindings: ["active"],
+  click: function() {
```
jjaffeux:
```suggestion
  click() {
```

app/assets/javascripts/discourse/components/tap-tile.js.es6
```diff
@@ -0,0 +1,15 @@
...
+    this.onSelect(this.tileId);
+  },
+
+  @discourseComputed("activeTile", "tileId")
```
jjaffeux: check `propertyEqual` in codebase

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+    this.setProperties({
+      errorMessage: null,
+      loading: true,
+      name: null,
+      selectedReminderType: null,
+      closeWithoutSaving: false,
+      saveButtonClicked: false
+    });
+
+    this.set("loading", false);
```
jjaffeux: not sure I understand here:

we have `loading:true` and `loading:false` in the same runloop that feels wrong. Could you explain the reason please?

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+    if (!this.closeWithoutSaving && !this.saveButtonClicked) {
+      this.saveBookmark();
+    }
+    if (this.onCloseWithoutSaving && this.closeWithoutSaving) {
+      this.onCloseWithoutSaving();
+    }
+  },
+
+  @discourseComputed()
+  usingMobileDevice() {
```
jjaffeux:
```suggestion
  usingMobileDevice: reads("site.mobileView")
```

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+    return this.site.mobileView;
+  },
+
+  @discourseComputed()
+  reminderTypes() {
```
jjaffeux: this would work probably, I have a doubt:

``` 
reminderTypes: REMINDER_TYPES
```

or 

```
reminderTypes: computed(function() { return REMINDER_TYPES; })
```

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+    return htmlSafe(
+      I18n.t("bookmarks.reminders.next_month", {
+        date: this.nextMonth().format(I18n.t("dates.month_day_time"))
+      })
+    );
+  },
+
+  @discourseComputed()
+  userHasTimezoneSet() {
+    return !_.isEmpty(this.userTimezone());
```
jjaffeux:
```
userHasTimezoneSet: notEmpty("userTimezone")
```

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+    return later.minutes() < 30
+      ? later.minutes(30)
+      : later.add(30, "minutes").startOf("hour");
+  },
+
+  actions: {
+    saveAndClose() {
+      this.saveButtonClicked = true;
+      this.saveBookmark()
+        .then(() => {
```
jjaffeux: could be inline:

```
then(() => this.send("closeModal"))
```
```

File: app/assets/javascripts/discourse/controllers/bookmark.js.es6 [hunk 8]
```diff
@@ -0,0 +1,224 @@
...
+          this.send("closeModal");
+        })
+        .catch(e => {
+          this.saveButtonClicked = false;
```
jjaffeux: think the name is a little bit misleading here. Actually the button has been clicked, but it's not "trying to save anymore" as it failed. So maybe something like: `isSavingBookmark` ?

app/assets/javascripts/discourse/lib/transform-post.js.es6
```diff
@@ -35,6 +35,8 @@ export function transformBasicPost(post) {
     username: post.username,
     avatar_template: post.avatar_template,
     bookmarked: post.bookmarked,
+    bookmarked_with_reminder: post.bookmarked_with_reminder,
```
jjaffeux:
```suggestion
    bookmarkedWithReminder: post.bookmarked_with_reminder,
```

app/assets/javascripts/discourse/lib/transform-post.js.es6
```diff
@@ -35,6 +35,8 @@ export function transformBasicPost(post) {
...
+    bookmark_reminder_at: post.bookmark_reminder_at,
```
jjaffeux:
```suggestion
    bookmarkReminderAt: post.bookmark_reminder_at,
```

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -336,6 +337,34 @@ const Post = RestModel.extend({
...
+      });
+      controller.setProperties({
+        onCloseWithoutSaving: () => {
+          this.toggleProperty("bookmarked_with_reminder");
+          this.appEvents.trigger("post-stream:refresh", { id: this.id });
+        }
+      });
+    } else {
+      return Post.destroyBookmark(this.id)
+        .then(() => {
```
jjaffeux: can be inlined

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -301,6 +301,41 @@ registerButton("bookmark", attrs => {
   };
 });
 
+registerButton("bookmarkWithReminder", (attrs, state, siteSettings) => {
+  if (!attrs.canBookmark || !siteSettings.enable_bookmarks_with_reminders) {
+    return;
+  }
+
+  let className = "bookmark with-reminder";
```
jjaffeux: I think a cleaner pattern for classnames is to do this:

```
const classNames = ["bookmark", "with-reminder"];

if (foo) {
  classNames.push("bookmarked");
}

return classnames.join(" ");
```

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -301,6 +301,41 @@ registerButton("bookmark", attrs => {
...
+      };
+    } else {
+      title = "bookmarks.created";
+    }
+  }
+
+  return {
+    id: attrs.bookmarked_with_reminder ? "unbookmark" : "bookmark",
+    action: "toggleBookmarkWithReminder",
+    title: title,
```
jjaffeux:
```suggestion
    title,
```

app/assets/javascripts/discourse/widgets/post-menu.js.es6
```diff
@@ -301,6 +301,41 @@ registerButton("bookmark", attrs => {
...
+    titleOptions: titleOptions,
```
jjaffeux:
```suggestion
    titleOptions,
```

app/assets/stylesheets/common/components/tap-tile.scss
```diff
@@ -0,0 +1,39 @@
...
+  .tap-tile {
+    padding: 10px 5px;
+    display: flex;
+    flex-direction: column;
+    text-align: center;
+    align-items: center;
+    border: 2px solid #c7d2db;
+    margin: 0 0 20px;
+    width: 28%;
+    font-size: 12px;
```
jjaffeux: look at using $font-size

app/assets/stylesheets/common/components/tap-tile.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  .tap-tile {
+    padding: 10px 5px;
+    display: flex;
+    flex-direction: column;
+    text-align: center;
+    align-items: center;
+    border: 2px solid #c7d2db;
+    margin: 0 0 20px;
+    width: 28%;
```
jjaffeux: feels too magical to me, could we have something less arbitrary?

app/assets/stylesheets/common/components/tap-tile.scss
```diff
@@ -0,0 +1,39 @@
...
+    .svg-icon-title {
+      width: 2.5em;
+      height: 2.5em;
+      margin-bottom: 10px;
+    }
+  }
+
+  &::after {
+    content: "";
+    width: 31%;
```
jjaffeux: same than before this feels a little bit brittle to me

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+  tomorrow() {
+    return this.startOfDay(this.now().add(1, "day"));
+  },
+
+  startOfDay(momentDate) {
+    return momentDate.hour(START_OF_DAY_HOUR).startOf("hour");
+  },
+
+  userTimezone() {
+    return this.currentUser.timezone;
```
jjaffeux: are we always sure we have a currentUser here ?

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+  saveBookmark() {
+    const reminderAt = this.reminderAt();
+    const data = {
+      reminder_type: this.selectedReminderType,
+      reminder_at: reminderAt ? reminderAt.toISOString() : null,
+      name: this.name,
+      post_id: this.model.postId
+    };
+
+    return ajax("/bookmarks", { type: "POST", data: data });
```
jjaffeux:
```suggestion
    return ajax("/bookmarks", { type: "POST", data });
```

app/assets/javascripts/discourse/controllers/bookmark.js.es6
```diff
@@ -0,0 +1,224 @@
...
+  },
+
+  reminderAt() {
+    if (this.selectedReminderType === null) {
```
jjaffeux: Should be enough?

```suggestion
    if (!this.selectedReminderType) {
```

app/assets/stylesheets/common/components/bookmark-modal.scss
```diff
@@ -0,0 +1,27 @@
+.bookmark-with-reminder.modal {
+  .modal-body {
+    max-width: 410px;
+    min-width: 380px;
+
+    .control-label {
+      font-weight: 700;
```
jjaffeux: $font-size

config/locales/server.en.yml
```diff
@@ -561,6 +565,11 @@ en:
           attributes:
             word:
               too_many: "Too many words for that action"
+        bookmark:
+          attributes:
+            reminder_at:
+              blank: "time must be provided for all reminders except 'Next time I'm at my Desktop'"
```
gschlager: It might be a good idea to replace `Next time I'm at my Desktop` with a placeholder like `%{reminder_name}` to ensure that the name doesn't get out of sync from the actual translation of `bookmarks.reminders.at_desktop`. There's already a minor difference right now. 😉

app/assets/javascripts/discourse/controllers/bookmark.js.es6
martin-brennan: @jjaffeux this was a goof, I didn't mean to leave that in there; was just boilerplate from other modal controllers.

app/assets/javascripts/discourse/widgets/post-menu.js.es6
martin-brennan: That is much better and more flexible, thank you

app/assets/javascripts/discourse/widgets/post-menu.js.es6
martin-brennan: I'm still not used to this ES6 property syntax, it's only been almost 5 years :roll_eyes:

app/assets/javascripts/discourse/controllers/bookmark.js.es6
martin-brennan: Yes, you cannot open the bookmark modal (and the old bookmark button does not work either) unless you have a current user. Will just double check that I can't get to it anyway

app/assets/stylesheets/common/components/tap-tile.scss
martin-brennan: For posterity this was referring to font-0, font-down-*, and font-up-* in `app/assets/stylesheets/common/foundation/variables.scss`

PR #8467, opened 2019-12-06
FIX: Muted tags are respected by TopicTrackingState

lis2:
When the tag is muted and topic contains that tag, we should not mark that message as NEW.

There are 3 possible settings which site admin can set.
remove_muted_tags_from_latest - always
It means that if the topic got at least one muted tag, we should not mark that topic as NEW

remove_muted_tags_from_latest - only muted
Similar to above, however, if at least one tag is not muted, the topic is still marked as NEW

remove_muted_tags_from_latest - never
Basically, mute tag setting is ignored and all topics are set as NEW

### DIFF
--- a/app/assets/javascripts/discourse/models/topic-tracking-state.js.es6
+++ b/app/assets/javascripts/discourse/models/topic-tracking-state.js.es6
@@ -31,6 +31,18 @@ function isUnseen(topic) {
   return !topic.is_seen;
 }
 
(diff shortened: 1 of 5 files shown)


SamSaffron:
This is a nice change I say go ahead and merge tomorrow. 

Just a quick one, can you confirm using rack mini profiler that muted tags are only selected once per req from SQL.

PR #8468, opened 2019-12-06
Rename 'target usernames' with 'target recipients' in Composer

nbianca:
This introduces a series of deprecation notices for 'usernames' and replaces some occurrences


eviltrout:
Is this just the first step in this refactoring? 

Because while it is an improvement to not call it usernames anymore, a front end theme/plugin still doesn't know if any value is a username/group/email. I'd imagined storing the recipients as a list of objects with a type. For example:

`{ value: 'eviltrout', type: 'username' }, { value: 'admins', type: 'group' }`

nbianca:
Sorry for late reply here, @eviltrout. 😊Yes, you are correct, this is the first step and the next one is going to be huge.

The biggest problem right now is that 'user-selector' loses all information about types (group, users or email) and it also concatenates all usernames in one huge string. This is definitely not ideal and it requires a pretty big refactoring. I decided to submit this PR first, because I wanted to put in the deprecation notice as soon as possible. 🙏

eviltrout:
@nbianca great! I just wanted to confirm that there was more coming. Approved!

app/assets/javascripts/discourse/components/composer-messages.js.es6
```diff
@@ -97,17 +97,9 @@ export default Component.extend({
...
-      if (usernames) {
-        usernames = usernames.split(",");
-      }
-
-      if (
-        usernames &&
-        usernames.length === 1 &&
-        usernames[0] === this.currentUser.get("username")
-      ) {
+      if (recipients.every(r => r === this.currentUser.get("username"))) {
```
ZogStriP: I think you also need to check for the length of recipients because `[].every(() => false) == true`

app/assets/javascripts/discourse/controllers/composer.js.es6
```diff
@@ -909,19 +910,24 @@ export default Controller.extend({
       isWarning: false
     });
 
-    if (opts.usernames && !this.get("model.targetUsernames")) {
-      this.set("model.targetUsernames", opts.usernames);
+    if (!this.model.targetRecipients) {
+      if (opts.usernames) {
+        deprecated("`usernames` is deprecated, use `recipients` instead.");
+        this.model.set("targetRecipients", opts.usernames);
```
ZogStriP: I think all those `.set()` should now be replaced with the more direct `this.model.targetRecipients = opts.usernames`.

PR #8469, opened 2019-12-06
FEATURE: Add 'groups' option to polls

nbianca:


plugins/poll/config/locales/server.en.yml
```diff
@@ -57,6 +57,7 @@ en:
     post_is_deleted: "Cannot act on a deleted post."
 
     user_cant_post_in_topic: "You can't vote because you can't post in this topic."
+    user_cant_vote: "You can't vote because your trust level is too low."
```
ZogStriP: Could we change this to say something like "You need to be at least XYZ to be able to vote on this poll"?

nbianca:
Updated the PR with new UI:

Screenshot 2019-12-10 at 22 13 45

SamSaffron:
I hate to put a spanner in this works, but I wonder if we should just add support for groups instead here? if we did groups then the change would be extremely similar in magnitude but much more flexible. 

We ship with the group: https://meta.discourse.org/g/trust_level_1 and so on, so the exact same feature will work, except that a poll could be set up restricted to a subset of users.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -333,16 +333,39 @@ createWidget("discourse-poll-container", {
...
-        options.map(option => {
-          return this.attach("discourse-poll-option", {
-            option,
-            isMultiple: attrs.isMultiple,
-            vote: attrs.vote
-          });
-        })
+      const contents = [];
+
+      const pollGroups = poll.group_name && poll.group_name.split(",");
```
ZogStriP: I wonder, why is this called `group_name` when it can hold several groups? Why not call this `groups`?

ZogStriP:
Where are we on this?

plugins/poll/app/models/poll.rb
```diff
@@ -78,6 +78,7 @@ def can_see_voters?(user)
 #  created_at       :datetime         not null
 #  updated_at       :datetime         not null
 #  chart_type       :integer          default("bar"), not null
+#  groups           :string
```
ZogStriP: I wonder if we should instead use an array of group_ids rather than a list of group names... 🤔

nbianca: We had a discussion about this and we realized that there will be some conversions anyway, from group names to group IDs and `.to_i`s or `parseInt`s. I see no advantage or disadvantage to either of the options.

PR #8471, opened 2019-12-06
FIX: Optimize images in Onebox

udan11:


eviltrout:
Can you give some more context for this PR please?

udan11:
`CookedPostProcessor` optimizes some images, usually the same that are also lightboxed and the images in Onebox were not optimized.

Last week I submitted PR #8427 which was fixing a similar bug, but for images inside quotes.

ZogStriP:
LGTM but this definitely needs a test 👍

PR #8480, opened 2019-12-08
DEV: See if the store is external before checking disk space

CvX:
`available_disk_space` calls `df` which exits with an error if the `uploads` path doesn't exist. That's often the case when the `Discourse.store.external?` is true.

By doing the `external?` check first the `disable_if_low_on_disk_space` does less work and doesn't output any errors to the console.


SamSaffron:
Nice change, I see this in logs occasionally.

PR #8482, opened 2019-12-09
FEATURE: Warning when component is added and not assigned

lis2:
When a component is installed and not assigned to any theme and the user is changing page, we should display a warning.

If the user decides to skip warning or come back later, a warning should not be shown anymore.

Also, when the user clicks "delete" button right after installation, warning about forgotten themes should not be shown.

leave-confirmation


SamSaffron:
This looks good to me

PR #8496, opened 2019-12-09
FIX: default draft key in openComposerWithTopicParams

markvanlan:


markvanlan:
I need to use the this route action, and no draftKey is passed in. This is for creating a new topic, so I dont see why we wouldnt default to a new_draft_key, and the tests all pass

https://github.com/discourse/discourse/blob/aa24be1a9a1907870fc68b2450a601fd221c2990/app/assets/javascripts/discourse/routes/application.js.es6#L214

PR #8505, opened 2019-12-09
DEV: revert webpush version bump

markvanlan:


darix:
some explanation *why* you rever it would have been good.

PR #8516, opened 2019-12-10
UX: Include public groups in mentionable groups set

nbianca:


eviltrout:
Could you please provide a little more context for the change?

nbianca:
This commit includes public groups in autocomplete and also wraps group mentions in a link as long as the group is visible to everyone, even though the group is not mentionable.

app/services/post_alerter.rb
```diff
@@ -458,7 +458,7 @@ def push_notification(user, payload)
   def expand_group_mentions(groups, post)
     return unless post.user && groups
 
-    Group.mentionable(post.user).where(id: groups.map(&:id)).each do |group|
+    Group.mentionable(post.user, include_public: false).where(id: groups.map(&:id)).each do |group|
```
ZogStriP: That API feels a little weird. I think the UI should instead use something like `Group.visible` and on the backend, we rely on `Group.mentionnable` to generate the notifications.

spec/requests/groups_controller_spec.rb
```diff
@@ -451,6 +451,11 @@ def expect_type_to_return_right_groups(type, expected_group_ids)
     it "should return the right response" do
       sign_in(user)
 
+      group.update!(
```
ZogStriP: For this feature, I was expecting more like an integration test that would 

1. create a public group that can't be mentioned (with at least one member)
2. create a post with an @mention to that group
3. ensure no mention notifications were sent to members of the group
4. ensure the mention gets wrapped in the cooked version of the post

app/services/post_alerter.rb
nbianca: It feels a little bit weird and I was not happy at first either, but I have a few arguments why this makes sense: 🤔 

We already have `Group.visible_groups`, but that's not what we want here. What we want is really just mentionable groups, which sometimes can include the public ones as well. Also, it is consistent with the API for `Group.visible_groups`, which has a `include_everyone` option.

spec/requests/groups_controller_spec.rb
nbianca: Each individual step is tested somewhere, cooked_post_processor_spec, post_alerter_spec or pretty_text_spec, but I agree that an integration test would be ideal, so I wrote one. :)

spec/requests/users_controller_spec.rb
```diff
@@ -3163,26 +3163,27 @@ def post_user
     context 'groups' do
       let!(:mentionable_group) do
         Fabricate(:group,
-          mentionable_level: 99,
-          messageable_level: 0,
-          visibility_level: 0,
+          mentionable_level: Group::ALIAS_LEVELS[:everyone],
```
ZogStriP: Nice 👍

PR #8520, opened 2019-12-10
FEATURE: Create IE Support Plugin

eviltrout:
This core plugin, which could be split off in the future, allows us to load IE specific code on demand.

Co-authored-by: jjaffeux <j.jaffeux@gmail.com>


plugins/discourse-internet-explorer/plugin.rb
```diff
@@ -0,0 +1,57 @@
...
+          return unless request.env['HTTP_USER_AGENT'] =~ /MSIE|Trident/
+        end
+
+        previous_discourse_stylesheet_link_tag(name, opts)
+      end
+    end
+  end
+
+  register_anonymous_cache_key(:ie) do
+    unless defined?(@is_ie)
```
jjaffeux: can't we do something like:

```ruby
@is_ie ||= begin
session = @env[self.class::RACK_SESSION]
    # don't initialize params until later
    # otherwise you get a broken params on the request
    params = {}

    BrowserDetection.browser(@env[self.class::USER_AGENT]) == :ie
end
```

eviltrout: That will run again if it's false though.

jjaffeux: indeed 🙃

SamSaffron:
This is awesome!

I wonder if we should use user-agent sniffing on the server to avoid shipping the bundle to non IE? This would be less about performance vs making sure the polyfills we think we don't need we actually do not need.

plugins/discourse-internet-explorer/spec/requests/bootstrap_request_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      SiteSetting.discourse_internet_explorer_enabled = true
+    end
+
+    it "includes the IE js and css" do
+      get "/categories", headers: { "HTTP_USER_AGENT" => ie_agent }
+      expect(response.body).to match(/discourse-internet-explorer\/js\/ie.js/)
+      expect(response.body).to match(/stylesheets\/discourse-internet-explorer/)
+    end
+
+    it "doesn't include IE stuff for non-IE browsers" do
```
SamSaffron: Awesome, I love that we will not ship this to non IE, will help catch bugs as well with polyfills we think we don't need that we do.

jjaffeux:
Just a note here if it's unclear. There more css code to move into this. I didn't do it so far, but I expect kris and johani to have a look this.

PR #8521, opened 2019-12-10
FIX: Use updated_at date to denote expired invites

pmusaraj:
Fixes an issue where resending invites did not reset the expired state of invites in the admin UI.

### DIFF
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -55,7 +55,7 @@ class Invite < ActiveRecord::Base
   end
 
   def expired?
-    created_at < SiteSetting.invite_expiry_days.days.ago
+    updated_at < SiteSetting.invite_expiry_days.days.ago
   end
 
   # link_valid? indicates whether the invite link can be used to log in to the site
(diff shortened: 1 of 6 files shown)


pmusaraj:
I added one on `invited_by_id` given that there are many calls to `find_all_invites_from`, and the inviter's ID is the common denominator among all of them. 

I'm not sure how to check whether this benefits the query, though. Any ideas?

lis2:
How about the index on `updated_at`? I know that `created_at` didn't have index before, but I am curious if that might be beneficial

eviltrout:
@pmusaraj you can test by running the select queries from your console (or mini profiler) in postgres, using explain and such. But I think invited_by is a big enough improvement that we are probably OK to go ahead.

@lis2's idea might be worth a follow up, but again you'd need some test data and a DB console to try it out.

Once merged we should check the performance in production on sites with invites and see if all is well.

techAPJ:
I believe this line needs to be updated as well?

https://github.com/discourse/discourse/blob/eb9155f3fef34ae7b474bdf3d2655bd1a043cbd3/app/models/invite_redeemer.rb#L91

Other than that the PR looks good! 👍

PR #8528, opened 2019-12-11
FIX: Replace deprecated URI.encode, URI.escape, URI.unescape and URI.unencode

martin-brennan:
The following methods have long been deprecated in ruby due to flaws in their implementation per http://blade.nagaokaut.ac.jp/cgi-bin/vframe.rb/ruby/ruby-core/29293?29179-31097:

* URI.escape
* URI.unescape
* URI.encode
* URI.unencode

escape/encode are just aliases for one another. This PR uses the [Addressable gem](https://github.com/sporkmonger/addressable) to replace these methods with its own `encode`, `unencode`, and `encode_component` methods where appropriate.

I have put all references to `Addressable::URI` here into the `UrlHelper` to keep them corralled in one place to make changes to this implementation easier.


SamSaffron:
Minor but worth cleaning this up.

This makes addressable an explicit vs implicit dep.

We should add it to the Gemfile explicitly then. At the moment this is working due to the luck of `rotp` not being `require: false`

Once that is done I am happy with this, go ahead and merge.

martin-brennan:
Done, leaving out `require: false` for now. I am going to make another PR after this to make ROTP `require: false` now that you mention it.

PR #8529, opened 2019-12-11
UX: attempts to make charts loading less laggy

jjaffeux:


jjaffeux:
Basically it's removing the chart animation and using a simple CSS fading

PR #8533, opened 2019-12-11
FIX: Granting staff status should auto-approve users waiting approval

romanrizzi:


app/controllers/admin/users_controller.rb
```diff
@@ -471,7 +471,6 @@ def invite_admin
     end
 
     user.active = true
-    user.approved = true
```
SamSaffron: do we need an extra test somewhere to confirm that if there is no reviewable the user still gets approved?

romanrizzi: This is already covered by the `users_controller_spec`. We could also add an additional test for the user class.

app/models/concerns/roleable.rb
```diff
@@ -48,4 +50,15 @@ def set_permission(permission_name, value)
     save_and_refresh_staff_groups!
   end
 
+  private
+
+  def auto_approve_user
+    reviewable = ReviewableUser.where(status: Reviewable.statuses[:pending]).find_by(target: self)
```
eviltrout:
```suggestion
    reviewable = ReviewableUser.find_by(target: self, status: Reviewable.statuses[:pending])
```
This is equivalent I think and slightly shorter.

eviltrout: You could also do `if reviewable = ...` and cut out a line.

PR #8540, opened 2019-12-12
FIX: require: false for rotp gem

martin-brennan:
The ROTP gem is only used in a very small amount of places in the app, we don't need to globally require it.

Also set the Addressable gem to not have a specific version range, as it has not been a problem yet.

Some slight refactoring of `UserSecondFactor` here too to use `SecondFactorManager` to avoid code repetition

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -134,7 +134,7 @@ gem 'highline', '~> 1.7.0', require: false
 gem 'rack-protection' # security
 gem 'cbor', require: false
 gem 'cose', require: false
-gem 'addressable', '~> 2.7.0'
+gem 'addressable'
 
 # Gems used only for assets and not required in production environments by default.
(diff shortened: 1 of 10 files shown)


SamSaffron:
Yeah Robin is totally right here, if we want to defer load the gems we need to do it at runtime with something like:

```
def some_method
   require 'rotp' if !defined? ROTP
   ...
end
```

I think it is totally legit to do this for 2fa stuff cause it is so rarely called. I have a feeling though Addressable is going to be loaded pretty early.

martin-brennan:
@SamSaffron @eviltrout that makes total sense, I didn't even think of the models + controllers being loaded at all times, but of course they are. I will make the ROTP change, and probably just take `require: false` off of Addressable, as it is likely to be loaded vert early

PR #8542, opened 2019-12-12
FIX: allow focus of input on Android mobile devices

pmusaraj:
This reverts commit bfcf8ed61bebba6be3082507dafa6766024ddf3a.

We shouldn't disable focus on select-kit inputs on mobile devices, I think it's a better user experience to allow focus. Unfortunately, iOS already blocks focusing on inputs via a JS call. This revert only affects Android devices.


jjaffeux:
As discussed LGTM :+1: This will change with select-kit 2, so we should make sure to track this in the next weeks.

PR #8543, opened 2019-12-13
FIX: Adds a message when the passwords doesn't match in task admin:create

hectorbus:
If the password and password confirmation doesn't match when using `rake admin:create` to create an admin user, it shows a message indicating that. Previously it didn't say anything and it was confusing.

Screen Shot 2019-12-12 at 5 25 09 PM


SamSaffron:
sure this looks fine to me, thanks!

PR #8545, opened 2019-12-13
FEATURE: Modal for profile featured topic & admin wrench refactor

markvanlan:


app/assets/javascripts/discourse/components/choose-topic.js.es6
```diff
@@ -4,14 +4,38 @@ import Component from "@ember/component";
...
   topics: null,
+  selectedTopic: null,
   selectedTopicId: null,
   currentTopicId: null,
-  topicTitle: null,
+  additionalFilters: "",
+  topicTitle: "",
+  label: null,
+  loadOnInit: false,
+  topicChangedCallback: () => {},
```
jjaffeux: This is probably leaking, shouldn't init anything but string, null and booleans here.

Should be set at null here, and do this in init

app/assets/javascripts/discourse/controllers/preferences/profile.js.es6
```diff
@@ -48,6 +50,31 @@ export default Controller.extend(PreferencesTabController, {
...
+        title: "user.feature_topic_on_profile.title"
+      });
+    },
+
+    clearFeaturedTopicFromProfile() {
+      bootbox.confirm(
+        I18n.t("user.feature_topic_on_profile.clear.warning"),
+        result => {
+          if (result) {
+            return ajax(`/u/${this.model.username}/clear-featured-topic`, {
```
jjaffeux: Don't know much about bootbox usage, but what are the uses of these return here and in then ?

markvanlan: Yeah, removing both returns does not change functionality.

Gemfile.lock
```diff
@@ -45,7 +45,6 @@ GEM
       rake (>= 10.4, < 14.0)
     ast (2.4.0)
     aws-eventstream (1.0.3)
-
```
eviltrout: It looks like this change was included by mistake. The gems didn't change so the whitespace in the lock file should not either.

app/assets/javascripts/discourse/components/choose-topic.js.es6
```diff
@@ -4,14 +4,38 @@ import Component from "@ember/component";
...
 
 export default Component.extend({
   loading: null,
   noResults: null,
   topics: null,
+  selectedTopic: null,
   selectedTopicId: null,
   currentTopicId: null,
-  topicTitle: null,
+  additionalFilters: "",
```
eviltrout: It's slightly dangerous to initialize to `""` as all instances will point at the same string. In general with strings we replace them, but there are operations in Javascript that mutate them. Instead you should set these to `null` and initialize them in `init`

app/assets/javascripts/discourse/components/choose-topic.js.es6
```diff
@@ -24,6 +48,17 @@ export default Component.extend({
     this.search(this.topicTitle);
   },
 
+  @observes("selectedTopic")
```
eviltrout: We are trying to avoid observers wherever possible. Could you instead have the action that selects a topic trigger this operation rather than observing?

app/assets/javascripts/discourse/components/topic-footer-buttons.js.es6
```diff
@@ -46,6 +40,7 @@ export default Component.extend({
 
   @discourseComputed
   showAdminButton() {
+    return true;
```
eviltrout: If `showAdminButton` is always true it probably doesn't need to exist does it?

app/assets/javascripts/discourse/controllers/feature-topic-on-profile.js.es6
```diff
@@ -0,0 +1,33 @@
...
+
+  actions: {
+    save() {
+      return ajax(`/u/${this.model.username}/feature-topic`, {
+        type: "PUT",
+        data: { topic_id: this.newFeaturedTopic.id }
+      })
+        .then(() => {
+          this.model.set("featured_topic", this.newFeaturedTopic);
+          return this.send("closeModal");
```
eviltrout:
```suggestion
          this.send("closeModal");
```
this doesn't return a promise so the `return` is probably unecessary here.

app/assets/javascripts/discourse/components/topic-footer-buttons.js.es6
markvanlan: woops! temporary override made its way in

app/assets/javascripts/discourse/components/choose-topic.js.es6
```diff
@@ -4,14 +4,41 @@ import Component from "@ember/component";
...
+    this.topicTitle = this.topicTitle || "";
+
+    if (this.loadOnInit && !isEmpty(this.additionalFilters)) {
+      searchForTerm(this.additionalFilters, {}).then(results => {
+        if (results && results.posts && results.posts.length > 0) {
+          this.set(
+            "topics",
+            results.posts
+              .mapBy("topic")
+              .filter(t => t.id !== this.currentTopicId)
```
spirobel: Can you access "this" here? I think the code should be similar to: https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/choose-topic.js.es6#L92

PR #8548, opened 2019-12-15
FIX: prevents crash in discourse_tagging with empty term

jjaffeux:


lib/discourse_tagging.rb
```diff
@@ -278,9 +278,7 @@ def self.filter_allowed_tags(guardian, opts = {})
 
     term = opts[:term]
     if term.present?
-      term = term.gsub("_", "\\_")
-      clean_tag(term)
-      term.downcase!
+      term = term.gsub("_", "\\_").downcase
```
nlalonde: Oops, the intention was `term = clean_tag(term)` so that `TAGS_FILTER_REGEXP` is applied before putting using it in the query.

jjaffeux: We already do it there https://github.com/discourse/discourse/blob/master/app/controllers/tags_controller.rb#L220

And the only other place we use `filter_allowed_tags` is in https://github.com/discourse/discourse/blob/master/app/controllers/topics_controller.rb#L329 and there's not term so no need for cleaning as far as I can see.

Now I understand that there might be other cases or future usages. But I think it should be better to decide that cleaning has to happen inside or before but not both.

nlalonde: Ah ok I see what you mean. It has essentially always been this way, so I'll approve it.

PR #8552, opened 2019-12-15
FEATURE: allows plugins to add a global notice 

jjaffeux:
Usage:

```javascript
api.addGlobalNotice(id, text, options = {});
```

Options can be:

```javascript
dismissable // boolean: will display a button to hide the notice if true
html // string: will prepend html to the text if present
level // string: alert level, info, danger...
persistentDismiss // boolean: if true won't show notice again on reload
onDismiss // function: execute a custom action on dismiss
visibility // function: defines custom logic for notice visibility
```

### DIFF
--- a/app/assets/javascripts/discourse/components/global-notice.js.es6
+++ b/app/assets/javascripts/discourse/components/global-notice.js.es6
@@ -196,7 +196,7 @@ export default Component.extend({
         visibility() {
           return !LogsNotice.currentProp("hidden");
         },
-        onDismiss: function() {
+        onDismiss() {
           LogsNotice.currentProp("hidden", true);
           LogsNotice.currentProp("text", "");
         }
--- a/app/assets/javascripts/discourse/components/global-notice.js.es6
+++ b/app/assets/javascripts/discourse/components/global-notice.js.es6
@@ -47,7 +47,7 @@ export default Component.extend({
   willDestroyElement() {
     this._super(...arguments);
 
-    this._teardDownObservers();
+    this._tearDownObservers();
   },
 
   notices: Ember.computed(
--- a/app/assets/javascripts/discourse/components/global-notice.js.es6
+++ b/app/assets/javascripts/discourse/components/global-notice.js.es6
@@ -1,119 +1,208 @@
-import { bind } from "@ember/runloop";
+import { bind, cancel } from "@ember/runloop";
 import Component from "@ember/component";
-import { on } from "discourse-common/utils/decorators";
-import { iconHTML } from "discourse-common/lib/icon-library";
 import LogsNotice from "discourse/services/logs-notice";
-import { bufferedRender } from "discourse-common/lib/buffered-render";
+import EmberObject from "@ember/object";
 
-export default Component.extend(
-  bufferedRender({
-    rerenderTriggers: ["site.isReadOnly", "siteSettings.disable_emails"],
+const _pluginNotices = [];
 
-    buildBuffer(buffer) {
+export function addGlobalNotice(text, id, options = {}) {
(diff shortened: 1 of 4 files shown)


eviltrout:
This should be dismissable per @coding-horror

jjaffeux:
@eviltrout @ZogStriP can you review again please ?

Also linked: https://github.com/discourse/discourse/pull/8577

Note, I tried to write tests for this, but global-notice code is using many things which are not available in tests so it was starting to become sever scope creep.

jjaffeux:
@awesomerobot I added you as a reviewer as I need your validation on alert.scss changes please

app/assets/stylesheets/common/base/alert.scss
```diff
@@ -1,19 +1,28 @@
...
+    align-self: flex-start;
+    margin-left: auto;
+
     .d-icon {
       color: $primary-low-mid;
     }
   }
 
+  .text {
+    flex: 1;
```
awesomerobot: We should use `flex: 1 1 auto;` here.

In modern browsers `flex: 1;` is already `flex: 1 1 auto;` but IE has different defaults so it would be `flex: 1 0 auto;`.

jjaffeux:
```suggestion
    flex: 1 1 auto;
```

app/assets/javascripts/discourse/components/global-notice.js.es6
```diff
@@ -1,119 +1,208 @@
...
+  init() {
+    this._super(...arguments);
+
+    this._setupObservers();
+  },
+
+  willDestroyElement() {
+    this._super(...arguments);
+
+    this._teardDownObservers();
```
eviltrout: typo here, an extra `d`

app/assets/javascripts/discourse/components/global-notice.js.es6
```diff
@@ -1,119 +1,208 @@
...
+  },
+
+  _handleLogsNoticeUpdate() {
+    const logNotice = Notice.create({
+      text: LogsNotice.currentProp("message"),
+      id: "alert-logs-notice",
+      options: {
+        dismissable: true,
+        persistentDismiss: false,
+        visibility: function() {
```
eviltrout:
```suggestion
        visibility() {
```

app/assets/javascripts/discourse/components/global-notice.js.es6
```diff
@@ -1,119 +1,208 @@
...
+          return !LogsNotice.currentProp("hidden");
+        },
+        onDismiss: function() {
```
eviltrout:
```suggestion
        onDismiss() {
```

app/assets/javascripts/discourse/components/global-notice.js.es6
jjaffeux:
```suggestion
    this._tearDownObservers();
```

app/assets/javascripts/discourse/components/global-notice.js.es6
```diff
@@ -1,119 +1,208 @@
...
+    }
+  },
+
+  _setupObservers() {
+    this._boundLogsNoticeHandler = bind(this, this._handleLogsNoticeUpdate);
+    LogsNotice.current().addObserver("hidden", this._boundLogsNoticeHandler);
+    LogsNotice.current().addObserver("text", this._boundLogsNoticeHandler);
+  },
+
+  _teardDownObservers() {
```
jjaffeux:
```suggestion
  _tearDownObservers() {
```

jjaffeux:
This PR is removing a bufferedrenderer can you give it  a look please?

PR #8555, opened 2019-12-16
FIX: category id in filterCategory

lis2:
Small regression was created here: https://github.com/discourse/discourse/commit/374534f00ee#diff-4af46675500edc092f9224ca5835a7ddR106

After that change `listFilter` is including `categoryId` like `c/cat/subcat/6/l/latest` instead of `c/cat/subcat/l/latest`

Therefore, `trackIncoming` function in topic-tracking-state couldn't properly filter new messages


lis2:
@danielwaterworth could you take a look?

PR #8568, opened 2019-12-17
FEATURE: Keyboard shortcut for opening the topic admin menu

hectorbus:
A new keyboard shortcut is implemented. When <kbd>Shift</kbd> + <kbd>a</kbd>, it opens the topic admin actions.

This PR came from [this topic](https://meta.discourse.org/t/keyboard-shortcut-for-closing-a-topic/135102/8).

Screen Shot 2019-12-16 at 9 50 34 PM

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-admin-menu-button.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-admin-menu-button.js.es6
@@ -7,5 +7,23 @@ export default MountWidget.extend({
 
   buildArgs() {
     return this.getProperties("topic", "fixed", "openUpwards", "rightSide");
+  },
+
+  toggleAdminMenu() {
+    $(".toggle-admin-menu")
+      .first()
+      .click();
+  },
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    this.appEvents.on("topic:toggleAdminMenu", this, this.toggleAdminMenu);
+  },
+
+  willDestroyElement() {
+    this._super(...arguments);
+
+    this.appEvents.off("topic:toggleAdminMenu", this, this.toggleAdminMenu);
   }
 });
--- a/app/assets/javascripts/discourse/components/topic-admin-menu-button.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-admin-menu-button.js.es6
@@ -7,23 +7,5 @@ export default MountWidget.extend({
 
   buildArgs() {
     return this.getProperties("topic", "fixed", "openUpwards", "rightSide");
-  },
-
-  toggleAdminMenu() {
-    $(".toggle-admin-menu")
-      .first()
-      .click();
-  },
-
-  didInsertElement() {
-    this._super(...arguments);
-
-    this.appEvents.on("topic:toggleAdminMenu", this, this.toggleAdminMenu);
-  },
-
-  willDestroyElement() {
-    this._super(...arguments);
-
-    this.appEvents.off("topic:toggleAdminMenu", this, this.toggleAdminMenu);
   }
 });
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/discourse/components/topic-admin-menu-button.js.es6
```diff
@@ -7,5 +7,23 @@ export default MountWidget.extend({
...
+  didInsertElement() {
+    this._super(...arguments);
+
+    this.appEvents.on("topic:toggleAdminMenu", this, this.toggleAdminMenu);
+  },
+
+  willDestroyElement() {
+    this._super(...arguments);
+
+    this.appEvents.off("topic:toggleAdminMenu", this, this.toggleAdminMenu);
```
eviltrout: Can we instead use `this.dispatch` to send the event to the widget, which can then toggle itself rather than using a jQuery click event?

hectorbus: Sure, I'll make the change right away.

hectorbus: I'm having some trouble understanding how to implement `this.dispatch`. There is a method called `showAdminMenu` in the `topic-admin-menu-button` widget. This method shows the menu by changing the state `expanded`. But I don't know how to trigger that method when executing the keyboard shortcut.

eviltrout: So here's an example from the site header. In didInsertElement you add `this.dispatch`:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/components/site-header.js.es6#L206

In this case, when `header:keyboard-trigger` is triggered as an app event, it will be dispatched to the `header` widget:

https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/widgets/header.js.es6#L533

the method name `headerKeyboardTrigger` is made automatically from `header:keyboard-trigger`

So in your case you want to use the same pattern. In your equivalent method to `headerKeyboardTrigger` you want to toggle expanded. Does that make sense?

hectorbus: Yeah! It makes a lot of sense. I didn't know about the method name made automatically, that helps a lot! Thank you!

app/assets/javascripts/discourse/widgets/topic-admin-menu.js.es6
```diff
@@ -52,7 +52,10 @@ createWidget("topic-admin-menu-button", {
         this.attach("button", {
           className:
             "btn-default toggle-admin-menu" +
-            (attrs.fixed ? " show-topic-admin" : ""),
+            (attrs.fixed ? " show-topic-admin" : "") +
+            (attrs.triggerKeyboardShortcut
```
eviltrout: This is a bit of a nit pick, but could we call this `addKeyboardTargetClass` and then call it `keyboard-target-admin-menu` or something similar? I think it's more clear that the attribute means this is a target for keyboard operations rather than a trigger for them.

hectorbus: Sure, I completely agree with you.

eviltrout:
Thanks very much @hectorbus, I've merged it now.

hectorbus:
> Thanks very much @hectorbus, I've merged it now.

You're welcome! Any idea why the Core QUnit tests failed on CORE-FRONTEND?

eviltrout:
They seemed completely unrelated. I think it was a random failure due to date/times.

PR #8569, opened 2019-12-17
DEV: Remove unsafe-eval from development CSP

davidtaylorhq:
- Remove `eval` when loading JS modules in development
- Always compile HBS templates on the server
- Load the HBS compiler only for qunit routes
- Extend the CSP middleware to allow unsafe-eval only on qunit routes


SamSaffron:
I would love to see this merged, it should make debugging in firefox a bit easier. It struggles a bit with eval.

PR #8571, opened 2019-12-17
DEV: allows this.get in widgets

jjaffeux:


eviltrout:
This is good, but could we add a test to make sure it's there?

jjaffeux:
@eviltrout sure thing!

done ✅

PR #8578, opened 2019-12-18
FEATURE: Make 'Reorder Categories' work with nested categories

udan11:


app/assets/javascripts/discourse/controllers/reorder-categories.js.es6
```diff
@@ -31,123 +26,113 @@ export default Controller.extend(ModalFunctionality, Ember.Evented, {
...
+    this.reorder();
   },
 
   actions: {
-    change(cat, e) {
-      let position = parseInt($(e.target).val(), 10);
-      let amount = Math.min(
-        Math.max(position, 0),
+    change(category, event) {
+      let newPosition = parseInt($(event.target).val(), 10);
```
ZogStriP: Is there a way we could do this without using jQuery?

test/javascripts/controllers/reorder-categories-test.js.es6
```diff
@@ -30,7 +30,7 @@ QUnit.test("fixIndices set unique position number", function(assert) {
 });
 
 QUnit.test(
-  "fixIndices places subcategories after their parent categories, while maintaining the relative order",
+  "reorder places subcategories after their parent categories, while maintaining the relative order",
```
ZogStriP: We probably want to add another test that ensure sub-sub-categories work properly?

udan11: Amended the current test to include a sub-sub-category.

app/assets/javascripts/discourse/controllers/reorder-categories.js.es6
udan11: Looks like a `DOMElement` is already passed. I do not know the reason why jQuery was used, probably just to make the test work.

PR #8579, opened 2019-12-18
FIX: Show parent and subcategories for 2nd level categories

udan11:


ZogStriP:
Since this is a UX change, can you add some before/after screenshots?

Also, can you add a commit message that explains why this change is needed? (_future-you thanks you_)

udan11:
Before:

Screen Shot 2019-12-18 at 17 03 55

After (with wider parent category chooser):

Screen Shot 2019-12-18 at 17 04 15

ZogStriP:
Have you tested on mobile?

udan11:
Mobile was left untouched because I could not widen the category chooser (not too much space there).

PR #8580, opened 2019-12-18
DEV: Add smoke test to github actions

davidtaylorhq:


.github/workflows/smoke.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Bundle
+        run: unbuffer d/bundle install
+
+      - name: Migrate
+        run: unbuffer d/rake db:migrate
+
+      - name: Create script
+        run: | 
+          touch tmp/script.rb
+          echo "User.create!(username:'smoke_user', password:'P4ssw0rd', admin:true)" >> tmp/script.rb
```
ZogStriP: Why not directly call the rails console?

```suggestion
          echo "User.create!(username:'smoke_user', password:'P4ssw0rd', admin:true)" | d/rails c
```

davidtaylorhq: The `d/rails` wrapper doesn't hook up stdin 😢

ZogStriP:
Any chances getting this to work reliably?

PR #8589, opened 2019-12-18
DEV: Remove buffered rendering from topic-list-item

oblakeerickson:
This is another refactoring in the multi-step process to remove all uses
of our custom Render Buffer.

Previous commit: 1c7305c0f1b26444dbd12918efb50e55a1231e73 in this
series.

This is just a refactor and should not change any functionality.

### DIFF
--- a/app/assets/javascripts/discourse/components/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-list-item.js.es6
@@ -37,7 +37,7 @@ export default Component.extend({
   attributeBindings: ["data-topic-id"],
   "data-topic-id": alias("topic.id"),
 
-  init() {
+  didReceiveAttrs() {
     this._super(...arguments);
     this.renderTopicListItem();
   },
@@ -75,13 +75,8 @@ export default Component.extend({
     }
   },
 
-  @observes("bulkSelectEnabled", "topic.{pinned,tags}")
+  @observes("topic.pinned")
   rerenderTriggers() {
-    this.renderTopicListItem();
-  },
-
-  @discourseComputed("topic.{archived,visible,closed,category}")
-  rerenderComputedTriggers() {
     this.renderTopicListItem();
   },
 
--- a/app/assets/javascripts/discourse/components/topic-list-item.js.es6
+++ b/app/assets/javascripts/discourse/components/topic-list-item.js.es6
@@ -1,8 +1,7 @@
-import discourseComputed from "discourse-common/utils/decorators";
+import discourseComputed, { observes } from "discourse-common/utils/decorators";
 import { alias } from "@ember/object/computed";
 import Component from "@ember/component";
 import DiscourseURL from "discourse/lib/url";
-import { bufferedRender } from "discourse-common/lib/buffered-render";
 import { findRawTemplate } from "discourse/lib/raw-templates";
 import { wantsNewWindow } from "discourse/lib/intercept-click";
 import { on } from "@ember/object/evented";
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -64,6 +75,19 @@ export const ListItemDefaults = {
     }
   },
 
+  @observes(
```
jjaffeux: I would REALLY prefer if we didn't have to use observers.

If you have to,  I think you can do:
@observes("bulkSelectEnabled", "topic.{pinned,visible,archived,closed,category,tags}")
```

Comment (oblakeerickson) [reply app/assets/javascripts/discourse/components/topic-list-item.js.es6]: Okay updated. How about a compromise? I switched to using computed properties where possible. 

These were using observers already with the buffer render,
```
rerenderTriggers: ["bulkSelectEnabled", "topic.pinned"],
```

so the only one I'm really adding is for the tags.

**Comment by oblakeerickson** [comment]
Also looks like this change breaks the discourse-assign plugin

https://github.com/discourse/discourse-assign/blob/master/assets/javascripts/discourse/components/assigned-topic-list-item.js.es6

I'll add a pr their with a fix.

**Comment by oblakeerickson** [comment]
Here is the PR for the discourse-assign plugin:

https://github.com/discourse/discourse-assign/pull/59

File: app/assets/javascripts/discourse/components/topic-list-item.js.es6 [hunk 2]
```diff
@@ -64,6 +75,16 @@ export const ListItemDefaults = {
     }
   },
 
+  @observes("bulkSelectEnabled", "topic.{pinned,tags}")
+  rerenderTriggers() {
+    this.renderTopicListItem();
+  },
+
+  @discourseComputed("topic.{archived,visible,closed,category}")
```
jjaffeux: hmmm this weird

computedProperties are supposed to a value, here you return nothing, so nothing ever get cached, so i'm not even sure what it does 😅

oblakeerickson: It has to do with the bulk selected items and the actions you take on them. If you bulk select a bunch of topics, then use the wrench to close them all, this will trigger a re-render of the raw template and show the closed icon next to each topic. Without this the icons never get added unless you hard-refresh the page.

image

jjaffeux: I'm curious if only having:

```
didReceiveAttrs() {
  this._super(...arguments);
  this.renderTopicListItem();
}
```

Work? (You can probably remove init with this too.)

oblakeerickson: Okay updated it to use didReceiveAttrs which is working great! Thanks!

I'm not sure why I didn't use it this time cause that was the pattern I've been using for all the other updates I've done.

The only thing which still needs an observer now is just for pinning a topic, because you can actually click the pin and toggle it. None of the other icons have this functionality.

discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/dev-remove-buffered-rendering-from-topic-list-item-pr-8589/7893/11):

One thing we could do is refactor pinning topic to trigger an app event we could consume here. Would be cleaner IMO. But we wan do this later.

app/assets/javascripts/discourse/components/topic-list-item.js.es6
```diff
@@ -64,6 +75,11 @@ export const ListItemDefaults = {
     }
   },
 
+  @observes("topic.pinned")
```
jjaffeux: is this sill needed?

jjaffeux: Also if you can't remove it, if it's only doing this, you can directly put it on top of renderTopicListItem, it's not because a function is decorated that you can't call it.

oblakeerickson: yep it is still needed

PR #8590, opened 2019-12-19
FIX: correct user serializer user method for extended serializer

featheredtoast:
A small fix for Basic User Serializers where some downstream serializers do not correctly set user objects. This causes some issues in certain plugins that depend on the user method to return a user.

Submitting as a PR so I don't break anything in an end-of-day push-to-master rush. I will merge this tomorrow if this looks OK.

### DIFF
--- a/spec/serializers/basic_user_serializer_spec.rb
+++ b/spec/serializers/basic_user_serializer_spec.rb
@@ -15,8 +15,7 @@ describe BasicUserSerializer do
     end
 
     describe 'extended serializers' do
-      Fabricator(:post_action)
-      let(:post_action) { Fabricate.build(:post_action, user: user) }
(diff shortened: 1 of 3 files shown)


spec/serializers/basic_user_serializer_spec.rb
```diff
@@ -14,6 +14,16 @@
       expect(json[:avatar_template]).to eq(user.avatar_template)
     end
 
+    describe 'extended serializers' do
+      Fabricator(:post_action) do
```
jjaffeux: is this empty block needed ?

featheredtoast: No, it's not necessary. I'll clean that up!

PR #8592, opened 2019-12-19
UX: Show grandchildren in categories list

udan11:
It looks like this:

Screen Shot 2019-12-19 at 11 51 04

### DIFF
--- a/app/assets/javascripts/discourse/models/category.js.es6
+++ b/app/assets/javascripts/discourse/models/category.js.es6
@@ -65,6 +65,16 @@ const Category = RestModel.extend({
     return (parentLevel || -1) + 1;
   },
 
+  @discourseComputed("subcategories")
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/models/category.js.es6
```diff
@@ -65,6 +65,16 @@ const Category = RestModel.extend({
     return (parentLevel || -1) + 1;
   },
 
+  @discourseComputed("subcategories")
+  isGrandParent(subcategories) {
+    return (
+      subcategories &&
+      subcategories.some(
```
ZogStriP: Does `some` work in IE11?

udan11: [Yes.](https://caniuse.com/#search=some)

PR #8594, opened 2019-12-19
DEV: Run linting just for core

udan11:
`CORE-LINT` and `PLUGIN-LINT` were doing the same thing.


davidtaylorhq:
Official plugins are only cloned when `env.TARGET == 'PLUGINS'`, so that's why the lint job is set to run in both environments.

PR #8598, opened 2019-12-19
FEATURE: Allow complex post params from plugin

markvanlan:
This allows hashes and arrays to be permitted into the create post params

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -685,12 +685,12 @@ class PostsController < ApplicationController
     ]
 
     Post.plugin_permitted_create_params.each do |key, value|
-      permission =  case value[:type].class.name
-                    when "String"
+      permission =  case value[:type]
+                    when :string
                       key.to_sym
-                    when "Array"
+                    when :array
                       { key => [] }
-                    when "Hash"
+                    when :hash
                       { key => {} }
       end
       permitted << permission if value[:plugin].enabled?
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -684,8 +684,16 @@ def create_params
       :draft_key
     ]
 
-    Post.plugin_permitted_create_params.each do |key, plugin|
-      permitted << key if plugin.enabled?
+    Post.plugin_permitted_create_params.each do |key, value|
```
ZogStriP: Could you add some tests to ensure this API does not regress?

markvanlan: @eviltrout Tests have been added

lib/plugin/instance.rb
```diff
@@ -250,9 +250,9 @@ def add_preloaded_topic_list_custom_field(field)
   end
 
   # Add a permitted_create_param to Post, respecting if the plugin is enabled
-  def add_permitted_post_create_param(name)
+  def add_permitted_post_create_param(name, type = "")
```
eviltrout: This API is a bit odd to me. You pass an object of the type you want?

I would prefer an enum of something or a symbol.

markvanlan: @eviltrout I made the change to use symbols. The internals look a lot better, but the api is less consistent now with how strong params are handled in controllers.

PR #8599, opened 2019-12-20
UX: Update time-based strings to be SI-compliant

getsnoopy:
The units of second, minute, hour, and day are all SI units or are
associated with the SI, and therefore have symbols and not
abbreviations. The month is not an SI unit nor associated with it,
but is almost universally abbreviated as "mo". Finally, all units
should have a (non-breaking) space between the quantity and the
symbol/abbreviation. This commit addresses all these issues.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -71,37 +71,37 @@ en:
       wrap_ago: "%{date} ago"
 
       tiny:
-        half_a_minute: "< 1m"
+        half_a_minute: "< 1&nbsp;min"
         less_than_x_seconds:
-          one: "< %{count}s"
-          other: "< %{count}s"
+          one: "< %{count}&nbsp;s"
+          other: "< %{count}&nbsp;s"
         x_seconds:
-          one: "%{count}s"
-          other: "%{count}s"
+          one: "%{count}&nbsp;s"
+          other: "%{count}&nbsp;s"
         less_than_x_minutes:
-          one: "< %{count}m"
-          other: "< %{count}m"
+          one: "< %{count}&nbsp;min"
+          other: "< %{count}&nbsp;min"
         x_minutes:
-          one: "%{count}m"
-          other: "%{count}m"
+          one: "%{count}&nbsp;min"
+          other: "%{count}&nbsp;min"
         about_x_hours:
-          one: "%{count}h"
-          other: "%{count}h"
+          one: "%{count}&nbsp;h"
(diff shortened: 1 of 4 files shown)


ZogStriP:
Thanks but why the change? I've never heard any report of people having trouble understand what "1s" or "2m" means.

getsnoopy:
Well I just saw the errors and wanted to correct them. For "2m" more specifically, it means 2 metres unless a context is provided, and even in a "time" context, it could be confused with 2 months.

ZogStriP:
Thanks but Discourse is a discussion software, not a science software. I don't think we need to be SI compliant here. Plus, if people had trouble figuring out what these acronyms meant, we would have heard complaints by now.

Discourse is 100% customizable, so feel free to apply these changes to your Discourse 😉

getsnoopy:
The purpose of SI is for all uses, not just science. Using seconds, minutes, and hours automatically means the usage of SI units; there are no "SI seconds, minutes, and hours" and "generic seconds, minutes , and hours".

Also, I'm confused. If the change makes it clearer and more correct regardless of complaints, why reject it? Being compliant with SI doesn't hurt anyone, but helps many by making these units consistent with the units they use daily.

getsnoopy:
I'm still waiting on a reasonable response to my question about why perpetuating this error is preferable to fixing it beforehand.

eviltrout:
> Also, I'm confused. If the change makes it clearer and more correct regardless of complaints, why reject it?

We literally have millions of users and have iterated these abbreviations over years of development, so with all due respect what is clearer to you is not necessarily the same as what is clear to a wide audience.

Discourse also makes customizing text easy - you can do it in your admin section or via a plugin, so we don't see the point of changing this.

getsnoopy:
> We literally have millions of users and have iterated these abbreviations over years of development, so with all due respect what is clearer to you is not necessarily the same as what is clear to a wide audience.

It's not what I think is clearer, it's an international standard. And with all due respect, there are _billions_ of people who've come together from practically every country in the world to define standards which the BIPM has released, namely the SI. There is a reason every SI symbol is unique: it's because it's unambiguous and consistent. The point of an international standard is that everyone uses it.

I don't understand what the reason for resistance here is. It's an error to refer to minutes using the "m" symbol, which means metres. Fixing this hurts no one, but helps everyone.

eviltrout:
@getsnoopy we've given you viable workarounds. If you still disagree I suggest finding other software that suits your needs.

PR #8600, opened 2019-12-20
FIX: Update topic/post counter correctly when category has zero topics

davidtaylorhq:
Previously, categories without any topics were being excluded from the UPDATE query. This means the counter gets stuck, and the category cannot be deleted. This change ensures that the counters get correctly set to zero.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -188,11 +188,16 @@ class Category < ActiveRecord::Base
 
     DB.exec <<~SQL
       UPDATE categories c
-         SET topic_count = x.topic_count,
-             post_count = x.post_count
-        FROM (#{topics_with_post_count}) x
+         SET topic_count = COALESCE(x.topic_count, 0),
+             post_count = COALESCE(x.post_count, 0)
+        FROM (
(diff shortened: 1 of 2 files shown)


davidtaylorhq:
This does work... but having read it again I think we'll be better off splitting this into multiple queries. This one huge query is very difficult to understand.

 I will take another look.

davidtaylorhq:
Ok, much much simpler now. Ready for review

davidtaylorhq:
I checked the performance locally, and it was identical. Of course, with a bigger database the story is slightly different. With my 'simpler' method, it takes about 40% longer on a massive database. I've gone back to the 'hard to read, but working' single query, and there is no noticeable performance impact. I guess here we have to sacrifice readability for performance.

On meta: (average over 5 runs)
  - Original (broken) method: 7.5s
  - New 'simpler' method: 9s
  - 'Complicated' query method: 7.5s

On a larger site with 500k topics (1 run):
  - Old method: 83.1s
  - New 'simpler' method: 116.7s
  - 'Complicated' query method: 77.8

PR #8605, opened 2019-12-20
UX: Do not use avatars as fallback opengraph images for replies

davidtaylorhq:
People rarely want to have their avatars show up as the preview image on social media platforms. Instead, we should fall back to the site opengraph image.

### DIFF
--- a/lib/topic_view.rb
+++ b/lib/topic_view.rb
@@ -236,11 +236,8 @@ class TopicView
   end
 
   def image_url
-    if @post_number > 1 && @desired_post.present?
-      @desired_post.image_url
-    else
-      @topic.image_url
-    end
+    url = @desired_post&.image_url if @post_number > 1
+    url || @topic.image_url
   end
 
   def filter_posts(opts = {})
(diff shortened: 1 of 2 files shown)


lib/topic_view.rb
```diff
@@ -237,12 +237,7 @@ def published_time
 
   def image_url
     if @post_number > 1 && @desired_post.present?
-      if @desired_post.image_url.present?
-        @desired_post.image_url
-      elsif @desired_post.user
-        # show poster avatar
-        @desired_post.user.avatar_template_url.gsub("{size}", "200")
-      end
+      @desired_post.image_url
```
ZogStriP: I think we should be more clever here and try to always get an image.

```ruby
def image_url
  url = @desired_post&.image_url if @post_number > 1
  url || @topic.image_url
end
```

ZogStriP: Also, I think this is worth testing.

davidtaylorhq: :+1: agreed. Updated the logic and added some specs

PR #8606, opened 2019-12-20
Rename 'target usernames' with 'target recipients' in Composer

nbianca:
This reverts commit 9fe11d0fc33850d46d4e1261ba6f66f187652613 which
reverted ebb288dc2cec966fd26ee46f10e1e4d8a596ba5a.


ZogStriP:
Can you also add a test to make sure we can reply and/or whisper in a message?

ZogStriP:
Where are we on this?

nbianca:
I rebased this PR and added the test for reply to PM (the use case that broke before).

I'll wait for CI checks to pass and will merge it. :blush:

PR #8607, opened 2019-12-20
UX: Use '-' as default category slug

udan11:
This replaces the default slug from 'ID-category' to '-'.

### DIFF
--- a/test/javascripts/models/category-test.js.es6
+++ b/test/javascripts/models/category-test.js.es6
@@ -17,13 +17,13 @@ QUnit.test("slugFor", assert => {
   );
   slugFor(
     store.createRecord("category", { id: 123, slug: "" }),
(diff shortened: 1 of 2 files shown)


eviltrout:
Can you give a few more examples here? Also should there be a test?

ZogStriP:
Where are we on this?

test/javascripts/models/category-test.js.es6
```diff
@@ -17,12 +17,12 @@ QUnit.test("slugFor", assert => {
   );
   slugFor(
     store.createRecord("category", { id: 123, slug: "" }),
-    "123-category",
+    "-",
     "It returns id-category for empty strings"
```
ZogStriP: You should also update the tests description ;)

PR #8609, opened 2019-12-20
FEATURE: add mybb.ru import script

AlexP11223:


eviltrout:
Thanks! We're always happy to add more importers.

PR #8610, opened 2019-12-20
PERF: Cache user badge count in user_stats table

davidtaylorhq:
This means that we no longer need to run a `count()` on the user_badges table when serializing a user


SamSaffron:
cool, I support this change, feel free to merge!

PR #8612, opened 2019-12-22
FIX: category routes model params should decode their URL parts

featheredtoast:
Ember's route star globbing does not uri decode by default. This is
problematic for subcategory globs with encoded URL site settings enabled.

Subcategories with encoded URLs will 404 without this decode.

I found this https://github.com/tildeio/route-recognizer/pull/91
which explicitly explains that globbing does not decode automatically.

### DIFF
--- a/app/assets/javascripts/discourse/routes/build-category-route.js.es6
+++ b/app/assets/javascripts/discourse/routes/build-category-route.js.es6
@@ -39,6 +39,12 @@ export default (filterArg, params) => {
             .filter(x => x)
             .join("/");
         }
+      } else if (Discourse.SiteSettings.slug_generation_method === "encoded") {
+        let separator = "/";
+        modelParams.category_slug_path_with_id = modelParams.category_slug_path_with_id
+          .split(separator)
+          .map(urlPart => decodeURI(urlPart))
+          .join(separator);
       }
 
       return modelParams;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/routes/build-category-route.js.es6
```diff
@@ -39,6 +39,12 @@ export default (filterArg, params) => {
             .filter(x => x)
             .join("/");
         }
+      } else if (Discourse.SiteSettings.slug_generation_method === "encoded") {
+        let separator = "/";
```
udan11: I do not think it is required to define a variable for the separator. We hardcode `"/"` in so many places already.

```suggestion
```

app/assets/javascripts/discourse/routes/build-category-route.js.es6
```diff
@@ -39,6 +39,12 @@ export default (filterArg, params) => {
...
+        modelParams.category_slug_path_with_id = modelParams.category_slug_path_with_id
+          .split(separator)
```
udan11:
```suggestion
          .split("/")
```

app/assets/javascripts/discourse/routes/build-category-route.js.es6
```diff
@@ -39,6 +39,12 @@ export default (filterArg, params) => {
...
+          .map(urlPart => decodeURI(urlPart))
+          .join(separator);
```
udan11:
```suggestion
          .join("/");
```

featheredtoast:
per Daniel, I'm instead implementing this in findBySlugPathWithID; slash and splitting is already handled in there :)

PR #8619, opened 2019-12-24
DEV: Improve support for extending flags.

romanrizzi:
- Ensure that the 'notify_moderators' flag is always the last flag when using custom flags.
- Support passing a custom FlagSettings object when replacing flags to reuse existing ones.

### DIFF
--- a/app/serializers/site_serializer.rb
+++ b/app/serializers/site_serializer.rb
@@ -54,17 +54,16 @@ class SiteSerializer < ApplicationSerializer
 
   def post_action_types
     cache_fragment("post_action_types_#{I18n.locale}") do
-      types = PostActionType.types.values.map { |id| PostActionType.new(id: id) }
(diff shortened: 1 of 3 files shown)


app/serializers/site_serializer.rb
```diff
@@ -163,4 +162,16 @@ def include_shared_drafts_category_id?
     scope.can_create_shared_draft?
   end
 
+  private
+
+  def ordered_flags(flags)
+    notify_moderators_type = 7
```
eviltrout: Isn't this constant available elsewhere?

romanrizzi: It's not. I'll make it a constant 👍

romanrizzi: I realized I should use `PostActionType.flag_types[:notify_moderators]` instead.

PR #8620, opened 2019-12-27
FEATURE: Topic admin menu sticks to bottom on mobile.

lis2:
Our current topic admin menu is not always fully visible on a mobile device, therefore some options are difficult to click.

To solve this issue, we can display the admin menu on the bottom of the screen on mobile devices.

image


eviltrout:
This looks good to me. What do you think @awesomerobot ?

PR #8624, opened 2019-12-29
FIX: makes highlighting last viewed topic more resilient

jjaffeux:
`highlight` was called from `didInsertElement` which technically doesn't ensure the list is rendered. By wrapping the highlighting code in `afterRender` we ensure it works more consistently.


SamSaffron:
I see, this looks safe to me, feel free to merge! :+1:

PR #8629, opened 2019-12-30
PERF: Reduce DB queries when serializing ignore/mute information

davidtaylorhq:
This avoids needing two database queries every time another user is serialized


davidtaylorhq:
Thanks for reviewing @eviltrout. The data will be refreshed when `.reload` is called on the activerecord object, and we use this pattern already for a few other variables in `user.rb`. I think it should be safe 🤞

PR #8639, opened 2019-12-31
FIX: Add check for blank params in Webauthn::SecurityKeyBaseValidationService

martin-brennan:
Added a fix to gracefully error with a Webauthn::SecurityKeyError if somehow a user provides blank params when validating a security key, e.g. on login. From this error in the dev logs:

```
Message (2 copies reported)
NoMethodError (undefined method `hash_function' for nil:NilClass)
/var/www/discourse/lib/webauthn/security_key_authentication_service.rb:67:in `authenticate_security_key'

Backtrace
/var/www/discourse/lib/webauthn/security_key_authentication_service.rb:67:in `authenticate_security_key'
/var/www/discourse/app/controllers/session_controller.rb:303:in `create'
```

### DIFF
--- a/lib/webauthn.rb
+++ b/lib/webauthn.rb
@@ -14,6 +14,7 @@ module Webauthn
 
   class SecurityKeyError < StandardError; end
 
+  class InvalidArgumentError < SecurityKeyError; end
   class InvalidOriginError < SecurityKeyError; end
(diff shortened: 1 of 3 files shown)


eviltrout:
Oops I missed that this was a draft before I approved it! Sorry :)

PR #8643, opened 2019-12-31
FIX: hide topic admin menu button for regular users

lis2:
This bug was introduced here https://github.com/discourse/discourse/commit/b725252cd0c60d17521499d70f021cfde2114ccf#diff-a75b5e5a8561229e772a6c8e14804786R49

Wrench button is displayed on mobile for all users, not just admins with no actions available

image


lis2:
I am closing that PR because @davidtaylorhq made same fix and it is already in master

PR #8649, opened 2020-01-01
FIX: Catch error when unknown COSE algorithm is supplied for Security Key

martin-brennan:
Added a fix to gracefully error with a Webauthn::SecurityKeyError if somehow a user provides an unkown COSE algorithm when logging in with a security key. From this error in the dev logs:

```
Message (2 copies reported)
NoMethodError (undefined method `hash_function' for nil:NilClass)
/var/www/discourse/lib/webauthn/security_key_authentication_service.rb:67:in `authenticate_security_key'

Backtrace
/var/www/discourse/lib/webauthn/security_key_authentication_service.rb:67:in `authenticate_security_key'
/var/www/discourse/app/controllers/session_controller.rb:303:in `create'
```

If `COSE::Algorithm.find` returns nil we now fail gracefully and log the algorithm used along with the user ID and the security key params for debugging, as this will help us find other common algorithms to implement for webauthn

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -943,6 +943,7 @@ en:
       public_key_error: "The public key verification for the credential failed."
       ownership_error: "The security key is not owned by the user."
       not_found_error: "A security key with the provided credential ID could not be found."
+      unknown_cose_algorithm_error: "The algorithm used for the security key is not recognized."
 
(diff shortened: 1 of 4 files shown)


SamSaffron:
This looks good to me ... feel free to merge.

PR #8653, opened 2020-01-02
DEV: Don't run the PLUGINS-LINT stage

CvX:
It's unlikely that changes introduced in a PR would trigger lint errors in non-default official plugins.
What might happen though (and did at least once) is that an unrelated lint error in a plugin can fail the master build and all PR builds.

@davidtaylorhq, @jjaffeux?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,14 +10,14 @@ on:
     
 jobs:
   build:
-    name: "${{ matrix.target }}-${{ matrix.build_types }}"
+    name: "${{ matrix.target }}-${{ matrix.build_type }}"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
     env:
       DISCOURSE_HOSTNAME: www.example.com
       RUBY_GLOBAL_METHOD_CACHE_SIZE: 131072
-      BUILD_TYPE: ${{ matrix.build_types }}
+      BUILD_TYPE: ${{ matrix.build_type }}
       TARGET: ${{ matrix.target }}
       RAILS_ENV: test
       PGHOST: localhost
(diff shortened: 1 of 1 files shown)


davidtaylorhq:
Yeah, I agree with idea. However, it gets complicated because of our internal build. @jjaffeux and I designed the GitHub actions to match the internal build as closely as possible. If you have a green tick on GitHub, it should pass the internal build and get added to tests-passed. If you have a red cross on GitHub, it will fail the internal build, and never make it to tests-passed.

Now that we have so many plugins, it might be worth reconsidering this, but I think we might have to make the same change on the internal build system as well 🤔

CvX:
Yup, I’m for changing this in the internal build  too.

Our linting rules live outside the core, so it would make sense to do the plugin linting only in plugin repo builds (and maybe in lint configuration repos? Though that may be overcomplicating things)

davidtaylorhq:
> Our linting rules live outside the core

Almost... although I think rubocop rules still live in core? But anyway, I agree it's weird to lint plugins in the core build.

At the moment, the core build failing is the only thing which actually forces us to keep plugins linted properly. If we remove it, then we need some other warning system. Maybe worth opening a dev topic on it?

CvX:
> I think rubocop rules still live in core

You're right. I'm closing the PR for now. Before merging this, we'd need to:
1. Extract rubocop rules from core into https://github.com/discourse/rubocop-discourse
2. Update plugins to use the the new URL
3. Add a Github Actions CI workflow to every plugin, that will run frontend and backend tests, and do the linting
4. Maybe also add a workflow to rubocop-discourse and https://github.com/discourse/eslint-config-discourse that will do a lint pass (after `bin/rake plugin:install_all_official`)
5. Remove the plugins-lint stage from both the internal build

I'll play with 3./4. and post the updated plan for a review on dev

PR #8662, opened 2020-01-05
A minor copyedit

rizka10:
High search priority cannot be smaller than 1 but can be higher.


SamSaffron:
Change looks good to me! thanks!

PR #8664, opened 2020-01-06
FEATURE: Secure media allowing duplicated uploads with category-level privacy and post-based access rules

martin-brennan:
### General Changes and Duplication

* We now consider a post `with_secure_media?` if it is in a read-restricted category.
* When uploading we now set an upload's secure status straight away, using the following flow:

SecureMediaUploadController

* When uploading if `SiteSetting.secure_media` is enabled, we do not check to see if the upload already exists using the `sha1` digest of the upload. The `sha1` column of the upload is filled with a `SecureRandom.hex(20)` value which is the same length as `Upload::SHA1_LENGTH`. The `original_sha1` column is filled with the _real_ sha1 digest of the file. 
* Whether an upload `should_be_secure?` is now determined by whether the `access_control_post` is `with_secure_media?` (if there is no access control post then we leave the secure status as is).
* When serializing the upload, we now cook the URL if the upload is secure. This is so it shows up correctly in the composer preview, because we set secure status on upload.

### Viewing Secure Media

* The secure-media-upload URL will take the post that the upload is attached to into account via `Guardian.can_see?` for access permissions
* If there is no `access_control_post` then we just deliver the media. This should be a rare occurrance and shouldn't cause issues as the `access_control_post` is set when `link_post_uploads` is called via `CookedPostProcessor`

### Removed

We no longer do any of these because we do not reuse uploads by sha1 if secure media is enabled.

* We no longer have a way to prevent cross-posting of a secure upload from a private context to a public context.
* We no longer have to set `secure: false` for uploads when uploading for a theme component.

### Missing

* Functionality for moving an upload between contexts is still not supported. For example, if splitting one or more posts out of a PM and placing them into a public topic, then we need to loop through the uploads associated with those posts and set them to `secure: false` and update their ACLs.


app/models/upload.rb
```diff
@@ -392,30 +399,38 @@ def short_url_basename
...
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#  sha1                   :string(40)
+#  origin                 :string(1000)
+#  retain_hours           :integer
+#  extension              :string(10)
+#  thumbnail_width        :integer
+#  thumbnail_height       :integer
+#  etag                   :string
+#  secure                 :boolean          default(FALSE), not null
```
SamSaffron: do we need the "secure" column give we have "access_control_post_id" ?

martin-brennan: Yes, because we don't know the `access_control_post_id` when the upload is first created but we still want to know it was created in a secure context, as the ACL is now set at upload time (and the URL for /secure-media-uploads/ is used straight away for posts).

SamSaffron: aha yeah ... makes sense

spec/lib/upload_creator_spec.rb
```diff
@@ -246,6 +236,105 @@
...
+            upload = UploadCreator.new(file_from_fixtures(filename), filename, type: "avatar").create_for(user.id)
+            expect(upload.secure).to eq(false)
+            upload.destroy!
+          end
+        end
+
+        context "if type of upload is in the composer" do
+          let(:opts) { { type: "composer" } }
+          it "sets the upload to secure, because we don't know the context of the composer" do
+            expect(result.secure).to eq(true)
```
SamSaffron: we should check for sha and original sha here as well no? To ensure we are generating a new one?

app/controllers/uploads_controller.rb
```diff
@@ -139,6 +141,14 @@ def show_secure
     redirect_to upload.secure? ? signed_secure_url : Discourse.store.cdn_url(upload.url)
   end
 
+  def handle_secure_upload_request(upload, path_with_ext)
+    if upload.access_control_post_id.present?
+      raise Discourse::InvalidAccess if !guardian.can_see?(upload.access_control_post)
```
SamSaffron: is `guardian.can_see?(nil)` going to return true - eg post was trashed? perhaps if there is no post we can unconditionally raise an invalid access?

SamSaffron:
perhaps we add an extra where here: 

https://github.com/discourse/discourse/blob/710eafdd35d422eb5732b39fad915c2dbfb7ce89/app/jobs/scheduled/clean_up_uploads.rb#L70-L70

then we can give secure uploads a bit of extra protection? atm revising a post with a an upload not to have the upload link will result in removal, but with secure uploads we have a far stronger link so it is trivial to protect history. 

In the context of this change it may make sense to add this extra protection. 

Given we already have all the refs in place maybe we add `where('uploads.access_control_post_id IS NULL')` 

On the fence here but it seems like it is reasonable if you are being totally paranoid about uploads to just keep them around in revisions.

app/controllers/uploads_controller.rb
martin-brennan: If we unconditionally raise invalid access then we will break previews in the composer, because at time of upload creation we don't know the access control post.

Maybe if the upload was created > 1h ago with no post attached or something?

martin-brennan:
@SamSaffron with the cleanup job comments above, if a post no longer contains its upload the upload will still have the `access_control_post_id`, which is fine. It probably is a good idea to add the where clause, because otherwise if we delete the upload any posts that have a direct link to the /secure-media-uploads/ URL for the upload will get a 404 error.

Maybe at some point we could have a secure upload deleter...that deletes the upload and finds all posts that mention it and replace the URL/lightbox with "(secure upload has been removed)" or something.

lib/file_store/base_store.rb
```diff
@@ -9,7 +9,7 @@ def store_upload(file, upload, content_type = nil)
       store_file(file, path)
     end
 
-    def store_optimized_image(file, optimized_image)
+    def store_optimized_image(file, optimized_image, content_type = nil, secure: false)
```
pmusaraj: Minor thing: do we have to define `content_type` and `secure` here? They aren't used when storing locally.

app/models/optimized_image.rb
```diff
@@ -106,7 +106,7 @@ def self.create_for(upload, width, height, opts = {})
 
           # store the optimized image and update its url
           File.open(temp_path) do |file|
-            url = Discourse.store.store_optimized_image(file, thumbnail)
+            url = Discourse.store.store_optimized_image(file, thumbnail, nil, secure: upload.secure?)
```
pmusaraj: Related to the comment in `base_store.rb`, I don't think we need to pass `nil` here for content type. Simply setting it to `nil` when using the S3 store should suffice.

lib/upload_creator.rb
```diff
@@ -117,12 +127,16 @@ def create_for(user_id)
       @upload.for_export          = true if @opts[:for_export]
       @upload.for_site_setting    = true if @opts[:for_site_setting]
       @upload.for_gravatar        = true if @opts[:for_gravatar]
+      uploading_in_public_context = @upload.for_theme || @upload.for_site_setting || @opts[:type] == "avatar"
+
+      # determine whether upload should be secure when first created
```
pmusaraj: Right now there are two places where we determine whether an upload should be marked secure: here and in `models/upload.rb#should_be_secure?`. I'm wondering whether it would make sense to abstract the logic into one central place. There are some differences (due to the upload existing in one case and not the other) but there is a fair amount of overlap, too.

app/jobs/scheduled/clean_up_uploads.rb
```diff
@@ -78,6 +78,7 @@ def execute(args)
         .where("tf.upload_id IS NULL")
         .where("ue.upload_id IS NULL")
         .where("ss.value IS NULL")
+        .where("uploads.access_control_post_id IS NULL")
```
ZogStriP: I think this `where` should go up, next to the other `where` on the `uploads` table (right before the first `join`).

lib/file_store/base_store.rb
martin-brennan: The problem is @pmusaraj both the methods had a different signature, but the S3 has the additional secure + content_type parameters. In `OptimizedImage` I had:

```
url = Discourse.store.store_optimized_image(file, thumbnail, nil, secure: upload.secure?)
```

But this would fail all the local specs. I could add a conditional here e.g. if its S3 store use the extra params, but I thought it would be better for the method signature to be the same if we are overriding it. I guess the other alternative is just adding `opts` to the base_store method and using those where needed.

lib/upload_creator.rb
martin-brennan: @pmusaraj very good point, I have moved this stuff out into a new `UploadSecurity` class :+1:

lib/file_store/base_store.rb
pmusaraj: That makes sense, thanks!

PR #8665, opened 2020-01-06
FEATURE: Turn csp on by default

oblakeerickson:


oblakeerickson:
image

There should be output here while the tests are running. I'm pretty sure the qunit tests don't even start.

PR #8668, opened 2020-01-06
DEV: Allow CSP to be enabled during QUnit tests

davidtaylorhq:
The QUnit rake task starts a server in test mode. Therefore we need some tweaks to allow dynamic hostnames and dynamic CSP hostnames in test mode. These tweaks are already present in development mode.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -224,8 +224,9 @@ module Discourse
     # supports etags (post 1.7)
     config.middleware.delete Rack::ETag
 
-    if Rails.env.production?
+    if !(Rails.env.development? || ENV['SKIP_ENFORCE_HOSTNAME'] == "1")
       require 'middleware/enforce_hostname'
+      puts "enforcing hostname"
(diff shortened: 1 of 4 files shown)


SamSaffron:
some weird fails:

```
Users::OmniauthCallbacksController Google Oauth2 when user has been verified with full screen login doesn't attempt redirect to external origin
     Failure/Error: expect(response.location).to eq "http://test.localhost/"

       expected: "http://test.localhost/"
            got: "http://www.example.com/"

       (compared using ==)
     # ./spec/requests/omniauth_callbacks_controller_spec.rb:445:in `block (5 levels) in <main>'
```

Also wondering, if we need this in `RAILS_ENV=profile` cause `ruby script/bench.rb` uses that (it also needs a bit of TLC due to the api key changes)

davidtaylorhq:
👍 looking at the weird fails. Looks like those tests were introduced after the enforce_hostname middleware, so need to figure out the best option there. (either normalise on www.example.com, or on test.localhost). I'll take a look at RAILS_ENV=profile as well

PR #8670, opened 2020-01-07
DEV: Update bench.rb for core changes

davidtaylorhq:
- Use new api key rake task
- Switch to header-based API auth
- Stop hard-coding topic id

### DIFF
--- a/script/profile_db_generator.rb
+++ b/script/profile_db_generator.rb
@@ -65,7 +65,7 @@ unless Rails.env == "profile"
 end
 
 def ensure_perf_test_topic_has_right_title!
-  t = Topic.find(179)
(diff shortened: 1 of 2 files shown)


script/profile_db_generator.rb
```diff
@@ -65,7 +65,7 @@ def create_admin(seq)
 end
 
 def ensure_perf_test_topic_has_right_title!
-  t = Topic.find(179)
+  t = Topic.where(archetype: :regular).last
```
SamSaffron: minor but I think this picks an inconsistent topic, topics have random numbers of posts so we really want to pick a topic with exactly N posts in it. will adjust.

SamSaffron:
This is looking good!

PR #8671, opened 2020-01-07
FIX: Make scrolling to bottom post in topic more consistent

martin-brennan:
Meta topic: https://meta.discourse.org/t/dragging-the-timeline-handle-to-the-end-brings-you-to-the-bottom-of-the-page-not-the-bottom-of-the-discussion/137899

* When using the slider to scroll past the bottom post,
  we now scroll to the bottom of the post/page IF the
  post height is > the window height (e.g. really long
  posts). if the post height is smaller, then we lock
  onto and jump to the top of the post
* This also removes the mobile hack that would always jump
  to the top of the last post on mobile

This works on both mobile and desktop. To test, try dragging the timeline handle past the end of the last post and letting go. For small height posts we jump to the top of the last posts. For large height posts we jump to the end of the post which often just shows the end of the page with the bottom of the post.

The only main concern I still have here is whether the post height > window height rule is OK. On my laptop my window height for Chrome is ~811 so any post taller than that will make the slider jump to the end of the page. I am not sure how to make this rule more dynamic without adding arbitrary heights to the window height in the comparison or doing ranges e.g. if your window height is 800-1200 the post must be > 1400 height to jump to end of the page etc.


SamSaffron:
Cool, let's try this out!

mamr235689:
https://meta.discourse.org/t/dragging-the-timeline-handle-to-the-end-brings-you-to-the-bottom-of-the-page-not-the-bottom-of- ที่อภิปราย / 137899/8

mamr235689:
http://www.martin-brennan.com

nebulousdog:
seems a bit bugged. I'm constantly scrolling through months to newest messages. :grimacing:

PR #8674, opened 2020-01-07
enqueue spam/dmarc failing emails instead of hiding

LeoMcA:
This is how an email which fails DMARC, for instance, appears (mostly through previous work from @eviltrout):

Screenshot from 2020-01-07 14-54-12

At the moment this doesn't show the reason that this topic needs approval, and I think in the DMARC case at least, this is important (otherwise you'll be left wondering why a post from an admin, for instance, is in the reviewable queue). But this reason is already stored in the reviewable score, so @eviltrout I wondered if you'd already had thoughts about exposing this.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -4655,7 +4655,7 @@ en:
       category: "Posts in this category require manual approval by staff. See the category settings."
       must_approve_users: "All new users must be approved by staff. See `must_approve_users`."
       invite_only: "All new users should be invited. See `invite_only`."
-      email_auth_res_enqueue: "This email failed a DMARC check, it most likely isn\'t from whom it seems to be from. Check the raw email headers for more information."
+      email_auth_res_enqueue: "This email failed a DMARC check, it most likely isn't from whom it seems to be from. Check the raw email headers for more information."
       email_spam: "This email was flagged as spam by the header defined in `email_in_spam_header`."
 
     actions:
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -2402,6 +2402,7 @@ en:
       fixed_primary_email: "Fixed primary email for staged user"
       same_ip_address: "Same IP address (%{ip_address}) as other users"
       inactive_user: "Inactive user"
+    email_in_spam_header: "User's first email was flagged as spam"
 
   reviewables_reminder:
     submitted:
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -4655,6 +4655,8 @@ en:
       category: "Posts in this category require manual approval by staff. See the category settings."
       must_approve_users: "All new users must be approved by staff. See `must_approve_users`."
       invite_only: "All new users should be invited. See `invite_only`."
+      email_auth_res_enqueue: "This email failed a DMARC check, it most likely isn\'t from whom it seems to be from. Check the raw email headers for more information."
+      email_spam: "This email was flagged as spam by the header defined in `email_in_spam_header`."
 
     actions:
       agree:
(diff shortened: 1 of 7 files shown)


LeoMcA:
A-ha! That's very nice!

davidtaylorhq:
@LeoMcA looks like the test failures are showing something legit here

```
Psych::SyntaxError: (/home/runner/work/discourse/discourse/config/locales/server.en.yml): found unknown escape character while parsing a quoted scalar at line 4659 column 31
```
I don't think you need to escape the single quote in the translation.

eviltrout:
Could we default this to off for now? It seems a bit safer. Then we can give it a try and if it all works nicely we can default it on.

LeoMcA:
> I don't think you need to escape the single quote in the translation.

Oh yeah... for some reason Atom's syntax highlighting bugs out with an unescaped quote, so I assumed it needed to be escaped without checking. Odd that that exception isn't thrown locally for me, though.

> Could we default this to off for now?

@eviltrout Which aspect of this you thinking? The spam checking element of this has existed as a feature for ages, so I wouldn't want to effectively turn that off on instances that have enabled it. The authentication results checking was added much more recently - but it's possible some instances are already relying on it.

LeoMcA:
Any update here? We have near-daily phishing emails coming through which are marked as spam by SES or fail dmarc, but which still trigger email notifications to users because the current behaviour of posting + hiding doesn't prevent email notifications being sent.

davidtaylorhq:
I think we would like to have the whole DMARC feature disabled by default for now. That matches the 'disabled by default' behaviour of the old spam detection site setting. The only DMARC failures we've had hitting our meta inbox are 'real' customers with misconfigured email servers, so this is just causing  confusion on our end. Any actual spam emails are being filtered upstream by google.

What do you think about disabling the feature completely when `email_in_authserv_id` is blank?

Aside: how do the 'well known values' [in the meta post](https://meta.discourse.org/t/configuring-authentication-checks-on-incoming-email/134358?u=david) get used? As far as I can tell, when `email_in_authserv_id` is blank, we read all `Authentication-Results` headers, and take the worst one. I can't see any specific checks for amazon/google.

LeoMcA:
> I think we would like to have the whole DMARC feature disabled by default for now. That matches the 'disabled by default' behaviour of the old spam detection site setting. The only DMARC failures we've had hitting our meta inbox are 'real' customers with misconfigured email servers, so this is just causing confusion on our end.

Fair enough, I'll change the behaviour when `email_in_authserv_id` is blank from "if there's a failing header, observe it" to "just do nothing". I made that decision from a naïve security perspective (where everyone has their email servers set up correctly!)

> Aside: how do the 'well known values' [in the meta post](https://meta.discourse.org/t/configuring-authentication-checks-on-incoming-email/134358?u=david) get used? As far as I can tell, when `email_in_authserv_id` is blank, we read all `Authentication-Results` headers, and take the worst one. I can't see any specific checks for amazon/google.

Those 'well known values' are for users to paste into the `email_in_authserv_id` setting if they know they're using amazon/google, so they don't have to grep through their received emails. We can't hard code them in, since, say a user's using amazon as their email receiver, an attacker could then include a `Authentication-Results: mx.google.com;` header in their email, and amazon wouldn't strip it out (because it's not `mx.google.com`).

davidtaylorhq:
> where everyone has their email servers set up correctly!

I wish that were the case! It would make this so much easier!

> Those 'well known values' are for users to paste into the email_in_authserv_id

:+1: makes sense, thanks for clarifying

LeoMcA:
> +1 makes sense, thanks for clarifying

The meta topic could probably do with a clarification (something like "Paste these values into your `email_in_authserv_id` setting:" under the "Well known values" header) but I don't have permission to edit the post.

eviltrout:
Great, thanks for the legwork @LeoMcA  - I've merged it now.

PR #8681, opened 2020-01-08
DEV: Fix function prototype deprecations

CvX:
Fixes following two deprecations (and cleans up some imports):

> DEPRECATION: Function prototype extensions have been deprecated, please migrate from function(){}.on('foo') to on('foo', function() {}). [deprecation id: function-prototype-extensions.on] See https://deprecations.emberjs.com/v3.x/#toc_function-prototype-extensions-on for more details.

and

> DEPRECATION: Function prototype extensions have been deprecated, please migrate from function(){}.observes('foo') to observer('foo', function() {}). [deprecation id: function-prototype-extensions.observes] See https://deprecations.emberjs.com/v3.x#toc_function-prototype-extensions-observes for more details.


ZogStriP:
Would have been easier to review if each deprecation fix was it its own PR.

CvX:
I've grouped them together since the event listener one is just a single change (https://github.com/discourse/discourse/pull/8681/commits/c532f98929710f45fc7259c3909d1f1685d889ab) and both are about function prototype extensions.

jjaffeux:
Looks quite safe to me :+1:

jjaffeux:
I fixed the merge conflict as my FIX is at the root of this conflict, should be good :+1:

CvX:
Given the updated beta timeline I'm going ahead and merging this.
The risk is rather low since this PR only replaces one syntax sugar with another, and we still have a couple weeks to catch any issues before the next beta lands.

PR #8683, opened 2020-01-08
FIX: Use new tag routes

udan11:
Daniel added unambiguous routes for tags in 1fb7a6297c7b5ebcd8433cdecd8bd41ec9fc4f3e. This PR ensures that the new routes are used.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -191,10 +191,10 @@ class TagsController < ::ApplicationController
     discourse_expires_in 1.minute
 
     tag_id = params[:tag_id]
-    @link = "#{Discourse.base_url}/tags/#{tag_id}"
+    @link = "#{Discourse.base_url}/tag/#{tag_id}"
     @description = I18n.t("rss_by_tag", tag: tag_id)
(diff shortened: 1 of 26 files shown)


app/controllers/tags_controller.rb
```diff
@@ -191,10 +191,10 @@ def tag_feed
     discourse_expires_in 1.minute
 
     tag_id = params[:tag_id]
-    @link = "#{Discourse.base_url}/tags/#{tag_id}"
+    @link = "#{Discourse.base_url}/tag/#{tag_id}"
```
ZogStriP: Should we use the rails url helper here instead?

app/controllers/tags_controller.rb
```diff
@@ -191,10 +191,10 @@ def tag_feed
     discourse_expires_in 1.minute
 
     tag_id = params[:tag_id]
-    @link = "#{Discourse.base_url}/tag/#{tag_id}"
+    @link = Discourse.base_url + Rails.application.routes.url_helpers.tag_show_path(tag_id: tag_id)
```
ZogStriP: Hmm, yeah that's actually worse 😢

PR #8684, opened 2020-01-08
Minor update (*.ru.yml)

iPopstop:


eviltrout:
Please use transifex to update text for Discourse:
https://blog.discourse.org/2014/05/translating-discourse-with-transifex/

PR #8692, opened 2020-01-09
FIX: Change rootNone behavior in category-chooser

CvX:
This breaking change was originally a deprecation fix for overriding a computed property `none`.

There are 4 uses of `rootNone` in core and "all-the-plugins":

1. in **discourse-chat-integration**, admin-plugins-chat-edit-rule.hbs - changed behavior, that I'd consider a fix - `rootNoneLabel` is now used regardless of `siteSettings.allow_uncategorized_topics` value, which I believe was an originally intended behavior (i.e. it most likely hasn't been tested with disabled uncategorized topics)
2. in **discourse-slack-official**, plugins-slack.hbs - the same as 1.
3. in **core**, edit-category-general.hbs (in this PR) - no change in behavior
4. in **discourse-googlebooks**, edit-category-general.hbs - no change in behavior (since `allowUncategorized="true"` is also passed as an argument)

@jjaffeux what do you think about this bugfix?

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/edit-category-general.hbs
+++ b/app/assets/javascripts/discourse/templates/components/edit-category-general.hbs
@@ -12,7 +12,7 @@
     <section class='field'>
       <label>{{i18n 'category.parent'}}</label>
       {{category-chooser
-        none="category.none"
+        rootNone=true
         value=category.parent_category_id
         excludeCategoryId=category.id
         categories=parentCategories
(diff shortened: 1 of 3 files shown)


jjaffeux:
Should be fine, this is all gone in my branch anyways

CvX:
> this is all gone in my branch anyways

awesome! 🔥😃

jjaffeux:
Can we merge this rather fast please ? I need to backport it into sk2

PR #8693, opened 2020-01-09
FIX: prevents url of file from being pasted when pasting file on iOS

jjaffeux:


jjaffeux:
@vinothkannans on iOS (maybe android too) when you pasted a file in the composer it would upload the file and also paste the URL of the file.

Like this for example:
Screenshot 2020-01-09 at 15 49 28

What do you think of this fix?

./cc @ZogStriP

app/assets/javascripts/discourse/components/d-editor.js.es6
```diff
@@ -892,7 +892,7 @@ export default Component.extend({
       }
     }
 
-    if (handled) {
+    if (handled || (!handled && canUpload)) {
```
ZogStriP: I think you can shorten that to

```javascript
if (handled || canUpload) {
```

jjaffeux: Yes we already had this discussion, I just think my version much more readable. Making both conditions explicit, instead of making the second one relying on the first one. But I'm fine changing it no problem.

eviltrout: I have to side with @ZogStriP here - the negative makes me personally read it as much more complicated than it is.

vinothkannans: I agree with @ZogStriP & @eviltrout. We can add JS comment to prevent confusion.

jjaffeux:
```suggestion
    if (handled || canUpload) {
```

jjaffeux: No a comment is not needed, all right let's go with this 👍

jjaffeux:
@vinothkannans feel free to merge when you start your day

PR #8694, opened 2020-01-09
Feature: Mass award badge

romanrizzi:
The mass award will happen in batches of 200 users.

No badge selected:

Screen Shot 2020-01-09 at 16 00 29

When a badge is selected:

Screen Shot 2020-01-09 at 16 00 48

### DIFF
--- a/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
@@ -1,14 +1,28 @@
 import Controller from "@ember/controller";
+import { ajax } from "discourse/lib/ajax";
+import { popupAjaxError } from "discourse/lib/ajax-error";
 
 export default Controller.extend({
   saving: false,
 
   actions: {
     massAward() {
-      this.set("saving", true);
-      setTimeout(() => {
-        this.set("saving", false);
-      }, 3000);
+      const file = document.querySelector("#massAwardCSVUpload").files[0];
+
+      const options = {
+        type: "POST",
+        processData: false,
+        contentType: false,
+        data: new FormData()
+      };
+
+      options.data.append("file", file);
+
+      this.set("saving", false);
+
+      ajax(`/admin/badges/award/${this.model.id}`, options)
+        .then(() => this.set("saving", false))
+        .catch(popupAjaxError);
     }
   }
 });
--- a/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
@@ -1,0 +1,14 @@
+import Controller from "@ember/controller";
+
+export default Controller.extend({
+  saving: false,
+
+  actions: {
+    massAward() {
+      this.set("saving", true);
+      setTimeout(() => {
+        this.set("saving", false);
+      }, 3000);
+    }
+  }
+});
--- a/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
+++ b/app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
@@ -9,20 +9,27 @@ export default Controller.extend({
     massAward() {
       const file = document.querySelector("#massAwardCSVUpload").files[0];
 
-      const options = {
-        type: "POST",
-        processData: false,
-        contentType: false,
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
```diff
@@ -0,0 +1,35 @@
...
+          data: new FormData()
+        };
+
+        options.data.append("file", file);
+
+        this.set("saving", true);
+
+        ajax(`/admin/badges/award/${this.model.id}`, options)
+          .then(() => {
+            bootbox.alert(I18n.t("admin.badges.mass_award.success"));
```
eviltrout: Maybe this should redirect to the front end view so they can see who was awarded the badge?

app/assets/javascripts/admin/controllers/admin-badges.js.es6
```diff
@@ -1,2 +1,19 @@
 import Controller from "@ember/controller";
-export default Controller.extend();
+import { inject as service } from "@ember/service";
+
+export default Controller.extend({
+  selectedRoute: "adminBadges.show",
+  router: service(),
+
+  actions: {
+    showBadge() {
```
eviltrout: Could we convert these actions to links? Buttons can have `href`s

app/assets/javascripts/admin/templates/badges-award.hbs
```diff
@@ -0,0 +1,22 @@
+{{#d-section class="award-badge"}}
+  <form class="form-horizontal">
+    <h1>{{i18n 'admin.badges.mass_award.title'}}</h1>
+    <div class='badge-preview'>
+      {{#if model}}
+        {{icon-or-image model}}
```
eviltrout: I noticed in the screenshot that this is extremely big. Could we shrink it a bit?

app/controllers/admin/badges_controller.rb
```diff
@@ -33,6 +35,38 @@ def new
...
+
+    batch_number = 1
+    batch = []
+
+    File.open(csv_file) do |csv|
+      csv.each_line do |email_line|
+        batch.concat CSV.parse_line(email_line)
+
+        # Split the emails in batches of 200 elements.
+        full_batch = csv.lineno % (BadgeGranter::MAX_ITEMS_FOR_DELTA * batch_number) == 0
```
eviltrout: I wonder if 200 emails will be too large as a redis value for sidekiq? Might be ok, but we should definitely test with batches larger than 200 and perhaps tune it.

app/services/badge_granter.rb
```diff
@@ -12,6 +12,40 @@ def self.grant(badge, user, opts = {})
...
   end
 
+  def self.mass_grant(badge, users)
+    return unless badge.enabled?
+
+    system_user_id = Discourse.system_user.id
+    user_badges = users.map { |u| { badge_id: badge.id, user_id: u.id, granted_by_id: system_user_id, granted_at: Time.now } }
+    granted_badges = UserBadge.insert_all(user_badges, returning: %i[user_id])
+
+    users_to_notify = users.each do |user|
```
eviltrout: This variable seems unused?

app/services/badge_granter.rb
```diff
@@ -12,6 +12,40 @@ def self.grant(badge, user, opts = {})
...
+      use_default_locale = !SiteSetting.allow_user_locale && user.locale.blank?
+      notification_locale = use_default_locale ? SiteSetting.default_locale : user.locale
+
+      notification = I18n.with_locale(notification_locale) do
```
eviltrout: This feels a little repetitive from the #grant method below - could you split it out to a function and use it in both places?

app/assets/javascripts/admin/controllers/admin-badges-award.js.es6
romanrizzi: The mass award will happen asynchronously, does the redirect still make sense?

app/assets/javascripts/admin/controllers/admin-badges.js.es6
romanrizzi: I replaced the transition with an "href". However, I still need the action to dynamically set the badges link-to url.

PR #8695, opened 2020-01-09
FEATURE: pass in excluded usernames to user-selector

markvanlan:


markvanlan:
The core frontend checks failed, but passed for me locally

test/javascripts/components/user-selector-test.js.es6
```diff
@@ -2,6 +2,12 @@ import componentTest from "helpers/component-test";
 
 moduleForComponent("user-selector", { integration: true });
 
+const paste = (element, text) => {
```
eviltrout:
```suggestion
function paste(element, text) {
```
If it's not a closure why not use the function syntax?

PR #8697, opened 2020-01-10
DEV: Refactor 2FA authentication flows for sanity

martin-brennan:
* Move confusing 2FA logic involving whether to authenticate with security key or totp to second factor manager and clean up, and move authenticate webauthn functionality to second factor manager.
* Replace confusing code in session controller with new calls to second factor manager
* Add specs for new manager code and ensure current session controller specs work fine

### DIFF
--- a/app/models/concerns/second_factor_manager.rb
+++ b/app/models/concerns/second_factor_manager.rb
@@ -5,6 +5,31 @@ module SecondFactorManager
 
   extend ActiveSupport::Concern
 
+  class SecondFactorAuthenticationResult
+    attr_reader :ok, :error, :reason, :backup_enabled, :multiple_second_factor_methods
+
+    def initialize(ok, params)
+      @ok = ok
+      @error = params[:error]
+      @reason = params[:reason]
+      @backup_enabled = params[:backup_enabled]
+      @multiple_second_factor_methods = params[:multiple_second_factor_methods]
+    end
+
+    def to_h
+      {
+        error: error,
+        reason: reason,
+        backup_enabled: backup_enabled,
+        multiple_second_factor_methods: multiple_second_factor_methods
+      }
+    end
+
+    def ok?
+      @ok
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


app/models/concerns/second_factor_manager.rb
```diff
@@ -5,6 +5,31 @@ module SecondFactorManager
...
+    def to_h
+      {
+        error: error,
+        reason: reason,
+        backup_enabled: backup_enabled,
+        multiple_second_factor_methods: multiple_second_factor_methods
+      }
+    end
+
+    def ok?
```
ZogStriP: Is this really necessary? Since `ok` is already `attr_reader`

app/models/concerns/second_factor_manager.rb
```diff
@@ -5,6 +5,31 @@ module SecondFactorManager
 
   extend ActiveSupport::Concern
 
+  class SecondFactorAuthenticationResult
```
ZogStriP: Do we need this to be a `class`? Wouldn't a `struct` be enough?

```ruby
SecondFactorAuthenticationResult = Struct.new(:ok, :error, :reason, :backup_enabled, :multiple_second_factor_methods)
```

app/models/concerns/second_factor_manager.rb
```diff
@@ -68,26 +93,98 @@ def security_keys_enabled?
...
+  def only_totp_or_backup_codes_enabled?
+    !security_keys_enabled? && totp_or_backup_codes_enabled?
   end
 
   def remaining_backup_codes
     self&.user_second_factors&.backup_codes&.count
   end
 
   def authenticate_second_factor(token, second_factor_method)
+    second_factor_method = second_factor_method.to_i if !second_factor_method.is_a?(Integer)
```
ZogStriP: Any reasons we need that test?

```suggestion
    second_factor_method = second_factor_method.to_i
```

PR #8698, opened 2020-01-10
PERF: Cache ranks for featured badges, to simplify user serialization

davidtaylorhq:
This introduces a new `featured_rank` column on the `user_badges` table. This is kept up-to-date automatically, and makes it much easier to load badges for user profiles and user cards.

### DIFF
--- a/app/jobs/scheduled/badge_grant.rb
+++ b/app/jobs/scheduled/badge_grant.rb
@@ -22,6 +22,7 @@ module Jobs
       end
 
       BadgeGranter.revoke_ungranted_titles!
+      UserBadge.ensure_consistency! # Badge granter sometimes uses raw SQL, so hooks do not run. Clean up data
     end
 
   end
--- a/app/models/user_badge.rb
+++ b/app/models/user_badge.rb
@@ -100,6 +100,7 @@ end
 #  post_id         :integer
 #  notification_id :integer
 #  seq             :integer          default(0), not null
+#  featured_rank   :integer
 #
 # Indexes
 #
--- a/app/models/user_badge.rb
+++ b/app/models/user_badge.rb
@@ -75,7 +75,7 @@ class UserBadge < ActiveRecord::Base
       )
       -- Now use that data to update the featured_rank column
       UPDATE user_badges SET featured_rank = rank_number
-      FROM ranks WHERE ranks.badge_id = user_badges.badge_id AND ranks.user_id = user_badges.user_id
+      FROM ranks WHERE ranks.badge_id = user_badges.badge_id AND ranks.user_id = user_badges.user_id AND featured_rank IS DISTINCT FROM rank_number
     SQL
 
     DB.exec query
--- a/app/models/user_badge.rb
+++ b/app/models/user_badge.rb
@@ -10,13 +10,12 @@ class UserBadge < ActiveRecord::Base
   scope :grouped_with_count, -> {
     group(:badge_id, :user_id)
       .select(UserBadge.attribute_names.map { |x| "MAX(user_badges.#{x}) AS #{x}" },
-                  'COUNT(*) AS "count"',
-                  'MAX(badges.badge_type_id) AS badges_badge_type_id',
-                  'MAX(badges.grant_count) AS badges_grant_count')
-      .joins(:badge)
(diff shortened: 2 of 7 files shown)


app/models/user_badge.rb
```diff
@@ -19,15 +29,58 @@ class UserBadge < ActiveRecord::Base
...
+  end
+
+  def self.update_featured_ranks!(user_id = nil)
+    query = <<~SQL
+      WITH featured_tl_badge AS -- Find the best trust level badge for each user
+      (
+        SELECT user_id, max(badge_id) as badge_id
+        FROM user_badges
+        WHERE badge_id IN (1,2,3,4)
+        #{"AND user_id = #{user_id}" if user_id}
```
eviltrout:
```suggestion
        #{"AND user_id = #{user_id.to_i}" if user_id}
```
Protect ourselves against injection in the future?

app/models/user_badge.rb
```diff
@@ -19,15 +29,58 @@ class UserBadge < ActiveRecord::Base
...
+                    MAX(featured_tl_badge.user_id) NULLS LAST, -- Best tl badge first
+                    CASE WHEN user_badges.badge_id IN (1,2,3,4) THEN 1 ELSE 0 END ASC, -- Non-featured tl badges last
+                    MAX(badges.badge_type_id) ASC,
+                    MAX(badges.grant_count) ASC,
+                    user_badges.badge_id DESC
+          ) rank_number
+        FROM user_badges
+        INNER JOIN badges ON badges.id = user_badges.badge_id
+        LEFT JOIN featured_tl_badge ON featured_tl_badge.user_id = user_badges.user_id AND featured_tl_badge.badge_id = user_badges.badge_id
+        #{"WHERE user_badges.user_id = #{user_id}" if user_id}
```
eviltrout:
```suggestion
        #{"WHERE user_badges.user_id = #{user_id.to_i}" if user_id}
```

app/jobs/scheduled/badge_grant.rb
```diff
@@ -22,6 +22,7 @@ def execute(args)
       end
 
       BadgeGranter.revoke_ungranted_titles!
+      UserBadge.ensure_consistency! # Badge granter sometimes uses raw SQL, so hooks do not run. Clean up data
```
SamSaffron: This worries me a bit, what is the cost of the new big query on one of our larger sites?

db/migrate/20200107161405_add_featured_rank_to_user_badges.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+class AddFeaturedRankToUserBadges < ActiveRecord::Migration[6.0]
+  def change
+    add_column :user_badges, :featured_rank, :integer, null: true
+
+    execute <<~SQL
```
SamSaffron: will this lock up the table in a transaction? do we need to be careful here and run it outside of a transaction?

SamSaffron:
Overall I really like this change, we just have to be super careful with upgrade and need to make sure we are not introducing extreme costs to our daily schedule.

db/migrate/20200107161405_add_featured_rank_to_user_badges.rb
davidtaylorhq: During the transaction, it holds `RowExclusiveLock`s over the rows being updated. That means existing rows in the UserBadge table cannot be updated by any other process during this time. I think new rows could still be added, since no table-level locks area being held.

davidtaylorhq:
Yes, I did a few tests to make sure it's not too expensive. Running this as a `SELECT`, rather than `UPDATE`, I tested this on a few sites.

On meta with 41k users, and 94k user_badges, it took 0.6 seconds.
On a site with 4.2 million users, and 590k user_badges, it took 2.3 seconds

db/migrate/20200107161405_add_featured_rank_to_user_badges.rb
SamSaffron: 5 seconds on a large db sounds fine to me. 5 minutes and stuff can get a bit tricky.

SamSaffron:
This is looking good, feel free to merge

PR #8709, opened 2020-01-13
FIX: Make topic query include topics from sub-sub-categories

udan11:


lib/topic_query.rb
```diff
@@ -677,13 +677,15 @@ def default_results(options = {})
...
         sql = <<~SQL
             categories.id IN (
-              SELECT c2.id FROM categories c2 WHERE c2.parent_category_id = :category_id
-              UNION ALL
-              SELECT :category_id
-            ) AND
-            topics.id NOT IN (
-              SELECT c3.topic_id FROM categories c3 WHERE c3.parent_category_id = :category_id AND c3.topic_id IS NOT NULL
-            )
+              WITH RECURSIVE subcategories AS (
```
ZogStriP: Do we need to limit the recursion level based on the site setting?

spec/components/topic_query_spec.rb
```diff
@@ -153,13 +153,25 @@
...
       let!(:subcategory) { Fabricate(:category_with_definition, parent_category_id: category.id) }
+      let(:subsubcategory) { Fabricate(:category_with_definition, parent_category_id: subcategory.id) }
 
       it "works with subcategories" do
         expect(TopicQuery.new(moderator, category: category.id).list_latest.topics.size).to eq(1)
         expect(TopicQuery.new(moderator, category: subcategory.id).list_latest.topics.size).to eq(1)
         expect(TopicQuery.new(moderator, category: category.id, no_subcategories: true).list_latest.topics.size).to eq(1)
       end
 
+      it "works with subsubcategories" do
```
ZogStriP: 😍

PR #8713, opened 2020-01-14
Make version the same as install docs

rimian:
Is this version correct?

https://github.com/techAPJ/install-rails/blob/master/mac


SamSaffron:
oh I wonder why we even need this file. agree it should be updated, thanks Rim

PR #8723, opened 2020-01-15
FIX: show error message if the topic deletion fails

techAPJ:
When staff tries to delete category definition topic the topic does not gets deleted but the client side code indicates that topic has been deleted. This commit will show error message from server side if topic can not be deleted.


ZogStriP:
Looks good 👍 

Any chances you could also add a test?

techAPJ:
I have added a test but skipping it for now as I need to debug why it's not showing "recover" button on delete. Will fix, but I need to merge this PR first to verify another fix.

PR #8726, opened 2020-01-15
FEATURE: text line with only one emoji will now display a bigger emoji

jjaffeux:
Screenshot 2020-01-15 at 17 21 16

### DIFF
--- a/app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
+++ b/app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
@@ -220,14 +220,6 @@ function applyEmoji(
 
       result.push(token);
 
-      if (i === 0 && offset === content.length) {
-        token.attrs.forEach(attr => {
-          if (attr[0] === "class") {
-            attr[1] = `${attr[1]} only-emoji`;
-          }
-        });
-      }
-
       end = start = i + offset;
       i += offset - 1;
     }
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -220,6 +220,14 @@ function applyEmoji(
 
       result.push(token);
 
+      if (i === 0 && i + offset === content.length) {
```
ZogStriP: If `i == 0`, then there's no need to use it after 😉 

```suggestion
      if (i === 0 && offset === content.length) {
```

app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -220,6 +220,14 @@ function applyEmoji(
...
+        token.attrs.forEach(attr => {
+          if (attr[0] === "class") {
+            attr[1] = `${attr[1]} all-emoji`;
```
ZogStriP: I would have used `only-emoji`.

jjaffeux: I followed what mattermost was using on this, but sure only-emoji works for me too

app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -257,5 +265,10 @@ export function setup(helper) {
     );
   });
 
-  helper.whiteList(["img[class=emoji]", "img[class=emoji emoji-custom]"]);
+  helper.whiteList([
```
ZogStriP: That feels a bit weird. Isn't there a better way to whitelist 3 classes on the `img` tag?

jjaffeux: No I don't think there's a better way

spec/components/pretty_text_spec.rb
```diff
@@ -109,6 +109,28 @@ def cook(*args)
 
           expect(cook(md)).to eq(html.strip)
         end
+
+        it "adds an all-emoji class when a line has only one emoji" do
+          md = <<~MD
+            foo 😀
+            foo 😀 bar
```
ZogStriP: Same with `😉 foo` ?

app/assets/stylesheets/common/base/emoji.scss
```diff
@@ -4,6 +4,11 @@ img.emoji {
   vertical-align: middle;
 }
 
+img.emoji.all-emoji {
```
ZogStriP: Needs to be changed to `.only-emoji` as well 😉

jjaffeux: weird it was working with this on my screen 🤔  probably thought I rebaked and didnt :+1:

app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
jjaffeux: maybe @SamSaffron  can advise here, but AFAIK you have to define each combination

SamSaffron:
I think the whitelister is kind of silly I don't think we have any shortcuts. This looks good to me.

BTW I noticed on iOS that if you have 2 emojis by themselves or even 3 they are all embiggened. You have to have 4 for them to be small. Twitter requires 11 emojis to stop the embiggening.

I think at a minimum we should embiggen with 3 and on the 4th stop.

app/assets/javascripts/pretty-text/engines/discourse-markdown/emoji.js.es6
```diff
@@ -231,9 +231,66 @@ function applyEmoji(
     result.push(text);
   }
 
+  // we check for a result <= 5 because we support maximum 3 large emojis
+  // EMOJI SPACE EMOJI SPACE EMOJI => 5 tokens
+  if (result && result.length <= 5 && result.length > 0) {
```
ZogStriP: Tiny, but that condition order feels weird 😜

```suggestion
  if (result && result.length > 0 && result.length <= 5) {
```

PR #8727, opened 2020-01-15
FIX: Styling for feature topic on profile modal

markvanlan:


ZogStriP:
Can we get before/after screenshots?

markvanlan:
Before the categories were just squashed next to the title 

Screenshot from 2020-01-15 14-44-39

Now they are below, and the save button is in the modal footer

Screenshot from 2020-01-15 14-39-37

PR #8729, opened 2020-01-15
FEATURE: Drop "backup" schema 7 days after restore

gschlager:
The "backup" schema is used to rollback a failed restore. It isn't useful after a longer period of time and turns into a waste of disk space.

### DIFF
--- a/app/jobs/scheduled/drop_backup_schema.rb
+++ b/app/jobs/scheduled/drop_backup_schema.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DropBackupSchema < ::Jobs::Scheduled
+    every 1.day
+
+    def execute(_)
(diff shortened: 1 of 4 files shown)


app/jobs/scheduled/drop_backup_schema.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DropBackupSchema < ::Jobs::Scheduled
+    every 1.day
+
+    def execute(_)
```
eviltrout: Why not `def execute`?

gschlager: The `args` parameter isn't needed in this job, but it's required. Removing the argument would lead to a `Job exception: wrong number of arguments (given 1, expected 0)`

eviltrout: Ah yes of course!

PR #8730, opened 2020-01-16
FIX: Mobile topic-timeline not closing on scrolling past last post

martin-brennan:
See https://meta.discourse.org/t/post-excerpts-are-missing-from-the-expanded-mobile-timeline/132768/4.

* When we dragged the topic-timeline handle past the last post
  in a topic we were not closing the timeline as we would if
  scrolling to a specific post
* This also fixes the issue where when scrolling past the end of
  the topic with a massive last post, none of the post could be
  seen


SamSaffron:
lets give this a shot!

PR #8731, opened 2020-01-16
FEATURE: Update upload security status on post move, topic conversion, category change

martin-brennan:
Add TopicUploadSecurityManager to handle post moves. When a post moves around or a topic changes between categories and public/private message status the uploads connected to posts in the topic need to have their secure status updated, depending on the security context the topic now lives in.


app/jobs/regular/update_topic_upload_security.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Jobs
+
+  class UpdateTopicUploadSecurity < ::Jobs::Base
+
+    def execute(args)
+      TopicUploadSecurityManager.new(Topic.find(args[:topic_id])).run
```
ZogStriP: Do we want this job to throw an exception when the topic can't be found?

martin-brennan: Good point, probably no need to throw an exception, no need to update upload security in a topic that no longer exists

PR #8736, opened 2020-01-16
REFACTOR: Rename `post_replies.reply_id` column to `post_replies.reply_post_id`

gschlager:
Follow-up to bac1bcc79f73d8890c5d0e1489ab9a4caf3ed276

### DIFF
--- a/db/migrate/20200116140132_rename_reply_id_column.rb
+++ b/db/migrate/20200116140132_rename_reply_id_column.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class RenameReplyIdColumn < ActiveRecord::Migration[6.0]
+  def up
+    Migration::ColumnDropper.mark_readonly(:post_replies, :reply_id)
+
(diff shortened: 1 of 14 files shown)


db/migrate/20200116140132_rename_reply_id_column.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class RenameReplyIdColumn < ActiveRecord::Migration[6.0]
+  def up
+    Migration::ColumnDropper.mark_readonly(:post_replies, :reply_id)
+
+    add_column :post_replies, :reply_post_id, :integer
```
gschlager: I'm quite surprised that it's possible to add a non-nullable column without problems. Maybe it works because there's an `UPDATE` inside the same transaction.

gschlager:
I updated the PR and split up the migration in order to add and drop indexes concurrently. In my eyes that makes sense since `post_replies` could be a fairly large table.

PR #8737, opened 2020-01-17
FIX: Re-allow admin login fro SSO enabled sites

martin-brennan:
* When we refactored away the admin-login route we introduced a bug where admins could not log into an SSO enabled site, because of a check in the email_login route that disallowed this.
* Allow admin to get around this check.


SamSaffron:
looks good to me, feel free to merge!

PR #8743, opened 2020-01-17
FIX: Don't cause exceptions due to rename of `reply_id` column

gschlager:
This syncs the value of the `reply_id` column into the `reply_post_id` column until all servers have been deployed and the post migrations ran.

Follow-up to ab07b945c28682ce58c5149003d65e6f3fd26dff

I tested it by executing the new `20200117172135_add_trigger_to_sync_post_replies` migration and running old code with the migrated database. Works fine. The trigger intercepts writes into the `reply_id` column and saves the value into the `reply_post_id` column. That way new code and old code will work until the deploy has finished. Then the post migration kicks in, deletes the trigger that does all the syncing and creates a new trigger to make the `reply_id` column read-only.

### DIFF
--- a/db/migrate/20200117172135_add_trigger_to_sync_post_replies.rb
+++ b/db/migrate/20200117172135_add_trigger_to_sync_post_replies.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class AddTriggerToSyncPostReplies < ActiveRecord::Migration[6.0]
+  def up
+    # we don't want this column to be readonly yet
+    Migration::ColumnDropper.drop_readonly(:post_replies, :reply_id)
+
+    DB.exec <<~SQL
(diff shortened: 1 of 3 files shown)


db/migrate/20200117172135_add_trigger_to_sync_post_replies.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class AddTriggerToSyncPostReplies < ActiveRecord::Migration[6.0]
+  def up
+    # we don't want this column to be readonly yet
+    Migration::ColumnDropper.drop_readonly(:post_replies, :reply_id)
```
eviltrout: Is this safe in the situation where it was never marked as readonly?

gschlager: Yes, it's executing `DROP FUNCTION IF EXISTS`

PR #8746, opened 2020-01-19
FIX: Quoting a quote preserves the original post information

nbianca:
Let's say post #2 quotes post number #1. If a user decides to quote the
quote in post #2, it should keep the information of post #1
("user_1, post: 1, topic: X"), instead of replacing with current post
info ("user_2, post: 2, topic: X").

### DIFF
--- a/app/assets/javascripts/discourse/components/quote-button.js.es6
+++ b/app/assets/javascripts/discourse/components/quote-button.js.es6
@@ -51,7 +51,7 @@ export default Component.extend({
     let opts;
     for (
       let element = selectedElement();
-      element;
+      element && element.tagName !== "ARTICLE";
       element = element.parentElement
     ) {
       if (element.tagName === "ASIDE" && element.classList.contains("quote")) {
--- a/app/assets/javascripts/discourse/components/quote-button.js.es6
+++ b/app/assets/javascripts/discourse/components/quote-button.js.es6
@@ -1,7 +1,7 @@
 import { scheduleOnce } from "@ember/runloop";
 import Component from "@ember/component";
 import discourseDebounce from "discourse/lib/debounce";
-import { selectedText } from "discourse/lib/utilities";
+import { selectedText, selectedElement } from "discourse/lib/utilities";
 
 export default Component.extend({
   classNames: ["quote-button"],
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/discourse/components/quote-button.js.es6
```diff
@@ -48,8 +48,28 @@ export default Component.extend({
       }
     }
 
+    let opts;
+    for (
+      let element = selectedElement();
+      element;
+      element = element.parentElement
```
ZogStriP: Do we need to climb the whole DOM tree up to the "root"? Could we just instead bail out as soon as we're out of the body of a post?

app/assets/javascripts/discourse/components/quote-button.js.es6
```diff
@@ -48,8 +48,28 @@ export default Component.extend({
...
+    ) {
+      if (element.tagName === "ASIDE" && element.classList.contains("quote")) {
+        opts = {
+          username:
+            element.dataset.username ||
+            element
+              .querySelector(".title")
+              .textContent.trim()
+              .replace(/:$/, ""),
```
ZogStriP: When do we have `$` in the username of a quote?

app/assets/javascripts/pretty-text/engines/discourse-markdown/quotes.js.es6
```diff
@@ -73,6 +73,10 @@ const rule = {
       token.attrs.push(["class", "quote no-group"]);
     }
 
+    if (username) {
+      token.attrs.push(["data-username", username]);
```
ZogStriP: Does this mean we need to rebake all the posts with a quote?

app/assets/javascripts/discourse/components/quote-button.js.es6
```diff
@@ -165,8 +185,8 @@ export default Component.extend({
   },
 
   click() {
-    const { postId, buffer } = this.quoteState;
-    this.attrs.selectText(postId, buffer).then(() => this._hideButton());
+    const { postId, buffer, opts } = this.quoteState;
+    this.attrs.selectText(postId, buffer, opts).then(() => this._hideButton());
```
jjaffeux: Think this should work

```suggestion
    this.attrs.selectText(postId, buffer, opts).then(this._hideButton);
```

PR #8747, opened 2020-01-19
FIX: Workaround limitation in jquery.autoellipsis

nbianca:
Calling $.ellipsis() on an element containing <br> elements would throw
an exception.

### DIFF
--- a/app/assets/javascripts/discourse/components/text-overflow.js.es6
+++ b/app/assets/javascripts/discourse/components/text-overflow.js.es6
@@ -7,6 +7,7 @@ export default Component.extend({
       const $this = $(this.element);
 
       if ($this) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/text-overflow.js.es6
```diff
@@ -7,6 +7,7 @@ export default Component.extend({
       const $this = $(this.element);
 
       if ($this) {
+        $this.find("br").replaceWith(" ");
```
ZogStriP: Are `br` and `hr` the only tags that are breaking the `ellipsis()` method?

nbianca: I couldn't find any other element that could break the `ellipsis()` method.

PR #8748, opened 2020-01-19
FEATURE: Export all types of reports

nbianca:
There is a single stacked_chart which was not exportable

### DIFF
--- a/app/jobs/regular/export_csv_file.rb
+++ b/app/jobs/regular/export_csv_file.rb
@@ -198,8 +198,25 @@ module Jobs
         end
       end
 
+      if report.modes == [:stacked_chart]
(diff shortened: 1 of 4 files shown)


app/jobs/regular/export_csv_file.rb
```diff
@@ -198,8 +198,25 @@ def report_export
         end
       end
 
+      if report.modes == [:stacked_chart]
```
ZogStriP: When we're exporting a `stacked_chart` report, we should probably avoid doing the loop above (starting on line `191`)

nbianca: That loop is necessary to populate the `titles` hash. 😊

PR #8749, opened 2020-01-19
FIX: Incorrect locale in badge granter

lis2:
We want to use default locale when:
a) Site settings are not allowing for user locale
OR
b) User locale are blank


ZogStriP:
Any chances you could add a test for that?

lis2:
@ZogStriP good point, specs added. I will wait for CI and I will merge it

PR #8753, opened 2020-01-20
DEV: Upgrade Ember to version 3.12.2

CvX:


spec/lib/theme_javascript_compiler_spec.rb
```diff
@@ -83,7 +83,7 @@ def statement(template)
     end
 
     # might change/break when updating ember
-    EMBER_INTERNAL_ID = 29
+    EMBER_INTERNAL_ID = 28
```
eviltrout: Could this use the `Ember::VERSION` or `BASE_COMPILER_VERSION` instead?

jjaffeux: No I don't think so, this is some thing in Ember compiler, it's not something "to bump" at each version.

eviltrout: Ah I understand. My recommendation then would be to ignore that value in the test.

jjaffeux:
@eviltrout I would like to merge this tomorrow morning if that's fine with you. We have been using for few days, seems rather safe.

spec/lib/theme_javascript_compiler_spec.rb
CvX: A bit more background, that's a glimmer vm opcode for helper (see: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/wire-format/lib/opcodes.ts#L35) that has changed after this commit: https://github.com/glimmerjs/glimmer-vm/commit/b4707985fe462052454e15ca5a537b3103156fdd

I've changed the test so it doesn't hard-code this value and updated the names so they're a bit more self-explanatory.

CvX:
I can't merge this in yet.
Ruby test failures were unrelated (https://github.com/discourse/discourse/commit/23b9da6b1cb94014497ae2724668594cfb9a85fe), so was the core-lint failure, but the frontend ones are legit as in: those aren't flaky tests and are caused by this upgrade, though I'm pretty confident it's only a test incompatibility and not a development/production build issue.
I'm still debugging this. I think I have to update qunit/ember-test-helpers, they haven't seen a bump in ~2.5 years.

jjaffeux:
> I can't merge this in yet.
> Ruby test failures were unrelated ([23b9da6](https://github.com/discourse/discourse/commit/23b9da6b1cb94014497ae2724668594cfb9a85fe)), so was the core-lint failure, but the frontend ones are legit as in: those aren't flaky tests and are caused by this upgrade, though I'm pretty confident it's only a test incompatibility and not a development/production build issue.
> I'm still debugging this. I think I have to update qunit/ember-test-helpers, they haven't seen a bump in ~2.5 years.

Have you made any work on updating qunit ?

CvX:
Not yet, but definitely will try to tackle this during the Ember 3.13 upgrade. In this PR I managed to make the CI pass with a bugfix in [discourse-canned-replies](https://github.com/discourse/discourse-canned-replies/commit/4d9430417635cca60ffa03fa94365e3e408b1add) and a change in wizard templates. 🚀 

Locally tests still either did stop or fail because of incompatibility between two (or more) plugins and the new Ember. I've split all plugins into two groups and separately they succeed. Anyway, save for that particular set of plugins I had and their last week's versions, it seems 2.12.2 does work now with the plugins installed in CI tests.

PR #8760, opened 2020-01-21
FEATURE: Use upload:// short URL for videos and audio in composer

martin-brennan:
For consistency this PR introduces using custom markdown and short upload:// URLs for video and audio uploads, rather than just treating them as links and relying on the oneboxer. The markdown syntax for videos is `file text|video` and for audio it is `file text|audio`.

This is achieved in discourse-markdown-it by modifying the rules for images in mardown-it via `md.renderer.rules.image`. We return HTML instead of the token when we encounter `audio` or `video` after `|` and the preview renders that HTML. Also when uploading an audio or video file we insert the relevant markdown into the composer.


app/assets/javascripts/discourse/lib/uploads.js.es6
```diff
@@ -220,37 +228,33 @@ export function uploadIcon(staff) {
...
+}
+
+function attachmentMarkdown(upload) {
+  return `[${upload.original_filename}|attachment](${
+    upload.short_url
+  }) (${I18n.toHumanSize(upload.filesize)})`;
 }
 
 export function getUploadMarkdown(upload) {
   if (isAnImage(upload.original_filename)) {
```
ZogStriP: Maybe we should rename this method to `isImage` so that it's the same as `isAudio` and `isVideo`?

PR #8765, opened 2020-01-22
FIX: Notify staged users about private categories

mintsaxon:
Group membership and `CategoryUser` notification level should be
respected to determine whether to notify staged users about activity in
private categories, instead of only ever generating notifications for staged
users' own topics (which has been the behaviour since
0c4ac2a)

### DIFF
--- a/lib/guardian/topic_guardian.rb
+++ b/lib/guardian/topic_guardian.rb
@@ -154,7 +154,7 @@ module TopicGuardian
 
     category = topic.category
     can_see_category?(category) &&
-      (!category.read_restricted || !is_staged? || topic.user == user)
+      (!category.read_restricted || !is_staged? || secure_category_ids.include?(category.id) || topic.user == user)
   end
 
   def can_get_access_to_topic?(topic)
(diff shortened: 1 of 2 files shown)


mintsaxon:
I left the behaviour of categories with `email_in_allow_strangers = true`alone, because I'm not clear whether it needs to change.

Based on [this discussion](https://meta.discourse.org/t/no-category-notifications-for-staged-users-in-private-catgeory/139230/3)

mintsaxon:
Also, I wasn't 100% sure about correct test naming or location, very happy to take any suggestions on that.

eviltrout:
@romanrizzi I'll defer to you because you made the previous fix and have a better idea of what's going on under the hood. Whatever you think is best!

mintsaxon:
@romanrizzi thanks for having a look!

> I recommend checking for `user.secure_category_ids.include?(category.id)` instead.

It seemed iffy to me to duplicate that check from `can_see_category?`, for some reason, but you're right that it makes perfect sense :) Updating the PR now.

romanrizzi:
I just tried this on "try" and it's working as expected. Thanks for fixing!

PR #8770, opened 2020-01-22
FEATURE: Replace existing badge owners when using the bulk award feature

romanrizzi:


app/services/badge_granter.rb
```diff
@@ -104,6 +104,27 @@ def self.revoke(user_badge, options = {})
     end
   end
 
+  def self.revoke_all(badge)
+    custom_badge_name = TranslationOverride.where(translation_key: badge.translation_key).pluck(:value)
+    custom_badge_name = custom_badge_name.reduce("") { |m, v| m += "'#{v}'," }.delete_suffix(',')
```
eviltrout: Won't this fail if there's a single quote in the translation? We should probably re-write this query to use active record and sanitization.

config/locales/client.en.yml
```diff
@@ -4502,6 +4502,7 @@ en:
           upload_csv: Upload a CSV with user emails
           aborted: Please upload a CSV containing user emails
           success: Your CSV was received and users will receive their badge shortly.
+          replace_owners: Replace existing badge owners
```
```suggestion
          replace_owners: Remove the badge from previous owners
```

PR #8772, opened 2020-01-23
FEATURE: Improving bookmarks part 2 -- Lists and Topics

martin-brennan:
This PR makes quite a few changes, but it can be safely merged because all the changes are gated between the `enable_bookmarks_with_reminders` setting. If it is off none of this new stuff will happen.

The aim of this PR is to get the new bookmarks feature in a state where we can turn the setting on and it just works as a replacement for the old bookmarks, and then we can test in dev and meta. Then it's just a matter of adding the background reminder functionality.

### Testing

* Enable `SiteSetting.enable_bookmarks_with_reminders`.
* Test topic-level bookmarking and interaction with post bookmarks & clearing all bookmarks for topic.
* Go to user Activity tab and click Bookmarks, make sure it goes to the new list.
* Test deleting bookmarks from the new list and that clicking the link opens the correct post. Also check the name of the bookmark (if given) shows here.

### Main Changes

* Added a rake task to sync the old PostAction bookmarks to the new Bookmark table, which can be run as many times as we want for a site (it will not create duplicates).
* Changed the bookmarks topic filter at /bookmarks to load new bookmark records. This involved changing several topic serializers. This will show multiple records per topic because each bookmarked post can have its own reminder settings and name. Bookmarks can be deleted from here from the context menu. The icon changes based on whether the bookmark has a reminder attached. I had to make my own bookmark-topic-list-item.hbs template here to do the customization I needed; if there is a better way I am all ears.
* Disabled selecting the "reminder type" for bookmarks in the UI because the backend functionality is not done yet (of sending users notifications etc.)
* Hook up topic-level bookmarking to work with the new Bookmark table.

image

### Not Addressed

I will do the following in another PR:

* I have not yet changed the reports for bookmarks at /admin/reports/bookmarks.
* I have not made the topic-level bookmarks button trigger the new reminder modal.


app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,30 @@
+import DropdownSelectBoxComponent from "select-kit/components/dropdown-select-box";
+
+export default DropdownSelectBoxComponent.extend({
+  classNames: ["bookmark-actions-dropdown"],
+  headerIcon: null,
+  title: "...",
+  allowInitialValueMutation: false,
```
jjaffeux:
```suggestion
```

app/assets/javascripts/discourse/components/bookmark-topic-list-item.es6
```diff
@@ -0,0 +1,16 @@
+import TopicListItem from "discourse/components/topic-list-item";
+import Bookmark from "discourse/models/bookmark";
+import discourseComputed from "discourse-common/utils/decorators";
+
+export default TopicListItem.extend({
+  @discourseComputed("topic.bookmarked_post_numbers")
```
jjaffeux: can probably rewrite all of this with:

```
bookmakedPostNumber: readOnly("topic.bookmarked_post_numbers.firstObject")
```

app/assets/javascripts/discourse/components/bookmark-topic-list-item.es6
```diff
@@ -0,0 +1,16 @@
...
+  bookmakedPostNumber(bookmarked_post_numbers) {
+    return bookmarked_post_numbers[0];
+  },
+  actions: {
+    removeBookmark(id) {
+      let bookmark = new Bookmark({ id: id });
```
jjaffeux:
```suggestion
      let bookmark = new Bookmark({ id });
```

ZogStriP:
Can you ping me once it's good for review?

martin-brennan:
@ZogStriP yes for sure, Joffrey is going to review as well.

app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,31 @@
+import DropdownSelectBoxComponent from "select-kit/components/dropdown-select-box";
+
+export default DropdownSelectBoxComponent.extend({
+  classNames: ["bookmark-actions-dropdown"],
+  headerIcon: null,
+  title: "...",
+  showFullTitle: true,
+
```
jjaffeux: should now be:

```
selectKitOptions: {
  icon: null,
  translatedNone: "...",
  showFullTitle: true
}
```

app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,31 @@
+import DropdownSelectBoxComponent from "select-kit/components/dropdown-select-box";
+
+export default DropdownSelectBoxComponent.extend({
+  classNames: ["bookmark-actions-dropdown"],
```
jjaffeux: need a pluginApiIdentifier:

```
pluginApiIdentifiers: ["bookmark-actions-dropdown"]
```

app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,31 @@
...
+  headerIcon: null,
+  title: "...",
+  showFullTitle: true,
+
+  computeContent() {
```
jjaffeux: should be now be:

```
content: computed(function() { ...
```

app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,31 @@
...
+        name: I18n.t("post.bookmarks.actions.delete_bookmark.name"),
+        description: I18n.t(
+          "post.bookmarks.actions.delete_bookmark.description"
+        )
+      }
+    ];
+  },
+
+  actions: {
+    onSelect(id) {
```
jjaffeux:
```suggestion
    onChange(id) {
```

app/assets/javascripts/discourse/models/bookmark.js.es6
```diff
@@ -0,0 +1,24 @@
+import discourseComputed from "discourse-common/utils/decorators";
+import { none } from "@ember/object/computed";
+import { ajax } from "discourse/lib/ajax";
+import { Promise } from "rsvp";
+import RestModel from "discourse/models/rest";
+
+const Bookmark = RestModel.extend({
+  newBookmark: none("id"),
+
+  @discourseComputed
```
jjaffeux: I would prefer computed here, discourseComputed should only be used when you have params, as its main advantage is to provide computed properties as params.

app/assets/javascripts/discourse/models/post.js.es6
```diff
@@ -353,14 +353,18 @@ const Post = RestModel.extend({
           this.appEvents.trigger("post-stream:refresh", { id: this.id });
         },
         afterSave: reminderAtISO => {
+          this.set("topic.bookmarked", true);
```
jjaffeux: group both line in setProperties

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -106,8 +106,8 @@ const Topic = RestModel.extend({
     return users;
   },
 
-  @discourseComputed("fancy_title")
-  fancyTitle(title) {
+  @discourseComputed("fancy_title", "bookmark_name")
+  fancyTitle(title, bookmark_name) {
```
jjaffeux: should be bookmarkName in js

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -117,6 +117,7 @@ const Topic = RestModel.extend({
       const titleDir = isRTL(title) ? "rtl" : "ltr";
       return `<span dir="${titleDir}">${fancyTitle}</span>`;
     }
+    fancyTitle = fancyTitle + (bookmark_name ? " (" + bookmark_name + ")" : "");
```
jjaffeux: not very readable, might need intermediate variable

app/assets/javascripts/discourse/models/topic.js.es6
```diff
@@ -411,6 +415,13 @@ const Topic = RestModel.extend({
                 post.set("bookmarked", false);
                 updated.push(post.get("id"));
               }
+              if (
+                this.siteSettings.enable_bookmarks_with_reminders &&
+                post.get("bookmarked_with_reminder")
+              ) {
+                post.set("bookmarked_with_reminder", false);
+                updated.push(post.get("id"));
```
jjaffeux:
```suggestion
                updated.push(post.id);
```

jjaffeux:
I feel like the debug preload data thing and store map additions should be in separate PRs

jjaffeux:
Also we would need tests for this store map id and and at least one acceptance test for bookmarks

app/assets/javascripts/discourse/components/topic-list.js.es6
```diff
@@ -31,6 +33,22 @@ export default Component.extend(LoadMore, {
...
+  showBookmarkListItem() {
+    return (
+      this.get("router.currentRoute.name") === "discovery.bookmarks" &&
+      this.siteSettings.enable_bookmarks_with_reminders
+    );
+  },
+
+  @discourseComputed("showBookmarkListItem")
+  listItemComponentName(showBookmarkListItem) {
+    if (showBookmarkListItem) {
```
ZogStriP: This could be a one-liner

```js
return showBookmarkListItem ? "bookmark-topic-list-item" : "topic-list-item";
```

app/assets/javascripts/discourse/models/topic.js.es6
ZogStriP: How about this?

```js
if (bookmark_name) fancyTitle += ` (${bookmark_name})`;
```

app/serializers/listable_topic_serializer.rb
```diff
@@ -45,7 +45,11 @@ def liked
   end
 
   def bookmarked
-    object.user_data && object.user_data.bookmarked
+    if SiteSetting.enable_bookmarks_with_reminders?
+      object.bookmark_ids.any?
+    else
+      object.user_data && object.user_data.bookmarked
```
ZogStriP:
```suggestion
      object.user_data&.bookmarked
```

lib/guardian/bookmark_guardian.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module BookmarkGuardian
+  def can_delete_bookmark?(bookmark)
```
ZogStriP: Do we want staff/admins to be able to delete another user's bookmarks?

martin-brennan:
Thanks @jjaffeux and @ZogStriP I'll address these today, and put the store & debug preload data thing in separate PRs.

lib/guardian/bookmark_guardian.rb
martin-brennan: Hmm good question...now that the Bookmarks link takes you to /bookmarks, which only shows your own bookmarks, I don't think there is a situation in which the admin could even _see_ the user's bookmarks. I wouldn't think it would be necessary for admin to mess with them?

app/assets/javascripts/discourse/models/topic-list.js.es6
```diff
@@ -174,6 +174,14 @@ TopicList.reopenClass({
         t.participants.forEach(p => (p.user = users[p.user_id]));
       }
 
+      // the same topic can show multiple times in the bookmark
+      // list because each post bookmark can have a different
+      // name + reminder setting, so we use the bookmark_id to make
+      // the topic more unique
+      if (result.filter === "bookmarks" && result.siteSettings.enable_bookmarks_with_reminders) {
```
martin-brennan: Paging @eviltrout, this is what I came up with in place of changing store, do you think this is a better place for this?

eviltrout: I don't think this will work either, unfortunately. This code seems to be replacing the topic id with the bookmark id. It's still using the same identity map, so now another piece of code that uses the same id as the bookmark id is getting back something strange.

I think a better solution here would be to make a new route to list these bookmarks that doesn't use topic list. It would return a list of bookmarks, with each bookmark having a topic id that side loads to a topic object.

martin-brennan:
Closing this PR to go with the new approach of replacing the user activity bookmarks route instead of messing with TopicList and TopicQuery per discussions with Sam

PR #8773, opened 2020-01-23
UX: Invalid CSV error message now includes information about the malformed line

romanrizzi:


app/controllers/admin/badges_controller.rb
```diff
@@ -44,11 +44,13 @@ def mass_award
     raise Discourse::InvalidParameters if csv_file.try(:tempfile).nil? || badge.nil?
 
     batch_number = 1
+    lineno = 1
```
ZogStriP: To be consistent with `batch_number`, I would have used `line_number` instead of ` lineno`.

config/locales/server.en.yml
```diff
@@ -4134,7 +4134,7 @@ en:
   badges:
     mass_award:
       errors:
-        invalid_csv: We couldn't read that CSV file. Please confirm it has one email per line.
+        invalid_csv: We encountered an error in line %{lineno}. Please confirm the CSV has one email per line.
```
```suggestion
        invalid_csv: We encountered an error on line %{lineno}. Please confirm the CSV has one email per line.
```

PR #8780, opened 2020-01-24
DEV: Ignore `ls` errors when clearing FileStore cache

CvX:
A race condition issue is possible when multiple thread/processes are calling this method.
`ls` prints out to stderr "cannot access '...': No such file or directory" if any of the files it's currently trying to list are being removed by the `xargs rm -rf` in an another process. That doesn't affect the result, but it did raise an error before this change.

Tested on a production instance where the original issue was observed. This solution has @davidtaylorhq's blessing 😉

### DIFF
--- a/lib/file_store/base_store.rb
+++ b/lib/file_store/base_store.rb
@@ -164,7 +164,7 @@ module FileStore
       # Exit status `1` in `ls` occurs when e.g. "listing a directory
       # in which entries are actively being removed or renamed".
       # It's safe to ignore it here.
-      unless ls.exitstatus.in?([0, 1]) && processes.all?(&:success?)
+      if ![0, 1].include?(ls.exitstatus) || !processes.all?(&:success?)
         raise "Error clearing old cache"
       end
     end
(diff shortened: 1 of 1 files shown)


lib/file_store/base_store.rb
```diff
@@ -150,14 +150,23 @@ def cache_file(file, filename)
...
         "xargs rm -f"
       )
-      raise "Error clearing old cache" if !processes.all?(&:success?)
+
+      ls = processes.shift
+
+      # Exit status `1` in `ls` occurs when e.g. "listing a directory
+      # in which entries are actively being removed or renamed".
+      # It's safe to ignore it here.
+      unless ls.exitstatus.in?([0, 1]) && processes.all?(&:success?)
```
ZogStriP: Might be easier to understand when using `if` instead of `unless`?

```suggestion
      if ![0, 1].include?(ls.exitstatus) || !processes.all?(&:success?)
```

CvX: Do the two negations have a lower cognitive cost than `unless`? I don't know 😅 I'm ok with either one.

But why not `.in?`? 😛 I like how it reads naturally, similar to usual equality comparisons:

```rb
status == 1
status.in?([1, 2])
```

vs the reversed order:

```rb
1 == status
[1, 2].include?(status)
```

ZogStriP: `unless` feels always harder to understand when there’s more than 1 condition.

As for the not in? it’s just an habit I have to always put the “constants” on the left of a condition to prevent typoing = instead of ==.

Not really concerned about that though ;)

PR #8782, opened 2020-01-24
Ember linting

markvanlan:
This adds the new Ember linting found in this repo - https://github.com/discourse/eslint-discourse-ember


markvanlan:
@eviltrout This is rebased onto the select-kit changes now in master. Good to commit?

eviltrout:
Let's give Select Kit one more day just in case it has to be rolled back or anything.

PR #8783, opened 2020-01-24
FEATURE: Use Contact Picker API for invites

xfalcox:
output

### DIFF
--- a/app/assets/javascripts/discourse/components/invite-panel.js.es6
+++ b/app/assets/javascripts/discourse/components/invite-panel.js.es6
@@ -1,5 +1,6 @@
 import discourseComputed from "discourse-common/utils/decorators";
 import { isEmpty } from "@ember/utils";
+import { computed } from "@ember/object";
 import { alias, and, equal } from "@ember/object/computed";
 import EmberObject from "@ember/object";
 import Component from "@ember/component";
@@ -7,6 +8,7 @@ import { emailValid } from "discourse/lib/utilities";
 import Group from "discourse/models/group";
 import Invite from "discourse/models/invite";
 import { i18n } from "discourse/lib/computed";
+import { getNativeContact } from "discourse/lib/pwa-utils";
 
 export default Component.extend({
   tagName: null,
@@ -179,6 +181,10 @@ export default Component.extend({
       (emailValid(emailOrUsername) || isPrivateTopic || !invitingToTopic)
     );
   },
+
+  showContactPicker: computed(function() {
+    return this.capabilities.hasContactPicker;
+  }),
 
   @discourseComputed("emailOrUsername")
   showCustomMessage(emailOrUsername) {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/components/invite-panel.js.es6
```diff
@@ -180,6 +182,10 @@ export default Component.extend({
     );
   },
 
+  showContactPicker: Ember.computed(function() {
```
jjaffeux: would be better to use `computed` directly here, didn't do it when I gave you this example.

```
import { computed } from "@ember/object";

showContactPicker: computed(function() {
  return this.capabilities.hasContactPicker;
});
```

app/assets/javascripts/discourse/components/invite-panel.js.es6
```diff
@@ -433,6 +439,13 @@ export default Component.extend({
       } else {
         this.set("customMessage", null);
       }
+    },
+
+    searchContact() {
+      getNativeContact(['email'], false)
+        .then(result => {
+          this.setProperties({emailOrUsername: result[0].email[0]});
```
jjaffeux:
```suggestion
          this.set("emailOrUsername", result[0].email[0]);
```

app/assets/javascripts/discourse/lib/pwa-utils.js.es6
```diff
@@ -26,3 +26,20 @@ export function nativeShare(data) {
...
+  return new Promise((resolve, reject) => {
+    if (!caps.hasContactPicker) {
+      reject();
+      return;
+    }
+
+    navigator.contacts
+      .select(properties, {multiple: multiple})
+      .then(resolve)
+      .catch(() => {
```
jjaffeux:
```suggestion
      .catch(reject)
```

jjaffeux:
```
 .catch(reject)
```

app/assets/javascripts/discourse/lib/pwa-utils.js.es6
```diff
@@ -26,3 +26,20 @@ export function nativeShare(data) {
...
+export function getNativeContact(properties, multiple) {
+  const caps = Discourse.__container__.lookup("capabilities:main");
+  return new Promise((resolve, reject) => {
+    if (!caps.hasContactPicker) {
+      reject();
+      return;
+    }
+
+    navigator.contacts
+      .select(properties, {multiple: multiple})
```
jjaffeux:
```suggestion
      .select(properties, { multiple })
```

xfalcox:
> This looks quite cool but the button is definitely the wrong size there. Can it be customized?

Moved it. Also asked @awesomerobot  to take a look last week, as this UI using input group is not something we do in Discourse, but I couldn't come with a better one.

PR #8784, opened 2020-01-24
FEATURE: allow searching public topics and private messages simultaneously

adqm:
As discussed in [this thread](https://meta.discourse.org/t/search-public-topics-and-pms-simultaneously), this PR is a first attempt at allowing users to use a single query to search both public and private messages.

I've done a little bit of testing on my own machine and things kind of seem to work (also added a few little unit test cases), but I'm not sure I quite grok the way everything is hooked up, so it's possible I missed something.

Feedback and criticism are welcome!  I'm happy to discuss, and to modify as necessary.

### DIFF
--- a/lib/search.rb
+++ b/lib/search.rb
@@ -28,7 +28,7 @@ class Search
   end
 
   def self.facets
-    %w(topic category user private_messages tags)
+    %w(topic category user private_messages tags all_topics)
   end
 
   def self.ts_config(locale = SiteSetting.default_locale)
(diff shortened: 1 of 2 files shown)


davidtaylorhq:
Thanks for this @adqm. I just made a couple of small rearrangements.

Will merge tomorrow

adqm:
Woo-hoo, thanks!  And agreed that your changes are a big improvement.

PR #8789, opened 2020-01-27
PERF: Use a separate route for user cards, and split user serializer

davidtaylorhq:
This change is behind a hidden site setting, so we can work on plugin interfaces, and roll out slowly. The setting will be removed in the near future.

### DIFF
--- a/app/assets/javascripts/discourse/components/user-card-contents.js.es6
+++ b/app/assets/javascripts/discourse/components/user-card-contents.js.es6
@@ -140,13 +140,11 @@ export default Component.extend(CardContentsBase, CanCheckEmails, CleansUp, {
     this._positionCard($target);
     this.setProperties({ visible: true, loading: true });
 
-    const args = {};
+    const args = {
+      forCard: this.siteSettings.enable_new_user_card_route,
+      include_post_count_for: this.get("topic.id")
+    };
 
-    if (this.siteSettings.enable_new_user_card_route) {
-      args.forCard = true;
-    }
-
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/components/user-card-contents.js.es6
```diff
@@ -140,7 +140,12 @@ export default Component.extend(CardContentsBase, CanCheckEmails, CleansUp, {
     this._positionCard($target);
     this.setProperties({ visible: true, loading: true });
 
-    const args = { stats: false };
+    const args = {};
+
+    if (this.siteSettings.enable_new_user_card_route) {
```
eviltrout: Minor but in Javascript it's generally a good idea for performance to keep all objects the same shape. It would be preferred to do `args.forCard = this.siteSettings.enable_new_user_card_route`. Use `false` over `undefined`.

spec/requests/users_controller_spec.rb
```diff
@@ -2943,6 +2943,57 @@ def post_user
...
+  describe "#show_card" do
+    context "anon" do
+      let(:user) { Discourse.system_user }
+
+      it "returns success" do
+        get "/u/#{user.username}/card.json"
+        expect(response.status).to eq(200)
+        parsed = JSON.parse(response.body)["user"]
+
+        expect(parsed['username']).to eq(user.username)
```
eviltrout: Super super super minor but why use single quotes for some attributes and double for others?

PR #8792, opened 2020-01-27
FIX: Post reviser picking up edits for hidden posts

markvanlan:


markvanlan:
> No tests?

Should have made this a draft PR; tests coming.

markvanlan:
@eviltrout I added a test. The test does fail when `remove_flags_and_unhide_post` is called where is was previously.

The issue is that the `modification` column is set by using 
```
@post.previous_changes
```
and when the post is hidden, the previous_changes were being overwritten.

spec/components/post_revisor_spec.rb
```diff
@@ -227,6 +227,18 @@
       end
     end
 
+    describe 'hidden post' do
+      it "created stores the modification value" do
```
eviltrout:
```suggestion
      it "correctly stores the modification value" do
```

PR #8795, opened 2020-01-27
FIX: set showFooter to true on group topic list

awesomerobot:
If using `{{#if showFooter}}` in a template, `showFooter` is never set to `true` on a group's `g/groupname/activity/topics` route (it's correctly set on other group routes like `group-activity-posts`)

Not sure if this is the best way to do it here, but it's also what we do for some of the user activity routes.

### DIFF
--- a/app/assets/javascripts/discourse/routes/group-activity-topics.js.es6
+++ b/app/assets/javascripts/discourse/routes/group-activity-topics.js.es6
@@ -1,6 +1,8 @@
 import DiscourseRoute from "discourse/routes/discourse";
 
 export default DiscourseRoute.extend({
+  showFooter: true,
+
   titleToken() {
     return I18n.t(`groups.topics`);
   },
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/routes/group-activity-topics.js.es6
```diff
@@ -9,5 +9,12 @@ export default DiscourseRoute.extend({
     return this.store.findFiltered("topicList", {
       filter: `topics/groups/${this.modelFor("group").get("name")}`
     });
+  },
+
+  actions: {
+    didTransition() {
+      this.controllerFor("application").set("showFooter", true);
```
markvanlan: It looks like you can set showFooter to true in a more straightforward manor like this:
```javascript
export default DiscourseRoute.extend({
  showFooter: true,
})
```

That is the pattern in these files just to mention a few
```
./app/assets/javascripts/discourse/routes/group-manage-membership.js.es6:  showFooter: true,
./app/assets/javascripts/discourse/routes/preferences-tags.js.es6:  showFooter: true
./app/assets/javascripts/discourse/routes/preferences-emails.js.es6:  showFooter: true
./app/assets/javascripts/discourse/routes/tag-groups-new.js.es6:  showFooter: true
```

markvanlan:
@awesomerobot I just approved this, but I wanted to make sure that you saw you made my suggested change on a different file than the original one.

PR #8802, opened 2020-01-28
FIX: Mitigate issue where legacy pre-secure hotlinked media would not be redownloaded

martin-brennan:
Basically, say you had already downloaded a certain image from a certain URL
using pull_hotlinked_images and the onebox. The upload would be stored
by its sha as an upload record. Whenever you linked to the same URL again
in a post (e.g. in our case an og:image on review.discourse) we would
would reuse the original upload record because of the sha1.

However when you turned on secure media this could cause problems as
the first post that uses that upload after secure media is enabled
will set the access control post for the upload to the new post.
Then if the post is deleted every single onebox/link to that same image
URL will fail forever with 403 as the secure-media-uploads URL fails
if the access control post has been deleted.

To fix this when cooking posts and pulling hotlinked images, we only
allow using an original upload by URL if its access control post
matches the current post, and if the original_sha1 is filled in,
meaning it was uploaded AFTER secure media was enabled. otherwise
we just redownload the media again to be safe, as the URL will always
be new then.

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -330,7 +330,15 @@ class CookedPostProcessor
       img["height"] = height
     end
 
-    if upload = Upload.get_from_url(src)
+    upload = Upload.get_from_url(src)
+
+    # if the upload already exists and is attached to a different post,
+    # or the original_sha1 is missing meaning it was created before secure
+    # media was enabled. we want to re-thumbnail and re-optimize in this case
+    # to avoid using uploads linked to many other posts
+    upload = Upload.invalid_secure_upload_reuse?(upload, @post) ? nil : upload
+
+    if upload.present?
       upload.create_thumbnail!(width, height, crop: crop)
 
       each_responsive_ratio do |ratio|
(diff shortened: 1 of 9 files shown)


lib/cooked_post_processor.rb
```diff
@@ -330,7 +330,15 @@ def convert_to_link!(img)
...
     end
 
-    if upload = Upload.get_from_url(src)
+    upload = Upload.get_from_url(src)
+
+    # if the upload already exists and is attached to a different post,
+    # or the original_sha1 is missing meaning it was created before secure
+    # media was enabled. we want to re-thumbnail and re-optimize in this case
+    # to avoid using uploads linked to many other posts
+    upload = Upload.invalid_secure_upload_reuse?(upload, @post) ? nil : upload
```
eviltrout: It seems in both places in the PR where this new function is used, you also use a ternary operator to consider changing the upload to nil. 

Why not have the method return an `upload` or `nil`? Something like:

`upload = Upload.consider_for_reuse(upload, @post)`

martin-brennan: That is very true...will change!

PR #8804, opened 2020-01-28
UX: Introduce automatic 'categories topics' setting

udan11:
When 'categories topics' setting is set to 0, the system will
automatically try to find a value to keep the two columns (categories
and topics) symmetrical.

The value is computed as 1.5x the number of top level categories.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -8,6 +8,9 @@ class CategoriesController < ApplicationController
   before_action :initialize_staff_action_logger, only: [:create, :update, :destroy]
   skip_before_action :check_xhr, only: [:index, :categories_and_latest, :categories_and_top, :redirect]
 
+  SYMMETRICAL_CATEGORIES_TO_TOPICS_FACTOR = 1.5
+  MIN_CATEGORIES_TOPICS = 5
+
   def redirect
     return if handle_permalink("/category/#{params[:path]}")
     redirect_to path("/c/#{params[:path]}")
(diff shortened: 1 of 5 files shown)


app/controllers/categories_controller.rb
```diff
@@ -226,6 +229,18 @@ def find_by_slug
   end
 
   private
+
+  def self.topics_per_page
+    if SiteSetting.categories_topics > 0
+      categories_count = SiteSetting.categories_topics
```
ZogStriP: I think if the user set it to > 0 and < 5, it's on them. We should

```ruby
def self.topics_per_page
  return SiteSetting.categories_topics if SiteSetting.categories_topics > 0
  ...
end
```

config/locales/server.en.yml
```diff
@@ -1686,7 +1686,7 @@ en:
     alert_admins_if_errors_per_minute: "Number of errors per minute in order to trigger an admin alert. A value of 0 disables this feature. NOTE: requires restart."
     alert_admins_if_errors_per_hour: "Number of errors per hour in order to trigger an admin alert. A value of 0 disables this feature. NOTE: requires restart."
 
-    categories_topics: "Number of topics to show in /categories page."
+    categories_topics: "Number of topics to show in /categories page. If set to 0, it will automatically try to find a value to keep the two columns symmetrical (categories and topics). At least 5 topics will always be returned."
```
ZogStriP: Not sure about the "at least 5 topics will always be returned" part. We should definitely have a minimum when it's "automatic", but if the user wants less topics, they should be able to set it

PR #8808, opened 2020-01-29
FIX: Properly convert quotes to Markdown

nbianca:
When quoting a quote it used to convert the quote header, including the user avatar and username, into an image and some text and then the contents. This also caused issues when quoting full paragraphs (or when selecting paragraphs by triple-clicking) because the user avatar and name from the following quote would also be included.

This commit implements the support necessary to convert `<aside class="quote">` elements to proper Discourse quotes.

### DIFF
--- a/app/assets/javascripts/discourse/lib/to-markdown.js.es6
+++ b/app/assets/javascripts/discourse/lib/to-markdown.js.es6
@@ -44,7 +44,6 @@ export class Tag {
     return [
       "address",
       "article",
-      "aside",
       "dd",
       "div",
       "dl",
@@ -89,6 +88,7 @@ export class Tag {
       ...Tag.blocks(),
       ...Tag.headings(),
       ...Tag.slices(),
+      "aside",
       "li",
       "td",
       "th",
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -126,6 +126,53 @@ export class Tag {
...
+        if (this.name === "p" && parent && parent.name === "li") {
+          // fix for google docs
+          this.gap = "";
+        }
+
+        return `${this.gap}${this.prefix}${text}${this.suffix}${this.gap}`;
+      }
+
+      toMarkdown() {
+        if (!this.element.attributes.class.split(" ").includes("quote")) {
```
ZogStriP: Might be more efficient to use a regex?

```javascript
if (!/\bquote\b/.test(this.element.attributes.class))
```

app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -126,6 +126,53 @@ export class Tag {
...
+      constructor() {
+        super("aside");
+        this.gap = "\n\n";
+      }
+
+      decorate(text) {
+        const parent = this.element.parent;
+
+        if (this.name === "p" && parent && parent.name === "li") {
+          // fix for google docs
```
ZogStriP: What's the issue with Google Docs?

app/assets/javascripts/discourse/lib/to-markdown.js.es6
```diff
@@ -126,6 +126,53 @@ export class Tag {
...
+
+        return `${this.gap}${this.prefix}${text}${this.suffix}${this.gap}`;
+      }
+
+      toMarkdown() {
+        if (!this.element.attributes.class.split(" ").includes("quote")) {
+          return super.toMarkdown();
+        }
+
+        const blockquote = this.element.children.find(
```
ZogStriP: Is is possible that there are no `blockquote` as children of the `aside`?

app/assets/javascripts/discourse/lib/to-markdown.js.es6
nbianca: I do not know. I needed to extend the class that is returned by `Tag.block('aside')` and did not realize that I could call that method with `extend`, so I just copied that class and modified it.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
nbianca: Thanks, that seems to be 3x faster.

app/assets/javascripts/discourse/lib/to-markdown.js.es6
nbianca: Yes, that is a possibility. Added a check.

PR #8809, opened 2020-01-29
FEATURE: Users can feature any public topic on his/her profile

markvanlan:


spec/requests/users_controller_spec.rb
```diff
@@ -3937,13 +3931,19 @@ def revoke(user, skip_remote: false)
...
 
-      it 'sets the user_profiles featured_topic correctly' do
+      it 'sets featured_topic correctly for user created topic' do
         sign_in(user)
         topic.update(user_id: user.id)
         put "/u/#{user.username}/feature-topic.json", params: { topic_id: topic.id }
         expect(response.status).to eq(200)
         expect(user.user_profile.featured_topic).to eq topic
       end
+      it 'sets featured_topic correctly for non-user-created topic' do
```
ZogStriP: Might want an empty line between the two specs.

markvanlan: @ZogStriP Agreed, good catch.

PR #8811, opened 2020-01-29
FEATURE: Send suspect users to the review queue

romanrizzi:


app/jobs/scheduled/enqueue_suspect_users.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Jobs
+  class EnqueueSuspectUsers < ::Jobs::Scheduled
+    every 2.hours
+
+    def execute(_args)
+      return unless SiteSetting.approve_suspect_users
+
+      AdminUserIndexQuery.new(query: 'suspect').find_users(10).each do |user|
```
eviltrout: Does this mean it'll run against the entire suspect list over and over? Do those entries ever get cleaned up or over time will this just grow and grow?

PR #8813, opened 2020-01-29
UX: Show vertical topic timeline while composing

pmusaraj:
When enough vertical space is available, we should show the vertical topic timeline, it provides easier access to navigating within the topic while replying. (This is particularly useful on long topics.)

The change is best described using before/after videos. 

**Before**
current-timeline

**After**
proposed-timeline


pmusaraj:
Yes, agreed. It's probably not too tricky to infer from a calculation of window height minus (composer height plus header height plus some padding). I'll see if I can do that now, while we're at it.

eviltrout:
Awesome, I always prefer that.

CvX:
@pmusaraj I found a bug that seems to originate in this PR:

timeline2

PR #8816, opened 2020-01-30
FEATURE: Add logout functionality to SSO Provider protocol

mdoggydog:
This commit adds support for an optional "logout" parameter in the
payload of the /session/sso_provider endpoint.  If an SSO Consumer
adds a "logout=true" parameter to the encoded/signed "sso" payload,
then Discourse will treat the request as a logout request instead
of an authentication request.  The logout flow works something like
this:

 * User requests logout at SSO-Consumer site (e.g., clicks "Log me out!"
   on web browser).
 * SSO-Consumer site does whatever it does to destroy User's session on
   the SSO-Consumer site.
 * SSO-Consumer then redirects browser to the Discourse sso_provider
   endpoint, with a signed request bearing "logout=true" in addition
   to the usual nonce and the "return_sso_url".
 * Discourse destroys User's discourse session and redirects browser back
   to the "return_sso_url".
 * SSO-Consumer site does whatever it does --- notably, it cannot request
   SSO credentials from Discourse without the User being prompted to login
   again.

### DIFF
--- a/app/controllers/session_controller.rb
+++ b/app/controllers/session_controller.rb
@@ -54,6 +54,12 @@ class SessionController < ApplicationController
         return
       end
 
+      if sso.logout
+        params[:return_url] = sso.return_sso_url
+        destroy
+        return
+      end
+
       if current_user
         sso.name = current_user.name
         sso.username = current_user.username
--- a/lib/single_sign_on.rb
+++ b/lib/single_sign_on.rb
@@ -16,6 +16,7 @@ class SingleSignOn
     groups
     locale
     locale_force_update
+    logout
     name
     nonce
     profile_background_url
@@ -34,6 +35,7 @@ class SingleSignOn
     admin
     avatar_force_update
     locale_force_update
+    logout
     moderator
     require_activation
     suppress_welcome_message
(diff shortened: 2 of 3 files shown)


mdoggydog:
> This is going to need a test or two to confirm it works. 

Does this mean adding something to `spec/requests/session_controller_spec.rb`?  (I haven't found anything in the contribution guides which discusses how/when testing should work, and this is literally the first ruby code I have ever written.  If there is some standard testing methodology that is obvious to seasoned ruby/rails/etc coders, I apologize for my naivete.)

> Also, it might be a good idea to put this new behaviour behind a site setting.

To clarify:
 * Would a single site-wide setting be sufficient (versus a per-SSO-consumer parameter, like the shared-secret)?
 * What should be the response to a `logout=true` request if logout behavior is disabled?  I'm imagining returning a 404 just as if `enable_sso_provider` is turned off --- because it is a problem if the Discourse and the SSO-Consumer configurations disagree on whether or not logout is allowed, and the result should be noticeable.
 * When/why do you think a site would want to disable this behavior (in particular, at the Discourse SSO-Provider side, instead of at the SSO-Consumer side)?  (E.g., I'd rather avoid adding another site setting until there is a plausible need for it.)

mdoggydog:
@eviltrout, I have added a couple of tests to exercise the logout functionality.  

I did not add a site-setting, though --- I really can't imagine any practical scenario where someone would need to disable the logout behavior at the Discourse SSO provider.

eviltrout:
Thanks for the test. My big concern here was security. However, it does seem safe because logout is only exposed if the other site is trusted. Thanks!

mdoggydog:
You're welcome, and thanks!

PR #8819, opened 2020-01-30
FIX: Prevent race condition when post processing post

nbianca:
If a post is being cooked twice (for example after an edit), there is a chance the 'raw' and 'cooked' column to be inconsistent. This reduces the chances of that happening.

### DIFF
--- a/app/jobs/regular/process_post.rb
+++ b/app/jobs/regular/process_post.rb
@@ -7,43 +7,45 @@ module Jobs
   class ProcessPost < ::Jobs::Base
 
     def execute(args)
-      post = Post.find_by(id: args[:post_id])
-      # two levels of deletion
-      return unless post.present? && post.topic.present?
+      DistributedMutex.synchronize("process_post_#{args[:post_id]}") do
+        post = Post.find_by(id: args[:post_id])
+        # two levels of deletion
+        return unless post.present? && post.topic.present?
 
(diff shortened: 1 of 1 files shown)


app/jobs/regular/process_post.rb
```diff
@@ -33,7 +33,14 @@ def execute(args)
...
         else
-          post.update_column(:cooked, cp.html)
+          # Hack to prevent race conditions. If a post is cooked, edited and
+          # quickly rebaked, there is a chance that the first rebake will
+          # finish last. This leads to inconsistencies where the 'cooked'
+          # column contains the old baked 'raw' text.
+          Post
+            .where(id: post.id, raw: post.raw)
+            .update_all(cooked: cp.html)
+          post.reload
```
eviltrout: `post.reload` will `SELECT` it again from the database. Could we instead do `post.cooked = cp.html` since we know that's the only attribute that changed? You might also need to clear the dirty active record attributes after you do that because we know the value has been saved.

nbianca: That is how I wanted to implement this and it seemed safer that way. Having another look now, it really looks better. 😊

ZogStriP:
I wonder, should we add a mutex around that job to ensure it's never editing the same post at the same time?

nbianca:
Yes, that's a better idea. 😊

PR #8824, opened 2020-01-30
FIX: We don't want to update the post read count and user stats if the post timing wasn't created due to a conflict.

romanrizzi:


app/models/post_timing.rb
```diff
@@ -31,12 +31,14 @@ def self.pretend_read(topic_id, actual_read_post_number, pretend_read_post_numbe
...
   def self.record_new_timing(args)
-    DB.exec("INSERT INTO post_timings (topic_id, user_id, post_number, msecs)
+    timings_created = DB.exec("INSERT INTO post_timings (topic_id, user_id, post_number, msecs)
               SELECT :topic_id, :user_id, :post_number, :msecs
               ON CONFLICT DO NOTHING",
             args)
+
     # concurrency is hard, we are not running serialized so this can possibly
     # still happen, if it happens we just don't care, its an invalid record anyway
+    return if timings_created.zero?
```
SamSaffron: `.zero?` feels a bit fancy to me... `return if row_count == 0` the less fancy option is my pref

I know it is splitting hairs here, and totally a bit of a dinosaur.

Anyway feel free to keep this if you much prefer it... or not... up to you. 

Also feel free to merge, change looks very safe to me.

PR #8828, opened 2020-01-31
FIX: Opening new topic draft never creates reply post

markvanlan:
This PR addresses [this issue](https://meta.discourse.org/t/duplicating-a-tab-breaks-a-change-from-a-quote-to-create-a-new-topic/140243?u=markvanlan), where a reply post is being created rather than a new topic.

### DIFF
--- a/app/assets/javascripts/discourse/models/composer.js.es6
+++ b/app/assets/javascripts/discourse/models/composer.js.es6
@@ -912,6 +912,10 @@ const Composer = RestModel.extend({
   },
 
   createPost(opts) {
+    if (this.action === CREATE_TOPIC) {
+      this.set("topic", null);
+    }
+
     const post = this.post;
     const topic = this.topic;
     const user = this.user;


markvanlan:
@eviltrout I am trying to think of the best way to test this, if one is needed. 

topic_id was being sent the server on weird occations when the composer action is `createTopic`. This was creating reply posts instead of a new topic. Only happened when refreshing page with a "reply in new topic" draft was open.

ZogStriP:
🤔 Feels like adding a band-aid... 

> topic_id was being sent the server on **weird occations**

Have you managed to find out when we're adding the `topic_id` when we should not?

markvanlan:
@ZogStriP Yeah, it is in the link in my first comment (in a meta topic).

Its when the composer is open inside a topic, the composer is switched to `reply as linked topic`, and then the tab is duplicated and opened in a new tab. If you hit create, the `topic_id` is present.

This seems like a future-proof solution though. There is no case when creating a topic that you would want to pass the topic_id to the server

ZogStriP:
> Its when the composer is open inside a topic, the composer is switched to `reply as linked topic`, and then the tab is duplicated and opened in a new tab. If you hit create, the `topic_id` is present.

Should we fix that then? When switching the composer to "reply as linked topic", nullify the `topic_id`.
It does mean we need to re-set the `topic_id` when the composer is switched back to "reply"... 🤔 

> This seems like a future-proof solution though. There is no case when creating a topic that you would want to pass the topic_id to the server

Definitely. "band-aid" might not have been the best word to use. I meant that we were not fixing the root-cause.

PR #8834, opened 2020-02-01
FIX: Recovered posts with no user will be taken over by system user

udan11:


lib/post_destroyer.rb
```diff
@@ -91,6 +91,7 @@ def recover
       user_recovered
     end
     topic = Topic.with_deleted.find @post.topic_id
+    topic.update_column(:user_id, Discourse.system_user.id) if !topic.user_id
```
ZogStriP: Even thought we're caching the system user, I think it might be faster to only use the constant here?

```suggestion
    topic.update_column(:user_id, Discourse::SYSTEM_USER_ID) if !topic.user_id
```

lib/post_destroyer.rb
```diff
@@ -103,6 +104,7 @@ def recover
   end
 
   def staff_recovered
+    @post.update_column(:user_id, Discourse.system_user.id) if !@post.user_id
```
ZogStriP: Ditto regarding the constant.

SamSaffron:
feel free to merge once small sys user id change is applied.

PR #8840, opened 2020-02-03
Log error when optimized image file fails to store.

schungx:


eviltrout:
Seems reasonable to me! Thanks.

schungx:
@eviltrout  I have been thinking whether it is necessary to put in also a `rescue` section because `store_optimized_image` may throw (in external stores at least) when the store fails for any reason.

If it throws, it will probably leave the temp file open dangling, and create an `OptimizedImage` record with no `url`.

eviltrout:
@schungx you might want a follow up PR for that then.

schungx:
Guess I'll leave it until there appears to be mysterious image store failings... ;-)

PR #8844, opened 2020-02-04
UX: introduces icon-picker component for badges

jjaffeux:
Screenshot 2020-02-04 at 14 27 36
Screenshot 2020-02-04 at 14 27 41


app/assets/javascripts/select-kit/components/icon-picker.js.es6
```diff
@@ -0,0 +1,49 @@
...
+      }
+
+      this.attrs.onChange && this.attrs.onChange(value, item);
+    }
+  },
+
+  _transformIconName(icon) {
+    let processedIconName = icon;
+    if (!icon.startsWith("fa")) {
+      processedIconName = `fa-${processedIconName}`;
```
pmusaraj: I'm a bit torn about including the "fa-" prefix in the displayed icon name here. FontAwesome, when you search for icons, doesn't display the prefix. I think we should only display the `fab-` and `far-` prefixes in the picker.

jjaffeux: so value should have fa- but we never display it, right?

pmusaraj: No, we don't need `fa-` at all, only `fab-` (for brand icons) and `far-` (for the regular icon set).

pmusaraj:
@jjaffeux I made the changes to have the search display all available icons. A few nice-to-haves that I didn't manage to do: 

- on the first search (with no filter set) it would be nice to show a message to the user along the lines of "Showing the first 200 results, please use a keyword to search the full FontAwesome set)
- debouncing the search would nice too
- also, I couldn't get this to work without resorting to jQuery, the icons wouldn't render properly when inserted in the DOM via vanilla JS

jjaffeux:
> @jjaffeux I made the changes to have the search display all available icons. A few nice-to-haves that I didn't manage to do:
> 
> * on the first search (with no filter set) it would be nice to show a message to the user along the lines of "Showing the first 200 results, please use a keyword to search the full FontAwesome set"

No way to do this ATM, will think about it, but don't want to stop PR for this

> * debouncing the search would nice too

It should be the case and in my tests seems to work, I will merge as is, if it's not debounced it's an sk bug anyways, not related to this PR

> * also, I couldn't get this to work without resorting to jQuery, the icons wouldn't render properly when inserted in the DOM via vanilla JS

I tried too, not much luck, I think there's some load event to use for svg in vanilla js maybe? Fine like this for now

PR #8850, opened 2020-02-04
FIX: HTML parsed in Featured topic title on profile

markvanlan:


markvanlan:
Addresses [this meta topic](https://meta.discourse.org/t/html-entities-not-being-parsed-in-featured-topic-title-on-profile/140546?u=markvanlan)

PR #8853, opened 2020-02-04
FEATURE: Allow single option polls

markvanlan:


markvanlan:
@eviltrout Oh right, the message does need to change.. So you want me to completely remove to translation to it is re-translated, correct?

eviltrout:
Just leave those files as-is. Transifex should pick up that it changed in English and flag it for translation.

PR #8864, opened 2020-02-05
DEV: adds afterRender decorator

jjaffeux:
This decorator, will make any DOM actiopn perfectly safe:

```
import { afterRender } from "discourse-common/utils/decorators";

@afterRender
_setStyle() {
  this.element.style.backgroundColor = "blue";
}
```

### DIFF
--- a/app/assets/javascripts/discourse-common/utils/decorators.js.es6
+++ b/app/assets/javascripts/discourse-common/utils/decorators.js.es6
@@ -1,6 +1,7 @@
 import handleDescriptor from "ember-addons/utils/handle-descriptor";
 import isDescriptor from "ember-addons/utils/is-descriptor";
 import extractValue from "ember-addons/utils/extract-value";
+import { schedule, next } from "@ember/runloop";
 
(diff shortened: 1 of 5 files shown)


jjaffeux:
@eviltrout note that the test is kinda useless with our current stack, as a test will not fail if there's an exception and try/catch block can't currently catch an exception in a deferred action. We should be able to address this once we watch up on ember/qunit.

app/assets/javascripts/discourse-common/utils/decorators.js.es6
```diff
@@ -13,6 +14,19 @@ export default function discourseComputedDecorator(...params) {
   }
 }
 
+export function afterRender(target, name, descriptor) {
+  const originalFunction = descriptor.value;
+  descriptor.value = function() {
+    next(() => {
+      schedule("afterRender", () => {
+        if (this.element && !this.isDestroying && !this.isDestroyed) {
```
ZogStriP: 😍

PR #8867, opened 2020-02-05
DEV: Make image resize controls more resilient

udan11:
Commit aa24be1 made it possible to build data attributes from image's
Markdown and this changes ensure that the resize controls still work
with some data parameters present.


SamSaffron:
I support merging this once we have an acceptance test of sorts. It is fiddly stuff.

PR #8870, opened 2020-02-05
PERF: Load users in batches when generating notifications

davidtaylorhq:
Previously we were trying to load all the users into memory simultaneously, which can cause Sidekiq to run out of memory for large sites.

notify_first_post_watchers now matches the code for sending other types of notifications.

### DIFF
--- a/app/services/post_alerter.rb
+++ b/app/services/post_alerter.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class PostAlerter
+  USER_BATCH_SIZE = 1000
+
   def self.post_created(post, opts = {})
     PostAlerter.new(opts).after_save_post(post, true)
     post
@@ -140,8 +142,7 @@ class PostAlerter
     users = User.where(id: user_ids)
 
     DiscourseEvent.trigger(:before_create_notifications_for_users, users, post)
-    users.pluck(:id).each do |user_id|
-      user = User.find_by(id: user_id)
+    each_user_in_batches(users) do |user|
       create_notification(user, Notification.types[:watching_first_post], post)
     end
   end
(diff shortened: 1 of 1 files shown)


app/services/post_alerter.rb
```diff
@@ -140,7 +140,8 @@ def notify_first_post_watchers(post, user_ids)
     users = User.where(id: user_ids)
 
     DiscourseEvent.trigger(:before_create_notifications_for_users, users, post)
-    users.each do |user|
+    users.pluck(:id).each do |user_id|
```
eviltrout: Maybe `find_in_batches` would be better here?

davidtaylorhq: It would probably be more efficient, but find_in_batches can be [affected by race conditions](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches) (e.g. if a user at the beginning of the list is deleted while the notifications are being sent, it could cause a user near the end of the list to be double-notified)

Loading one-by-one matches the existing behaviour of [`PostAlerter#notify_post_users`](https://github.com/discourse/discourse/blob/master/app/services/post_alerter.rb#L614-L618)

eviltrout: That's a nice solution. I think we should do it here and elsewhere.

davidtaylorhq: :+1: updated `notify_first_post_watchers` to use batches, and also `#notify_post_users`

app/services/post_alerter.rb
```diff
@@ -140,7 +142,7 @@ def notify_first_post_watchers(post, user_ids)
     users = User.where(id: user_ids)
 
     DiscourseEvent.trigger(:before_create_notifications_for_users, users, post)
-    users.each do |user|
+    each_user_in_batches(users) do |user|
```
ZogStriP: Why not use `user_ids` here instead of `users`?

app/services/post_alerter.rb
```diff
@@ -611,14 +613,22 @@ def notify_post_users(post, notified, include_category_watchers: true, include_t
 
     already_seen_users = TopicUser.where(topic_id: post.topic.id).where("highest_seen_post_number >= ?", post.post_number).pluck(:user_id)
 
-    notify.pluck(:id).each do |user_id|
-      notification_type = already_seen_users.include?(user_id) ? Notification.types[:edited] : Notification.types[:posted]
-      user = User.find_by(id: user_id)
+    each_user_in_batches(notify) do |user|
+      notification_type = already_seen_users.include?(user.id) ? Notification.types[:edited] : Notification.types[:posted]
```
SamSaffron: this is not new ... but `already_seen_users` should be a `Set` so include is fast and we are not constantly walking it.

app/services/post_alerter.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class PostAlerter
+  USER_BATCH_SIZE = 1000
```
SamSaffron: maybe 100 to keep memory down a tiny bit?

SamSaffron:
a couple of minor changes, but once you apply them feel free to merge, nice change.

app/services/post_alerter.rb
ZogStriP: Also it should probably be called `already_seen_user_ids` since it's only holding user ids.

app/services/post_alerter.rb
davidtaylorhq: 👍 Agreed. I took the 1000 default from Activerecord's `find_in_batches`, but our user object it pretty heavy so 100 makes more sense.

PR #8874, opened 2020-02-06
FIX: Change store.js.es6 to allow for random ID generation for duplicate records

martin-brennan:
Extracted from https://github.com/discourse/discourse/pull/8772

In some cases we want to use a random value instead of the object's id, e.g. if we have a list where the same ID database record shows multiple times but with different attached data. In the bookmarks PR for example the topic list shows multiple copies of the same topic because you can bookmark multiple posts in a topic with different reminder dates + times.

### DIFF
--- a/app/assets/javascripts/discourse/models/store.js.es6
+++ b/app/assets/javascripts/discourse/models/store.js.es6
@@ -198,9 +198,11 @@ export default EmberObject.extend({
     });
   },
 
-  createRecord(type, attrs) {
+  createRecord(type, attrs, opts = {}) {
     attrs = attrs || {};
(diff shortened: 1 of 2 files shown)


eviltrout:
This is probably not the right approach to solve your problem. The store is heavily built around the concept of an identity map and re-using objects. In other words, if you request the same record by id more than once, you will get back the same object. This helps saving memory and keeping things bound together.

If you need the same ID to refer to different objects, they must not be the same thing? I would recommend using a different ID in this case. For example, if it's the topic id but you are bookmarking posts, could you use the post id instead?

eviltrout:
@martin-brennan in that case could you use the post action (bookmark) id? There is probably some unique ID you can use for the store. If you need more help identifying it please do ask.

PR #8877, opened 2020-02-06
Various fixes to support posts with no user

udan11:


db/post_migrate/20200206101739_remove_not_null_from_user_id.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveNotNullFromUserId < ActiveRecord::Migration[6.0]
+  def up
+    execute "ALTER TABLE topic_links ALTER COLUMN user_id DROP NOT NULL"
```
ZogStriP: Do you know when this could be happening? When are we deleting an user but forgetting to delete `topic_links` records?

udan11: As part of the post processing pipeline we are calling `TopicLink.extract_from`. When processing a post with no user, that method will raise an exception as it attempts to create `TopicLink` records (because `user_id` is `nil`).

ZogStriP: Maybe we should instead bypass the post processing altogether if there are no user associated to a post?

udan11: I do not think it makes sense skipping post processing because some things may not work anymore, but I made it skip `TopicLink` creation which was the problem in the first place.

eviltrout:
@udan11 please merge when you are around to see through any bugs, perhaps the beginning of a day.

PR #8892, opened 2020-02-07
UX: adds visual feedback to topic-notifications-button

jjaffeux:
Screenshot 2020-02-07 at 14 48 13

### DIFF
--- a/app/assets/javascripts/select-kit/components/topic-notifications-button.js.es6
+++ b/app/assets/javascripts/select-kit/components/topic-notifications-button.js.es6
@@ -1,15 +1,24 @@
 import Component from "@ember/component";
+import { computed } from "@ember/object";
+import { later } from "@ember/runloop";
 
 export default Component.extend({
   layoutName: "select-kit/templates/components/topic-notifications-button",
   classNames: ["topic-notifications-button"],
   appendReason: true,
   showFullTitle: true,
+  isLoading: false,
+  icon: computed("isLoading", function() {
+    return this.isLoading ? "spinner" : null;
+  }),
 
   actions: {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/select-kit/components/topic-notifications-button.js.es6
```diff
@@ -1,15 +1,24 @@
...
+  }),
 
   actions: {
     changeTopicNotificationLevel(newNotificationLevel) {
       if (newNotificationLevel !== this.notificationLevel) {
-        this.topic.details.updateNotifications(newNotificationLevel);
+        this.set("isLoading", true);
+        this.topic.details
+          .updateNotifications(newNotificationLevel)
+          .finally(() => later(() => this.set("isLoading", false), 250));
```
eviltrout: Why do we need the 250ms delay here?

jjaffeux: Sorry forgot to answer, this is to have a consistent loading animation. Often the loading is so fast that you barely see it and end up just having a blinking/clunky UI. Given this is only for animation purpose and doesn't block UI I think it's fine to always have this delay.

eviltrout: Sounds good to me. Thanks for the clarification.

SamSaffron:
@jjaffeux  looks like we have conflict :( can you fix and merge this, change looks pretty safe to me.

jjaffeux:
@SamSaffron  I want to try a better a solution, this is why I postponed merged for now.

SamSaffron:
I worry that we are keeping this open for too long @jjaffeux  closing for now pending a new PR

PR #8895, opened 2020-02-07
FIX: Make inline oneboxes work with secured topics in secured contexts

udan11:


udan11:
That behaviour is already tested in `oneboxer_spec.rb`, but I have added another test in `inline_oneboxer_spec.rb`.

spec/components/inline_oneboxer_spec.rb
```diff
@@ -57,6 +62,39 @@
...
+
+    let(:admin) { Fabricate(:admin) }
+
+    it "can lookup private topics if in same category" do
+      topic = Fabricate(:topic, category: category)
+      topic1 = Fabricate(:topic, category: category)
+      topic2 = Fabricate(:topic, category: category2)
+
+      # Link to `topic` from new topic (same category)
+      onebox = InlineOneboxer.lookup("#{Discourse.base_url}#{topic.relative_url}", user_id: admin.id, category_id: category.id, skip_cache: true)
```
ZogStriP: Why use `topic.relative_url` instead of `topic.url`?

spec/fabricators/category_fabricator.rb
```diff
@@ -10,23 +10,13 @@
...
-  slug 'happy'
-  user
-end
-
 Fabricator(:private_category, from: :category) do
   transient :group
 
-  name 'Private Category'
-  slug 'private'
+  name { sequence(:name) { |n| "Private Category #{n}" } }
```
ZogStriP: Why that change?

udan11: Fabricating two `:private_category` in the same context would raise a 'duplicate name / slug' validation error.

PR #8906, opened 2020-02-10
DEV: Replace User.unstage and User#unstage API with User#unstage!

CvX:
Quoting @SamSaffron:

> * User.unstage mixes concerns of both unstaging users and updating params which is fragile/surprising.
> * u.unstage destroys notifications and raises a user_unstaged event prior to the user becoming unstaged and the user object being saved.

`User#unstage!` no longer updates user attributes and saves the object before triggering the `user_unstaged` event.

---

_To be merged after the next release?_


ZogStriP:
@CvX can you fix the conflict and merge sooner rather than later?

PR #8907, opened 2020-02-10
DEV: enforces no-invalid-interactive linting rule

jjaffeux:


app/assets/javascripts/admin/templates/components/themes-list.hbs
```diff
@@ -1,10 +1,17 @@
...
+  {{d-button
+    action=(action "changeView")
+    actionParam=THEMES
+    class=(concat "themes-tab " "tab " (if themesTabActive 'btn-danger active' ''))
+    label="admin.customize.theme.title"
+  }}
+  {{d-button
+    action=(action "changeView")
+    actionParam=COMPONENTS
+    class=(concat "themes-tab " "tab " (if componentsTabActive 'btn-danger active' ''))
```
CvX:
```suggestion
    class=(concat "components-tab " "tab " (if componentsTabActive 'btn-danger active' ''))
```

jjaffeux: nice catch thx

jjaffeux:
> This seems good but is more likely to break stuff than some of the other linting changes right? Considering we have a release scheduled for next week, how confident do you feel about it?

I'm a little bit worried, the alert change I made is probably risky given the lot of different ways we use alerts and with different markup... maybe we wait

PR #8910, opened 2020-02-10
DEV: introducing user-chooser

jjaffeux:
user-chooser is a new select-kit component which aims to replace user-selector in the long term.


app/assets/javascripts/select-kit/components/user-chooser.js.es6
```diff
@@ -0,0 +1,85 @@
...
+      exclude: this.excludedUsers,
+      includeGroups: options.includeGroups,
+      allowedUsers: options.allowedUsers,
+      includeMentionableGroups: options.includeMentionableGroups,
+      includeMessageableGroups: options.includeMessageableGroups,
+      groupMembersOf: options.groupMembersOf,
+      allowEmails: options.allowEmails
+    }).then(result => {
+      if (typeof result === "string") {
+        // do nothing promise probably got cancelled
```
eviltrout: This is a bit odd - you saw this error in testing?

jjaffeux: Not that's not odd, this is fully expected. I guess this is an artifact of the past. But when we cancel userSearch will return some _CANCELED string.

Promise cancelation can happen if you have an empty filter and no topicId/categoryId,  these files (userSearch, category tag search...) would need a huge refactoring, but didn't want it touch it for now. So I handled this case like this for now.

eviltrout: So it's our code that produces _CANCELLED? In that case this is OK but it's just an odd pattern to have to check for a string as an API. If it's not new here let's leave it for now.

discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/dev-introducing-user-chooser-pr-8910/8914/8):

yes our code creates this. it’s in lib/user-search

PR #8912, opened 2020-02-10
FIX: Render HTML for pie chart options

markvanlan:
We have to render our own legends for Chartjs charts if we want HTML to be rendered. That is what this does. Turns off legends by default but then renders them after the chart is rendered.

Legends are rendered beneath the chart now so the chart is never hidden inside a scrollable section on load.

I stripped all HTML tags out of the mouse popup when hovering over the pie. That will always be rendered as text, not html so this makes it readable :+1: 

# Old

Screenshot from 2020-02-10 11-05-38

# New
Screenshot from 2020-02-10 11-10-37


plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -620,13 +622,14 @@ createWidget("discourse-poll-pie-chart", {
...
+    contents.push(h(`div#poll-results-legend-${attrs.id}.pie-chart-legends`));
     return contents;
   }
 });
 
 function pieChartConfig(data, labels, opts = {}) {
-  const aspectRatio = "aspectRatio" in opts ? opts.aspectRatio : 2.0;
-  const displayLegend = "displayLegend" in opts ? opts.displayLegend : true;
+  const aspectRatio = "aspectRatio" in opts ? opts.aspectRatio : 2.2;
+  const strippedLabels = labels.map(l => l.replace(/<[^>]*>?/gm, "").trim());
```
eviltrout: Stripping HTML can be error prone. I just want to confirm that there's no other way to have the chart library omit this?

markvanlan: @eviltrout Chartjs does not have a strip out html feature.

I could do something more robust like this
```
function strip(html){
   var doc = new DOMParser().parseFromString(html, 'text/html');
   return doc.body.textContent || "";
}
```

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
```diff
@@ -636,13 +639,26 @@ function pieChartConfig(data, labels, opts = {}) {
...
     options: {
       responsive: true,
       aspectRatio,
       animation: { duration: 400 },
-      legend: { display: displayLegend }
+      legend: { display: false },
+      legendCallback: function(chart) {
+        let text = [];
+        for (let i = 0; i < labels.length; i++) {
+          text.push(
```
jjaffeux: As much as possible I think it's better create DOM objects here instead of string string concatenation.

https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
eviltrout: Sure, try it out but be sure to check performance and if it works in IE11.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
markvanlan: I believe this introducing more complexity because we are inside a widget.

I changed the legendCallback to this: 
```javascript
      legendCallback: function(chart) {
        let legendsContainer = document.createElement('div')
        for (let i = 0; i < labels.length; i++) {
          let legend = document.createElement("div")
          let swatch = document.createElement("span")
          legend.classList.add('legend')
          swatch.classList.add('swatch')
          swatch.style.backgroundColor = chart.data.datasets[0].backgroundColor[i]
          let legendLabel = document.createElement('span')
          legendLabel.innerHTML = labels[i]
          legend.appendChild(swatch)
          legend.appendChild(option)
          legendsContainer.appendChild(legend)
        }
        return legendsContainer
      }
```

And then inside the widget
```javascript
      document.getElementById(
        `poll-results-legend-${attrs.id}`
      ).appendChild(chart.generateLegend());
```

But the problem is now that those DOM  elements are being held outside of scope of the widget, so they are not being garbage collected. Whenever the widget is re-rendered, the legend is duplicated.

Screenshot from 2020-02-10 15-57-42

I think returning an HTML string here is better than doing manual cleanup. What do you think @jjaffeux ?

jjaffeux: there's probably a way but chart.js doesnt make it easy for sure... Let's give up on this. Thanks for trying.

plugins/poll/assets/javascripts/widgets/discourse-poll.js.es6
markvanlan: @eviltrout Do you trust caniuse? It says that it has full support for every major browser.

https://caniuse.com/#feat=mdn-api_domparser_parsefromstring_html

PR #8913, opened 2020-02-10
DEV: Use classes for styling user and group cards

davidtaylorhq:
With hard-coded element-ids, it is impossible for themes/plugins to display multiple cards on a single page. Using classes is a more flexible approach.

This is a low-risk change for core, but it will likely break a number of themes/plugins. One mitigation would be to maintain the element-ids, in addition to the new classes. Then themes/plugins which need multiple cards per page can override the element-id manually. We could then look at removing the element-id later 🤔

Leaving this as a draft for now while we decide the best approach... do you have any thoughts @awesomerobot


davidtaylorhq:
Updated to maintain the element-ids for compatibility. Will merge tomorrow if there are no objections :)

PR #8915, opened 2020-02-10
FIX: Make category slug validation less strict

udan11:
This was changed recently and caused issues saving old categories which
already had digits at the beginning of the slug (for example, '30-days').


udan11:
I avoided that one because we also support URLs such as `/c/ID-SLUG/`.

PR #8919, opened 2020-02-10
add postmark webhook handling

pfaffman:
Add webhook processing for postmark bounces and spam complaints.

It appears to do the Right Thing for Bounce and Spam complaint webhook processing.


eviltrout:
Looks good to me! Thanks.

PR #8921, opened 2020-02-10
FIX: when unread reply notification exists don't create new

lis2:
From time to time, the user is creating a reply post and then they want to add additional details. They edit an existing post and for example, add a quote from a previous one.

In that situation, if the user to whom reply was directed to already have the unread notification, we should not create the new one.

That behaviour was mentioned here: https://meta.discourse.org/t/reply-then-edit-to-add-quote-notification-redundancy/138358

### DIFF
--- a/app/services/post_alerter.rb
+++ b/app/services/post_alerter.rb
@@ -261,6 +261,7 @@ class PostAlerter
   end
 
   def should_notify_previous?(user, post, notification, opts)
+    return false unless notification
     case notification.notification_type
     when Notification.types[:edited] then should_notify_edit?(notification, post, opts)
     when Notification.types[:liked]  then should_notify_like?(user, notification)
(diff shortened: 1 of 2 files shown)


app/services/post_alerter.rb
```diff
@@ -332,6 +332,9 @@ def create_notification(user, type, post, opts = {})
 
     return if existing_notification_of_same_type && !should_notify_previous?(user, post, existing_notification_of_same_type, opts)
 
+    existing_unread_reply_notification = existing_notifications.find { |n| !n.read && n.notification_type == Notification.types[:replied] }
```
ZogStriP: You can merge both lines 😉 

```ruby
return if existing_unread_reply_notification = existing_notifications.find { |n| !n.read && n.notification_type == Notification.types[:replied] }
```

lis2:
@SamSaffron could you review it? Regis and Robin reviewed and accepted it before our conversation worth 1 million dollars.

PR #8933, opened 2020-02-12
UX: Differentiate Between PMs and Topics in Search Results

adqm:
Per the discussion [here](https://meta.discourse.org/t/search-public-topics-and-pms-simultaneously/139624), this PR adds a little indicator to PMs in the search results.  I'm not sure that this is ready for prime-time, and there are at least a couple of things I'm not sure about:

* Right now, I've styled things like [this](https://hz.mit.edu/discourse/display_1.png), with the envelope next to the title.  I also played around a little bit with putting the indicator down next to the categories, which looked like [this](https://hz.mit.edu/discourse/display_2.png) instead.  I'm not sure which of those is preferable.  And if it's the second, there's probably a way to properly internationalize the "PM" string?

    * Alternatively, it's probably possible to set the envelope on top of the author's avatar or something like that...

* Does `private-message-icon` deserve its own file?  Is there a more appropriate place to put that function?

* Right now, this puts the envelope next to PMs in all search results, not just those using `in:all`.  So they will show up next to every result in an `in:personal` search, for example.  That matches my preference (versus only showing the envelope next to PMs for `in:all` searches), but I'm not sure it's objectively preferable.

Any feedback / suggestions would be welcome, and I'd be interested to know if there are better ways of doing this.


ZogStriP:
> * Right now, I've styled things like [this](https://hz.mit.edu/discourse/display_1.png), with the envelope next to the title

I think that's the best approach 👍 as it's consistent with the way we show the enveloppe when viewing a PM.

> * Does `private-message-icon` deserve its own file?  Is there a more appropriate place to put that function?

Not sure about that one. Why not just do `{{#if topic.isPM}}{{d-icon "enveloppe"}}{{/if}}` or something similar?

> * Right now, this puts the envelope next to PMs in all search results, not just those using `in:all`.

I think that's fine.

adqm:
> Not sure about that one. Why not just do `{{#if topic.isPM}}{{d-icon "enveloppe"}}{{/if}}` or something similar?

In part because I didn't notice that `d-icon` existed as a helper already :).  That's obviously a lot cleaner.  I'll rearrange.

adqm:
Poking around some more, it's possible that using `{{private-message-glyph}}` might actually be better, for consistency with how it is displayed when actually viewing the message, just with some custom styling on top to make it look nicer in this context.

app/serializers/listable_topic_serializer.rb
```diff
@@ -9,6 +9,8 @@ class ListableTopicSerializer < BasicTopicSerializer
              :last_posted_at,
              :bumped,
              :bumped_at,
+             :archetype,
+             :is_pm,
```
davidtaylorhq: I don't think we need both of these new parameters. Can we manage with just `archetype`? That would match other places where topics are serialized. 

If you include the archetype, then you can [magically](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/models/topic.js.es6#L352) use the `topic.isPrivateMessage` boolean in handlebars templates.

davidtaylorhq:
This is looking good - I left one comment about the serializer change.

In terms of styling, I think we should try to match the existing "topic status" indicators:

Screenshot 2020-02-12 at 16 04 27

So I would check how the font-size/line-height/margin/colour is set on those indicators, and do the same here. It may be useful to wrap the new icon in something like `<span class='pm-indicator'>`

app/serializers/listable_topic_serializer.rb
adqm: I had no idea, thanks!  There seems to be a lot of black magic here that I don't yet understand :)

adqm:
All sounds good, thanks!  I'll make those changes and let you know when it's ready for another look.

adqm:
Just wanting to continue the conversation here a bit as I think about things...sorry if this conversation should be happening someplace else (if so, just let me know where).

From some perspective, this PR is ready for another look:

* Per your suggestion, I got rid of `is_pm` from the serializer, and used `topic.isPrivateMessage` instead.

* I also changed the styling of the envelope icon to match the topic statuses.  I'm not sure whether you think there might be a better place for those SCSS rules; it feels a little bad to duplicate those lines from `.topic-statuses` almost verbatim, but it would also have felt wrong just to wrap the PM glyph in `.topic-statuses`...

* The display when something is a warning looks weird (it displays both envelopes).  Which leads me to my real thought, outlined below.

I've also started wandering down a different rabbit hold that I think might actually make more sense than this, though it's not _quite_ working yet and there may be things I'm missing.  In [this branch](https://github.com/adqm/discourse/tree/pm_as_topic_status), I started instead implementing the PM indicator just by modifying the topic status stuff, as I thought it seemed to fit there a little better.  I almost have that working (the exception, AFAICT, being that the "messages" view currently doesn't work...).  My gut says that's a more promising path to explore, though I'm happy to continue that conversation, too.

app/assets/javascripts/discourse/templates/full-page-search.hbs
```diff
@@ -85,6 +85,7 @@
                   {{/if}}
 
                   <a class='search-link' href='{{unbound result.url}}' {{action "logClick" result.topic_id}}>
+                    {{#if result.topic.isPrivateMessage}}<span class="pm-indicator" title="{{I18n 'topic_statuses.personal_message.title'}}">{{d-icon "envelope"}}</span>{{/if}}
```
davidtaylorhq: We normally use a lowercase for `i18n` in handlebars
```suggestion
                    {{#if result.topic.isPrivateMessage}}<span class="pm-indicator" title="{{i18n 'topic_statuses.personal_message.title'}}">{{d-icon "envelope"}}</span>{{/if}}
```

davidtaylorhq:
This PR is looking good! 😄 

> modifying the topic status stuff, as I thought it seemed to fit there a little better

Maybe... Having the double envelope for warnings is weird, so we should definitely avoid that. I see a couple of options:

1. Keep this PR as-is, and wrap the new icon with `{{#unless topic.is_warning}}...{{/unless}}`. This would match the implementation in `topic.hbs`

2. Do as you suggest, and modify the topic status component. The problem we have is that the topic-status component is used in a few different places. To ensure we don't affect any other locations, I suggest adding another parameter to the component, so it would be called like:

    ```
    {{topic-status topic=result.topic disableActions=true showIsPm=true}}
    ```
    Then, **if** the showIsPm variable is true, show the envelope icon for PMs. If not, maintain the existing behavior. With that approach, we don't need to touch any other parts of the code (e.g. topic.hbs)

I'm torn 50/50 between the options - (1) is definitely easier, and lower risk. (2) is probably the cleanest, but more work and more risk. Happy to go with whichever option you would like to work on. Nothing is set in stone, so we can always refactor again in the future.

app/assets/javascripts/discourse/templates/full-page-search.hbs
ZogStriP: TIL if that's case insensitive 😲

adqm:
Sorry for disappearing here, @davidtaylorhq.  I finally had a chance to come back to this, and in the interest of not letting it drag on longer, I decided to go with the easier of the two options here (the existence proof of something like this in `topic.hbs` made me feel more OK about it).

I also made a slight additional change to the CSS, making `search-link` have `display: flex` to avoid extra whitespace creeping in between the regular grey envelope and the title, so that the PM envelope and the warning envelope line up with each other.  Here's what things look like as of now:

icons lined up

app/assets/javascripts/discourse/templates/full-page-search.hbs
```diff
@@ -85,6 +85,11 @@
                   {{/if}}
 
                   <a class='search-link' href='{{unbound result.url}}' {{action "logClick" result.topic_id}}>
+                    {{#unless result.topic.is_warning}}
+                      {{#if result.topic.isPrivateMessage}}
+                      <span class="pm-indicator" title="{{i18n 'topic_statuses.personal_message.title'}}">{{d-icon "envelope"}}</span>
+                      {{/if}}
```
davidtaylorhq: Nitpick, this should be indented like

```suggestion
                      {{#if result.topic.isPrivateMessage}}
                        <span class="pm-indicator" title="{{i18n 'topic_statuses.personal_message.title'}}">{{d-icon "envelope"}}</span>
                      {{/if}}
```

davidtaylorhq:
@adqm, @ZogStriP and I discussed this, and decided that we should try and avoid duplicating code here. Especially the 'magic numbers' for the icon size in the SCSS. You were 100% right to suggest moving this into the `topic-status` - sorry I confused things earlier.

I've moved things around a little bit so that the icon is handled by topic-status, so I think we're good to merge now.

adqm:
Thanks!

A couple of lingering thoughts:

1. It feels a little weird to me to have the styling be inconsistent between the search results (orange envelope = warning, dark grey envelope = PM) and the topic views themselves (dark grey envelope = warning, light grey and slightly larger envelope = PM).

    Is there a reason, after this change, still to have the PM envelope as a special case in `topic.hbs`, as opposed to just letting `topic-status` do its thing there?  It seems like doing the latter might help with consistency (I had started down that road in [this branch](https://github.com/adqm/discourse/tree/pm_as_topic_status) when I was first looking into this).

1. Will this affect the way the subject lines display in the PM Inbox view?  I can't get that view to load locally as of right now, but maybe that's just something wrong with my local setup (and I had had a similar issue on the branch linked above when I was first testing it).

davidtaylorhq:
> inconsistent between the search results (orange envelope = warning, dark grey envelope = PM) and the topic views themselves (dark grey envelope = warning, light grey and slightly larger envelope = PM)

Agreed, the warning envelope should be the orange colour in PM headings. I believe that's taken care of by this change (because of the new `.topic-status-warning` class)

> Is there a reason, after this change, still to have the PM envelope as a special case in topic.hbs, as opposed to just letting topic-status do its thing there?

I had a quick look. The PM icon in `topic.hbs` is a link to the inbox, so that would need some more custom logic in `topic-status`.

> Will this affect the way the subject lines display in the PM Inbox view?

It shouldn't because I've put this behind a switch on `topic-status`. It only shows the PM icon when you explicitly enable it like:

```
{{topic-status topic=topic showPrivateMessageIcon=true}}
```

PR #8936, opened 2020-02-12
DEV: enforces link-rel-noopener linting rule

jjaffeux:


jjaffeux:
@eviltrout this one should be quite safe

app/assets/javascripts/discourse/templates/user.hbs
```diff
@@ -104,13 +104,16 @@
               </h3>
             {{/if}}
 
+
+            AAAAAA
```
CvX: 👀

app/assets/javascripts/discourse/templates/user/summary.hbs
```diff
@@ -77,10 +77,11 @@
         <ul>
           {{#each model.links as |link|}}
             <li>
+              {{!-- template-lint-disable  --}}
               <a class='domain'
                  href='{{unbound link.url}}'
                  title='{{unbound link.title}}'
-                 rel='{{unless user.removeNoFollow 'nofollow ugc noopener'}}'
+                 rel='{{unless user.removeNoFollow 'nofollow ugc noopener' 'noopener'}}'
```
CvX: This should work, right?

```suggestion
                 rel='noopener {{unless user.removeNoFollow 'nofollow ugc'}}'
```

CvX: (Same with the other `template-lint-disable` instances)

app/assets/javascripts/discourse/templates/user.hbs
ZogStriP: BBBBBB

jjaffeux: damn was sure I had removed it

app/assets/javascripts/discourse/templates/user/summary.hbs
jjaffeux: yes I think it would, will try :+1:

jjaffeux: Sadly it still errors in the linting, but I changed it to your syntax as it still prevents duplicating noopener

eviltrout:
Merging, I think the plugin failure is not related.

PR #8937, opened 2020-02-12
FEATURE: adds a new plugin api to decorate plugin outlets

jjaffeux:
```
api.decoratePluginOutlet(
  "discovery-list-container-top",
  elem => {
    if (elem.classList.contains("foo")) {
      elem.style.backgroundColor = "yellow";
    }
  }
);
```


app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -983,6 +984,31 @@ class PluginApi {
   addGlobalNotice(id, text, options) {
     addGlobalNotice(id, text, options);
   }
+
+  /**
+   * Used for decorating the `cooked` content of a a plugin-outlet after it's been rendered
```
ZogStriP:
```suggestion
   * Used for decorating the `cooked` content of a plugin-outlet after it's been rendered
```

davidtaylorhq:
So this is decorating the rendered handlebars from the outlet? What happens if the content of the outlet changes after initial render? Do we re-decorate?

jjaffeux:
> So this is decorating the rendered handlebars from the outlet? What happens if the content of the outlet changes after initial render? Do we re-decorate?

I think it makes sense :+1: See my last commit

davidtaylorhq:
Cool! What's the use case here? When would this be more useful than simply adding to the plugin outlet directly? 

I think we need to be super careful before adding yet another extensibility method. Whatever we add will need to be maintained for a long time.

jjaffeux:
The use case is mostly quality of life when you try to apply DOM functions over cooked elements.

If I have some markdown blocks : `[my-custom-ui option=foo][/my-custom-ui]` which I want to show in posts but also in group pages, or user card or category top... 

I will usually have an initializer.js.es6 file which will have an `api.decorateCooked` method but doesn't have the equivalent for plugin-outlets, making it kinda cumberstone to share code between some custom component connector code  and this initializer. With this you can now have this kind of code:


```
function applyYellowBackground(element, args) {
  element.style.backgroundColor = "yellow";
}

api.decorateCooked(applyYellowBackground);
api.decoratePluginConnector("discovery-list-container-top", applyYellowBackground);
```

test/javascripts/acceptance/plugin-outlet-decorator-test.js.es6
```diff
@@ -0,0 +1,64 @@
...
+        { id: "yellow-decorator" }
+      );
+    });
+  },
+
+  afterEach() {
+    delete Ember.TEMPLATES[`${PREFIX}/discovery-list-container-top/foo`];
+    delete Ember.TEMPLATES[`${PREFIX}/discovery-list-container-top/bar`];
+
+    resetDecorators();
```
eviltrout: This should probably be done in the test helpers. Otherwise if people have acceptance tests for their plugins they could leak.

app/assets/javascripts/discourse/components/plugin-connector.js.es6
```diff
@@ -45,6 +57,19 @@ export default Component.extend({
...
+  didReceiveAttrs() {
+    this._super(...arguments);
+
+    this._decoratePluginOutlets();
+  },
+
+  @afterRender
+  _decoratePluginOutlets() {
+    _decorators
+      .filterBy("connector", this.connector.outletName)
```
eviltrout: You could replace this `filterBy` with a hash lookup based on the outlet name. 

for example if decorators looked like:

`{ "my-outlet-name": [ dec0, dec1, dec2 ] }`

You could then:

`(_decorators[this.connector.outletName] || []) .forEach(dec => dec.callback(this.element, this.args))`

app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -983,6 +984,31 @@ class PluginApi {
   addGlobalNotice(id, text, options) {
     addGlobalNotice(id, text, options);
   }
+
+  /**
+   * Used for decorating the `cooked` content of a plugin-outlet after it's been rendered
```
eviltrout:
```suggestion
   * Used for decorating the rendered HTML content of a plugin-outlet after it's been rendered
```
Cooked means something specific to us -- taking raw text and rendering markdown -- this change makes it clear that it's the HTML in the outlet.

app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -983,6 +984,31 @@ class PluginApi {
...
+   *     }
+   *   }
+   * );
+   * ```
+   *
+   **/
+  decoratePluginConnector(pluginConnector, callback, opts) {
+    opts = opts || {};
+
+    addPluginConnectorDecorator(pluginConnector, callback, opts);
```
eviltrout:
```suggestion
    addPluginConnectorDecorator(pluginConnector, callback, opts || {});
```
Could make this change and remove the `opts = opts || {}` above.

jjaffeux:
This should be safe before release as it's not used yet.

PR #8939, opened 2020-02-12
PERF: Memoize PostRevisor#diff_size

udan11:
Although this method is usually called twice, its execution can take
about 1s.


udan11:
It took me ~9s to edit a post. This combined with 8e27fa1bd871272ee923ef948703afc02fb12e25 reduces it to ~1s, which is still more than I'd like, but makes it a bit more manageable.

PR #8941, opened 2020-02-12
FEATURE: export defaultRenderTag function

artemv:
This can be helpful for plugins e.g. when combined with replaceTagRenderer: a plugin may want to customize rendering for some tags and let others be rendered with default code


app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -51,7 +51,7 @@ import Composer from "discourse/models/composer";
 import { on } from "@ember/object/evented";
 
 // If you add any methods to the API ensure you bump up this number
-const PLUGIN_API_VERSION = "0.8.37";
+const PLUGIN_API_VERSION = "0.8.38";
```
eviltrout: The plugin-api surface didn't change here so it doesn't make sense to bump the version.

How are you using render-tag in your code? By importing?

artemv: > How are you using render-tag in your code? By importing?

Currently no (as the `replaceTagRenderer` is exposed via plugin-api) , but if PR is merged then yes that's what I want.

PR #8943, opened 2020-02-12
UX: Prevent category-drop topic count from wrapping to new line

pmusaraj:
This addresses a recent regression where the topic count on a subcategory drops on a second row: 

Discourse 2020-02-12 12-05-03


jjaffeux:
LGTM thanks

PR #8950, opened 2020-02-12
DEV: Test theme js compiler append_raw_template

markvanlan:


spec/lib/theme_javascript_compiler_spec.rb
```diff
@@ -110,4 +110,21 @@ def statement(template)
     end
   end
 
+  describe "#append_raw_template" do
+    let(:compiler) { ThemeJavascriptCompiler.new(1, 'marks') }
+    it 'adds the correct template to "Discourse.RAW_TEMPLATES"' do
+      template = "<h1>hello</h1>"
+      name = "/path/to/templates1"
+      compiler.append_raw_template("#{name}.raw", template)
```
davidtaylorhq: I think this test should be `.raw.hbs`? I don't think we ever used `.raw` by itself.

```suggestion
      compiler.append_raw_template("#{name}.raw.hbs", template)
```
(and then I think it will fail, because the new implementation doesn't remove the `.hbs` part)

markvanlan: @davidtaylorhq If you add this to the head of a templates:

Screenshot from 2020-02-18 08-14-27

The topic list will not be template will not be overwritten because the extension `.raw.hbs`. If you use just `.raw`, the template will be overriden. `.hbr` works as well.

I believe the `hbs` gets removed earlier in the pipeline for templates that are in a theme, and templates added in the way done above (admin -> customize -> theme html/css), `.hbs` is never appended. This is confirmed by @awesomerobot

PR #8958, opened 2020-02-13
Invalid activation link in Turkish locale

midorikocak:
Users cannot activate their accounts due to bad url.


gschlager:

PR #8963, opened 2020-02-13
FIX: Topics should honor auto-close when published to category

tshenry:
This is my attempt to fix the issue outlined at:
https://meta.discourse.org/t/not-autoclosing-topics-published-from-another-category/138898

I suspect there may be more to the puzzle, but in any case, I'm looking forward to any feedback.


eviltrout:
This looks like it could be the correct solution, but we're going to need a test to confirm it's working. 

The test should run the auto publish job with an auto-close set on the category, then confirm that the new published topic has the correct setting.

tshenry:
This is my first go at writing a test, so let me know if I need to tweak anything!

app/jobs/regular/publish_topic_to_category.rb
```diff
@@ -12,6 +12,8 @@ def execute(args)
       TopicTimer.transaction do
         TopicPublisher.new(topic, Discourse.system_user, topic_timer.category_id).publish!
       end
+
+      Topic.find(topic.id).inherit_auto_close_from_category
```
ZogStriP: Any reasons you need to "reload" the topic from the database?

```suggestion
      topic.inherit_auto_close_from_category
```

tshenry: You'll notice that `topic` is taken from the original `TopicTimer`. This ends up storing the topic with its starting `category_id` in `topic`. We need to run `inherit_auto_close_from_category` against the topic _after_ it's published to the new category and has an updated `category_id`. There may be a better way to handle this, but hopefully that at least explains the rationale behind the reload.

PR #8969, opened 2020-02-14
FEATURE: Users cannot reflag recently handled items using the same reason

romanrizzi:
cc @featheredtoast

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -4650,6 +4650,7 @@ en:
       restored: Restored
 
   reviewables:
+    already_handled: "Thanks, but we've already reviewed that post and determined it does not need to be flagged again."
     priorities:
       low: "Low"
       medium: "Medium"
--- a/lib/post_action_creator.rb
+++ b/lib/post_action_creator.rb
@@ -66,6 +66,13 @@ class PostActionCreator
 
     PostAction.limit_action!(@created_by, @post, @post_action_type_id)
 
+    reviewable = Reviewable.includes(:reviewable_scores).find_by(target: @post)
+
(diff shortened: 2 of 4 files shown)


config/locales/server.en.yml
```diff
@@ -4650,6 +4650,7 @@ en:
       restored: Restored
 
   reviewables:
+    already_handled: "A staff member already reviewed this post because someone else already flagged it for the same reason. Please wait for at least 24hs before flagging it again."
```
```suggestion
    already_handled: "Thanks, but we've already reviewed that post and determined it does not need to be flagged again."
```
A couple suggested changes here: we allow category moderation so it might not always be reviewed by staff. Secondly, we might want to make the 24 hours a setting and it's probably a good idea to not let them know how long they need to wait in this particular case.

lib/post_action_creator.rb
```diff
@@ -115,6 +122,19 @@ def perform
...
 
+  def flagging_post?
+    PostActionType.notify_flag_type_ids.include?(@post_action_type_id)
+  end
+
+  def cannot_flag_again?(reviewable)
+    return false if @post_action_type_id == PostActionType.types[:notify_moderators]
+    flag_type_already_used = reviewable.reviewable_scores.any? { |rs| rs.reviewable_score_type == @post_action_type_id }
+    not_edited_since_last_review = @post.last_version_at.blank? || reviewable.updated_at > @post.last_version_at
+    handled_recently = (Time.now - reviewable.updated_at).to_i / 3600 < 24
```
eviltrout:
```suggestion
    handled_recently = reviewable.updated_at > 24.hours.ago
```

PR #8970, opened 2020-02-14
PERF: Allow passing an existing list of user field ids when loading

davidtaylorhq:
This avoids the need for running `UserField.pluck(:id)` for each user that is serialized

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1093,16 +1093,17 @@ class User < ActiveRecord::Base
 
   USER_FIELD_PREFIX ||= "user_field_"
 
-  def user_fields
-    return @user_fields if @user_fields
-    user_field_ids = UserField.pluck(:id)
-    if user_field_ids.present?
-      @user_fields = {}
-      user_field_ids.each do |fid|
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -1093,16 +1093,17 @@ def associated_accounts
...
   USER_FIELD_PREFIX ||= "user_field_"
 
-  def user_fields
-    return @user_fields if @user_fields
-    user_field_ids = UserField.pluck(:id)
-    if user_field_ids.present?
-      @user_fields = {}
-      user_field_ids.each do |fid|
-        @user_fields[fid.to_s] = custom_fields["#{USER_FIELD_PREFIX}#{fid}"]
+  def user_fields(field_ids = nil)
```
eviltrout: This will rebuild the hash every time you call it though versus the memoized one which didn't.

I understand why you can't memoize it anymore due to the parameter, but is that actually faster? If it's called repeatedly it's going to build that hash repeatedly.

davidtaylorhq: 👍 have added memoization based on the requested fields

PR #8971, opened 2020-02-14
FEATURE: Support uploading a CSV with either user emails or usernames

romanrizzi:


app/jobs/regular/mass_award_badge.rb
```diff
@@ -3,9 +3,12 @@
 module Jobs
   class MassAwardBadge < ::Jobs::Base
     def execute(args)
+      return unless mode = args[:mode]
       badge = Badge.find_by(id: args[:badge_id])
-      users = User.select(:id, :username, :locale).with_email(args[:user_emails])
 
+      users = User.select(:id, :username, :locale)
+
+      users = mode == 'email' ? users.with_email(args[:users_batch]) : users.where(username: args[:users_batch])
```
SamSaffron: usually we would match on username_lower and lower case the username, saves us on all sorts of odd errors and we are unique on username_lower anyway

PR #8974, opened 2020-02-17
DEV: Use `type` instead of `method` in ajax calls

CvX:
Even though `type` is an alias for `method`, we have custom logic in `/discourse/lib/ajax` that checks only `type`, and ~200 other ajax calls in the codebase already use `type` param.


jjaffeux:
:+1: Look safe, but maybe we should wait after release though? What if some things were relying on the fact that this check was not happening (probably unknowingly) ?

ZogStriP:
Can we also have a lint rule or similar to prevent us from using `method` when we should be using `type`?

CvX:
> What if some things were relying on the fact that this check was not happening (probably unknowingly) ?

I don't know if they're relying on it, but definitely using `method` had quite a few unintended consequences:

1. It defaulted `args.dataType` to `"json"` for all requests instead of just `GET`
2. It forced `args.cache` to `true` for on all requests instead of just `GET` (though I think jQ just ignores it for non-GET and non-HEAD calls)
3. It didn't update CSRF token when it was missing, before making the call (that's the reason I had to update the tests: https://github.com/discourse/discourse/pull/8974/commits/9e94e6226a53379b7a0548556336657cbbf488ff)

None of these are backbreaking (otherwise we would catch that earlier) but yeah, I'd wait with merging after the release. 😉

---

> Can we also have a lint rule or similar to prevent us from using `method` when we should be using `type`?

I thought about it, but soon(-ish) we will most likely migrate to `fetch` (polyfilled or not) to drop the jQuery dependency, so a custom lint rule would be short-lived. (and fetch has has only one argument for this: `method` 😉)

The other way to approach this would be to deprecate or soft-deprecate the old param:

```js
if (args.method) {
  args.type = args.method;
  delete args.method;
}
```

What do you think?

eviltrout:
Bump for @ZogStriP

ZogStriP:
> The other way to approach this would be to deprecate or soft-deprecate the old param:

Sure, that would work 👍

ZogStriP:
@CvX any chances you could merge this sooner rather than later?

app/assets/javascripts/discourse/lib/ajax.js
```diff
@@ -139,6 +139,14 @@ export function ajax() {
       });
     };
 
+    if (args.method) {
+      args.type = args.method;
+      delete args.method;
+
+      // eslint-disable-next-line no-console
+      console.warn("`method` option is deprecated, use `type` instead");
```
davidtaylorhq: Can we use the `deprecated` helper here for consistency?

https://github.com/discourse/discourse/blob/7ec124fc897dced11eaf2e658011cb09b99f25e7/app/assets/javascripts/discourse-common/lib/deprecated.js.es6#L1

davidtaylorhq: And also some short reason for the deprecation would be useful

PR #8975, opened 2020-02-17
DEV: Use `discourse/lib/ajax` instead of Ember global

CvX:
Submitting this one separately from the other import changes since it has a (slightly) higher risk of unintentional change of behavior.


jjaffeux:
❤️  probably safer after release though

CvX:
This one still needs some work. Closing for now.

PR #8977, opened 2020-02-17
DEV: Import `makeArray` from `discourse-common` instead using `Ember` global

CvX:


jjaffeux:
👍 This one looks safe to me even before release @eviltrout

PR #8987, opened 2020-02-17
FIX: Disable poll options if user groups do not allow them

udan11:


udan11:
Yes, it was already there and it looks like this:

Screen Shot 2020-02-18 at 11 57 16

PR #8999, opened 2020-02-19
FEATURE: MVP Bookmarks with reminders user list changes

martin-brennan:
**Merge https://github.com/discourse/discourse/pull/9071 before this PR**

This PR changes the user activity bookmarks stream to show a new list of bookmarks based on the Bookmark record:

image

* If a bookmark has a name or reminder it will be shown as metadata above the topic title in the list
* The categories, tags, topic status, and assigned show for each bookmarked post based on the post topic
* Bookmarks can be deleted from the [...] menu in the list

image

As well as this, the list of bookmarks from the quick access panel is now drawn from the Bookmarks table for a user:

image

### Notes

* All of this new functionality is gated behind the `enable_bookmarks_with_reminders` site setting
* The `/bookmarks/` route now redirects directly to `/user/:username/activity/bookmarks-with-reminders`
* The structure of the Ember for the list of bookmarks is not ideal, this is an **MVP** PR so we can start testing this functionality internally. There is a little repeated code from topic.js.es6. There is an ongoing effort to start standardizing these lists that will be addressed in future PRs.

### Related (to be merged _after_ this)

https://github.com/discourse/discourse-assign/pull/62


app/assets/javascripts/discourse/models/bookmark.js.es6
```diff
@@ -18,6 +22,88 @@ const Bookmark = RestModel.extend({
...
+  },
+
+  @discourseComputed("created_at")
+  createdAt(created_at) {
+    return new Date(created_at);
+  },
+
+  @discourseComputed("tags")
+  visibleListTags(tags) {
+    if (!tags || !Discourse.SiteSettings.suppress_overlapping_tags_in_list) {
```
eviltrout: Is `this.siteSettings` not available in your model? It's preferred to the global constant.

app/assets/javascripts/discourse/routes/discovery.js.es6
```diff
@@ -13,8 +13,21 @@ export default DiscourseRoute.extend(OpenComposer, {
...
 
   beforeModel(transition) {
-    const user = User;
+    // the new bookmark list is radically different to this topic-based one,
+    // including being able to show links to multiple posts to the same topic
+    // and being based on a different model. better to just redirect
     const url = transition.intent.url;
+    if (
+      this.siteSettings.enable_bookmarks_with_reminders &&
+      url === "/bookmarks"
```
eviltrout: With this work in a subfolder? I suspect not.

app/assets/javascripts/discourse/routes/discovery.js.es6
```diff
@@ -13,8 +13,21 @@ export default DiscourseRoute.extend(OpenComposer, {
...
+    ) {
+      this.transitionTo(
+        "userActivity.bookmarksWithReminders",
+        this.currentUser
+      );
+    }
+
+    const user = User;
```
eviltrout: This is slightly confusing to me - `user` in lower case usually indicates an local or member variable. Could the code below use `User` instead of assigning it like this?

app/assets/javascripts/discourse/templates/user/bookmarks.hbs
```diff
@@ -0,0 +1,51 @@
+{{#if noContent}}
+  <div class='alert alert-info'>{{ noResultsHelp }}</div>
```
eviltrout:
```suggestion
  <div class='alert alert-info'>{{noResultsHelp}}</div>
```

app/assets/javascripts/discourse/templates/user/bookmarks.hbs
```diff
@@ -0,0 +1,51 @@
...
+              {{/if}}
+              {{#if bookmark.bookmark_reminder_at}}
+                <span class="bookmark-metadata-item">
+                  {{d-icon "far-clock"}}{{bookmark.formattedReminder}}
+                </span>
+              {{/if}}
+            </div>
+
+            {{topic-status topic=bookmark}}
+            <a class="title" href="/t/{{bookmark.topic_id}}/{{bookmark.bookmark_post_number}}">{{bookmark.title}}</a>
```
eviltrout: This is not going to work in subfolder installs.

lib/bookmark_query.rb
```diff
@@ -0,0 +1,41 @@
...
+    @user = user
+    @params = params
+  end
+
+  def list_all
+    results = user_bookmarks
+      .joins('INNER JOIN topics ON topics.id = bookmarks.topic_id')
+      .joins('INNER JOIN posts ON posts.id = bookmarks.post_id')
+      .joins('INNER JOIN users ON users.id = posts.user_id')
+      .select(<<-SQL
```
eviltrout: `select` in a case like this is quite dangerous. If in the future someone adds a column to these tables and doesn't update this code, another user of this method will have objects that look like `Bookmark` but are missing fields.

It is much safer to return the default ActiveRecord objects. Is there a specific reason you chose to do it this way?

lib/topic_view.rb
```diff
@@ -349,6 +349,11 @@ def topic_user
     end
   end
 
+  def has_bookmarks?
+    return false if @user.blank?
+    @topic.bookmarks.where(user_id: @user.id).any?
```
eviltrout:
```suggestion
    @topic.bookmarks.exists?(user_id: @user.id)
```

app/assets/javascripts/discourse/routes/discovery.js.es6
martin-brennan: This was there already, not sure why it is assigned, will clean up

app/assets/javascripts/discourse/routes/discovery.js.es6
martin-brennan: @eviltrout in this same area there is also the following:

```
if (
      (url === "/" || url === "/latest" || url === "/categories") &&
      transition.targetName.indexOf("discovery.top") === -1 &&
      User.currentProp("should_be_redirected_to_top")
    ) {
      User.currentProp("should_be_redirected_to_top", false);
      const period = User.currentProp("redirected_to_top.period") || "all";
      this.replaceWith(`discovery.top${period.capitalize()}`);
    }
```

Would that break on subfolder too? Is the alternative to just use `indexOf`?

app/assets/javascripts/discourse/templates/user/bookmarks.hbs
martin-brennan: Fixed by just using topic-link helper instead :+1:

lib/bookmark_query.rb
martin-brennan: I was just trying to be more efficient here by selecting only the columns I need. If it is a premature optimization I will just change this to include the user/post/topic instead!

app/assets/javascripts/discourse/controllers/user-activity-bookmarks-with-reminders.js.es6
```diff
@@ -0,0 +1,54 @@
...
+  noResultsHelp: null,
+
+  loadItems() {
+    this.setProperties({
+      content: [],
+      loading: true,
+      noResultsHelp: null
+    });
+
+    this.model
```
jjaffeux: It would be nice to have a return here this way the promise is passed around

app/assets/javascripts/discourse/controllers/user-activity-bookmarks-with-reminders.js.es6
```diff
@@ -0,0 +1,54 @@
...
+  },
+
+  @discourseComputed("loaded", "content.[]")
+  noContent(loaded, content) {
+    return loaded && content.length === 0;
+  },
+
+  actions: {
+    removeBookmark(bookmark) {
+      return bookmark.destroy().then(() => this.loadItems());
```
jjaffeux: I think this should work:

```suggestion
      return bookmark.destroy().then(this.loadItems);
```

app/assets/javascripts/discourse/controllers/user-activity-bookmarks-with-reminders.js.es6
```diff
@@ -0,0 +1,54 @@
...
+      })
+      .finally(() =>
+        this.setProperties({
+          loaded: true,
+          loading: false
+        })
+      );
+  },
+
+  @discourseComputed("loaded", "content.[]")
```
jjaffeux:
```suggestion
  @discourseComputed("loaded", "content.length")
```

app/assets/javascripts/discourse/components/bookmark-actions-dropdown.js.es6
```diff
@@ -0,0 +1,33 @@
...
+        icon: "trash-alt",
+        name: I18n.t("post.bookmarks.actions.delete_bookmark.name"),
+        description: I18n.t(
+          "post.bookmarks.actions.delete_bookmark.description"
+        )
+      }
+    ];
+  }),
+
+  actions: {
```
jjaffeux: as it's new code you should try to use our new pattern here:


```
import { action } from "@ember/object";

@action
onChange(action) {}
```

app/assets/javascripts/discourse/models/bookmark.js.es6
```diff
@@ -18,6 +26,107 @@ const Bookmark = RestModel.extend({
...
+
+  @discourseComputed("highest_post_number", "url")
+  lastPostUrl(highestPostNumber) {
+    return this.urlForPostNumber(highestPostNumber);
+  },
+
+  // Helper to build a Url with a post number
+  urlForPostNumber(postNumber) {
+    let url = Discourse.getURL(`/t/${this.topic_id}`);
+    if (postNumber && postNumber > 0) {
```
jjaffeux:
```suggestion
    if (postNumber > 0) {
```

In what case would we need to check on postNumber?

app/assets/javascripts/discourse/routes/discovery.js.es6
eviltrout: My mistake - looks like `url` here already has had the subfolder removed (I tested it.)

lib/bookmark_query.rb
eviltrout: If you want to select only certain columns you should use `pluck` or write a custom SQL query using `DB.exec`

`select` is dangerous. It returns objects that look like ActiveRecord objects but are missing certain attributes. I wrote about it here:

https://eviltrout.com/2013/10/05/enemy-of-the-state.html.html

In this case I do think it's a premature optimization and probably not worth the effort of migrating to mini-sql or pluck, but if you took those approaches I'd feel much safer. I consider `select` a bug waiting to happen in the future.

martin-brennan: Thank you for the blog article, I never thought about it in that way. Probably because I have never worked with many other people so everyone else usually knows all the columns :sweat_smile: I will change it to a regular query.

app/assets/javascripts/discourse/models/bookmark.js.es6
martin-brennan: This is from topic.js model, I will change there too

martin-brennan:
Thanks for your additional review here @jjaffeux  :)

app/assets/javascripts/discourse/controllers/user-activity-bookmarks-with-reminders.js.es6
martin-brennan: For the record this did not work, because inside loadItems `this` is pointing to the wrong thing.

PR #9000, opened 2020-02-19
FIX: prevents saving an apparently single key as an all key

jjaffeux:


ZogStriP:
Should this be tested as well?

davidtaylorhq:
This is supposed to already be handled by https://github.com/discourse/discourse/blob/499b57a9edab04d67a1fc9a76f5ea5bb2b4e3f43/app/assets/javascripts/admin/controllers/admin-api-keys-new.js.es6#L16-L21

🤔

davidtaylorhq:
I opened #9005, which is how I always intended it to work. Looks like I just missed out the `disabled=` parameter on the button.

@jjaffeux do you think that is an ok solution, instead of a bootbox?

PR #9001, opened 2020-02-19
FIX: When admin changes another user's email auto-confirm the change

martin-brennan:
Discussions in meta topic: https://meta.discourse.org/t/howto-change-a-users-email-when-they-cant-access-the-email-account-and-forgot-password/56322

When admin changes a user's email from the preferences page of that user:

* The user will not be sent an email to confirm that their
  email is changing. They will be sent a reset password email
  so they can set the password for their account at the new
  email address.
* The user will still be sent an email to their old email to inform
  them that it was changed.
* Admin and staff users still need to follow the same old + new
  confirm process, as do users changing their own email.


lib/email_updater.rb
```diff
@@ -5,9 +5,10 @@ class EmailUpdater
 
   attr_reader :user
 
-  def initialize(guardian = nil, user = nil)
+  def initialize(guardian = nil, user = nil, initiating_user = nil)
```
ZogStriP: 3 optional arguments might be the threshold were it's better to use optional _named_ arguments

```suggestion
  def initialize(guardian: nil, user: nil, initiating_user: nil)
```

spec/components/email_updater_spec.rb
```diff
@@ -17,6 +17,92 @@
     expect(updater.errors.messages[:base].first).to be I18n.t("change_email.error_staged")
   end
 
+  context "when an admin is changing the email of a user" do
+    let(:email_change_requesting_user) { Fabricate(:admin) }
```
ZogStriP: Calling this `admin` makes it easier to read

```suggestion
    let(:admin) { Fabricate(:admin) }
```

spec/components/email_updater_spec.rb
```diff
@@ -17,6 +17,92 @@
...
+    let(:updater) { EmailUpdater.new(user.guardian, user, email_change_requesting_user) }
+
+    def expect_old_email_job
+      Jobs.expects(:enqueue).with(:critical_user_email, has_entries(to_address: "old.email@example.com", type: :notify_old_email, user_id: user.id))
```
ZogStriP: Might be better to get rid of the string when we already have it in a variable?

```suggestion
      Jobs.expects(:enqueue).with(:critical_user_email, has_entries(to_address: old_email, type: :notify_old_email, user_id: user.id))
```

spec/components/email_updater_spec.rb
```diff
@@ -17,6 +17,92 @@
...
+
+      it "sends a reset password email to the user so they can set a password for their new email" do
+        expect_old_email_job
+        expect_forgot_password_job
+        updater.change_to(new_email)
+      end
+    end
+
+    context "for a staff user" do
+      let(:user) { Fabricate(:moderator, email: old_email) }
```
ZogStriP: Calling this `moderator` makes it easier to read

```suggestion
      let(:moderator) { Fabricate(:moderator, email: old_email) }
```

spec/components/email_updater_spec.rb
```diff
@@ -17,6 +17,92 @@
     expect(updater.errors.messages[:base].first).to be I18n.t("change_email.error_staged")
   end
 
+  context "when an admin is changing the email of a user" do
```
ZogStriP:
```suggestion
  context "when an admin is changing the email of another user" do
```

lib/email_updater.rb
martin-brennan: Damn I knew I should have done that!

PR #9004, opened 2020-02-19
FIX: Improve styling of share-and-invite modal

udan11:
Before:

Screen Shot 2020-02-19 at 11 24 22

After:

Screen Shot 2020-02-19 at 11 24 40


eviltrout:
What about in the case where an error is displayed? Does it still look OK?

PR #9005, opened 2020-02-19
FIX: Disable save button for API key creation when invalid

davidtaylorhq:
Possible alternative to https://github.com/discourse/discourse/pull/9000


jjaffeux:
> I like it. What do you think @jjaffeux ?

Yes, we have a discussion about it https://review.discourse.org/t/fix-prevents-saving-an-apparently-single-key-as-an-all-key-pr-9000/9210/5?u=jjaffeux

I'm all ok with this

PR #9006, opened 2020-02-19
Improve SMF2 Importer

justindirose:
Two major changes here:

- The addition of a bbcode conversion method. 
- Improving the script to work with frozen string literals.

Most of the changes I've been able to test on a live migration with the exception of the string changes.

Would appreciate a review prior to merge.

### DIFF
--- a/script/import_scripts/smf2.rb
+++ b/script/import_scripts/smf2.rb
@@ -259,7 +259,7 @@ class ImportScripts::Smf2 < ImportScripts::Base
     tags.each do |tag|
       post = tag.post
       Post.transaction do
-        post.raw = convert_quotes(post.raw)
+        post.raw = convert_bbcode(post.raw)
         post.rebake!
         post.save
         tag.destroy!
@@ -320,8 +320,8 @@ class ImportScripts::Smf2 < ImportScripts::Base
     end
     body.gsub!(XListPattern) do |s|
       r = +"\n[ul]"
-      s.lines.each { |l| r << '[li]' << l.strip.sub(/^\[x\]\s*/, '') << '[/li]' }
-      r << "[/ul]\n"
+      s.lines.each { |l| "#{r}[li]#{l.strip.sub(/^\[x\]\s*/, '')}[/li]" }
+      "#{r}[/ul]\n"
     end
 
     if attachments.present?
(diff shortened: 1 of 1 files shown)


script/import_scripts/smf2.rb
```diff
@@ -250,6 +250,23 @@ def import_attachment(post, attachment)
...
 
   def postprocess_posts
+    puts '', 'converting bbcode'
+
+    current = 0
+    max = Post.count
+
+    Post.find_each do |post|
+      begin
+        new_raw = convert_bbcode(post.raw.dup)
```
ZogStriP: Shouldn't this be done during the import? (ie. in `import_posts`)

justindirose: Ideally, yes. I was having trouble getting that to directly work while running the import. I'm not sure if that was due to encoding errors that weren't getting thrown or not. This PR is an effort to try to consolidate my improvements after the fact because they got to be a bit of a mess.

justindirose: I added this in to happen after quote conversion and in the postprocess section. You can change a variable to handle quotes later and mark those posts for rebaking. The reason for doing this is the bbcode conversion would need to happen after quote conversion to make sure all the quoted elements are handled appropriately.

ZogStriP: Right, quotes often have to be processed afterwards.

script/import_scripts/smf2.rb
```diff
@@ -260,6 +260,7 @@ def postprocess_posts
       post = tag.post
       Post.transaction do
         post.raw = convert_quotes(post.raw)
+        post.raw = convert_bbcode(post.raw.dup)
```
ZogStriP: Why not call `convert_quotes` inside `convert_bbcode` and then just do

```ruby
post.raw = convert_bbcode(post.raw)
```

script/import_scripts/smf2.rb
```diff
@@ -336,16 +337,21 @@ def convert_message_body(body, attachments = [], **opts)
...
           end
         end
       end
     end
 
-    opts[:ignore_quotes] ? body : convert_quotes(body)
+    if opts[:ignore_quotes]
+      body
+    else
+      body = convert_quotes(body)
```
ZogStriP: Same here, I would move `convert_quotes` inside `convert_bbcode` and just have one call

```ruby
body = convert_bbcode(body)
```

PR #9008, opened 2020-02-20
DEV: Add RakeHelpers.print_message to wrap puts

martin-brennan:
* Rake tasks often have a heavy output to inform the person
  running it what is going on. however if we want to test the
  rake task we don't want all that output clogging up the CI
  build logs.

  RakeHelpers.print_message can be used in place of puts in rake
  tasks, and when running the spec for the task the output is
  not printed RAILS_ENABLE_TEST_LOG is switched on

### DIFF
--- a/lib/rake_helpers.rb
+++ b/lib/rake_helpers.rb
@@ -10,4 +10,10 @@ class RakeHelpers
     return if Rails.env.test? && !ENV['RAILS_ENABLE_TEST_LOG']
     print "\r\033[K%9d / %d (%5.1f%%)" % [current, max, ((current.to_f / max.to_f) * 100).round(1)]
   end
+
+  def self.print_message(*messages)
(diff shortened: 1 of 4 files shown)


martin-brennan:
Hey @ZogStriP can you please take a look at this; my take on fixing the fly-bazooka problem here https://review.discourse.org/t/try-fix-upload-spec-flakys-and-remove-logging-from-tasks-uploads-spec/9174/2?u=martin

SamSaffron:
I am not sure about this ... you can use this pattern instead: 

https://github.com/discourse/discourse/blob/a3d576534a801ad3afe468b873d2a77770272160/spec/components/s3_inventory_spec.rb#L114-L116

Then we don't need to remember to use special helpers.

martin-brennan:
@SamSaffron ah I totally missed that that was a thing...will close this

PR #9009, opened 2020-02-20
FIX: Only mark attachments as secure media if SiteSetting.secure_media?

martin-brennan:
* Attachments (non media files) were being marked as secure if just
  SiteSetting.prevent_anons_from_downloading_files was enabled. this
  was not correct as nothing should be marked as actually "secure" in
  the DB without that site setting enabled
* Also add a proper standalone spec file for the upload security class


SamSaffron:
looks good to me

PR #9011, opened 2020-02-20
DEV: Deprecate `ember` module imports

CvX:
I don't think we'd like to support `import Ember from "ember"` much longer.
It be great to remove the remaining uses and deprecate it now, in 2.4, so it can be removed in 2.5.

---

1. We polyfill `Array.isArray`, no need for Ember utils
2. ember-shim isn't a transpiled module, so the warning has to be done by hand instead of importing `discourse-common/lib/deprecated`


CvX:
@eviltrout 

Importing the `ember` shim does work:
```js

```

CvX:
@eviltrout 

1. Importing the `ember` shim does work:

```js
import Ember from "ember";
Ember.debug("test");
// => DEBUG: test
```

2. `import Evented from "@ember/object/evented"` does resolve correctly.
  The `reorder-categories` controller doesn't need it though, since `this.trigger` calls have been removed since their introduction in https://github.com/discourse/discourse/commit/342eba437468b2f98f47dfa0a255a2af607d4b74#diff-d6dbcf4672b1bf573ebb12ec07f81658R61 but that's something for another PR

3. `isArray` in `ember-addons/fmt.js.es6` – looks like we polyfill `Array.isArray` only for use inside `discourse-loader`. `Array.isArray` is widely supported (we're talking IE9+ here: https://caniuse.com/#search=isarray) but to be on the safe side I'm replacing it here with `_.isArray` that I've confirmed is available inside this module (though at some point we'll probably replace some uses of lodash with JS builtins)

4. `import { computed, get } from "@ember/object"` does resolve correctly.

eviltrout:
Okay, sounds good to merge then. But please be around after you do so we can catch any regressions if they occur.

PR #9014, opened 2020-02-20
FIX: downloaded image URLs incorrectly replaced in post raw.

vinothkannans:
Previously, while replacing the downloaded image URL `http://wiki.mozilla.org/images/2/2e/Longcat1.png` similar non-image URL `http://wiki.mozilla.org/images/2` was replaced wrongly.


vinothkannans:
The issue is found in the meta post [here](https://meta.discourse.org/t/discourse-a2a-bbcode/140864?u=vinothkannans).

PR #9019, opened 2020-02-21
FEATURE: Publish a message when reviewable claimer changes

udan11:


app/assets/javascripts/discourse/routes/review-index.js.es6
```diff
@@ -30,6 +31,24 @@ export default DiscourseRoute.extend({
     });
   },
 
+  activate() {
+    this.messageBus.subscribe("/reviewable_claimed", data => {
+      const reviewables = this.controller.reviewables;
+      if (reviewables) {
+        const user = data.user ? User.create(data.user) : null;
```
eviltrout: This will skip the identity map in the store that I suspect is already used for `claimed_by`. Could you use a `store` method to rehyrdrate the user?

Also, does this work when it's unclaimed, by setting `claimed_by` to null?

app/controllers/reviewable_claimed_topics_controller.rb
```diff
@@ -6,13 +6,18 @@ class ReviewableClaimedTopicsController < ApplicationController
   def create
     topic = Topic.find_by(id: params[:reviewable_claimed_topic][:topic_id])
     guardian.ensure_can_claim_reviewable_topic!(topic)
-    ReviewableClaimedTopic.create(user_id: current_user.id, topic_id: topic.id)
+
+    begin
+      ReviewableClaimedTopic.create!(user_id: current_user.id, topic_id: topic.id)
+    rescue ActiveRecord::RecordInvalid
+      return render_json_error(I18n.t('reviewables.conflict'), status: 409)
```
eviltrout: Nice

app/controllers/reviewable_claimed_topics_controller.rb
```diff
@@ -26,6 +31,26 @@ def destroy
...
     end
 
+    notify_users(topic, nil)
     render json: success_json
   end
+
+  private
+
+  def notify_users(topic, claimed_by)
+    user_ids = User.staff.pluck(:id)
```
eviltrout: What about in the case of category moderation or when moderators cannot see it? We can't assume that only staff reviews content.

A reviewable has fields `reviewable_by_moderator` and `reviewable_by_group` to handle this.

app/assets/javascripts/discourse/routes/review-index.js.es6
udan11: Thanks for the `store` suggestion, that is really the best way.

Yes, it works when `claimed_by` is set to null.

PR #9020, opened 2020-02-21
FIX: When appending tags to restricted category posts, we need to pass the category id, or it won't work.

romanrizzi:


eviltrout:
Is there any way you could add a test for this?

PR #9029, opened 2020-02-24
FIX: ensures we don't attempt to create a new PM on an existing topic

jjaffeux:
This fix attempts to both fix it at UI level and server side. A previous attempt related to this behavior has been made in commit: https://github.com/discourse/discourse/commit/49c750ca7890490b1405f940b1eb2e7a4f3f69f6

### DIFF
--- a/app/assets/javascripts/discourse/models/composer.js.es6
+++ b/app/assets/javascripts/discourse/models/composer.js.es6
@@ -912,7 +912,7 @@ const Composer = RestModel.extend({
   },
 
   createPost(opts) {
-    if (this.action === CREATE_TOPIC) {
+    if (CREATE_TOPIC === this.action || PRIVATE_MESSAGE === this.action) {
       this.set("topic", null);
     }
 
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -137,11 +137,9 @@ class PostCreator
     else
       @topic = Topic.find_by(id: @opts[:topic_id])
 
-      if @topic.present?
-        if @opts[:archetype] == Archetype.private_message
-          errors.add(:base, I18n.t(:create_pm_on_existing_topic))
-          return false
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/discourse/models/composer.js.es6
```diff
@@ -912,7 +912,7 @@ const Composer = RestModel.extend({
   },
 
   createPost(opts) {
-    if (this.action === CREATE_TOPIC) {
+    if (CREATE_TOPIC === this.action || PRIVATE_MESSAGE === this.action) {
```
ZogStriP: PSA: when sending the information to the server, we need to make sure the "topic_id" is [null/blank](https://github.com/discourse/discourse/blob/31f3ed8d3610b63c7b1146e1a67ee51b703dbc41/lib/post_creator.rb#L593-L595), otherwise, the server will think it's a reply instead of a new topic/message.

lib/post_creator.rb
```diff
@@ -136,6 +136,14 @@ def valid?
       return false unless skip_validations? || validate_child(topic_creator)
     else
       @topic = Topic.find_by(id: @opts[:topic_id])
+
+      if @topic.present?
+        if @opts[:archetype] == Archetype.private_message
```
ZogStriP: PSA: We added this test to ensure that any content that were supposed to be "private" never gets public.

lib/post_creator.rb
```diff
@@ -136,6 +136,14 @@ def valid?
       return false unless skip_validations? || validate_child(topic_creator)
     else
       @topic = Topic.find_by(id: @opts[:topic_id])
+
+      if @topic.present?
```
markvanlan: Not a big deal, but you could just `&&` there 2 conditions to avoid a nested if.

PR #9030, opened 2020-02-24
FIX: Translate none-tag and all-tags labels in tag filter

pmusaraj:


SamSaffron:
This looks correct to me, since we want this to land for next release I am merging this in @jjaffeux

jjaffeux:
yes should be good

PR #9037, opened 2020-02-25
FIX: Use full URL for secure attachments when secure media enabled

martin-brennan:
When secure media is enabled and an attachment is marked as secure we want to use the full url instead of the short-url so we get the same access control post protections as secure media uploads


app/assets/javascripts/pretty-text/upload-short-url.js.es6
```diff
@@ -52,6 +51,26 @@ function retrieveCachedUrl($upload, dataAttribute, callback) {
...
+
+  // non-attachments always use the full URL
+  if (dataAttribute !== "orig-href") {
+    return cachedUpload.url;
+  }
+
+  // attachments should use the full /secure-media-uploads/ URL
+  // in this case for permission checks
+  if (
+    Discourse.SiteSettings.secure_media &&
```
eviltrout: Is there a way to avoid using the `Discourse.SiteSettings` global here? It is deprecated and should be passed into the pretty text rendering engine already.

martin-brennan: Ah yep I had a feeling it might have been deprecated. I was able to get the site settings a different way in a different area; I will change this as well :+1:

app/assets/javascripts/pretty-text/upload-short-url.js.es6
```diff
@@ -97,31 +118,31 @@ function _loadCachedShortUrls($uploads) {
...
-    _loadCachedShortUrls($shortUploadUrls);
+    _loadCachedShortUrls($shortUploadUrls, siteSettings);
 
     $shortUploadUrls = $(attributes);
     if ($shortUploadUrls.length > 0) {
       // this is carefully batched so we can do a leading debounce (trigger right away)
       return debounce(
         null,
-        () => _loadShortUrls($shortUploadUrls, ajax),
+        () => _loadShortUrls($shortUploadUrls, ajax, siteSettings),
```
jjaffeux: are you sure this debouncing is working as you expect?

martin-brennan: I am not sure, the debounce was there before. All I added was the siteSettings @jjaffeux

eviltrout: I think if you didn't add the debounce it's safe to merge, but maybe a todo for the future @jjaffeux

eviltrout:
Feel free to merge once conflicts are resolved.

PR #9039, opened 2020-02-25
FIX plugin generator: mobile, desktop stylesheets registering

spirobel:
After adding styles in the mobile scss they will also be applied to desktop without this change.
https://meta.discourse.org/t/registering-assets-for-mobile-and-desktop-making-site-responsive/65041/2?u=spirobel


ZogStriP:
Thanks 👍

PR #9040, opened 2020-02-25
FIX: Make reviewable claiming work with deleted topics

udan11:
This fixes the case when the reviewed topic is deleted (for example, in
discourse-akismet).


spec/requests/reviewable_claimed_topics_controller_spec.rb
```diff
@@ -36,6 +36,17 @@
         expect(messages[0].data[:user][:id]).to eq(moderator.id)
       end
 
+      it "works with deleted topics" do
```
ZogStriP: Maybe we should also add a test for the `destroy` action?

PR #9042, opened 2020-02-25
FIX: ensures destroying a user with security keys doesn't fail

jjaffeux:


spec/models/user_spec.rb
```diff
@@ -2220,6 +2220,16 @@ def filter_by(method)
     end
   end
 
+  describe "Destroying a user with security key" do
+    let!(:security_key) { Fabricate(:user_security_key_with_random_credential, user: user) }
+    fab!(:admin) { Fabricate(:admin) }
```
ZogStriP: Isn't `admin` fabricated "earlier"?

ZogStriP:
> This looks good to me. It should be safe to bring in to 2.4, do you agree?

I think it's safe to bring in 2.4

spec/models/user_spec.rb
jjaffeux: not in a parent block, there are other describe at the same hierarchy level with an admin being fabricated, but that would work right?

PR #9046, opened 2020-02-26
FEATURE: auto archive group message if topic is closed.

vinothkannans:


app/models/topic.rb
```diff
@@ -502,6 +502,17 @@ def self.similar_to(title, raw, user = nil)
   def update_status(status, enabled, user, opts = {})
     TopicStatusUpdater.new(self, user).update!(status, enabled, opts)
     DiscourseEvent.trigger(:topic_status_updated, self, status, enabled)
+
+    if private_message? && ["closed", "autoclosed"].include?(status.to_s) && enabled
```
ZogStriP: Micro-optimization but I  always try to do the "lightest" tests in the beginning have the "heaviest" tests towards the end.

The `status.to_s["closed"]` might be a little too clever though.

```suggestion
    if enabled && private_message? && status.to_s["closed"]
```

PR #9048, opened 2020-02-26
FIX: Add aria-labels to topic list items

eviltrout:
Before this fix you could navigate the topic list using a screen reader
and a keyboard but some of the items were not as descriptive as they
could be. The newly added labels make it easier to understand what you
are tabbing over.

context:
https://meta.discourse.org/t/accessibility-aria-attributes-are-not-defined-for-links-under-replies-category/142539

### DIFF
--- a/app/assets/javascripts/discourse/helpers/application.js.es6
+++ b/app/assets/javascripts/discourse/helpers/application.js.es6
@@ -41,6 +41,13 @@ registerUnbound("number", (orig, params) => {
   if (n.toString() !== title.toString() && addTitle) {
     result += " title='" + Handlebars.Utils.escapeExpression(title) + "'";
   }
+  if (params.ariaLabel) {
+    result +=
+      " aria-label='" +
+      Handlebars.Utils.escapeExpression(params.ariaLabel) +
+      "'";
+  }
+
   result += ">" + n + "</span>";
 
   return new safe(result);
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/helpers/application.js.es6
```diff
@@ -41,6 +41,13 @@ registerUnbound("number", (orig, params) => {
   if (n.toString() !== title.toString() && addTitle) {
     result += " title='" + Handlebars.Utils.escapeExpression(title) + "'";
   }
+  if (params.ariaLabel) {
+    result +=
+      " aria-label='" +
+      Handlebars.Utils.escapeExpression(params.ariaLabel) +
+      "'";
```
ZogStriP: I don't really like this string spanning 4 lines... Maybe we could write it this way?

```javascript
const ariaLabel = Handlebars.Utils.escapeExpression(params.ariaLabel);
result += ` aria-label='${ariaLabel}'`;
```

app/assets/javascripts/discourse/lib/utilities.js.es6
```diff
@@ -66,9 +66,12 @@ export function avatarImg(options, getURL) {
 
   const classes =
     "avatar" + (options.extraClasses ? " " + options.extraClasses : "");
-  const title = options.title
-    ? " title='" + escapeExpression(options.title || "") + "'"
-    : "";
+
+  let title = "";
+  if (options.title) {
+    let escaped = escapeExpression(options.title || "");
```
ZogStriP:
```suggestion
    const escaped = escapeExpression(options.title || "");
```

PR #9052, opened 2020-02-26
FEATURE: Always disable customizations on the `/safe-mode` route

davidtaylorhq:
This makes it easier to enter safe mode when a customization has made the UI unusable

Context: https://meta.discourse.org/t/-/142666/8


davidtaylorhq:
> integration test

✅

PR #9057, opened 2020-02-27
FIX: Polyfill Promise for IE11

davidtaylorhq:
This is required for popper.js


plugins/discourse-internet-explorer/public/js/ie.js
```diff
@@ -970,3 +970,12 @@ if (!String.prototype.startsWith) {
   return ES6;
 });
 /* eslint-enable */
+
+// Polyfill Promise - used by popper.js
+window.addEventListener(
```
eviltrout: It's so nice that this :wastebasket: only lives in the `ie.js` file!

PR #9061, opened 2020-02-27
PERF: skip existing users if large no of notifications expected

vinothkannans:
We shouldn't update default categories for existing users if large no. of notifications are expected.


vinothkannans:
More details at https://meta.discourse.org/t/very-slow-sidekiq-issue-with-large-queue-due-to-massive-numbers-of-unread-user-notifications/140716/79?u=vinothkannans

app/controllers/admin/site_settings_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Admin::SiteSettingsController < Admin::AdminController
+  MAX_NOTIFICATION_COUNT = 10000
```
eviltrout:
```suggestion
  MAX_NOTIFICATION_COUNT ||= 10000
```
This is a handy trick in case the case is re-evaluated for any reason.

SamSaffron:
sorry @vinothkannans  I am closing this as a bit too risky, we have sat on this for quite a while and it churns through some very very critical code. I prefer to hold off on this risk.

PR #9064, opened 2020-02-27
DEV: Run prettier on staged files before commit using lefthook

markvanlan:
I don't know about you, but it annoys me how often I will forget to run prettier or maybe I make a tiny commit on an existing PR and use double quotes instead of single quotes. Then tests fail. 

### Lets catch that early!

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -4,6 +4,9 @@ pre-commit:
     rubocop:
       glob: "*.rb"
       run: bundle exec rubocop {staged_files}
+    prettier:
+      glob: "*.{js,es6}"
+      run: yarn prettier --list-different {staged_files}
     eslint:
       glob: "*.{js,es6}"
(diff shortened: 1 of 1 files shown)


markvanlan:
@eviltrout Are there files that we would want prettier to ignore?

eviltrout:
@markvanlan I think it would be safe to start with the same ones that eslint does.

markvanlan:
@eviltrout I am not sure if this is a problem or not. Some issues (like missing semicolons) are going to get caught from both prettier and eslint

Screenshot from 2020-02-27 15-36-23

We could 
* Change our eslint config so there is less overlap.
* Merge this and both will error
* Close this PR.

eviltrout:
I don't think there's a problem with the same error being reported twice. It's a valid error right, and fixing it once will close the objections from both tools?

PR #9066, opened 2020-02-27
DEV: Replace save-button partial with a component

pmusaraj:
- Refactors user profile preferences screens, and reuses the `save-control` component instead of a partial
- Fixes an issue where the "Saved!" notice after saving changes would persist when navigating away from that specific tag and returning to it

The deleted `user/preferences/save-button` partial is also used in the Discourse Saved Search plugin, see accompanying PR.


app/assets/javascripts/admin/components/save-controls.js.es6
```diff
@@ -3,10 +3,14 @@ import { or } from "@ember/object/computed";
 import Component from "@ember/component";
 
 export default Component.extend({
-  classNames: ["controls"],
+  classNames: ["controls", "save-button"],
 
   buttonDisabled: or("model.isSaving", "saveDisabled"),
 
+  didInsertElement() {
```
jjaffeux: should have a:

```
this._super(...arguments);
```

app/assets/javascripts/discourse/templates/preferences/account.hbs
```diff
@@ -215,11 +215,7 @@
 {{plugin-outlet name="user-custom-controls" args=(hash model=model)}}
 
 {{#if canSaveUser}}
-  <div class="control-group save-button">
-    <div class="controls">
-      {{partial 'user/preferences/save-button'}}
-    </div>
-  </div>
+  {{#save-controls model=model action=(action "save") saved=saved}}{{/save-controls}}
```
jjaffeux: why is save-controls using block form if it's not yielding anything?

pmusaraj:
This needs some more work, will reopen shortly.

pmusaraj:
The plugin spec failure is to expected, it comes from the Saved Searches plugin: 

```
Module Failed: Acceptance: Saved Searches
  Test Failed: Saved Search preferences
    Assertion Failed: Error: Element .save-user not found.
      Expected: true, Actual: false
```

There is a fix for it already, I'll add it once we merge this in core. 

@eviltrout I wanted to add an acceptance test with a non-admin user (since I moved the component out of the `adminjs` path), but it doesn't look like we can test against that.

PR #9071, opened 2020-02-28
FEATURE: Send notifications for time-based and At Desktop bookmark reminders

martin-brennan:
This PR implements the scheduling and notification system for bookmark reminders. Every 5 minutes a schedule runs to check any reminders that need to be sent before now, limited to **300** reminders at a time. Any leftover reminders will be sent in the next run. This is to avoid having to deal with fickle sidekiq and reminders in the far-flung future, which would necessitate having a background job anyway to clean up any missing `enqueue_at` reminders.

If a reminder is sent its `reminder_at` time is cleared and the `reminder_last_sent_at` time is filled in. Notifications are only user-level notifications for now.

All JavaScript and frontend code related to displaying the bookmark reminder notification is contained here. The reminder functionality is now re-enabled in the bookmark modal as well.

image

This PR also implements the "Remind me next time I am at my desktop" bookmark reminder functionality. When the user is on a mobile device they are able to select this option. When they choose this option we set a key in Redis saying they have a pending at desktop reminder. The next time they change devices we check if the new device is desktop, and if it is we send reminders using a DistributedMutex. There is also a job to ensure consistency of these reminders in Redis (in case Redis drops the ball) and the at desktop reminders expire after 20 days.

Also in this PR is a fix to delete all Bookmarks for a user via `UserDestroyer`


martin-brennan:
@SamSaffron if you aren't reviewing yet please hold off, I am almost done with the "At Desktop" reminders now as well, and it would be easier to review all at once I think.

app/controllers/topics_controller.rb
```diff
@@ -548,10 +548,12 @@ def bookmark
     first_post = topic.ordered_posts.first
 
     if SiteSetting.enable_bookmarks_with_reminders?
-      if Bookmark.exists?(user: current_user, post: first_post)
-        return render_json_error(I18n.t("bookmark.topic_already_bookmarked"), status: 403)
+      bookmark_manager = BookmarkManager.new(current_user)
+      bookmark_manager.create(post_id: first_post.id)
+
+      if bookmark_manager.errors
+        return render_json_error(bookmark_manager.errors.full_messages, status: 400)
```
eviltrout: You should be able to say:

```suggestion
        return render_json_error(bookmark_manager, status: 400)
```

I think our error handler is smart enough to extract the standard errors from objects.

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+
+module Jobs
+
+  # Runs periodically to send out bookmark reminders, capped at 300 at a time.
+  # Any leftovers will be caught in the next run, because the reminder_at column
+  # is set to NULL once a reminder has been sent.
+  class BookmarkReminderNotifications < ::Jobs::Scheduled
+    MAX_REMINDER_NOTIFICATIONS_PER_RUN = 300
+    JOB_RUN_NUMBER_KEY = 'jobs_bookmark_reminder_notifications_job_run_num'.freeze
+    AT_DESKTOP_CONSISTENCY_RUN_NUMBER = 6
```
eviltrout:
```suggestion
    MAX_REMINDER_NOTIFICATIONS_PER_RUN ||= 300
    JOB_RUN_NUMBER_KEY ||= 'jobs_bookmark_reminder_notifications_job_run_num'.freeze
    AT_DESKTOP_CONSISTENCY_RUN_NUMBER ||= 6
```

This removes error messages in case the class is reloaded.

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+      bookmarks.limit(MAX_REMINDER_NOTIFICATIONS_PER_RUN).each do |bookmark|
+        BookmarkReminderNotificationHandler.send_notification(bookmark)
+      end
+
+      # we only want to ensure the desktop consistency every X runs of this job
+      # (every 30 mins in this case) so we don't bother redis too much, and the
+      # at desktop consistency problem should not really happen unless people
+      # are setting the "at desktop" reminder, going out for milk, and never coming
+      # back
+      current_job_run_number = Discourse.redis.get(JOB_RUN_NUMBER_KEY)&.to_i || 0
```
eviltrout:
```suggestion
      current_job_run_number = Discourse.redis.get(JOB_RUN_NUMBER_KEY).to_i
```
`nil.to_i` returns 0

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+      Discourse.redis.set(JOB_RUN_NUMBER_KEY, new_job_run_number)
+    end
+
+    def ensure_at_desktop_consistency
+      pending_at_desktop_bookmark_reminders = \
+        Bookmark.includes(:user)
+          .references(:user)
+          .pending_at_desktop_reminders
+          .where('users.last_seen_at >= :one_day_ago', one_day_ago: Time.now.utc - 24.hours)
```
eviltrout:
```suggestion
          .where('users.last_seen_at >= :one_day_ago', one_day_ago: 1.day.ago.utc)
```

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      unique_users = pending_at_desktop_bookmark_reminders.map(&:user).uniq
+      pending_reminders_for_redis_check = unique_users.map do |user|
+        "#{BookmarkReminderNotificationHandler::PENDING_AT_DESKTOP_KEY_PREFIX}#{user.id}"
+      end
+
+      Discourse.redis.mget(pending_reminders_for_redis_check).each.with_index do |value, idx|
+        next if value.present?
+        user_id = pending_reminders_for_redis_check[idx][/\d+/].to_i
+        user = unique_users.find { |u| u.id == user_id }
```
eviltrout: This involves looping through all the `unique_users` for each iteration. If that array is only ever small it's probably fine, but if you expect more than a handful of users to be in it, I recommend converting `unique_users` to be a `Hash` so you can look it up in constant time.

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+          bookmark.user == user
+        end
+
+        user_expired_bookmark_reminders = user_pending_bookmark_reminders.select do |bookmark|
+          bookmark.reminder_set_at <= expiry_limit_datetime
+        end
+
+        next if user_pending_bookmark_reminders.length == user_expired_bookmark_reminders.length
+
+        # only put tell the cache-gods that this user has pending "at desktop" reminders
```
eviltrout:
```suggestion
        # only tell the cache-gods that this user has pending "at desktop" reminders
```

app/jobs/scheduled/bookmark_reminder_notifications.rb
```diff
@@ -0,0 +1,101 @@
...
+      Bookmark
+        .pending_at_desktop_reminders
+        .where('reminder_set_at <= :expiry_limit_datetime', expiry_limit_datetime: expiry_limit_datetime)
+        .update_all(
+          reminder_set_at: nil, reminder_type: nil
+        )
+    end
+
+    def expiry_limit_datetime
+      Time.now.utc - BookmarkReminderNotificationHandler::PENDING_AT_DESKTOP_EXPIRY_DAYS.days
```
eviltrout:
```suggestion
      BookmarkReminderNotificationHandler::PENDING_AT_DESKTOP_EXPIRY_DAYS.days.ago.utc
```

app/models/bookmark.rb
```diff
@@ -10,6 +10,41 @@ class Bookmark < ActiveRecord::Base
...
+
+  def unique_per_post_for_user
+    existing_bookmark = Bookmark.find_by(user_id: user_id, post_id: post_id)
+    return if existing_bookmark.blank? || existing_bookmark.id == id
+    self.errors.add(:base, I18n.t("bookmarks.errors.already_bookmarked_post"))
+  end
+
+  def ensure_sane_reminder_at_time
+    return if reminder_at.blank?
+    if reminder_at < Time.now.utc
```
eviltrout:
```suggestion
    if reminder_at < Time.zone.now
```
should be in UTC and we are moving to `Time.zone.now` in our app everywhere.

app/models/bookmark.rb
```diff
@@ -10,6 +10,41 @@ class Bookmark < ActiveRecord::Base
...
+      self.errors.add(:base, I18n.t("bookmarks.errors.cannot_set_past_reminder"))
+    end
+    if reminder_at > (Time.now.utc + 10.years)
```
eviltrout:
```suggestion
    if reminder_at > 10.years.from_now.utc
```

app/serializers/post_serializer.rb
```diff
@@ -338,6 +343,10 @@ def bookmark_reminder_at
     post_bookmark&.reminder_at
   end
 
+  def bookmark_reminder_type
+    Bookmark.reminder_types[post_bookmark&.reminder_type].to_s
```
eviltrout: This is a bit odd - do we want to look up `nil` in a hash then call `to_s` on the resulting `nil`? Should be an empty string but it seems weird to me.

lib/bookmark_manager.rb
```diff
@@ -0,0 +1,74 @@
...
+    reminder_type = Bookmark.reminder_types[reminder_type.to_sym] if reminder_type.present?
+
+    bookmark = Bookmark.create(
+      user_id: @user.id,
+      topic_id: topic_id_for_post(post_id),
+      post_id: post_id,
+      name: name,
+      reminder_type: reminder_type,
+      reminder_at: reminder_at,
+      reminder_set_at: Time.now.utc
```
eviltrout:
```suggestion
      reminder_set_at: Time.zone.now
```

lib/bookmark_manager.rb
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    if bookmark.errors.any?
+      add_errors_from(bookmark)
+      return
```
eviltrout:
```suggestion
      return add_errors_from(bookmark)
```
(super trivial!!!)

lib/bookmark_manager.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def self.send_reminder_notification(id)
+    bookmark = Bookmark.find_by(id: id)
+    BookmarkReminderNotificationHandler.send_notification(bookmark)
+  end
+
+  private
+
+  def topic_id_for_post(post_id)
+    Post.select(:topic_id).find(post_id).topic_id
```
eviltrout:
```suggestion
    Post.where(id: post_id).pluck_first(:topic_id)
```
`select` is dangerous and pluck is faster in this case because we don't instantiate active record objects.

lib/bookmark_reminder_notification_handler.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class BookmarkReminderNotificationHandler
+  PENDING_AT_DESKTOP_KEY_PREFIX = 'pending_at_desktop_bookmark_reminder_user_'.freeze
+  PENDING_AT_DESKTOP_EXPIRY_DAYS = 20
```
eviltrout:
```suggestion
  PENDING_AT_DESKTOP_KEY_PREFIX ||= 'pending_at_desktop_bookmark_reminder_user_'.freeze
  PENDING_AT_DESKTOP_EXPIRY_DAYS ||= 20
```

lib/bookmark_reminder_notification_handler.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def self.clear_reminder(bookmark)
+    Rails.logger.debug(
+      "Clearing bookmark reminder for bookmark_id #{bookmark.id}. reminder info: #{bookmark.reminder_at} | #{Bookmark.reminder_types[bookmark.reminder_type]}"
+    )
+
+    bookmark.update(
+      reminder_at: nil,
+      reminder_type: nil,
+      reminder_last_sent_at: Time.now.utc,
```
eviltrout:
```suggestion
      reminder_last_sent_at: Time.zone.now
```

lib/bookmark_reminder_notification_handler.rb
```diff
@@ -0,0 +1,74 @@
...
+      data: {
+        topic_title: bookmark.topic.title,
+        display_username: user.username,
+        bookmark_name: bookmark.name
+      }.to_json
+    )
+  end
+
+  def self.user_has_pending_at_desktop_reminders?(user)
+    Discourse.redis.get("#{PENDING_AT_DESKTOP_KEY_PREFIX}#{user.id}").present?
```
eviltrout:
```suggestion
    Discourse.redis.exists("#{PENDING_AT_DESKTOP_KEY_PREFIX}#{user.id}")
```

spec/jobs/bookmark_reminder_notifications_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    it "is not included in the run, because it is not time-based" do
+      BookmarkManager.any_instance.expects(:send_reminder_notification).with(bookmark1.id).never
+      subject.execute
+    end
+  end
+
+  context "when the number of notifications exceed MAX_REMINDER_NOTIFICATIONS_PER_RUN" do
+    it "does not send them in the current run, but will send them in the next" do
+      begin
+        original_const = Jobs::BookmarkReminderNotifications::MAX_REMINDER_NOTIFICATIONS_PER_RUN
```
eviltrout: This is a bit awkward - I would prefer if we added a class variable for this and this code was changed to be a setter.

app/jobs/scheduled/bookmark_reminder_notifications.rb
martin-brennan: Great point, I don't expect there to be many but a hash is still so much faster here. I was amazed at how fast hash lookups are when I first found that trick

lib/bookmark_manager.rb
martin-brennan: Didn't know about `pluck_first` thanks!

spec/jobs/bookmark_reminder_notifications_spec.rb
martin-brennan: So you mean changing this to @@max_reminder_notifications_per_run in the class and then passing in a different value via the execute method of the job (if neeeded, otherwise default like it does now to 300)?

martin-brennan:
> One thing that kept coming up is utc references and Time.now. Unless there's a specific reason this code needs utc I think Time.zone.now should suffice.....we are moving to Time.zone.now in our app everywhere.

This one was a little odd to me, shouldn't all datetimes in the DB use UTC? Especially things like reminders that would need to be set based on the user's timezone (which they are + stored in UTC in the DB). Time.zone.now (should) return UTC on the server I guess but not everywhere else. There is probably a good reason for this I just need a little more explanation

spec/jobs/bookmark_reminder_notifications_spec.rb
eviltrout: Yes, it could be changed to have a getter/setter and the test could call `Jobs::BookmarkReminderNotifications.max_reminder_notifications_per_run = 2`

eviltrout:
> This one was a little odd to me, shouldn't all datetimes in the DB use UTC?

I could be mistaken here, but as I understand it ActiveRecord already stores everything in UTC in the database and you don't need to convert it. Maybe you could investigate this and confirm?

martin-brennan:
> > This one was a little odd to me, shouldn't all datetimes in the DB use UTC?
> 
> I could be mistaken here, but as I understand it ActiveRecord already stores everything in UTC in the database and you don't need to convert it. Maybe you could investigate this and confirm?

Ah I think you may be right there. I am just super paranoid about timezone stuff because it has bitten me before. I found this in-depth article on timezones in rails that seems to confirm that it is UTC by default https://nandovieira.com/working-with-dates-on-ruby-on-rails. I'll change my code and see if I can notice any difference!

PR #9072, opened 2020-02-28
DEV: adds a loading property to d-button

jjaffeux:
Usage:

```
{{d-button icon="times" label="foo.bar" isLoading=true}}
```

Note that  a button loading without an icon will shrink text size to prevent button to jump in size.

A button while loading is disabled.


jjaffeux:
@awesomerobot carefully review this please, as I changed d-button from inline-block to inline-flex

jjaffeux:
@awesomerobot @eviltrout I have updated the css code and changed from `loading` to `isLoading` to be consistent with {{conditional-loading-section}}

Should now be ready to merge :+1:

Screenshot 2020-03-29 at 16 07 11

PR #9073, opened 2020-02-28
FEATURE: Use YouTube's enhanced privacy embed mode by default

davidtaylorhq:


davidtaylorhq:
This version of the embed removes some useful features like 'watch later', so it is not a suitable default

PR #9074, opened 2020-02-28
PERF: Allow user serializer to make use of preloaded custom fields

davidtaylorhq:
This will be used to allow serializing multiple user cards without N+1 queries

### DIFF
--- a/app/serializers/user_card_serializer.rb
+++ b/app/serializers/user_card_serializer.rb
@@ -149,14 +149,14 @@ class UserCardSerializer < BasicUserSerializer
   end
 
   def custom_fields
-    fields = User.whitelisted_user_custom_fields(scope)
-
-    if scope.can_edit?(object)
-      fields += DiscoursePluginRegistry.serialized_current_user_fields.to_a
-    end
+    fields = custom_field_keys
 
     if fields.present?
-      User.custom_fields_for_ids([object.id], fields)[object.id] || {}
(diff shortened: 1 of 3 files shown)


app/serializers/user_card_serializer.rb
```diff
@@ -197,4 +197,11 @@ def featured_topic
   def card_background_upload_url
     object.card_background_upload&.url
   end
+
+  private
+
+  def custom_field_keys
+    # Can be extended by other serializers
```
eviltrout: If it's meant to be extended shouldn't the method be `protected`?

eviltrout: Also this isn't calling `super` so isn't the logic above about `can_edit` ignored?

davidtaylorhq: I don't follow, why is `protected` required for inheritence?

Also, the inheritence goes `UserSerializer < UserCardSerializer`, so `UserSerializer` is the one which calls `super`.

eviltrout: Thanks I assumed that card extended user and not the inverse.

As for protected vs private I think that's my Java/C++ background. I always use `protected` to indicate that a method is meant to be used by the class and all its subclasses. In ruby you're right it doesn't matter, private allows that.

`protected` is somewhat useless in ruby I think as a result :) You can keep it this way if you like and consider my labelling antiquated :)

PR #9075, opened 2020-02-28
PERF: Cache subcategory IDs to avoid recursive CTE

udan11:


app/models/category.rb
```diff
@@ -73,6 +73,7 @@ class Category < ActiveRecord::Base
   after_save :publish_discourse_stylesheet
   after_save :publish_category
   after_save :reset_topic_ids_cache
+  after_save :clear_subcategory_ids_cache
```
udan11: It would be ideal to clear the cache only for new records or if `parent_category_id` changed, but I wanted to play it safe for now.

eviltrout:
I'll leave this review to @SamSaffron but I should note that despite the perf improvement here it does open an new class of bugs with a stale cache. If the message bus can't send the message for whatever reason some processes will be out of date.

That might not be so bad because eventually the cache will update, but it's something we should factor into our decision here about whether to accept this. Are we ok with this data being a little stale?

SamSaffron:
OK I went with https://github.com/discourse/discourse/commit/b4999acadd3783e0953f04729cafc2fe0598db87 which is one more query but avoids a global cache. 

This solution is great, but has a downside, when a forum has enormous amounts of categories this means that any category change will blast out a large update on message bus. We can introduce a cache here later on if we want.

PR #9076, opened 2020-02-28
PERF: Allow preloading 'recent time read' for a user

davidtaylorhq:
This will be used when serializing multiple user cards


app/models/user.rb
```diff
@@ -1207,10 +1207,22 @@ def secondary_emails
     self.user_emails.secondary.pluck(:email)
   end
 
+  RECENT_TIME_READ_THRESHOLD = 60.days
```
eviltrout:
```suggestion
  RECENT_TIME_READ_THRESHOLD ||= 60.days
```
This helps in case the class is reloaded for any reason.

PR #9080, opened 2020-02-28
FIX: We need to skip users with associated reviewables when auto-approving

romanrizzi:


spec/initializers/track_setting_changes_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'Setting changes' do
+  describe '#must_approve_users' do
+    before { SiteSetting.must_approve_users = false }
+
+    it 'does not approve a users with an associated reviewables' do
```
eviltrout:
```suggestion
    it 'does not approve a user with associated reviewables' do
```

spec/initializers/track_setting_changes_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      user_pending_approval = Fabricate(:reviewable_user).target
+
+      SiteSetting.must_approve_users = true
+
+      expect(user_pending_approval.reload.approved?).to eq(false)
+    end
+
+    it 'approves a users with no associated reviewables' do
```
eviltrout:
```suggestion
    it 'approves a user with no associated reviewables' do
```

PR #9086, opened 2020-03-01
DEV: Use aliased `inject` imports

CvX:
This is a common Ember idiom for easier differentiation between controller and service imports.


eviltrout:
This is excellent. Let's merge later this week.

PR #9089, opened 2020-03-02
FEATURE: Enable service worker on iOS PWA

xfalcox:


xfalcox:
@pmusaraj DiscourseHub doesn't support webview so I think this won't impact it.

PR #9095, opened 2020-03-02
FIX: Check if auth token exists before revocation

nbianca:


nbianca:
That error happens when the user has two open tabs and revokes the same token from both tabs.

PR #9096, opened 2020-03-03
FIX: Stop infinite lookup-urls issue for video/audio on page

martin-brennan:
Meta report: https://meta.discourse.org/t/excessive-requests-to-uploads-lookup-urls-leading-to-429-response/143119

* The data-orig-src attribute was not being removed from cooked
  video and audio so the composer was infinitely trying to get the
  URLs for them, which would never resolve to anything
* Also the code that retrieved the short URL was unscoped, and was
  getting everything on the page. if running from the composer we
  now scope to the preview window
* Also fixed a minor issue where the <a> element href for the video
  and audio tags was not being set when the short URL was found


coding-horror:
this should be backported probably??

PR #9097, opened 2020-03-03
FEATURE: Allow themes to specify modifiers in their about.json file

davidtaylorhq:
Modifiers are specified in the `about.json` file like

```
{
...
  "modifiers": {
    "serialize_topic_excerpts": true,
    "svg_icons": ["newicon1", "newicon2"],
    "csp_extensions": ["script-src: https://from-theme-flag.script"],
  },
...
}
```
There are three modifiers:
- serialize_topic_excerpts (boolean)
- csp_extensions (array of strings)
- svg_icons (array of strings)

When multiple themes are active, the values will be combined. The combination method varies based on the setting. CSP/SVG arrays will be combined. serialize_topic_excerpts will use `Enumerable#any`.


davidtaylorhq:
Thanks @eviltrout , renamed to 'modifiers'

PR #9098, opened 2020-03-03
FIX: Dismiss notifications on middle click

nbianca:


app/assets/javascripts/discourse/widgets/default-notification-item.js.es6
```diff
@@ -154,6 +155,17 @@ export const DefaultNotificationItem = createWidget(
           );
         }
       });
+    },
+
+    mouseUp(e) {
+      // dismiss notification on middle click
+      if (e.which === 2 && !this.attrs.read) {
+        this.attrs.set("read", true);
+        ajax("/notifications/mark-read", {
```
ZogStriP: Shouldn't we do that in https://github.com/discourse/discourse/blob/de81bf76dc249dd5723d2ab85f87c08da40a0016/app/assets/javascripts/discourse/widgets/default-notification-item.js.es6#L139-L141 instead of in a different event?

app/assets/javascripts/discourse/widgets/default-notification-item.js.es6
```diff
@@ -154,6 +155,17 @@ export const DefaultNotificationItem = createWidget(
           );
         }
       });
+    },
+
+    mouseUp(e) {
+      // dismiss notification on middle click
+      if (e.which === 2 && !this.attrs.read) {
```
SamSaffron: do we need to look at `e.button` as well? https://stackoverflow.com/a/21224428/17174

discoursereviewbot: [Régis Hanol posted](https://review.discourse.org/t/fix-dismiss-notifications-on-middle-click-pr-9098/9514/4):

I think that’s already taken care of by the `wantsNewWindow` method?

app/assets/javascripts/discourse/widgets/default-notification-item.js.es6
nbianca: The `click` event handler is called for left click only, so I had to use `mouseUp` which is triggered on both left and middle clicks.

app/assets/javascripts/discourse/widgets/default-notification-item.js.es6
nbianca: I believe `e.which` is the "normalized" version of `e.button`. `e.button` can be either `2` or `4` for middle clicks, while `e.which` is always `2`. This felt like a more robust solution.

ZogStriP:
Can you also add a test for that?

app/assets/javascripts/discourse/widgets/default-notification-item.js.es6
```diff
@@ -154,6 +155,17 @@ export const DefaultNotificationItem = createWidget(
           );
         }
       });
+    },
+
+    mouseUp(e) {
```
jjaffeux: I think using `event` is much better, because it's way easier to grep for it in the codebase.

ZogStriP:
@nbianca can you fix the conflict and merge?

PR #9100, opened 2020-03-03
FIX: Quick access panel undefined loading property

zcuric:
### This PR
- [x] References the correct `loading` property which is on `state` object instead on `this` 
- [x] Changes the access to state directly instead of passing it around, like in other widgets

### DIFF
--- a/app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
+++ b/app/assets/javascripts/discourse/widgets/quick-access-panel.js.es6
@@ -85,21 +85,21 @@ export default createWidget("quick-access-panel", {
     return limit;
   },
 
-  refreshNotifications(state) {
-    if (state.loading) {
+  refreshNotifications() {
+    if (this.state.loading) {
       return;
     }
 
     if (this.getItems().length === 0) {
-      state.loading = true;
+      this.state.loading = true;
     }
 
     this.findNewItems()
       .then(newItems => this.setItems(newItems))
       .catch(() => this.setItems([]))
       .finally(() => {
-        state.loading = false;
-        state.loaded = true;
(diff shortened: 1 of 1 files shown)


eviltrout:
Can you offer more context as to what bug this fixes? Is there a post on discourse meta for example?

Additionally, could you contribute a test for it? We generally don't accept PRs without tests.

zcuric:
@eviltrout It's in the PR description.  `this.loading` it's undefined, it doesn't exists on `this`. And when the widget is in `state.loading` true, the `return` never happens. That's the bug.

 I can write a test, no problem, but then again this files in the core and it's not covered with tests.  😕 

> We generally don't accept PRs without tests.
That's debatable. 😉

eviltrout:
@zcuric I saw the description but it describes internal application state. My question was how does this manifest as a bug? Does clicking on a thing not work? How could a tester confirm that this fix works?

Also you are wrong about lack of tests - I added a `return` from `refreshNotifications` and qunit tests did fail. I'm not saying the tests couldn't be improved but there is *some* coverage there.

zcuric:
@eviltrout I want to apologise, I might have sound a bit rude. 

The bug is not reproducible from the UI side of things, because from my read of the code, this `if` statement just prevents the rest of the code in `refreshNotifications` function from executing. In original form it doesn't do that, because it's never true, and that's the fix. 

Other commit is just a sort of clean up and aligning this widget to the code style of others i.e. accessing state directly on `this`.

I'm not expert in `ember` nor in `ruby`, I just recently started working on it in project. I like discourse, so I decided to contribute in this way, by fixing small thing that I found. Also, that typo PR #9101 was also one for the consistency around the code base,  but hey...

PR #9101, opened 2020-03-03
FIX: Change atts to attrs

zcuric:
### This PR
- [x] Fixes typo `atts` to `attrs` across the codebase


eviltrout:
I don't consider this a typo really. Both are abbreviations and this PR only seems to change variable names?

PR #9110, opened 2020-03-04
DEV: Reserve webhook event types to be used in plugins

oblakeerickson:
Based on feedback on the following PR's:

https://github.com/discourse/discourse-solved/pull/85

https://github.com/discourse/discourse-assign/pull/61

This commit reserves ID's to be used for webhook event types to ensure
that some other webhook or plugin doesn't end up using the same ID.

### DIFF
--- a/spec/jobs/emit_web_hook_event_spec.rb
+++ b/spec/jobs/emit_web_hook_event_spec.rb
@@ -239,7 +239,7 @@ describe Jobs::EmitWebHookEvent do
       stub_request(:post, post_hook.payload_url)
         .to_return(body: 'OK', status: 200)
 
-      WebHookEventType.all.pluck(:name).each do |name|
+      WebHookEventType.limit(10).pluck(:name).each do |name|
         web_hook_id = Fabricate("#{name}_web_hook").id
 
         expect do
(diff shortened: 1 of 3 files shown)


spec/jobs/emit_web_hook_event_spec.rb
```diff
@@ -239,7 +239,7 @@
       stub_request(:post, post_hook.payload_url)
         .to_return(body: 'OK', status: 200)
 
-      WebHookEventType.all.pluck(:name).each do |name|
+      WebHookEventType.limit(10).pluck(:name).each do |name|
```
eviltrout: Can you clarify what this is doing? Only emitting core webhooks?

oblakeerickson: Correct. It is only selecting the first 10 webhook event types which all happen to be in core.

oblakeerickson: For this to be a "good" test I don't think it even needs to loop through ALL of them. It could probably just grab the first one and validate that this functionality is working.

eviltrout: I think I would prefer that rather than having 10 as a constant here.

PR #9117, opened 2020-03-05
FIX: CSP with subfolder expects the subfolder included in the CDN path.

riking:


lib/content_security_policy/default.rb
```diff
@@ -25,16 +25,18 @@ def initialize
...
       ['/theme-javascripts/',  false, true],
-      ['/svg-sprite/',         false, true],
+      ['/svg-sprite/',         false, :subfolder],
     ]
 
     def script_assets(base = base_url, s3_cdn = GlobalSetting.s3_cdn_url, cdn = GlobalSetting.cdn_url)
       SCRIPT_ASSET_DIRECTORIES.map do |dir, can_use_s3_cdn, can_use_cdn|
         if can_use_s3_cdn && s3_cdn
           s3_cdn + dir
+        elsif (can_use_cdn == :subfolder) && cdn
```
eviltrout: Although ruby lets you do this with duck typing, I am not a fan of having a variable use multiple types. I would prefer them all by symbols than some be booleans and some symbols.

riking:
Need to change to alternate solution.

riking:
Found the alternate solution: highlight-js also expects the subfolder, so just fix theme-javascripts to use the subfolder and switch everything.. to the subfolder.

pmusaraj:
@riking is it possible to add a spec for this?

riking:
I'm having trouble figuring out how we would test the CSP... I found https://github.com/d4l3k/go-csp-engine which is intended for unit testing but it's in a different language.

app/models/javascript_cache.rb
```diff
@@ -8,7 +8,7 @@ class JavascriptCache < ActiveRecord::Base
   before_save :update_digest
 
   def url
-    "#{GlobalSetting.cdn_url}#{GlobalSetting.relative_url_root}/theme-javascripts/#{digest}.js?__ws=#{Discourse.current_hostname}"
+    "#{GlobalSetting.cdn_url}#{Discourse.base_uri}/theme-javascripts/#{digest}.js?__ws=#{Discourse.current_hostname}"
```
davidtaylorhq: Does this make a difference? I'm pretty sure these two things evaluate to the same result. This is from the console on a live subfolder site

```
[1] pry(main)> GlobalSetting.relative_url_root
=> "/forum"
[2] pry(main)> Discourse.base_uri
=> "/forum"
```

davidtaylorhq:
> how we would test the CSP

I would add it in here: https://github.com/discourse/discourse/blob/master/spec/lib/content_security_policy_spec.rb

`Discourse.base_uri` will probably need to be stubbed [like this](https://github.com/discourse/discourse/blob/c63e84dc62c58abc1ead3bec13a2de0de3cb909e/plugins/discourse-narrative-bot/spec/discourse_narrative_bot/new_user_narrative_spec.rb#L641)

app/models/javascript_cache.rb
davidtaylorhq: So... maybe this is already working. And all we need to do it change the CSP :)

Switching to use `base_uri` is a good change anyway, so let's leave it in 👍

riking: The site I diagnosed this on is definitely getting an empty string back for this.

riking:
@davidtaylorhq You mean `set_subfolder()` :)

riking:
Ready to merge once head is green.

PR #9120, opened 2020-03-05
go permalinks: Add path suffixing, query parameter handling

riking:
This allows people to write links like:

```
go/meta
go/meta/c/praise
go/meta/c/praise?order=views
go/meta?safe_mode=no_custom
```

Additionally, change go links to a 302 instead of 301 as we anticipate that forum admins might edit them + avoiding premature optimization.

### DIFF
--- a/app/controllers/permalinks_controller.rb
+++ b/app/controllers/permalinks_controller.rb
@@ -3,35 +3,35 @@
 class PermalinksController < ApplicationController
   skip_before_action :check_xhr, :preload_json
 
+  def go
+    url = request.fullpath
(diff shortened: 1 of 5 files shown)


SamSaffron:
Wait, I thought we said we were just going to fix the client routing, and existing permalink solution so it works on the client?

riking:
As I went to try and test them out using github.com/discourse/discourse-linkify-words I found myself wanting some extra features like extra / payloads and query parameters, just from "okay, I can write go/meta in this box, what else that's similar would I like to write in the box?" Extra path segments and query parameters just kinda... fell out of the desire box

PR #9123, opened 2020-03-06
FIX: Check for permalinks before showing the 404 page

riking:
Implements the "internal routing from posts" part of https://meta.discourse.org/t/feature-interest-survey-go-permalinks/142605

### DIFF
--- a/app/assets/javascripts/discourse/routes/unknown.js
+++ b/app/assets/javascripts/discourse/routes/unknown.js
@@ -1,8 +1,22 @@
-import DiscourseRoute from "discourse/routes/discourse";
 import { ajax } from "discourse/lib/ajax";
+import DiscourseURL from "discourse/lib/url";
+import DiscourseRoute from "discourse/routes/discourse";
 
 export default DiscourseRoute.extend({
-  model() {
-    return ajax("/404-body", { dataType: "html" });
+  model(params, transition) {
+    const path = params.path;
+    return ajax("/permalink-check.json", {
+      data: { path }
+    }).then(results => {
+      if (results.found) {
+        // Avoid polluting the history stack for external links
+        transition.abort();
+        DiscourseURL.routeTo(results.target_url);
+        return "";
+      } else {
+        // 404 body HTML
+        return results.html;
+      }
+    });
   }
 });
--- a/app/controllers/permalinks_controller.rb
+++ b/app/controllers/permalinks_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PermalinksController < ApplicationController
-  skip_before_action :check_xhr, :preload_json
+  skip_before_action :check_xhr, :preload_json, only: [:show]
 
   def show
     url = request.fullpath
@@ -14,6 +14,30 @@ class PermalinksController < ApplicationController
       redirect_to permalink.target_url, status: :moved_permanently
     else
       raise Discourse::NotFound
+    end
+  end
+
+  def check
+    begin
+      raise Discourse::NotFound if params[:path].blank?
+
+      permalink = Permalink.find_by_url(params[:path])
+
+      raise Discourse::NotFound unless permalink
+
+      data = {
+        found: true,
+        internal: permalink.external_url.nil?,
+        target_url: permalink.target_url,
+      }
+
+      render json: MultiJson.dump(data)
+    rescue Discourse::NotFound
+      data = {
+        found: false,
+        html: build_not_found_page(status: 200),
+      }
+      render json: MultiJson.dump(data)
     end
   end
 
--- a/test/javascripts/acceptance/unknown-test.js.es6
+++ b/test/javascripts/acceptance/unknown-test.js.es6
@@ -1,8 +1,49 @@
 import { acceptance } from "helpers/qunit-helpers";
+import pretender from "helpers/create-pretender";
 acceptance("Unknown");
 
-QUnit.test("Unknown URL", async assert => {
-  assert.expect(1);
+QUnit.test("Permalink Unknown URL", async assert => {
   await visit("/url-that-doesn't-exist");
   assert.ok(exists(".page-not-found"), "The not found content is present");
+});
+
+QUnit.test("Permalink URL to a Topic", async assert => {
+  pretender.get("/permalink-check.json", () => {
+    return [
+      200,
+      { "Content-Type": "application/json" },
+      {
+        found: true,
+        internal: true,
+        target_url: "/t/internationalization-localization/280"
+      }
+    ];
+  });
+
+  await visit("/viewtopic.php?f=8&t=280");
+  assert.ok(exists(".topic-post"));
+});
+
+QUnit.test("Permalink URL to a static page", async assert => {
+  pretender.get("/permalink-check.json", () => {
+    return [
+      200,
+      { "Content-Type": "application/json" },
+      {
+        found: true,
+        internal: true,
+        target_url: "/faq"
+      }
+    ];
+  });
+
+  await visit("/not-the-url-for-faq");
+
+  // body is outside of #ember-testing-container and needs to be targeted
+  // through document instead of find
+  assert.ok(
+    document.querySelector("body.static-faq"),
+    "routed to the faq page"
+  );
+  assert.ok(exists(".body-page"));
 });
--- a/app/assets/javascripts/discourse/lib/url.js
+++ b/app/assets/javascripts/discourse/lib/url.js
@@ -24,7 +24,6 @@ const SERVER_SIDE_ONLY = [
   /^\/posts\/\d+\/raw/,
   /^\/raw\/\d+/,
   /^\/wizard/,
-  /^\/go\//, // EXPERIMENTAL: https://meta.discourse.org/t/-/142605
   /\.rss$/,
   /\.json$/,
   /^\/admin\/upgrade$/,
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -4578,7 +4578,7 @@ en:
 
       permalink:
         title: "Permalinks"
-        description: "Please note this only applies to external sources, links posted on your forum will not be redirected."
+        description: "Redirections to apply for URLs not known by the forum."
         url: "URL"
         topic_id: "Topic ID"
         topic_title: "Topic"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -954,6 +954,8 @@ Discourse::Application.routes.draw do
 
   resources :csp_reports, only: [:create]
 
+  get "/permalink-check", to: 'permalinks#check'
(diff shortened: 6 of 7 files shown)


SamSaffron:
This certainly needs testing (both client and server), I think the approach is sound

app/assets/javascripts/discourse/routes/unknown.js.es6
```diff
@@ -1,8 +1,28 @@
...
 
 export default DiscourseRoute.extend({
-  model() {
-    return ajax("/404-body", { dataType: "html" });
+  model(params) {
+    const bodyReq = ajax("/404-body", { dataType: "html" });
+    const testReq = ajax("/permalink-check.json", {
+      data: { path: params.path }
+    });
+    const self = this;
```
eviltrout: We prefer to use the fat arrow rather than `self` - looks like it should be possible in this case.

riking:
JS test added. Couldn't test external URL properly because it redirects the entire test page.

riking:
Test failures are real.

test/javascripts/acceptance/unknown-test.js.es6
```diff
@@ -1,8 +1,49 @@
...
+      {
+        found: true,
+        internal: true,
+        target_url: "/faq"
+      }
+    ];
+  });
+
+  await visit("/not-the-url-for-faq");
+  assert.ok($("body.static-faq").length, "routed to the faq page");
```
jjaffeux: We try to not use jquery as much as possible

```suggestion
  assert.ok(exists("body.static-faq"), "routed to the faq page");
```

jjaffeux:
It's the first time I'm reviewing some js code you wrote so I will have few nitpicks, feel free to ignore, just saying it once for science, and maybe you might agree :p

app/assets/javascripts/discourse/routes/unknown.js.es6
```diff
@@ -1,8 +1,27 @@
...
 export default DiscourseRoute.extend({
-  model() {
-    return ajax("/404-body", { dataType: "html" });
+  model(params) {
+    const bodyReq = ajax("/404-body", { dataType: "html" });
+    const testReq = ajax("/permalink-check.json", {
+      data: { path: params.path }
+    });
+    return RSVP.allSettled([bodyReq, testReq]).then(results => {
+      if (results[1].state === "fulfilled" && results[1].value.found) {
```
jjaffeux: I would probably give meaningful names to `results[0]` and `results[1]` so it's easier to follow logic after.

test/javascripts/acceptance/unknown-test.js.es6
```diff
@@ -1,8 +1,49 @@
 import { acceptance } from "helpers/qunit-helpers";
+import pretender from "helpers/create-pretender";
 acceptance("Unknown");
 
 QUnit.test("Unknown URL", async assert => {
+  pretender.get("/permalink-check.json", () => {
```
jjaffeux: when you have only one state you can write it this way:

```
pretender.get("/permalink-check.json", () => [404, {}, null]);
```

app/assets/javascripts/discourse/routes/unknown.js.es6
jjaffeux: I'm unclear what's the purpose of found given it will raise 404 server side if not found?

app/assets/javascripts/discourse/routes/unknown.js.es6
```diff
@@ -1,8 +1,27 @@
 import DiscourseRoute from "discourse/routes/discourse";
+import RSVP from "rsvp";
 import { ajax } from "discourse/lib/ajax";
 
 export default DiscourseRoute.extend({
-  model() {
-    return ajax("/404-body", { dataType: "html" });
+  model(params) {
```
jjaffeux: very nitpick but given you use params only for path, you could write:

```
model({ path } = params) {

and then later do:  `data : { path }`

File: app/controllers/permalinks_controller.rb [hunk 6]
```diff
@@ -17,4 +17,18 @@ def show
     end
   end
 
+  def check
+    permalink = Permalink.find_by_url(params[:path])
```
jjaffeux: should we handle case where path is not set ? Especially if used through an API, it would raise a 404 which is slightly unexpected. As it's more an invalid param.

riking: find_by_* will return nil instead of throwing.

app/assets/javascripts/discourse/routes/unknown.js.es6
```diff
@@ -1,8 +1,27 @@
...
+    const bodyReq = ajax("/404-body", { dataType: "html" });
+    const testReq = ajax("/permalink-check.json", {
+      data: { path: params.path }
+    });
+    return RSVP.allSettled([bodyReq, testReq]).then(results => {
+      if (results[1].state === "fulfilled" && results[1].value.found) {
+        if (results[1].value.internal) {
+          this.transitionTo(results[1].value.target_url);
+        } else {
+          window.location = results[1].value.target_url;
```
riking: Can we reuse the logic for "open external links in new tab" here somehow?

app/assets/javascripts/discourse/routes/unknown.js.es6
riking: eslint complained when I tried to do this.. how are you supposed to spell the parameter?

test/javascripts/acceptance/unknown-test.js.es6
riking: For some reason this started failing...

riking: Is there something special about $ in tests that makes it work for `body` ?

jjaffeux: Yes I think I know what's going on.

find is probably scoped to #ember-testing-container when $ is scoped to document.

So I guess this is a special case. In this case you could do : 

```
// body is outside of #ember-testing-container and needs to be targeted
// through document instead of find
assert.ok(document.querySelector("body.static-faq"), "routed to the faq page");
```

I think that should work.

app/assets/javascripts/discourse/routes/unknown.js.es6
jjaffeux: hmmm weird thought this should work, nevermind. Maybe I will try when pushed.

app/controllers/permalinks_controller.rb
jjaffeux: yes but the line after your raise no ?

test/javascripts/acceptance/unknown-test.js.es6
riking: Yep, this worked!

app/assets/javascripts/discourse/routes/unknown.js
```diff
@@ -1,8 +1,29 @@
...
+    const testReq = ajax("/permalink-check.json", {
+      data: { path }
+    });
+    return RSVP.allSettled([bodyReq, testReq]).then(results => {
+      const [bodyResp, testResp] = results;
+      if (testResp.state === "fulfilled" && testResp.value) {
+        if (testResp.value.internal) {
+          this.transitionTo(testResp.value.target_url);
+        } else {
+          window.location = testResp.value.target_url;
```
jjaffeux: I think it's safer to use: https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/lib/url.js#L303

jjaffeux:
LGTM :+1:

app/assets/javascripts/discourse/routes/unknown.js
riking: Wait, this can just use `DiscourseURL.routeTo()`. All the correct behavior we want.

jjaffeux: Yes most likely.

riking:
Changed to use DiscourseURL.routeTo, which required special handling for the "open external links in new tab" setting. Testing was confounded by the fact that the linkify theme component sets all links to a new tab already.

app/assets/javascripts/discourse/routes/unknown.js
```diff
@@ -1,8 +1,41 @@
...
+          "external_links_in_new_tab"
+        );
+
+        if (!isInternal && openExternalInNewTab) {
+          const newWindow = window.open(href, "_blank");
+          newWindow.opener = null;
+          newWindow.focus();
+          transition.abort();
+        } else {
+          DiscourseURL.routeTo(testResp.value.target_url);
```
jjaffeux:
```suggestion
          DiscourseURL.routeTo(href);
```

?

app/controllers/permalinks_controller.rb
riking: Oh, you mean returning something like found: false? I don't think this is too bad of an API as it is right now.

app/assets/javascripts/discourse/routes/unknown.js
```diff
@@ -1,8 +1,28 @@
...
-    return ajax("/404-body", { dataType: "html" });
+  model(params, transition) {
+    const path = params.path;
+    const bodyReq = ajax("/404-body", { dataType: "html" });
+    const testReq = ajax("/permalink-check.json", {
+      data: { path }
+    });
+    return RSVP.allSettled([bodyReq, testReq]).then(results => {
+      const [bodyResp, testResp] = results;
+      if (testResp.state === "fulfilled" && testResp.value) {
```
udan11: Checking promise response looks a bit hacky. Also, it is always doing two requests even if it should not be the case, but I understand the intention of having parallel requests. 

Is there any chance you could make the `/permalink-check` endpoint return the 404-body if permalink does not exist?

```javascript
ajax("/permalink-check", { data: { path }, dataType: "html" })
  .then(url => { /* redirect */ })
  .catch(body404 => { /* show error */ });
```

eviltrout: Agreed, I don't think we should count on `state`/`value` for this.

app/controllers/permalinks_controller.rb
eviltrout: I actually think 404 is fine in this case.

app/assets/javascripts/discourse/routes/unknown.js
riking: Ported the code from exceptions#not_found_body into the check_permalink handler. This looks even cleaner on the client now!

app/controllers/permalinks_controller.rb
riking: Moot: we now return json with the 404 body directly on not found.

riking:
Fixed the JS test, changed the copy on `/admin/customize/permalinks`, and rebased.

PR #9134, opened 2020-03-06
FEATURE: Add embed_set_canonical_url setting

markvanlan:
This sets the canonical_url to the topic's `topic_embed.embed_url` when the `embed_set_canonical_url` site setting is enabled.

Defaulted the site setting to `false`, because Falco was worried about the number of Meta complaints this will raise.

### DIFF
--- a/spec/components/topic_view_spec.rb
+++ b/spec/components/topic_view_spec.rb
@@ -249,20 +249,6 @@ describe TopicView do
       end
     end
 
-    describe "#get_canonical_path with embed_set_canonical_url true" do
-      fab!(:topic_embed) { Fabricate(:topic_embed, embed_url: "https://markvanlan.com") }
-      let(:path) { "/1234" }
-
-      before do
-        SiteSetting.embed_set_canonical_url = true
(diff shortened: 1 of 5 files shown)


spec/components/topic_view_spec.rb
```diff
@@ -249,6 +249,20 @@
       end
     end
 
+    describe "#get_canonical_path with embed_set_canonical_url true" do
+      fab!(:topic_embed) { Fabricate(:topic_embed, embed_url: "https://markvanlan.com") }
+      let(:path) { "/1234" }
+
+      before do
+        SiteSetting.embed_set_canonical_url = true
+        TopicView.any_instance.stubs(:find_topic).with(1234).returns(topic_embed.topic)
```
eviltrout: Could you do this without a stub? I'd always prefer that.

It might be easier to do a request spec for this particular thing.

markvanlan: @eviltrout Agreed; moved to topics_controller_spec without stubbing. Also this now tests all the way from server to crawler specifically which rules. Great call.

PR #9135, opened 2020-03-06
FIX: prevent scheduled publishing to deleted category

tshenry:
This change will destroy any `TopicTimer` that is associated with a category upon that category's deletion. This prevents an edge case that can happen when a pending "Schedule Publishing" `TopicTimer` is set to a category that is deleted after the timer is created.


ZogStriP:
There should be a test as well 😉

SamSaffron:
sorted per: 

https://github.com/discourse/discourse/commit/46a962224621eef623e208f2810af67ad55553c9 

Thanks @tshenry

PR #9136, opened 2020-03-06
DEV: shutdown_ok parameter to /srv/status

riking:
This allows probers to distinguish between liveness and readiness conditions, which you need to avoid k8s killing your container early in the shutdown sequence.

k8s does not allow checking the content of the probe response.

https://meta.discourse.org/t/do-hits-to-srv-status-count-as-crawlers/143431/9?u=riking


ZogStriP:
Looks ok to me. @SamSaffron what do you think?

riking:
> There should be a test for this.

Added, was unsure of the practices for modifying global variables like this. PTAL

PR #9137, opened 2020-03-06
FEATURE: Add support for custom gravatar-like services

hellcp:
Adds 3 config values that allow to set a custom provider of Gravatar-like API accessible from gravatar_base_url. The gravatar_name is purely cosmetic, but helps with associating name with the service that actually provides the avatars. gravatar_login_url is a link relative to gravatar_base_url, which provides the user with the login to the Gravatar service


hellcp:
@eviltrout outside of the obvious user model url substitution test (which I added quickly), what kind of tests would you like to see? I don't think there is much of a reason to add a reliance in the tests on services which most instances won't use, so just testing if the correct url is in correct place seems like it would be enough.

PR #9138, opened 2020-03-07
DEV: enforces ember-template-lint: no-unused-block-params

jjaffeux:


jjaffeux:
@eviltrout user summary is a very edge case, as we already disable template for another rule, which causes it to no parse it and think we don't use link.

PR #9144, opened 2020-03-09
DEV: Merge imports from the same module

CvX:


jjaffeux:
Great thanks :+1:

We should merge this ASAP as this kind of commits age badly :p

PR #9147, opened 2020-03-09
DEV: enforces ember-template-lint: no-unbound

jjaffeux:
From ember-template-lint documentation (https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-unbound.md):

```
{{unbound}} is a legacy hold over from the days in which Ember's template engine was less performant. Its use today is vestigial, and it no longer offers performance benefits.

It is also a poor practice to use it for rendering only the initial value of a property that may later change.
```

### DIFF
--- a/app/assets/javascripts/admin/templates/users-list-show.hbs
+++ b/app/assets/javascripts/admin/templates/users-list-show.hbs
@@ -49,7 +49,7 @@
               {{/if}}
             </td>
             <td class='email'>
-              {{user.email~}}
+              {{~user.email~}}
             </td>
             <td class="last-emailed">
               <div class="label">{{i18n 'admin.users.last_emailed'}}</div>
--- a/app/assets/javascripts/admin/templates/users-list-show.hbs
+++ b/app/assets/javascripts/admin/templates/users-list-show.hbs
@@ -40,16 +40,16 @@
         {{#each model as |user|}}
           <tr class="user {{user.selected}} {{unless user.active 'not-activated'}}">
             <td class="username">
-              <a href={{unbound user.path}} data-user-card={{unbound user.username}}>
+              <a href={{user.path}} data-user-card={{user.username}}>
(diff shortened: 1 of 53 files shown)


app/assets/javascripts/admin/templates/users-list-show.hbs
```diff
@@ -40,16 +40,16 @@
...
               </a>
-              {{#link-to 'adminUser' user}}{{unbound user.username}}{{/link-to}}
+              {{#link-to 'adminUser' user}}{{user.username}}{{/link-to}}
               {{#if user.staged}}
                 {{d-icon "far-envelope" title="user.staged" }}
               {{/if}}
             </td>
             <td class='email'>
-              {{~unbound user.email~}}
+              {{user.email~}}
```
CvX:
```suggestion
              {{~user.email~}}
```

jjaffeux:
@eviltrout this one is a little bit risky, but it's hard to know if we were relying on some weird side effect of unbound to make anything work. I suggest we try and see.

davidtaylorhq:
Hmm yes, I think this will introduce a few regressions, because we have been using it for:

> rendering only the initial value of a property

One example I can think of: https://github.com/discourse/discourse-theme-creator/commit/d1ab92d3a20ba1f6778f97b7b6bed170658991b9

But it's still a good change - we should try to clean up these uses.

jjaffeux:
> Hmm yes, I think this will introduce a few regressions, because we have been using it for:
> 
> > rendering only the initial value of a property
> 
> One example I can think of: [discourse/discourse-theme-creator@d1ab92d](https://github.com/discourse/discourse-theme-creator/commit/d1ab92d3a20ba1f6778f97b7b6bed170658991b9)
> 
> But it's still a good change - we should try to clean up these uses.

Yep this is a very bad pattern, we need to get rid of it

PR #9149, opened 2020-03-09
FIX: Error message for 403 when featuring topic on profile

markvanlan:
Problem described in [this meta topic](https://meta.discourse.org/t/can-t-set-featured-topic-in-profile-to-restricted-categories/143380?u=markvanlan)

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1362,7 +1362,7 @@ class UsersController < ApplicationController
     user = fetch_user_from_params
     topic = Topic.find(params[:topic_id].to_i)
 
-    if topic.nil? || !guardian.can_feature_topic?(user, topic)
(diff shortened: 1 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -1362,7 +1362,10 @@ def feature_topic
     user = fetch_user_from_params
     topic = Topic.find(params[:topic_id].to_i)
 
-    raise Discourse::InvalidAccess.new unless topic && guardian.can_feature_topic?(user, topic)
+    if !topic || !guardian.can_feature_topic?(user, topic)
```
eviltrout: Very minor suggestion - the guardian could check if `topic.nil?` so you are only checking one condition here, and if in the future someone else uses the guardian method it'll return false for this case too.

PR #9160, opened 2020-03-10
Support for transpiling `.js` files

eviltrout:
This PR adds support for `.js` transpilation to migrate us away from `.js.es6`

Once merged I will follow up with a commit that renames all existing files.


spec/lib/discourse_js_processor_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    it "returns false for a regular js file" do
+      expect(DiscourseJsProcessor.should_transpile?("file.js")).to eq(false)
+    end
+
+    it "returns true for deprecated .es6 files" do
+      expect(DiscourseJsProcessor.should_transpile?("file.es6")).to eq(true)
+      expect(DiscourseJsProcessor.should_transpile?("file.js.es6")).to eq(true)
+      expect(DiscourseJsProcessor.should_transpile?("file.js.es6.erb")).to eq(true)
```
markvanlan: What about for `.js.erb`? Does this happen?

markvanlan: ah this test is for es6 extension. carry on!

PR #9163, opened 2020-03-10
FIX: word boundary regex (\b) not working in Unicode languages.

vinothkannans:
Removed the jQuery Highlight plugin and created a new es6 highlighting library.


app/assets/javascripts/discourse/lib/highlight-html.js.es6
```diff
@@ -0,0 +1,88 @@
+function highlight(node, pattern, nodeName, className) {
+  if (
+    ![1, 3].includes(node.nodeType) ||
```
eviltrout: Can we use `Node.ELEMENT_NODE` and other enums rather than hard coding integers here? It's much more clear.

app/assets/javascripts/discourse/lib/highlight-html.js.es6
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  if (node.nodeType === 1 && node.childNodes) {
+    for (let i = 0; i < node.childNodes.length; i++) {
+      i += highlight(node.childNodes[i], pattern, nodeName, className);
+    }
+    return 0;
+  }
+
+  if (node.nodeType === 3) {
```
eviltrout: Let's remove this constant too.

app/assets/javascripts/discourse/lib/highlight-html.js.es6
```diff
@@ -0,0 +1,88 @@
...
+    matchCase: false
+  };
+
+  Object.assign(settings, opts);
+  words = typeof words === "string" ? [words] : words;
+  words = words
+    .filter(Boolean)
+    .map(word => word.replace(/[-\/\\^$*+?.()|[\]{}]/g, "\\$&"));
+
+  if (words.length === 0) return node;
```
jjaffeux:
```suggestion
  if (!words.length) return node;
```

ZogStriP:
The `highlight-html` lib should be tested 😉

SamSaffron:
@vinothkannans are we just waiting on a few tests here prior to merging? Agree it makes sense to test this, especially since it is so easy to test.

vinothkannans:
Yes, I'll make it today 👍

vinothkannans:
I did some improvements and added a test [here](https://github.com/discourse/discourse/pull/9163/commits/e3e033596643fa7692a4690327c37be867b21f81) @ZogStriP. The highlight function is performing well in big paragraphs and the large number of posts @SamSaffron. Anyway, we're going keep using `\b` regex if Unicode words are not found.

PR #9164, opened 2020-03-10
UX: Add quick filters to reviewable page

riking:
It was annoying me to have to set two different dropdowns whenever I wanted to check for a recently resolved flag. Add a place for buttons for commonly used review filters, which replaces the status chooser above the mobile fold.

"Reset to default" gets special handling in the form of a disable handler.

Additional suggestions for other common filters welcome or can be added in followups.

### DIFF
--- a/app/assets/javascripts/discourse/templates/review-index.hbs
+++ b/app/assets/javascripts/discourse/templates/review-index.hbs
@@ -17,16 +17,36 @@
   </div>
 
   <div class='reviewable-filters'>
-    <div class='reviewable-filter'>
-      <label class='filter-label'>{{i18n "review.filters.status"}}</label>
-      {{combo-box
-        value=filterStatus
-        content=statuses
-        onChange=(action (mut filterStatus))
-      }}
+    <div class='reviewable-filter quick-filters'>
+      {{i18n "review.filters.quick"}}
+      <div>
+        {{d-button
+          label="review.filters.reset"
+          icon="times"
+          class="btn-default"
+          disabled=disableFilterReset
+          action=(action "resetFilters")}}
+
+        {{plugin-outlet name="review-quick-filters" args=(hash model=model)}}
+
+        {{d-button
+          label="review.filters.resolved"
+          class="btn-default"
+          action=(action "setFilters")
+          actionParam=(hash sort_order="created_at" status="reviewed")}}
+      </div>
     </div>
 
     {{#if filtersExpanded}}
+
+      <div class='reviewable-filter'>
+        <label class='filter-label'>{{i18n "review.filters.status"}}</label>
+        {{combo-box
+          value=filterStatus
+          content=statuses
+          onChange=(action (mut filterStatus))
+        }}
+      </div>
 
       {{plugin-outlet name="above-review-filters" args=(hash model=model additionalFilters=additionalFilters)}}
 
--- a/app/assets/javascripts/discourse/controllers/review-index.js
+++ b/app/assets/javascripts/discourse/controllers/review-index.js
@@ -1,5 +1,6 @@
 import discourseComputed from "discourse-common/utils/decorators";
 import Controller from "@ember/controller";
+import { not } from "@ember/object/computed";
 
 export default Controller.extend({
   queryParams: [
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/discourse/templates/review-index.hbs
```diff
@@ -17,17 +17,37 @@
...
+          disabled=disableFilterReset
+          action=(action "resetFilters")}}
+
+        {{plugin-outlet name="review-quick-filters" args=(hash model=model)}}
+
+        {{d-button
+          label="review.filters.resolved"
+          class="btn-default"
+          action=(action "setFilters")
+          actionParam='{"sort_order": "created_at", "status": "reviewed"}'}}
```
riking: How do I pass an object in from a Handlebars template?

app/assets/javascripts/discourse/templates/review-index.hbs
```diff
@@ -17,17 +17,37 @@
...
+      <div class='reviewable-filter'>
+        <label class='filter-label'>{{i18n "review.filters.status"}}</label>
+        {{combo-box
+          value=filterStatus
+          content=statuses
+          onChange=(action (mut filterStatus))
+        }}
+      </div>
+
       {{plugin-outlet name="above-review-filters" args=(hash model=model additionalFilters=additionalFilters)}}
```
riking: Should status go above or below the plugin outlet?

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -127,6 +168,27 @@ export default Controller.extend({
...
+        to_date: null,
+        sort_order: "priority",
+        additional_filters: null
+      });
+
+      this.send("refreshRoute");
+    },
+
+    setFilters(new_filters) {
+      this.setProperties(JSON.parse(new_filters));
```
riking: This is fairly jank - how do I filter this to approved properties and ideally accept an object from Handlebars?

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -85,6 +85,47 @@ export default Controller.extend({
...
+      username !== "" ||
+      from_date !== null ||
+      to_date !== null ||
+      sort_order !== "priority" ||
+      (additional_filters !== null &&
+        JSON.stringify(additional_filters) !== "{}")
+    );
+  },
+
+  @discourseComputed("anyFilters")
```
jjaffeux: Can probably be replaced by: https://api.emberjs.com/ember/3.12/classes/@ember%2Fobject%2Fcomputed/methods/not?anchor=not

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -127,6 +168,27 @@ export default Controller.extend({
...
+        from_date: null,
+        to_date: null,
+        sort_order: "priority",
+        additional_filters: null
+      });
+
+      this.send("refreshRoute");
+    },
+
+    setFilters(new_filters) {
```
jjaffeux:
```suggestion
    setFilters(newFilters) {
```

app/assets/javascripts/discourse/controllers/review-index.js.es6
jjaffeux:
```suggestion
      this.setProperties(JSON.parse(newFilters));
```

app/assets/javascripts/discourse/templates/review-index.hbs
jjaffeux:
```
actionParam=(hash foo=1 bar=2)
```

app/assets/javascripts/discourse/controllers/review-index.js.es6
```diff
@@ -85,6 +85,47 @@ export default Controller.extend({
...
+      type != null ||
+      priority !== "low" ||
+      status !== "pending" ||
+      category_id !== null ||
+      username !== "" ||
+      from_date !== null ||
+      to_date !== null ||
+      sort_order !== "priority" ||
+      (additional_filters !== null &&
+        JSON.stringify(additional_filters) !== "{}")
```
jjaffeux: I think this would be a better pattern:

```
Object.keys(additional_filters).length
```

oblakeerickson:
Just checked this PR out to keep it from getting too stale. I think for a new feature like this it would be good to post it on meta first to get feedback, but perhaps before doing so @eviltrout could give his thoughts because he has a lot more review queue experience and ideas for its future.

This is how the feature looks:

image

Overall I think it is a good idea, because I bet it can get tedious to always select the same drop downs over and over again. My concern though is that these buttons are hard coded and every user is going to want their own version of their "Reviewed" button or any future buttons that we would add. For example they might want it to select the "Minimum Priority" of "High" instead of "(any)".

I feel like for this feature to be complete it would have to be automatic where we either automatically populate the quick filters based on usage or we add a save button that when clicked would add the current selected filters to a new quick access button. Maybe as a minimum first feature we could add a SiteSetting instead that you could populate with the selected filters you want to always have for the review queue.

---

As a side note: In the mean time, you could bookmark the "reviewed" filter, use the hamburger-links-component, or even use a go-link ;) because it looks like the selected filters get populated as query params.

eviltrout:
I think @oblakeerickson's reply is excellent. This really should have some more discussion first.

I made the filters query parameters so you should be able to bookmark a filter in your browser to easily come back to it at any time, so these buttons seem to re-creating that? My big fear is what blake says - that no two people will want exactly the same filters. 

@riking maybe let's move this to meta for discussion instead?

PR #9169, opened 2020-03-11
DEV: defines a constant for INPUT_DELAY

jjaffeux:
This should be used for standard debouncing/throttling in the app.


app/assets/javascripts/discourse/components/emoji-picker.js.es6
```diff
@@ -9,7 +9,10 @@ import {
   emojiSearch
 } from "pretty-text/emoji";
 import { safariHacksDisabled } from "discourse/lib/utilities";
-import ENV from "discourse-common/config/environment";
+import {
+  default as ENV,
```
CvX:
```suggestion
import ENV, {
```

PR #9172, opened 2020-03-11
UX: align modals to top and full width on mobile

jjaffeux:
Discourse Meta - The Official Support Forum for Discourse

### DIFF
--- a/app/assets/stylesheets/mobile/modal.scss
+++ b/app/assets/stylesheets/mobile/modal.scss
@@ -7,6 +7,7 @@
 .modal-middle-container {
   max-width: 100%;
   margin-bottom: 30px; // For iOS Safari issues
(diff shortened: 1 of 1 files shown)


jjaffeux:
@awesomerobot how do you feel about this ?

CvX:
Hm, not sure about this one. It reduces reachability of the UI for no upside. (other than a bit of visual alignment?)

jjaffeux:
At the very least I think the width makes sense.

It's very annoying to have all modals with various widths.

pmusaraj:
I think modals look better center-aligned, both vertically and horizontally. It would be good to have uniform widths, but I would prefer at least some left/right margin.

jjaffeux:
Ok Ok :p I will rethink this

PR #9173, opened 2020-03-11
DEV: enforces ember-template-lint: no-partial

jjaffeux:
selected-posts parial is kept and calling the new component to prevent errors with users who would have rewritten topic.hbs

dashboard-problems and version-checks seem less risky and have only been converted to components


app/assets/javascripts/admin/templates/components/dashboard-problems.hbs
```diff
@@ -18,7 +18,7 @@
         </div>
 
         <p class="actions">
-          {{d-button action=(action "refreshProblems") class="btn-default" icon="sync" label="admin.dashboard.refresh_problems"}}
+          {{d-button action=refreshProblems class="btn-default" icon="sync" label="admin.dashboard.refresh_problems"}}
```
CvX: Is this change related to partials?

jjaffeux: I think yes ? AFAIK variables in context are passed to partials automatically, but as it's converted to a component it's not the case anymore, so I'm create the action when I call the component {{dashboard-problems}} which is in the context of having the "refreshProblems" action handler.

Or am I missing your point ?

CvX: Ah, yes, it's an arg now. Carry on. 🙂

Btw. we should start using named `@args` (https://github.com/emberjs/rfcs/blob/e3d18e6206546f1c846ce593f2314ab3f536d253/text/0276-named-args.md) from what I can tell they've been around since Ember 3.0.0-beta1 😃 (that and `this.` for properties)

jjaffeux: indeed, do you want to do a post on dev about it please?

PR #9175, opened 2020-03-11
DEV: applies default modal class to a modal when shown

jjaffeux:


jjaffeux:
Today most modals only have a class on body. This is very annoying, because you can't interact with a specific modal.

With this change, if you don't specify a `modalClass` when calling `showModal("foo")`, the `#discourse-modal` will have a `.foo-modal` added by default.

eviltrout:
Although maybe we should dasherize/lower case the name if it's not already?

PR #9177, opened 2020-03-11
DEV: Support adding keybindings via plugins

markvanlan:


app/assets/javascripts/discourse/lib/plugin-api.js.es6
```diff
@@ -50,6 +51,7 @@ import { addCategorySortCriteria } from "discourse/components/edit-category-sett
 import { queryRegistry } from "discourse/widgets/widget";
 import Composer from "discourse/models/composer";
 import { on } from "@ember/object/evented";
+import KeyboardShortcuts, { bindings } from "discourse/lib/keyboard-shortcuts";
 
 // If you add any methods to the API ensure you bump up this number
 const PLUGIN_API_VERSION = "0.8.38";
```
eviltrout: Let's bump this due to new method.

PR #9179, opened 2020-03-11
FIX: When must_approve_users is enabled, we don't want to send suspect users to the review queue

romanrizzi:
Context: https://meta.discourse.org/t/approve-suspect-user/143836

- Only non-approved users should be sent. 
- Provide a migration to auto-approve every problematic review item

### DIFF
--- a/db/migrate/20200311135425_clear_approved_users_from_the_review_queue.rb
+++ b/db/migrate/20200311135425_clear_approved_users_from_the_review_queue.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+class ClearApprovedUsersFromTheReviewQueue < ActiveRecord::Migration[6.0]
+  def up
+    reviewables = DB.query_single <<~SQL
+      UPDATE reviewables r
+      SET status = #{Reviewable.statuses[:approved]}
+      FROM users u
+      WHERE u.approved = true AND r.type = 'ReviewableUser' AND r.status = #{Reviewable.statuses[:pending]}
+      RETURNING r.id
+    SQL
+
+    system_user_id = Discourse::SYSTEM_USER_ID
+    scores = reviewables.map do |id|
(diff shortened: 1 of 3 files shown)


db/migrate/20200311135425_clear_approved_users_from_the_review_queue.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+class ClearApprovedUsersFromTheReviewQueue < ActiveRecord::Migration[6.0]
+  def up
+    stuck_user_ids = User.joins("INNER JOIN reviewables r ON users.id = r.target_id")
```
eviltrout: Can we replace this with SQL queries? We've found that using AR objects often introduces failures in the future when the structure of the objects change (or they are removed!)

romanrizzi: Done!

db/migrate/20200311135425_clear_approved_users_from_the_review_queue.rb
```diff
@@ -0,0 +1,22 @@
...
+      WHERE u.approved = true AND r.type = 'ReviewableUser' AND r.status = #{Reviewable.statuses[:pending]}
+    SQL
+
+    DB.exec(
+      "UPDATE users SET approved = false WHERE id IN (:user_ids)",
+      user_ids: stuck_user_ids
+    )
+
+    system_user = Discourse.system_user
+    ReviewableUser.where(target_id: stuck_user_ids).find_each { |ru| ru.perform(system_user, :approve_user) }
```
eviltrout: Even this is unfortunately not going to work in a migration for the long term :(

Is there any way you could use SQL to approve them? If it's only for stuck users it needn't be perfect.

PR #9180, opened 2020-03-11
FIX: Allow CSP to work correctly for non-default hostnames/schemes

davidtaylorhq:


lib/content_security_policy/middleware.rb
```diff
@@ -12,12 +12,12 @@ def call(env)
       _, headers, _ = response = @app.call(env)
 
       return response unless html_response?(headers)
-      ContentSecurityPolicy.base_url = request.host_with_port if !Rails.env.production?
+      base_url = request.host_with_port
```
ZogStriP: Is there any ways `request.host_with_port` could be "manipulated" by the users or is always under "our" control?

spec/components/middleware/enforce_hostname_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Middleware::EnforceHostname do
+
+  before do
+    RailsMultisite::ConnectionManagement.stubs(:current_db_hostnames).returns(['primary.example.com', 'secondary.example.com'])
```
riking: Split this into a pair of describe blocks so you can test the case for 1 and empty.

PR #9184, opened 2020-03-12
FIX: Remote themes Github link should go to custom branch

danielhollas:
I am trying to improve the remote theme "Source" link in Admin->Customize->Themes so that it points to the branch that was selected during import (if it was non-default). Without this, admins have no good way of knowing which branch was used in the initial import.

image

Nevertheless, my code does not work for me. :sob: I am unable to access the `model.remote_theme.branch`, neither from the template nor in the controller.


vinothkannans:
Thanks for the PR @danielhollas. It's merged in the commit https://github.com/discourse/discourse/commit/48d690ae01e6a65c2c5ac7995b7e0b8a741599c2.

PR #9191, opened 2020-03-12
FIX: prevents i18n helper to return a SafeString

jjaffeux:
For convenience the i18n helper has been made returning a SafeString, but when used with other helpers (eg: replace-emoji), a String is expected and will cause unexpected behaviors.

This is the root cause of the initial bug fixed in https://github.com/discourse/discourse/commit/d2bb127e2c8c4a4df30024883ee36472fc86b688

This commit is kept as it's a better security in case of unexpected behavior.

### DIFF
--- a/app/assets/javascripts/admin/templates/components/admin-report-storage-stats.hbs
+++ b/app/assets/javascripts/admin/templates/components/admin-report-storage-stats.hbs
@@ -15,7 +15,7 @@
 
       {{#if backupStats.last_backup_taken_at}}
         <br>
-        {{i18n "admin.dashboard.lastest_backup" date=(format-date backupStats.last_backup_taken_at leaveAgo="true")}}
+        {{html-safe (i18n "admin.dashboard.lastest_backup" date=(format-date backupStats.last_backup_taken_at leaveAgo="true"))}}
       {{/if}}
     </p>
   </div>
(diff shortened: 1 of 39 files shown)


ZogStriP:
Not sure I like that change as it will introduce another thing to think about when using translations... Is there another way we could fix the issue?

jjaffeux:
> Not sure I like that change as it will introduce another thing to think about when using translations... Is there another way we could fix the issue?

There's no other thing to think. We were already doing `{{{i18n "foo"}}}` and now it's `{{html-safe (i18n "foo")}}` it's the exact translation of `{{{`. 

The alternative is to make every helper work with text or SafeString which basically means to do things like: `text.string || text` I'm not sure we want to change all the helpers code.

jjaffeux:
Another alternative could be to have a {{safe-i18n}} helper

ZogStriP:
> There's no other thing to think. We were already doing `{{{i18n "foo"}}}` and now it's `{{html-safe (i18n "foo")}}` it's the exact translation of `{{{`.

Oh, I think I misunderstood. I thought we would have to change all the `{{i18n ...}}`. Now if the `{{{` syntax is being deprecated, then `{{safe-html (i18n ...)}}` is fine 👍

PR #9194, opened 2020-03-12
FIX: Use delete_all_posts_max to improve delete user protection consistency 

romanrizzi:
cc @techAPJ

### DIFF
--- a/app/assets/javascripts/discourse/models/user.js
+++ b/app/assets/javascripts/discourse/models/user.js
@@ -45,7 +45,8 @@ const User = RestModel.extend({
 
   @discourseComputed("can_be_deleted", "post_count")
   canBeDeleted(canBeDeleted, postCount) {
-    return canBeDeleted && postCount <= 5;
+    const maxPostCount = Discourse.SiteSettings.delete_all_posts_max;
+    return canBeDeleted && postCount <= maxPostCount;
   },
 
   @discourseComputed()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/admin/controllers/admin-user-index.js.es6
```diff
@@ -199,7 +199,8 @@ export default Controller.extend(CanCheckEmails, {
 
     destroy() {
       const postCount = this.get("model.post_count");
-      if (postCount <= 5) {
+      const maxPostCount = Discourse.SiteSettings.delete_all_posts_max;
```
eviltrout: Can this be `this.siteSettings.delete_all_posts_max` instead?

app/assets/javascripts/discourse/models/user.js
```diff
@@ -45,7 +45,8 @@ const User = RestModel.extend({
 
   @discourseComputed("can_be_deleted", "post_count")
   canBeDeleted(canBeDeleted, postCount) {
-    return canBeDeleted && postCount <= 5;
+    const maxPostCount = Discourse.SiteSettings.delete_all_posts_max;
```
eviltrout: ditto here

romanrizzi: `this.siteSettings` is undefined here 🤔

eviltrout: Ah yes it's a model. Can't avoid this here for now unless you set it when the model is built. Fine to ignore this one.

PR #9197, opened 2020-03-13
UX: display avatar flair in categories route topic list items

vinothkannans:


app/assets/javascripts/discourse/models/topic.js
```diff
@@ -63,9 +63,19 @@ const Topic = RestModel.extend({
...
         p => p.extras && p.extras.indexOf("latest") >= 0
       )[0];
-      user = latest && latest.user;
+      user = latest;
     }
-    return user || this.creator;
+    return user || posters.firstObject;
+  },
+
+  @discourseComputed("lastPoster")
```
eviltrout: We could use `alias` for these computed properties. Would be shorter.

app/serializers/category_and_topic_lists_serializer.rb
```diff
@@ -15,4 +16,14 @@ def users
     users
   end
 
+  def primary_groups
+    groups = object.topic_list.topics.map do |t|
+      t.posters.map { |poster| poster.try(:primary_group) }
```
eviltrout: Could this be causing a n+1 query? It seems to me like it could be and we should confirm it is not.

vinothkannans: It looks like we already preloading the `posters` of all topics with the `primary_group` summary in the [`TopicList` class](https://github.com/discourse/discourse/blob/60184a290c45926583393cf53895e063b4798ff5/app/models/topic_list.rb#L126). So it won't create N+1 query.

PR #9198, opened 2020-03-13
FEATURE: Add enable_bookmark_at_desktop_reminders site setting

martin-brennan:
Add enable_bookmark_at_desktop_reminders site setting default to false a new hidden site setting to hide the "At Desktop" reminder option so we can restrict this further until it is polished.


SamSaffron:
looks good to me, merge away

PR #9206, opened 2020-03-13
FIX: Don't display webhooks for inactive plugins

oblakeerickson:
This commit ensures that we don't show webhooks for plugins that are not
installed or that are disabled.

Bug report:

https://meta.discourse.org/t/webhookeventtype-and-the-solved-and-assign-plugins/144180

### DIFF
--- a/app/models/web_hook_event_type.rb
+++ b/app/models/web_hook_event_type.rb
@@ -20,7 +20,7 @@ class WebHookEventType < ActiveRecord::Base
 
   validates :name, presence: true, uniqueness: true
 
-  def self.all_excluding_disabled_plugin_web_hooks
+  def self.active
     ids_to_exclude = []
     ids_to_exclude << SOLVED unless defined?(SiteSetting.solved_enabled) && SiteSetting.solved_enabled
     ids_to_exclude << ASSIGN unless defined?(SiteSetting.assign_enabled) && SiteSetting.assign_enabled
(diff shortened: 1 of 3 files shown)


app/models/web_hook_event_type.rb
```diff
@@ -19,6 +19,15 @@ class WebHookEventType < ActiveRecord::Base
   default_scope { order('id ASC') }
 
   validates :name, presence: true, uniqueness: true
+
+  def self.all_excluding_disabled_plugin_web_hooks
```
ZogStriP: I love descriptive names but I think this one went over the top 😉 

What about `WebHookEventType.active` instead? (or something similar?)

app/models/web_hook_event_type.rb
```diff
@@ -19,6 +19,15 @@ class WebHookEventType < ActiveRecord::Base
...
+    ids_to_exclude = []
+    ids_to_exclude << SOLVED unless defined?(SiteSetting.solved_enabled) && SiteSetting.solved_enabled
```
ZogStriP: Hmmm 🤔 Shouldn't this be the responsability of the plugin to "add" their webhook types?

oblakeerickson: Yes, originally that was the plan

https://github.com/discourse/discourse-assign/pull/61#discussion_r385752247

But if we reserve the webhook id here we prevent future id conflicts.

PR #9213, opened 2020-03-16
FIX: Bookmark reminders and improvements changes

martin-brennan:
* Cosmetic fixes for the bookmark modal
* Do not show "later today" when the later time will be > 5pm
* When a custom reminder time is selected, store it in localStorage. The next time the modal is opened, if the last datetime is > now, then a new tile with "Last" will be shown that lets the user reselect that same time.
* Also add an explicit "No Reminder" option that is selected by default

image

### DIFF
--- a/app/assets/javascripts/discourse/controllers/bookmark.js
+++ b/app/assets/javascripts/discourse/controllers/bookmark.js
@@ -14,7 +14,9 @@ const REMINDER_TYPES = {
   TOMORROW: "tomorrow",
   NEXT_WEEK: "next_week",
   NEXT_MONTH: "next_month",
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/controllers/bookmark.js
```diff
@@ -71,9 +99,29 @@ export default Controller.extend(ModalFunctionality, {
     return REMINDER_TYPES;
   },
 
+  @discourseComputed()
```
jjaffeux: you can make this a one liner with the and computed property

http://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Fcomputed/and

PR #9215, opened 2020-03-16
FEATURE: Include optimized thumbnails for topics

davidtaylorhq:
This PR introduces new APIs for obtaining optimized thumbnails for topics. There are a few building blocks required for this:

- Introduces new `image_upload_id` columns on the `posts` and `topics` table. This replaces the old `image_url` column, which means that thumbnails are now restricted to uploads. Hotlinked thumbnails are no longer possible. In normal use (with pull_hotlinked_images enabled), this has no noticeable impact

- A migration attempts to match existing urls to upload records. If a match cannot be found then the posts will be queued for rebake

- Optimized thumbnails are generated during post_process_cooked. If thumbnails are missing when serializing a topic list, then a sidekiq job is queued

- Topic lists and topics now include a `thumbnails` key, which includes all the available images:
   ```
   "thumbnails": [
   {
     "max_width": null,
     "max_height": null,
     "url": "//example.com/original-image.png",
     "width": 1380,
     "height": 1840
   },
   {
     "max_width": 1024,
     "max_height": 1024,
     "url": "//example.com/optimized-image.png",
     "width": 768,
     "height": 1024
   }
   ]
  ```

- Themes can request additional thumbnail sizes by using a modifier in their `about.json` file:
   ```
    "modifiers": {
      "topic_thumbnail_sizes": [
        [200, 200],
        [800, 800]
      ],
      ...
  ```
  Remember that these are generated asynchronously, so your theme should include logic to fallback to other available thumbnails if your requested size has not yet been generated

- Two new raw plugin outlets are introduced, to improve the customisability of the topic list. `topic-list-before-columns` and `topic-list-before-link`


davidtaylorhq:
@SamSaffron would you mind reviewing this if you have some time?

app/models/topic.rb
```diff
@@ -135,6 +200,9 @@ def recover!(recovered_by = nil)
   has_one :first_post, -> { where post_number: 1 }, class_name: 'Post'
   has_one :topic_search_data
   has_one :topic_embed, dependent: :destroy
+  belongs_to :image_upload, class_name: "Upload"
+
+  belongs_to :image_upload, class_name: 'Upload'
```
SamSaffron, line 205: typo.. line is here twice.

db/migrate/20200415095034_add_image_upload_id_to_posts_and_topics.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddImageUploadIdToPostsAndTopics < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :posts, :image_upload, foreign_key: { to_table: :uploads, on_delete: :nullify }
```
SamSaffron, line 5: A slight concern here I have is that we are adding DB consistency foreign keys in some spots when a whole bunch of columns on the same table do not have them .... like user_id post_id and the list goes on. 

Feels a bit inconsistent as a rule to apply. 

I think we should just stick to adding the column and indexes we want by hand.

db/migrate/20200415095035_migrate_image_url_to_image_upload_id.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+class MigrateImageUrlToImageUploadId < ActiveRecord::Migration[6.0]
```
SamSaffron, line 3: feels like a post migration is what we want here, no need to hold off the deploy for this, just run it after all app servers have the new code.

app/models/theme_modifier_set.rb
```diff
@@ -50,6 +50,26 @@ def self.resolve_modifier_for_themes(theme_ids, modifier_name)
     end
   end
 
+  def topic_thumbnail_sizes
+    array = super
```
SamSaffron, line 54: use of super is a bit surprising here... `read_attribute(:topic_thumbnail_sizes)` is a bit less surprising

app/models/theme_modifier_set.rb
```diff
@@ -50,6 +50,26 @@ def self.resolve_modifier_for_themes(theme_ids, modifier_name)
...
+      [parts[0].to_i, parts[1].to_i]
+    end.filter(&:present?)
+  end
+
+  def topic_thumbnail_sizes=(val)
+    return super(val) if val.nil?
+    return super(val) if !val.is_a?(Array)
+    return super(val) if !val.all? { |v| v.is_a?(Array) && v.length == 2 }
+
+    super(val.map { |dim| "#{dim[0]}x#{dim[1]}" })
```
SamSaffron, line 70: write_attribute

SamSaffron:
should we leave image_url around and print a deprecation when it is accessed? how many plugins are we going to break?

davidtaylorhq:
Thanks for reviewing @SamSaffron 

> should we leave image_url around and print a deprecation when it is accessed

IMO we should keep it, and we don't need a deprecation. There is zero cost in keeping it, and it is a much simpler API than the thumbnails array. It's still used in core for things like opengraph/sharing images, and I have seen people use it for things like IFTTT/Zapier integrations. 

I set it to prefer a 1024x1024 thumbnail, and fall back to the original image.

db/migrate/20200415095035_migrate_image_url_to_image_upload_id.rb
davidtaylorhq: If we run this as a post-migration, then sites will stop serving opengraph images on all topics for a few minutes. I guess that's ok... but if we keep it as a pre-deploy migration then the change should be seamless. I think it should only take a few seconds.

PR #9224, opened 2020-03-17
FIX: Perform crop using user-specified image sizes

nbianca:
It used to resize the images to max width and height first and then
perform the crop operation. This is wrong because it ignored the user
specified image sizes from the Markdown.

### DIFF
--- a/lib/cooked_post_processor.rb
+++ b/lib/cooked_post_processor.rb
@@ -308,29 +308,34 @@ class CookedPostProcessor
   end
 
   def convert_to_link!(img)
+    w, h = img["width"].to_i, img["height"].to_i
+    user_width, user_height = (w > 0 && h > 0 && [w, h]) ||
+                              get_size_from_attributes(img) ||
+                              get_size_from_image_sizes(img["src"], @opts[:image_sizes])
+
+    limit_size!(img)
+
     src = img["src"]
     return if src.blank? || is_a_hyperlink?(img) || is_svg?(img)
 
-    width, height = img["width"].to_i, img["height"].to_i
-    # TODO: store original dimentions in db
     original_width, original_height = (get_size(src) || [0, 0]).map(&:to_i)
-
-    # can't reach the image...
     if original_width == 0 || original_height == 0
       Rails.logger.info "Can't reach '#{src}' to get its dimension."
       return
     end
 
-    return if original_width <= width && original_height <= height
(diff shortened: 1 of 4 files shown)


SamSaffron:
oops I think I asked @vinothkannans  to look at something similar as well.

Since this is a very visual change @nbianca  can you post some screenshots of before and after with markdown?

nbianca:
Before
localhost_3000_t_test-image-crop_129_2

After
localhost_3000_t_test-image-crop_129_2 (1)

lib/cooked_post_processor.rb
```diff
@@ -700,10 +705,7 @@ def html
 
   def post_process_images
     extract_images.each do |img|
-      unless add_image_placeholder!(img)
-        limit_size!(img)
-        convert_to_link!(img)
-      end
+      convert_to_link!(img) unless add_image_placeholder!(img)
```
davidtaylorhq: Does this mean we now onebox every image in a post, regardless of size?

nbianca: There is another [check (line 325)](https://github.com/discourse/discourse/pull/9224/files#diff-9cbdf282d05bfa042991776a62127eb9L325) which returns early if the image does not exceed maximum dimensions.

davidtaylorhq:
The code is looking ok, but I'm not sure exactly what the intention is here? 

We had a feature which crops very tall images. Does this PR remove that feature?

nbianca:
The images are still cropped when necessary, but it gives user some control over image sizes.

As an example, I used the image from [this topic](https://meta.discourse.org/t/min-ratio-to-crop-site-setting-should-respect-w-h-ratio-as-defined-in-markdown/138374) and the following raw post:

```markdown
A very tall image, unmodified (original size 1097x3351):

![image](upload://dBlV2poMFtso5zGLgXpcBraTVxg.png)

Very tall image, with a given size of 489x1500:

![image|489x1500](upload://dBlV2poMFtso5zGLgXpcBraTVxg.png) 

Very tall image, with a given size of 326x1000:

![image|326x1000](upload://dBlV2poMFtso5zGLgXpcBraTVxg.png) 

Very tall image, with a given size of 163x500:

![image|163x500](upload://dBlV2poMFtso5zGLgXpcBraTVxg.png)
```

There are two results (before and after this pull request):

before fixes

after fixes

PR #9225, opened 2020-03-17
FIX: Condense line codes in emails

nbianca:
Before
Screenshot 2020-03-17 at 16 41 30

After
Screenshot 2020-03-17 at 16 41 32


eviltrout:
Is there a context for this change? One thing we have to watch out for is email clients are really varied with respect to formatting. What works in mailcatcher often won't in gmail and such :(

kacf:
This change breaks email rendering in Thunderbird. See example:

[](https://postimg.cc/1gxGnC3P)

eviltrout:
@kacf we've followed it up here: 

https://github.com/discourse/discourse/commit/ba1a08510eb9574803955a9a2e6e7a8f7b2d1a43

Once that builds please try again.

kacf:
I can confirm, this is fixed!

PR #9230, opened 2020-03-18
FIX: Improve user timezone saving

martin-brennan:
Based on issues identified in https://meta.discourse.org/t/improved-bookmarks-with-reminders/144542/20

* Implement the resolvedTimezone() function on the user model where we return the user's timezone if it has been set, or we guess it using moment and save it to the user using an update call if it has not yet been set. This covers the cases of users who do not log out/in often who will not get their timezone set via login. This also makes sure the guess + save is done in a non-obtrusive way not on every page -- only when it is needed.

* Before if a user's timezone was blank when they visited their profile page we were autofilling the dropdown with the guessed timezone from moment. However this was confusing as it would appear you have that timezone saved in the DB when you really didn't. Now we do not autofill the dropdown and added a button to automatically guess the current timezone to make everything more explicit.

image


app/assets/javascripts/discourse/models/user.js
```diff
@@ -840,6 +840,21 @@ const User = RestModel.extend({
       !secondFactorEnabled &&
       (enforce === "all" || (enforce === "staff" && staff))
     );
+  },
+
+  resolvedTimezone() {
```
jjaffeux: I'm unsure this should be in the model.

At the very least I think async event from models should return the promise.

app/assets/javascripts/discourse/models/user.js
```diff
@@ -840,6 +840,21 @@ const User = RestModel.extend({
...
+    if (this.timezone) {
```
eviltrout: Minor but for private variables like this I'd suggest using `this._timezone` - it makes it obvious to someone building on top of this that they should be careful about using that variable rather than the `resolvedTimezone()` method.

test/javascripts/controllers/bookmark-test.js.es6
```diff
@@ -181,14 +181,23 @@ QUnit.test(
...
-    Discourse.currentUser.timezone = null;
-    BookmarkController.onShow();
-    assert.equal(BookmarkController.userHasTimezoneSet, false);
-    Discourse.currentUser.timezone = "Australia/Brisbane";
-    BookmarkController.onShow();
-    assert.equal(BookmarkController.userHasTimezoneSet, true);
-  }
-);
+QUnit.test("user timezone updates when the modal is shown", function(assert) {
+  Discourse.currentUser.timezone = null;
```
eviltrout: It's better to use the `updateCurrentUser` helper

app/assets/javascripts/discourse/models/user.js
martin-brennan: @jjaffeux I did it this way because IMO the result of the AJAX call does not matter, we just want to get the timezone or guess it and transparently save in the background. Where would you put this if not the model? Because the user model has the timezone on it and this is updating something on the user model.

I will defer to your experience here and return the promise if you think that will work better, just wanted to explain my reasoning.

app/assets/javascripts/discourse/models/user.js
```diff
@@ -840,10 +840,40 @@ const User = RestModel.extend({
...
       (enforce === "all" || (enforce === "staff" && staff))
     );
+  },
+
+  resolvedTimezone() {
+    if (this._timezone) {
+      return this._timezone;
+    }
+
+    this.set("_timezone", moment.tz.guess());
```
eviltrout: If you're not going to use `_timezone` in a computed property or observer (and you shouldn't!) you can omit `set` here and do `this._timezone = `

martin-brennan: Thanks!

PR #9236, opened 2020-03-19
FIX: N1 issues for bookmark list

martin-brennan:
* Preload custom fields for BookmarkQuery and add preload callback. Copy TopicQuery preload methodology to allow plugins to preload data for the BookmarkQuery. This fixes assigned plugin custom fields N1
* Include topic tags in initial query to avoid tags N1

Related: https://github.com/discourse/discourse-assign/pull/63


SamSaffron:
Looks good, merge at will

PR #9237, opened 2020-03-19
FIX: Bug in setting notification level when the post was moved

lis2:
When the post was moved from one topic into a new one, and the user with `Automatically track topics I enter` preferences, visits that topic for the first time notification_level was not modified.

### DIFF
--- a/app/models/topic_user.rb
+++ b/app/models/topic_user.rb
@@ -131,9 +131,6 @@ class TopicUser < ActiveRecord::Base
 
         if rows == 0
           create_missing_record(user_id, topic_id, attrs)
-        elsif topic_user = TopicUser.find_by(topic_id: topic_id, user_id: user_id, notifications_reason_id: nil, notifications_changed_at: nil)
-          set_notification_level_attributes(user_id, topic_id, attrs)
-          topic_user.update(attrs)
         end
       end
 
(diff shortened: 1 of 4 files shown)


app/models/topic_user.rb
```diff
@@ -131,6 +131,9 @@ def change(user_id, topic_id, attrs)
 
         if rows == 0
           create_missing_record(user_id, topic_id, attrs)
+        elsif topic_user = TopicUser.find_by(topic_id: topic_id, user_id: user_id, notifications_reason_id: nil, notifications_changed_at: nil)
+          find_notifiation_level(user_id, topic_id, attrs)
```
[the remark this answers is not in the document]
eviltrout: It looks to me like it doesn’t modify any data - however there is a typo in the name. It’s missing a c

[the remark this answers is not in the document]
jjaffeux: what about this line 

https://github.com/discourse/discourse/pull/9237/files#diff-a453d5e1c29398203488ff879ca69664R226

app/models/topic_user.rb
```diff
@@ -219,26 +234,19 @@ def create_missing_record(user_id, topic_id, attrs)
...
     end
 
     def track_visit!(topic_id, user_id)
       now = DateTime.now
       rows = TopicUser.where(topic_id: topic_id, user_id: user_id).update_all(last_visited_at: now)
 
-      if rows == 0
-        change(user_id, topic_id, last_visited_at: now, first_visited_at: now)
-      end
+      change(user_id, topic_id, last_visited_at: now, first_visited_at: now)
```
SamSaffron: first_visited_at ... now? how does that make sense, are we 10000% track_visit! is only every called once?

app/models/topic_user.rb
```diff
@@ -158,6 +161,17 @@ def notification_level_change(user_id, topic_id, notification_level, reason_id)
...
+
+      TopicUser.create!(attrs.merge!(
+        user_id: user_id,
+        topic_id: topic_id,
+        first_visited_at: now,
+        last_visited_at: now
+      ))
+    end
+
+    def set_notification_level_attributes(user_id, topic_id, attrs)
```
SamSaffron: is this protected? cause the way it works just changed here

app/models/topic_user.rb
```diff
@@ -131,6 +131,9 @@ def change(user_id, topic_id, attrs)
 
         if rows == 0
           create_missing_record(user_id, topic_id, attrs)
+        elsif topic_user = TopicUser.find_by(topic_id: topic_id, user_id: user_id, notifications_reason_id: nil, notifications_changed_at: nil)
```
SamSaffron: I am not sure here, this is changing 

update
- if missing create

to 

update
- if existing ... lookup
- then update again

This is an extra query... I would prefer to keep the usual use case here as single query.

lis2: This is a tricky part. TopicUser record are created in two cases. When the user visits the topic for the first time or when post which was already seen by the user is moved to another topic (then we create topic_user with notification level set as regular)

So logic we have here is
- If there is no topic_user record, create one and while you are creating, check if the user wanted to automatically and immediately track topic 

- if there is already topic_user, without notifications_reason_id (so it was created when the post was moved and notification level was not changed by the user) - check if the user wanted to automatically track topic and do that for them. 

I was thinking as well if we should set correct notification level when a post is moved and we create that topic_user. However, if the user was tracking the original topic, it doesn't mean that they want to track automatically new topic to which post was moved to

I will try to solve without adding a new query

lis2:
I changed that flow a little bit:
1. I extracted logic which is finding notification level based on `automatic track topic I enter` and etc. to separate method called `set_notification_level_attribute`
2. Then here: https://github.com/discourse/discourse/pull/9237/files#diff-a453d5e1c29398203488ff879ca69664R175 
We track topic_user and if that is not having `notifications_reason_id` (notification level was not chosen by the user) we try to assign automatic one.
It added one additional query, previously it was just an update query, now we got select and update. It is in defer block so it should not affect users
```
    Scheduler::Defer.later "Track Visit" do
      TopicViewItem.add(topic_id, ip, user_id)
      TopicUser.track_visit!(topic_id, user_id) if track_visit
    end
```
3. If notification level is changed we are sending information to frontend to properly update UI (same thing is happening if your settings are set to for example automatically track topic after 30 seconds)
```
notification_level_change(user_id, topic_id, attrs[:notification_level], attrs[:notifications_reason_id]) if attrs[:notification_level]
```

@SamSaffron - could you take another look? What do you think?

SamSaffron:
I am still super defensive of queries here cause this happens every time you visit a topic.

The change means we load the topic, then update it.

Previous implementation was that we update it... if update fails we create

So at a minimum it is 2 queries now and you pull a full model down. 

I prefer that it just updates, if update fails, do a create cause usually you are updating cause you tend to visit topics twice or more.

PR #9241, opened 2020-03-19
DEV: allows to run up/down migrations of a plugin

jjaffeux:
This is mostly useful while developping a plugin, to avoid manual actions of deleting tables and schema_migrations rows.

Usage:
bundle exec rake plugin:migrate:down[discourse-calendar]

### DIFF
--- a/lib/tasks/plugin.rake
+++ b/lib/tasks/plugin.rake
@@ -146,3 +146,32 @@ task 'plugin:qunit', [:plugin, :timeout] do |t, args|
 
   sh cmd
 end
+
+desc 'run all migrations of a plugin'
+namespace 'plugin:migrate' do
+  def list_migrations(plugin_name)
+    plugin_root = File.join(Rails.root, "plugins", plugin_name)
+    migrations_root = File.join(plugin_root, "db", "migrate", "*.rb")
+    Dir[migrations_root]
+      .map do |migration_filename|
+        File.basename(migration_filename)[/(^.*?)_/, 1].to_i
+      end
+      .sort
+  end
+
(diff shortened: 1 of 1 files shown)


lib/tasks/plugin.rake
```diff
@@ -146,3 +146,32 @@ task 'plugin:qunit', [:plugin, :timeout] do |t, args|
 
   sh cmd
 end
+
+desc 'run all migrations of a plugin'
+namespace 'plugin:migrate' do
+  def list_migrations(plugin_name)
+    plugin_root = File.join(Rails.root, "plugins", plugin_name)
+    migrations_root = File.join(plugin_root, "db", "migrate", "*.rb")
```
davidtaylorhq: I think we might need `post_migrate` in there as well [e.g.](https://github.com/discourse/discourse/tree/master/plugins/poll/db/post_migrate)

jjaffeux: Can you confirm this is correct:
- one migration in january
- one post migratrion in february
- one migration in march

the post migration has to happen before march for up and after for down ?

davidtaylorhq: I think for this purpose we should just run them in order of date. That's how it works for self-hosters, and in development. The only time it makes a difference is when you set RUN_POST_DEPLOYMENT_MIGRATIONS=0

jjaffeux: oh so I can just mix them all together, cool

jjaffeux: Ok That should do it now :+1:

PR #9242, opened 2020-03-19
DEV: Fix inconsistent time-sensitive specs

CvX:
@martin-brennan The following spec now consistently fails, is it correct? 
https://github.com/discourse/discourse/blob/9a4e33c3f3797479ac1768298bbf4edb6c8e2bf6/spec/tasks/uploads_spec.rb#L83-L95

From what I gathered `UploadSecurity#uploading_in_secure_context?` is false for all uploads in this test, so these posts don't get rebaked.


martin-brennan:
Closed because I messed this PR up, see https://github.com/discourse/discourse/pull/9283

PR #9248, opened 2020-03-20
FEATURE: Show votes in an "on voted" poll to the creator

riking:
Implements https://meta.discourse.org/t/poll-creator-should-be-able-to-see-poll-results-even-when-shown-on-vote/138108

Had to plumb the Guardian through to the serializer. Notably, the client will not show results _by default_ to the poll author if they haven't voted - they need to click the button.


riking:
(rebased & updated commit message)

PR #9249, opened 2020-03-20
FIX: Backfill topic timer duration

romanrizzi:
Since `TopicTimer#duration` can be nil for existing timers, we should backfill them or `TopicStatusUpdater#message_for_autoclosed` will fail.

cc @vinothkannans


romanrizzi:
Done 🙆‍♂  @eviltrout

PR #9254, opened 2020-03-22
FIX: allows adapters to define a custom primaryKey

jjaffeux:


jjaffeux:
> It seems fine to me, but I am curious what the motivation for the change is? In what situation is `id` not available?

rails allows you to override primary_key of a model. And it has bitten me multiple times last year, and I just decided to fix it :p

I'm about to write an article about using store instead if ajax as many people don't use store, and I want to reduce at the minimum the limitations of using store.

PR #9259, opened 2020-03-23
DEV: Add test case for /srv/status probers

riking:
Test case I wrote while investigating https://meta.discourse.org/t/do-hits-to-srv-status-count-as-crawlers/143431/13?u=riking .

Figured I'd submit it to ensure we don't regress here.


SamSaffron:
seems safe to me, cool

PR #9260, opened 2020-03-23
FIX: Use 1 column instead of 4 for permalink destination

riking:
Fixes a major UX issue with the permalinks page. Having 4 columns meant it was nearly impossible to see where the links were actually going.

After:

image

Not sure about including the type in the field - thoughts?


coding-horror:
fantastic change 😍

SamSaffron:
I love the change, feel free to merge it in once you apply the feedback from dev.

riking:
New screenshot:

image

PR #9265, opened 2020-03-24
UX: Align bulk select menu toggle relative to main wrapper when possible

awesomerobot:
The goal is to align the bulk-select menu toggle relative to the content rather than the viewport as discussed here: https://meta.discourse.org/t/better-bulk-select-wrench-icon-placement/141396

I'm grabbing the padding on the top of `#main-outlet` so I can set the `top` value for `position:sticky;`... this should ensure that if the header height is increased/decreased the button would still align with the content (because the padding would also need to be increased/decreased).

Also checking for `position: sticky` support in the CSS and falling back to the current implementation if it's not. 

Before:
Screen Shot 2020-03-24 at 2 39 59 PM

After: 
Screen Shot 2020-03-24 at 2 39 21 PM

### DIFF
--- a/app/assets/javascripts/discourse/components/bulk-select-button.js
+++ b/app/assets/javascripts/discourse/components/bulk-select-button.js
@@ -4,6 +4,18 @@ import showModal from "discourse/lib/show-modal";
 export default Component.extend({
   classNames: ["bulk-select-container"],
 
+  didInsertElement() {
+    this._super(...arguments);
+
+    let mainOutletPadding =
+      window.getComputedStyle(document.querySelector("#main-outlet"))
+        .paddingTop || 0;
+
+    document.querySelector(
+      ".bulk-select-container"
+    ).style.top = mainOutletPadding;
+  },
+
   actions: {
     showBulkActions() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/bulk-select-button.js
```diff
@@ -4,6 +4,18 @@ import showModal from "discourse/lib/show-modal";
 export default Component.extend({
   classNames: ["bulk-select-container"],
 
+  didInsertElement() {
+    this._super(...arguments);
+
+    let mainOutletPadding =
```
jjaffeux: I think should be in an afterRender block

awesomerobot: ok, just updated it... can you double-check that I went about it the right way?

jjaffeux: Yes should be good, your only risk now is that the node is actually not available because some fast transition removed the component before it rendered but still called the block. That could show an error but wouldn't cause any crash, so no need to worry about it probably.

PR #9279, opened 2020-03-25
Completely remove read only mode during backups

romanrizzi:
We removed the `readonly_mode_during_backup` setting in #9112 and according to [this](https://meta.discourse.org/t/when-to-use-readonly-mode-during-backup/136009/5) read-only mode is not necessary for backups anymore. I didn't notice that only removing the setting would always enable read-only mode during backups.

### DIFF
--- a/lib/backup_restore/backuper.rb
+++ b/lib/backup_restore/backuper.rb
@@ -34,9 +34,6 @@ module BackupRestore
 
       update_metadata
 
-      ### READ-ONLY / START ###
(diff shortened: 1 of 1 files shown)


eviltrout:
Looks good to me but I'll defer to @gschlager

gschlager:
> I didn't notice that only removing the setting would always enable read-only mode during backups.

No worries. I missed that as well. :wink:

PR #9284, opened 2020-03-26
FIX: First pass to improve efficiency of secure uploads rake task

martin-brennan:
Get rid of harmful each loop over uploads to update. Instead we put all the unique access control posts for the uploads into a map for fast access (vs using the slow .find through array) and look up the post when it is needed when looping through the uploads in batches.

On a Discourse instance with ~93k uploads, a simplified version of the old method takes **> 1 minute**:

```
unique_access_control_posts = Post.unscoped.select(:id, :topic_id).includes(topic: :category).where(id: uploads_to_update.pluck(:access_control_post_id).uniq)
uploads_to_update.each do |upload|
  upload.access_control_post = unique_access_control_posts.find { |post| post.id == upload.access_control_post_id }
end
```

A simplified version of the new method takes **~18s** and uses a lot less memory:

```
unique_access_control_posts = {}
Post.unscoped.select(:id, :topic_id).includes(topic: :category)
       .where(id: uploads_to_update.pluck(:access_control_post_id).uniq).find_each do |post|
  unique_access_control_posts[post.id] = post
end

uploads_to_update.find_each do |upload|
  upload.access_control_post = unique_access_control_posts[upload.access_control_post_id]
end
```

### DIFF
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -710,15 +710,6 @@ task "uploads:secure_upload_analyse_and_update" => :environment do
       # secure based on site settings.
       uploads_to_update = Upload.includes(:posts, :optimized_images).joins(:post_uploads)
 
-      # we do this to avoid a heavier post query, and to make sure we only
-      # get unique posts AND include deleted posts (unscoped)
-      unique_access_control_posts = {}
-      Post.unscoped.select(:id, :topic_id)
-        .includes(topic: :category)
-        .where(id: uploads_to_update.pluck(:access_control_post_id).uniq).find_each do |post|
-        unique_access_control_posts[post.id] = post
-      end
-
       puts "There are #{uploads_to_update.count} upload(s) that could be marked secure.", ""
 
       # Simply mark all these uploads as secure if login_required because no anons will be able to access them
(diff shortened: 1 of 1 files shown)


lib/tasks/uploads.rake
```diff
@@ -840,8 +833,20 @@ def determine_upload_security_and_posts_to_rebake(uploads_to_update, mark_secure
   uploads_to_adjust_acl_for = []
   posts_to_rebake = {}
 
+  # we do this to avoid a heavier post query, and to make sure we only
+  # get unique posts AND include deleted posts (unscoped)
+  unique_access_control_posts = {}
+  Post.unscoped.select(:id, :topic_id)
+    .includes(topic: :category)
+    .where(id: uploads_to_update.pluck(:access_control_post_id).uniq).find_each do |post|
+    unique_access_control_posts[post.id] = post
```
SamSaffron: Does this not pull a huge number of posts into memory? Why not just keep the ids and then instead just update `access_control_post_id = id` ?

martin-brennan: @SamSaffron because the post's topic and that topic's category are used in `UploadSecurity` to determine whether the post has secure media. See:

https://github.com/discourse/discourse/blob/master/lib/upload_security.rb#L51-L53
https://github.com/discourse/discourse/blob/master/app/models/post.rb#L500-L504

martin-brennan: In the forum I have tested on this results in ~3400 posts in memory for ~93k uploads.

discoursereviewbot: [SamSaffron posted](https://review.discourse.org/t/fix-first-pass-to-improve-efficiency-of-secure-uploads-rake-task-pr-9284/10096/5):

I see, then why not just keep a list of ids, then you do a `Post.find(id)` in the loop. That way you don't keep 3000 posts in memory.

PR #9285, opened 2020-03-26
FIX: custom SQL with a trailing comment might break BadgeGranter SQL

Supermathie:
For example given a custom badge with SQL:
```
SELECT 1
-- I am a comment
```

You end up with

```
FROM (SELECT 1
-- I am a comment) q
```

This fix adds newlines so you end up with the now-valid:

```
FROM (
  SELECT 1
-- I am a comment
) q
```


danielwaterworth:
@Supermathie, there are other lines with similar problems in the same file, https://github.com/discourse/discourse/blob/7180fcd39f20ca2a77ed33d4b09960193033e97c/app/services/badge_granter.rb#L225 for example.

PR #9286, opened 2020-03-26
Ember template linting: rc config & easy fixes

riking:
Clean up the .template-lintrc.js, and enable the "recommended" category with explicit disables for everything we're failing. As these lints get fixed, they can be flipped on.


app/assets/javascripts/discourse/templates/composer.hbs
```diff
@@ -213,7 +213,7 @@
       {{else}}
         <div class='saving-text'>
           {{#if model.createdPost}}
-            {{i18n 'composer.saved'}} <a class='permalink' href={{createdPost.url}} {{action "viewNewReply"}}>{{i18n 'composer.view_new_post'}}</a>
+            {{i18n 'composer.saved'}} <a href class='permalink' href={{createdPost.url}} {{action "viewNewReply"}}>{{i18n 'composer.view_new_post'}}</a>
```
jjaffeux: we have two hrefs here

PR #9290, opened 2020-03-27
FEATURE: High priority bookmark reminder notifications

martin-brennan:
Introduce the concept of "high priority notifications" which include PM and bookmark reminder notifications. Now bookmark reminder notifications act in the same way as PM notifications (float to top of recent list, show in the green bubble) and most instances of `unread_private_messages` in the UI have been replaced with `unread_high_priority_notifications`.

The user email digest is changed to just have a section about unread high priority notifications, the unread PM section has been removed.

A `high_priority` boolean column has been added to the Notification table and relevant indices added to account for it.

`unread_private_messages` has been kept on the `User` model purely for backwards compat, but now just returns `unread_high_priority_notifications` count so this may cause some inconsistencies in the UI.


SamSaffron:
There is an index change we need as well to account for this: 

https://github.com/discourse/discourse/blob/40b6e278a08d58ffd9525c52318daac397465b37/app/models/notification.rb#L288-L288

https://github.com/discourse/discourse/blob/40b6e278a08d58ffd9525c52318daac397465b37/app/models/notification.rb#L291-L291

Without this the whole thing will be super slow on many sites

martin-brennan:
@SamSaffron good catch I did not see that index

db/migrate/20200329222246_add_priority_column_to_notifications.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class AddPriorityColumnToNotifications < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+  def up
+    if !column_exists?(:notifications, :priority)
+      add_column :notifications, :priority, :integer, default: nil
```
SamSaffron: integer is huge... instead I would say since this is binary ... boolean is enough for now, maybe tinyint or byte.

app/assets/stylesheets/common/base/discourse.scss
```diff
@@ -422,7 +422,8 @@ table {
   align-items: center;
 }
 
-.unread-private-messages {
+.unread-private-messages,
```
SamSaffron: do we need the class `.unread-private-messages` ?

app/models/notification.rb
```diff
@@ -64,10 +68,10 @@ def self.purge_old!
   end
 
   def self.ensure_consistency!
-    DB.exec(<<~SQL, Notification.types[:private_message])
+    DB.exec(<<~SQL)
       DELETE
         FROM notifications n
-       WHERE notification_type = ?
+       WHERE high_priority = TRUE
```
SamSaffron: `where high_priority` should do the trick, I think the capital true is a bit too verbose.

app/models/notification.rb
```diff
@@ -280,14 +295,17 @@ def send_email
...
 #  idx_notifications_speedup_unread_count                       (user_id,notification_type) WHERE (NOT read)
 #  index_notifications_on_post_action_id                        (post_action_id)
-#  index_notifications_on_read_or_n_type                        (user_id,id DESC,read,topic_id) UNIQUE WHERE (read OR (notification_type <> 6))
 #  index_notifications_on_topic_id_and_post_number              (topic_id,post_number)
 #  index_notifications_on_user_id_and_created_at                (user_id,created_at)
-#  index_notifications_on_user_id_and_id                        (user_id,id) UNIQUE WHERE ((notification_type = 6) AND (NOT read))
 #  index_notifications_on_user_id_and_topic_id_and_post_number  (user_id,topic_id,post_number)
+#  index_notifications_read_or_not_high_priority                (user_id,id DESC,read,topic_id) WHERE (read OR (high_priority = false))
+#  index_notifications_unique_unread_high_priority              (user_id,id) UNIQUE WHERE ((NOT read) AND (high_priority = true))
+#  index_notifications_unread_high_priority                     (user_id,high_priority) WHERE ((NOT read) AND (high_priority = true))
```
SamSaffron: hmmm why are we adding 2 extra indexes beyond what we had to start with? every index has a pretty high disk cost on this table

db/migrate/20200329222246_add_high_priority_column_to_notifications.rb
```diff
@@ -0,0 +1,34 @@
...
+  disable_ddl_transaction!
+  def up
+    if !column_exists?(:notifications, :high_priority)
+      add_column :notifications, :high_priority, :boolean, default: nil
+    end
+
+    # type 6 = private message, 24 = bookmark reminder
+    # priority 0 = low, 1 = normal, 2 = high
+    execute <<~SQL
+      UPDATE notifications SET high_priority = TRUE WHERE notification_type IN (6, 24);
```
SamSaffron: hmmm I would still keep it to one statement per execute, that way if it blows up it will be easier to find out where it exploded

db/migrate/20200329222246_add_high_priority_column_to_notifications.rb
```diff
@@ -0,0 +1,34 @@
...
+      UPDATE notifications SET high_priority = FALSE WHERE notification_type NOT IN (6, 24);
+    SQL
+
+    execute <<~SQL
+      ALTER TABLE notifications ALTER COLUMN high_priority SET DEFAULT FALSE;
+      ALTER TABLE notifications ALTER COLUMN high_priority SET NOT NULL;
+    SQL
+
+    execute <<~SQL
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS index_notifications_read_or_not_high_priority ON notifications(user_id, id DESC, read, topic_id) WHERE (read OR (high_priority = FALSE));
```
SamSaffron: You have "if not exists" here, but don't have it above in the alter table ... I am guessing you should be 100% re-runnable is that is the intent here.

db/post_migrate/20200330233427_drop_old_unread_pm_notification_indices.rb
```diff
@@ -0,0 +1,17 @@
...
+class DropOldUnreadPmNotificationIndices < ActiveRecord::Migration[6.0]
+  def up
+    DB.exec("DROP INDEX IF EXISTS index_notifications_on_user_id_and_id")
+    DB.exec("DROP INDEX IF EXISTS index_notifications_on_read_or_n_type")
+  end
+
+  def down
+    add_index :notifications, [:user_id, :id], unique: true, where: 'notification_type = 6 AND NOT read'
+    execute <<~SQL
+      CREATE UNIQUE INDEX CONCURRENTLY index_notifications_on_read_or_n_type
```
SamSaffron: I don't think this will work cause it will still have a DDL transaction, just chuck a not allowed on down?

SamSaffron:
OK  pending any overnight reviews, let's merge it first thing tomorrow morning! Then we have a full day to watch it!

PR #9294, opened 2020-03-27
DEV: allows to decorate topic list item

jjaffeux:


app/assets/javascripts/discourse/components/topic-list-item.js
```diff
@@ -67,6 +69,16 @@ export default Component.extend({
         }
       });
     }
+
+    schedule("afterRender", () => {
+      const rawTopicLink = this.element.querySelector(".raw-topic-link");
```
eviltrout: Do we need a check here in case `this.element` is `null` say on destroy?

jjaffeux: yes I was thinking of it, at the minimum I think it's safer will add a guard :+1:

app/assets/javascripts/discourse/lib/plugin-api.js
```diff
@@ -1017,6 +1018,24 @@ class PluginApi {
   decoratePluginOutlet(outletName, callback, opts) {
     addPluginOutletDecorator(outletName, callback, opts || {});
   }
+
+  /**
+   * Allows to alter topic list item node
```
davidtaylorhq:
```suggestion
   * Allows altering the topic title in the topic list, and in the topic view
```

Also would be useful to add the possible values of `topicTitleType`

PR #9295, opened 2020-03-28
FEATURE: Add anonymous topic read setting when login is required

joallard:
This adds a setting to allow reading topics even when login is required.

### The problem I'm trying to solve
We have a gated community for our coworking coop where login is required, but privacy isn't of capital importance. It's just a nice add-on.

Users are still not totally on board with the forum, so we often share forum links by email, on Slack, Facebook, etc.

When this is done: (1) no preview appears, and (2) the linked topic is blank when you're logged out.

It's been talked about on Meta [about Slack](https://meta.discourse.org/t/allow-slack-to-unfurl-expand-links-to-login-required-discourse-instance/39547/4), for which there is now a plugin; and a request for ["special links"](https://meta.discourse.org/t/closed-forum-but-special-links-to-access-content/42975). If we look at the Slack topic, at least 95 people have been interested enough to click on the plugin link.

### The dream
It would be nice if we could share topic links as eg Google Docs and other services do it. You click a button, you get a nice hash in your URL to share. However it would take quite some work to get there.

For now, just being able to paste a URL somewhere and have anyone be able to read it is plenty.

### The fix
Going step by step, this setting just allows topic reads for anyone who has the URL. 

Yes, even guessed URLs like `/t/7`. Known limitation. We could make the code more complex by requiring a full URL when we're in this case. Not a road my intuition tells me to take, as this feature could grow in another direction eventually.

I figure this is a good first step since the login_required setting is quite hardcoded in some places. I would argue that permission to read is actually a function of the topic (or more broadly, category).

Other implementation notes:
- The setting name and configuration is maybe not ideal. Would another name fit better? Would it be better to turn `login_required` into a select that enums `:strict (true), :topic_reads_allowed, :no (false)`?

- At present, this is broken. Topic#show shows a spinner, and JS is stuck here:

image

And so am I. I don't know where to go from here. Guidance would be appreciated.

I'm happy to go back to the forum if this is the more appropriate place for this.

I've pondered making this into a plugin, but I have the impression that hooking into `ApplicationController#redirect_to_login_if_required` and other tricky places might set me up for a maintenance failure down the line when the underlying code changes.

Overall guidance, advice and feedback are appreciated. Cheers!

### DIFF
--- a/app/assets/javascripts/discourse/controllers/application.js
+++ b/app/assets/javascripts/discourse/controllers/application.js
@@ -19,7 +19,11 @@ export default Controller.extend({
 
   @discourseComputed
   loginRequired() {
-    return Discourse.SiteSettings.login_required && !this.currentUser;
+    return (
+      Discourse.SiteSettings.login_required &&
+      !Discourse.SiteSettings.allow_reading_topics &&
+      !this.currentUser
+    );
   },
 
   @discourseComputed
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -741,6 +741,10 @@ class ApplicationController < ActionController::Base
     end
 
     if !current_user && SiteSetting.login_required?
+      if SiteSetting.allow_reading_topics && controller_name == "topics"
+        return
+      end
+
       flash.keep
       redirect_to_login
       return
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -1592,6 +1592,7 @@ en:
     invite_only: "All new users must be explicitly invited by trusted users or staff. Public registration is disabled."
 
     login_required: "Require authentication to read content on this site, disallow anonymous access."
+    allow_reading_topics: "If authentication is required, still allow reading topics with the URL."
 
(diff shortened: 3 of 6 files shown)


coding-horror:
Has to be discussed on meta.discourse.org not here

I see an immediate showstopper though:

> this setting just allows topic reads for anyone who has the URL.

That's.. uh.. crazy bad? It's super unlikely we would merge anything like this due to the danger of the setting. Just make your community public, if security is of so little concern. That's FAR FAR easier.

joallard:
Well, as of now, we have login_required, but anybody can sign up and just... read topics. So I wouldn't say it's much worse than that.

> Just make your community public, if security is of so little concern. That's FAR FAR easier.

Quite clearly, yes.

However as I pointed out, that is not my use case. I was hoping for a way when you don't want to feel like the eyes of the internet are upon you, but people with the URL (assuming it can't be guessed with just sequential ids and need at least title) is fine. I can say that the need exists in one place, at least.

If merging is unlikely and there's no fix insight, then looks like I need to make a plugin that would patch ApplicationController and cope with any future change. Maybe.

Anyway, sounds like this shouldn't be here.

PR #9296, opened 2020-03-28
DEV: Allows moduleForWidget to use before/afterEach options

jjaffeux:


test/javascripts/helpers/widget-test.js
```diff
@@ -1,7 +1,15 @@
...
 
-export function moduleForWidget(name) {
-  moduleForComponent(name, `widget:${name}`, { integration: true });
+export function moduleForWidget(name, options = {}) {
+  moduleForComponent(
+    name,
+    `widget:${name}`,
+    Object.assign(
+      { integration: true },
+      getProperties(options, ["beforeEach", "afterEach"])
```
eviltrout: `{ beforeEach: options.beforeEach, afterEach: options.afterEach }` is simpler and doesn't require an import.

This isn't a dealbreaker for me but I would prefer it. If you disagree feel free to merge.

PR #9297, opened 2020-03-28
DEV: adds a new dropdown widget usable in any widget

jjaffeux:
image

Note this commit needs: https://github.com/discourse/discourse/pull/9296 tests will fail until this is merged.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/dropdown.js
+++ b/app/assets/javascripts/discourse/widgets/dropdown.js
@@ -143,10 +143,6 @@ export const DropdownClass = {
     if (!attrs.label && !attrs.translatedLabel) {
       throw "A widget-dropdown expects at least a `label` or `translatedLabel`";
     }
-
-    if (!attrs.onChange) {
-      throw "A widget-dropdown expects an `onChange` handler.";
-    }
   },
 
   buildKey: attrs => {
--- a/app/assets/javascripts/discourse/widgets/dropdown.js
+++ b/app/assets/javascripts/discourse/widgets/dropdown.js
@@ -1,0 +1,267 @@
+import { createWidget } from "discourse/widgets/widget";
+import hbs from "discourse/widgets/hbs-compiler";
+
+/*
+
+  dropdown widget
+
+  Usage
+  -----
+
+  {{attach
+    widget="dropdown"
+    attrs=(hash
+      id=id
+      label=label
+      content=content
+      onChange=onChange
+      options=(hash)
+    )
+  }}
+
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/widgets/dropdown.js
```diff
@@ -0,0 +1,266 @@
...
+
+createWidget("dropdown-header", DropdownHeaderClass);
+
+export const DropdownItemClass = {
+  tagName: "div",
+
+  transform(attrs) {
+    return {
+      content:
+        attrs.item === "separator"
```
eviltrout: I can't decide if I hate this or not :smiley: 

On one hand nested ternary operators make my brain hurt, but on the other I fully understand what this is doing. I am leaning towards "let's keep it."

app/assets/javascripts/discourse/widgets/dropdown.js
```diff
@@ -0,0 +1,266 @@
...
+      [
+        { id: 1, label: "foo.bar" },
+        "separator",
+        { id: 2, translatedLabel: "FooBar" },
+        { id: 3 label: "foo.baz", icon: "times" }
+      ]
+      ```
+
+    - options: accepts a hash of optional attributes
+      - smallHeader: true/false, if true "btn-small" class will be added to header.
```
CvX: A bool for adding a class doesn't scale well if you'd ever want to add another class to that header/button. Wouldn't explicitly passing a class list be better?

app/assets/javascripts/discourse/widgets/dropdown.js
```diff
@@ -0,0 +1,266 @@
...
+    )
+  }}
+
+  Mandatory attributes:
+
+    - id: must be unique in the application
+
+    - label: an i18n key to be translated and displayed on the header
+
+    - translatedLabel: an already translated label to display on the header
```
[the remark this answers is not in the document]
jjaffeux: Yes right will clarify this :+1:

app/assets/javascripts/discourse/widgets/dropdown.js
jjaffeux: Yes I feel the same :D

app/assets/javascripts/discourse/widgets/dropdown.js
jjaffeux: yes agreed it feels a little bit hacky... I just thought this is a very common case, so I could make it explicit as `btn-small` requires you to know the internals of the dropdown

PR #9300, opened 2020-03-28
FIX: correctly take category/group filters into csv export

jjaffeux:


spec/jobs/export_csv_file_spec.rb
```diff
@@ -97,6 +97,23 @@
...
 
+    it 'works with filters' do
+      user.user_visits.create!(visited_at: '2010-01-01', posts_read: 42)
+
+      group = Fabricate(:group)
+      user1 = Fabricate(:user)
+      group_user = Fabricate(:group_user, group: group, user: user1)
+      user1.user_visits.create!(visited_at: '2010-01-03', posts_read: 420)
+
+      exporter.instance_variable_get(:@extra)['name'] = 'visits'
```
eviltrout: I think it would be nicer to add an `attr_reader` for this so you could do `exporter.extra` instead. 

I know there's an argument that we shouldn't add interfaces for tests, but in this case exposing the extra arguments is not a bad thing to have anyway.

PR #9301, opened 2020-03-28
DEV: trigger :username_changed when a user is renamed

jjaffeux:


jjaffeux:
@gschlager do you think this is the right place to hook this ?

PR #9305, opened 2020-03-29
FIX: ensures unchanged draft model is correctly clear on escape

jjaffeux:
This fix will also make sure the bootbox will be correctly shown when using escape.

Test cases:
- edit post, wait for draft check, escape => should totally hide composer
- edit post, add a char, escape => should show bootbox, and reduce or hide depending on choice

Same behavior is expected when pressing on cancel.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/composer.js
+++ b/app/assets/javascripts/discourse/controllers/composer.js
@@ -1062,7 +1062,6 @@ export default Controller.extend({
             }
           }
         );
-        // debugger;
       } else {
         // it is possible there is some sort of crazy draft with no body ... just give up on it
         this.destroyDraft().then(destroyedDraftModel => {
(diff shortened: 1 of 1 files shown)


jjaffeux:
@nbianca @SamSaffron can you carefully review this as you kind of know the draft code very well please ?

@eviltrout do you see a better way to prevent the loss of the context in `Draft.clear` ? (a comment explains it in my commit) and maybe a better explanation than me please?

SamSaffron:
I am not sure about this fix, instead when you click edit it should not be saving a draft until something changes. To me that is the big bug here.

jjaffeux:
> I guess there is the case of ... hit edit ... add a... remove a... hit escape and it should operate same as cancel.
> 
> What worries me though is that I don't want to change ESC minimizing if there is any legit draft in play.


Im a little bit lost here, because these two sentences are opposite I think. Currently when you "ESC" with a legit draft it minimises, when you press [cancel] it shows the bootbox. So they are operating diffrently, and say esc should operate as cancel, but you then say you don't can to change "ESC".

PR #9308, opened 2020-03-29
FEATURE: allows multiple custom emoji groups

jjaffeux:
Previous PR suffered a bug from github so I'm reopening it here.

mobile:
Screenshot 2020-01-16 at 12 49 23

desktop:
Screenshot 2020-01-16 at 12 48 53

Few notes about it:
- it adds multiple tests which were totally missing (registerEmoji and custom emoji

 plugin instance)
- I took this opportunity to rework the customise emoji page:

Screenshot 2020-03-29 at 22 16 19

It will also now give visual feedback that an emoji is getting uploaded:
image

You can also now filter by groups:
Screenshot 2020-03-29 at 22 22 02

Also the recommended way to add translations for custom emoji groups is via theme component:

```
<script>
    I18n.translations.en.js.emoji_picker.group_name = "Group name";
</script>
```

### DIFF
--- a/app/assets/javascripts/discourse/components/emoji-picker.js
+++ b/app/assets/javascripts/discourse/components/emoji-picker.js
@@ -14,9 +14,26 @@ import ENV, { INPUT_DELAY } from "discourse-common/config/environment";
 const { run } = Ember;
 
 const PER_ROW = 11;
-const customEmojis = _.keys(extendedEmojiList()).map(code => {
-  return { code, src: emojiUrlFor(code) };
-});
+function customEmojis() {
+  const list = extendedEmojiList();
+  const emojis = Object.keys(list)
+    .map(code => {
+      const { group } = list[code];
+      return {
+        code,
+        src: emojiUrlFor(code),
+        group,
+        key: `emoji_picker.${group || "default"}`
+      };
+    })
+    .reduce((acc, curr) => {
+      if (!acc[curr.group]) acc[curr.group] = [];
+      acc[curr.group].push(curr);
+      return acc;
+    }, {});
+
+  return Object.values(emojis);
+}
 
 export default Component.extend({
   automaticPositioning: true,
@@ -35,7 +52,9 @@ export default Component.extend({
   },
 
   show() {
-    const template = findRawTemplate("emoji-picker")({ customEmojis });
+    const template = findRawTemplate("emoji-picker")({
+      customEmojis: customEmojis()
+    });
     this.$picker.html(template);
 
     this.$filter = this.$picker.find(".filter");
@@ -579,7 +598,7 @@ export default Component.extend({
       this.$picker.width() -
       this.$picker.find(".categories-column").width() -
       this.$picker.find(".diversity-picker").width() -
-      32;
+      60;
     this.$picker.find(".info").css("max-width", infoMaxWidth);
   },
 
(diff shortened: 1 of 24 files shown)


lib/plugin/instance.rb
```diff
@@ -6,21 +6,29 @@
 require_dependency 'auth'
 
 class Plugin::CustomEmoji
+  CACHE_KEY = "plugin-emoji"
```
CvX:
```suggestion
  CACHE_KEY ||= "plugin-emoji"
```

app/assets/javascripts/discourse/components/emoji-uploader.js
```diff
@@ -1,31 +1,61 @@
 import { notEmpty, not } from "@ember/object/computed";
+import { action } from "@ember/object";
 import Component from "@ember/component";
-import discourseComputed from "discourse-common/utils/decorators";
+import { default as discourseComputed } from "discourse-common/utils/decorators";
```
CvX: We'll need to agree on a single style here. 😅 Because I've been doing the opposite:

```js
//  Replacing e.g.
import { default as computed, other } from "x";
// with
import computed, { other } from "x";
```

app/assets/javascripts/discourse/components/emoji-picker.js
```diff
@@ -14,9 +14,28 @@ import ENV, { INPUT_DELAY } from "discourse-common/config/environment";
...
+
+  Object.keys(list).forEach(code => {
+    const { group } = list[code];
+    emojis.push({
+      code,
+      src: emojiUrlFor(code),
+      group,
+      key: `emoji_picker.${group}`
+    });
+  });
```
CvX: Nit pick, but instead of `.forEach()` and `.push()`, I'd use `.map()`.

app/assets/javascripts/discourse/templates/emoji-picker.hbr.erb
```diff
@@ -9,10 +9,12 @@
     </div>
   <% end %>
 
-  <% if !Emoji.custom.blank? %>
-    <div class='category-icon'>
-      <button data-tabicon="<%= Emoji.custom.first.name %>" type="button" class="emoji" tabindex="-1" data-section="ungrouped" title="{{i18n 'emoji_picker.custom'}}"></button>
-    </div>
+  <% Emoji.custom.group_by { |emoji| emoji.group }.each do |group, emojis| %>
+    <% if !emojis.blank? %>
```
CvX:
```suggestion
    <% if emojis.present? %>
```

app/models/emoji.rb
```diff
@@ -6,9 +6,11 @@ class Emoji
 
   FITZPATRICK_SCALE ||= [ "1f3fb", "1f3fc", "1f3fd", "1f3fe", "1f3ff" ]
 
+  DEFAULT_GROUP = "default"
```
CvX:
```suggestion
  DEFAULT_GROUP ||= "default"
```

app/assets/javascripts/discourse/templates/emoji-picker.hbr.erb
```diff
@@ -49,18 +51,22 @@
...
+      {{#if emojis.length}}
+        <div class='section' data-section='custom-{{emojis.firstObject.group}}'>
+          <div class='section-header'>
+            <span class="title">
+              {{i18n emojis.firstObject.key}}
+            </span>
+          </div>
+          <div class='section-group'>
+            {{#each emojis as |emoji|}}
+              <button style="background-url: url("{{emoji.src}}")" type="button" class="emoji" tabindex="-1" title="{{emoji.code}}"></button>
```
eviltrout: This likely won't pass linting. Is there a way to avoid an inline style?

jjaffeux: I guess it will because it's not new code :)

PR #9313, opened 2020-03-30
Update DEVELOPER-ADVANCED.md

benbowler:
Since Ubuntu 12.04 `software-properties-properties` has been renamed `software-properties-common`


SamSaffron:
legit change, thanks!

PR #9314, opened 2020-03-30
FIX: jobs/delete_replies: Add Time+Duration, not Time+Time

riking:
Fixes `Job exception: can't convert ActiveSupport::TimeWithZone into an exact number`

### DIFF
--- a/app/jobs/regular/delete_replies.rb
+++ b/app/jobs/regular/delete_replies.rb
@@ -22,7 +22,7 @@ module Jobs
         PostDestroyer.new(topic_timer.user, post, context: I18n.t("topic_statuses.auto_deleted_by_timer")).destroy
       end
 
-      topic_timer.execute_at = (replies.minimum(:created_at) || Time.zone.now) + topic_timer.duration.days.ago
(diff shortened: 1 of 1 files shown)


app/jobs/regular/delete_replies.rb
```diff
@@ -22,7 +22,7 @@ def execute(args)
         PostDestroyer.new(topic_timer.user, post, context: I18n.t("topic_statuses.auto_deleted_by_timer")).destroy
       end
 
-      topic_timer.execute_at = (replies.minimum(:created_at) || Time.zone.now) + topic_timer.duration.days.ago
+      topic_timer.execute_at = (replies.minimum(:created_at) || Time.zone.now) + topic_timer.duration.days
```
eviltrout: Can we get a test for this?

SamSaffron:
@vinothkannans  can you make sure there is a test here ?

vinothkannans:
A spec test is added in the commit https://github.com/discourse/discourse/commit/82201fa466bda0d620a65e1f143b5a41b900b81a. Thanks for the fix @riking :+1:

PR #9317, opened 2020-03-31
FEATURE: Add lazy loading to user bookmarks list

martin-brennan:
This is so users with huge amount of bookmarks do not have to wait a long time to see results.

* Add a bookmark list and list serializer to server-side to be able to handle paging and load more URL
* Use load-more component to load more bookmark items, 20 at a time in user activity
* Change the way current user is loaded for bookmark ember models because it was breaking/losing resolvedTimezone when loading more items


SamSaffron:
Are we preloading here? when you hit refresh on the page, ideally it should not make an extra http request.

martin-brennan:
@SamSaffron  a question RE: Preloading. All of the other links in user activity send off another HTTP request:

image

Just wondering why preload the bookmarks and nothing else? Also would we add the bookmarks to the `UserSerializer` so they get preloaded here https://github.com/discourse/discourse/blob/master/app/controllers/users_controller.rb#L87-L91 or preload them separately in the same place? This is the route that is hit when you go to the u/:username/activity/:filter URL.

PR #9323, opened 2020-03-31
FEATURE: add setting `auto_approve_email_domains` to auto approve users

techAPJ:
This commit adds a new site setting `auto_approve_email_domains` to
auto approve users based on their email address domain.

Note that if a domain already exists in `email_domains_whitelist` then
`auto_approve_email_domains` needs to be duplicated there as well,
since users won’t be able to register with email address that is
not allowed in `email_domains_whitelist`.


config/locales/server.en.yml
```diff
@@ -1582,6 +1582,7 @@ en:
     allow_index_in_robots_txt: "Specify in robots.txt that this site is allowed to be indexed by web search engines. In exceptional cases you can permanently <a href='%{base_path}/admin/customize/robots'>override robots.txt</a>."
     email_domains_blacklist: "A pipe-delimited list of email domains that users are not allowed to register accounts with. Example: mailinator.com|trashmail.net"
     email_domains_whitelist: "A pipe-delimited list of email domains that users MUST register accounts with. WARNING: Users with email domains other than those listed will not be allowed!"
+    auto_approve_email_domains: "A list of email domains that users register accounts with for auto approval."
```
```suggestion
    auto_approve_email_domains: "Users with email addresses from this list of domains will be automatically approved."
```

techAPJ:
Thanks for the review @eviltrout ❤️

PR #9324, opened 2020-03-31
FIX: Restore failed if schema contained objects not owned by the current DB user

gschlager:


rudygodoy:
This is a real issue right now. Can't even do `rebuild` to get back on track.

rudygodoy:
> This is a real issue right now. Can't even do `rebuild` to get back on track.

Nevermind. The issue was actually related to the `discourse-emoji` plugin.

PR #9326, opened 2020-03-31
FIX: Bypass serviceworker cache for auth routes

davidtaylorhq:
This avoids samesite cookie-related issues in iOS PWAs during the authentication flow

### DIFF
--- a/app/assets/javascripts/service-worker.js.erb
+++ b/app/assets/javascripts/service-worker.js.erb
@@ -7,11 +7,15 @@ workbox.setConfig({
   debug: false
 });
 
+var authUrl = "<%= Discourse.base_path %>/auth/";
+
 var cacheVersion = "1";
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/service-worker.js.erb
```diff
@@ -7,11 +7,15 @@ workbox.setConfig({
...
 
+var authUrl = "<%= Discourse.base_path %>/auth/";
+
 var cacheVersion = "1";
 
 // Cache all GET requests, so Discourse can be used while offline
 workbox.routing.registerRoute(
-  new RegExp('.*?'), // Matches all, GET is implicit
+  function(args) {
+    return !(args.url.origin === location.origin && args.url.pathname.startsWith(authUrl));
```
eviltrout: `startsWith` doesn't work in old browsers, but I guess it's not a problem because service workers don't either!

PR #9328, opened 2020-03-31
FIX: FlagSockpuppets should not flag a post if a post of that user was already rejected by staff

romanrizzi:


spec/services/flag_sockpuppets_spec.rb
```diff
@@ -150,5 +150,28 @@
...
+
+      expect(flagged_post.reload.status).to eq(Reviewable.statuses[:rejected])
+    end
+
+    it "doesn't flag the post if another post of the same user was rejected by staff before" do
+      another_post = Fabricate(:post, user: user1)
+      flagged_post = Fabricate(
+        :reviewable_flagged_post,
+        target: another_post, status: Reviewable.statuses[:rejected], target_created_by: another_post.user
+    )
```
eviltrout:
```suggestion
      )
```

PR #9329, opened 2020-04-01
FEATURE: Bookmark reminder type changes and bugfixes

martin-brennan:
### New Reminder Types

* Add a "later this week" reminder which is today + 2 days, will not show if we are on the days Thu-Sun
* Add a "start of next business week" reminder which is 8am Monday

image

### Bugfixes and Tweaks

* Move dates out of translation for reminder types and yield HTML for tap-tile for more customizable content and styling
* Make sure double clicking the bookmark icon in quick access takes users to the new bookmarks-with-reminders page
* Sane default to 8am (start of day) for custom reminder with no time


SamSaffron:
This is looking good! Merge away.

PR #9331, opened 2020-04-01
FIX: jQuery deprecation warning

zcuric:
### This PR
- [x] Updates code to remove [Replace jQuery APIs depraction warning](https://deprecations.emberjs.com/v3.x/#toc_jquery-apis)

### DIFF
--- a/app/assets/javascripts/discourse-common/lib/icon-library.js
+++ b/app/assets/javascripts/discourse-common/lib/icon-library.js
@@ -1,6 +1,7 @@
 import { h } from "virtual-dom";
 import attributeHook from "discourse-common/lib/attribute-hook";
 import deprecated from "discourse-common/lib/deprecated";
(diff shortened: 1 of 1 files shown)


eviltrout:
The linting is failing for this one. Please run prettier.

app/assets/javascripts/discourse-common/lib/icon-library.js
```diff
@@ -1,6 +1,7 @@
 import { h } from "virtual-dom";
 import attributeHook from "discourse-common/lib/attribute-hook";
 import deprecated from "discourse-common/lib/deprecated";
+import jQuery from 'jquery';
```
pmusaraj, line 4: I think this should satisfy Prettier (double instead of single quotes): 
```suggestion
import jQuery from "jquery";
```

PR #9332, opened 2020-04-01
UX: Remove share as link fallback on touch devices

pmusaraj:
Previously, there were three workflows when sharing a topic or a post: 

- sharing URL displayed in popup or modal, in an input or textarea element
- native sharing dialog on touch devices with native sharing support
- and a fallback to the URL displayed as link element on touch devices with no native sharing support

This PR removes the third option. We were only displaying this on non-iOS/non-Android touch devices like laptops with a touch screen, and it was buggy, see https://meta.discourse.org/t/user-experience-on-discourse-with-optional-touch-input/40190/24 for details.

### DIFF
--- a/app/assets/javascripts/discourse/components/share-panel.js
+++ b/app/assets/javascripts/discourse/components/share-panel.js
@@ -42,9 +42,12 @@ export default Component.extend({
 
   didInsertElement() {
     this._super(...arguments);
-
-    const $linkInput = $(this.element.querySelector(".topic-share-url"));
-    window.setTimeout(() => $linkInput.select().focus(), 200);
+    window.setTimeout(() => {
+      const textArea = this.element.querySelector(".topic-share-url");
+      textArea.style.height = textArea.scrollHeight + "px";
+      textArea.focus();
+      textArea.setSelectionRange(0, this.shareUrl.length);
+    }, 200);
   },
 
   actions: {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/components/share-panel.js
```diff
@@ -43,31 +42,12 @@ export default Component.extend({
...
-        $linkInput.remove();
-
-        $linkForTouch.attr("href", shareUrl).text(shareUrl);
-
-        const range = window.document.createRange();
-        range.selectNode($linkForTouch[0]);
-        window.getSelection().addRange(range);
-      }
-    });
+    window.setTimeout(() => {
```
jjaffeux, line 45: setInterval/setTimeout should never be used, we still have some in the codebase I think... but prefer `Ember.run.later`

app/assets/javascripts/discourse/components/share-popup.js
```diff
@@ -35,21 +35,14 @@ export default Component.extend({
...
-      $linkForTouch.text(link);
-      const range = window.document.createRange();
-      range.selectNode($linkForTouch[0]);
-      window.getSelection().addRange(range);
-    }
+    const linkInput = this.element.querySelector("#share-link input");
+    linkInput.value = this.link;
+
+    // Wait for the fade-in transition to finish before selecting the link:
+    window.setTimeout(() => {
```
jjaffeux, line 42: Ember.run.later

jjaffeux:
happy to see this code gone :p

jjaffeux:
This is minor, but I feel like I have to tell this. Given you now use later with 200ms in didInsert, there's a possibility that a transition will be called before 200ms, destroying the element and making this block fail. You can try to add a longer later, and do a transition to see if the consequences are minor (just an error logged) or more problematic, if yes, you should ensure the element has been found.

PR #9338, opened 2020-04-02
FIX: remove word boundary regex (\b) for search result highlights.

vinothkannans:
And removed the jQuery Highlight plugin and created a new es6 highlighting library. Also, since our server-side search is not limited to whole words matching I removed it in our JS highlight library too.

https://github.com/discourse/discourse/commit/d62d258fe5948e609c9cd26bf43aef9d4960c14a


app/assets/javascripts/discourse/components/highlight-search.js
```diff
@@ -1,12 +1,12 @@
...
-import highlightText from "discourse/lib/highlight-text";
+import highlightSearch from "discourse/lib/highlight-search";
 
 export default Component.extend({
   tagName: "span",
 
   _highlightOnInsert: function() {
     const term = this.highlight;
-    highlightText($(this.element), term);
+    highlightSearch($(this.element), term);
```
jjaffeux, line 9: Could we try to avoid using jquery while we are at it ? I see highlight-html is using the node and not jquery object.

app/assets/javascripts/discourse/lib/highlight-html.js
```diff
@@ -0,0 +1,83 @@
...
+
+export default function(node, words, opts = {}) {
+  let settings = {
+    nodeName: "span",
+    className: "highlighted",
+    matchCase: false
+  };
+
+  Object.assign(settings, opts);
+  words = typeof words === "string" ? [words] : words;
```
jjaffeux, line 44: Screenshot 2020-04-03 at 15 21 16

We have `makeArray` for this pattern

app/assets/javascripts/discourse/lib/highlight-html.js
```diff
@@ -0,0 +1,83 @@
...
+  return node;
+}
+
+export function unhighlightHTML(opts = {}) {
+  let settings = {
+    nodeName: "span",
+    className: "highlighted"
+  };
+
+  Object.assign(settings, opts);
```
jjaffeux, line 74: I think it's better to do this:

```
settings = Object.assign({}, settings, opts);
```

This way you don't rely on an implicit mutation of setting but create a new object instead, much more explicit.

app/assets/javascripts/discourse/lib/highlight-html.js
```diff
@@ -0,0 +1,82 @@
...
+
+export default function(node, words, opts = {}) {
+  let settings = {
+    nodeName: "span",
+    className: "highlighted",
+    matchCase: false
+  };
+
+  settings = Object.assign({}, settings, opts);
+  words = Ember.makeArray(words)
```
jjaffeux, line 44: should be called through an import

app/assets/javascripts/discourse/lib/highlight-html.js
```diff
@@ -0,0 +1,82 @@
...
+  let settings = {
+    nodeName: "span",
+    className: "highlighted"
+  };
+
+  settings = Object.assign({}, settings, opts);
+
+  document
+    .querySelectorAll(`${settings.nodeName}.${settings.className}`)
+    .forEach(e => {
```
jjaffeux, line 77: could we use `element` instead of `e` here please?

app/assets/javascripts/discourse/widgets/post-cooked.js
```diff
@@ -52,13 +56,13 @@ export default class PostCooked {
 
     if (highlight && highlight.length > 2) {
       if (this._highlighted) {
-        $html.unhighlight();
+        unhighlightHTML($html[0]);
```
jjaffeux, line 59: probably worth having a variable with a good name, instead of reach for $html[0] each time

jjaffeux:
Only added few minor comments @vinothkannans , seems very good to me 👍 But I think robin should have a look at this too, so I added him.

app/assets/javascripts/discourse/components/highlight-search.js
```diff
@@ -1,12 +1,12 @@
 import Component from "@ember/component";
```
eviltrout, line 1: Renaming `highlight-text` to `highlight-search` will break at least two themes in our repos. Is there a way you can add backwards compatibility with a deprecation?

vinothkannans: sure :+1:

pmusaraj: @eviltrout @vinothkannans Just a heads up, this broke a theme component for me today, previously `highlight-text` was in `discourse/lib/highlight-text`, now it is in `discourse/component/highlight-text`.

vinothkannans: @pmusaraj previously, `highlight-text` was in both `lib` and `component` folders. I added backward compatibility only for the component file since we're using it in ember templates. I will add backward compatibility for lib file too if it's going to break more than one theme/plugin.

PR #9342, opened 2020-04-03
FIX: guardian always got user but sometimes it is anonymous

lis2:
```
  def initialize(user = nil, request = nil)
    @user = user.presence || AnonymousUser.new
    @request = request
  end
```

AnonymouseUser defines `blank?` method
```
  class AnonymousUser
    def blank?
      true
    end
    ...
  end
```
so if we would use `@user.present?` it would be correct, however, just `@user` is always true

### DIFF
--- a/spec/components/guardian_spec.rb
+++ b/spec/components/guardian_spec.rb
@@ -2743,6 +2743,7 @@ describe Guardian do
   end
 
   describe '#can_export_entity?' do
+    let(:anonymous_guardian) { Guardian.new() }
     let(:user_guardian) { Guardian.new(user) }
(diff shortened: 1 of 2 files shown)


spec/components/guardian_spec.rb
```diff
@@ -2743,6 +2743,7 @@
   end
 
   describe '#can_export_entity?' do
+    let(:anonymous_guardian) { Guardian.new() }
```
eviltrout, line 2746:
```suggestion
    let(:anonymous_guardian) { Guardian.new }
```

eviltrout:
Nice fix :+1:

PR #9346, opened 2020-04-03
UX: removes color on categories if no style chosen

jordanvidrine:
Removing Color on categories when "none" is selected for category style

### DIFF
--- a/app/assets/javascripts/discourse/templates/user/messages.hbs
+++ b/app/assets/javascripts/discourse/templates/user/messages.hbs
@@ -1,7 +1,7 @@
 {{#d-section class="user-secondary-navigation" pageClass="user-messages"}}
   {{#unless site.mobileView}}
     {{#if showNewPM}}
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/discourse/templates/user/messages.hbs
```diff
@@ -1,7 +1,7 @@
 {{#d-section class="user-secondary-navigation" pageClass="user-messages"}}
   {{#unless site.mobileView}}
     {{#if showNewPM}}
-      {{d-button class="btn-primary new-private-message" action=(route-action "composePrivateMessage") icon="envelope" label="user.new_private_message"}}
+      {{d-button class="btn-primary new-" action=(route-action "composePrivateMessage") icon="envelope" label="user.new_private_message"}}
```
pmusaraj, line 4: This "new-" class looks weird, can you describe why we are changing this?

PR #9350, opened 2020-04-04
PERF: Cache Category.subcategory_ids

udan11:


app/models/category.rb
```diff
@@ -153,23 +154,40 @@ def reset_topic_ids_cache
...
-      max_category_nesting: SiteSetting.max_category_nesting
-    )
+      end
+  end
+
+  def self.clear_subcategory_ids
+    @@subcategory_ids.clear
+  end
+
+  DiscourseEvent.on(:site_setting_changed) do |name, _, _|
```
davidtaylorhq, line 185: I think we try to keep these types of things in `config/initializers/014-track-setting-changes.rb`, so that it's easier to reason with.

Also, do we need to clear the cache in any other situations? E.g. restoring a backup?

PR #9362, opened 2020-04-06
DEV: Support for `onChange` on `{{text-field}}`

eviltrout:
This will automatically be debounced and only fired when the value
changes.

There is also `onChangeImmediate` which is not debounced in case you
need that, but in almost all cases when observing text in an element you
should debounce.

### DIFF
--- a/app/assets/javascripts/discourse/components/text-field.js
+++ b/app/assets/javascripts/discourse/components/text-field.js
@@ -1,12 +1,13 @@
 import { TextField } from "@ember/component";
 import discourseComputed from "discourse-common/utils/decorators";
 import { siteDir, isRTL, isLTR } from "discourse/lib/text-direction";
-import { next, debounce } from "@ember/runloop";
+import { next, debounce, cancel } from "@ember/runloop";
 
 const DEBOUNCE_MS = 500;
 
 export default TextField.extend({
   _prevValue: null,
+  _timer: null,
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/components/text-field.js
```diff
@@ -11,6 +16,27 @@ export default TextField.extend({
...
+  },
+
+  didUpdateAttrs() {
+    this._super(...arguments);
+    if (this._prevValue !== this.value) {
+      if (this.onChangeImmediate) {
+        next(() => this.onChangeImmediate(this.value));
+      }
+      if (this.onChange) {
+        debounce(this, this._debouncedChange, DEBOUNCE_MS);
```
jjaffeux, line 31: This will throw an exception if you leave page during debounce timer

jjaffeux:
I have a more general question though, why would we want to use {{text-field}} over {{input}} ?

app/assets/javascripts/discourse/components/text-field.js
eviltrout: I now cancel it, good catch.

eviltrout:
> I have a more general question though, why would we want to use {{text-field}} over {{input}} ?

The main reason is to include stuff we always want to do. This used to include a placeholder (which is now doable other ways, but it still has RTL helpers which are useful.

PR #9365, opened 2020-04-06
DEV: changes so plugins can customize invites UI

nlalonde:
Need to add to the "invited" json, and extend the send invite modal.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/share-and-invite.js
+++ b/app/assets/javascripts/discourse/controllers/share-and-invite.js
@@ -1,4 +1,4 @@
 import Controller from "@ember/controller";
 import ModalFunctionality from "discourse/mixins/modal-functionality";
 
-export default Controller.extend(ModalFunctionality, {});
+export default Controller.extend(ModalFunctionality);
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/controllers/share-and-invite.js
```diff
@@ -0,0 +1,4 @@
+import Controller from "@ember/controller";
+import ModalFunctionality from "discourse/mixins/modal-functionality";
+
+export default Controller.extend(ModalFunctionality, {});
```
eviltrout, line 4:
```suggestion
export default Controller.extend(ModalFunctionality);
```

app/serializers/invited_serializer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class InvitedSerializer < ApplicationSerializer
+  attributes :invites, :can_see_invite_details
+
+  def invites
```
eviltrout, line 6: Can you not use a `has_many` here?

nlalonde: I can't find a way to pass the `show_emails` option when using has_many.

PR #9369, opened 2020-04-07
FEATURE: Promote bookmarks with reminders to core functionality

martin-brennan:
#### IMPORTANT :mega: 

When merged, the new bookmarks with reminders will go live for everyone https://meta.discourse.org/t/improved-bookmarks-with-reminders/144542. Only merge if we are sure we want this to happen.

The main thrust of this PR is to take all the conditional checks based on the `enable_bookmarks_with_reminders` away and only keep the code from the `true` path, making bookmarks with reminders the core bookmarks feature. There is also a migration to create `Bookmark` records out of `PostAction` bookmarks for a site.

### Summary

* Remove logic based on whether enable_bookmarks_with_reminders is true. This site setting is now obsolete, the old bookmark functionality is being removed. Retain the setting and set the value to `true` in a migration.
* Use the code from the rake task to create a database migration that creates bookmarks from post actions.
* Change the bookmark report to read from the new table.
* Get rid of old endpoints for bookmarks
* Link to the new bookmarks list from the user summary page


app/controllers/topics_controller.rb
```diff
@@ -232,10 +232,10 @@ def posts
     fetch_topic_view(options)
 
     render_json_dump(TopicViewPostsSerializer.new(@topic_view,
-      scope: guardian,
-      root: false,
-      include_raw: !!params[:include_raw]
-    ))
+                                                  scope: guardian,
```
eviltrout, line 235: for the record I prefer the original format here. One problem with lining up is that if the original line changes in length you have to move all subsequent lines which pollutes the history. This is a bikeshed though.

db/migrate/20200409033412_create_bookmarks_from_post_action_bookmarks.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+class CreateBookmarksFromPostActionBookmarks < ActiveRecord::Migration[6.0]
+  def up
+    SiteSetting.enable_bookmarks_with_reminders = true
+
+    post_action_bookmarks = PostAction
```
eviltrout, line 7: Can you use manual SQL here? Using AR in migrations leads to many problems down the road.

coding-horror:
I think in the case of bookmarks the old feature was so useless that break
it and see who complains is the right choice.

app/controllers/topics_controller.rb
martin-brennan: Ah I think my vim autoformatter did this, I agree with you that the original was better

db/migrate/20200409033412_create_bookmarks_from_post_action_bookmarks.rb
martin-brennan: @eviltrout you are right, I wrote this as a rake task originally and moved it into a migration but didn't consider the AR issue. I will change this.

martin-brennan:
@coding-horror @eviltrout I will just remove the old API and see who complains as Jeff suggests. The new endpoints in `bookmarks_controller.rb` can be used in their place, if there is something special I need to do to make an endpoint work for API keys I will do that otherwise just leave it.

PR #9373, opened 2020-04-07
FEATURE: Show noscript view to unsupported browsers

udan11:


vendor/assets/javascripts/browser-update.js.erb
```diff
@@ -4,12 +4,18 @@
 (function() {
 
 var $buo = function() {
-
-  var badAndroid = false, ua = null;
+  if (window.msCrypto) {
```
davidtaylorhq, line 7: This will detect IE11, but I think we should try to make this work for other old browsers as well.

Can we do `if(!window.Promise)` instead? Just like we check for `window.WeakMap` below.

We also need to check that this works with polyfills. If the IE11 polyfills **are** loaded, window.unsupportedBrowser should be `false`.

xfalcox: > We also need to check that this works with polyfills. If the IE11 polyfills are loaded, window.unsupportedBrowser should be false.

Won't that mean we can't ever use newer CSS ?

davidtaylorhq: I mean for the next couple of months. I think we want to be able to merge this right now, then once we turn off the polyfills it will kick into action.

PR #9376, opened 2020-04-07
FEATURE: screenreader landmarks for main, suggested topics

riking:
Had screenreaders on my mind recently so did this while blocked on other stuff.

Testing info: https://www.scottohara.me/blog/2018/03/03/landmarks.html#using-screen-readers-to-navigate-landmarks

 - role="main" is nicer to implement than a "jump to content" href, because you can jump from anywhere.
 - Suggested topics is also an important jump point that you should to be able to reach from anywhere in the topic, so give it role="complementary".
   - In order to avoid a boatload of attributeBindings, I moved the root element of the suggested-topics component into the template. Also, autoformat their hbs files.
 - Flag the modal root with aria-modal=true to avoid the screenreader accidentally escaping the modal. https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/
     There's no need to ever toggle the attribute to false, because we display:none the modal root when it's closed. Except for maybe an iOS bug. I can't test iOS VO, so going to wait for complaints.

### DIFF
--- a/app/assets/javascripts/discourse/components/d-modal.js
+++ b/app/assets/javascripts/discourse/components/d-modal.js
@@ -10,7 +10,7 @@ export default Component.extend({
     "modalStyle",
     "hasPanels"
   ],
-  attributeBindings: ["data-keyboard"],
+  attributeBindings: ["data-keyboard", "aria-modal"],
   dismissable: true,
   title: null,
   subtitle: null,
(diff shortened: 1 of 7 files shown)


riking:
P.S. I have no clue how in the world we're going to write a screenreader integration test or how that test would execute.

riking:
oh right, the other reference doc I used for this: https://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page

riking:
I actually wasn't touching the topic list here, just topic view. I'm not sure how to get the jump to the topic list content while also covering every non-topiclist page.

PR #9378, opened 2020-04-07
FIX: Quoting posts

CvX:
Fixes to the quote feature. Most important changes listed below:

* FIX: Correctly attribute quotes when using Reply button
* FIX: Correctly attribute quotes when using replyAsNewTopic
* FIX: Allow quoting a quote
* FIX: Correctly mark quotes as "full"
* FIX: Don't try to create a quote if it's empty
* DEV: Remove an obsolete method `loadQuote`
  It isn't used in core anymore, the only use in core has been removed over 4 years ago in 3251bcb. It's not used in any plugins in all-the-plugins and all references to it on GitHub are from outdated forks (https://github.com/search?q=%22Post.loadQuote%22&type=Code)


CvX:
Nothing in `all-the-plugins` imports `discourse/lib/quote`, and there are no references to `Quote` object.

PR #9381, opened 2020-04-08
DEV: Refactor and test plugin addKeyboardShortcut

martin-brennan:
Refactor plugin-api `addKeyboardShortcut` to point to `KeyboardShortcuts`.
* Do not add shortcuts to the default object directly.
* Create an addShortcut function in keyboard-shortcuts to add shortcuts safely and call to bindKey to be able to use opts.
* Refactor controllers/bookmark.js to use new addShortcut func and emove unnecessary addBindings.
* No longer export keyboard shortcut bindings, rename to DEFAULT_BINDINGS and remove export, these do not need to be accessed by anything else.

**Note:** This has been tested locally with discourse-command-pallete and the bookmarks shortcuts to ensure no breakages.


jjaffeux:
It looks extremely good to me 👍  awesome work!

I have nothing to say, but requested @markvanlan to have  a look at this as he recently worked on this.

PR #9388, opened 2020-04-08
DEV: Drop 'key' column from user_api_keys

udan11:


davidtaylorhq:
Looks good to me, just a couple of test failures in CI to clean up

SamSaffron:
I am fine with this, merge at will

davidtaylorhq:
I think we're waiting on a theme-creator update which @udan11 is working on, then this will be good to merge

davidtaylorhq:
Are we good to merge this now @udan11 ?

PR #9391, opened 2020-04-08
FEATURE: New site setting `embed_unlisted`

eviltrout:
If enabled, posts imported to discourse via embeddings will default to
unlisted until they receive a reply.

### DIFF
--- a/lib/post_creator.rb
+++ b/lib/post_creator.rb
@@ -186,6 +186,7 @@ class PostCreator
         @post.link_post_uploads
         update_uploads_secure_status
         ensure_in_allowed_users if guardian.is_staff?
+        make_visible
         unarchive_message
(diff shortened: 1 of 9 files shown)


lib/post_creator.rb
```diff
@@ -414,6 +415,16 @@ def ensure_in_allowed_users
     end
   end
 
+  def make_visible
+    return unless SiteSetting.embed_unlisted?
+    return unless @post.post_number > 1
+    return if @post.topic.visible?
```
riking, line 421: Can you also assert that the hidden reason == imported_as_unlisted here?

eviltrout: I looked at that but the reason is attached to a post and means the first post is hidden, as opposed to the topic being unlisted.

PR #9392, opened 2020-04-08
FEATURE: Webhooks and Event for user being granted a badge

riking:
Adding a webhook for badge revocation is left for future work as it's relatively rare.

The webhook type is named 'user_badge' instead of 'badge' just in case we want to add hooks for modifying badge properties in the future, I guess...?

### DIFF
--- a/config/initializers/012-web_hook_events.rb
+++ b/config/initializers/012-web_hook_events.rb
@@ -74,6 +74,16 @@ end
   end
 end
 
+%i(
+  user_badge_granted
+).each do |event|
+  # user_badge_revoked
+  DiscourseEvent.on(event) do |badge, user_id|
+    ub = UserBadge.find_by(badge: badge, user_id: user_id)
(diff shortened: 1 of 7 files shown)


config/initializers/012-web_hook_events.rb
```diff
@@ -74,6 +74,16 @@
   end
 end
 
+%i(
+  user_badge_granted
+).each do |event|
+  # user_badge_revoked
+  DiscourseEvent.on(event) do |badge, user_id|
+    ub = UserBadge.find_by(badge: badge, user_id: user_id)
```
riking, line 82: I'm fairly sure this is something like a 4*N query count on badge grant jobs, but I'm not sure how to do any better. Moving the UserBadge.find_by into the DEvent.trigger doesn't exactly help, at all.

riking:
Plugin tests are hanging in "Create Database", odd...

riking:
Plugin specs passed locally, merging.

PR #9393, opened 2020-04-08
REFACTOR: Resize controls

romanrizzi:
The current implementation of this feature is too complex. I replaced most of the code by extending the image renderer function instead of using Tokens.

The new implementation has two parts: The first one is a rule that appends additional data to the IMG token. It adds an index that will let us identify the image we must resize and also the default scale of it (50%, 75%, 100%).

The second part extends the existing image renderer function. It wraps the original function with the resize controls HTML.

### DIFF
--- a/app/assets/javascripts/pretty-text/engines/discourse-markdown/resize-controls.js
+++ b/app/assets/javascripts/pretty-text/engines/discourse-markdown/resize-controls.js
@@ -1,3 +1,5 @@
+const SCALES = ["100", "75", "50"];
+
 function isUpload(token) {
   return token.content.includes("upload://");
 }
@@ -6,16 +8,8 @@ function hasMetadata(token) {
   return token.content.match(/(\d{1,4}x\d{1,4})/);
 }
 
-function buildToken(state, type, tag, klass, nesting) {
-  const token = new state.Token(type, tag, nesting);
-  token.block = true;
-  token.attrs = [["class", klass]];
-  return token;
-}
-
-function wrapImage(tokens, index, state, imgNumber) {
-  const imgToken = tokens[index];
-  const sizePart = imgToken.content
+function appendMetaData(index, token) {
+  const sizePart = token.content
     .split("|")
     .find(x => x.match(/\d{1,4}x\d{1,4}(,\s*\d{1,3}%)?/));
   let selectedScale =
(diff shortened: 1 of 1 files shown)


eviltrout:
I'm not super familiar with this pipeline so I've requested @SamSaffron do a review here.

jjaffeux:
I think @udan11 did the previous refactoring of this, he should take a look

app/assets/javascripts/pretty-text/engines/discourse-markdown/resize-controls.js
```diff
@@ -6,131 +6,63 @@ function hasMetadata(token) {
...
-      newElements.push(separatorToken);
-      newElements.push(buildToken(state, "separator_close", "span", "", -1));
-    }
-  });
-  newElements.push(buildToken(state, "wrap_button_close", "span", "", -1));
+      .trim()
+      .replace("%", "");
 
-  newElements.push(buildToken(state, "wrap_image_close", "span", "", -1));
+  const scales = ["100", "75", "50"];
```
udan11, line 21: Nitpick: This should be a proper const which can be reused in  `rule` too.

app/assets/javascripts/pretty-text/engines/discourse-markdown/resize-controls.js
```diff
@@ -147,6 +79,36 @@ export function setup(helper) {
...
+        if (scaleIndex !== -1) {
+          var scale = token.attrs[scaleIndex][1];
+          var index = token.attrs[imageIndex][1];
+
+          let result = "<span class='image-wrapper'>";
+          result += oldRule(tokens, idx, options, env, slf);
+          result +=
+            "<span class='button-wrapper' data-image-index='" + index + "'>";
+
+          result += buildScaleButton(scale, "100");
```
udan11, line 98: Then, using the constant above defined per the comment above you could replace this block with:

```suggestion
          result += PREDEFINED_SCALES
            .map(s => buildScaleButton(scale, s))
            .join("<span class='separator'>&nbsp;•&nbsp;</span>");
```

`PREDEFINED_SCALES` and `s` are not the best names, but I cannot think of better ones (maybe `s` could be `scale` and rename outer scope `scale` to `selectedScale`). :thinking:

PR #9394, opened 2020-04-08
DEV: Buff theme install rake task - more options

markvanlan:
This rake task was created from someone outside Discourse, and does not quite fit our needs. This PR keeps the existing functionality, while extending it.

Instead of "skipping" previously installed themes, call `update_from_remote` on them. Calling this rake task systematically, will install new themes and keep themes/components updated.

Add option to install a theme component to all existing themes: `add_to_all_themes`. Even if a theme component is previously installed, so it gets updated, the `add_to_all_themes` option will be in effect.

Also add ability to pass a JSON string into `rake themes:install`
```bash
bin/rake themes:install -- '--{"placeholder": {"url": "https://github.com/discourse/discourse-placeholder-theme-component.git", "install_to_all_themes": true}}'
```

This diff sucks - looks kind of tough to review.

### DIFF
--- a/app/services/themes_install_task.rb
+++ b/app/services/themes_install_task.rb
@@ -6,6 +6,7 @@ class ThemesInstallTask
     log = []
     themes.each do |name, val|
       installer = new(val)
+      next if installer.url.nil?
 
       if installer.theme_exists?
         installer.update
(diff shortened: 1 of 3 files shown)


markvanlan:
Not ready for review.

PR #9397, opened 2020-04-09
FEATURE: add after-reviewable-post-user plugin outlet

featheredtoast:
Add a plugin outlet after reviewable post user

Additionally adds custom field user serializer

Add a basic user serializer that includes custom fields.
Allows review queue serializer to include custom fields for its users

### DIFF
--- a/app/serializers/user_custom_fields_serializer.rb
+++ b/app/serializers/user_custom_fields_serializer.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# A basic user serializer, with custom fields
+class UserCustomFieldsSerializer < BasicUserSerializer
+  attributes :custom_fields
+
+  def custom_fields
+    fields = custom_field_keys
+
+    if fields.present?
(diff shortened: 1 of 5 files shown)


featheredtoast:
I'll merge this tomorrow - this is to allow staff notes to appear in the review queue like it did in the old flag queue.

app/serializers/user_custom_fields_serializer.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# A basic user serializer, with custom fields
+class UserCustomFieldsSerializer < BasicUserSerializer
```
eviltrout, line 4: One thing that's confusing here is we already have a model called `UserCustomField` and that means this would be the default serializer for that model.

Could we use a name like `UserWithCustomFields` instead to differentiate?

featheredtoast: Oh, that's a good one - can do! Naming is hard.

PR #9398, opened 2020-04-09
FEATURE: Plugin support for transpiling regular `.js` files

eviltrout:
This adds support for a new piece of metadata to your plugin.rb
files. If you add:

```
transpile_js: true
```

Then Discourse will support transpilation of assets in your
`assets/javascripts` directory. Previously they had to be named
`.js.es6` but now regular `.js` will work.

Note this is opt-in because some plugins currently have `.js` files in
app/assets that are not meant to be transpiled.

Going forward all plugins should migrate to this setting as they are
comfortable able to do so.

### DIFF
--- a/lib/discourse_js_processor.rb
+++ b/lib/discourse_js_processor.rb
@@ -4,6 +4,10 @@ require 'mini_racer'
 
 class DiscourseJsProcessor
 
+  def self.plugin_transpile_paths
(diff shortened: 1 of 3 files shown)


eviltrout:
> Some specs are failing because `Plugin` instances created in those specs don't have metadata and `#transpile_js`/`#activate!` require that now.

Thanks! I've fixed it.

PR #9413, opened 2020-04-13
FIX: Keyboard navigation fixes in setup wizard

riking:
- Delete a positive tabindex from a reused component
 - Copy :hover styles to :focus
 - Replace an 'outline: 0' rule with a TODO for a custom :focus style

Discovered while fixing the no-positive-tabindex lint. Needed to test wizard w/ keyboard to reach the tabindex deletion, and encountered these other blockers.


jjaffeux:
Fixed merge conflicts due to double quotes

PR #9414, opened 2020-04-13
DEV: Fix and enable the 'require-button-type' lint 

riking:
Pretty simple. One button got type=submit instead of type=button.


jjaffeux:
Fixed merge conflicts due to double quotes

PR #9415, opened 2020-04-13
DEV: Fix all 'require-valid-alt-text' lints except in reused compoments

riking:
Three violations require parameters from the caller to generate valid alt text.

### DIFF
--- a/app/assets/javascripts/discourse/templates/components/group-flair-inputs.hbs
+++ b/app/assets/javascripts/discourse/templates/components/group-flair-inputs.hbs
@@ -35,13 +35,13 @@
 
   <div class="avatar-flair-preview">
     <div class="avatar-wrapper">
-      <img width="45" height="45" src={{demoAvatarUrl}} class="avatar actor">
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/discourse/templates/components/group-flair-inputs.hbs
```diff
@@ -35,13 +35,13 @@
 
   <div class="avatar-flair-preview">
     <div class="avatar-wrapper">
-      <img width="45" height="45" src={{demoAvatarUrl}} class="avatar actor">
+      <img width="45" height="45" src={{demoAvatarUrl}} class="avatar actor" alt="" role="presentation">
```
eviltrout, line 38: Should we add a value here? "sample avatar"

riking: I did this because there is no alt text on regular post avatars.

riking:
Checks look bugged, merging w/ override.

PR #9419, opened 2020-04-14
FIX: Extend bookmark cutoff time for "Later Today" to 5pm

martin-brennan:
As per discussions at https://meta.discourse.org/t/improved-bookmarks-with-reminders/144542/88?u=mjrbrennan

* users can now use the Later Today option for bookmark reminders up until 5PM
* the time for later today maxes out at 6pm, so any time it is used after 3pm it is maxed to 6pm
* round to hour instead of half-hour for Later Today as well


SamSaffron:
cool, lets give this a shot!

PR #9421, opened 2020-04-14
FIX: allows custom groups updates to be reflected without recompilation

jjaffeux:


jjaffeux:
@eviltrout few notes on this:
- it's been broken for quite sometimes, but given people don't add custom emojis everyday, it was not a big issue
- however with custom groups, that might become more popular
- I dropped the RAW_TEMPLATE, but kept the precompiled html template from erb which is where the biggest perf gain is
- tested on android device works fine
- I  took the opportunity to add `loading="lazy"` as it's essentially free win when supported, and it is on edge/firefox/chrome, it's only for custom emojis, but I might apply this to the whole list later

PR #9422, opened 2020-04-14
DEV: pass widget back when cleaning up stream

jjaffeux:


jjaffeux:
@eviltrout would you be ok with this? I think it will make easier to clean state in some cases.

PR #9428, opened 2020-04-15
DEV: attempts to fix various leaks

jjaffeux:


test/javascripts/test_helper.js
```diff
@@ -105,6 +105,8 @@ function resetSite(siteSettings, extras) {
   Site.resetCurrent(Site.create(siteAttrs));
 }
 
+let baseTransalations = I18n.translations[I18n.locale].js || {};
```
eviltrout, line 108: This isn't copying them right? It's a reference to that object. So if that object is mutated setting it back does nothing?

jjaffeux: Oh yes I changes the impelmentation of this at the end and didnt check but yes you are right I think. This one doesn’t matter that much anyways, it mosty prevents to carry locales over time, will come up with something better.

jjaffeux:
@eviltrout I have removed the I18NN change because it's not needed, and also I changed every scheduleOnce with anon function to schedule because this is broken. A scheduleOnce with an anonymous function will work as a simple schedule as there's no way it can know it's the same thing.

danielwaterworth:
> A scheduleOnce with an anonymous function will work as a simple schedule as there's no way it can know it's the same thing.

Aside from doing toString on it :trollface:

PR #9430, opened 2020-04-15
UX: drop the `automatic_membership_retroactive` column from groups model.

vinothkannans:
Instead, display the automatic membership user count details in a bootbox alert.


app/assets/javascripts/discourse/controllers/groups-new.js
```diff
@@ -1,18 +1,54 @@
...
+import { ajax } from "discourse/lib/ajax";
 import { popupAjaxError } from "discourse/lib/ajax-error";
 
+export function popupAutomaticMembershipAlert(group_id, email_domains) {
+  if (!email_domains) {
+    return;
+  }
+
+  const data = {};
+  data["automatic_membership_email_domains"] = email_domains;
```
eviltrout, line 12:
```suggestion
  data.automatic_membership_email_domains = email_domains;
```

app/assets/javascripts/discourse/controllers/groups-new.js
```diff
@@ -1,18 +1,54 @@
...
+
+  if (group_id) {
+    data["id"] = group_id;
```
eviltrout, line 15:
```suggestion
    data.id = group_id;
```

app/models/group.rb
```diff
@@ -842,6 +840,32 @@ def update_primary_group
     end
   end
 
+  def self.automatic_membership_users(domains, group_id = nil)
+    domains = domains.gsub('.', '\.')
+
+    users = User.joins(:user_emails).where("user_emails.email ~* '@(#{domains})$'").activated.where(staged: false)
```
eviltrout, line 846: I don't like how domains is inserted here as it looks like an opportunity for SQL injection. Can you change it to be parameter instead?

(I realize the old code did this too, but this is an opportunity to harden our app.)

spec/models/group_spec.rb
```diff
@@ -407,6 +407,7 @@ def real_staff
     Group.refresh_automatic_groups!
 
     groups = Group.includes(:users).to_a
+    puts groups.pluck(:name).inspect
```
eviltrout, line 410: Looks like you left in debugging by accident :)

PR #9434, opened 2020-04-15
DEV: Fix no-implicit-this for all user-* components

riking:


riking:
Yep, I made sure to heavily inspect the first few changes I made, and verified the user card functionality that I could. There's a lot in there!

PR #9437, opened 2020-04-16
FEATURE: Allow editing bookmark reminders

martin-brennan:
Users can now edit the bookmark name and reminder time from their list of bookmarks:

image

image

We use "Custom" for the date and time in the modal because if the user set a reminder for "tomorrow" then edit the reminder "tomorrow", the definition of what "tomorrow" is has changed.


app/assets/javascripts/discourse/controllers/bookmark.js
```diff
@@ -73,12 +75,34 @@ export default Controller.extend(ModalFunctionality, {
...
+    }
+  },
+
+  parseExistingBookmarkReminderAt() {
+    if (this.editingExistingBookmark() && this.existingBookmarkHasReminder()) {
+      let parsedReminderAt = moment.tz(
+        this.model.reminderAt,
+        this.userTimezone
+      );
+      this.set("customReminderDate", parsedReminderAt.format("YYYY-MM-DD"));
```
jjaffeux, line 95: group with setProperties

app/assets/javascripts/discourse/controllers/bookmark.js
```diff
@@ -136,6 +160,11 @@ export default Controller.extend(ModalFunctionality, {
 
   showBookmarkReminderControls: true,
 
+  @discourseComputed("model.reminderAt")
+  showExistingReminderAt(existingReminderAt) {
+    return existingReminderAt != null;
```
jjaffeux, line 165:
```suggestion
    return existingReminderAt !== null;
```

app/assets/javascripts/discourse/controllers/user-activity-bookmarks-with-reminders.js
```diff
@@ -63,6 +64,24 @@ export default Controller.extend({
...
+          postId: bookmark.post_id,
+          id: bookmark.id,
+          reminderAt: bookmark.reminder_at,
+          name: bookmark.name
+        },
+        title: "post.bookmarks.edit",
+        modalClass: "bookmark-with-reminder"
+      });
+      controller.setProperties({
+        afterSave: () => {
```
jjaffeux, line 79: one line

app/assets/javascripts/discourse/templates/modal/bookmark.hbs
```diff
@@ -9,9 +9,15 @@
     {{/if}}
 
     <div class="control-group">
-      {{input value=name name="name" class="bookmark-name" placeholder=(i18n "post.bookmarks.name_placeholder")}}
+      {{input value=model.name name="name" class="bookmark-name" placeholder=(i18n "post.bookmarks.name_placeholder")}}
     </div>
 
+    {{#if showExistingReminderAt }}
+      <div class="alert alert-info existing-reminder-at-alert">
+        {{d-icon "far-clock"}} {{i18n "bookmarks.reminders.existing_reminder"}} {{existingReminderAtFormatted}}
```
jjaffeux, line 17: indentation looks weird here

app/assets/javascripts/discourse/templates/user/bookmarks.hbs
```diff
@@ -42,7 +42,10 @@
               <td>{{format-date bookmark.created_at format="tiny"}}</td>
               {{raw "list/activity-column" topic=bookmark class="num" tagName="td"}}
               <td>
-                {{bookmark-actions-dropdown bookmark=bookmark removeBookmark=(action "removeBookmark")}}
+                {{bookmark-actions-dropdown
```
jjaffeux, line 45: {{bookmark-actions-dropdown
  bookmark=bookmark
  removeBookmark=(action "removeBookmark")
  editBookmark=(action "editBookmark")
}}

app/assets/javascripts/discourse/controllers/bookmark.js
```diff
@@ -73,12 +75,34 @@ export default Controller.extend(ModalFunctionality, {
       userTimezone: this.currentUser.resolvedTimezone()
     });
 
+    this.parseExistingBookmarkReminderAt();
```
jjaffeux, line 78: There's a lot of methods with a a kinda unexpected naming, let me explain my reasoning.

When you say: `this.parseExistingBookmarkReminderAt();` I expect to actually parse something as a param, here the parsing is actually happening inside the function through some state and mutates another state.

I feel like a better separation of public/private APIs should be though here, as it currently makes the code a little bit hard to follow in my opinion.

martin-brennan: That makes sense, keeping mutations like this as pure functions. I had a moment of doubt yesterday when I made and named it too but just brushed past it. Will make this more clear :+1:

PR #9439, opened 2020-04-16
FIX: flaky groups_controller_spec

lis2:
Sometimes spec which is testing order groups by user count is failing - line 141

My theory is that cause is the randomness of Postgres when the order value is the same for 2 rows.

In spec, we got three groups
`moderator_group` - 0 users
`group` - 1 user
`other_group` - 1 user

And we are expecting that controller will return them in ascending order [moderator, group, other_group]

Because `group` and `other_group` contain the same amount of users, we are dealing with luck

Therefore, I believe that adding one more user to other_group should make that query reliable.

After adding a user to other_group, I needed to change the order in another spec which was testing DESC order of groups - line 120

It was not crashing on my local machine, so I am not 100% sure.

### DIFF
--- a/spec/requests/groups_controller_spec.rb
+++ b/spec/requests/groups_controller_spec.rb
@@ -4,6 +4,7 @@ require 'rails_helper'
 
 describe GroupsController do
   fab!(:user) { Fabricate(:user) }
+  let(:other_user) { Fabricate(:user) }
   let(:group) { Fabricate(:group, users: [user]) }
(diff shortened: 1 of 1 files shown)


CvX:
> My theory is that cause is the randomness of Postgres when the order value is the same for 2 rows.

If that's the case, would it be better add a secondary `ORDER BY` clause in the query that gives those results? It's not the spec that exhibits occasionally random behavior but the underlying code. 🙂 The same could happen in a live application – it would flip-flop between two different responses to the same request even if the data hasn't changed.

PR #9451, opened 2020-04-16
FEATURE: Add copy button to codeblocks

justindirose:
This PR converts the copy codeblocks theme component to a core feature. The codeblocks are only shown when the `show copy button on codeblocks` site setting is enabled. This is defaulted to off.

Additionally, I've also fixed an issue where codeblocks inside of GitHub oneboxes have a copy button added. This uses the `:scope` selector, which is not compatible in IE 11. I did find a [possible polyfill](https://github.com/jonathantneal/element-qsa-scope/) we could use but my knowledge is limited there.

### DIFF
--- a/app/assets/javascripts/discourse/initializers/copy-codeblocks.js
+++ b/app/assets/javascripts/discourse/initializers/copy-codeblocks.js
@@ -54,8 +54,8 @@ function clipboardCopy(text) {
       );
 }
 
-let _clickHandlerElement = null;
-let _runLater = null;
+let _copyCodeblocksClickHandlers = {};
+let _fadeCopyCodeblocksRunners = {};
 
 export default {
   name: "copy-codeblocks",
@@ -63,14 +63,16 @@ export default {
   initialize(container) {
     withPluginApi("0.8.7", api => {
       function _cleanUp() {
-        if (_clickHandlerElement) {
-          _clickHandlerElement.removeEventListener("click", _handleClick);
-          _clickHandlerElement = null;
-        }
-        if (_runLater) {
-          cancel(_runLater);
-          _runLater = null;
-        }
+        Object.values(_copyCodeblocksClickHandlers || {}).forEach(handler =>
+          handler.removeEventListener("click", _handleClick)
+        );
+
+        Object.values(_fadeCopyCodeblocksRunners || {}).forEach(runner =>
+          cancel(runner)
+        );
+
+        _copyCodeblocksClickHandlers = {};
+        _fadeCopyCodeblocksRunners = {};
       }
 
       function _handleClick(event) {
@@ -82,39 +84,58 @@ export default {
         const code = button.nextSibling;
 
         if (code) {
-          clipboardCopy(code.innerText.trim());
+          clipboardCopy(code.innerText.trim()).then(() => {
+            button.classList.add("copied");
 
-          button.classList.add("copied");
+            const commandId = button.dataset.commandId;
 
-          _runLater = later(() => button.classList.remove("copied"), 3000);
+            if (_fadeCopyCodeblocksRunners[commandId]) {
+              cancel(_fadeCopyCodeblocksRunners[commandId]);
+              delete _fadeCopyCodeblocksRunners[commandId];
+            }
+
+            _fadeCopyCodeblocksRunners[commandId] = later(
+              () => button.classList.remove("copied"),
+              3000
+            );
+          });
         }
       }
 
-      function _attachCommands(postElements) {
+      function _attachCommands(postElements, helper) {
+        if (!helper) {
+          return;
+        }
+
         const siteSettings = container.lookup("site-settings:main");
         const { isIE11 } = container.lookup("capabilities:main");
         if (!siteSettings.show_copy_button_on_codeblocks || isIE11) {
           return;
         }
+
         const commands = postElements[0].querySelectorAll(
           ":scope > pre > code"
         );
 
-        if (!commands.length) {
+        const post = helper.getModel();
+
+        if (!commands.length || !post) {
           return;
         }
 
-        _clickHandlerElement = postElements[0];
+        const postElement = postElements[0];
+        _copyCodeblocksClickHandlers[post.id] = postElement;
 
-        commands.forEach(command => {
+        commands.forEach((command, commandIndex) => {
           const button = document.createElement("button");
           button.classList.add("btn", "nohighlight", "copy-cmd");
           button.innerHTML = iconHTML("copy");
+          button.dataset.commandId = `${post.id}-${commandIndex}`;
           command.before(button);
           command.parentElement.classList.add("copy-codeblocks");
         });
 
-        _clickHandlerElement.addEventListener("click", _handleClick, false);
+        postElement.addEventListener("click", _handleClick, false);
       }
 
       api.decorateCooked(_attachCommands, { id: "copy-codeblocks" });
--- a/app/assets/javascripts/discourse/initializers/copy-codeblocks.js
+++ b/app/assets/javascripts/discourse/initializers/copy-codeblocks.js
@@ -82,12 +82,7 @@ export default {
         const code = button.nextSibling;
 
         if (code) {
-          let string = code.innerText;
-
-          if (string) {
-            string = string.trim();
-            clipboardCopy(string);
-          }
+          clipboardCopy(code.innerText.trim());
 
           button.classList.add("copied");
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,121 @@
...
+      throw err !== undefined
+        ? err
+        : new DOMException("The request is not allowed", "NotAllowedError");
+    });
+  }
+
+  // ...Otherwise, use document.execCommand() fallback
+
+  // Put the text to copy into a <span>
+  var span = document.createElement("span");
```
jjaffeux, line 21: no var, use let/const

jjaffeux, line 21: no var, use let/const, I know it's third party code, but we can't have this in the codebase.

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,121 @@
...
+
+          button.classList.add("copied");
+
+          later(() => button.classList.remove("copied"), 3000);
+        }
+      }
+
+      function _attachCommands($elem) {
+        const siteSettings = container.lookup("site-settings:main");
+        if (siteSettings.show_copy_button_on_codeblocks) {
```
jjaffeux, line 95: it's usually better to do:

```
if (!siteSettings.show_copy_button_on_codeblocks) {
  return;
}
```

It avoids adding one un-necessary indentation level

jjaffeux:
I don't think we want to be bothered by IE here, however, please test how bad it fails on IE please, if it just doesn't show we don't care.

Maybe add this in the site setting description, eg: not supported on IE.

jjaffeux:
Now that I think about it, as it's a core PR? if you want you can add https://github.com/jonathantneal/element-qsa-scope/blob/master/index.js in the discourse-internet-explorer ie.js file

justindirose:
@jjaffeux fixes should be wrapped up in the last push. Removed IE 11 support since the button was rendering without polyfill but failed to work in general.

config/locales/server.en.yml
```diff
@@ -2064,6 +2064,7 @@ en:
     warn_reviving_old_topic_age: "When someone starts replying to a topic where the last reply is older than this many days, a warning will be displayed. Disable by setting to 0."
     autohighlight_all_code: "Force apply code highlighting to all preformatted code blocks even when they didn't explicitly specify the language."
     highlighted_languages: "Included syntax highlighting rules. (Warning: including too many languages may impact performance) see: <a href='https://highlightjs.org/static/demo/' target='_blank'>https://highlightjs.org/static/demo</a> for a demo"
+    show_copy_button_on_codeblocks: "Add a button to codeblocks to copy the block contents to the user's clipboard. This feature is not supported on Internet Explorer 11."
```
jjaffeux, line 2067:
```suggestion
    show_copy_button_on_codeblocks: "Add a button to codeblocks to copy the block contents to the user's clipboard. This feature is not supported on Internet Explorer."
```

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,123 @@
...
+          let string = code.innerText;
+
+          if (string) {
+            string = string.replace(/^\s+|\s+$/g, "");
+            clipboardCopy(string);
+          }
+
+          button.classList.add("copied");
+
+          later(() => button.classList.remove("copied"), 3000);
```
jjaffeux, line 89: I think we should hold a reference to later and cancel it in cleanup http://api.emberjs.com/ember/3.14/functions/@ember%2Frunloop/cancel

Let me know if you need help on this.

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        const button = event.target;
+        const code = button.nextSibling;
+
+        if (code) {
+          let string = code.innerText;
+
+          if (string) {
+            string = string.replace(/^\s+|\s+$/g, "");
```
jjaffeux, line 88: any reason we don't use trim() here ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,128 @@
...
+
+          button.classList.add("copied");
+
+          _runLater = later(() => button.classList.remove("copied"), 3000);
+        }
+      }
+
+      function _attachCommands($elem) {
+        const siteSettings = container.lookup("site-settings:main");
+        const { isIE11 } = api.container.lookup("capabilities:main");
```
jjaffeux, line 100:
```suggestion
        const { isIE11 } = container.lookup("capabilities:main");
```

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,128 @@
...
+        if (!siteSettings.show_copy_button_on_codeblocks || isIE11) {
+          return;
+        }
+        const commands = $elem[0].querySelectorAll(":scope > pre > code");
```
jjaffeux, line 104: give $elem[0] a proper name so we can reuse it after

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,128 @@
...
+      function _handleClick(event) {
+        if (!event.target.classList.contains("copy-cmd")) {
+          return;
+        }
+
+        const button = event.target;
+        const code = button.nextSibling;
+
+        if (code) {
+          let string = code.innerText;
```
jjaffeux, line 85: now that I think about it, I think we can just all have in one line:

```
clipboardCopy(code.innerText.trim());
```

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
justindirose: I think I took care of this in the latest commit. Have a look to make sure I understood correctly.

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+        const commands = postElements[0].querySelectorAll(
+          ":scope > pre > code"
+        );
+
+        if (!commands.length) {
+          return;
+        }
+
+        _clickHandlerElement = postElements[0];
```
eviltrout, line 107: I might be missing something here, but it seems to me like `_clickHandlerElement` could be overwritten here.

What if two posts have commands? The first would capture the element, and the second would overwrite it, which means it cannot be cleaned up.

To confirm this you could add an assertion here that throws an error if `_clickHandlerElement` is not null.

jjaffeux: Oh yes!

Thanks robin, totally forgot this, this is for this use case I made this commit the other day: https://github.com/discourse/discourse/commit/6023ea197909c42f4fb4c21ff1dcda3be90a161f

I think it will allow to cleanly remove create event listeners for each post.

justindirose: Added a patch thanks to @jjaffeux to clean this up.

justindirose:
@jjaffeux @eviltrout are we good to merge this now?

app/assets/javascripts/discourse/initializers/copy-codeblocks.js
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        const postElement = postElements[0];
+        _copyCodeblocksClickHandlers[post.id] = postElement;
+
+        commands.forEach((command, commandIndex) => {
+          const button = document.createElement("button");
+          button.classList.add("btn", "nohighlight", "copy-cmd");
+          button.innerHTML = iconHTML("copy");
+          button.dataset.commandId = `${post.id}-${commandIndex}`;
```
eviltrout, line 133: I hope that in the case when a post is edited to add/remove a comand we call `cleanUp` before decorating again because these would stamp on the other `id`s. Something to test!

jjaffeux: OFC we don't :p good catch, will think about it...

justindirose: Most recent commit (patch from @jjaffeux) should address this.

justindirose:
I will plan to merge tomorrow morning barring any objections!

Chaboi45:
This feature helped a bunch! Thanks :)

PR #9454, opened 2020-04-17
FEATURE: Edit bookmark reminders from post and explicit delete button

martin-brennan:
There is now an explicit "Delete Bookmark" button in the edit modal:

image

A confirmation is shown before deleting

image

Along with this, when the bookmarked post icon is clicked the modal is now shown instead of just deleting the bookmark. Also, the "Delete Bookmark" button from the user bookmark list now confirms the action.

Add a `d d` shortcut in the modal to delete the bookmark.


coding-horror:
I don’t think you need the word “bookmark” in the delete button? Then the
copy is not unique and easier to localize. Heck maybe just use the glyph...

PR #9455, opened 2020-04-17
FEATURE: Edit bookmark reminders from post and explicit delete button

martin-brennan:
There is now an explicit "Delete Bookmark" button in the edit modal:

image

A confirmation is shown before deleting

image

Along with this, when the bookmarked post icon is clicked the modal is now shown instead of just deleting the bookmark. Also, the "Delete Bookmark" button from the user bookmark list now confirms the action.

Add a `d d` shortcut in the modal to delete the bookmark.


martin-brennan:
@coding-horror moved your comment here, I had to close that other PR because weird stuff was happening

> I don’t think you need the word “bookmark” in the delete button? Then the
copy is not unique and easier to localize. Heck maybe just use the glyph...

Will do!

app/assets/javascripts/discourse/controllers/bookmark.js
```diff
@@ -413,6 +439,22 @@ export default Controller.extend(ModalFunctionality, {
         .finally(() => (this.saving = false));
     },
 
+    delete() {
+      this.deleting = true;
+      bootbox.confirm(I18n.t("bookmarks.confirm_delete"), result => {
+        if (result) {
+          this.closeWithoutSaving = true;
```
eviltrout, line 446: I know this variable didn't start in this commit but as a general note - if you are only using a variable in a particular controller, and nothing is computed based on it, we prefer to underscore them as private. For example `this._closeWithoutSaving`. 

Absolutely a minor thing, though and just something to keep in mind for the future.

martin-brennan: @eviltrout cool I was talking to Joffrey about this as well RE: functions because I wasn't sure, so I will make sure to do it from now on :)

PR #9456, opened 2020-04-17
FIX: when category or tag is muted, update user

lis2:
Currently, when category or tag is muted, only after the hard refresh, these new muted categories are really muted. Without a hard refresh, you will still receive "new topic" messages.

Therefore, when tag or category is muted, we should update the user object right away.

### DIFF
--- a/app/assets/javascripts/discourse/controllers/tags-show.js
+++ b/app/assets/javascripts/discourse/controllers/tags-show.js
@@ -155,7 +155,18 @@ export default Controller.extend(BulkTopicSelection, FilterModeMixin, {
     },
 
     changeTagNotificationLevel(notificationLevel) {
-      this.tagNotification.update({ notification_level: notificationLevel });
+      this.tagNotification
+        .update({ notification_level: notificationLevel })
+        .then(response => {
+          this.currentUser.set(
+            "muted_tag_ids",
+            this.currentUser.calculateMutedIds(
+              notificationLevel,
+              response.responseJson.tag_id,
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/discourse/controllers/tags-show.js
```diff
@@ -155,7 +156,18 @@ export default Controller.extend(BulkTopicSelection, FilterModeMixin, {
     },
 
     changeTagNotificationLevel(notificationLevel) {
-      this.tagNotification.update({ notification_level: notificationLevel });
+      this.tagNotification
+        .update({ notification_level: notificationLevel })
+        .then(response => {
+          User.currentProp(
```
eviltrout, line 162: Because this is a controller I would prefer if you used `this.currentUser` to get the user rather than the `User` singleton. Saves an import and makes testing easier due to dependency injection.

app/assets/javascripts/discourse/models/user.js
```diff
@@ -859,6 +860,15 @@ const User = RestModel.extend({
 
   changeTimezone(tz) {
     this._timezone = tz;
+  },
+
+  calculateMutedIds(notificationLevel, id, type) {
+    const muted_ids = User.currentProp(type);
```
eviltrout, line 866: Is there a case where this wouldn't be `this`? Not clear why you'd need to use the singleton here.

lis2:
@eviltrout thank you for your comments, you were right, fixed :)

PR #9458, opened 2020-04-17
FIX: Do not raise an error if the post action type is nil

romanrizzi:


spec/requests/post_action_users_controller_spec.rb
```diff
@@ -79,4 +79,19 @@
 
     expect(total).to eq(5)
   end
+
+  it 'returns no users when the action type id is invalid' do
+    get "/post_action_users.json", params: {
+      id: post.id, post_action_type_id: "ThZ3m3Dx');select%20pg_sleep(9);%20--%20"
```
riking, line 85: Just use a generic invalid string here.

spec/requests/post_action_users_controller_spec.rb
```diff
@@ -79,4 +79,19 @@
...
     expect(total).to eq(5)
   end
+
+  it 'returns no users when the action type id is invalid' do
+    get "/post_action_users.json", params: {
+      id: post.id, post_action_type_id: "invalid_action_type"
+    }
+
+    expect(response.status).to eq(200)
+    json = JSON.parse(response.body)
```
eviltrout, line 89: There's now a cool `response.parsed_body` so we don't have to do this anymore :)

PR #9459, opened 2020-04-17
improvements to the SSO fields

mentalstring:
- add `location` as a SSO field, with corresponding `sso_overrides_location` setting to allow override of local value
- add `sso_overrides_website` setting that allows to prevent local changes

### DIFF
--- a/spec/services/user_updater_spec.rb
+++ b/spec/services/user_updater_spec.rb
@@ -249,6 +249,22 @@ describe UserUpdater do
       end
     end
 
+    context 'when sso overrides location' do
+      it 'does not change location' do
+        SiteSetting.sso_url = "https://www.example.com/sso"
+        SiteSetting.enable_sso = true
+        SiteSetting.sso_overrides_location = true
+
+        user = Fabricate(:user)
+        updater = UserUpdater.new(acting_user, user)
+
+        expect(updater.update(location: "new location")).to be_truthy
+
+        user.reload
+        expect(user.user_profile.location).not_to eq 'new location'
+      end
+    end
+
     context 'when updating primary group' do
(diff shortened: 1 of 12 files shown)


eviltrout:
Looks like some tests are failing here. Can you fix those please?

mentalstring:
Fixed one, but not sure what's the best way to address the other. Seems the test doesn't find the location field that is now dependent on the `sso_override_location` setting (defaults to false). Quite possible that I'm doing something wrong as I'm new to Discourse/Ruby. Could use some help?

eviltrout:
It looks to me like you have to add the `can_edit_location` that you added to the user serializer to the qunit test fixtures to pass. My advice would be to run the qunit tests locally, and find the failure. Then you can re-run that one and figure out why the field is not showing up.

mentalstring:
Thanks for the tip — seems it passes now. Please double-check the code as this is a new stack for me.

SamSaffron:
Looks generally fine to me, but given it is a protocol level change  can we add bit more testing to cover both on and off cases servers side?

spec/services/user_updater_spec.rb
```diff
@@ -249,6 +249,38 @@
...
+        user = Fabricate(:user)
+        updater = UserUpdater.new(acting_user, user)
+
+        expect(updater.update(website: "https://google.com")).to be_truthy
+
+        user.reload
+        expect(user.user_profile.website).not_to eq 'https://google.com'
+      end
+    end
+
```
mentalstring, line 283: Something like this? Just following along what is done for the `sso_overrides_bio`.

SamSaffron:
This is looking good to me, the change looks safe! 

Thanks @mentalstring

PR #9461, opened 2020-04-17
FIX: Set user timezone on password reset login

markvanlan:


eviltrout:
Can you provide some context for this change please?

markvanlan:
> Can you provide some context for this change please?

When a new clicks an invite link, it takes them to the password reset page to set their first password.

They are then logged in. That user's timezone was not set in the process, but while logging in normally, the timezone is set. This follows the same pattern, as it sets the user timezone when they log in, if the timezone was previously unset.

PR #9463, opened 2020-04-17
FEATURE: Add created_at column to user_badges

riking:
This makes scraping for newly granted badges easier.

### DIFF
--- a/app/services/badge_granter.rb
+++ b/app/services/badge_granter.rb
@@ -16,7 +16,7 @@ class BadgeGranter
     return unless badge.enabled?
 
     system_user_id = Discourse.system_user.id
-    now = Time.now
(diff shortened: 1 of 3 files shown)


app/services/badge_granter.rb
```diff
@@ -16,7 +16,8 @@ def self.mass_grant(badge, users)
     return unless badge.enabled?
 
     system_user_id = Discourse.system_user.id
-    user_badges = users.map { |u| { badge_id: badge.id, user_id: u.id, granted_by_id: system_user_id, granted_at: Time.now } }
+    now = Time.now
```
eviltrout, line 19:
```suggestion
    now = Time.zone.now
```

PR #9475, opened 2020-04-20
DEV: Add class to header-topic-info widget

markvanlan:
There is an `.extra-info-wrapper` holding the `.extra-info` div. Right now, a `.two-rows` class gets added to `.extra-info` when viewing a topic with a category showing. If the topic is uncategorized, and suppress uncategorized is true, the `.two-rows` class is not present.

In **some plugins** we decorate this widget like so: `header-topic-info:after`. The problem here is that we need the `two-rows` class to be on the PARENT of `header-topic-info` so that we can use it as a css selector.

### DIFF
--- a/app/assets/javascripts/discourse/widgets/header-topic-info.js
+++ b/app/assets/javascripts/discourse/widgets/header-topic-info.js
@@ -57,6 +57,10 @@ createWidget("topic-header-participant", {
 export default createWidget("header-topic-info", {
   tagName: "div.extra-info-wrapper",
 
+  buildClasses(attrs) {
+    return this.showCategory(attrs.topic) ? "two-rows" : "";
+  },
+
   buildFancyTitleClass() {
     const baseClass = ["topic-link"];
     const flatten = array => [].concat.apply([], array);
@@ -112,11 +116,7 @@ export default createWidget("header-topic-info", {
     const category = topic.get("category");
 
     if (loaded || category) {
-      if (
-        category &&
-        (!category.get("isUncategorizedCategory") ||
-          !this.siteSettings.suppress_uncategorized_badge)
-      ) {
+      if (this.showCategory(topic)) {
         const parentCategory = category.get("parentCategory");
         const categories = [];
         if (parentCategory) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/widgets/header-topic-info.js
```diff
@@ -208,6 +208,14 @@ export default createWidget("header-topic-info", {
     );
   },
 
+  showCategory(topic) {
+    return (
+      topic.category &&
+      (!topic.category.get("isUncategorizedCategory") ||
```
jjaffeux, line 214: do we need the `.get` here ?

markvanlan: I can remove it!

app/assets/javascripts/discourse/widgets/header-topic-info.js
```diff
@@ -112,11 +116,7 @@ export default createWidget("header-topic-info", {
     const category = topic.get("category");
 
     if (loaded || category) {
-      if (
-        category &&
-        (!category.get("isUncategorizedCategory") ||
-          !this.siteSettings.suppress_uncategorized_badge)
-      ) {
+      if (this.showCategory(topic)) {
```
jjaffeux, line 119: I'm curious why don't we directly use category here ?

app/assets/javascripts/discourse/widgets/header-topic-info.js
```diff
@@ -57,6 +57,10 @@ createWidget("topic-header-participant", {
 export default createWidget("header-topic-info", {
   tagName: "div.extra-info-wrapper",
 
+  buildClasses(attrs) {
+    return this.showCategory(attrs.topic) ? "two-rows" : "";
```
jjaffeux, line 61: and `attrs.topic.category`  here ?

app/assets/javascripts/discourse/widgets/header-topic-info.js
markvanlan: I like the function name with topic as argument.

> Should we show the category for the given topic?

app/assets/javascripts/discourse/widgets/header-topic-info.js
markvanlan: Same as above. I can change it, but I do like how it reads.

app/assets/javascripts/discourse/widgets/header-topic-info.js
jjaffeux: well ok then :+1:

PR #9478, opened 2020-04-20
FIX: Set category description to first posts cooked value

markvanlan:
We do not notice this bug right now, because the default category description has no markdown that needs to be converted into html.

Once the category definition topic/post are created, simply set the description of the category to the posts `cooked` value.


eviltrout:
Could you please create a test?

markvanlan:
@eviltrout Alright, the specs were actually failing. Category descriotion does not get set to the first post's cooked value if the description was actually blank. 

I added a check to keep the categories description unset if no description is passed into the Category.

PR #9481, opened 2020-04-20
FEATURE: Digital Ocean Spaces compatibility for Object Storage

xfalcox:
DO does not implement tagging support for S3 objects. Removing our default
empty tag fixes compatibility.

The expire_missing_assets rake task can't be used with that service still,
but this patch allows normal operation.


SamSaffron:
was the empty tag added for a reason? are there any concerns around this change @andrewschleifer ?

andrewschleifer:
I do not anticipate any problems, as an empty value shouldn't actually apply a tag to the object anyway.

PR #9482, opened 2020-04-21
FEATURE: don't display new/unread notification for muted topics

lis2:
Currently, even if user mute topic, when a new reply to that topic arrives, the user will get "See 1 new or updated topic" message. After clicking on that link, nothing is visible (because the topic is muted)

To solve that problem, we will send background message to all users who recently muted that topic that update is coming and they can ignore the next message about that topic.

### DIFF
--- a/app/assets/javascripts/discourse/models/topic-tracking-state.js
+++ b/app/assets/javascripts/discourse/models/topic-tracking-state.js
@@ -53,11 +53,7 @@ const TopicTrackingState = EmberObject.extend({
     const tracker = this;
 
     const process = data => {
-      if (
-        tracker
-          .trackMutedTopics(data)
-          .findBy("topicId", data.topic_id)
-      ) {
+      if (tracker.trackMutedTopics(data).findBy("topicId", data.topic_id)) {
         return;
       }
 
--- a/app/assets/javascripts/discourse/models/topic-tracking-state.js
+++ b/app/assets/javascripts/discourse/models/topic-tracking-state.js
@@ -53,6 +53,14 @@ const TopicTrackingState = EmberObject.extend({
     const tracker = this;
 
     const process = data => {
+      if (
+        tracker
+          .trackMutedTopics(data)
+          .find(mutedTopic => mutedTopic.topicId === data.topic_id)
+      ) {
+        return;
+      }
+
       if (data.message_type === "delete") {
         tracker.removeTopic(data.topic_id);
         tracker.incrementMessageCount();
@@ -113,6 +121,10 @@ const TopicTrackingState = EmberObject.extend({
         "/unread/" + this.currentUser.get("id"),
         process
       );
+      this.messageBus.subscribe(
+        "/muted/" + this.currentUser.get("id"),
+        process
+      );
     }
 
     this.messageBus.subscribe("/delete", msg => {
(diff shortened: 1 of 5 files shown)


SamSaffron:
No need to post per user messages, instead publish to a /muted-topics channel and when you publish the message specify the user ids as a param

This saves a lot of message bus work

app/assets/javascripts/discourse/models/topic-tracking-state.js
```diff
@@ -132,6 +141,28 @@ const TopicTrackingState = EmberObject.extend({
     });
   },
 
+  trackMutedTopics(data) {
+    if (!this.currentUser) {
+      return [];
+    }
+    let mutedTopics = this.currentUser.muted_topics || [];
+    let now = new Date();
```
SamSaffron, line 149: in general `Date.now()` is best for usages like this.

app/models/topic_tracking_state.rb
```diff
@@ -71,6 +72,15 @@ def self.unread_channel_key(user_id)
     "/unread/#{user_id}"
   end
 
+  def self.publish_muted(post)
+    user_ids = post.topic.topic_users.where(notification_level: NotificationLevels.all[:muted]).order(notifications_changed_at: :desc).limit(100).pluck(:user_id)
```
SamSaffron, line 76: I prefer to look at last user activity ... join the `users` table looking at `last_seen_at`

When you muted matters a lot less vs when you were last around. 

In fact I would filter out all users that have not been seen in the last 7 days as well.

app/models/topic_tracking_state.rb
```diff
@@ -71,6 +72,15 @@ def self.unread_channel_key(user_id)
...
+    message = {
```
SamSaffron, line 77: move the if here, no need to create the object if you have no user_ids.

app/assets/javascripts/discourse/models/topic-tracking-state.js
```diff
@@ -132,6 +141,28 @@ const TopicTrackingState = EmberObject.extend({
...
+    let mutedTopics = this.currentUser.muted_topics || [];
+    let now = Date.now();
+
+    if (data.message_type === "muted") {
+      mutedTopics = mutedTopics.concat({
+        topicId: data.topic_id,
+        createdAt: now
+      });
+    } else {
+      mutedTopics = mutedTopics.filter(
```
SamSaffron, line 157: shouldn't it run this unconditionally? always do a cleanup run when you track a new muted topic?

SamSaffron:
except for minor comment here on cleanup run, I think this is good to merge, lets let one more person review this.

app/assets/javascripts/discourse/models/topic-tracking-state.js
```diff
@@ -53,6 +53,14 @@ const TopicTrackingState = EmberObject.extend({
     const tracker = this;
 
     const process = data => {
+      if (
+        tracker
+          .trackMutedTopics(data)
+          .find(mutedTopic => mutedTopic.topicId === data.topic_id)
```
jjaffeux, line 59:
```suggestion
          .findBy("topicId", data.topic_id)
```

davidtaylorhq:
Does message_bus guarantee that messages on **different** channels arrive in exactly chronological order? Using two different channels and relying on the order smells like a possible race condition to me 🤔 

Could we put the muted message on the regular `/latest` channel, but scoped to a set of users, and with a different payload? That way, we guarantee that the "muted" message will arrive before the "new topic" message.

SamSaffron:
> Does message_bus guarantee that messages on different channels arrive in exactly chronological order? 

It does, we have a global channel that takes care of that, but yeah cutting out a subscription here makes sense, just feed it in via latest

app/assets/javascripts/discourse/models/topic-tracking-state.js
lis2: I don't think it is necessary. When you mute topic, you will receive 2 messages. First "muted-topic" will just add that topic to an array of muted topics. Second message "latest", will filter muted topics and decide if the message should be processed or not. 

If we filter on "muted" message as well we will do it twice in a very short time and "latest" is a more important message because if you unmute topic you will not receive "muted" message anymore.

SamSaffron: aha... ok lets refactor a bit, I feel this method is doing a bit too much work...

```
if (data.message_type == "muted") {
   tracker.trackMutedTopic(data.topic_id)
   return;
}

pruneOldMutedTopics();

if (isMutedTopic(data.topic_id)) {
    return;
}


```

a lot more verbose, but significantly less full of side effects so it is far simpler to reason about

SamSaffron:
Looking good :+1:

PR #9483, opened 2020-04-21
DEV: Fix topic_timer_spec

CvX:
File it under "how did it even work".

Removes workaround introduced in 003df147a59e53ef6e94a330137fa07e2651ebb9.


SamSaffron:
yeah no idea how it used to work :) thanks for the refactor!

PR #9487, opened 2020-04-21
FEATURE: Redesign discourse-presence to track state on the client side. 

tgxworld:


plugins/discourse-presence/plugin.rb
```diff
@@ -22,154 +22,50 @@ class Engine < ::Rails::Engine
...
+      topic = Topic.find_by(id: topic_id)
+      guardian.ensure_can_see!(topic)
+
+      MessageBus.publish(
+        "/presence/#{topic_id}",
+        {
+          user: BasicUserSerializer.new(current_user, root: false).as_json,
+          state: permitted_params[:state]
+        },
+        group_ids: topic.secure_group_ids,
```
SamSaffron, line 43: if it is a whisper you are editing group_ids is only going to be the intersection of "staff" with topic secure groups.

Also the blew away all support for pms... you are going to need to reconstruct that ... topic allowed users / groups for messages.

tgxworld: Yup I've not tackled that yet. I've only gotten replying and closing of the composer to publish the right messages so far.

SamSaffron: no worries at all, the approach here looks a-ok to me.

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -0,0 +1,197 @@
...
+
+    const props = attrs || {};
+    props.last_seen = Date.now();
+
+    if (existingUser) {
+      existingUser.setProperties(props);
+    } else {
+      const limit = this.get("siteSettings.presence_max_users_shown");
+
+      if (usersLength < limit) {
```
tgxworld, line 190: This acts a guard to ensure that memory doesn't blow up as we store the state of users in the `users` array. For `editingUsers`, we can technically still end up storing a lot of users in the array if everyone edits at the same time which seems kind of rare so I'm not tackling it yet. If we're worried about the memory growth here, we can't rely on the same messageBus channel that is used for replying to a topic.

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
+    presenceManager && presenceManager.subscribe();
   },
 
-  @observes("action", "post.id", "topic.id")
-  composerStateChanged() {
-    once(this, "updateState");
+  @computed("post.id", "editingUsers.@each.last_seen", "users.@each.last_seen")
+  presenceUsers(postId, editingUsers, users) {
+    if (postId) {
+      return editingUsers.filter(user => {
```
jjaffeux, line 30:
```
return editingUsers.filterBy("post_id", postId)
```
Should be the same

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
 
+  shouldDisplay: gt("presenceUsers.length", 0),
+
   @observes("reply", "title")
   typing() {
-    if (new Date() - this._lastPublish > keepAliveDuration) {
-      this.publish({ current: this.currentState });
-    }
-  },
+    const presenceManager = this.get("presenceManager");
```
jjaffeux, line 42: I think you can just use: `this.presenceManager` in all the codeblock

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
-    let state = null;
-    const action = this.action;
-
-    if (action === "reply" || action === "edit") {
-      state = { action };
-      if (action === "reply") state.topic_id = this.get("topic.id");
-      if (action === "edit") state.post_id = this.get("post.id");
+      this._throttle = presenceManager.throttlePublish(
+        postId ? EDITING : REPLYING,
+        this.get("whisper"),
```
jjaffeux, line 49:
```suggestion
        this.whisper,
```

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
-    }
+  @observes("whisper")
+  cancelThrottle() {
+    this._cancelThrottle();
+  },
 
-    this.clear();
+  @observes("post.id")
+  stopEditing() {
+    const presenceManager = this.get("presenceManager");
```
jjaffeux, line 62: same remark about presenceManager

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
 
-  clear() {
-    if (!this.isDestroyed) this.set("presenceUsers", []);
-  },
-
-  publish(data) {
-    this._lastPublish = new Date();
+  @on("willDestroyElement")
+  composerClosing() {
+    const presenceManager = this.get("presenceManager");
```
jjaffeux, line 71: same

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
 
-    // Don't publish presence if disabled
-    if (this.currentUser.hide_profile_and_presence) {
-      return Promise.resolve();
+    if (presenceManager) {
+      this._cancelThrottle();
+      presenceManager.publish(CLOSED, this.get("whisper"));
```
jjaffeux, line 75:
```suggestion
      presenceManager.publish(CLOSED, this.whisper);
```

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
   title: null,
+  isWhispering: null,
 
-  // Internal variables
-  previousState: null,
-  currentState: null,
-  presenceUsers: null,
-  channel: null,
-
+  presenceManager: alias("topic.presenceManager"),
```
jjaffeux, line 16: Generally speaking, I'm not a big fan of alias, in your case I think https://api.emberjs.com/ember/3.17/functions/@ember%2Fobject%2Fcomputed/readOnly is a better choice, given you don't want to do set on these aliased properties

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -0,0 +1,197 @@
...
+    this.get("messageBus").unsubscribe(this.get("channel"));
+    this.set("subscribed", false);
+  },
+
+  @computed("topic.id")
+  channel(topicId) {
+    return `/presence/${topicId}`;
+  },
+
+  @observes("users.@each.last_seen")
```
jjaffeux, line 90: can you try to use the `didReceiveAttrs` hook instead of all these observers ? We generally try to avoid observers which are mostly an implementation detail of Ember.

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -0,0 +1,197 @@
...
+          this._removeUser(user);
+          break;
+      }
+    });
+
+    this.set("subscribed", true);
+  },
+
+  unsubscribe() {
+    this.get("messageBus").unsubscribe(this.get("channel"));
```
jjaffeux, line 81: I won't report all of them, but there's mostly no reason to do this.get("xxx") instead of this.xxx, get is only usefull when you have multiple nesting levels: this.get("foo.bar.baz")

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +11,72 @@ export default Component.extend({
...
-    once(this, "updateState");
+  subscribe() {
+    const presenceManager = this.get("presenceManager");
+    presenceManager && presenceManager.subscribe();
   },
 
-  @observes("action", "post.id", "topic.id")
-  composerStateChanged() {
-    once(this, "updateState");
+  @computed("post.id", "editingUsers.@each.last_seen", "users.@each.last_seen")
```
jjaffeux, line 27: this computed should now use @discourseComputed

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
tgxworld: :thinking: Would this work here since the class isn't an Ember component? I thought observers were provided so that run callbacks when Ember detects that a specific attribute has changed. I don't think we get the same with `didReceiveAttrs`.

plugins/discourse-presence/assets/javascripts/discourse/components/composer-presence-display.js.es6
```diff
@@ -15,115 +14,68 @@ export default Component.extend({
...
-  composerOpened() {
-    this._lastPublish = new Date();
-    once(this, "updateState");
-  },
-
-  @observes("action", "post.id", "topic.id")
-  composerStateChanged() {
-    once(this, "updateState");
+  subscribe() {
+    const presenceManager = this.presenceManager;
```
jjaffeux, line 26: I would just use this.presenceManager directly, no need for this variable IMO

tgxworld: o good catch :+1:

jjaffeux:
js code looks good to me now, but this is quite complex, will have to pull the PR and try locally, will try to do it today.

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -0,0 +1,205 @@
...
+    const props = attrs || {};
+    props.last_seen = Date.now();
+
+    if (existingUser) {
+      existingUser.setProperties(props);
+    } else {
+      const limit = this.get("siteSettings.presence_max_users_shown");
+
+      if (usersLength < limit) {
+        users.pushObject(EmberObject.create(Object.assign(user, props)));
```
jjaffeux, line 171:
```suggestion
        users.pushObject(EmberObject.create({}, Object.assign(user, props)));
```

Otherwise you are actually mutating user here, not sure this is what you wanted to do.

CvX: @jjaffeux The suggested change is still mutating the user with props. 😛

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -0,0 +1,205 @@
...
+      data.post_id = postId;
+    }
+
+    return ajax("/presence/publish", {
+      type: "POST",
+      data
+    });
+  },
+
+  _debounceSetProperties(properties) {
```
jjaffeux, line 119: where is this used?

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
jjaffeux: Oh yes didn't put it at right position :+1:

jjaffeux:
Been testing this quite a lot today, and it seems to work very well ❤️ 

I have one last recommendation, given there are no JS tests, we should make this disabled during client tests.

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
tgxworld: Mutating the `user` object is fine here because it is from the message's payload.

plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
tgxworld: Oops good catch :+1: I thought I removed this.

SamSaffron:
Lets try this out!!! thanks @tgxworld  !

PR #9488, opened 2020-04-21
FIX: Copying image markdown for secure media loading full image

martin-brennan:
* When copying the markdown for an image between posts, we were not adding the `srcset` and `data-small-image` attributes which are done by calling optimize_image! in cooked post processor
* Refactored the code which was confusing in its current state (the `consider_for_reuse` method was super confusing) and fixed the issue


app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -181,9 +181,14 @@ def should_download_image?(src, post = nil)
         # media was enabled, then we definitely want to redownload again otherwise
         # we end up reusing existing uploads which may be linked to many posts
         # already.
-        upload = Upload.consider_for_reuse(Upload.get_from_url(src), post)
+        upload = Upload.get_from_url(src)
+        return true if upload.blank?
 
-        return !upload.present?
+        if SiteSetting.secure_media && upload.access_control_post_id.present?
+          return true if upload.copied_from_other_post?(post) || upload.uploaded_before_secure_media_enabled?
```
davidtaylorhq, line 188: So... this code will copy secure images from one post to another, simply using the URL? 

Could someone use this to steal images which they're not supposed to have access to?

martin-brennan: @davidtaylorhq I will check this out, this is a good point. Definitely want to redownload if `uploaded_before_secure_media_enabled?` but yeah maybe not in the other case. Thanks for pointing this out!

SamSaffron:
Do we really want to pull this in if we are just going to be removing this anyway? Are we going to keep any of this code after we add the stricter rules?

martin-brennan:
@SamSaffron still worth pulling in, I think its good to have the server-side checks there. At any rate it may be a little while until I get stricter rules merged so would be good to have it fixed until then. Oh also it has a legit fix for pull_hotlinked_images when the hotlinked image is secure so that needs to be merged anyway.

app/jobs/regular/pull_hotlinked_images.rb
```diff
@@ -181,9 +185,22 @@ def should_download_image?(src, post = nil)
...
+        upload = Upload.get_from_url(src)
+        return true if upload.blank?
+
+        if SiteSetting.secure_media && upload.access_control_post_id.present?
+          if upload.copied_from_other_post?(post)
+            # only redownload if the person making this post can access the
+            # upload in the original post. this is to stop the loophole of
+            # having a URL for a post that you should not have and copying it
+            # into a public post to redownload that upload
+            return Guardian.new(post.user).can_see?(upload.access_control_post)
```
davidtaylorhq, line 197: I don't think this is strong enough. What if an admin creates a wiki post, and then a TL1 edits it? Now the TL1 can steal any uploads which the admin can access.

I don't think we can solve this cleanly. I think we should never try and "pull hotlinked" on secure uploads.

app/models/upload.rb
```diff
@@ -123,14 +123,21 @@ def short_url
     "upload://#{short_url_basename}"
   end
 
-  def short_path
-    self.class.short_path(sha1: self.sha1, extension: self.extension)
+  def uploaded_before_secure_media_enabled?
```
davidtaylorhq, line 126: Why is this function needed? Don't we backfill the uploads table when secure media is enabled?

lib/cooked_post_processor.rb
```diff
@@ -339,12 +339,7 @@ def convert_to_link!(img)
       width, height = ImageSizer.resize(width, height)
     end
 
-    # if the upload already exists and is attached to a different post,
-    # or the original_sha1 is missing meaning it was created before secure
-    # media was enabled. we want to re-thumbnail and re-optimize in this case
-    # to avoid using uploads linked to many other posts
-    upload = Upload.consider_for_reuse(Upload.get_from_url(src), @post)
-
```
davidtaylorhq, line 347: This is a nice change. I agree there is no need to prevent thumbnail generation for copied secure uploads 👍

app/models/upload.rb
martin-brennan: @davidtaylorhq you don't _have_ to run the rake task that goes and marks old uploads as secure, and it is not automatic. Depending on the forum it may not even be wise if there are a lot of uploads reused all over the place, because the first time it is used in a post that post will be used as the access control post, which may break subsequent uses of the upload.

You can just turn on secure uploads + have it apply from that point in time. From an issue we had internally this check on pull hotlinked is used so if someone copies an old non-secure image into a secure context it is always created as a new upload so it does not get an access_control_post attached to the old upload.

PR #9490, opened 2020-04-21
DEV: Update mocha

CvX:
The spec that was blocking the update was fixed in c08753dc3431ce2f85916b5a7e68db52891917d8.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -160,10 +160,7 @@ group :test, :development do
   gem 'listen', require: false
   gem 'certified', require: false
   gem 'fabrication', require: false
-
-  # TODO: upgrading to 1.10.1 cause it breaks our test suite.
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -161,9 +161,7 @@ group :test, :development do
   gem 'certified', require: false
   gem 'fabrication', require: false
 
-  # TODO: upgrading to 1.10.1 cause it breaks our test suite.
-  # We want our test suite fixed though to support this upgrade.
-  gem 'mocha', '1.8.0', require: false
+  gem 'mocha', '1.11.2', require: false
```
davidtaylorhq, line 164: Can we remove the pinned version completely, so that we get auto-PRs for new versions?

CvX: doh, of course 😅

PR #9498, opened 2020-04-21
UX: improves existing date-time-inputs

jjaffeux:
Few examples:
Screenshot 2020-04-21 at 22 51 39
Screenshot 2020-04-21 at 22 48 13
Screenshot 2020-04-21 at 22 48 29

This can be test in the current app at /review and /admin/reports/signups but most interesting use cases are coming in the future.

### DIFF
--- a/app/assets/javascripts/select-kit/components/select-kit.js
+++ b/app/assets/javascripts/select-kit/components/select-kit.js
@@ -257,7 +257,7 @@ export default Component.extend(
       autoFilterable: "autoFilterable",
       filterIcon: "search",
       filterPlaceholder: "filterPlaceholder",
-      translatedfilterPlaceholder: null,
+      translatedFilterPlaceholder: null,
       icon: null,
       icons: null,
       maximum: null,
--- a/app/assets/javascripts/select-kit/components/select-kit/select-kit-filter.js
+++ b/app/assets/javascripts/select-kit/components/select-kit/select-kit-filter.js
@@ -1,6 +1,6 @@
 import Component from "@ember/component";
 import discourseComputed from "discourse-common/utils/decorators";
-import { isEmpty } from "@ember/utils";
+import { isPresent } from "@ember/utils";
 import { computed } from "@ember/object";
 import { not } from "@ember/object/computed";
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/discourse/components/time-input.js
```diff
@@ -1,76 +1,164 @@
-import { oneWay, or } from "@ember/object/computed";
-import { schedule } from "@ember/runloop";
+import { isPresent } from "@ember/utils";
+import { computed, action } from "@ember/object";
 import Component from "@ember/component";
-import { isNumeric } from "discourse/lib/utilities";
+
+function convertMinutes(n) {
+  const num = n;
```
eviltrout, line 6: Could be simplified to: 
```
return { hours: Math.floor(num / 60), minutes: num % 60 }
```

app/assets/javascripts/discourse/templates/components/date-time-input-range.hbs
```diff
@@ -1,35 +1,16 @@
...
-    date=to
-    onChange=(action "_onChange" (hash prop="to"))
-    showTime=showToTime
-    clearable=true
-  }}
-</div>
+{{date-time-input
+  date=to
+  relativeDate=from
+  onChange=(action "_onChange" (hash prop="to"))
```
eviltrout, line 11: This is a bit unusual. An underscore implies it's a private API yet it's used in the template? I recommend a different name.

app/assets/javascripts/discourse/templates/components/time-input.hbs
```diff
@@ -1,40 +1,6 @@
...
-      input=(action "onInput" (hash prop="minutes"))
-      focus-in=(action "onFocusIn")
-    }}
-  {{/if}}
-</div>
+{{combo-box
+  value=time
+  content=timeOptions
+  onChange=(action "onChangeTime")
+  options=(hash translatedNone="00:00" allowAny=true filterable=false autoInsertNoneItem=false translatedfilterPlaceholder="Set time...")
```
eviltrout, line 5: Looks like some missing translations here: "Set time..."

PR #9501, opened 2020-04-21
DEV: Export openNewCategoryModal from discovery-categories

markvanlan:
Exporting this function allows plugins to be able to open the "Create Category" modal, without repeating any code.


app/assets/javascripts/discourse/routes/discovery-categories.js
```diff
@@ -148,4 +133,22 @@ const DiscoveryCategoriesRoute = DiscourseRoute.extend(OpenComposer, {
   }
 });
 
+export const openNewCategoryModal = context => {
```
eviltrout, line 136:
```suggestion
export function openNewCategoryModal(context) {
```

PR #9502, opened 2020-04-21
DEV: Update lefthook to run lints only on pre-push.

tgxworld:


tgxworld:
Closing this for now while we figure out a way to make it configurable in the dev env.

PR #9504, opened 2020-04-22
DEV: Add checks that assets do not modify cookies to smoke-test.js

riking:
Test cases as a followup from a past postmortem related to nginx configs.

This test will fail when executed against a Rails server without the container's nginx (or an improperly configured nginx) in front of it that blocks the application from setting cookies on cached asset requests.

Haven't figured out how to build discourse_test locally, so I'm throwing this against CI. EDIT: Done :)

### DIFF
--- a/test/smoke_test.js
+++ b/test/smoke_test.js
@@ -197,14 +197,14 @@ async function getCookie(name, page) {
     "service-worker.js does not set cookies",
     checkNoCookies("/service-worker.js")
   );
-  await exec("cookie check works", async () => {
+  await exec("application paths do clear invalid cookies", async () => {
     const fn = checkNoCookies("/about");
     let failure = false;
     try {
       await fn();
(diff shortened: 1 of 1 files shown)


test/smoke_test.js
```diff
@@ -129,10 +142,75 @@ const path = require("path");
...
+    checkNoCookies("/service-worker.js")
+  );
+  await exec("cookie check works", async () => {
+    const fn = checkNoCookies("/about");
+    let failure = false;
+    try {
+      await fn();
+      failure = true;
+    } catch (e) {
+      // pass
```
featheredtoast, line 207: Might be good to update the comment here a little more explicitly say that it passes because we expect checkNoCookies to fail+throw; this may be difficult to reason about/read through later down the road otherwise.

riking:
Test failure is a flake.

PR #9505, opened 2020-04-22
FIX: If bookmarking discobot posts, skip the reminder modal

martin-brennan:
This is to streamline the new user narrative. only works when creating the bookmark, if editing/deleting the modal is shown.

### DIFF
--- a/app/assets/javascripts/discourse/models/post.js
+++ b/app/assets/javascripts/discourse/models/post.js
@@ -336,20 +336,6 @@ const Post = RestModel.extend({
   },
 
   toggleBookmarkWithReminder() {
-    // if we are talking to discobot then any bookmarks should just
-    // be created without reminder options, to streamline the new user
-    // narrative.
-    if (this.username === "discobot" && !this.bookmarked_with_reminder) {
-      return ajax("/bookmarks", { type: "POST", data: { post_id: this.id } }).then((response) => {
-        this.setProperties({
-          "topic.bookmarked": true,
-          bookmarked_with_reminder: true,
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/models/post.js
```diff
@@ -336,6 +336,20 @@ const Post = RestModel.extend({
   },
 
   toggleBookmarkWithReminder() {
+    // if we are talking to discobot then any bookmarks should just
+    // be created without reminder options, to streamline the new user
+    // narrative.
+    if (this.username === "discobot" && !this.bookmarked_with_reminder) {
```
tgxworld, line 342: Hmm I'm wondering if we can override this method in a plugin initializer instead so that discobot specific code can stay within the discobot plugin.

plugins/discourse-narrative-bot/assets/javascripts/initializers/new-user-narrative.js.es6
```diff
@@ -12,6 +13,28 @@ function initialize(api) {
     }
   });
 
+  api.modifyClass("model:post", {
+    toggleBookmarkWithReminder() {
+      // if we are talking to discobot then any bookmarks should just
+      // be created without reminder options, to streamline the new user
+      // narrative.
+      if (this.username === "discobot" && !this.bookmarked_with_reminder) {
```
tgxworld, line 21: We should try and use the user id here since the bot can be renamed.

tgxworld:
LGTM :+1:

PR #9511, opened 2020-04-22
FEATURE: add class to <aside> quote block when quoting an ignored user

pnoeric:
Adds the class 'ignored-user' to the `<aside>` quote block, when we are looking at a quote from an ignored user.

Per https://meta.discourse.org/t/tiny-feature-request-for-core-add-3-classes/148725


pnoeric:
No idea why some tests are failing. As you can see, my submission is exactly one line of code, with spaces instead of tabs, etc. 🤷🏻‍♂️

eviltrout:
It looks like the failures are linting related. Can you use double quotes instead of single quotes?

eviltrout:
Thanks!

pnoeric:
Thank YOU!

PR #9515, opened 2020-04-22
Improve support for old browsers

udan11:
Before:

before_home
before_topic

After:

after_home
after_topic


app/views/list/list.erb
```diff
@@ -44,33 +44,63 @@
...
                 <%= t.excerpt.html_safe %>
               </p>
             <% end %>
           </td>
+          <td class='posters'>
+            <% t.posters.each do |poster| %>
+              <% if poster.moreCount %>
+                <a class="posters-more-count"><% poster.moreCount %></a>
+              <% else %>
+                <a href="<%= Discourse.base_url %>/u/<%= poster.user.username %>" class="<%= poster.extraClasses %>">
```
riking, line 95: Let's maybe drop the topic list avatars, or put them inside an `if false`, until the topic view has them too.

SamSaffron:
Very nice changes, what is holding us up on merging?

PR #9517, opened 2020-04-22
PERF: Refactor decorateCooked to run in a detached DOM

davidtaylorhq:
This means that the HTML can be modified without triggering the download of remote resources (e.g. images)

In some rare cases (e.g. IntersectionObservers in Chromium), decorating needs to happen in the real DOM. For this, pass `afterAdopt: true` to `decorateCooked`

### DIFF
--- a/app/assets/javascripts/discourse/lib/lazy-load-images.js
+++ b/app/assets/javascripts/discourse/lib/lazy-load-images.js
@@ -85,7 +85,7 @@ function show(image) {
 }
 
 function forEachImage($post, callback) {
-  $("img", $post).each((_, img) => {
+  $post[0].querySelectorAll("img").forEach(img => {
     if (img.width >= MINIMUM_SIZE && img.height >= MINIMUM_SIZE) {
       callback(img);
     }
--- a/app/assets/javascripts/discourse/lib/lazy-load-images.js
+++ b/app/assets/javascripts/discourse/lib/lazy-load-images.js
@@ -84,6 +84,14 @@ function show(image) {
   }
 }
 
+function forEachImage($post, callback) {
+  $("img", $post).each((_, img) => {
+    if (img.width >= MINIMUM_SIZE && img.height >= MINIMUM_SIZE) {
+      callback(img);
+    }
+  });
+}
+
 export function setupLazyLoading(api) {
   const observer = new IntersectionObserver(entries => {
     entries.forEach(entry => {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/lib/lazy-load-images.js
```diff
@@ -84,6 +84,14 @@ function show(image) {
   }
 }
 
+function forEachImage($post, callback) {
+  $("img", $post).each((_, img) => {
```
jjaffeux, line 88: can't this be:

```
$post[0].querySelectorAll("img").forEach(xxx
```

app/assets/javascripts/discourse/widgets/post-cooked.js
```diff
@@ -41,14 +41,31 @@ export default class PostCooked {
...
+
+    return cookedDiv;
+  }
+
+  _decorateAndAdopt(cooked) {
+    const $cooked = $(cooked);
+
+    _decorators
+      .filter(d => !d.afterAdopt)
+      .forEach(d => d.callback($cooked, this.decoratorHelper));
```
jjaffeux, line 62: we want to move away from sending jquery object in decorate cooked, I understand you do this for consistency... but I'm unsure, maybe we should just give regular object? Wha do you think @eviltrout ?

app/assets/javascripts/discourse/widgets/post-cooked.js
```diff
@@ -273,19 +292,29 @@ export default class PostCooked {
     });
   }
 
+  _createDetachedElement(nodeName) {
```
eviltrout, line 295: Minor but this method doesn't access any instance variables and could be more private by putting it at the top of the module as a regular function.

app/assets/javascripts/discourse/widgets/post-cooked.js
```diff
@@ -41,14 +41,31 @@ export default class PostCooked {
...
+
+    this._decorateAndAdopt(cookedDiv);
+
+    return cookedDiv;
+  }
+
+  _decorateAndAdopt(cooked) {
+    const $cooked = $(cooked);
+
+    _decorators
```
eviltrout, line 60: This is going to be called a LOT and the filter is the same result every time.

I would recommend making separate `_beforeAdoptDecorators` and `_afterAdoptDecorators` so you needn't filter every time.

app/assets/javascripts/discourse/lib/lazy-load-images.js
davidtaylorhq: Nice! I removed a lot of JQuery but missed this one. Updated :+1:

app/assets/javascripts/discourse/widgets/post-cooked.js
davidtaylorhq: :+1: done, thanks. I also refactored so that the detached document is re-used, rather than recreating each time

app/assets/javascripts/discourse/widgets/post-cooked.js
davidtaylorhq: Good idea, thanks ❤️

PR #9522, opened 2020-04-22
DEV: DiscourseEvents fired when users are added/removed from groups

markvanlan:
These are generally useful hooks that plugins for centuries to come will likely hook into.

There could also be put into `GroupHistory` and fired whenever a GroupHistory record is created. I thought 'GroupActionLogger' sounded like a proper place though.


SamSaffron:
looks good to me!

PR #9525, opened 2020-04-22
FEATURE: dismiss first notification on click anywhere

featheredtoast:
Quicker jumpstart for those already familiar with the platform:

Allow dismissal of first notification mask from any click.
On the dismissal click, we also need to send a "yes I saw it" confirmation
by grabbing a batch of notifications. This prevents the dialog from appearing
again on refresh, or other browsers and ensures we only see it once.

### DIFF
--- a/app/assets/javascripts/discourse/app/components/site-header.js
+++ b/app/assets/javascripts/discourse/app/components/site-header.js
@@ -205,6 +205,30 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     this.dispatch("search-autocomplete:after-complete", "search-term");
 
     this.appEvents.on("dom:clean", this, "_cleanDom");
+
+    // Allow first notification to be dismissed on a click anywhere
+    if (
+      !this.get("currentUser.read_first_notification") &&
+      !this.get("currentUser.enforcedSecondFactor")
+    ) {
+      document.addEventListener(
+        "click",
+        e => {
+          if (
+            !e.target.closest("#current-user") &&
+            !e.target.closest(".ring-backdrop") &&
+            !this.currentUser.get("read_first_notification") &&
+            !this.currentUser.get("enforcedSecondFactor")
+          ) {
+            this.eventDispatched(
+              "header:dismiss-first-notification-mask",
+              "header"
+            );
+          }
+        },
+        { once: true }
+      );
+    }
   },
 
   _cleanDom() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/widgets/header.js
```diff
@@ -276,10 +276,51 @@ export function attachAdditionalPanel(name, toggle, transformAttrs) {
...
+  init() {
+    // Allow first notification to be dismissed on a click anywhere
+    if (
+      this.currentUser &&
+      !this.currentUser.get("read_first_notification") &&
+      !this.currentUser.get("enforcedSecondFactor") &&
+      !addedDismissHandler
+    ) {
+      addedDismissHandler = true;
+      document.addEventListener(
```
eviltrout, line 293: I don't love adding a click handler in a widget `init()`.

It might be easier to add the click handler in the header component `didInsertElement`. Then you could trigger an appEvent that this widget listens to.

featheredtoast: @eviltrout That's right, I forgot about app events for widgets - I've moved the addition of the event listener out to the site-header component :+1: 

Hopefully this looks a bit better!

app/assets/javascripts/discourse/components/site-header.js
```diff
@@ -205,6 +205,30 @@ const SiteHeaderComponent = MountWidget.extend(Docking, PanEvents, {
     this.dispatch("search-autocomplete:after-complete", "search-term");
 
     this.appEvents.on("dom:clean", this, "_cleanDom");
+
+    // Allow first notification to be dismissed on a click anywhere
+    if (
+      !this.get("currentUser.read_first_notification") &&
+      !this.get("currentUser.enforcedSecondFactor")
+    ) {
+      document.addEventListener(
```
eviltrout, line 214: There should be an equivalent removeEventListener in the willDestroyElement. 

(I realize in normal Discourse operation the header is never removed, but in an test environment you want to make sure to get this stuff right.)

featheredtoast: That's what I get for brain fogging my way through this, thanks for catching that - I was thinking that because it was a "once" event, it'd remove itself (in most cases), but I was forgetting about test case cleanup! Updated once more.

featheredtoast:
Oops, just needed to shuffle the event function around a little bit to make the closures correct, it works and I can merge it now. :+1:

PR #9526, opened 2020-04-23
UX: Improve second factor UI

tshenry:
This will make a few minor improvements to the second factor user interface. Highlights include:

- Using the site's title to prefix the backup code filename. If non-ascii characters are detected, then prefix "discourse" instead.
- Add icons and change the text on some of the buttons for better clarity and consistency
- Add an education link to the security key modal


SamSaffron:
@martin-brennan  can you review / merge

app/assets/javascripts/discourse/app/components/backup-codes.js
```diff
@@ -42,6 +42,23 @@ export default Component.extend({
     return backupCodes.join("\n").trim();
   },
 
+  @discourseComputed()
+  siteTitleLowerCase() {
```
tgxworld, line 46: I think having a unit test here will be great as documentation of what this function does :)

martin-brennan: Have to agree here, especially `^[\000-\177]`, I have no idea what that does.

tshenry:
@martin-brennan that one is making sure that the string has all ASCII characters using an octal range. (see [here](https://stackoverflow.com/a/5185722)). I do agree that entire function is convoluted and needs some kind of explanation. What you see is me trying my best to work together a javascript equivalent to ruby's `parameterize`. Extremely open to alternative ideas if you have any, otherwise I'll work on a test next week!

tgxworld:
> together a javascript equivalent to ruby's parameterize.

If that is the case, I think it'll be good to move the function into a utilities directory. The function name will probably need to be updated because `siteTitleLowerCase` seems to suggest that it is just returning a lower cased site title.

martin-brennan:
Ah interesting @tshenry thanks for explaining :) Utilities is the way to go then IMO

jjaffeux:
@tshenry do you need help on this? Ping me today if needed, I got some time.

test/javascripts/lib/utilities-test.js
```diff
@@ -175,6 +177,39 @@ QUnit.test("caretRowCol", assert => {
   document.body.removeChild(textarea);
 });
 
+QUnit.test("toAsciiPrintable", assert => {
```
jjaffeux, line 180: should have a test for _

PR #9527, opened 2020-04-23
FEATURE: Optionally show local time for user in card

martin-brennan:
This adds a site setting (default off) to optionally show a user's local time and timezone in their user card. For example, I live in Brisbane, and if at 3:30PM my time I were to open a user who lives in California's card I would see 22:30 (PST).

image


app/assets/javascripts/discourse/components/user-card-contents.js
```diff
@@ -38,7 +38,11 @@ export default Component.extend(CardContentsBase, CanCheckEmails, CleansUp, {
   showMoreBadges: gt("moreBadgesCount", 0),
   showDelete: and("viewingAdmin", "showName", "user.canBeDeleted"),
   linkWebsite: not("user.isBasic"),
-  hasLocationOrWebsite: or("user.location", "user.website_name"),
+  hasLocaleOrWebsite: or(
+    "user.location",
+    "user.website_name",
+    "user._timezone"
```
eviltrout, line 44: We should not be making computed properties dependant on private member variables.

We have to either convert `_timezone` to be a public variable or remove it from this.

app/assets/javascripts/discourse/templates/components/user-card-contents.hbs
```diff
@@ -158,6 +158,12 @@
                 <span>{{this.user.location}}</span>
               </span>
             {{/if}}
+            {{#if showUserLocalTime}}
+              <span class="local-time">
```
eviltrout, line 162: Should probably check for this in an acceptance test.

PR #9536, opened 2020-04-23
Fixing the plural plugin with spanish translations

leog:
Keywords "other" and "one" in the plural plugin should not be translated, they are necessary to configure singular and plural choices of the texts. Moved the translations inside the brackets to be taken into account in the translation of the text itself.


gschlager:

leog:
@gschlager Ok. If you can add me to the project in Transifex I can replicate these changes. My username is `lgiovanetti`.

PR #9541, opened 2020-04-24
FIX: Do not save bookmark if close (X) on modal is clicked

martin-brennan:
* After this change the bookmark will still be saved if clicking out of the modal or pressing escape
* To achieve this I implemented an initiatedBy parameter for modal closing from d-modal. If clicking on the cross it is initiated by close, if clicking out of the modal it is by click out.
* These options can then be compared in controllers consuming `onClose`

### DIFF
--- a/app/assets/javascripts/discourse/app/templates/components/d-modal.hbs
+++ b/app/assets/javascripts/discourse/app/templates/components/d-modal.hbs
@@ -3,7 +3,7 @@
     <div class="modal-inner-container">
       <div class="modal-header">
         {{#if dismissable}}
-          {{d-button icon="times" action=(route-action "closeModal") class="btn btn-flat modal-close close" title="modal.close"}}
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/discourse/app/templates/components/d-modal.hbs
```diff
@@ -3,7 +3,7 @@
     <div class="modal-inner-container">
       <div class="modal-header">
         {{#if dismissable}}
-          {{d-button icon="times" action=(route-action "closeModal") class="btn btn-flat modal-close close" title="modal.close"}}
+          {{d-button icon="times" action=(route-action "closeModal" "initiatedByCloseButton") class="btn btn-flat modal-close close" title="modal.close"}}
```
jjaffeux, line 6:
```suggestion
          {{d-button icon="times" action=(route-action "closeModal" "initiatedByCloseButton") class="btn-flat modal-close close" title="modal.close"}}
```

PR #9545, opened 2020-04-24
DEV: Move `discourse.js` to `app/app.js`

eviltrout:
This is another thing to get our application in line with what Ember CLI
expects.


config/application.rb
```diff
@@ -324,7 +324,14 @@ def config.database_configuration
 
         ActionView::Base.precompiled_asset_checker = -> logical_path do
           default_checker[logical_path] ||
-            %w{qunit.js qunit.css test_helper.css test_helper.js wizard/test/test_helper.js}.include?(logical_path)
+            %w{qunit.js
```
jjaffeux, line 327: I don't understand what is this file doing in this commit? how is this related?

eviltrout: Sorry these errors came up while I was building this, but basically it means if there is a node_modules or dist anywhere in discourse everything would break. I can split it up if you prefer.

jjaffeux: No up to you :)

Was just wondering if I was missing something :+1:

PR #9548, opened 2020-04-26
FIX: correctly shows no bookmark message

jjaffeux:


app/assets/javascripts/discourse/app/controllers/user-activity-bookmarks-with-reminders.js
```diff
@@ -38,69 +35,72 @@ export default Controller.extend({
...
-          this.setProperties({
-            loadingMore: false
-          })
-        );
+    response = response.user_bookmark_list;
+    this.model.more_bookmarks_url = response.more_bookmarks_url;
+
+    if (response.bookmarks) {
+      this.content.pushObjects(
+        response.bookmarks.map(bookmark => Bookmark.create(bookmark))
```
eviltrout, line 102: I find this kind of operation is much better suited for a model than a controller. Is there a reason we can't do the transformation in the model?

discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/fix-correctly-shows-no-bookmark-message-pr-9548/10934/3):

Agreed, I might merge it like this and refactor with martin in another commit if that's ok though.

martin-brennan:
Thanks @jjaffeux! :+1:

PR #9554, opened 2020-04-27
DEV: Always use a tmp directory for storing uploads in tests

davidtaylorhq:
This avoids development-mode upload files from polluting the test environment


eviltrout:
Now that you've made this (great!) change, I wonder if you should follow up by removing `is_parallel_test?` from `discourse.rb`

My thinking is now the app never needs to know. It's only used in `db.rake` and that could check the ENV values directly (or have a local helper function.)

In other words, why pollute the public API with something so specific?

PR #9555, opened 2020-04-27
FIX: currentUser might be undefined

jjaffeux:


jjaffeux:
https://github.com/discourse/discourse/commit/e638d43f0a7549a75eb9de57bb8508b36f11543d was causing a lot of silent exceptions when running tests, can you review this please @featheredtoast

SamSaffron:
Looks like a safe change to me :+1:

PR #9556, opened 2020-04-27
FIX: Only show the review page to users that can see it. Do not publish the reviewable count update message to everyone.

romanrizzi:


app/controllers/reviewables_controller.rb
```diff
@@ -6,6 +6,7 @@ class ReviewablesController < ApplicationController
   PER_PAGE = 10
 
   before_action :version_required, only: [:update, :perform]
+  before_action :can_see_review_queue
```
eviltrout, line 9: nitpick but I would call this something different than `can_` - something simple like `ensure_can_see` would be closer to the operation.

app/jobs/regular/notify_reviewable.rb
```diff
@@ -48,6 +48,8 @@ def count_for_group(group_id)
   end
 
   def notify(count, user_ids)
+    return if user_ids.empty?
```
eviltrout, line 51:
```suggestion
    return if user_ids.blank?
```
`blank?` here will work if it's `nil` too - just in case. More defensive.

PR #9562, opened 2020-04-27
DEV: adds support for header-title in topicTitleDecorators

jjaffeux:


jjaffeux:
Two notes:
- will add tests for this tomorrow
- I noticed header-title is re-rendered two times in less than 25ms, I will investigate to see if I could avoid this in another PR

eviltrout:
So far so good.

jjaffeux:
@eviltrout I'm merging as is, writing a test was too complex.

PR #9563, opened 2020-04-28
FIX: Topic user bookmarked column logic was not correct

martin-brennan:
Make sure the topic_user.bookmarked column is set correctly when user bookmarks/unbookmarks any post in a topic. For example, you bookmarked a post in the topic that was not the OP, the bookmark icon in the topic list would not be shown. Same if deleting a bookmark for the last bookmarked post in a topic, the bookmark icon in the topic list would not be removed.

image

* Previously this was only setting it to true if bookmarking the OP/topic, which was not correct -- we want to show the icon on the topic list if any post is bookmarked.
* Also set to false if unbookmarking the last bookmarked post in the topic.

Also in this PR is a migration to correct any out of sync `topic_user.bookmarked` columns, based on the new logic.

### DIFF
--- a/db/migrate/20200428014005_correct_topic_user_bookmarked_boolean.rb
+++ b/db/migrate/20200428014005_correct_topic_user_bookmarked_boolean.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class CorrectTopicUserBookmarkedBoolean < ActiveRecord::Migration[6.0]
+  def up
+    # if the relation exists then we set to bookmarked because
+    # at least 1 bookmark for the user + topic exists
+    DB.exec(
+      <<~SQL
+        UPDATE topic_users SET bookmarked = true
+        FROM bookmarks AS b
+        WHERE topic_users.bookmarked IS false AND topic_users.topic_id = b.topic_id AND topic_users.user_id = b.user_id
+      SQL
+    )
+
+    DB.exec(
+      <<~SQL
+        UPDATE topic_users SET bookmarked = false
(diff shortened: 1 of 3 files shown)


db/migrate/20200428014005_correct_topic_user_bookmarked_boolean.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class CorrectTopicUserBookmarkedBoolean < ActiveRecord::Migration[6.0]
+  def up
+    # if the relation exists then we set to bookmarked because
+    # at least 1 bookmark for the user + topic exists
+    DB.exec(
+      <<~SQL
+        UPDATE topic_users SET bookmarked = true
+        FROM bookmarks AS b
+        WHERE topic_users.bookmarked IS false AND topic_users.topic_id = b.topic_id AND topic_users.user_id = b.user_id
```
SamSaffron, line 11: `IS false` feels a bit weird ... `NOT topic_users.bookmarked` ... for the `true` case `where topic_users.bookmarked`

SamSaffron:
Overall change looks good! Feel free to merge the fix after the minor style change.

spec/lib/bookmark_manager_spec.rb
```diff
@@ -134,6 +136,17 @@ def create_bookmark
       expect(result[:topic_bookmarked]).to eq(true)
     end
 
+    context "if the bookmark is the last one bookmarked in the topic" do
+      before do
+        TopicUser.create(user: user, topic: bookmark.post.topic, bookmarked: true)
```
SamSaffron, line 141: why put this in a before block just for one test?

martin-brennan: ...good point

PR #9566, opened 2020-04-28
FIX: Change bookmarks-with-reminders URL back to bookmarks for user activity

martin-brennan:
* Bookmarks with reminders is a core feature now, no need to have a separate URL
* Keep around the old /u/:username/activity/bookmarks-with-reminders route for backwards compat in Ember but just redirect to user activity bookmarks.


app/assets/javascripts/discourse/app/routes/user-activity-bookmarks-with-reminders.js
```diff
@@ -1,29 +1,7 @@
...
-    controller.set("model", model);
-    controller.loadItems();
-  },
-
-  actions: {
-    didTransition() {
-      this.controllerFor("user-activity")._showFooter();
-      return true;
-    }
+  redirect: function() {
```
jjaffeux, line 4:
```suggestion
  redirect() {
```

martin-brennan:
@eviltrout the new route is still there and redirects to /bookmarks, see @jjaffeux's review comment. Will keep it around for a while before nuking it.

PR #9568, opened 2020-04-28
FIX: Improve clearing store cache

CvX:
1. Shorter
2. Simpler
3. Doesn't depend on external binaries
4. Doesn't fail on large amounts of files
5. Hopefully eliminates flaky spec errors


davidtaylorhq:
Looks good - much cleaner than before 😍 

Looks like there's a small linting thing to clean up

PR #9570, opened 2020-04-28
FEATURE: allows to to style published page with themes/plugins

jjaffeux:


jjaffeux:
> Nice, this change was less complicated than I suspected it would be.

me too

app/controllers/published_pages_controller.rb
```diff
@@ -15,6 +15,18 @@ def show
     guardian.ensure_can_see!(pp.topic)
     @topic = pp.topic
     @canonical_url = @topic.url
+
+    @body_classes = [
```
eviltrout, line 19: If you make this a `Set.new` instead it will avoid duplicates which is nice to do defenstively.

jjaffeux:
Note I'm a little bit concerned, the Set might break the test as there's no order guarantees, but it seems reliable so far.

PR #9571, opened 2020-04-28
FEATURE: Add user_profile to user_archive CSV export

udan11:


app/jobs/regular/export_csv_file.rb
```diff
@@ -29,15 +30,22 @@ def execute(args)
...
-      else
-        "#{@entity.split('_').join('-')}-#{Time.now.strftime("%y%m%d-%H%M%S")}"
+      entities.each do |entity|
+        entity[:method] = :"#{entity[:name]}_export"
+        raise Discourse::InvalidParameters.new(:entity) unless respond_to?(entity[:method])
+
+        @timestamp ||= Time.now.strftime("%y%m%d-%H%M%S")
+        entity[:filename] =
+          if entity[:name] == "user_archive" || entity[:name] === "user_archive_profile"
+            "#{entity[:name].split('_').join('-')}-#{@current_user.username}-#{@timestamp}"
```
eviltrout, line 43: This would read a lot easier if you put `#{entity[:name].split('_').join('-')}` into a variable

udan11: That's right. I just moved the code and did not realize how weird that line is (why not `gsub` in the first place?). I am using `dasherize` now which is more succinct and hopefully does a better job at formatting.

PR #9573, opened 2020-04-28
FIX: Unassign user titles when a badge is deleted

riking:
Taking suggestions for improvements to where the code for `clear_titles_for_granted_badge` lives.

### DIFF
--- a/app/controllers/admin/badges_controller.rb
+++ b/app/controllers/admin/badges_controller.rb
@@ -129,10 +129,12 @@ class Admin::BadgesController < Admin::AdminController
   end
 
   def destroy
-    badge = find_badge
-    StaffActionLogger.new(current_user).log_badge_deletion(badge)
-    Jobs::BulkUserTitleUpdate.clear_titles_for_granted_badge(badge.id)
-    badge.destroy!
+    Badge.transaction do
+      badge = find_badge
+      StaffActionLogger.new(current_user).log_badge_deletion(badge)
+      badge.clear_user_titles!
+      badge.destroy!
+    end
     render body: nil
   end
 
--- a/app/controllers/admin/badges_controller.rb
+++ b/app/controllers/admin/badges_controller.rb
@@ -131,7 +131,8 @@ class Admin::BadgesController < Admin::AdminController
   def destroy
     badge = find_badge
     StaffActionLogger.new(current_user).log_badge_deletion(badge)
-    badge.destroy
+    Jobs::BulkUserTitleUpdate.clear_titles_for_granted_badge(badge.id)
+    badge.destroy!
     render body: nil
   end
 
--- a/app/jobs/regular/bulk_user_title_update.rb
+++ b/app/jobs/regular/bulk_user_title_update.rb
@@ -18,6 +18,23 @@ module Jobs
       end
     end
 
+    # Not called from the job because this needs to happen during the destroy request
+    def self.clear_titles_for_granted_badge(granted_badge_id)
+      ActiveRecord::Base.transaction do
+        DB.exec(<<~SQL, granted_title_badge_id: granted_badge_id, title: '', updated_at: Time.now)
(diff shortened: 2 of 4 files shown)


app/jobs/regular/bulk_user_title_update.rb
```diff
@@ -2,6 +2,7 @@
 
 module Jobs
   class BulkUserTitleUpdate < ::Jobs::Base
+    DESTROY_ACTION = 'destroy'.freeze
```
eviltrout, line 5: This new constant seems unused and should probably be removed.

app/controllers/admin/badges_controller.rb
```diff
@@ -131,7 +131,8 @@ def update
   def destroy
     badge = find_badge
     StaffActionLogger.new(current_user).log_badge_deletion(badge)
```
martin-brennan, line 133: I feel like these 3 calls should now be in a `Badge.transaction do` block now, it would be awkward if we cleared all the titles but then failed to destroy the badge. Should be all-or-nothing.

app/jobs/regular/bulk_user_title_update.rb
```diff
@@ -18,6 +19,23 @@ def execute(args)
       end
     end
 
+    # Not called from the job because this needs to happen during the destroy request
+    def self.clear_titles_for_granted_badge(granted_badge_id)
+      ActiveRecord::Base.transaction do
+        DB.exec(<<~SQL, granted_title_badge_id: granted_badge_id, title: '', updated_at: Time.now)
+          UPDATE users AS u
+          SET title = :title, updated_at = :updated_at
```
martin-brennan, line 27: Nitpick, but why pass in `title` if you are just setting it to `''`? `SET title = ''`

app/jobs/regular/bulk_user_title_update.rb
```diff
@@ -18,6 +19,23 @@ def execute(args)
...
+          FROM user_profiles AS up
+          WHERE up.user_id = u.id AND up.granted_title_badge_id = :granted_title_badge_id
+        SQL
+        DB.exec(<<~SQL, granted_title_badge_id: granted_badge_id, title: '', updated_at: Time.now)
+          UPDATE user_profiles AS up
+          SET badge_granted_title = false, granted_title_badge_id = NULL
```
martin-brennan, line 33: You are not setting the `updated_at` here

app/jobs/regular/bulk_user_title_update.rb
```diff
@@ -18,6 +19,23 @@ def execute(args)
       end
     end
 
+    # Not called from the job because this needs to happen during the destroy request
+    def self.clear_titles_for_granted_badge(granted_badge_id)
```
martin-brennan, line 23: I think this should just be a method on the `Badge` model rather than putting it in the job, seeing as how the job doesn't use it. Then you don't have to pass the id either, so something like:

```
badge.clear_granted_user_titles
```

The naming of the method is a little tricky because it's like "get rid of all the user titles that have been set based on me (the badge) being granted"

app/jobs/regular/bulk_user_title_update.rb
riking: user_profiles does not have an updated_at

martin-brennan: Ah OK, you can just remove it from the query params then

app/jobs/regular/bulk_user_title_update.rb
SamSaffron: Also a note ... all those .freeze calls are not needed since `# frozen_string_literal: true` will fix

app/jobs/regular/bulk_user_title_update.rb
riking: Done!

PR #9576, opened 2020-04-28
FEATURE: add category banner for why a user cannot post

featheredtoast:
Adds a category banner for why a user is unable to post in a category.

Additionally adds the same message to the disabled new topic button's title

Screenshot from 2020-04-28 15-47-36

Screenshot from 2020-04-28 15-53-44


coding-horror:
Why not also have it appear when you click the button as an error-ish
dialog?

featheredtoast:
The new topic button is disabled if they don't have permissions, so clicking on it wouldn't do anything. Between the banner and the title, I consider those as sufficient nuclear options for forums looking to tack up "Look, you really can't post here yet" signs

coding-horror:
Right, I'm saying enable it, then clicking or tapping it also shows that
same text? For extra teachings!

featheredtoast:
Ah, gotcha, I misunderstood - we can do that!

app/assets/javascripts/discourse/app/templates/discovery.hbs
```diff
@@ -3,6 +3,7 @@
 {{else}}
   <div class="container">
     {{discourse-banner user=currentUser banner=site.banner}}
+    {{category-read-only-banner category=category readOnly=navigationCategory.cannotCreateTopicOnCategory user=currentUser}}
```
eviltrout, line 6: You probably don't need to pass `currentUser` in here, as all components have access to `this.currentUser`.

PR #9578, opened 2020-04-29
DEV: Make `discourse-common` an Ember addon.

eviltrout:
This is to help with the migration to Ember CLI. In the current running
version of Discourse everything should be the same as before, just with
a few extra files that are not used. However, using Ember CLI this can
be installed as an Ember addon.


.git-blame-ignore-revs
```diff
@@ -23,9 +23,6 @@ acc5cbdf8ecb9293a0fa9474ee73baf499c02428
 # Rename wizard from es6 -> js
 1ac02422011f89716ab27250d39b0e0212e03892
 
-# Rename discourse-common es6 -> js
```
eviltrout, line 26: This is pointless now!

package.json
```diff
@@ -4,7 +4,7 @@
   "main": "index.js",
   "repository": "git@github.com:discourse/discourse.git",
   "author": "Discourse",
-  "license": "MIT",
+  "license": "GPLv2",
```
eviltrout, line 7: I noticed discourse had MIT in our package.json so I took the opportunity to correct it.

CvX: I see it generates a warning:

> warning package.json: License should be a valid SPDX license expression

Looking at the list (https://spdx.org/licenses/), it apparently should be:

```suggestion
  "license": "GPL-2.0-only",
```

(or `GPL-2.0-or-later`)

eviltrout: Weird I literally got that string from a list of valid values! I guess it was wrong or outdated.

PR #9579, opened 2020-04-29
FIX: Rename all instances of bookmarkWithReminder to just bookmark

martin-brennan:
* Rename all instances of bookmarkWithReminder and bookmark_with_reminder to just bookmark
* Delete old bookmark code at the same time
* Add migration to remove the bookmarkWithReminder post menu item if people have it set in site settings


db/migrate/20200429045956_remove_bookmarks_with_reminder_post_menu_item.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class RemoveBookmarksWithReminderPostMenuItem < ActiveRecord::Migration[6.0]
+  def up
+    post_menu = SiteSetting.post_menu
```
eviltrout, line 5: We need to use manual SQL in migrations rather than Ruby objects.

Can you adjust this to update the value in the DB instead?

martin-brennan: Argh sorry @eviltrout this is just an ingrained habit, I keep forgetting. Will change before merging.

PR #9580, opened 2020-04-29
FIX: discourse-presence breaks composer for users that hide presence.

tgxworld:


tgxworld:
Merged in https://github.com/discourse/discourse/commit/ae54a337195db3afc4447e81e5231a3e2f41ac1a

PR #9587, opened 2020-04-29
FEATURE: Allow user creation with admin api when local logins disabled

davidtaylorhq:


SamSaffron:
Cool edge case seems good to me

PR #9594, opened 2020-04-30
DEV: Refactor presence manager to deal with multiple composer states.

tgxworld:


plugins/discourse-presence/assets/javascripts/discourse/lib/presence-manager.js.es6
```diff
@@ -200,4 +214,73 @@ const PresenceManager = EmberObject.extend({
...
+  init() {
+    this._super(...arguments);
+
+    this.setProperties({
+      presences: {}
+    });
+  },
+
+  subscribe(topicId, type) {
+    if (!topicId) return;
```
tgxworld, line 232: Not sure how I can get around having to litter this every where to handle the case where the composer does not have a `topic.id`. Mainly when creating new topics.

plugins/discourse-presence/assets/javascripts/discourse/initializers/discourse-presence.js.es6
```diff
@@ -0,0 +1,32 @@
+import { withPluginApi } from "discourse/lib/plugin-api";
+import PresenceManager from "../lib/presence-manager";
+import ENV from "discourse-common/config/environment";
+
+function initializeDiscoursePresence(api, { app }) {
+  const currentUser = api.getCurrentUser();
+
+  if (currentUser) {
+    app.register(
+      "presence-manager:main",
```
davidtaylorhq, line 10: I wonder if this would be better as an Ember service? That seems to be the most 'ember-y' way to create a singleton, and it's how I handled a [similar problem in discourse-whos-online](https://github.com/discourse/discourse-whos-online/blob/master/assets/javascripts/discourse/services/online-service.js.es6). Then you can [inject the service](https://github.com/discourse/discourse-whos-online/blob/master/assets/javascripts/discourse/components/whos-online.js.es6#L17) into components as needed. That would avoid the need for `getOwner(this).lookup("presence-manager:main"))`

SamSaffron: totally support changing this, feel free to make the refactor, it is a bit fiddly from what I can tell and I really want to get these fixes in today so I will merge this anyway once I test it.

tgxworld: @davidtaylorhq what are the differences between factories and services? I was mainly following https://guides.emberjs.com/release/applications/dependency-injection/ to register an singleton with the application.

SamSaffron:
OK merged this in by hand, thanks @tgxworld  really happy with this now ... we can clean up the factory stuff next week I guess, but it is not urgent.

plugins/discourse-presence/assets/javascripts/discourse/initializers/discourse-presence.js.es6
davidtaylorhq: @tgxworld I think Services are the 'easy-mode' for singletons. Factories are 'hard-mode', but more flexible. From the link you shared

> Generally, Services are Ember's primary method for sharing state via dependency injection. In most cases, you shouldn't need to learn about how to work with Ember's DI system directly, or how to manually register and setup dependencies. 

Nothing wrong with factories though, so this is totally fine to leave as-is for now.

tgxworld: Thank you for the explanation :+1: Do you happen to know how to initialize the service with attributes? I was looking at https://github.com/discourse/discourse-whos-online/blob/master/assets/javascripts/discourse/services/online-service.js.es6 and it seems to be grabbing `messageBus` and `currentUser` from globals and it isn't clear from the [Ember docs](https://guides.emberjs.com/release/services/) whether that is possible

davidtaylorhq: Hmm you're right - the global lookup is not pretty! I guess it should be possible to use `inject()` there rather than a global lookup.

Maybe we should be adding currentUser and messageBus to services [via inject-discourse-objects](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/app/pre-initializers/inject-discourse-objects.js#L13). 🤔

tgxworld: O yea we should probably do that. TBH though, I do feel that services and factories are more or less the same as least in the context of how I'm using it :grin: Replacing `getOwner(this).lookup("presence-manager:main"))` with `presenceManager: inject()` and not having to register the factory manually wasn't that painful. 

Any thoughts about this @eviltrout? I feel like I might be missing something on why directly registering factories can be bad?

davidtaylorhq: > I do feel that services and factories are more or less the same as least in the context of how I'm using it 😁

Agreed, nothing bad about it. In a perfect world I think siteSettings and appEvents would also be ember services. I even [found a TODO for it](https://github.com/discourse/discourse/blob/a51b8d9c6604fa0a1ff4bf07aabbb1bdce7c4b51/app/assets/javascripts/admin/services/admin-tools.js#L18-L18) ;)

eviltrout: `appEvents` and `currentUser` should absolutely be services too and I would very much like to fix that eventually.

@tgxworld we should be minimizing `getOwner` and `lookup` - those APIs are private-ish and are discouraged in the community unless you absolutely have to use them. Over time I have been trying to remove as many as possible outside of initializers.

For now, we should go with @davidtaylorhq's suggestion of injecting `currentUser` and `appEvents` into services in the initializer that does the other injecting automatically. That way it keeps the `lookup` apis in one place and not in various objects spread throughout our project.

PR #9595, opened 2020-04-30
UX: display info about extra files and uploader

jjaffeux:
A component with not extra files but uploaded:
Screenshot 2020-04-30 at 13 17 04

A component with extra files from a remote source:
Screenshot 2020-04-30 at 13 17 22

A component with extra files with a remote source:
Screenshot 2020-04-30 at 13 17 35


config/locales/client.en.yml
```diff
@@ -3759,6 +3759,9 @@ en:
           settings: "Settings"
           translations: "Translations"
           extra_scss: "Extra SCSS"
+          extra_files: "Extra files"
+          extra_files_upload: "Use Export to access these files."
+          extra_files_remote: "Use Export or download the source to access these files."
```
davidtaylorhq, line 3764:
```suggestion
          extra_files_upload: "Export theme to view these files."
          extra_files_remote: "Export theme or check the git repository to view these files."
```

app/assets/javascripts/admin/templates/customize-themes-show.hbs
```diff
@@ -126,6 +126,15 @@
         </span>
       {{/if}}
     </div>
+  {{else}}
+    <div class="control-unit">
+      <span class="heading">{{i18n "admin.customize.theme.uploader"}}</span>
+      <span>
+        {{#user-link user=model.user}}
```
davidtaylorhq, line 133: This is the original creator, right? Someone else [could upload another version of the theme later](https://github.com/discourse/discourse/blob/501b19b6e005f00ffbaa126965b80bad3eaca47b/app/models/remote_theme.rb#L39) (normally via theme cli)

Should we change this to "Created by {{user}}"?

jjaffeux: Yes wasn't sure about this... Guess created is slightly better, will do the change :+1:

PR #9598, opened 2020-04-30
FEATURE: allows to limit visible reports and tabs in dashboard

jjaffeux:


eviltrout:
I like the idea here but that setting is enormous, plus new reports have to remember to add themselves to it when they are created. Maybe we should have a setting to hide reports instead?

jjaffeux:
> I like the idea here but that setting is enormous, plus new reports have to remember to add themselves to it when they are created. Maybe we should have a setting to hide reports instead?

And I would give suggestion? You would have to type what you think is a right report name? If you are ok with this then yes it will be shorter and probably good enough.

jjaffeux:
@eviltrout I went ahead and inverted the logic in the last commit, I think you are right this is the way to go. Will be much more future proof.

Screenshot 2020-04-30 at 16 41 25

eviltrout:
@jjaffeux yes I would vastly prefer if you had to type it exactly rather than having a huge input like that! Also new reports have less boilerplate. I think it's a win.

I think most people will not hide reports so it's okay if the experience is not perfect.

app/assets/javascripts/admin/controllers/admin-dashboard-general.js
```diff
@@ -23,9 +23,44 @@ export default Controller.extend(PeriodComputationMixin, {
...
+      .split("|")
+      .filter(Boolean);
+  }),
+
+  isActivityMetricsVisible: computed(
+    "activityMetrics",
+    "hiddenReports",
+    function() {
+      return (
+        this.activityMetrics.length > 0 &&
```
jjaffeux, line 40:
```suggestion
        this.activityMetrics.length &&
```

app/assets/javascripts/admin/controllers/admin-dashboard-general.js
```diff
@@ -23,9 +23,44 @@ export default Controller.extend(PeriodComputationMixin, {
...
+  isCommunityHealthVisible: computed("hiddenReports", function() {
+    return [
+      "consolidated_page_views",
+      "signups",
+      "topics",
+      "posts",
+      "dau_by_mau",
+      "daily_engaged_users",
+      "new_contributors"
+    ].filter(x => !this.hiddenReports.includes(x)).length;
```
eviltrout, line 61: I would recommend using `.some` for this. No point looping through them all if you only care if one is visible.

PR #9609, opened 2020-05-01
Fix theme concurrency issues

davidtaylorhq:
(see commit messages)


davidtaylorhq:
Added commits to master manually

PR #9612, opened 2020-05-02
DEV: api documentation updates

oblakeerickson:
- Created a script to convert json responses to rswag
- Documented several api endpoints
- Switched rswag to use header based auth


script/json_to_rswag.rb
```diff
@@ -0,0 +1,102 @@
...
+#   success: { type: :boolean },
+#   active: { type: :boolean },
+#   message: { type: :string },
+#   user_id: { type: :integer },
+# }
+
+class JsonToRswag
+
+  def initialize
+    @s = " "
```
jjaffeux, line 25: what's "s" here ? I think it would deserve a more explicit name

jjaffeux:
Does it fail if the request response has a new property?

PR #9615, opened 2020-05-02
DEV: Use `response.parsed_body` in specs

CvX:
Autofixed with rubocop-discourse 2.1.1.


eviltrout:
Excellent change. It does look like specs are failing though and might be legit.

PR #9616, opened 2020-05-02
DEV: Update rubocop-discourse

CvX:
Our rubocop configuration moved there.

Since `rubocop` and `rubocop-rspec` are dependencies of `rubocop-discourse`, **do we want to keep them as explicit `discourse` dependencies?** This should affect only dependabot. For now I removed them from Gemfile but can restore those lines before merging.


eviltrout:
> do we want to keep them as explicit discourse dependencies?

I don't think it's necessary.

PR #9617, opened 2020-05-02
DEV: Update rspec-rails from 4.0.0.beta2 to 4.0.0

ryanwi:
Upgrades to rspec-rails 4.0.0 release.


SamSaffron:
Apologies Ryan ... looks like a CLA is missing, I went ahead and upgrades this gem by hand without looking at the source code change you made.

ryanwi:
No problem, thanks for the update and for updating.

PR #9623, opened 2020-05-04
REFACTOR: minor tweaks to user's notifications-filter

jjaffeux:


Ahmedgagan:
‘ensure we only show filter if we have notifications’
@jjaffeux the issue here comes when there are no notifications, and filter doesn’t shows, then we can’t change the filter and notifications page becomes empty.

Like this:
image

So, when filter is hidden we cannot change the filter and get the notifications which are available.
And this causes issue.

Please have a look.
Thanks

jjaffeux:
Hmm yess I see. I think the name "hasNotifications" is not good. Because this is not what it means, it actually means "hasNoFilteredNotifications".

We have notifications, but not for this filter.

Brought this back in https://github.com/discourse/discourse/commit/5285cbee62ad7b0cee5b3bf0b6449174fb4f4491 thanks

PR #9627, opened 2020-05-04
DEV: Add remove button function to PluginAPI

zcuric:
### This PR
- [x] Introduces `removeButton` function to `post-menu` widget;
- [x] Introduces `removePostMenuButton` function to PluginAPI to complement `addPostMenuButton`
- [x] Adds tests for both `addButton` and `removeButton` functions in the `post-menu-test`

### DIFF
--- a/app/assets/javascripts/discourse/app/lib/plugin-api.js
+++ b/app/assets/javascripts/discourse/app/lib/plugin-api.js
@@ -5,7 +5,7 @@ import { addPluginOutletDecorator } from "discourse/components/plugin-connector"
 import { addTopicTitleDecorator } from "discourse/components/topic-title";
 import ComposerEditor from "discourse/components/composer-editor";
 import DiscourseBanner from "discourse/components/discourse-banner";
-import { addButton } from "discourse/widgets/post-menu";
+import { addButton, removeButton } from "discourse/widgets/post-menu";
 import { includeAttributes } from "discourse/lib/transform-post";
 import { registerHighlightJSLanguage } from "discourse/lib/highlight-syntax";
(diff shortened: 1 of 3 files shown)


eviltrout:
You probably noticed some tests are failing. The issue here is probably that the buttons you are removing are not being added back for tests that need them. Between tests you will need to reset the state of the buttons for this to work.

zcuric:
@eviltrout I tried to that in `afterEach`, but no success. Any example in the code for that kind of situation?

Btw, is this useful?

zcuric:
@eviltrout I've ended removing the `coffee` button in previous test, as I assume, if `admin` button remains remove,  `coffee` button should  remains there when added and removed when removed.

eviltrout:
Thanks @zcuric it looks like tests are passing now. Now in terms of usefulness, presumably you had a use case before creating this? Were you trying to remove a core or extra button?

zcuric:
@eviltrout My particular use case is to replace "native" buttons with custom ones. Custom share, instead of regular share etc. For example: I need to make sure that share button only apears on first post (main post), not on replies and buttons need to have labels also. So instead of rewriting the whole `post-menu` widget, my goal was to remove existing buttons and add new ones, with basically same functionality. 

Also, I have to override button from discourse-solved plugin. 

If I had more time, I would try to extend plugin API for buttons.  

Btw, post menu widget is too large, and it's html function could be refactor. 

I have some strange JIRA tickets 😅 . I'll show you someday what I did, when it get's live 😄

eviltrout:
This seems good! Thanks for your hard work :)

zcuric:
@eviltrout Oh, well, thank you!

PR #9630, opened 2020-05-04
DEV: Rename `Discourse.RAW_TEMPLATES`

eviltrout:
We were sharing `Discourse` both as an application object and a
namespace which complicated things for Ember CLI. This patch
moves raw templates into `__DISCOURSE_RAW_TEMPLATES` and adds
a couple helper methods to create/remove them.


app/assets/javascripts/discourse-common/addon/lib/raw-templates.js
```diff
@@ -0,0 +1,38 @@
+import { getResolverOption } from "discourse-common/resolver";
+
+export const __DISCOURSE_RAW_TEMPLATES = {};
```
jjaffeux, line 3: should we reset this const between tests ?

eviltrout: If we reset it all tests will fail! Those templates are not reset.

PR #9641, opened 2020-05-05
DEV: Add show_created_at option for invite_serializer

markvanlan:
In a plugin, I want to show when an invite was created. This allows me the ability to do that.


eviltrout:
Wouldn't it be easier for your plugin to use `add_to_serializer` for this?

markvanlan:
@eviltrout yyeeeep.

PR #9643, opened 2020-05-05
UX: don't disable "create account" button & display error message for required fields.

vinothkannans:


app/assets/javascripts/discourse/app/components/user-field.js
```diff
@@ -1,11 +1,33 @@
...
 import Component from "@ember/component";
 import { fmt } from "discourse/lib/computed";
+import { action } from "@ember/object";
 
 export default Component.extend({
   classNameBindings: [":user-field", "field.field_type", "customFieldClass"],
   layoutName: fmt("field.field_type", "components/user-fields/%@"),
 
+  didInsertElement() {
+    this.field.component = this;
```
jjaffeux, line 11: this._super(...arguments);

app/assets/javascripts/discourse/app/components/user-field.js
```diff
@@ -1,11 +1,33 @@
...
+  },
+
+  @action
+  focus() {
```
jjaffeux, line 15: We would need a more descriptive name for the action please, it's doesn't clash with native ember events (focusIn, focusOut) but still it would be safer.

app/assets/javascripts/discourse/app/components/user-field.js
```diff
@@ -1,11 +1,33 @@
...
+    if (this.element.querySelector("input")) {
```
jjaffeux, line 16: would be worth storing the element lookup

app/assets/javascripts/discourse/app/components/user-field.js
jjaffeux, line 11: I don't know much all this code, but it looks like a smell to me to carry this component like this.

Also I see it's used as "element" later, when most of the times these elements are DOM nodes, and not Ember components.

app/assets/javascripts/discourse/app/components/user-field.js
vinothkannans: I'm using the name `focus` since I have to call it [here](https://github.com/discourse/discourse/pull/9643/commits/962ef37108252288700981a3c7a89aba446a36b0#diff-092d88170d619c3d0d097005cfa089bbR319). I need to add one more if condition if I change it.

vinothkannans:
@jjaffeux the requested changes are now completed.

jjaffeux:
Can you show me what it does please ?

vinothkannans:
@jjaffeux like in the below screenshots, instead of disabling the "Create Account" button it will show the error message when the form submitted.

| Before | After |
| - | - |
| image | image |

PR #9646, opened 2020-05-05
Add script for compiling copyright deposits

kemitchell:
This small PR adds a script for compiling deposits for copyright registrations.


kemitchell:
Pushed.

PR #9647, opened 2020-05-06
DEV: Upgrade sinon and fix time based bookmark tests

martin-brennan:
Update sinon.js to 9.0.2 to access async fake timers https://sinonjs.org/releases/v9.0.2/fake-timers/ which can then be used with acceptance tests (previously `useFakeTimers` didn't work with await, e.g. for `visit`).

Fix the bookmark acceptance test that was time based to use these new fake timers.

Add a `fakeTime` function that uses moment and the provided date string + timezone to freeze time using `useFakeTimers` and return a `clock`.

Add a `timeStep` function that accepts a `clock` from `fakeTime` and a function to run. Once the function is run we call `clock.tickAsync(1000)` to progress the fake clock forward 1s to progress promises/callbacks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "puppeteer": "1.20",
     "qunit": "2.8.0",
     "route-recognizer": "^0.3.3",
-    "sinon": "^7.2.5"
+    "sinon": "^9.0.2"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


jjaffeux:
That’s awesome!!!

PR #9649, opened 2020-05-06
FIX: correct edit notification username for PMs

lis2:
When the tag is added to PM by admin, notification has an incorrect username (post author username instead of admin username)

### DIFF
--- a/app/services/post_alerter.rb
+++ b/app/services/post_alerter.rb
@@ -628,7 +628,9 @@ class PostAlerter
 
     each_user_in_batches(notify) do |user|
       notification_type = already_seen_user_ids.include?(user.id) ? Notification.types[:edited] : Notification.types[:posted]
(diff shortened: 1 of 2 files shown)


app/services/post_alerter.rb
```diff
@@ -628,7 +628,8 @@ def notify_post_users(post, notified, include_category_watchers: true, include_t
 
     each_user_in_batches(notify) do |user|
       notification_type = already_seen_user_ids.include?(user.id) ? Notification.types[:edited] : Notification.types[:posted]
-      create_notification(user, notification_type, post)
+      opts = notification_type == Notification.types[:edited] ? { display_username: post.last_editor.username } : {}
```
eviltrout, line 631: I think this would be more clear as two lines:

```
opts = {}
opts[:display_username] = post.last_editor_username if notification_type == Notification.types[:edited]
```

PR #9653, opened 2020-05-06
DEV: improves sk api

jjaffeux:
- reduces the API to 3 actions for now: appendContent/prependContent/onChange
- well tested
- removes all previous APIS which were only half supported or too dangerous as they could collide with other plugins or core behaviors
- this plugins also puts every sk test helpers in one file


jjaffeux:
@eviltrout a little bit of context for the review. I haven't actually updated the API when doing sk2, I wanted to see what people were actually using, from plugins, themes... My conclusion is that I have been proving a too vast API, so I have been reducing to three simple cases which should cover most use cases people have.

It shouldn't break any external plugin, I could add backward compat if needed.

PR #9658, opened 2020-05-06
FEATURE: Include category position when exporting categories

tshenry:
Previously, category exports excluded the `position` attribute. Depending on the number of categories being exported, this could result in a good amount of time wasted manually trying to match the original display order.


SamSaffron:
This seems fine to me! Thanks

PR #9660, opened 2020-05-07
DEV: Allow specifying if a notification is high_priority on create

martin-brennan:
This allows for special cases where we may not want a certain notification type to ALWAYS be high priority (e.g. a topic timer).

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -45,7 +45,9 @@ class Notification < ActiveRecord::Base
   end
 
   before_create do
-    self.high_priority = Notification.high_priority_types.include?(self.notification_type)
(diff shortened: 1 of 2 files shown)


app/models/notification.rb
```diff
@@ -45,7 +45,9 @@ class Notification < ActiveRecord::Base
   end
 
   before_create do
-    self.high_priority = Notification.high_priority_types.include?(self.notification_type)
+    # if we have manually set the notification to high_priority on create then
+    # make sure that is respected
+    self.high_priority = self.high_priority || Notification.high_priority_types.include?(self.notification_type)
```
SamSaffron, line 50: `self.high_priority ||= `

Feel free to merge once done :)

PR #9661, opened 2020-05-07
DEV: remove no-longer-necessary MULTISITE_PREFIX

andrewschleifer:
We prevent s3_backup_bucket from being the parent or the same as s3_upload_bucket -- spec/lib/site_settings/validations_spec.rb#validate_s3_upload_bucket -- so this looks like it's not useful anymore.

Currently, with s3_upload_bucket of "bucket_name" and s3_backup_bucket of "bucket_name/backups", multisite backups are actually ending up in "bucket_name/backups/backups/site_name".

Is my understanding correct @gschlager?
Is there anything else that needs to change?
How can we move existing backups?


andrewschleifer:
Aha. I think that prefix is to prevent pathologic behavior when the s3_backup_bucket is the same as s3_upload_bucket. It would be better to specifically prevent that.

PR #9686, opened 2020-05-07
REFACTOR: removes overwritten properties and duplicates fn

jjaffeux:


jjaffeux:
@eviltrout this one would probably require a slightly deeper check than the others

PR #9689, opened 2020-05-07
DEV: Move pretty-text into an ember-addon format

eviltrout:
The Ember CLI-ification of Discourse continues!


jjaffeux:
wow that's exciting!

PR #9691, opened 2020-05-07
FEATURE: Improve UX for user emails

udan11:


app/assets/javascripts/discourse/app/controllers/preferences/account.js
```diff
@@ -147,6 +164,14 @@ export default Controller.extend(CanCheckEmails, {
         .catch(popupAjaxError);
     },
 
+    setPrimaryEmail(email) {
+      this.model.setPrimaryEmail(email).then(() => {
+        this.model.secondary_emails.removeObject(email);
```
jjaffeux, line 169: shouldn't we do all of this this from inside `setPrimaryEmail` in models/user.js file given it's all model logic ?

app/assets/javascripts/discourse/app/components/email-dropdown.js
```diff
@@ -0,0 +1,63 @@
...
+        icon: "times",
+        name: I18n.t("user.email.destroy"),
+        description: ""
+      });
+    }
+
+    return content;
+  }),
+
+  actions: {
```
jjaffeux, line 46: if possible please use @action also maybe a more descriptive name like: `onChangeEmail`

app/assets/javascripts/discourse/app/components/email-dropdown.js
```diff
@@ -0,0 +1,63 @@
...
+    onChange(id) {
+      switch (id) {
+        case "updateEmail":
+          getOwner(this)
```
jjaffeux, line 50: I think you could get the router with a service, that would be better

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,49 @@ def check_emails
...
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
+
+    if user.primary_email.email == params[:email]
+      return render json: success_json
+    end
+
+    new_primary = user.user_emails.find_by(email: params[:email])
+    if new_primary.blank?
+      return render json: failed_json, status: 428
```
ZogStriP, line 224: Looks like we never used that HTTP status before. How does the UI handle that?

app/jobs/regular/user_email.rb
```diff
@@ -153,7 +153,7 @@ def message_for_email(user, post, type, notification, args = nil)
       raise Discourse::InvalidParameters.new("type=#{type}") unless UserNotifications.respond_to?(type)
 
       email_args[:email_token] = email_token if email_token.present?
-      email_args[:new_email] = user.email if type.to_s == "notify_old_email"
+      email_args[:new_email] = user.email if type.to_s == "notify_old_email" || type.to_s == "notify_old_email_add"
```
ZogStriP, line 156: What's the "type" of the `type` variable? Isn't it a symbol? If it is, when are we converting it to a string instead of checking directly against other symbols?

lib/email_updater.rb
```diff
@@ -35,54 +27,72 @@ def change_to(email_input)
       end
     end
 
-    if errors.blank? && existing_user.nil?
-      args = {
-        old_email: @user.email,
-        new_email: email,
-      }
+    return if !errors.blank? || existing_user.present?
```
ZogStriP, line 30:
```suggestion
    return if errors.present? || existing_user.present?
```

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,49 @@ def check_emails
     render json: failed_json, status: 403
   end
 
+  def update_primary_email
```
ZogStriP, line 212: I wonder if we should add an entry in the staff logs about this operation?

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,49 @@ def check_emails
...
+
+    User.transaction do
+      user.primary_email.update!(primary: false)
+      new_primary.update!(primary: true)
+    end
+
+    render json: success_json
+  end
+
+  def destroy_email
```
ZogStriP, line 235: I wonder if we should add an entry in the staff logs about this operation?

app/jobs/regular/user_email.rb
udan11: Sidekiq converts symbol arguments to strings, so it should be a string, unless the job is executed directly.

app/controllers/users_controller.rb
udan11: It should only happen if the user deletes the same email address. I will show a bootbox if this happens.

app/controllers/users_controller.rb
udan11: That sounds like a good idea.

app/controllers/users_controller.rb
udan11: That sounds like a good idea.

app/controllers/users_email_controller.rb
```diff
@@ -36,12 +36,18 @@ def update
...
 
     updater = EmailUpdater.new(guardian: guardian, user: user, initiating_user: current_user)
-    updater.change_to(params[:email])
+    updater.change_to(params[:email], add: request.post?)
 
     if updater.errors.present?
       return render_json_error(updater.errors.full_messages)
     end
 
+    if current_user.staff? && current_user != user
```
ZogStriP, line 45: Shouldn't this be done in the `EmailUpdater` class instead of the controller?

Also, we should not be logging the whole email address (cf. the `moderators_view_emails` site setting). I think just having a log that some staff member changed the email address of user X to be enough.

lib/email_updater.rb
```diff
@@ -35,78 +27,103 @@ def change_to(email_input)
       end
     end
 
-    if errors.blank? && existing_user.nil?
-      args = {
-        old_email: @user.email,
-        new_email: email,
-      }
+    return if errors.present? || existing_user.present?
```
ZogStriP, line 30: Can we do the `return if errors.present?` earlier (eg. right after we validate the email) so that we avoid doing any work if the email is not valid?

ZogStriP:
@tgxworld can you review this? With an "angle" on the security-side. 

I'd like to merge it before the end of the week and I'd love another pair of eyes.

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,62 @@ def check_emails
...
+
+    User.transaction do
+      old_primary.update!(primary: false)
+      new_primary.update!(primary: true)
+    end
+
+    if current_user.staff? && current_user != user
+      StaffActionLogger.new(current_user).log_update_email(user, previous_value: old_primary.email, new_value: new_primary.email)
+    else
+      UserHistory.create(action: UserHistory.actions[:update_email], target_user_id: user.id, previous_value: old_primary.email, new_value: new_primary.email)
```
tgxworld, line 236:
```suggestion
      UserHistory.create!(action: UserHistory.actions[:update_email], target_user_id: user.id, previous_value: old_primary.email, new_value: new_primary.email)
```

tgxworld: I feel like logging should be part of the same transaction as updating the primary status of the email address as well.

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,62 @@ def check_emails
...
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
+
+    user_email = user.user_emails.find_by(email: params[:email])
+    if user_email&.primary
+      return render json: failed_json, status: 428
+    end
+
+    if user_email
+      user.user_emails.where(email: params[:email]).destroy_all
```
tgxworld, line 254: Hmm why do we need to query for `user_email` again? It should be the same object as `user_email`?

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,62 @@ def check_emails
...
+    elsif
+      user.email_change_requests.where(new_email: params[:email]).destroy_all
+    end
+
+    if current_user.staff? && current_user != user
+      StaffActionLogger.new(current_user).log_destroy_email(user, previous_value: params[:email])
+    else
+      UserHistory.create(action: UserHistory.actions[:destroy_email], target_user_id: user.id, previous_value: params[:email])
```
tgxworld, line 262:
```suggestion
      UserHistory.create!(action: UserHistory.actions[:destroy_email], target_user_id: user.id, previous_value: params[:email])
```

tgxworld: Same comment about having this in the same transaction as deleting the email.

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,62 @@ def check_emails
...
+    end
+
+    render json: success_json
+  end
+
+  def destroy_email
+    params.require(:email)
+
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
```
tgxworld, line 246: We have to use `UserGuardian#can_edit_email?` instead.

app/controllers/users_controller.rb
```diff
@@ -209,6 +209,62 @@ def check_emails
     render json: failed_json, status: 403
   end
 
+  def update_primary_email
+    params.require(:email)
+
+    user = fetch_user_from_params
+    guardian.ensure_can_edit!(user)
```
tgxworld, line 216: We have to use `UserGuardian#can_edit_email?` instead.

app/controllers/users_controller.rb
tgxworld: Sorry actually we don't need this because primary email cannot be destroyed.

app/controllers/users_controller.rb
tgxworld: I think this is still valid because we can't allow the primary email to change if `SiteSetting.email_editable` is false or if `SiteSetting.sso_overrides_email` is true

app/services/staff_action_logger.rb
```diff
@@ -744,6 +744,33 @@ def log_unpublished_page(topic_id, slug)
     ))
   end
 
+  def log_add_email(user, opts = {})
```
tgxworld, line 747: This is minor but I think it is worth specifying the keys that needs to be passed in so that we always log with the right information.

lib/email_updater.rb
```diff
@@ -11,18 +11,10 @@ def initialize(guardian: nil, user: nil, initiating_user: nil)
     @initiating_user = initiating_user
```
tgxworld, line 11: @martin-brennan Do you recall why we have to pass in the `initiating_user` which can be `nil`? It doesn't seems right to me because the initiating use can always be retrieved via `guardian` and `EmailUpdater` does not allow any anonymous actions.

lib/email_updater.rb
```diff
@@ -35,78 +27,103 @@ def change_to(email_input)
...
+    old_email = @user.email if !add
 
-      if initiating_admin_changing_another_user_email? && !changing_staff_user_email?
-        auto_confirm_and_send_password_reset(email_token)
-        return
-      end
+    if @initiating_user&.admin? && !@user.staff? && @initiating_user != @user
+      send_email_notification(@user.email, email)
+      update_user_email(old_email, email)
+      send_email(:forgot_password, @user.email_tokens.create(email: @user.email))
```
tgxworld, line 37:
```suggestion
      send_email(:forgot_password, @user.email_tokens.create!(email: @user.email))
```

lib/email_updater.rb
```diff
@@ -35,78 +27,103 @@ def change_to(email_input)
...
+          end
+          update_user_email(change_req.old_email, change_req.new_email)
           confirm_result = :complete
         end
+
+        if confirm_result == :complete
+          if @initiating_user&.staff? && @initiating_user != @user
+            StaffActionLogger.new(@initiating_user).log_add_email(@user, previous_value: change_req.old_email, new_value: change_req.new_email)
+          else
+            UserHistory.create(action: UserHistory.actions[:add_email], target_user_id: @user.id, previous_value: change_req.old_email, new_value: change_req.new_email)
```
tgxworld, line 103:
```suggestion
            UserHistory.create!(action: UserHistory.actions[:add_email], target_user_id: @user.id, previous_value: change_req.old_email, new_value: change_req.new_email)
```

spec/serializers/web_hook_user_serializer_spec.rb
```diff
@@ -23,7 +23,7 @@
 
   it 'should only include the required keys' do
     count = serializer.as_json.keys.count
-    difference = count - 45
+    difference = count - 47
```
tgxworld, line 26: I don't think we should expose `unconfirmed_emails` to the web hook yet.

lib/email_updater.rb
martin-brennan: @tgxworld guess just a brain-fart on my part, I don't see any reason to keep `initiating_user` here if we can just do `guardian.user`

tgxworld: @martin-brennan Cool :+1: Let me see if I can remove it while not breaking any tests :P

ZogStriP:
Thanks @tgxworld for the review 👍 

Also @udan11 can you add a hidden site setting that will be used to toggle this feature (default off) so that we can merge it without fear or breaking anything (or if it does, we can always disable this using the site setting)

udan11:
I see there are some merge conflicts caused by @tgxworld's fixes discussed in the PR above. I wll resolve them later.

app/assets/javascripts/discourse/app/templates/preferences/account.hbs
```diff
@@ -46,12 +46,49 @@
   <div class="control-group pref-email">
     <label class="control-label">{{i18n "user.email.title"}}</label>
     {{#if model.email}}
-      <div class="controls">
-        <span class="static">{{model.email}}</span>
-        {{#if model.can_edit_email}}
-          {{#link-to "preferences.email" class="btn btn-default btn-small btn-icon pad-left no-text"}}{{d-icon "pencil-alt"}}{{/link-to}}
-        {{/if}}
-      </div>
+      {{#if siteSettings.enable_secondary_emails}}
```
ZogStriP, line 49: What happens if the site setting is disabled?

app/assets/javascripts/discourse/app/controllers/preferences/account.js
```diff
@@ -95,6 +96,36 @@ export default Controller.extend(CanCheckEmails, {
...
+    "model.email",
+    "model.secondary_emails.[]",
+    "model.unconfirmed_emails.[]"
+  )
+  emails(primaryEmail, secondaryEmails, unconfirmedEmails) {
+    const emails = [];
+
+    if (primaryEmail) {
+      const obj = { email: primaryEmail, primary: true, confirmed: true };
+      emails.push(EmberObject.create(obj));
```
ZogStriP, line 109:
```suggestion
      emails.push(EmberObject.create({ 
        email: primaryEmail, 
        primary: true, 
        confirmed: true
      }));
```

app/assets/javascripts/discourse/app/controllers/preferences/account.js
```diff
@@ -95,6 +96,36 @@ export default Controller.extend(CanCheckEmails, {
...
+    }
+
+    if (secondaryEmails) {
+      secondaryEmails.forEach(email => {
+        const obj = { email, confirmed: true };
+        emails.push(EmberObject.create(obj));
```
ZogStriP, line 115:
```suggestion
        emails.push(EmberObject.create({ email, confirmed: true }));
```

app/assets/javascripts/discourse/app/controllers/preferences/account.js
```diff
@@ -95,6 +96,36 @@ export default Controller.extend(CanCheckEmails, {
...
+      });
+    }
+
+    if (unconfirmedEmails) {
+      unconfirmedEmails.forEach(email => {
+        const obj = { email };
+        emails.push(EmberObject.create(obj));
```
ZogStriP, line 122:
```suggestion
        emails.push(EmberObject.create({ email }));
```

app/controllers/users_controller.rb
```diff
@@ -205,6 +205,72 @@ def check_emails
     render json: failed_json, status: 403
   end
 
+  def update_primary_email
+    if !SiteSetting.enable_secondary_emails
```
ZogStriP, line 209: Just to be 100% sure, when the site setting is disabled, people can still change their "primary" email address, right?

app/controllers/users_email_controller.rb
```diff
@@ -29,14 +29,18 @@ def index
   end
 
   def update
+    if !SiteSetting.enable_secondary_emails && request.post?
```
ZogStriP, line 32: I don't think we should be using the same route for both `POST` and `PUT`.

Usually, the `POST` is to add/create a new item whereas a `PUT` is to update a specific item (and the route usually has an `id`)

app/services/staff_action_logger.rb
ZogStriP: Yes, I agree, You can use named args here.

app/assets/javascripts/discourse/app/templates/preferences/account.hbs
udan11: It shows the old user interface.

app/controllers/users_controller.rb
udan11: Yes, they can. This endpoint is used just to switch between emails. To change the email address, `PUT /u/:username/preferences/email` is used instead.

app/services/staff_action_logger.rb
udan11: `opts` is no longer used after I removed `old_value` and `new_value` as they may contain sensitive user information (email addresses).

lib/email_updater.rb
udan11: I can do that too.

app/assets/javascripts/discourse/app/templates/preferences/account.hbs
ZogStriP: Nevermind I missed the `{{else}}` 🙈

app/controllers/users_controller.rb
```diff
@@ -205,6 +205,72 @@ def check_emails
     render json: failed_json, status: 403
   end
 
+  def update_primary_email
```
ZogStriP, line 208: Should definitely be tested 😉

app/controllers/users_controller.rb
```diff
@@ -205,6 +205,72 @@ def check_emails
...
+        StaffActionLogger.new(current_user).log_update_email(user)
+      else
+        UserHistory.create!(action: UserHistory.actions[:update_email], target_user_id: user.id)
+      end
+    end
+
+    render json: success_json
+  end
+
+  def destroy_email
```
ZogStriP, line 242: Should definitely be tested 😉

app/controllers/users_email_controller.rb
```diff
@@ -28,12 +28,35 @@ class UsersEmailController < ApplicationController
   def index
   end
 
+  def create
```
ZogStriP, line 31: Should definitely be tested 😉

PR #9703, opened 2020-05-08
DEV: calling preload-store is deprecated

jjaffeux:


eviltrout:
Oh weird! How were tests working?!!?

jjaffeux:
Well I think it was called, but issuing a deprecation message

PR #9706, opened 2020-05-08
FIX: Adjust margin on bookmark advanced btn

jordanvidrine:


app/assets/stylesheets/common/components/bookmark-modal.scss
```diff
@@ -71,6 +71,7 @@
 
   .bookmark-options-button {
     margin-left: 0.5em;
+    margin-bottom: 9px;
```
jjaffeux, line 74: can we use 1em or 0.5em ? I'm not a big fan of magic number like "9"

jordanvidrine: I think we can, I did it to match the `margin-bottom` of the input that is to it's left. Otherwise it was not centered

PR #9710, opened 2020-05-08
DEV: Add theme name as class

markvanlan:
This is useful for plugins to affect specific themes. It is certainly an edge case, and in general we do not want this to happen. This gives the flexibility to do so, and could be helpful for future projects.

For the material design theme for example the body looks like this:

Screenshot from 2020-05-08 11-27-21


davidtaylorhq:
I'm not too sure about this... users can change the name of themes whenever they like - they will not necessarily match the name in the git repository. What use-case do you have in mind?

PR #9716, opened 2020-05-08
DEV: allows to define an ariaLabel on d-button

jjaffeux:
This topic also adds this function to topic-footer-buttons, simplifies the whole logic of titile/label/arialabel in d-button and adds tests for these properties.

### DIFF
--- a/test/javascripts/components/d-button-test.js
+++ b/test/javascripts/components/d-button-test.js
@@ -105,7 +105,7 @@ componentTest("aria-label", {
   template:
     "{{d-button ariaLabel=ariaLabel translatedAriaLabel=translatedAriaLabel}}",
 
-  beforEach() {
+  beforeEach() {
     I18n.translations[I18n.locale].js.test = { fooAriaLabel: "foo" };
   },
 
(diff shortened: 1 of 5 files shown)


test/javascripts/components/d-button-test.js
```diff
@@ -100,3 +100,76 @@ componentTest("disabled button", {
...
+    });
+
+    assert.equal(find("button")[0].getAttribute("title"), "bar");
+  }
+});
+
+componentTest("label", {
+  template: "{{d-button label=label translatedLabel=translatedLabel}}",
+
+  beforEach() {
```
eviltrout, line 156: This is probably not working :)

test/javascripts/components/d-button-test.js
```diff
@@ -100,3 +100,76 @@ componentTest("disabled button", {
...
+    });
+
+    assert.equal(find("button")[0].getAttribute("aria-label"), "bar");
+  }
+});
+
+componentTest("title", {
+  template: "{{d-button title=title translatedTitle=translatedTitle}}",
+
+  beforEach() {
```
eviltrout, line 132: Same typo here :)

jjaffeux:
@eviltrout haha nice catch thx :p it was not breaking the test, because if I don't set the locale it will just be `[en_US.test.fooTitle]`  instead of the expected translation.

PR #9720, opened 2020-05-09
FIX: prevents rendering topic-category if empty

jjaffeux:


jjaffeux:
@udan11 for context https://meta.discourse.org/t/missing-field-itemlistelement/150747

PR #9731, opened 2020-05-10
FIX: randomize file name when created from fixtures

lis2:
When a temporary file is created from fixtures it should have a unique name.
It is to prevent a collision in parallel specs evaluation

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -238,7 +238,9 @@ RSpec.configure do |config|
   end
 
   config.after(:suite) do
-    $files_from_fixtures.each { |tmp_file_path| FileUtils.rm(tmp_file_path) }
+    if SpecSecureRandom.value
+      FileUtils.remove_dir(file_from_fixtures_tmp_folder, true)
+    end
   end
 
   config.before :each, &TestSetup.method(:test_setup)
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -369,8 +373,11 @@ def unfreeze_time
 
 def file_from_fixtures(filename, directory = "images")
   FileUtils.mkdir_p("#{Rails.root}/tmp/spec") unless Dir.exists?("#{Rails.root}/tmp/spec")
-  FileUtils.cp("#{Rails.root}/spec/fixtures/#{directory}/#{filename}", "#{Rails.root}/tmp/spec/#{filename}")
-  File.new("#{Rails.root}/tmp/spec/#{filename}")
+  tmp_file_path = "#{Rails.root}/tmp/spec/#{SecureRandom.hex << filename}"
+  FileUtils.cp("#{Rails.root}/spec/fixtures/#{directory}/#{filename}", tmp_file_path)
+  $files_from_fixtures ||= []
+  $files_from_fixtures << tmp_file_path
+  File.new(tmp_file_path)
```
eviltrout, line 380: Are these never cleaned up? I know developers will often clear their tmp folders but I feel like that should be a last resort and not the default behaviour.

It also seems worse now because previously you'd only get one file per spec and now you'll get one every time the test runs!

lis2: You are right, in this scenario, we will create around 200 files for the whole suite - that's why I was thinking that cleaning afterwards would be a good idea.

The problem I am trying to solve is a random failure for parallel specs. When it was same file name from time to time one spec wanted to access file which was already deleted by another spec. 

I can make an adjustment and leave `file_from_fixtures` untouched and defined additional helper called `temporary_file_from_fixtures` and use that for specs which have a destructive effect. In that case, fewer files would be created. Do you think it would be a better approach?

eviltrout: I don't mind if temporary files are created by specs as long as they are cleaned up. There is a ruby API that creates a tmp file and yields a block:

https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-create

I understand in this case you want to copy a file so that API might not work but something similar would be great. You can yield a block for the tests and it's automatically cleaned up afterwards.

lis2: I couldn't find a good way to use TempFile here. I talked with @SamSaffron  today, and we concluded that we can store those files in `File.join(Dir.tmpdir, "rspec_#{Process.pid}_#{SecureRandom.hex}")` and remove the whole directory afterwards

PR #9734, opened 2020-05-11
FIX: CurrentUser now must be passed to resolveTimezone and user card local time issues

martin-brennan:
* This is to prevent user's timezones being changed accidentally
  e.g. by admin looking at a user
* This problem only occurred via the user card, however the user card
  was still calling userTimezone even if the setting to display user
  time in card was disabled


curiousdannii:
Oh, good catch. It looks like something that would be very easy to miss next time, so what about a console warning if you call `resolvedTimezone()` without passing in a user?

PR #9736, opened 2020-05-11
FIX: add noindex header to badges, groups, and /my pages

jomaxro:


SamSaffron:
Looks good to me! Thanks

PR #9738, opened 2020-05-11
FIX: default canonical URL

lis2:
As a fallback, we should always have a canonical URL defined as base_url + path


SamSaffron:
Looks good to me, let's give it a shot.

PR #9741, opened 2020-05-11
UX: Fix hard coded value in Crazy in Love badge description

rukku:
Currently, the Crazy in Love badge's description has a hard coded
value of 50.This should correspond to the max_likes_per_day value
instead.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -4373,7 +4373,7 @@ en:
         This badge is granted when you use all %{max_likes_per_day} of your daily likes for 5 days. Thanks for taking the time actively encouraging the best conversations every day!
     crazy_in_love:
       name: Crazy in Love
-      description: Used 50 likes in a day 20 times
+      description: Used %{max_likes_per_day} likes in a day 20 times
       long_description: |
         This badge is granted when you use all %{max_likes_per_day} of your daily likes for 20 days. Wow! You’re a role model for encouraging your fellow community members!
     thank_you:


config/locales/server.en.yml
```diff
@@ -4373,7 +4373,7 @@ en:
         This badge is granted when you use all %{max_likes_per_day} of your daily likes for 5 days. Thanks for taking the time actively encouraging the best conversations every day!
     crazy_in_love:
       name: Crazy in Love
-      description: Used 50 likes in a day 20 times
+      description: Used %{max_likes_per_day} likes in a day 20 times
```
tgxworld, line 4376: Where is this translation called from? Don't we have to pass in the `max_likes_per_day` when fetching the translation?

tgxworld:
Thank you for contributing @rukku

config/locales/server.en.yml
rukku: Hi @tgxworld, I'm new to this codebase. I naively assumed that they all inherit from the English locale. Upon checking, this has been hardcoded on the other locales as well. For example, the es and it locales have it too

https://github.com/discourse/discourse/blob/16137308b011146be29f88c3856b154d55d04abf/config/locales/server.es.yml#L3941

https://github.com/discourse/discourse/blob/16137308b011146be29f88c3856b154d55d04abf/config/locales/server.it.yml#L3464

I'll have to read up more on how translations are done here.

rukku:
> Thank you for contributing @rukku

Thanks for reviewing @tgxworld.Just realized that this is bigger than I initially thought 😲 I thought I could just fix it in one place 😅 Would you know where I can bring up errors in the copy of the site text? I'm assuming they'll have to fix the reference and then it will have to be propagated to the translators via Transifex or whatever tool is being used now for translations.

config/locales/server.en.yml
rukku: Got this response on Discourse

https://meta.discourse.org/t/crazy-in-love-badge-description-not-reflecting-max-likes-per-day-setting/151127/2?u=rkda

It appears this will be the reference for all the other locales.

tgxworld:
Looks like your PR will work as it is a spot that was missed in https://github.com/tgxworld/discourse/commit/9e4fcb1a0a3cbb0061d6b1f56f0260d44883cd8e

PR #9743, opened 2020-05-11
DEV: Allow plugins to add extra icons to category-link

markvanlan:
Screenshot from 2020-05-11 13-15-01

This allows for plugins to check a property on a category, add additional icons to the category-link html

### DIFF
--- a/app/assets/javascripts/discourse/app/helpers/category-link.js
+++ b/app/assets/javascripts/discourse/app/helpers/category-link.js
@@ -151,8 +151,8 @@ function defaultCategoryLinkRenderer(category, opts) {
   if (restricted) {
     html += iconHTML("lock");
   }
-  Category.extraBadgeIconRenderers.forEach(function(appender) {
-    const icon = appender(category);
+  Category.extraBadgeIconRenderers.forEach(function(renderer) {
+    const icon = renderer(category);
     if (icon) {
       html += icon;
     }
--- a/app/assets/javascripts/discourse/app/helpers/category-link.js
+++ b/app/assets/javascripts/discourse/app/helpers/category-link.js
@@ -1,4 +1,4 @@
-import { get } from "@ember/object";
+import EmberObject, { get } from "@ember/object";
 import { registerUnbound } from "discourse-common/lib/helpers";
 import { isRTL } from "discourse/lib/text-direction";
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/discourse/app/models/category.js
```diff
@@ -510,6 +510,11 @@ Category.reopenClass({
     return _.sortBy(data, category => {
       return category.get("read_restricted");
     });
+  },
+
+  extraBadgeIconRenderers: [],
```
eviltrout, line 515: If these are only used in the helper I'd recommend moving them into there, and adding a `plugin-api` method to add one.

markvanlan: Oh, I didnt think about adding the actual array to the PluginApi. Okay, great!

app/assets/javascripts/discourse/app/helpers/category-link.js
```diff
@@ -7,6 +7,12 @@ import Site from "discourse/models/site";
 import { escapeExpression } from "discourse/lib/utilities";
 import { htmlSafe } from "@ember/template";
 
+const CategoryLink = EmberObject.extend({});
```
eviltrout, line 10: It doesn't need to be a class does it? The pattern we normally use here is to create a local instance variable:

`let _extraIconRenderers = []` 

It'll be scoped to this module.

Then you can export a function to add a new one to that local variable.

markvanlan:
@eviltrout Third time is the charm here I think

app/assets/javascripts/discourse/app/helpers/category-link.js
```diff
@@ -149,12 +155,15 @@ function defaultCategoryLinkRenderer(category, opts) {
...
 
   if (restricted) {
-    html += `${iconHTML(
-      "lock"
-    )}<span class="category-name" ${categoryDir}>${categoryName}</span>`;
-  } else {
-    html += `<span class="category-name" ${categoryDir}>${categoryName}</span>`;
+    html += iconHTML("lock");
   }
+  _extraIconRenderers.forEach(function(renderer) {
```
jjaffeux, line 160: should be fat arrow probably

markvanlan: I literally never know.

PR #9744, opened 2020-05-11
DEV: Generate `emoji/data` on demand

eviltrout:
These emoji rarely change and can be re-generated when we make changes
to them.


jjaffeux:
LGTM :+1:

PR #9754, opened 2020-05-12
FIX: randomly failing user_spec

lis2:
When we evaluate `update_last_seen!` we relay on Redis to not run that code too often

https://github.com/discourse/discourse/blob/master/app/models/user.rb#L753

The problem is that not all specs which are running `update_last_seen!` are cleaning after themselves
```
after do
  Discourse.redis.flushall
end
```
For examples specs in that block https://github.com/discourse/discourse/blob/master/spec/models/user_spec.rb#L901

So issue can be replicated when you run a few times
`bundle exec rspec  ./spec/models/user_spec.rb -e "should not update the first seen value if it doesn't exist" -e "should have 0 for days_visited"`

We should flush Redis after each spec which is evaluating `update_last_seen!`

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -874,7 +874,7 @@ describe User do
     end
 
     after do
-      Discourse.redis.flushall
+      Discourse.redis.del("user:#{user.id}:#{Time.zone.now.to_date}")
     end
 
     it "should act correctly" do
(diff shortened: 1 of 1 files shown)


SamSaffron:
I worry a bit about flushall, I wonder if we can be more explicit, at least the test is aware of exactly what it mucked with... I guess I am okish with this cause we do it elsewhere

lis2:
@SamSaffron you are right, it may create random failure somewhere else. Let me fix it to drop that specific key

PR #9760, opened 2020-05-13
DEV: Migrate deprecated badge icon names

pmusaraj:


pmusaraj:
@eviltrout I'm not sure whether it is OK that this relies on a JSON file outside migrations.

db/migrate/20200513013022_update_deprecated_badge_icon_names.rb
```diff
@@ -0,0 +1,25 @@
...
+  def up
+    migrate_value("up")
+  end
+
+  def down
+    migrate_value("down")
+  end
+
+  def migrate_value(dir)
+    icons = File.open("#{Rails.root}/lib/svg_sprite/fa4-renames.json", "r:UTF-8") { |f| JSON.parse(f.read) }
```
eviltrout, line 13: I think it's OK to use a JSON file but putting it in `lib` implies it's part of the app and it's not clear it's part of a migration.

I would make a copy of the file in the migrations directory, and maybe name it the same as the migration number? `20200513013022_fa4_renames.json`

pmusaraj: That makes sense, thanks! I have made the change, but moved the PR back to draft for now, waiting to see whether we can do one and the same migration for group icons too (since it will use the same JSON file).

pmusaraj:
Closing, superseded by https://github.com/discourse/discourse/pull/9806

PR #9765, opened 2020-05-13
FEATURE: Add endpoint for individual SVG icons

pmusaraj:
This adds an endpoint to render an SVG icon extracted from the included SVG sprites at `/svg-sprites/HOSTNAME/icon/ICON-NAME.svg`. An optional color parameter (HEX only) is also available, via "/svg-sprite/HOSTNAME/icon/CC0000/ICON-NAME.svg`. 

This should allow SVG icons to be used in PWA shortcuts menu, see https://github.com/discourse/discourse/pull/9749. cc @xfalcox

### DIFF
--- a/app/controllers/svg_sprite_controller.rb
+++ b/app/controllers/svg_sprite_controller.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class SvgSpriteController < ApplicationController
-  skip_before_action :preload_json, :redirect_to_login_if_required, :check_xhr, :verify_authenticity_token, only: [:show, :search]
+  skip_before_action :preload_json, :redirect_to_login_if_required, :check_xhr, :verify_authenticity_token, only: [:show, :search, :svg_icon]
 
-  requires_login except: [:show]
+  requires_login except: [:show, :svg_icon]
 
   def show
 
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -503,6 +503,7 @@
   get "svg-sprite/:hostname/svg-:theme_ids-:version.js" => "svg_sprite#show", constraints: { hostname: /[\w\.-]+/, version: /\h{40}/, theme_ids: /([0-9]+(,[0-9]+)*)?/, format: :js }
   get "svg-sprite/search/:keyword" => "svg_sprite#search", format: false, constraints: { keyword: /[-a-z0-9\s\%]+/ }
   get "svg-sprite/picker-search" => "svg_sprite#icon_picker_search", defaults: { format: :json }
+  get "svg-sprite/icon(/:color)/:name.svg" => "svg_sprite#svg_icon", constraints: { name: /[-a-z0-9\s\%]+/, format: :svg }
```
davidtaylorhq, line 506: I think this needs a `:hostname` parameter, since the action in the controller checks for `params[:hostname]`

app/controllers/svg_sprite_controller.rb
```diff
@@ -49,4 +49,25 @@ def icon_picker_search
...
+      if icon.blank?
+        render body: nil, status: 404
+      else
+        doc = Nokogiri.XML(icon)
+        doc.at_xpath("symbol").name = "svg"
+        doc.at_xpath("svg")['xmlns'] = "http://www.w3.org/2000/svg"
+        doc.at_xpath("svg")['fill'] = "##{params[:color]}" if params[:color]
+
+        render plain: doc, disposition: nil, content_type: 'image/svg+xml'
+      end
```
xfalcox, line 70: Maybe add some cache headers? So the CDN can cache it for us. Since the icons can change between updates, I'd say a day?

PR #9769, opened 2020-05-13
PERF: Dematerialize topic_reply_count

riking:
Instead of updating topic_reply_count on every post creation, deletion, and ownership change, calculate it from scratch during trust level promotions. Additionally, replace the `topic_count + reply_count` on the client with `any_posts`, as it was only ever compared to 0.

### DIFF
--- a/app/assets/javascripts/admin/models/tl3-requirements.js
+++ b/app/assets/javascripts/admin/models/tl3-requirements.js
@@ -12,6 +12,11 @@ export default EmberObject.extend({
     return Math.round((minDaysVisited * 100) / timePeriod);
   },
 
+  @discourseComputed("num_topics_replied_to", "min_topics_replied_to")
+  capped_topics_replied_to(numReplied, minReplied) {
+    return numReplied === 2 * minReplied;
+  },
+
   @discourseComputed(
     "days_visited",
     "min_days_visited",
--- a/app/assets/javascripts/admin/models/tl3-requirements.js
+++ b/app/assets/javascripts/admin/models/tl3-requirements.js
@@ -14,7 +14,7 @@ export default EmberObject.extend({
 
   @discourseComputed("num_topics_replied_to", "min_topics_replied_to")
   capped_topics_replied_to(numReplied, minReplied) {
-    return numReplied === 2 * minReplied;
+    return numReplied > minReplied;
   },
 
   @discourseComputed(
--- a/app/assets/javascripts/discourse/app/components/notification-consent-banner.js
+++ b/app/assets/javascripts/discourse/app/components/notification-consent-banner.js
@@ -26,13 +26,13 @@ export default DesktopNotificationConfig.extend({
     isNotSupported,
     isEnabled,
     bannerDismissed,
-    any_posts
+    anyPosts
   ) {
     return (
       this.siteSettings.push_notifications_prompt &&
       !isNotSupported &&
       this.currentUser &&
-      any_posts &&
+      anyPosts &&
       Notification.permission !== "denied" &&
       Notification.permission !== "granted" &&
       !isEnabled &&
--- a/app/assets/javascripts/discourse/app/components/notification-consent-banner.js
+++ b/app/assets/javascripts/discourse/app/components/notification-consent-banner.js
@@ -20,21 +20,19 @@ export default DesktopNotificationConfig.extend({
     "isNotSupported",
     "isEnabled",
     "bannerDismissed",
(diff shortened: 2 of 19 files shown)


app/models/user_stat.rb
```diff
@@ -151,12 +151,24 @@ def update_distinct_badge_count
...
 
   # topic_reply_count is a count of posts in other users' topics
-  def update_topic_reply_count
-    self.topic_reply_count = Topic
-      .joins("INNER JOIN posts ON topics.id = posts.topic_id AND topics.user_id <> posts.user_id")
-      .where("posts.deleted_at IS NULL AND posts.user_id = ?", self.user_id)
-      .distinct
-      .count
+  def calc_topic_reply_count!(max)
+    self.topic_reply_count = DB.query_single(<<~SQL, self.user_id, max).first
```
riking, line 155: the column is now useless, drop it from the db

app/assets/javascripts/discourse/app/components/notification-consent-banner.js
```diff
@@ -20,21 +20,19 @@ export default DesktopNotificationConfig.extend({
...
-    "currentUser.topic_count"
+    "currentUser.any_posts"
   )
   showNotificationPromptBanner(
     isNotSupported,
     isEnabled,
     bannerDismissed,
-    replyCount,
-    topicCount
+    any_posts
```
ZogStriP, line 29: We use `camelCase` on the client-side (except very few places) 😉 

```suggestion
    anyPosts
```

app/models/trust_level3_requirements.rb
```diff
@@ -143,11 +143,7 @@ def min_days_visited
   end
 
   def num_topics_replied_to
-    @user.posts
-      .public_posts
-      .where("posts.created_at > ? AND posts.post_number > 1", time_period.days.ago)
-      .select("distinct topic_id")
-      .count
+    @user.user_stat.calc_topic_reply_count!(min_topics_replied_to * 2, time_period.days.ago)
```
ZogStriP, line 146: Why `*2` and not just `+1`?

app/models/user_stat.rb
```diff
@@ -151,12 +154,30 @@ def update_distinct_badge_count
...
+    sql = <<~SQL
+    SELECT COUNT(*) count
+    FROM (
+      SELECT DISTINCT posts.topic_id
+      FROM posts
+      INNER JOIN topics ON topics.id = posts.topic_id
+      WHERE posts.user_id = ?
+      AND topics.user_id <> posts.user_id
+      AND posts.deleted_at IS NULL AND topics.deleted_at IS NULL
+      AND topics.archetype <> 'private_message'
```
ZogStriP, line 167: Looks like this is a new condition. Is it intended?

app/assets/javascripts/admin/models/tl3-requirements.js
```diff
@@ -12,6 +12,11 @@ export default EmberObject.extend({
     return Math.round((minDaysVisited * 100) / timePeriod);
   },
 
+  @discourseComputed("num_topics_replied_to", "min_topics_replied_to")
+  capped_topics_replied_to(numReplied, minReplied) {
+    return numReplied === 2 * minReplied;
```
ZogStriP, line 17: I think `>=` would be more correct here.

```suggestion
    return numReplied >= minReplied;
```

riking: If I'm doing a +1, then `>` would be correct, no?

app/models/trust_level3_requirements.rb
riking: TL3 requirements checked for `.public_posts` https://github.com/discourse/discourse/pull/9769/files#diff-71b275fc6f97d344de4978707d0f2fdeL147

app/models/user_stat.rb
riking: TL3 requirements checked for `.public_posts` https://github.com/discourse/discourse/pull/9769/files#diff-71b275fc6f97d344de4978707d0f2fdeL147

app/assets/javascripts/admin/models/tl3-requirements.js
ZogStriP: I'm always off by one :P

app/models/user_stat.rb
ZogStriP: TL2 didn't as far as I know, but I guess it's better that way 👍

PR #9770, opened 2020-05-13
FIX: Specific email error for replies to digest emails

markvanlan:
Emails the are replies to Digest emails, are getting rejected. Currently They are rejected by `BadDestinationError`. This is the catchall, that multiple cases result in. 

This PR adds a check to make sure sure that the `In-Reply-To` header of incoming messages does not match the `message_id` of a Digest email that was sent out. If this is the case, we can send a much more specific error email, saying that you cannot reply to a digest email.

:tada: Email debugging is **fun**

### DIFF
--- a/lib/email/receiver.rb
+++ b/lib/email/receiver.rb
@@ -191,7 +191,7 @@ module Email
           end
         end
 
-        raise ReplyToDigestError if EmailLog.find_by(message_id: @mail.in_reply_to).present?
+        raise ReplyToDigestError if EmailLog.find_by(email_type: "digest", message_id: @mail.in_reply_to).present?
         raise BadDestinationAddress
       end
     end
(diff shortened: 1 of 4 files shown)


coding-horror:
I AM EXCITE 🎉

lib/email/receiver.rb
```diff
@@ -190,6 +191,7 @@ def process_internal
           end
         end
 
+        raise ReplyToDigestError if EmailLog.find_by(email_type: "digest", message_id: @mail.in_reply_to).present?
```
eviltrout, line 194:
```suggestion
        raise ReplyToDigestError if EmailLog.find_by(email_type: "digest", message_id: @mail.in_reply_to).exists?
```

markvanlan: @eviltrout I commited this and got `undefined method "exists?" for `EmailLog`. Switching back to `.present?`

markvanlan: Ah NVM, I switched the `find_by` to `where` so that `.exists?` should work.

eviltrout: Oh my bad it should be `EmailLog.where(email_type: "digest", message_id: @mail.in_reply_to).exists?`

PR #9772, opened 2020-05-13
DEV: Allow plugins to hide user stats by new guardian method

markvanlan:


markvanlan:
@eviltrout This allows plugins to override the `can_see_summary_stats?` method in UserGuardian. This will remove all the user stats from the serializer, and then the UI will hide the entire section.

It is the least DRY code I have seen......

app/assets/javascripts/discourse/app/controllers/user-summary.js
```diff
@@ -11,6 +11,41 @@ export default Controller.extend({
   userController: inject("user"),
   user: alias("userController.model"),
 
+  @discourseComputed(
```
jjaffeux, line 14: I think you can replace all of this by:

```
showStats: or("model.time_read", "model.days_visited", "model.topics_entered", ...)
```

markvanlan:
@eviltrout I am not sure if you prefer an attribute accessor, or the method to expose `@user` from UserSummary, but it would be really nice to expose it, to avoid extra queries for no reason.

app/assets/javascripts/discourse/app/controllers/user-summary.js
```diff
@@ -10,6 +10,7 @@ const MAX_BADGES = 6;
 export default Controller.extend({
   userController: inject("user"),
   user: alias("userController.model"),
+  showStats: readOnly("model.can_see_summary_stats"),
```
eviltrout, line 13: I would not even bother making this property, myself. You can have the template check `model.can_see_summary_stats` directly.

PR #9775, opened 2020-05-14
FEATURE: Trigger browser-detect on Internet Archive replays

riking:
... this mechanism of action is a bad idea.

Replay currently fails with an infinite recursion somewhere inside jQuery.


udan11:
My opinion is that we should be finding the cause for jQuery's infinite recursion and check for that, because it might affect other old browsers too. If that is no possible, then this solution is fine too.

SamSaffron:
I agree 100% we should fix the infinite recursion. 

But for wayback machine there is an argument that can be made that just having non JS pages is better. It makes stuff less fragile, means we can confidently render even if wayback throws away a JS dependency or so on.

PR #9776, opened 2020-05-14
Remove old web hooks in favor of 'reviewable' web hook

nbianca:


app/assets/javascripts/admin/routes/admin-web-hooks-show.js
```diff
@@ -15,7 +14,7 @@ export default DiscourseRoute.extend({
   },
 
   setupController(controller, model) {
-    if (model.get("isNew") || isEmpty(model.get("web_hook_event_types"))) {
```
eviltrout, line 18: I'm curious why this line changed?

nbianca: It was a bit odd. I removed two web hook types which led to some web hooks not having any type. When editing such web hooks, it would automatically check the ‘post event’ type which I found weird.

PR #9779, opened 2020-05-14
UX: use "icon-picker" & "image-uploader" fields to set group flair.

vinothkannans:


app/assets/javascripts/discourse/app/components/group-flair-inputs.js
```diff
@@ -90,5 +99,17 @@ export default Component.extend({
   flairPreviewLabel(flairPreviewImage) {
     const key = flairPreviewImage ? "image" : "icon";
     return I18n.t(`groups.flair_preview_${key}`);
+  },
+
+  @action
+  setFlairImage(upload) {
+    this.model.set("flair_url", Discourse.getURL(upload.url));
+    this.model.flair_image_id = upload.id;
```
jjaffeux, line 107: why not set here ?

app/assets/javascripts/discourse/app/components/group-flair-inputs.js
```diff
@@ -90,5 +99,17 @@ export default Component.extend({
...
+  },
+
+  @action
+  removeFlairImage() {
+    this.model.set("flair_url", null);
+    this.model.flair_image_id = null;
```
jjaffeux, line 113: why not set here ?

app/assets/javascripts/discourse/app/components/group-flair-inputs.js
```diff
@@ -15,23 +16,26 @@ export default Component.extend({
...
+  @discourseComputed("model.flair_icon")
+  flairPreviewIconUrl(flairIcon) {
+    return flairIcon ? convertIconClass(flairIcon) : "";
   },
 
-  @observes("model.flair_url")
-  _loadSVGIcon() {
-    debounce(this, this._loadIcon, 1000);
+  @on("didInsertElement")
+  @observes("model.flair_icon")
```
jjaffeux, line 30: maybe you can try to replace all of this by @on("didReceiveAttrs")

not sure it would work didn't pull the pr

davidtaylorhq:
I think this new upload_id column will need to be [added to `clean_up_uploads.rb`](https://github.com/discourse/discourse/blob/c979f1877bb1971e0d61fe80a9394be986fb5860/app/jobs/scheduled/clean_up_uploads.rb#L78-L78), to make sure the upload doesn't get deleted automatically.

db/migrate/20200511043818_add_more_flair_columns_to_group.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class AddMoreFlairColumnsToGroup < ActiveRecord::Migration[6.0]
+  def change
+    add_column :groups, :flair_icon, :string
+    add_reference :groups, :flair_image, foreign_key: { to_table: :uploads }
```
tgxworld, line 6: `add_reference` is discouraged in our code base now. It'll be better to be more explicit about the commands here.

tgxworld: Also I think `flair_upload_id` will be a better name since we're referencing an upload.

app/models/group.rb
```diff
@@ -26,6 +26,8 @@ class Group < ActiveRecord::Base
   has_many :category_reviews, class_name: 'Category', foreign_key: :reviewable_by_group_id, dependent: :nullify
   has_many :reviewables, foreign_key: :reviewable_by_group_id, dependent: :nullify
 
+  belongs_to :flair_image, class_name: 'Upload'
```
tgxworld, line 29: This association seems strange to me. I would think that a `Group` -> `has_one :flair_image` rather than `belongs_to` it

db/migrate/20200511043818_add_more_flair_columns_to_group.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    DB.exec(
+      <<~SQL
+        UPDATE groups SET flair_icon = flair_url
```
pmusaraj, line 10: Since we are migrating the icons to their own column, I think this would be a good time to simplify the format as well. Currently, the accepted formats are "fa-icon" or "fas fa-icon" for solid style icons, "far fa-icon" for regular and "fab fa-icon" for brands. This is finicky (and sadly, it's my fault) and it doesn't work with the icon picker. 

The following one-line change should work: 

```
UPDATE groups SET flair_icon = REPLACE(REPLACE(flair_url, 'fas fa-', ''), ' fa-', '-')
```

After this, the accepted formats are "icon", "far-icon" and "fab-icon". (And the picker enforces the correct format anyway.)

db/migrate/20200511043818_add_more_flair_columns_to_group.rb
```diff
@@ -0,0 +1,32 @@
...
+        SET flair_image_id = u.id
+        FROM groups g2
+        INNER JOIN uploads u ON g2.flair_url ~ CONCAT('\/', u.sha1, '[\.\w]*')
+        WHERE g1.id = g2.id
+      SQL
+    )
+
+    DB.exec(
+      <<~SQL
+        UPDATE groups SET flair_url = NULL
```
pmusaraj, line 27: I think we should be careful here, this erases `flair_url` columns when if you revert the migration for some reason. It might be safer to use `up` `down` methods instead.

app/models/group.rb
vinothkannans: Since the target table has primary key field (`upload.id`) and "group" table has child key we should use `belongs_to` here, I hope. We have a similar code in [`user.rb`](https://github.com/discourse/discourse/blob/6c1c8be79433f87bef9d768da7b8fa4ec9bb18d7/app/models/user.rb#L53) file.

Previously, I tried `has_one` and it returned an error.

vinothkannans:
All the requested changes are done :+1:

app/models/group.rb
tgxworld: Yup you’re right. I got confused about where the foreign key should be on.

PR #9780, opened 2020-05-14
UI: refinements to topic footer buttons

jjaffeux:
### On mobile:
Screenshot 2020-05-14 at 10 44 58

### Desktop screen size reduced:
Screenshot 2020-05-14 at 17 32 20

### Desktop full width:
Screenshot 2020-05-14 at 17 32 24


jjaffeux:
@awesomerobot what do you think of this ?

jjaffeux:
> I really don't like it when the text wraps under the button... I'd prefer something like this instead (now that I look more closely, we do this on mobile but not on desktop currently):

Actually I have been the doing this not wrapping under the button long time ago, but jeff didn't like this and I have had to change this back for years...

> Also one other thing while I'm looking at it... pin/unpin has the triangle, but tracking does not... probably out of scope for this change, but 🤔

Yes can do that in another PR

jjaffeux:
@awesomerobot I have updated the PR to remove the inline part.

Screenshot 2020-05-17 at 16 27 58
Screenshot 2020-05-17 at 16 28 11
Screenshot 2020-05-17 at 16 28 56

CvX:
@jjaffeux what happened to the space between the wrench button and the bookmark button? 😃

jjaffeux:
Yes fixed it, sadly I can't make all the changes I want here, because I can't make d-button flex, wish I could...

jjaffeux:
> Looks like a margin went missing?
> 
> Screen Shot 2020-05-18 at 6 29 35 PM

Yes indeed, I thought pinned was out of DOM when not present, but it's just hidden, fixed it thanks.

PR #9781, opened 2020-05-14
FIX: Use Discourse.system_user when we need a placeholder admin

riking:
No clue why a fake User object was created with `id = -1` when system_user is right there. This fixes some exceptions.

### DIFF
--- a/app/services/random_topic_selector.rb
+++ b/app/services/random_topic_selector.rb
@@ -25,7 +25,7 @@ class RandomTopicSelector
       end
 
       # don't leak private categories into the "everything" group
-      options[:guardian] = Guardian.new(CategoryFeaturedTopic.fake_admin)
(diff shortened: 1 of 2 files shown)


app/services/random_topic_selector.rb
```diff
@@ -25,7 +25,7 @@ def self.backfill(category = nil)
       end
 
       # don't leak private categories into the "everything" group
-      options[:guardian] = Guardian.new(CategoryFeaturedTopic.fake_admin)
+      options[:guardian] = Guardian.new(Discourse.system_user)
```
riking, line 28: @SamSaffron, blame says you're the last to change this - can you recheck the security implications of this?

SamSaffron:
sure this seems like a very safe change to me 

Code predates Discourse.system_user :)

riking:
Well, that would do it!

PR #9786, opened 2020-05-15
DEV: Refactor draft attributes for `CategoryList` and `TopicList`.

tgxworld:
Avoid repeating the same logic in a bunch of places which will allow us
to make changes to the draft attributes easier in the future.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -30,12 +30,6 @@ class CategoriesController < ApplicationController
     }
 
     @category_list = CategoryList.new(guardian, category_options)
-    @category_list.draft_key = Draft::NEW_TOPIC
-    @category_list.draft_sequence = DraftSequence.current(
-      current_user,
-      Draft::NEW_TOPIC
-    )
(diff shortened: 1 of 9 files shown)


SamSaffron:
I am fine with a refactor here but concerns ... concern me ... they hide where stuff is implemented I much prefer a static method here of sorts, can you refactor this elsewise?

tgxworld:
> I am fine with a refactor here but concerns ... concern me ... they hide where stuff is implemented I much prefer a static method here of sorts

I'm not sure I understand what the concern here with concerns are. Isn't composition encouraged?

SamSaffron:
In this case there is no need for any kind of concern class, just have both  inherit off BaseList or DraftableList? 

composition if fine, but I thing using concerns is a last resort after standard inheritance can not fit the bill.

tgxworld:
We've generally been avoiding inheritance and preferring composition in our codebase so I still don't understand why including a module is not preferred here.

tgxworld:
@SamSaffron I updated the PR per your review.

tgxworld:
Merged in https://github.com/discourse/discourse/commit/b979579c1bec92d5e786a2004d0af8c9b07d5eab

PR #9789, opened 2020-05-15
PERF: Update some `User` associations` to dependent: :delete_all`.

tgxworld:
`dependent: :destroy` creates an AR object and then calls `destroy` on
it which is slow. For certain associations, we do not run callbacks or
need to run callbacks so we should avoid creating all the AR objects.


tgxworld:
I've look through the associations and these associations do not have any `before` or `after` destroy callbacks.

tgxworld:
Closed by https://github.com/discourse/discourse/commit/6b92c78033a1a26eea56f0417b6811581fab7a38

PR #9791, opened 2020-05-15
FIX: Switching composer action does not refresh composer actions content

tgxworld:


eviltrout:
I'm merging this myself because I want to move select-kit files around.

PR #9793, opened 2020-05-15
UX: Add copied text upon copy button click

justindirose:
Adds copied text to the codeblock copy button on click. This improves the user experience by explicitly stating the copy has happened.


jjaffeux:
Can you show me how it looks please ?

PR #9794, opened 2020-05-15
FIX: displays a title on sk header if no selected name

jjaffeux:
none has to be defined.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1667,6 +1667,9 @@ en:
         one: "Select at least {{count}} item."
         other: "Select at least {{count}} items."
       invalid_selection_length: "Selection must be at least {{count}} characters."
+      components:
+        categories_admin_dropdown:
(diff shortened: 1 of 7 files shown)


config/locales/client.en.yml
```diff
@@ -1667,6 +1667,9 @@ en:
         one: "Select at least {{count}} item."
         other: "Select at least {{count}} items."
       invalid_selection_length: "Selection must be at least {{count}} characters."
+      components:
+        categories_admin_dropdown:
+          title: "Categories admin dropdown"
```
gschlager, line 1672: Was there a particular reason for adding this title? Would something break if we remove it?
It seems to confuse translators and personally I don't see a value in it. I still have to click on the button to find out what it's doing. If we need it, maybe "manage categories" or something like that would be more helpful?

jjaffeux: 100% open to a better copy :+1:

PR #9795, opened 2020-05-15
UX: focus category name when editing category

jjaffeux:


jjaffeux:
@eviltrout Couldn't find a less hacky way to make it work consistently (scheduleRender or autofocus don't work for example), but given we have plans to rework this screen now, I guess that's ok.

PR #9797, opened 2020-05-15
DEV: Move select kit to an addon

eviltrout:
It's mostly about moving it into an `app` directory and adding some boilerplate files.


jjaffeux:
I think we can trust the tests on this, select-kit has a lot of them, and its also used in tons of features which have tests... so we can be quite confident here :+1:

PR #9798, opened 2020-05-15
UX: Remove live theme previewing in favor of refresh

markvanlan:
@davidtaylorhq Just merged [a commit](https://github.com/discourse/discourse/commit/eee20a9814217376dd2e5fd48e2fb1a490bd2d1d) that sends plugin css to the browser for live theme previewing. This fixes some annoying issues if plugins use scss variables, as they are not adjusted when the theme changed previously.

The issue with this, is that theme JS is not magically swapped out, so we need a refresh anyways. This gives a note that the theme will change once the user pressed "Save Changes", and then refreshes automatically.

Screenshot from 2020-05-15 14-38-48


app/assets/javascripts/discourse/app/controllers/preferences/interface.js
```diff
@@ -2,19 +2,16 @@ import I18n from "I18n";
...
+import discourseComputed from "discourse-common/utils/decorators";
+import { listThemes, setLocalTheme } from "discourse/lib/theme-selector";
 import { popupAjaxError } from "discourse/lib/ajax-error";
 import {
   safariHacksDisabled,
   isiPad,
   iOSWithVisualViewport
 } from "discourse/lib/utilities";
 import { computed } from "@ember/object";
+import DiscourseEnv from "discourse-common/config/environment";
```
jjaffeux, line 14: @eviltrout I'm curious here, what should we be using?

```
Ember.testing
DiscourseEnv. environment
Discourse.testing
```

Is there anything which will play better with ember-cli ?

markvanlan: Right now I am trying to figure out what test is hitting this, and see if it would be better to just write `location.reload` to do nothing :shrug:

CvX: Yeah, we should avoid adding any test-specific code to the app. Instead, stub the `reload` function.

markvanlan: @CvX I cant figure out how to stub `location.reload`. I tried with sandbox, and the property is unwriteable 
```
Promise rejected during "update some fields": Cannot redefine property: reload
```

CvX: Right. We'll have to introduce a bit of indirection here: add a method (or a service) that does the `location.reload()` and stub that method/service.

markvanlan: @CvX Alright a made a little helper called `page-reloader` which has a method we can stub :+1:

jjaffeux: I'm still curious about my initial question though :p because we have so many ways to check for testing

eviltrout: I usually use `Ember.testing` but I am not sure yet if that breaks in Ember CLI. We should stick with that for now and if it doesn't work in the future we'll mass rename.

markvanlan:
@eviltrout I really don't know the best place to put the `discourseEnv` check to keep the tests from stalling. If I stub the reload in individual tests, the same stub will have to go in every test, and this will come up again in the future where the stub is forgotten. I guess I could stub in the `test_helper.js` for all JS tests. Or I could keep the check where it is currently in this PR. What is your preference?

eviltrout:
@markvanlan  does `Ember.testing` not work?

markvanlan:
@eviltrout Maybe it does, but I was trying to avoid using `Ember` global

PR #9799, opened 2020-05-16
FEATURE: Add site settings for images and links in excerpts

nrchandan:
The ExcertParser has the features to not strip out the images and
links from the excerpt. This change enables site admins to turn
this feature on or off. The default behaviour remains the same.


hnb-ku:
Please see 

https://meta.discourse.org/t/images-in-topic-excerpts/151607/7?u=johani

PR #9802, opened 2020-05-16
datetime: Show time when date is between yesterday & tomorrow

jessicah:
When showing formatted strings such as "today", if the format includes time, then also include in the resulting output.

This will produce a string such as "today 7:31 AM".

See: https://meta.discourse.org/t/date-should-show-the-time-instead-of-today-for-time-formats/151435


jessicah:
JavaScript isn't my strong suit, but this appears to achieve the desired outcome, bit rough around the edges.

jessicah:
Was just thinking, it may need a special-case for `LLL` as well, where it transforms into `LT` to get the time portion preserved?

jessicah:
Hmm, I think this may have actually been already fixed with the new LocalDateBuilder recently introduced.

PR #9803, opened 2020-05-16
UX: allows dropdon-select-box to display a caret

jjaffeux:
```
selectKitOptions: {
  showCaret: true // default false
}
```


jjaffeux:
Screenshot 2020-05-16 at 21 32 08

PR #9805, opened 2020-05-17
datetime: handle date string with encoded time portion.

jessicah:
* This will handle input of the form `date=2020-05-14T19:31`.


jjaffeux:
Hi, I'm sorry I don't want to support this at this time. Thanks for your contribution though.

eg: what if someone adds a Z at the end, which would be valid for a date time with T, this would break your implementation.

If we wanted to do this, you should parse the date with moment and reformat into date and time to have something more robust, but even this is annoying to me because why not expect this to set the timezone property too then?

I want to keep this simple solution for now, I have plans to make writing dates easier in the near future.

jessicah:
The problem is that without this, it breaks existing posts where this worked prior to your changes, as noted on the forum.

PR #9808, opened 2020-05-17
FIX: extract reset_last_seen_cache! in user_spec

lis2:
Helper method to avoid code duplication

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -874,7 +874,7 @@ describe User do
     end
 
     after do
-      Discourse.redis.del("user:#{user.id}:#{Time.zone.now.to_date}")
+      reset_last_seen_cache!(user)
     end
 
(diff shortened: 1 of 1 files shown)


spec/models/user_spec.rb
```diff
@@ -2374,4 +2374,8 @@ def filter_by(method)
       expect(user2.recent_time_read).to eq(50)
     end
   end
+
+  def reset_last_seen_cache!(user)
+    Discourse.redis.del("user:#{user.id}:#{Time.zone.now.to_date}")
+  end
```
SamSaffron, line 2380: Minor but I prefer keeping this in the model proper, that way you get to reused the key calculation code that the model uses. 

I know it can sometimes feel a bit weird to have stuff like this in the model, so this is a bit of a bike shed. But overall this reduces code duplication which is a good thing.

PR #9809, opened 2020-05-17
FIX: Handle non-AWS S3 to local migration for assets that do not use upload: pseudo-protocl

johnsonm:


eviltrout:
@martin-brennan  can you review this one please? Thanks!

johnsonm:
Discourse is my introduction to ruby development, so I very much appreciate the pointer. I hate writing code without automated tests.

I don't think a _minimal_ test is really meaningful without adding some baseline coverage though. I'll see what I can do!

johnsonm:
I think that all the video uploads in my google+ import might have been broken when I tried to do it; it's not at all clear to me that this PR is meaningful.

johnsonm:
I have found examples that do need to be fixed in my migration, and the tests I added in the `batch` limit PR will help here.

There is an additional problem that `UploadCreator` is doing strange things with videos; with this broken, it happens that it will ignore them; with this problem fixed, it actually breaks the posts. It modified the post to:

`https:/uploads/default/original/3X/b/a/ba9e06ebc2f4397f26793bb5cd4e169308dd371d.mp4`

It looks like `new_upload.url` isn't right for a video. It's definitely giving me `/uploads/default/original/3X/b/a/ba9e06ebc2f4397f26793bb5cd4e169308dd371d.mp4` which is not a base URL.

johnsonm:
This is replaced by #10093 which has many other fixes.

PR #9813, opened 2020-05-18
FEATURE: multiple use invite links

techAPJ:
This commit allows Admins to create an invite link that can be used multiple times.

Creating Invite Link:

Screen Shot 2020-05-28 at 11 02 58

Invite Link listing page:

Screen Shot 2020-05-18 at 12 12 45

Invite Link redemption page:

Screen Shot 2020-05-18 at 12 09 47

Invite "Redeemed" tab showing all the accepted invites:

Screen Shot 2020-05-18 at 12 10 51


app/assets/stylesheets/common/components/share-and-invite-modal.scss
```diff
@@ -62,7 +62,7 @@
   }
 }
 
-.share-and-invite.modal .invite.modal-panel {
+.share-and-invite.modal .invite.modal-panel,.invite-link.modal-panel {
```
jjaffeux, line 65: missing a space here, will fail linting

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
+import I18n from "I18n";
+import Component from "@ember/component";
+import Group from "discourse/models/group";
+import { alias } from "@ember/object/computed";
+import discourseComputed from "discourse-common/utils/decorators";
+import Invite from "discourse/models/invite";
+
+export default Component.extend({
+  inviteModel: alias("panel.model.inviteModel"),
```
jjaffeux, line 9: would prefer readOnly instead of all the alias here

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
...
+
+  groupFinder(term) {
+    return Group.findAll({ term, ignore_automatic: true });
+  },
+
+  errorMessage: I18n.t("user.invited.invite_link.error"),
+
+  // Reset the modal to allow a new user to be invited.
+  reset() {
+    this.setProperties({
```
jjaffeux, line 61: should be set there's only one property

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
...
+      const model = this.inviteModel;
+      model.setProperties({ saving: true, error: false });
+
+      return model
+        .generateMultipleUseInviteLink(
+          groupNames,
+          maxRedemptionAllowed,
+          inviteExpiresAt
+        )
+        .then((result) => {
```
jjaffeux, line 93:
```suggestion
        .then(result => {
```

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
...
+          model.setProperties({
+            saving: false,
+            finished: true,
+            inviteLink: result,
+          });
+
+          if (userInvitedController) {
+            Invite.findInvitedBy(
+              this.currentUser,
+              userInvitedController.get("filter")
```
jjaffeux, line 103: do we need a get here ?

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
...
+            ).then((inviteModel) => {
+              userInvitedController.setProperties({
+                model: inviteModel,
+                totalInvites: inviteModel.invites.length,
+              });
+            });
+          }
+        })
+        .catch((e) => {
```
jjaffeux, line 112:
```suggestion
        .catch(e => {
```

app/assets/javascripts/discourse/app/controllers/invites-show.js
```diff
@@ -26,11 +29,13 @@ export default Controller.extend(
     errorMessage: null,
     userFields: null,
     inviteImageUrl: getUrl("/images/envelope.svg"),
+    isInviteLink: alias("model.is_invite_link"),
+    rejectedEmails: A(),
```
jjaffeux, line 33: Why do you need a native array here ?

Also I would prefer this to be null, and the array initiated in init()

app/assets/javascripts/discourse/app/controllers/invites-show.js
```diff
@@ -45,18 +50,26 @@ export default Controller.extend(
     },
 
     @discourseComputed(
```
jjaffeux, line 52: you can replace all of this by using or()

app/assets/javascripts/discourse/app/controllers/invites-show.js
```diff
@@ -80,14 +122,15 @@ export default Controller.extend(
...
             name: this.accountName,
             password: this.accountPassword,
             user_custom_fields: userCustomFields,
-            timezone: moment.tz.guess()
-          }
+            timezone: moment.tz.guess(),
+          },
         })
-          .then(result => {
+          .then((result) => {
```
jjaffeux, line 133:
```suggestion
          .then(result => {
```

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
```diff
@@ -31,10 +31,16 @@ export default Controller.extend({
       this.user,
       this.filter,
       this.searchTerm
-    ).then(invites => this.set("model", invites));
+    ).then((invites) => this.set("model", invites));
```
jjaffeux, line 34:
```suggestion
    ).then(invites => this.set("model", invites));
```

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
```diff
@@ -31,10 +31,16 @@ export default Controller.extend({
...
   }, INPUT_DELAY),
 
   inviteRedeemed: equal("filter", "redeemed"),
+  invitePending: equal("filter", "pending"),
+
+  @discourseComputed("filter")
+  inviteLinks(filter) {
+    return filter === "links" && this.currentUser.get("admin");
```
jjaffeux, line 42: do we need the get here ? if currentUser is null having a get won't save you from an error

jjaffeux, line 42: maybe we should have a constant for "links" here

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
```diff
@@ -97,13 +117,16 @@ export default Controller.extend({
...
 
     reinviteAll() {
-      bootbox.confirm(I18n.t("user.invited.reinvite_all_confirm"), confirm => {
-        if (confirm) {
-          Invite.reinviteAll()
-            .then(() => this.set("reinvitedAll", true))
-            .catch(popupAjaxError);
+      bootbox.confirm(
+        I18n.t("user.invited.reinvite_all_confirm"),
+        (confirm) => {
```
jjaffeux, line 122:
```suggestion
        confirm => {
```

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
```diff
@@ -116,7 +139,7 @@ export default Controller.extend({
           this.filter,
           this.searchTerm,
           model.invites.length
-        ).then(invite_model => {
+        ).then((invite_model) => {
```
jjaffeux, line 142:
```suggestion
        ).then(invite_model => {
```

app/assets/javascripts/discourse/app/templates/components/generated-invite-link.hbs
```diff
@@ -1,5 +1,7 @@
 <p>{{i18n "user.invited.link_generated"}}</p>
 <p>
-  <input value={{link}} class="invite-link-input" style="width: 75%" type="text">
+  <input value={{link}} class="invite-link-input" style="width: 100%" type="text">
```
jjaffeux, line 3: can you try to remove this inline style while you are at it please ?

app/assets/javascripts/discourse/app/templates/components/invite-link-panel.hbs
```diff
@@ -0,0 +1,67 @@
+{{#if inviteModel.error}}
+  <div class="alert alert-error">
+    {{html-safe errorMessage}}
+  </div>
+{{/if}}
+
+<div class="body">
+  {{#if inviteModel.finished}}
+    {{generated-invite-link link=inviteModel.inviteLink email=null}}
```
jjaffeux, line 9: do we need email=null here ? why not just removed the email property

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,122 @@
...
+
+  @discourseComputed(
+    "isAdmin",
+    "inviteModel.saving",
+    "inviteModel.details.can_invite_to"
+  )
+  disabled(isAdmin, saving, can_invite_to) {
+    if (saving) return true;
+
+    if (!isAdmin) {
```
jjaffeux, line 44: everything else is on one line and this if is done on 3 ? :p

app/assets/javascripts/discourse/app/controllers/invites-show.js
```diff
@@ -26,11 +29,13 @@ export default Controller.extend(
     errorMessage: null,
     userFields: null,
     inviteImageUrl: getUrl("/images/envelope.svg"),
+    isInviteLink: alias("model.is_invite_link"),
```
jjaffeux, line 32: would prefer readOnly here

app/assets/javascripts/discourse/app/controllers/invites-show.js
```diff
@@ -113,10 +164,10 @@ export default Controller.extend(
               }
             }
           })
-          .catch(error => {
+          .catch((error) => {
```
jjaffeux, line 167:
```suggestion
          .catch(error => {
```

techAPJ:
Thanks for the review @jjaffeux! Working through the comments now.

app/assets/javascripts/discourse/app/components/invite-link-panel.js
techAPJ: Looks like we do not.

app/assets/javascripts/discourse/app/components/invite-link-panel.js
techAPJ: copy./paste turned bad 😛

app/assets/javascripts/discourse/app/templates/components/invite-link-panel.hbs
techAPJ: Good idea. Made it so.

app/assets/javascripts/discourse/app/controllers/invites-show.js
techAPJ: Excellent suggestion! Made it so.

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
techAPJ: Good catch! Fixed.

app/assets/javascripts/discourse/app/templates/components/generated-invite-link.hbs
techAPJ: Sure, done!

app/assets/javascripts/discourse/app/controllers/user-invited-show.js
techAPJ: It's not being used anywhere else... also we use the same pattern for checking for "pending" and "redeemed" tabs, so I am not sure about this.

app/assets/javascripts/discourse/app/controllers/invites-show.js
techAPJ: Done.

app/assets/javascripts/discourse/app/components/invite-link-panel.js
techAPJ: Done.

app/assets/javascripts/discourse/app/controllers/invites-show.js
techAPJ: I basically copied the same implementation we have in [`create-account.js`](https://github.com/discourse/discourse/blob/c014b9385448a67143ef401d5461681caa49633c/app/assets/javascripts/discourse/app/controllers/create-account.js#L36).

I made the changes you suggested here.

app/controllers/users_controller.rb
```diff
@@ -303,26 +303,36 @@ def invited
...
       invites = invites.where(filter_sql, filter: "%#{params[:search].downcase}%")
     end
 
-    render json: MultiJson.dump(InvitedSerializer.new(
-      OpenStruct.new(invite_list: invites.to_a, show_emails: show_emails, inviter: inviter),
-      scope: guardian,
-      root: false
-    ))
+    serialized = if filter_by == "links"
+      MultiJson.dump(invites: serialize_data(invites.to_a, InviteLinkSerializer), can_see_invite_details:  guardian.can_see_invite_details?(inviter))
```
SamSaffron, line 326: Are we returning 2 different formats of json depending on params?

app/models/user.rb
```diff
@@ -35,12 +35,13 @@ class User < ActiveRecord::Base
   has_many :topics_allowed, through: :topic_allowed_users, source: :topic
   has_many :email_tokens, dependent: :destroy
   has_many :user_visits, dependent: :destroy
-  has_many :invites, dependent: :destroy
   has_many :topic_links, dependent: :destroy
   has_many :uploads
   has_many :user_warnings
   has_many :user_archived_messages, dependent: :destroy
   has_many :email_change_requests, dependent: :destroy
+  has_many :invited_users
```
SamSaffron, line 43: dependent: :delete_all on invited_users

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddBulkInviteLinkToInvites < ActiveRecord::Migration[6.0]
+  def change
+    add_column :invites, :max_redemption_allowed, :integer, null: false, default: 1
```
SamSaffron, line 5: name is not ideal ... `max_redemptions_allowed` maybe?

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
```diff
@@ -0,0 +1,9 @@
...
+    add_column :invites, :redemption_count, :integer, null: false, default: 0
+    add_column :invites, :invite_link_expires_at, :datetime, null: true
```
SamSaffron, line 7: does this mean invites have 2 expirys?

db/post_migrate/20200430173433_remove_user_id_from_invites.rb
```diff
@@ -0,0 +1,34 @@
...
+        invite_id,
+        redeemed_at,
+        created_at,
+        updated_at
+      )
+      SELECT user_id, id, redeemed_at, created_at, updated_at
+      FROM invites WHERE user_id IS NOT NULL
+SQL
+
+    DROPPED_COLUMNS.each do |table, columns|
```
SamSaffron, line 26: ColumnDropper is only useful if you need `.mark_readonly` or you are disabling ddl transaction. In this case you are doing neither, so you can simply drop the column. also be sure to add something like this: 

https://github.com/discourse/discourse/blob/a2d939608d5b3adc7f037dc0ca38e7d1f6b895f6/app/models/topic.rb#L13-L14

spec/requests/invites_controller_spec.rb
```diff
@@ -203,6 +203,52 @@
     end
   end
 
+  context '#create_multiple_use_invite_link' do
+    it 'requires you to be logged in' do
+      post "/invites/invite-link.json", params: {
+        max_redemption_allowed: 5
+      }
+      expect(response.status).to eq(500)
```
SamSaffron, line 211: 500 ... we don't want 500 errors

https://github.com/discourse/discourse/blob/a2d939608d5b3adc7f037dc0ca38e7d1f6b895f6/spec/requests/topics_controller_spec.rb#L481-L481

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,116 @@
...
+
+    this.reset();
+  },
+
+  @discourseComputed(
+    "isAdmin",
+    "inviteModel.saving",
+    "inviteModel.details.can_invite_to"
+  )
+  disabled(isAdmin, saving, can_invite_to) {
```
jjaffeux, line 41:
```suggestion
  disabled(isAdmin, saving, canInviteTo) {
```

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,116 @@
...
+    if (saving) return true;
+    if (!isAdmin) return true;
+    if (can_invite_to) return false;
```
jjaffeux, line 44:
```suggestion
    if (canInviteTo) return false;
```

app/assets/javascripts/discourse/app/components/invite-link-panel.js
```diff
@@ -0,0 +1,116 @@
...
+    this.inviteModel.setProperties({
+      groupNames: null,
+      error: false,
+      saving: false,
+      finished: false,
+      inviteLink: null
+    });
+  },
+
+  actions: {
```
jjaffeux, line 68: use @action please

app/assets/javascripts/discourse/app/templates/invites/show.hbs
```diff
@@ -16,14 +16,26 @@
...
+          </p>
+        {{/unless}}
 
         <form>
+
+          {{#if isInviteLink}}
+            <div class="input email-input">
+              <label>{{i18n "user.email.title"}}</label>
+              {{input type="email" value=email id="new-account-email" name="email" autofocus="autofocus"}}
+              &nbsp;{{input-tip validation=emailValidation id="account-email-validation"}}
```
jjaffeux, line 34: not a big fan of &nbsp; could we add spacing through css ?

app/assets/javascripts/discourse/app/templates/user-invited-show.hbs
```diff
@@ -99,6 +115,19 @@
                         {{d-button icon="sync" action=(action "reinvite") actionParam=invite label="user.invited.reinvite"}}
                       {{/if}}
                     </td>
+                  {{else if inviteLinks}}
+                    <td>{{d-button icon="link" action=(route-action "editInvite" invite.invite_link) label="user.invited.copy_link"}}</td>
+                    <td>{{format-date invite.created_at}}</td>
+                    <td>{{number invite.redemption_count}} / {{number invite.max_redemption_allowed}}</td>
+                    <td colspan="2">{{ invite.group_names }}</td>
+                    <td>{{raw-date invite.invite_link_expires_at leaveAgo="true"}}</td>
```
jjaffeux, line 123: does true need to be a string here ?

app/controllers/users_controller.rb
techAPJ: We need to use different serializer for links tab and pending/redeemed tabs. So if the param is `links` we use `InviteLinkSerializer` and for pending/redeemed we use `InvitedSerializer` serializer.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
techAPJ: Yes, normal invites expires based on `invite_expiry_days` site setting. Invite Links can different expiry dates based on user input at the time of link create. Here is the method that decides if invite has expired based on invite type: https://github.com/discourse/discourse/pull/9813/files#diff-9592cb71c2bb6ae748d9ac2fc3596e86R66

techAPJ:
Note that the failing plugin specs is for [discourse-invite-tokens](http://github.com/discourse/discourse-invite-tokens) plugin. I have opened [a PR](https://github.com/discourse/discourse-invite-tokens/pull/4) to update failing specs for plugin and will merge that PR after this feature is merged in core.

app/assets/javascripts/discourse/app/templates/user-invited-show.hbs
techAPJ: Yes, seems like we need string here because that's what we are checking for.

https://github.com/discourse/discourse/blob/d615de9139aaede37dbb5f97d89e08b7541869ec/app/assets/javascripts/discourse/app/helpers/format-date.js#L15

spec/requests/invites_controller_spec.rb
techAPJ: Oops, fixed!

db/post_migrate/20200430173433_remove_user_id_from_invites.rb
techAPJ: Done.

app/jobs/regular/anonymize_user.rb
```diff
@@ -16,7 +16,11 @@ def execute(args)
     def make_anonymous
       anonymize_ips(@anonymize_ip) if @anonymize_ip
 
-      Invite.with_deleted.where(user_id: @user_id).destroy_all
+      Invite.includes(:invited_users)
+        .where('invited_users.user_id = ?', @user_id)
+        .references('invited_users')
+        .destroy_all
```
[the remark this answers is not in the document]
techAPJ: You're right. Fixed it to only delete `InvitedUser` record.

app/models/invite.rb
```diff
@@ -40,18 +51,30 @@ def user_doesnt_already_exist
     return if email.blank?
     user = Invite.find_user_by_email(email)
 
-    if user && user.id != self.user_id
+    if user && user.id != self&.user&.user_id
```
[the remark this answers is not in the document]
techAPJ: You're right about 1.

I think I wanted this to be `self.invited_users&.first&.user_id`. Because `Invite` model doesn't have direct relation to user.

app/models/invite.rb
```diff
@@ -160,6 +183,49 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
...
+  end
+
+  # redeem multiple use invite link
+  def redeem_invite_link(email: nil, username: nil, name: nil, password: nil, user_custom_fields: nil, ip_address: nil)
+    if is_invite_link? && !expired? && !redeemed? && !destroyed? && link_valid?
+      lower_email = Email.downcase(email)
+      user = User.find_by_email(lower_email)
+      raise UserExists.new I18n.t("invite.invite_link_user_exists") if user.present?
+
+      self.update_column(:email, email)
```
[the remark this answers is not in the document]
techAPJ: That has crossed my mind as well. What do you recommend here?

app/models/invite.rb
```diff
@@ -74,7 +97,7 @@ def self.invite_by_email(email, invited_by, topic = nil, group_ids = nil, custom
     )
   end
 
-  # generate invite link
+  # single use invite link
```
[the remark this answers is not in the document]
techAPJ: Done.

app/models/invite.rb
```diff
@@ -176,30 +242,64 @@ def self.get_group_ids(group_names)
...
+    find_all_pending_invites_from(inviter, 0, nil).reorder(nil).count
   end
 
-  def self.find_pending_invites_count(inviter)
-    find_all_invites_from(inviter, 0, nil).where('invites.user_id IS NULL').reorder(nil).count
+  def self.find_all_redeemed_invites_from(inviter, offset = 0, limit = SiteSetting.invites_per_page)
+    InvitedUser.includes(:invite)
+      .includes(user: :user_stat)
+      .includes(:user)
+      .where('invited_users.user_id IS NOT NULL')
```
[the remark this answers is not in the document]
techAPJ: I checked and `invited_users.user_id IS NOT NULL` is only used here and nowhere else. Can you point out where else did you see this condition?

Also `invited_users.user_id IS NOT NULL` is not same as `.single_use_invites` scope because it tells which invites have been redeemed. This condition is required to show all the redeemed invites from a particular user.

app/models/invite.rb
```diff
@@ -160,6 +183,49 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
...
+
+    if group_ids.present?
+      group_ids = group_ids - invite.invited_groups.pluck(:group_id)
+
+      group_ids.each do |group_id|
+        invite.invited_groups.create!(group_id: group_id)
+      end
+    end
+
+    invite.reload
```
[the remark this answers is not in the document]
techAPJ: Not needed, removed.

app/models/invite.rb
```diff
@@ -176,30 +242,64 @@ def self.get_group_ids(group_names)
     group_ids
   end
 
-  def self.find_all_invites_from(inviter, offset = 0, limit = SiteSetting.invites_per_page)
-    Invite.where(invited_by_id: inviter.id)
+  def self.find_all_pending_invites_from(inviter, offset = 0, limit = SiteSetting.invites_per_page)
+    Invite.single_use_invites
+      .includes(:invited_users)
```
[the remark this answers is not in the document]
techAPJ: You're right. Removed.

app/models/invite.rb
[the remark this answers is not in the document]
davidtaylorhq: I think `InviteRedeemer` will need updating so that it can be optionally be passed an email address. If it is supplied with an email address then it uses that instead of `invite.email`. 

Would that work?

app/models/invite.rb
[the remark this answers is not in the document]
davidtaylorhq: Oh yes you're right. I confused it with `invited_users.user_id IS NULL` :+1:

app/models/invite.rb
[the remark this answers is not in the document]
techAPJ: That sounds doable. Let me try implementing that.

app/assets/javascripts/discourse/app/templates/invites/show.hbs
techAPJ: Done.

app/models/invite.rb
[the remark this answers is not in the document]
techAPJ: Okay, I've updated `InviteRedeemer` to accept email and made it so that we pass email whenever `InviteRedeemer` is initialized.

techAPJ:
@davidtaylorhq Thanks for the review! Can you have a look again?

techAPJ:
Thanks for the detailed review @davidtaylorhq! ❤️

I've updated `InviteRedeemer` per your suggestion.

app/controllers/invites_controller.rb
```diff
@@ -100,7 +107,7 @@ def create
     end
   end
 
-  def create_invite_link
+  def create_single_use_invite_link
```
SamSaffron, line 110: this to me API wise is a bit confusing, why do we need a special end point for "single vs multiple" 

Isn't "single" simply an invite link with max_redemptions_allowed == 1 ?

app/controllers/users_controller.rb
SamSaffron: structure wise if the format is different would we go with "links.json" for links and "invites.json" for pending & redeemed? its just a bit strange the the json format you get depends on the param, makes it harder for @oblakeerickson to document the API

app/models/invite.rb
```diff
@@ -160,6 +183,44 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
     invite
   end
 
+  # multiple use invite link
+  def self.generate_multiple_use_invite_link(invited_by, opts = nil)
```
SamSaffron, line 187: again the question is... is this `generate_invite_link`?

also ... lets avoid ... opts = nil in favour of being explicit: `invited_by, group_ids: nil, max_redemptions_allowed: 1` and so on.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
SamSaffron: sorry ... what I meant ... why not call this column `expires_at` ? `invite_link_expires_at` is a bit strange.

SamSaffron: also ... we could either leave this as null and defer to site setting, or set it in stone on creation.

app/controllers/invites_controller.rb
techAPJ: You're correct. I think I can replace these two methods with a single method `create_invite_link` wherein single use and multiple use can be differentiated via a param.

app/models/invite.rb
techAPJ: The two methods `generate_single_use_invite_link` and `generate_multiple_use_invite_link` are doing very different things. `generate_single_use_invite_link` is falling back to `create_invite_by_email` method for doing all the heavy lifting while `generate_multiple_use_invite_link` is a standalone method and do not share any code with `create_invite_by_email`.  So combining these two methods will not be trivial.

Agreed with `opts = nil`. I'll replace this with keyword arguments.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
techAPJ: I'll rename `invite_link_expires_at` with `expires_at`. Noted regarding the options for single use invite.

app/controllers/users_controller.rb
techAPJ: Understood. I'll differentiate the routes based on JSON response.

app/controllers/invites_controller.rb
techAPJ: Done.

app/models/invite.rb
techAPJ: Updated `generate_multiple_use_invite_link` method to use keyword arguments.

app/controllers/users_controller.rb
techAPJ: This is now done.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
techAPJ: Renamed `invite_link_expires_at` to `expires_at`.

db/migrate/20200430072846_create_invited_users.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateInvitedUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :invited_users do |t|
+      t.integer :user_id
```
tgxworld, line 6: Can this column be empty? Do we need a `null: false` check here? Also we should probably add an index on this column as well.

db/migrate/20200430072846_create_invited_users.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.integer :invite_id
```
tgxworld, line 7: Same as https://github.com/discourse/discourse/pull/9813/files#r432265656

tgxworld:
If I click on copy invite link without changing any of the settings, the link will actually expire in a month even though I didn't select any expiry. 

Screenshot from 2020-05-29 14-55-31

tgxworld:
It is kind of odd to enable search for the dropdown here since the values aren't really searchable. 

Screenshot from 2020-05-29 14-59-05

tgxworld:
When I clock on copy link, it pops up the modal saying link is successfully generated. However, it was already generated and we're not really generating a new one. That might be confusing for users.

Screenshot from 2020-05-29 15-04-17

tgxworld:
We should probably add a max for the input field on the client side.

Screenshot from 2020-05-29 15-05-36

tgxworld:
`Invited Via` is abit odd here because clicking on the link brings me to the invite link which will show an error as I'm currently logged in.

Screenshot from 2020-05-29 15-13-22

tgxworld:
Visiting a bulk invite link when logged in shows the following error message. However, there should be no user attached to a bulk invite link?

Screenshot from 2020-05-29 15-14-35

tgxworld:
I'm able to generate a bulk invite link even when number of allowed users is set to 0

Screenshot from 2020-05-29 15-17-22

tgxworld:
If I set num of allowed users to 1, I am shown a link but the link is never created.

Screenshot from 2020-05-29 15-18-57

tgxworld:
The copy seems alittle weird when all the redemptions have been used up. It says `your invite` has been redeemed which would be weird for a user that is using the invite link for the first time.  

Screenshot from 2020-05-29 15-21-04

tgxworld:
I think we should just remove the row here once the invite link has been removed.

Screenshot from 2020-05-29 15-23-01

app/controllers/invites_controller.rb
```diff
@@ -101,27 +108,42 @@ def create
...
-    guardian.ensure_can_invite_to!(topic) if topic.present?
-
     group_ids = groups.map(&:id)
 
-    invite_exists = Invite.where(email: params[:email], invited_by_id: current_user.id).first
-    if invite_exists && !guardian.can_send_multiple_invites?(current_user)
-      return render json: failed_json, status: 422
+    if is_single_invite
+      topic = Topic.find_by(id: params[:topic_id])
+      guardian.ensure_can_invite_to!(topic) if topic.present?
```
tgxworld, line 127: If `params[:topic_id]` is blank, we probably don't have to execute the query.

app/models/invite.rb
```diff
@@ -160,6 +183,38 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
     invite
   end
 
+  # multiple use invite link
+  def self.generate_multiple_use_invite_link(invited_by:, max_redemptions_allowed: 5, expires_at: 1.month.from_now, group_ids: nil)
+    create_args = {
+      invited_by: invited_by,
+      max_redemptions_allowed: max_redemptions_allowed.to_i,
```
tgxworld, line 190: We'll need to validate that this falls into a reasonable set of values.

app/models/invite.rb
```diff
@@ -40,18 +51,30 @@ def user_doesnt_already_exist
...
   end
 
+  def is_invite_link?
+    max_redemptions_allowed > 1
+  end
+
   def redeemed?
-    redeemed_at.present?
+    if is_invite_link?
+      redemption_count >= max_redemptions_allowed
```
tgxworld, line 66: I think we need to be very careful here with regards to concurrency. The `redemption_count` is accurate at the point where the invite was loaded. In that time, other Rails processes can also be updating the count which means we might burst the redemption limit when alot of people are redeeming invites at the same time.

app/serializers/invite_link_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class InviteLinkSerializer < ApplicationSerializer
+  attributes :id, :invite_link, :created_at, :max_redemptions_allowed, :redemption_count, :expires_at, :group_names
+
+  def invite_link
+    "#{Discourse.base_url}/invites/#{object.invite_key}"
```
tgxworld, line 7: Is `Discourse.base_url` available on the client side? If so, we can avoid sending down the same string multiple times and have the client form it.

app/serializers/invite_link_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def group_names
+    object.groups.pluck(:name).join(", ")
```
tgxworld, line 11: We need to preload the `groups` here otherwise it is going to result in N+1

app/serializers/invite_serializer.rb
```diff
@@ -1,21 +1,39 @@
...
-  attributes :email, :updated_at, :redeemed_at, :expired, :user
+  attributes :id, :email, :updated_at, :redeemed_at, :expired, :user, :invite_source
 
   def include_email?
-    options[:show_emails] && !object.redeemed?
+    options[:type] == "pending" && options[:show_emails] && !object.redeemed?
+  end
+
+  def redeemed_at
+    options[:type] == "pending" ? nil : object.redeemed_at
```
tgxworld, line 11: We should avoid shipping the column down if it is `nil`

app/serializers/invite_serializer.rb
```diff
@@ -1,21 +1,39 @@
...
     ser.as_json
   end
 
+  def include_invite_source?
+    options[:type] == "redeemed"
+  end
+
+  def invite_source
+    if object.invite.is_invite_link?
+      "<a href='#{Discourse.base_url}/invites/#{object.invite.invite_key}' target='_blank'>#{object.invite.invite_key[0..5]}</a>"
```
tgxworld, line 34: Hmm crafting HTML in a serializer on the server side is a red flag to me. Can we do this on the client side instead?

db/migrate/20200430072846_create_invited_users.rb
tgxworld, line 7:
```suggestion
      t.integer :user_id
      t.integer :invite_id
```

When both columns are not null, do we need to ensure uniqueness?

db/post_migrate/20200430173433_remove_user_id_from_invites.rb
```diff
@@ -0,0 +1,29 @@
...
+        redeemed_at,
+        created_at,
+        updated_at
+      )
+      SELECT user_id, id, redeemed_at, created_at, updated_at
+      FROM invites WHERE user_id IS NOT NULL
+    SQL
+
+    execute <<~SQL
+      ALTER TABLE invites DROP COLUMN IF EXISTS user_id
```
tgxworld, line 18: If we don't drop this column, will there be any side effects? Because this migration is irreversible, I'll recommend keeping the column around first as backup until we're confident everything is working well.

db/post_migrate/20200430173433_remove_user_id_from_invites.rb
```diff
@@ -0,0 +1,29 @@
...
+    SQL
+
+    execute <<~SQL
+      ALTER TABLE invites DROP COLUMN IF EXISTS redeemed_at
```
tgxworld, line 22: Same as https://github.com/discourse/discourse/pull/9813/files#r432354568

lib/guardian.rb
```diff
@@ -380,7 +380,7 @@ def can_invite_via_email?(object)
   end
 
   def can_bulk_invite_to_forum?(user)
-    user.admin?
+    authenticated? && user.admin?
```
tgxworld, line 383: Hmm why do we need `authenticated?` here?

db/migrate/20200430072846_create_invited_users.rb
techAPJ: Yes, `user_id` is empty initially, as per: https://github.com/discourse/discourse/pull/9813/files#diff-2a355fe6078d7a0ee0e440532e4032e2R102

Will add index.

db/migrate/20200430072846_create_invited_users.rb
techAPJ: Done.

db/migrate/20200430072846_create_invited_users.rb
```diff
@@ -0,0 +1,14 @@
...
+class CreateInvitedUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :invited_users do |t|
+      t.integer :user_id
+      t.integer :invite_id, null: false
+      t.datetime :redeemed_at
+      t.timestamps null: false
+    end
+
+    add_index :invited_users, [:user_id, :invite_id], unique: true
```
tgxworld, line 12: We only need the index when `user_id` is not null if I'm not wrong.

techAPJ: ah yes, you're right. can't add index here because user_id can be null.

app/controllers/invites_controller.rb
```diff
@@ -40,12 +41,16 @@ def show
...
     params.require(:id)
-    params.permit(:username, :name, :password, :timezone, user_custom_fields: {})
+    params.permit(:email, :username, :name, :password, :timezone, user_custom_fields: {})
     invite = Invite.find_by(invite_key: params[:id])
 
     if invite.present?
       begin
-        user = invite.redeem(username: params[:username], name: params[:name], password: params[:password], user_custom_fields: params[:user_custom_fields], ip_address: request.remote_ip)
+        user = if invite.is_invite_link?
+          invite.redeem_invite_link(email: params[:email], username: params[:username], name: params[:name], password: params[:password], user_custom_fields: params[:user_custom_fields], ip_address: request.remote_ip)
```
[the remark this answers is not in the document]
techAPJ: Fixed.

app/models/invited_user.rb
```diff
@@ -0,0 +1,18 @@
...
+  belongs_to :invite
+end
+
+# == Schema Information
+#
+# Table name: invited_users
+#
+#  id          :bigint           not null, primary key
+#  user_id     :integer
+#  invite_id   :integer
```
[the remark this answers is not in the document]
techAPJ: Nope, added not null.

db/migrate/20200430072846_create_invited_users.rb
techAPJ: Since only one column (`invite_id`) is not null we can't check for uniqueness.

app/models/invite.rb
techAPJ: Done.

app/models/invite.rb
```diff
@@ -160,6 +183,38 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
...
+      expires_at: expires_at,
+      emailed_status: emailed_status_types[:not_required]
+    }
+    invite = Invite.create!(create_args)
+
+    if group_ids.present?
+      group_ids = group_ids - invite.invited_groups.pluck(:group_id)
+
+      group_ids.each do |group_id|
+        invite.invited_groups.create!(group_id: group_id)
```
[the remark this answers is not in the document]
techAPJ: Implemented both suggestions.

app/models/invite_redeemer.rb
```diff
@@ -86,35 +86,42 @@ def process_invitation
...
-    if count == 1
+    @invited_user_record = InvitedUser.create(invite_id: invite.id, redeemed_at: Time.zone.now)
+    if invite.is_invite_link? && @invited_user_record.present?
+      Invite.increment_counter(:redemption_count, invite.id)
+    elsif @invited_user_record.present?
       delete_duplicate_invites
     end
 
-    count
+    @invited_user_record.present? ? 1 : 0
```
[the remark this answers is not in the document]
techAPJ: Nothing specific, was keeping the change inline with old code. I've updated the code now to return `true`/`false` instead.

app/models/invite_redeemer.rb
```diff
@@ -153,6 +160,11 @@ def notify_invitee
   end
 
   def delete_duplicate_invites
-    Invite.where('invites.email = ? AND redeemed_at IS NULL AND invites.id != ?', invite.email, invite.id).delete_all
+    Invite.single_use_invites
+      .includes(:invited_users)
+      .where('invites.email = ? AND invites.id != ?', email, invite.id)
+      .where('invited_users.user_id IS NULL')
+      .references('invited_users')
```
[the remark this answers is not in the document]
techAPJ: I believe you meant -- `Invite.single_user_invites.joins("LEFT JOIN invited_users ON invites.id = invited_users.invited_id").where('invited_users.user_id IS NULL')`?

I've made the above change.

db/migrate/20200430072846_create_invited_users.rb
tgxworld: O no that wasn't what I meant. You can add a `where` clause when adding an index. See https://github.com/discourse/discourse/blob/921751f128f989cc58bc570509977de4590f2835/db/migrate/20150513094042_add_index_on_post_actions.rb#L5

techAPJ: Nice, thanks for the tip! Made the required changes.

db/post_migrate/20200430173433_remove_user_id_from_invites.rb
techAPJ: No side effects. Makes sense. I've made the required changes.

lib/guardian.rb
techAPJ: Because it checks for presence of `user`, and `user.admin?` raises error if `user` is not present.

techAPJ: Actually, we don't need to check for presence of user anymore because I updated `create_invite_link` method to check for `requires_login`. Will update accordingly.

app/models/invite.rb
techAPJ: I see. What should we do here to add more safeguards?

spec/requests/invites_controller_spec.rb
```diff
@@ -203,6 +203,52 @@
     end
   end
 
+  context '#create_multiple_use_invite_link' do
```
tgxworld, line 206:
```suggestion
  context 'create multiple use invite link' do
```

`#` is meant to represent an instance method

spec/requests/invites_controller_spec.rb
```diff
@@ -203,6 +203,52 @@
...
+      end
+
+      it "allows multiple group invite" do
+        Fabricate(:group, name: "security")
+        Fabricate(:group, name: "support")
+        sign_in(Fabricate(:admin))
+
+        post "/invites/link.json", params: {
+          max_redemptions_allowed: 5,
+          group_names: "security,support"
```
tgxworld, line 243: Hmm why do we allow group names to be used? A group can be renamed so the `Group#name` isn't as reliable as `Group#id`.

app/controllers/invites_controller.rb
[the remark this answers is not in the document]
tgxworld: It looks like we tackled `expired?` but there is `redeemed?`, `destroyed?` and `link_valid?` too.

spec/requests/invites_controller_spec.rb
```diff
@@ -135,7 +135,7 @@
     end
   end
 
-  context '#create_invite_link' do
+  context '#create_single_use_invite_link' do
```
tgxworld, line 138: `#` is meant to represent an instance method so I would probably rearrange this to 

```
describe "#create_invite_link" do
  describe "single use invite link" do
    ...
  end 
  describe "multiple use invite link" do
    ...
  end
end
```

spec/requests/invites_controller_spec.rb
```diff
@@ -203,6 +203,52 @@
...
+        sign_in(Fabricate(:trust_level_4))
+        post "/invites/link.json", params: {
+          max_redemptions_allowed: 5
+        }
+        expect(response).to be_forbidden
+      end
+
+      it "allows admins to invite to groups" do
+        group = Fabricate(:group)
+        sign_in(Fabricate(:admin))
```
tgxworld, line 225: Minor but it looks like there is an opportunity to `fab!(:admin)` once and share it will all the tests.

spec/requests/invites_controller_spec.rb
```diff
@@ -250,6 +296,21 @@
       end
     end
 
+    context 'with an expired invite' do
+      fab!(:admin) { Fabricate(:admin) }
+      fab!(:invite_link) { Fabricate(:invite, email: nil, max_redemptions_allowed: 5, expires_at: 1.day.ago, emailed_status: Invite.emailed_status_types[:not_required]) }
+
+      it "response is not success" do
```
tgxworld, line 303:
```suggestion
      it "response is not successful" do
```

app/controllers/invites_controller.rb
```diff
@@ -40,12 +41,16 @@ def show
 
   def perform_accept_invitation
```
tgxworld, line 42: It looks like we don't have tests for the `redeem_invite_link` flow.

app/assets/javascripts/discourse/app/templates/components/invite-link-panel.hbs
```diff
@@ -0,0 +1,59 @@
...
+        groupFinder=groupFinder
+        groupNames=inviteModel.groupNames
+        placeholderKey="topic.invite_private.group_name"}}
+    </div>
+
+    <div class="max-redemptions-allowed">
+      <label class="instructions">
+        {{i18n 'user.invited.invite_link.max_redemptions_allowed_label'}}
+      </label>
+      {{input type="number" value=maxRedemptionAllowed class="max-redemptions-allowed-input" min="2" max="1000000"}}
```
tgxworld, line 27: Should we use the site settings for `1000000` instead?

app/models/invite.rb
```diff
@@ -160,6 +182,39 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
     invite
   end
 
+  def self.generate_multiple_use_invite_link(invited_by:, max_redemptions_allowed: 5, expires_at: 1.month.from_now, group_ids: nil)
+    max_redemptions_allowed = max_redemptions_allowed.to_i
+    unless max_redemptions_allowed.between?(2, SiteSetting.invite_link_max_redemptions_limit)
```
tgxworld, line 187: I don't think this is the right place to do this check as it assumes all multiple use invite link has to go through this code path. Instead, I think this should be an active record validation.

app/models/invite_redeemer.rb
```diff
@@ -86,35 +86,42 @@ def process_invitation
   end
 
   def invite_was_redeemed?
-    # Return true if a row was updated
-    mark_invite_redeemed == 1
+    mark_invite_redeemed == true
```
tgxworld, line 89:
```suggestion
    mark_invite_redeemed
```

app/models/invited_user.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class InvitedUser < ActiveRecord::Base
+  belongs_to :user
+  belongs_to :invite
```
tgxworld, line 5: We should add active record validations for `invite_id`

tgxworld, line 5: We should add active record validations for `invite_id`. Mainly presence and uniqueness.

app/models/user.rb
```diff
@@ -77,6 +77,7 @@ class User < ActiveRecord::Base
   has_many :topics_allowed, through: :topic_allowed_users, source: :topic
   has_many :groups, through: :group_users
   has_many :secure_categories, through: :groups, source: :categories
+  has_many :invites, through: :invited_users
```
tgxworld, line 80: This is a breaking change because we're no longer deleting invites when the user is deleted.

app/serializers/invite_serializer.rb
```diff
@@ -1,21 +1,39 @@
...
-  attributes :email, :updated_at, :redeemed_at, :expired, :user
+  attributes :id, :email, :updated_at, :redeemed_at, :expired, :user, :invite_source
 
   def include_email?
-    options[:show_emails] && !object.redeemed?
+    options[:type] == "pending" && options[:show_emails] && !object.redeemed?
+  end
+
+  def include_redeemed_at?
+    options[:type] == "redeemed"
```
tgxworld, line 11: It'll probably be a good idea to move `redeemed` and `pending` into constants so we can avoid repeatedly hardcoding the strings in multiple places making it hard to change in the future.

app/serializers/invite_serializer.rb
```diff
@@ -1,21 +1,39 @@
 # frozen_string_literal: true
 
 class InviteSerializer < ApplicationSerializer
```
tgxworld, line 3: It seems like we need to rename this serializer since it isn't a serializer for the `Invite` model anymore and is a serializer for `InvitedUser`

config/locales/server.en.yml
```diff
@@ -235,6 +235,10 @@ en:
     max_rows: "First %{max_bulk_invites} invites has been sent. Try splitting the file in smaller parts."
     error: "There was an error uploading that file. Please try again later."
 
+  invite_link:
+    email_taken: "This email is already in use. Please use different email address."
```
tgxworld, line 239: This error message doesn't seem right to me because if an email is already in use, it means the user has signed up before. So they should either reset password or be redirected to the login page instead of being ask to use a different email.

spec/models/invite_redeemer_spec.rb
```diff
@@ -176,7 +182,41 @@
...
+
+    context 'invite_link' do
+      fab!(:invite_link) { Fabricate(:invite, max_redemptions_allowed: 5, expires_at: 1.month.from_now, emailed_status: Invite.emailed_status_types[:not_required]) }
+      let(:invite_redeemer) { InviteRedeemer.new(invite: invite_link, email: 'foo@example.com') }
+
+      it 'works as expected' do
+        user = invite_redeemer.redeem
+        invite_link.reload
+
+        expect(user.is_a?(User)).to eq(true)
```
tgxworld, line 196: We don't really need this check. Subsequent tests will fail if this isn't true.

db/migrate/20200430072846_create_invited_users.rb
```diff
@@ -0,0 +1,26 @@
...
+    execute <<~SQL
+      INSERT INTO invited_users (
+        user_id,
+        invite_id,
+        redeemed_at,
+        created_at,
+        updated_at
+      )
+      SELECT user_id, id, redeemed_at, created_at, updated_at
+      FROM invites WHERE user_id IS NOT NULL
```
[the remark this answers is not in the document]
techAPJ: Nope. We only need to migrate "redeemed" invites to `invited_users` table. If `user_id` is null that means invite has not been redeemed yet.

app/models/invite.rb
tgxworld: I'll probably go with a mutex around the redemption.

db/migrate/20200430072846_create_invited_users.rb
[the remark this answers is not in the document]
tgxworld: Hmm then shouldn't we check for `redeemed_at IS NULL` instead?

db/migrate/20200430072846_create_invited_users.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateInvitedUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :invited_users do |t|
+      t.integer :user_id
+      t.integer :invite_id, null: false
```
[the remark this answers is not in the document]
techAPJ: Added. :)

app/models/invite.rb
```diff
@@ -5,6 +5,12 @@ class UserExists < StandardError; end
   include RateLimiter::OnCreateRecord
   include Trashable
 
+  # TODO(2021-05-22): remove
+  self.ignored_columns = %w{
+    user_id
+    redeemed_at
```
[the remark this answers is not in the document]
techAPJ: Done.

db/migrate/20200430072846_create_invited_users.rb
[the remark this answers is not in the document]
techAPJ: Update the query to check for `redeemed_at IS NOT NULL` as well.

app/controllers/invites_controller.rb
```diff
@@ -40,12 +41,16 @@ def show
...
     invite = Invite.find_by(invite_key: params[:id])
 
-    if invite.present?
+    if invite.present? && !invite.expired?
       begin
-        user = invite.redeem(username: params[:username], name: params[:name], password: params[:password], user_custom_fields: params[:user_custom_fields], ip_address: request.remote_ip)
+        user = if invite.is_invite_link?
+          invite.redeem_invite_link(email: params[:email], username: params[:username], name: params[:name], password: params[:password], user_custom_fields: params[:user_custom_fields], ip_address: request.remote_ip)
+        else
+          invite.redeem(username: params[:username], name: params[:name], password: params[:password], user_custom_fields: params[:user_custom_fields], ip_address: request.remote_ip)
```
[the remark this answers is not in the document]
techAPJ: We were already testing for everything except `link_valid?`, added check for that too. Note that `!redeemed?` and `!destroyed?` is being checked for in `show` method which is a precursor to `perform_accept_invitation`.

app/models/user.rb
techAPJ: But we don't really want to delete an invite in case the invite is an invite link which can be redeemed by multiple users.

spec/requests/invites_controller_spec.rb
techAPJ: We have been allowing group names ever since we started supporting adding group access for invites.

app/serializers/invite_serializer.rb
techAPJ: That is not entirely correct. It serializes `Invite` model in case of "pending" tab. I'd prefer to leave it as is.

app/models/invite.rb
techAPJ: Done.

techAPJ: Actually I had to revert this change because single use invite has `max_redemptions_allowed` default to `1` which will fail active record validation. I added another active record validation which checks for numericality and verifies that value is greater than 0.

spec/requests/invites_controller_spec.rb
techAPJ: Agreed, done.

app/assets/javascripts/discourse/app/templates/components/invite-link-panel.hbs
techAPJ: We should, done.

app/serializers/invite_serializer.rb
tgxworld: Ooo this is confusing because a serializer can serialize two different models. I think we should split the serializer into two and call the right serializer in the controller rather than have one serializer handle this.

app/models/user.rb
tgxworld: What about the previous case where a user sends an invite to a particular email? We used to delete those when the user is deleted but we don't do that anymore now.

app/models/invite.rb
tgxworld: We should only run the validation for multiple invite links so that it doesn't affect single use invites.

app/controllers/invites_controller.rb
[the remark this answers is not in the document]
tgxworld: I feel like the flow of logic here is very complex because the conditions for `Invite#redeem_invite_link` and `Invite#redeem` is duplicated in the model and in the controller. Perhaps we can check that `user` is `nil` and return an error message instead?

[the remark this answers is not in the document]
tgxworld: > Note that !redeemed? and !destroyed? is being checked for in show method which is a precursor to perform_accept_invitation.

The method can be called directly via API as well so it doesn't always go through the `show` method.

db/post_migrate/20200602034543_make_invite_user_id_redeemed_at_read_only.rb
```diff
@@ -0,0 +1,11 @@
+class MakeInviteUserIdRedeemedAtReadOnly < ActiveRecord::Migration[6.0]
```
tgxworld, line 1: This should be done before the migration to copy `invites` over to `invited_users`. Otherwise, the running applications will still be writing to the `invites` table before the new code changes have been deployed.

app/models/invited_user.rb
techAPJ: I've added validation for presence of `invite_id` but not sure if we should check for uniqueness because there will be multiple records for invite redemptions for multiple use invites.

tgxworld: Oops I meant adding a uniqueness validation for invited id scoped to the user_id where user _id is not null. The logic should be the same as the uniqueness index which was added.

db/post_migrate/20200602034543_make_invite_user_id_redeemed_at_read_only.rb
techAPJ: Okay, I moved the migration block here just after making the columns readonly.

app/controllers/invites_controller.rb
techAPJ: Added a test.

app/controllers/invites_controller.rb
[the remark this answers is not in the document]
techAPJ: > Perhaps we can check that user is nil and return an error message instead?

Agreed. Updated the code to send response based on user record presence.

app/serializers/invite_serializer.rb
techAPJ: Okay, I split the serializer in two. Which serializer to use will be decided in `InvitedSerializer` based on object type.

app/serializers/invite_serializer.rb
techAPJ: Since we now have two different serializers for pending/redeemed no need to check for type anymore.

app/models/invited_user.rb
techAPJ: Understood, done.

config/locales/server.en.yml
techAPJ: Updated the message to reflect they should login or reset password.

app/controllers/invites_controller.rb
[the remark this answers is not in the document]
techAPJ: Updated the code to not return success response if user is `nil`.

app/models/invite.rb
techAPJ: To identify if an invite is multiple invite link we check for `max_redemptions_allowed > 1`... so how can we add condition of the same field that we are validating?

db/post_migrate/20200602034543_make_invite_user_id_redeemed_at_read_only.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class MakeInviteUserIdRedeemedAtReadOnly < ActiveRecord::Migration[6.0]
```
tgxworld, line 3: Sorry one more comment here. The migration name should reflect the migration of data here instead of marking the columns as readonly. Marking the columns as readonly is the side effect of having to migrate the data.

app/controllers/invites_controller.rb
```diff
@@ -40,24 +41,29 @@ def show
...
+        else
+          response = { success: false, message: I18n.t('invite.not_found_json') }
         end
 
-        response = { success: true }
         if user.present? && user.active?
           topic = invite.topics.first
           response[:redirect_to] = topic.present? ? path("#{topic.relative_url}") : path("/")
-        else
-          response[:message] = I18n.t('invite.confirm_email')
```
tgxworld, line 60: Just confirming that this is no longer required? Can the `user` still be `present?` but not `active?`

app/models/invite.rb
```diff
@@ -13,12 +19,15 @@ class UserExists < StandardError; end
...
 
+  has_many :invited_users
+  has_many :users, through: :invited_users
   has_many :invited_groups
   has_many :groups, through: :invited_groups
   has_many :topic_invites
   has_many :topics, through: :topic_invites, source: :topic
   validates_presence_of :invited_by_id
   validates :email, email: true, format: { with: EmailValidator.email_regex }, allow_blank: true
+  validates :max_redemptions_allowed, numericality: { only_integer: true, greater_than: 0 }
```
tgxworld, line 30: We need a custom validation here too. If it is a multi-invite link, it has to be greater than 2 and smaller than the site setting.

app/models/invite.rb
tgxworld: I thought we were leaving `max_redemptions_allowed` blank for single invite so that the database default would kick in?

tgxworld: We have `add_column :invites, :max_redemptions_allowed, :integer, null: false, default: 1` so single invite should just leave `max_redemptions_allowed` blank

app/controllers/invites_controller.rb
techAPJ: Yes, that can be the case, that is why I moved this block to `user.present?` check just above, as per: https://github.com/discourse/discourse/pull/9813/files#diff-322a08e87c3cc46449eac6235452f266R59

app/models/invite.rb
techAPJ: Does this validation looks good?

```
validates :max_redemptions_allowed,
    allow_blank: true,
    numericality: { only_integer: true, greater_than_or_equal_to: 2, less_than_or_equal_to: SiteSetting.invite_link_max_redemptions_limit, allow_blank: true }
```

I am having trouble with above validation clause when creating single use invite link. It's failing with `Validation failed: Max redemptions allowed must be greater than or equal to 2` even when `max_redemptions_allowed` is left blank.

tgxworld: I’ll recommend writing a custom validation instead of trying to work with the Rails options here. Much easier to control what we want to do. You can add a custom validation with validate :some_method

app/models/user.rb
techAPJ: That is true. What do you suggest here?

app/controllers/invites_controller.rb
tgxworld: I think something is still off here. If `user.present?` is true, we'll always set `message: I18n.t('invite.confirm_email')`. However, if a `user.active?` the message shouldn't be set. You can see this in the payload when running the following spec `describe "directs to the first topic the user was invited to"`

app/models/user.rb
tgxworld: Actually 

> But we don't really want to delete an invite in case the invite is an invite link which can be redeemed by multiple users.

Why do we not want to delete an invite link which can be redeemed by multiple users? If the user that created the link has been deleted, the any invite links which has been created by the user should not be valid.

app/controllers/invites_controller.rb
techAPJ: Good catch. I've fixed this case.

app/models/invite.rb
techAPJ: I tried adding custom validation but when I check for `max_redemptions_allowed.present?` it returns true even for single use invites. It seems like `max_redemptions_allowed` defaults to 1 at the time of custom validation, even when I have not set it explicitly.

tgxworld: O I think there is some active record magic here. Can you try setting `max_redemptions_allowed: nil` for single invites?

techAPJ: In that case I'm getting `ActiveRecord::NotNullViolation`. I am really inclined to keep this check in `generate_multiple_use_invite_link` since we're not going to create multi use invites anywhere else anytime soon. Also, we are already checking for numericality and min value of `1` in active record validation.

tgxworld: > I am really inclined to keep this check in `generate_multiple_use_invite_link` since we're not going to create multi use invites anywhere else anytime soon.

The rational for adding this as an ActiveRecord validation is because we don't know who/when will come along and create an invite link in the future without going through the proper methods. By having it as a ActiveRecord validation, we get to ensure consistency before the record hits the database. Can you see if the following works? 

```
  validate :ensure_max_redemptions_allowed

  def ensure_max_redemptions_allowed
    if self.max_redemptions_allowed.nil?
      self.max_redemptions_allowed = 1
    else
      # validation
    end
  end
```

tgxworld: The validation here is really to ensure that if someone provides us with a value of more than `1`,  we need to ensure that it falls between `2` and `SiteSetting.max`

tgxworld: Actually, can't we use the lack of an `email` to differentiate between a multi invite link vs a single invite link? Or can there be a single invite link? If so, then we just have to validate that the value falls between 1 and SiteSetting max.

techAPJ: We can't check for `email` because we still have official `discourse-invite-tokens` plugin which allows maximum redemption of `1` but still has no email attached to it.

I'll try your validation which checks for range `1..SiteSetting max`.

techAPJ: Done.

app/models/user.rb
techAPJ: The `has_many` association in `users` table refers to `user_id` column in `invites` table which <s>is</s> was the ID of the _invited_ user. So it seems like when the user is deleted the corresponding invite record which was redeemed is deleted. This has always been the case.

Now when a user is deleted we'll delete the corresponding `invited_users` record.

tgxworld: Thank you for clarifying. Let me try to explain where I'm confused.

1. As a staff user, I can generate an invite link. That invite link is created as an `Invite` record with the `invited_by_id` set to the staff user. When the staff user is deleted, we should delete the invite record so that the link is no longer valid.

1. On the `User` model, we have `has_many :invited_users, dependent: :delete_all`. This means `SELECT * FROM invited_users WHERE invited_users.user_id = <current user id>`. The `InvitedUser` table as I understand is to maintain the link between the invite and the user that was invited through that invite. A user can only be invited once so how can a user `has_many :invited_users`?

1. `has_many :invites, through: :invited_users`. Since we're going through the `invited_users` table, this is looking for the invite that the current user was invited from which I think is wrong. Instead shouldn't a user's relationship to `invites` be `has_many :invites, foreign_key: :invited_by_id`?

techAPJ: I think you're correct. We should link invites that user has created, not redeemed. I'll make the required changes.

techAPJ: Done.

app/models/invite.rb
```diff
@@ -160,6 +183,34 @@ def self.create_invite_by_email(email, invited_by, opts = nil)
...
+        InvitedGroup.insert_all(invited_groups)
+      end
+
+      "#{Discourse.base_url}/invites/#{invite.invite_key}"
+    end
+  end
+
+  # redeem multiple use invite link
+  def redeem_invite_link(email: nil, username: nil, name: nil, password: nil, user_custom_fields: nil, ip_address: nil)
+    if is_invite_link? && !expired? && !redeemed? && !destroyed? && link_valid?
```
tgxworld, line 208: @techAPJ I suggest we wrap this in a distributed mutex and within the mutex reload the invite object so that we get the latest `redemption_count`.

techAPJ: Done.

techAPJ:
Made a few more changes:

- Wrapped invite link redemption code in `DistributedMutex`.
- Made it so that both admin and moderators can send invite link.
- Updated `Invite.expired?` to check against `expires_at` column for both single and multiple use invites. Also wrote a migration to update `expires_at` for existing invites.

app/models/invite.rb
techAPJ: Done.

app/models/user.rb
```diff
@@ -59,6 +59,7 @@ class User < ActiveRecord::Base
   has_many :group_requests, dependent: :delete_all
   has_many :muted_user_records, class_name: 'MutedUser', dependent: :delete_all
   has_many :ignored_user_records, class_name: 'IgnoredUser', dependent: :delete_all
+  has_many :invited_users, dependent: :delete_all
```
tgxworld, line 62: This should be a `has_one`? A user can only be invited once.

app/serializers/invited_user_record_serializer.rb
```diff
@@ -0,0 +1,54 @@
...
+  def include_posts_read_count?
+    scope.can_see_invite_details?(invited_by)
+  end
+
+  def days_since_created
+    ((Time.now - object.created_at) / 60 / 60 / 24).ceil
+  end
+
+  def include_days_since_created
+    scope.can_see_invite_details?(invited_by)
```
tgxworld, line 51: Can we memoize this? We're doing the same check multiple times.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddBulkInviteLinkToInvites < ActiveRecord::Migration[6.0]
+  def change
+    add_column :invites, :max_redemptions_allowed, :integer, null: false, default: 1
+    add_column :invites, :redemption_count, :integer, null: false, default: 0
+    add_column :invites, :expires_at, :datetime, null: true
+
+    execute <<~SQL
+      UPDATE invites SET expires_at = updated_at + INTERVAL '#{SiteSetting.invite_expiry_days} days'
```
tgxworld, line 10: We need to hardcode `SiteSetting.invite_expiry_days` here otherwise this migration will break when the site setting is removed in the futuer.

tgxworld:
@techAPJ Feel free to merge once the latest comments have been resolved.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
techAPJ: We can hardcode the default value of `30`, but what about sites that have modified the site setting?

app/models/user.rb
techAPJ: Actually a user can be invited multiple times via "bulk invite from file" feature where an invite is sent to grant user group access.

tgxworld: Hmm how is that possible when we have a unique index on `:invite_id` and `:user_id` on the `invited_users` table.

db/migrate/20200428102014_add_bulk_invite_link_to_invites.rb
tgxworld: O this isn't a new site setting. We can write a SQL query to fetch the value then.

app/models/user.rb
techAPJ: Aha, I think we updated the bulk invite code to grant access to groups to existing user and do not create new invite record.

I'll update the association to `has_one`.

PR #9815, opened 2020-05-18
FEATURE: Set custom document title separator

zcuric:
### This PR
- [x] Adds `document_title_separator` setting to update the title separator to enable different types of SEO titles, for example: `Test Topic | Discourse`. Defaults to `" - "`, as it was before. 
- [x] Adds locale option for this new option

Setting is an enum with list of separators inspired by [Yoast SEO WP plugin](https://yoast.com/wordpress/plugins/seo/). 

#### Note about tests
I've tried to write a meaningful test for this new option, but because the `qunit` creates it's own titles I coudn't verify in the test that this option applies. `qunit` titles are separated with commas, for example: `Test Topic,Qunit Discourse Test`.

### DIFF
--- a/app/assets/javascripts/discourse/app/routes/application.js
+++ b/app/assets/javascripts/discourse/app/routes/application.js
@@ -52,7 +52,10 @@ const ApplicationRoute = DiscourseRoute.extend(OpenComposer, {
       ) {
         tokens.push(this.shortSiteDescription);
       }
-      Discourse.set("_docTitle", tokens.join(" - "));
+      Discourse.set(
+        "_docTitle",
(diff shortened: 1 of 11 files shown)


jjaffeux:
hi, thanks. Can you tell why you think being able to customise this is important?

SamSaffron:
Sorry for the huge delay here. I am a bit mixed here.

@coding-horror  is this a level of configurability you want admins to have? If not we are going to need to expose this as a plugin extension instead. 

@zcuric recommend you open a topic on meta to discuss this.

zcuric:
As per discussion on Discourse Meta, closing this PR.

PR #9819, opened 2020-05-18
FEATURE: Redirect '/setup' to '/wizard'

markvanlan:
Based on some feedback it seems like `/setup` is an expected location for the wizard. This adds a redirect from `/setup` to `/wizard`.

### DIFF
--- a/app/controllers/wizard_controller.rb
+++ b/app/controllers/wizard_controller.rb
@@ -19,6 +19,10 @@ class WizardController < ApplicationController
     end
   end
 
+  def setup_redirect
+    redirect_to "/wizard"
+  end
+
   def qunit
     raise Discourse::InvalidAccess.new if Rails.env.production?
   end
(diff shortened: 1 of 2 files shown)


eviltrout:
Can you provide some context for this change?

markvanlan:
@eviltrout sure. @tobiaseigen was running through the wizard to setup a new instance, and was trying to re-run the wizard. His first instinct was to go to `/setup` to find the wizard. This makes a lot of sense to me, and I think we should redirect requests from `/setup` to `/wizard`.

eviltrout:
I don't think I agree with this one. Maybe open a dev topic with the idea and see if you can get buy in from others?

I'm just not sure how many people will randomly do what tobias did.

tobiaseigen:
I agree with Robin. My suggestion actually was to link to it from the admin
dashboard. Feel like I'm getting mixed messages about rerunning the
wizard.. happy to discuss further on meta.

PR #9820, opened 2020-05-18
UX: Wizard navigation button improvements

markvanlan:
The "Finish" button for the wizard is in a bit of an odd spot, and I think this UI is much better.

Current:
Screenshot from 2020-05-18 09-51-04

New: 
Screenshot from 2020-05-18 09-51-34

### DIFF
--- a/app/assets/javascripts/wizard/templates/components/wizard-step.hbs
+++ b/app/assets/javascripts/wizard/templates/components/wizard-step.hbs
@@ -34,12 +34,6 @@
       <a href {{action "quit"}} tabindex="11" class="action-link quit">{{i18n "wizard.quit"}}</a>
     {{/if}}
 
-    {{#if showFinishButton}}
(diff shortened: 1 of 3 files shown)


jjaffeux:
It's been done this way for a reason:

Screenshot 2020-05-19 at 13 19 01
Screenshot 2020-05-19 at 13 19 02

After first step when clicking next, you will end up with finish under your mouse and might click it, the current behavior prevents this.

markvanlan:
@jjaffeux It looks sooo much better though... Maybe we just have a confirm bootbox on the `Finish & Exit` button?

jjaffeux:
I don't know; this doesn't bother me like this and we never had any complaint about it since we made it. I just told you why it's this way :p I don't have a definite answer and it's not my call to make.

markvanlan:
@jjaffeux I will just close this for now

PR #9835, opened 2020-05-19
FIX: adds missing tags to published page header

jjaffeux:


jjaffeux:
@eviltrout I don't see any reason to not just reuse the _head partial

PR #9838, opened 2020-05-20
FIX: Don't blow up when trying to parse invalid or non-ASCII URLs

OsamaSayegh:
Follow-up to https://github.com/discourse/discourse/commit/72f139191e9b3aeebeca31ffa8b75977afe5d816

### DIFF
--- a/lib/file_store/s3_store.rb
+++ b/lib/file_store/s3_store.rb
@@ -79,7 +79,6 @@ module FileStore
     def has_been_uploaded?(url)
       return false if url.blank?
 
-      parsed_url = nil
       begin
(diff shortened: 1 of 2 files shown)


lib/file_store/s3_store.rb
```diff
@@ -79,7 +79,13 @@ def copy_file(url, source, destination)
     def has_been_uploaded?(url)
       return false if url.blank?
 
-      parsed_url = URI.parse(url)
+      parsed_url = nil
```
danielwaterworth, line 82: Nitpick, but you don't actually need this line, because `begin` doesn't open a new scope.

danielwaterworth:
This looks fine to me.

PR #9844, opened 2020-05-20
FEATURE: Add scopes to API keys

romanrizzi:
I still need to add a couple more scopes. Also, running `bundle exec annotate` does not annotate the ApiKeyScope model, I need to figure that out.

### Screenshots:
Screen Shot 2020-05-20 at 17 59 38
Screen Shot 2020-05-20 at 18 00 19

### DIFF
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -5,9 +5,9 @@ class ApiKey < ActiveRecord::Base
 
   SCOPE_MAPPINGS = {
     topics: {
-      write: { action: 'PostsController#create', params: %i[topic_id] },
-      read: { action: 'TopicsController#show', params: %i[topic_id] },
-      feed: { action: 'TopicsController#feed', params: %i[topic_id] }
+      write: { action: 'posts#create', params: %i[topic_id] },
+      read: { action: 'topics#show', params: %i[topic_id], aliases: { topic_id: :id } },
+      feed: { action: 'topics#feed', params: %i[topic_id] }
     }
   }
 
@@ -68,6 +68,12 @@ class ApiKey < ActiveRecord::Base
 
   def self.hash_key(key)
     Digest::SHA256.hexdigest key
+  end
+
+  def request_allowed?(request, route_param)
+    return false if allowed_ips.present? && allowed_ips.none? { |ip| ip.include?(request.ip) }
+
+    api_key_scopes.blank? || api_key_scopes.any? { |s| s.permits?(route_param) }
   end
 end
 
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -3,6 +3,14 @@
 class ApiKey < ActiveRecord::Base
   class KeyAccessError < StandardError; end
 
+  SCOPE_MAPPINGS = {
+    topics: {
+      write: { action: 'PostsController#create', params: %i[topic_id] },
+      read: { action: 'TopicsController#show', params: %i[topic_id] },
+      feed: { action: 'TopicsController#feed', params: %i[topic_id] }
+    }
+  }
+
   belongs_to :user
   belongs_to :created_by, class_name: 'User'
 
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -17,38 +17,6 @@ class ApiKey < ActiveRecord::Base
 
   after_initialize :generate_key
 
-  def self.list_actions
-    actions = []
-
-    TopTopic.periods.each do |p|
-      actions.concat(["list#category_top_#{p}", "list#top_#{p}", "list#top_#{p}_feed"])
-    end
-
-    %i[latest unread new top].each { |f| actions.concat(["list#category_#{f}", "list##{f}"]) }
-
-    actions
-  end
-
-  def self.default_mappings
-    {
-      topics: {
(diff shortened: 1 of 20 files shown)


db/migrate/20200520124359_add_api_key_scopes.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddApiKeyScopes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :api_key_scopes do |t|
```
tgxworld, line 5: Do we need to ensure that certain columns cannot be `null`?

db/migrate/20200520124359_add_api_key_scopes.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.integer :api_key_id
```
tgxworld, line 6: We probably need an index on this column.

db/migrate/20200520124359_add_api_key_scopes.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :resource
```
tgxworld, line 7: What values can `resource` be? If it is from a fixed set, it may be more efficient to store an integer representation rather than a string. I think it'll be the same for `action` as well.

app/serializers/api_key_scope_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class ApiKeyScopeSerializer < ApplicationSerializer
+
+  attributes :resource,
+             :action,
+             :parameters,
+             :allowed_parameters
+
+  def parameters
+    ApiKey.scope_mappings.dig(object.resource.to_sym, object.action.to_sym, :params).to_a
```
tgxworld, line 11: `dig` may return `nil` so calling `nil.to_a` will blow up.

app/models/api_key_scope.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class ApiKeyScope < ActiveRecord::Base
```
tgxworld, line 3: Do we need any validations on this model?

spec/requests/admin/api_controller_spec.rb
```diff
@@ -154,6 +200,16 @@
...
       end
     end
+
+    describe '#scopes' do
+      it 'includes scopes' do
+        get '/admin/api/keys/scopes.json'
+
+        data = response.parsed_body
+
+        expect(data['scopes']).not_to be_empty
```
tgxworld, line 210: I usually prefer to not use generic assertion like `not_to be_empty` because this test will pass even if `data['scopes']` contains random data.

db/migrate/20200520124359_add_api_key_scopes.rb
davidtaylorhq: I suspect this gets tricky because plugins are allowed to contribute new resources/scopes, so we can't pre-allocate consistent integer ids. I think a string is more flexible in this case. This table is never going to a be a performance bottleneck, so I think we can take the hit.

lib/plugin/instance.rb
```diff
@@ -721,6 +721,16 @@ def add_custom_reviewable_filter(filter)
     end
   end
 
+  # Receives a new mapping:
+  # 1. A symbol that represents the name of the resource.
+  # 2. A symbol that represents the name of the action.
+  # 3. An action that contains the controller mapping, optional params, and aliases.
+  def add_api_key_scope(resource, action_name, action)
+    reloadable_patch do
+      ApiKey.new_mapping(resource, action_name, action)
```
davidtaylorhq, line 730: We should make sure that these mappings get removed when plugins are disabled. The easiest way is probably to use a [`filtered_register` in the DiscoursePluginRegistry](https://github.com/discourse/discourse/blob/master/lib/discourse_plugin_registry.rb#L70-L78).

config/locales/client.en.yml
```diff
@@ -3545,6 +3545,14 @@ en:
         delete: Permenantly Delete
         not_shown_again: This key will not be displayed again. Make sure you take a copy before continuing.
         continue: Continue
+        use_scopes: Use Scopes
```
davidtaylorhq, line 3548: Can we flip this round, so that it sounds a bit more scary. Maybe "Global Key (allows all actions)" with a checkbox?

app/models/api_key.rb
```diff
@@ -16,6 +17,24 @@ class KeyAccessError < StandardError; end
 
   after_initialize :generate_key
 
+  def self.scope_mappings
+    @mappings ||= {
+      topics: {
+        write: { action: 'posts#create', params: %i[topic_id] },
```
davidtaylorhq, line 23: Can we make it possible for `action` to be an array? So that later we have the option of making topics:write support multiple actions?

app/serializers/api_key_scope_serializer.rb
romanrizzi: `nil.to_a` will return an empty array.

tgxworld: Ah icic . Today I learned something new

romanrizzi:
@tgxworld @davidtaylorhq Thanks for the review. I have addressed all the comments, feel free to take another look!

PR #9854, opened 2020-05-22
FEATURE: notify admins about old credentials

lis2:
Security and API keys should be renewed periodically.
This additional notification should help admins keep their Discourse safe and secure.

### DIFF
--- a/app/jobs/scheduled/old_keys_reminder.rb
+++ b/app/jobs/scheduled/old_keys_reminder.rb
@@ -54,9 +54,9 @@ module Jobs
     end
 
     def keys_list
-      setting_key_messages = old_site_settings_keys.map { |key| "#{key.name} - #{key.updated_at}" }
-      api_key_messages = old_api_keys.map { |key| "#{[key.description, key.user&.username, key.created_at].compact.join(" - ")}" }
-      [setting_key_messages | api_key_messages].join("\n")
+      messages = old_site_settings_keys.map { |key| "#{key.name} - #{key.updated_at}" }
+      old_api_keys.each_with_object(messages) { |key, array| array << "#{[key.description, key.user&.username, key.created_at].compact.join(" - ")}" }
+      messages.join("\n")
     end
   end
 end
--- a/app/jobs/scheduled/old_keys_reminder.rb
+++ b/app/jobs/scheduled/old_keys_reminder.rb
@@ -24,33 +24,25 @@ module Jobs
     def old_site_settings_keys
       @old_site_settings_keys ||= SiteSetting.secret_settings.each_with_object([]) do |secret_name, old_keys|
         site_setting = SiteSetting.find_by(name: secret_name)
-        next unless site_setting
-        next if site_setting.value.blank?
+        next if site_setting&.value.blank?
         next if site_setting.updated_at + calculate_period > Time.zone.now
         old_keys << site_setting
-      end
+      end.sort_by { |key| key.updated_at }
     end
 
     def old_api_keys
-      @old_api_keys ||= ApiKey.all.each_with_object([]) do |api_key, old_keys|
+      @old_api_keys ||= ApiKey.all.order(created_at: :asc).each_with_object([]) do |api_key, old_keys|
         next if api_key.created_at + calculate_period > Time.zone.now
         old_keys << api_key
       end
     end
 
     def calculate_period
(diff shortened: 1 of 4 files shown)


app/jobs/scheduled/old_keys_reminder.rb
```diff
@@ -0,0 +1,70 @@
...
+        1.year
+      when '2 years'
+        2.years
+      when '3 years'
+        3.years
+      end
+    end
+
+    def admins
+      User.real.where(admin: true)
```
eviltrout, line 53:
```suggestion
      User.real.admins
```

app/jobs/scheduled/old_keys_reminder.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def old_api_keys
+      @old_api_keys ||= ApiKey.all.each_with_object([]) do |api_key, old_keys|
+        next if api_key.created_at + calculate_period > Time.zone.now
+        old_keys << api_key
+      end
+    end
+
+    def calculate_period
+      case SiteSetting.notify_about_secrets_older_than
```
eviltrout, line 42: Could probably do `SiteSetting.notify_about_secrets_older_than.to_i.years`

app/jobs/scheduled/old_keys_reminder.rb
```diff
@@ -0,0 +1,70 @@
...
+        )
+      end
+    end
+
+    private
+
+    def old_site_settings_keys
+      @old_site_settings_keys ||= SiteSetting.secret_settings.each_with_object([]) do |secret_name, old_keys|
+        site_setting = SiteSetting.find_by(name: secret_name)
+        next unless site_setting
```
eviltrout, line 27: Could save a line here with: 
`next if site_setting&.value.blank?`

spec/jobs/old_keys_reminder_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+describe Jobs::OldKeysReminder do
+  let!(:google_secret) { SiteSetting.create!(name: 'google_oauth2_client_secret', value: '123', data_type: 1) }
+  let!(:instagram_secret) { SiteSetting.create!(name: 'instagram_consumer_secret', value: '123', data_type: 1) }
+  let!(:api_key) { Fabricate(:api_key, description: 'api key description') }
+  let!(:admin) { Fabricate(:admin) }
+
+  it "detects old keys" do
+    SiteSetting.notify_about_secrets_older_than = "2 years"
+    freeze_time 2.years.from_now
+    expect(described_class.new.send(:old_site_settings_keys).sort).to eq([google_secret, instagram_secret].sort)
```
eviltrout, line 14: I don't think we should be testing private methods. A test should stick to the public API surface because this prevents us from refactoring in the future.

In this case I would recommend moving the `old_xxx` methods to `SiteSetting` as public and testing using those.

For the `title`, `body` etc you can look for the post that was created after the job finishes.

lis2:
@eviltrout thank you for the feedback. I fixed it. I think that old keys setting is very specific to this problem, so I didn't want to move that to SiteSettings model. Instead, I extended the test for "execute" method so there is no more need to test private methods.

app/jobs/scheduled/old_keys_reminder.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def body
+      I18n.t('old_keys_reminder.body', keys: keys_list)
+    end
+
+    def keys_list
+      setting_key_messages = old_site_settings_keys.map { |key| "#{key.name} - #{key.updated_at}" }
+      api_key_messages = old_api_keys.map { |key| "#{[key.description, key.user&.username, key.created_at].compact.join(" - ")}" }
+      [setting_key_messages | api_key_messages].join("\n")
```
tgxworld, line 59: Minor but instead of two intermediate arrays, because we can just push into an existing array?

lis2: Thank you! nice catch, fixed and I will merge that stuff tomorrow in the morning :)

tgxworld: Btw I left a small comment in dev about the design of this feature in https://dev.discourse.org/t/automatic-reminder-to-reset-old-machine-generated-secrets/16311/9

PR #9855, opened 2020-05-22
DEV: Deprecate ordered_posts in favor of posts

udan11:
`posts` becomes `ordered_posts` and `ordered_posts` will be removed in the future.


SamSaffron:
This is great, from my perspective you can merge at will.

SamSaffron:
Should we close this pending a PR to Rails to add the magical support we need?

udan11:
I agree.

PR #9861, opened 2020-05-22
Require rails_helper in spec

legalosLOTR:
This spec doesn't require rails_helper and so it fails when run in isolation.

## Testing plan
- Run spec without this change and see it fail
- Run spec after this change and see it pass


eviltrout:
Looks good, thanks!

PR #9863, opened 2020-05-22
FEATURE: Permalinks for tags

gschlager:
This adds the ability to create permalinks for tags. Admins can enter the tag name, but the system will save the tag ID, so renaming of tags works.


SamSaffron:
looks good, but I think we need 1 integration spec here for permalinks that makes sure this works.

PR #9867, opened 2020-05-25
FIX: do not allow tag with name 'none'

techAPJ:
https://meta.discourse.org/t/none-tag-is-uneditable/152003

### DIFF
--- a/db/migrate/20200525072638_remove_none_tags.rb
+++ b/db/migrate/20200525072638_remove_none_tags.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class RemoveNoneTags < ActiveRecord::Migration[6.0]
+  def up
+    none_tag_id = DB.query_single("SELECT id FROM tags WHERE lower(name) = 'none'").first
+    if none_tag_id.present?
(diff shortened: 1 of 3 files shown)


db/migrate/20200525072638_remove_none_tags.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class RemoveNoneTags < ActiveRecord::Migration[6.0]
+  def up
+    none_tag = Tag.where("lower(name) = ?", "none").first
```
nlalonde, line 5: We should be using sql here instead of using the AR models. It's annoying because there are so many associations to delete... This case is so simple it might be ok.

nlalonde: I think it's better if we iterate over the associations `[:tag_users, :topic_tags, ... ]` and delete the rows with sql. `delete from #{table_name} where tag_id = :none_tag_id`

techAPJ: Update the migration to use SQL.

PR #9868, opened 2020-05-25
FIX: add X-Robots-Tag header for check_xhr-covered GET actions, too

artemv:
see https://meta.discourse.org/t/missing-x-robots-tag/152593/3 for context

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -39,12 +39,12 @@ class ApplicationController < ActionController::Base
   before_action :redirect_to_login_if_required
   before_action :block_if_requires_login
   before_action :preload_json
+  before_action :add_noindex_header, if: -> { is_feed_request? || !SiteSetting.allow_index_in_robots_txt }
   before_action :check_xhr
   after_action  :add_readonly_header
   after_action  :perform_refresh_session
   after_action  :dont_cache_page
   after_action  :conditionally_allow_site_embedding
-  after_action  :add_noindex_header, if: -> { is_feed_request? || !SiteSetting.allow_index_in_robots_txt }
 
   layout :set_layout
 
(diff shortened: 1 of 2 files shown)


eviltrout:
Can you add a test for this? We rarely accept fixes without tests.

artemv:
> Can you add a test for this? We rarely accept fixes without tests.

It would be easy if there was an existing non-json GET action in Discourse that 
* allowed non-logged visitors
* didn't skip `check_xhr` filter

Do you know such action? Otherwise I can add such action in spec code similarly to what plugins do but this sounds hacky. Or fine?

eviltrout:
Why would a GET for html not skip xhr? I don't understand now that I've read in more detail.

artemv:
> Why would a GET for html not skip xhr? I don't understand now that I've read in more detail.

Initially my plugin' action didn't skip check_xhr, and it was happily rendering empty layout then loaded page contents with XHR - the same way e.g. `/admin` does (its HTML version doesn't skip `check_xhr`), but in my case for not logged in users. This doesn't look wrong to me by itself. But the `add_noindex_header` not working in this case does look wrong.

eviltrout:
@artemv I am still super confused here. Where does `/admin` render HTML without skipping xhr?

artemv:
> @artemv I am still super confused here. Where does `/admin` render HTML without skipping xhr?

At standard Discourse. Here, I've added logging to `check_xhr` and requested `localhost:3000/admin`: 
bundle exec rails server 2020-05-26 17-43-46

eviltrout:
Aha I understand now, sorry what confused me is `check_xhr` does not ensure that XHR is present, it bootstraps the app for any GET if nothing special is supposed to happen.

It looks like the `/groups` path will respond to anonymous requests and doesn't skip xhr, so your test can use that path.

artemv:
@eviltrout I'm glad you figured it out - I was like, ugh, let's annotate `check_xhr` filter to give that guy more context about it. Wait, he's the author? Eh. :) It was long ago though, understood!

artemv:
@eviltrout just in case - I've added a test, please review

PR #9869, opened 2020-05-25
DEV: allows to decorate username selector

jjaffeux:
Usage:

```
api.addUsernameSelectorDecorator(username => {
  return iconHTML("calendar-alt");
});
```


jjaffeux:
@eviltrout what do you think of this approach? It will allow me to cleanly add an icon like this for example:

Screenshot 2020-05-25 at 15 22 54

PR #9871, opened 2020-05-25
DEV: Remove support for FontAwesome 4.7 icon names

pmusaraj:


SamSaffron:
I support this :+1:  we should merge as soon as possible.

PR #9875, opened 2020-05-25
FIX: The parent of the thumbnail image needs to have the lazyYT-container class set.

romanrizzi:


eviltrout:
@davidtaylorhq how's this look?

ZogStriP:
Closing in favor of #9870.

PR #9878, opened 2020-05-26
DEV: attempts to reuse pretty text object in composer

jjaffeux:


jjaffeux:
@reviewers note that this commit is also attempting to exit very early from `loadMarkdownIt` even if the code after in loadScript is kinda optimised/cached, given this is happening on every keystroke I think that exiting as soon as possible is a good idea here.

jjaffeux:
> I would hope that `loadScript` was smart enough to do this check itself, but if you've benchmarked it as faster let's do it!

loadScript does various things for this already, but he can't know which variable on window it will define, so doing this basically we avoid:

- check to Discourse.MarkdownItURL
- check on URL
- check for url in loaded url cache

So this is very much micro optimisation and most likely not even 1ms but it's on every key stroke and very easy to do.

The real big thing of the PR is reusing the markdown.it object.

SamSaffron:
One concern I have is do different composers have different contexts under certain conditions? 

Reply to topic 1, then reply to topic 2, reply to PM ... does anything change about the context? 

Should we keep the context associated with the composer and poison it when we clear the composer / load a new one ?

davidtaylorhq:
Closing in favour of #9987

PR #9888, opened 2020-05-27
FIX: Do not destroy $.fileupload element

nbianca:
conditional-loading-section component rerendered the <input> element
and lost the necessary event handlers for jQuery-File-Upload.

### DIFF
--- a/app/assets/javascripts/discourse/app/mixins/upload.js
+++ b/app/assets/javascripts/discourse/app/mixins/upload.js
@@ -39,8 +39,10 @@ export default Mixin.create({
 
   _initialize: on("didInsertElement", function() {
     const $upload = $(this.element);
-    const reset = () =>
+    const reset = () => {
       this.setProperties({ uploading: false, uploadProgress: 0 });
+      document.getElementsByClassName("hidden-upload-field")[0].value = "";
+    };
     const maxFiles = this.getWithDefault(
       "maxFiles",
       this.siteSettings.simultaneous_uploads
--- a/app/assets/javascripts/discourse/app/templates/components/emoji-uploader.hbs
+++ b/app/assets/javascripts/discourse/app/templates/components/emoji-uploader.hbs
@@ -1,48 +1,51 @@
-{{#conditional-loading-section isLoading=uploading}}
-  <div class="emoji-uploader">
-    <div class="control">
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/discourse/app/mixins/upload.js
```diff
@@ -39,8 +39,10 @@ export default Mixin.create({
 
   _initialize: on("didInsertElement", function() {
     const $upload = $(this.element);
-    const reset = () =>
+    const reset = () => {
       this.setProperties({ uploading: false, uploadProgress: 0 });
+      $upload.find(".hidden-upload-field").val("");
```
jjaffeux, line 44: do we need jquery for this ?

app/assets/javascripts/discourse/app/templates/components/emoji-uploader.hbs
```diff
@@ -1,48 +1,51 @@
...
-    <div class="control">
-      <div class="input">
-        <label class="btn btn-default btn-primary {{if addDisabled "disabled"}}">
+  </div>
+  <div class="control">
+    <div class="input">
+      <label class="btn btn-default btn-primary {{if addDisabled "disabled"}}">
+        {{#if uploading}}
+          {{d-icon "spinner" class="loading-icon"}}
+          {{i18n "admin.emoji.uploading"}}
```
jjaffeux, line 38: should have a span around

app/assets/javascripts/discourse/app/templates/components/emoji-uploader.hbs
```diff
@@ -1,48 +1,51 @@
...
+        {{else}}
           {{d-icon "plus"}}
           {{i18n "admin.emoji.add"}}
```
jjaffeux, line 41: should have a span around too (I know it's not from the PR :p )

app/assets/javascripts/discourse/app/mixins/upload.js
nbianca: We already use a lot of jQuery around here and rely on jQuery-File-Upload. It will not be going away anytime soon, but I agree with your comment so I will make the changes.

SamSaffron:
This looks like a nice fix, merging it in.

PR #9890, opened 2020-05-27
FIX: Pull hotlinked images even when edited by system users

davidtaylorhq:
Previously the pull hotlinked images job was skipped after system edits. This ensured that we never had an infinite loop of system-edit/pull-hotlinked/system-edit/pull-hotlinked etc.

A side effect was that edits made by system for any other reason (e.g. API, removing full quotes) would prevent pulling hotlinked images. This commit removes the system edit check, and replaces it with another method to avoid an infinite job scheduling loop.

Context: https://meta.discourse.org/t/thumbnail-generation-markdown-rendering-issue/152701/10

### DIFF
--- a/spec/components/cooked_post_processor_spec.rb
+++ b/spec/components/cooked_post_processor_spec.rb
@@ -1486,8 +1486,8 @@ describe CookedPostProcessor do
       end
 
       it "does not run when requested to skip" do
-        Jobs.expects(:enqueue_in).never
         CookedPostProcessor.new(post, skip_pull_hotlinked_images: true).pull_hotlinked_images
+        expect(Jobs::PullHotlinkedImages.jobs.size).to eq(0)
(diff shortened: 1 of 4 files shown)


davidtaylorhq:
Requires https://github.com/discourse/discourse/pull/9897

spec/components/cooked_post_processor_spec.rb
```diff
@@ -1485,16 +1485,14 @@ def s3_setup
         expect(SiteSetting.download_remote_images_to_local).to eq(false)
       end
 
-      context "and there is enough disk space" do
+      it "does not run when requested to skip" do
+        Jobs.expects(:enqueue_in).never
```
tgxworld, line 1489: Is it possible to check that the jobs queue is empty instead?

PR #9892, opened 2020-05-27
FIX: Show tag count for existing tags

nbianca:
When the tag matched an existent one, the count was not displayed.


ZogStriP:
Should there be a test for that?

PR #9908, opened 2020-05-28
FIX: Invalidate database theme cache when hostname changes

davidtaylorhq:
Hostname can vary per-site on a multisite cluster, so this change requires converting the compiler_version from a constant into a class method which is evaluated at runtime. The value is stored in the theme DistributedCache, so performance impact should be negligible.

### DIFF
--- a/spec/models/theme_spec.rb
+++ b/spec/models/theme_spec.rb
@@ -716,7 +716,7 @@ HTML
       first_common_value = Theme.lookup_field(child.id, :desktop, "header")
       first_extra_js_value = Theme.lookup_field(child.id, :extra_js, nil)
 
-      stub_const(ThemeField, :COMPILER_VERSION, "SOME_NEW_HASH") do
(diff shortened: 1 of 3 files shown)


spec/models/theme_spec.rb
```diff
@@ -730,5 +730,18 @@ def stub_const(target, const, value)
...
     end
+
+    it 'recompiles when the hostname changes' do
+      theme.set_field(target: :settings, name: :yaml, value: "name: bob")
+      theme_field = theme.set_field(target: :common, name: :after_header, value: '<script>console.log("hello world");</script>')
+      theme.save!
+
+      expect(Theme.lookup_field(theme.id, :common, :after_header)).to include("_ws=#{Discourse.current_hostname}")
+
+      Discourse.stubs(:current_hostname).returns("someotherhostname.com")
```
tgxworld, line 741: We can change this to `SiteSetting.force_hostname = "someotherhostname.com"`

PR #9910, opened 2020-05-28
DEV: abstract new functions out of topic-tracking-state

markvanlan:


SamSaffron:
I am actually working on this right now... hold off on work there. 

I am sorting out new/unread counts for tags with sidebar theme... it is a very very tricky change.

PR #9924, opened 2020-05-29
FIX: Staff users can bypass tag validation rule

udan11:
The rule was already bypassed on the server side.


SamSaffron:
Seems like a safe change to me!

PR #9926, opened 2020-05-29
DEV: flexify post actions

jjaffeux:


jjaffeux:
How do we feel about this ?

jjaffeux:
> Seeing an alignment issue with the replies button, so I think `.post-controls` needs to be flex too?
> 
> Screen Shot 2020-05-29 at 4 50 32 PM

Yes I saw this after indeed, I would have to fix this too, but it's easy. Will update PR.

jjaffeux:
This is how it should look:

Screenshot 2020-05-29 at 23 00 38

jjaffeux:
will wait after beta, this is a little bit tricky

SamSaffron:
Merging as this was approved by @awesomerobot

PR #9927, opened 2020-05-29
DEV: Plugin api for adding category admin dropdown items

markvanlan:
It is suprising difficult to add an item to this dropdown right now.

Screenshot from 2020-05-29 10-33-50

The actions for the items are defined in `app/assets/javascripts/discourse/app/components/d-navigation.js`, but the items themselves are defined [here](https://github.com/discourse/discourse/blob/94cb5ab17274a26f7bd972eb06a2d90b73f7a2f1/app/assets/javascripts/select-kit/addon/components/categories-admin-dropdown.js#L20).

This plugin api gives a single function to add an item to the dropdown, as well as a function to call when the item is clicked.

___
**Possible change?**

I could see including the callback function in `item` like
```
{
  id: X
  label: X
  onClick: function() {...}
}
```

And cutting out the onClick out of the `item` to pass it into 
```
modifySelectKit("categories-admin-dropdown").appendContent(function() {
    return item;
});
```

### DIFF
--- a/app/assets/javascripts/discourse/app/components/d-navigation.js
+++ b/app/assets/javascripts/discourse/app/components/d-navigation.js
@@ -4,6 +4,14 @@ import { inject as service } from "@ember/service";
 import Component from "@ember/component";
 import FilterModeMixin from "discourse/mixins/filter-mode";
 
+let _pluginCategoryAdminDropdownActions = {};
+
+export function addCategoryAdminDropdownAction(actionId, handler) {
+  _pluginCategoryAdminDropdownActions[actionId] =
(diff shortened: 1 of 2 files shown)


jjaffeux:
> It is suprising difficult to add an item to this dropdown right now.

It's actually very easy :p I don't do it on every dropdown because there are many different cases, and for example historically categories-admin-dropdown has been defining actions a little bit differently than other cases. But with only a little tweak to the component, we can leverage select-kit plugin API without having to write a dedicated API.

I propose the following PR instead:

https://github.com/discourse/discourse/pull/9931

markvanlan:
@jjaffeux Yeah that is fine! What I do

PR #9932, opened 2020-05-29
UX: Position topic admin menu next to wrench

pmusaraj:
Moves the admin menu below the button when next to the timeline: 

image

and to the left (or right on RTL) of the button when next to the footer topic buttons: 

image


jjaffeux:
I think it would be worth 3-5 px more vertical spacing so it's not "stuck" to the wrench.

PR #9935, opened 2020-05-29
DEV: Add option to only show icons to header widget

markvanlan:
Note I also added the `.category` class to `body` when viewing a category. This required a change to the `removeClass` function to make sure it gets removed `afterRender`. 

This is for further styling for a plugin


jjaffeux:
It doesn't need to be a function, can just be a const, given you will use it in both cases

PR #9940, opened 2020-05-30
FIX: catch NoMethodError during precompile_css too

tmm1:
Ran into this on heroku:

    NoMethodError: undefined method `enabled?' for #<Theme:0x0000558f069e7718>
    vendor/bundle/ruby/2.6.0/gems/activemodel-6.0.1/lib/active_model/attribute_methods.rb:431:in `method_missing'
    app/models/theme.rb:155:in `block (2 levels) in transform_ids'
    vendor/bundle/ruby/2.6.0/gems/activerecord-6.0.1/lib/active_record/relation/delegation.rb:85:in `each'
    vendor/bundle/ruby/2.6.0/gems/activerecord-6.0.1/lib/active_record/relation/delegation.rb:85:in `each'
    vendor/bundle/ruby/2.6.0/gems/activerecord-6.0.1/lib/active_record/relation/query_methods.rb:260:in `select'
    vendor/bundle/ruby/2.6.0/gems/activerecord-6.0.1/lib/active_record/relation/query_methods.rb:260:in `select'
    app/models/theme.rb:155:in `block in transform_ids'
    app/models/theme.rb:105:in `get_set_cache'
    app/models/theme.rb:140:in `transform_ids'
    app/models/theme.rb:321:in `list_baked_fields'
    app/models/theme.rb:317:in `resolve_baked_field'
    lib/stylesheet/manager.rb:289:in `theme_digest'

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -57,7 +57,7 @@ task 'assets:precompile:css' => 'environment' do
         STDERR.puts "Compiling css for #{db} #{Time.zone.now}"
         begin
           Stylesheet::Manager.precompile_css
-        rescue PG::UndefinedColumn, ActiveModel::MissingAttributeError => e
+        rescue PG::UndefinedColumn, ActiveModel::MissingAttributeError, NoMethodError => e
           STDERR.puts "#{e.class} #{e.message}: #{e.backtrace.join("\n")}"
           STDERR.puts "Skipping precompilation of CSS cause schema is old, you are precompiling prior to running migrations."
         end


eviltrout:
This does not seem like the right fix to me. It's a catch all that could sweep up many other things.

Do you know why the method doesn't exist? Is a migration missing or something?

tmm1:
> Do you know why the method doesn't exist? Is a migration missing or something?

The migrations have not been run, as mentioned in the line below.

`you are precompiling prior to running migrations`

SamSaffron:
Yeah it is certainly rare to be eating exceptions but in this case I am fine with it cause we both print out clearly that we skipped them and handle this automatically anyway in runtime if for some reason the precompile fails. This is also super scoped to the rake task. 

An alternative here would be to confirm theme schema matches exactly what we want, but wiring that up would be pretty complex.

PR #9942, opened 2020-05-30
DEV: removes dead code of dropdown-menu

jjaffeux:


jjaffeux:
I can't find where this would be still used, can you ?

PR #9945, opened 2020-05-30
UX: Adjustments to topic admin menu on mobile

pmusaraj:
Admin wrench button is now placed next to "jump to menu" when the topic progress is expanded. 

Before >> After

image image

When clicked, previously we were sometimes displaying a cloaking element, this PR removes the cloaking entirely. 

Before >> After

image image


jjaffeux:
@pmusaraj we should merge this before beta

PR #9946, opened 2020-05-30
FIX: Only render group card if user title is from group

riking:
This was failing when a user with a primary_group chose to display a title coming from a badge.

### DIFF
--- a/app/serializers/post_serializer.rb
+++ b/app/serializers/post_serializer.rb
@@ -47,6 +47,7 @@ class PostSerializer < BasicPostSerializer
              :link_counts,
              :read,
              :user_title,
+             :title_is_group,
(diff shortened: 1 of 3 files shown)


app/serializers/post_serializer.rb
```diff
@@ -212,6 +213,14 @@ def user_title
     object&.user&.title
   end
 
+  def title_is_group
+    object&.user&.title == object.user&.primary_group&.title
```
jjaffeux, line 217: why do we object&.user on one side an object.user on the other ?

riking: Whoops, copy-paste error. Probably harmless, so I'm going to submit as is.

PR #9947, opened 2020-05-31
FIX: Correct the misalignment of popover arrows

CvX:
* Fixes position of the top arrow, and border width of both
* Merged `top` and `transform` properties
* .5px values are required to make arrow border appear the same with as overall popover border width

Before/after

Screen Shot 2020-05-25 at 14 59 50 Screen Shot 2020-05-25 at 15 01 27

### DIFF
--- a/app/assets/stylesheets/common/base/d-popover.scss
+++ b/app/assets/stylesheets/common/base/d-popover.scss
@@ -80,15 +80,14 @@ $d-popover-border: $primary-medium;
 
   .d-popover-top-arrow {
     border-color: transparent transparent $d-popover-border;
-    top: 8px;
-    transform: translate(0, -15px);
+    top: -8px;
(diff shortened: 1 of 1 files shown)


coding-horror:
👏👁

jjaffeux:
I'm curious if it works in non retina display ?

CvX:
@jjaffeux I did have a chance to try it out on a Windows machine with a non-high DPI display and there were no issues. And I just re-tested it with Firefox on macOS using the "Open in Low Resolution" option and it looked more or less ok (bearing in mind that it the whole app is then a blurry mess, and that using custom zoom levels causes the arrow to break slightly, but that was also the case on Firefox before this change)

jjaffeux:
great 👌 then please merge it before beta

PR #9950, opened 2020-06-01
FIX: Improvements and fixes to the image downsizing script

CvX:
Fixed bugs, added specs, extracted the upload downsizing code to a class, added support for non-S3 setups, changed it so that images aren't downloaded twice.


CvX:
Btw. this code has been tested on production and successfully resized ~180k uploads.

PR #9953, opened 2020-06-01
DEV: Drop unused image_url column from posts and topics

davidtaylorhq:
This has been superseded by image_upload_id. The image_url value in API responses is now generated dynamically from the upload record.


SamSaffron:
Looks good and safe to me, merging :+1:

PR #9955, opened 2020-06-01
DEV: Correct historical schema discrepancies

davidtaylorhq:


SamSaffron:
Overall this is great and should be very safe to merge. 

One change though I would like to see is to query the information schema in batches, its a small change, but means this migration in the happy state will only run 3 selects vs 30. 

Once that is done merge away.

```
select * from (
   select 1 id, 2 x_id
   union all 
   select 2, 4
) as X 
join table...
```

Or a CTE or something.

PR #9957, opened 2020-06-01
FIX: Support exporting reports which reference topics

davidtaylorhq:


spec/jobs/export_csv_file_spec.rb
```diff
@@ -152,6 +152,21 @@
...
     end
 
+    it 'works with topic reports' do
+      freeze_time DateTime.parse('2010-01-01 6:00')
+
+      exporter.extra['name'] = 'top_referred_topics'
+      post1 = Fabricate(:post)
+      post2 = Fabricate(:post)
+      IncomingLink.add(host: "a.com", referer: "http://twitter.com", post_id: post1.id, ip_address: '1.1.1.1')
+      # IncomingLink.add(host: "a.com", referer: "http://twitter.com", post_id: post2.id, ip_address: '1.1.1.2')
```
eviltrout, line 162: Did you mean to check this in commented out?

davidtaylorhq: Oops no, thanks

PR #9963, opened 2020-06-02
DEV: Use consistent interpolation key format in translations

gschlager:
From now on client strings can easily be reused on the server and you don’t have to think about choosing the right format anymore.


app/assets/javascripts/discourse/app/components/flag-action-type.js
```diff
@@ -20,7 +20,7 @@ export default Component.extend({
   )
   formattedName(name, nameKey, isCustomFlag, username) {
     if (isCustomFlag) {
-      return name.replace("{{username}}", username);
```
eviltrout, line 23: I wonder if we should temporarily support both, as this will break all non-english languages right away?

gschlager: Oh, good catch. I didn't think about that. I'll do that and remove it with the next update of the translations.

PR #9969, opened 2020-06-02
PERF: rate limit search, and add anon cache for search results

featheredtoast:


spec/requests/search_controller_spec.rb
```diff
@@ -217,6 +267,57 @@
...
+      end
+
+      context "and a logged in user" do
+        before { sign_in(Fabricate(:user)) }
+
+        it 'rate limits searches' do
+          RateLimiter.enable
+          RateLimiter.clear_all!
+
+          30.times do
```
eviltrout, line 302: Just curious here - how slow are these tests? `30.times` seems like a lot.

featheredtoast: It was. I added a hidden site setting for this, and dialed the rate limits down in them so tests are much faster now

SamSaffron:
Looks good to me, merge at will.

spec/requests/search_controller_spec.rb
```diff
@@ -179,6 +179,61 @@
...
+        get "/search/query.json", params: {
+              term: 'wookie'
+            }
+        expect(response.status).to eq(200)
+        json = response.parsed_body
+        expect(json["grouped_search_result"]["error"]).to eq(I18n.t("rate_limiter.slow_down"))
+      end
+
+      context "and a logged in user" do
+        before { sign_in(Fabricate(:user)) }
```
tgxworld, line 212:
```suggestion
        before { sign_in(user) }
```

spec/requests/search_controller_spec.rb
```diff
@@ -217,6 +272,63 @@
...
+              q: 'bantha'
+            }
+        expect(response.status).to eq(200)
+        json = response.parsed_body
+        expect(json["grouped_search_result"]["error"]).to eq(I18n.t("rate_limiter.slow_down"))
+
+      end
+
+      context "and a logged in user" do
+        before { sign_in(Fabricate(:user)) }
```
tgxworld, line 307:
```suggestion
        before { sign_in(user) }
```

app/controllers/search_controller.rb
```diff
@@ -140,6 +154,19 @@ def site_overloaded?
...
   end
 
+  def rate_limit_search
+    begin
+      if current_user.present?
+        RateLimiter.new(current_user, "search-min", SiteSetting.rate_limit_search_user, 1.minute).performed!
+      else
+        RateLimiter.new(nil, "search-min-#{request.remote_ip}", SiteSetting.rate_limit_search_anon, 1.minute).performed!
+      end
+    rescue RateLimiter::LimitExceeded => e
```
tgxworld, line 164: Hmm this is a little odd, we should just throw the error and let the `rescue_from` handler in `ApplicationController` handle it.

app/controllers/search_controller.rb
```diff
@@ -48,7 +52,10 @@ def show
     search_args[:ip_address] = request.remote_ip
     search_args[:user_id] = current_user.id if current_user.present?
 
-    if site_overloaded?
+    if rate_limit_errors
+      result = Search::GroupedSearchResults.new(search_args[:type_filter], @search_term, context, false, 0)
```
tgxworld, line 56: Ah is the intention for us to return the cached results if rate limit has been exceeded?

tgxworld: Oops sorry just read the thread on dev. This looks good to me.

PR #9970, opened 2020-06-03
UX: Add simple-list setting type

pmusaraj:
Adds a new site setting list type called simple list, to be used only with settings that **do not** have a list of choices available. 

screencast 2020-06-02 22-19-13

This PR only enables this new setting type for the `content security policy script src` setting, as a first step., Support for other site settings (and theme settings) will be added in subsequent PRs.

### DIFF
--- a/app/assets/javascripts/admin/components/simple-list.js
+++ b/app/assets/javascripts/admin/components/simple-list.js
@@ -1,0 +1,80 @@
+import { empty } from "@ember/object/computed";
+import Component from "@ember/component";
+import { on } from "discourse-common/utils/decorators";
+
+export default Component.extend({
+  classNameBindings: [":simple-list", ":value-list"],
+  inputEmpty: empty("newValue"),
+  inputDelimiter: null,
+  inputType: null,
+  newValue: "",
+  collection: null,
+  values: null,
+
+  @on("didReceiveAttrs")
+  _setupCollection() {
+    const values = this.values;
+    if (this.inputType === "array") {
+      this.set("collection", values || []);
+      return;
+    }
+
+    this.set(
+      "collection",
+      this._splitValues(values, this.inputDelimiter || "\n")
+    );
+  },
+
+  keyDown(event) {
+    if (event.keyCode === 13) this.send("addValue", this.newValue);
+  },
+
+  actions: {
+    changeValue(index, newValue) {
+      this._replaceValue(index, newValue);
+    },
+
+    addValue(newValue) {
+      if (this.inputInvalid) return;
+
+      this.set("newValue", null);
+      this._addValue(newValue);
+    },
+
+    removeValue(value) {
+      this._removeValue(value);
+    }
+  },
+
+  _addValue(value) {
+    this.collection.addObject(value);
+    this._saveValues();
+  },
+
+  _removeValue(value) {
+    this.collection.removeObject(value);
+    this._saveValues();
+  },
+
+  _replaceValue(index, newValue) {
+    this.collection.replace(index, 1, [newValue]);
+    this._saveValues();
+  },
+
+  _saveValues() {
+    if (this.inputType === "array") {
+      this.set("values", this.collection);
+      return;
+    }
+
+    this.set("values", this.collection.join(this.inputDelimiter || "\n"));
+  },
+
+  _splitValues(values, delimiter) {
+    if (values && values.length) {
+      return values.split(delimiter).filter(x => x);
+    } else {
+      return [];
+    }
+  }
+});
--- a/app/assets/javascripts/admin/templates/components/simple-list.hbs
+++ b/app/assets/javascripts/admin/templates/components/simple-list.hbs
@@ -1,0 +1,40 @@
+{{#if collection}}
+  <div class="values">
+    {{#each collection as |value index|}}
+      <div data-index={{index}} class="value">
+        {{d-button
+              action=(action "removeValue")
+              actionParam=value
+              icon="times"
+              class="remove-value-btn btn-small"
+            }}
+
+        {{input
+              title=value
+              value=value
+              class="value-input"
+              focus-out=(action "changeValue" index)
+            }}
+      </div>
+    {{/each}}
+  </div>
+{{/if}}
+
+<div class="simple-list-input">
+  {{input
+    type="text"
+    value=newValue
+    placeholderKey="admin.site_settings.simple_list.add_item"
+    class="add-value-input"
+    autocomplete="discourse"
+    autocorrect="off"
+    autocapitalize="off"}}
+
+  {{d-button
+      action=(action "addValue")
+      actionParam=newValue
+      disabled=inputEmpty
+      icon="plus"
+      class="add-value-btn btn-small"
+    }}
+</div>
--- a/app/assets/stylesheets/common/admin/admin_base.scss
+++ b/app/assets/stylesheets/common/admin/admin_base.scss
@@ -931,6 +931,20 @@ table#user-badges {
   }
 }
 
+.simple-list-input {
+  display: flex;
+
+  input {
+    margin: 0;
+    box-sizing: border-box;
+    flex: 1 0 0px;
+  }
+
+  button {
+    margin-left: 0.25em;
+  }
+}
+
 // Mobile view text-inputs need some padding
 .mobile-view .admin-contents {
   input[type="text"] {
--- a/app/assets/javascripts/admin/mixins/setting-component.js
+++ b/app/assets/javascripts/admin/mixins/setting-component.js
@@ -25,7 +25,8 @@ const CUSTOM_TYPES = [
   "upload",
   "group_list",
   "tag_list",
-  "color"
+  "color",
+  "simple_list"
 ];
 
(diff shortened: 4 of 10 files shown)


jjaffeux:
Looks good, thx! will do few comment.

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,80 @@
...
+  inputEmpty: empty("newValue"),
+  inputDelimiter: null,
+  inputType: null,
+  newValue: "",
+  collection: null,
+  values: null,
+
+  @on("didReceiveAttrs")
+  _setupCollection() {
+    const values = this.values;
```
jjaffeux, line 16: don't this this const is needed, I would just directly use this.values

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    this.set(
+      "collection",
+      this._splitValues(values, this.inputDelimiter || "\n")
+    );
+  },
+
+  keyDown(event) {
+    if (event.keyCode === 13) this.send("addValue", this.newValue);
```
jjaffeux, line 29: once you change to @action you can do this.addValue(this.newValue)

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,80 @@
...
+  },
+
+  actions: {
```
jjaffeux, line 32: please use @action

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,80 @@
...
+      this.set("values", this.collection);
+      return;
+    }
+
+    this.set("values", this.collection.join(this.inputDelimiter || "\n"));
+  },
+
+  _splitValues(values, delimiter) {
+    if (values && values.length) {
+      return values.split(delimiter).filter(x => x);
```
jjaffeux, line 75:
```suggestion
      return values.split(delimiter).filter(Boolean);
```

app/assets/javascripts/admin/templates/components/simple-list.hbs
```diff
@@ -0,0 +1,40 @@
+{{#if collection}}
+  <div class="values">
+    {{#each collection as |value index|}}
+      <div data-index={{index}} class="value">
+        {{d-button
```
jjaffeux, line 5: indentation of both is wrong here

app/assets/javascripts/admin/templates/components/simple-list.hbs
```diff
@@ -0,0 +1,40 @@
...
+  {{input
+    type="text"
+    value=newValue
+    placeholderKey="admin.site_settings.simple_list.add_item"
+    class="add-value-input"
+    autocomplete="discourse"
+    autocorrect="off"
+    autocapitalize="off"}}
+
+  {{d-button
```
jjaffeux, line 33: indentation is wrong

app/assets/stylesheets/common/admin/admin_base.scss
```diff
@@ -931,6 +931,20 @@ table#user-badges {
   }
 }
 
+.simple-list-input {
+  display: flex;
+
+  input {
+    margin: 0;
+    box-sizing: border-box;
+    flex: 1 0 0px;
```
jjaffeux, line 940: why 0px and not 0 ?

app/assets/stylesheets/common/admin/admin_base.scss
```diff
@@ -931,6 +931,20 @@ table#user-badges {
   }
 }
 
+.simple-list-input {
+  display: flex;
+
+  input {
```
jjaffeux, line 937: can we use a class please?

app/assets/stylesheets/common/admin/admin_base.scss
```diff
@@ -931,6 +931,20 @@ table#user-badges {
...
+    margin: 0;
+    box-sizing: border-box;
+    flex: 1 0 0px;
+  }
+
+  button {
```
jjaffeux, line 943: can we use a class please?

app/assets/javascripts/admin/templates/components/simple-list.hbs
pmusaraj: Hmm, I was forced to do this by pre-commit hooks though, before I had this indentation, there were linting issues: 

image

app/assets/stylesheets/common/admin/admin_base.scss
pmusaraj: I copied the styles from the value-list SCSS, but found that this was an IE11 fix: 

https://github.com/discourse/discourse/commit/ba5e17a7a98af178c75d00d7be817c471fc88911 

I know we don't support IE11 anymore, but I would keep this as 0px for consistency.

app/assets/javascripts/admin/templates/components/simple-list.hbs
jjaffeux: correct indentation is:

```
{{input
  title=value
  value=value
  class="value-input"
  focus-out=(action "changeValue" index)
}}
```

app/assets/stylesheets/common/admin/admin_base.scss
jjaffeux: 👍

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    this.set(
+      "collection",
+      this._splitValues(this.values, this.inputDelimiter || "\n")
+    );
+  },
+
+  keyDown(event) {
+    if (event.keyCode === 13) this.addValue(this.newValue);
```
jjaffeux, line 29: guess it would be cleaner to stop the event here if we do something with it

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,81 @@
...
+    this.collection.addObject(value);
+    this._saveValues();
+  },
+
+  _removeValue(value) {
+    this.collection.removeObject(value);
+    this._saveValues();
+  },
+
+  _replaceValue(index, newValue) {
```
jjaffeux, line 60: I'm unsure this has value as a fn here given we only use it once

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,81 @@
...
+  addValue(newValue) {
+    if (this.inputInvalid) return;
+
+    this.set("newValue", null);
+    this._addValue(newValue);
+  },
+
+  @action
+  removeValue(value) {
+    this._removeValue(value);
```
jjaffeux, line 47: I'm unsure this has value as a fn here given we only use it once

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,81 @@
...
+    this._saveValues();
+  },
+
+  _saveValues() {
+    if (this.inputType === "array") {
+      this.set("values", this.collection);
+      return;
+    }
+
+    this.set("values", this.collection.join(this.inputDelimiter || "\n"));
```
jjaffeux, line 71: I would rather want to have an onChange event, instead of mutating this inside the component, values should be read only inside the component, given it's coming from outside the component.

app/assets/javascripts/admin/components/simple-list.js
jjaffeux, line 29: AFAIK keyCode is getting deprecated, I know we use it the app, but well if we are writing new code, maybe we should avoid it.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

app/assets/javascripts/admin/components/simple-list.js
```diff
@@ -0,0 +1,81 @@
...
+    this._saveValues();
+  },
+
+  _removeValue(value) {
+    this.collection.removeObject(value);
+    this._saveValues();
+  },
+
+  _replaceValue(index, newValue) {
+    this.collection.replace(index, 1, [newValue]);
```
jjaffeux: This is something slightly concerning that you are mixing KVO stuff (removeObject) and non KVO stuff (replace) it all works because in the end you do a set I guess.

Maybe you should consider calling arrayContentDidChange when doing replace, this way it will be consistent: https://api.emberjs.com/ember/3.18/classes/MutableArray/methods/removeObject?anchor=removeObject

tgxworld:
Minor comment, how does the interface scale as the list grows? Do we collapse the list at some point?

pmusaraj:
I don’t think we currently collapse the list for any of the list type settings.

jjaffeux:
My feeling is that any gigantic list on this is probably a misusage and should be handled differently. Maybe we should enforce some max items limit ?

pmusaraj:
I would not add a max limit here, unless we start seeing complaints. For the vast majority of site settings, the items in the list will remain to a reasonable minimum. 

I do think the user fields dropdown lists should be collapsible beyond 20-30 items.

PR #9972, opened 2020-06-03
DEV: Deprecate Category#url_with_id in favor of Category#url

udan11:
I checked the plugins in all-the-plugins repo and they should be compatible with the change.


app/models/permalink.rb
```diff
@@ -80,7 +80,7 @@ def target_url
     return external_url if external_url
     return "#{Discourse::base_uri}#{post.url}" if post
     return topic.relative_url if topic
-    return "#{category.url}/#{category.id}" if category
+    return "#{category.url}" if category
```
ZogStriP, line 83:
```suggestion
    return category.url if category
```

PR #9979, opened 2020-06-04
FIX: select-kit components not being hidden.

tgxworld:


jjaffeux:
Ok so theres some history on this, ember changed support for isHidden at some point, so this might be why it ended up veing clunky. No hard feedlings on this, looks ok. It’s possible that something relies on DOM instead of class somewhere... We will see.

tgxworld:
Hmm I got a failing test, will look into this tomorrow.

tgxworld:
@jjaffeux Updated the PR.

PR #9985, opened 2020-06-05
DEV: adds SKIP_PLUGIN to plugin:install_all_official

jjaffeux:
Usage:

SKIP_PLUGINS=discourse-encrypt,discourse-calendar rake plugin:install_all_official


SamSaffron:
We can't use SKIP_PLUGINS cause it is used elsewise in the rake task 

I made a commit with SKIP_INSTALL_PLUGINS but trying to figure out why it is not picking up correctly

PR #9987, opened 2020-06-05
PERF: Cache PrettyText instance for rendering composer preview

davidtaylorhq:
Previously we were building pretty-text from scratch on every keypress. In my browser, this cuts cooking a simple post from 2.5ms to 0.5ms

Before:
Screenshot 2020-06-05 at 11 20 05

After:
Screenshot 2020-06-05 at 11 59 01

### DIFF
--- a/app/assets/javascripts/discourse/app/components/d-editor.js
+++ b/app/assets/javascripts/discourse/app/components/d-editor.js
@@ -9,7 +9,7 @@ import discourseComputed, {
 } from "discourse-common/utils/decorators";
 import { categoryHashtagTriggerRule } from "discourse/lib/category-hashtags";
 import { search as searchCategoryTag } from "discourse/lib/category-tag-search";
-import { cookAsync } from "discourse/lib/text";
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/app/components/d-editor.js
```diff
@@ -344,15 +344,26 @@ export default Component.extend({
     return toolbar;
   },
 
+  cachedCookAsync(text) {
```
eviltrout, line 347: You could probably make it even simpler by setting `this.cookFunction` in an `init()` method when the component is created and then you could call it without checking if it exists every time. I doubt it would change performance much but it would be simpler code.

davidtaylorhq: The problem is that loading pretty-text is asynchronous. So we can't set it synchronously during `init`. 

I guess we could trigger the loadScript during `init`, but then we would still have to check whether loading is done before cooking anything.

PR #9989, opened 2020-06-05
FIX: Redirects containing Unicode usernames didn't work

gschlager:
@eviltrout I wonder if we should remove the following two routes. I couldn't find any references in our code. AFAIK we are always using the `/my/...` routes.

https://github.com/discourse/discourse/blob/0f09fd22f3afa6ba22499e2fbf6d50a4574831aa/config/routes.rb#L318

https://github.com/discourse/discourse/blob/0f09fd22f3afa6ba22499e2fbf6d50a4574831aa/config/routes.rb#L377

I didn't even know that `/email_preferences` and `/user_preferences` were a thing. They exist since the beginning of the project.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -760,7 +760,7 @@ class ApplicationController < ActionController::Base
     return if !current_user
     return if !should_enforce_2fa?
 
-    redirect_path = "#{GlobalSetting.relative_url_root}/u/#{current_user.username}/preferences/second-factor"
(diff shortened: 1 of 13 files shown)


gschlager:
@eviltrout Thanks for the review. I'm mentioning you because I updated the PR to fix the usernames in avatar redirects and I added a commit to [remove the unused routes](https://github.com/discourse/discourse/pull/9989/commits/152a60716afc62da85ce2571ad2de18f0bede3f8).

PR #9991, opened 2020-06-05
FIX: Enforce maximum of 100 years on all site settings counted in days

riking:
This will prevent Postgres "timestamp out of range" errors on peoples' sites.

### DIFF
--- a/config/site_settings.yml
+++ b/config/site_settings.yml
@@ -148,11 +148,11 @@ basic:
   suggested_topics_unread_max_days_old:
     default: 90
     min: 0
-    max: 10000
+    max: 36500
   suggested_topics_max_days_old:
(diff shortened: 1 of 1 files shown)


riking:
Plugins frontend failure looks unrelated?

davidtaylorhq:
Do we need a migration to clean up any existing values? IIRC site setting validations are only enforced when the setting is being changed.

riking:
There's not really a danger for these unless they're actually causing Postgres errors, which should show up in a site's `/logs`. Proactive remediation is probably unnecessary.

SamSaffron:
This change looks good to me, agree we can just react on errors.

If this history becomes a systemic issue leading to support we can add a migration.

PR #9992, opened 2020-06-05
DEV: Raise special check_xhr exception in tests

riking:
Solves a papercut in tests when you forget .json on an endpoint by raising a special exception.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -107,6 +107,7 @@ class ApplicationController < ActionController::Base
   end
 
   class RenderEmpty < StandardError; end
+  class CheckXhrForTest < StandardError; end
   class PluginDisabled < StandardError; end
 
   rescue_from RenderEmpty do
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -661,7 +662,11 @@ def authorize_mini_profiler
   def check_xhr
     # bypass xhr check on PUT / POST / DELETE provided api key is there, otherwise calling api is annoying
     return if !request.get? && (is_api? || is_user_api?)
-    raise ApplicationController::RenderEmpty.new unless ((request.format && request.format.json?) || request.xhr?)
+    unless ((request.format && request.format.json?) || request.xhr?)
+      # Tests usually aren't trying to request HTML pages. Give a special error
+      raise ApplicationController::CheckXhrForTest.new if Rails.env.test?
```
tgxworld, line 667: Instead of doing this check at runtime, we can defined the error as a constant so that the check is only done.

riking: Do you mean switching out the method body based on Rails.env.test?

tgxworld: I was thinking something like

```
ERROR_CLASS = Rails.env.test? ? ApplicationController::CheckXhrForTest : ApplicationController::RenderEmpty

def check_xhr
    ....
    raise ERROR_CLASS.new
end
```

riking:
Still failing a few tests that are dependent on the old behavior.

app/controllers/application_controller.rb
SamSaffron: I think the cost of the check is so cheap and it only happens anyway in exceptional cases (which is key). Code is somewhat simpler this way, I say we just keep it.

SamSaffron:
@riking I would like to close this off, can you merge this when you start your day?

SamSaffron:
I am just too mixed on all of this, I guess we just do this in rubocop and can catch it on save at lint time?

PR #9994, opened 2020-06-05
Note on 1-2GB server size

zyphlar:
It's not obvious upon reading this that during Digitalocean's setup routine, a 1-2GB server size will cause a swap file to be created, which is bad practice on SSDs and can lead to hardware degradation. Choosing 3+ GB, at least for Digitalocean servers, is recommended.

### DIFF
--- a/docs/INSTALL-cloud.md
+++ b/docs/INSTALL-cloud.md
@@ -8,7 +8,7 @@ Create your new cloud server, for example [on DigitalOcean][do]:
 
 - The default of **Ubuntu 18.04 LTS x64** works fine. At minimum, a 64-bit Linux OS with a kernel version of 3.10+ is required.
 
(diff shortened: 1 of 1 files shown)


coding-horror:
This is not the case, swap is allowed on cloud hosts. It is largely an
insurance policy. Otherwise the minute you go one byte over memory used the
out of memory reaper will kill the process.

zyphlar:
I didn't say it wasn't allowed, I said it wasn't recommended. Digitalocean
explicitly discourages it, and in following their recommendation I had to
delete/recreate my server a few times to get the right settings:
https://www.digitalocean.com/community/tutorials/how-to-add-swap-space-on-ubuntu-20-04

securingmom:
Explicitly discourages it by providing directions to do it seems a tad conflicted

PR #9995, opened 2020-06-06
Support plugin compatibility version manifests

featheredtoast:
Adds a new rake task `plugin:checkout_compatible_all` and
`plugin:checkout_compatible[plugin-name]` that check out compatible plugin
versions.

Supports a .discourse-compatibility file in the root of plugins that list out
a plugin's compatibility with certain discourse versions:

eg: .discourse-compatibility
```
2.5.0.beta6: some-git-hash
2.4.4.beta4: some-git-tag
2.2.0: git-reference
```

This ensures older Discourse installs are able to find and install older
versions of plugins without intervention, through the manifest only.

It iterates through the versions in descending order. If the current Discourse
version matches an item in the manifest, it checks out the listed plugin target.
If the Discourse version is greater than an item in the manifest, it checks out
the next highest version listed in the manifest.

If no versions match, it makes no change.

It will also be pretty simple to integrate this task into docker_manager and our pups files once merged.

### DIFF
--- a/lib/tasks/plugin.rake
+++ b/lib/tasks/plugin.rake
@@ -88,6 +88,45 @@ task 'plugin:update', :plugin do |t, args|
   abort('Unable to pull latest version of plugin') unless update_status
 end
 
+desc 'pull compatible plugin versions for all plugins'
+task 'plugin:pull_compatible_all' do |t|
+  # Loop through each directory
+  plugins = Dir.glob(File.expand_path('plugins/*')).select { |f| File.directory? f }
+  # run plugin:pull_compatible
+  plugins.each do |plugin|
+    next unless File.directory?(plugin + "/.git")
+    Rake::Task['plugin:pull_compatible'].invoke(plugin)
+    Rake::Task['plugin:pull_compatible'].reenable
+  end
+end
+
+desc 'pull a compatible plugin version'
+task 'plugin:pull_compatible', :plugin do |t, args|
+
+  plugin = ENV['PLUGIN'] || ENV['plugin'] || args[:plugin]
+  plugin_path = plugin
+  plugin = File.basename(plugin)
+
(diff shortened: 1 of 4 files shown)


riking:
Question about the developer experience here: does the `.discourse-compatibility` file need to be present on both the master branch and the branch it's pointing to, or just one of those? What happens if it's inconsistent?

featheredtoast:
The local version watches the "upstream" branch version, wherever the repo was pulled via `git -C '#{path}' show HEAD@{upstream}:.discourse-compatibility` so any updates to the file on the upstream branch will be properly honored in future pulls.

So if you did something like `git clone -b my-custom-branch github.com/discourse-plugin` it'd follow the `.discourse-compatibility` file from my-custom-branch. Updates to the file on the my-custom-branch will be honored. Changes on any other branch (origin/master for instance) will not have any effect here.

If you did a `git clone github.com/discourse-plugin` and that was set up to be by default origin/master, it'd follow the file on `origin/master` as the single source of truth.

This means that: We can change/update compatibility files as we need and they will be honored within the main origin master versions, and also if there is a fork for a feature on a separate branch, but within the same repo that we can still separately track compatibility in its own branch.

If there is no file in the current branch's upstream, it does nothing, even if it exists in other branches, or in history.

lib/tasks/plugin.rake
```diff
@@ -88,6 +88,45 @@ task 'plugin:update', :plugin do |t, args|
   abort('Unable to pull latest version of plugin') unless update_status
 end
 
+desc 'checkout all compatible plugin versions'
+task 'plugin:checkout_compatible_all' do |t|
```
eviltrout, line 92: I suggest we change the name of the tasks to `plugin:pull_compatible` - because it's already been checked out and this is about upgrading the plugin.

PR #10000, opened 2020-06-09
FIX: flaky post_alerter_spec.rb because of missing order

lis2:
Randomly failing spec - https://github.com/discourse/discourse/blob/master/spec/services/post_alerter_spec.rb#L756


lis2:
P1059429

PR #10001, opened 2020-06-09
FIX: Make Email::Styles preserve email structure when it exists

danielwaterworth:
Applying styles shouldn't alter the structure of the HTML that is being
manipulated. In particular, meta tags were being omitted during this
process.

This does mean that, now, when we create emails, they have a html5
doctype by default. I changed the doctype in the default template to
match since it's being ignored anyway (because we're using an html5
parser).


lis2:
@danielwaterworth do you think we should have additional method like `contains_html?` to decide if we should use `Nokogiri::HTML5.parse(@html).to_html` or `Nokogiri::HTML5.parse(@html).at("body").to_html` . My only concern is that if you parse fragment without <HTML> tag, nokogumbo will add it:
image

danielwaterworth:
I don't really have a handle on how useful that would be. The default template has an html element, so I assume, in the vast majority cases, that we will be dealing with html email with that structure. I wonder, in fact, whether we might be able to remove the body check and always use `HTML5.parse`.

SamSaffron:
I support this fix, but we have waited a bit too long here to merge it. Let's merge first thing post 2.5 release.

PR #10003, opened 2020-06-09
DEV: adds mouse{Over,Out} and touch{Start,End} to widgets

jjaffeux:


app/assets/javascripts/discourse/app/widgets/hooks.js
```diff
@@ -112,6 +120,18 @@ WidgetClickHook.setupDocumentCallback = function() {
     }
   };
 
+  $(document).on("mouseover.discourse-widget", e => {
```
ZogStriP, line 123: Do we need to use jQuery here?

jjaffeux: No we dont, the whole file is following this structure, I can note to fix the whole thing after the release :+1:

PR #10004, opened 2020-06-09
DEV: allows to call a widget function without re-rendering the widget

jjaffeux:


app/assets/javascripts/discourse/app/widgets/widget.js
```diff
@@ -323,6 +323,13 @@ export default class Widget {
...
+  callWidgetFunction(name, param) {
+    const widget = this._findAncestorWithProperty(name);
+    if (widget) {
+      return widget[name].call(widget, param);
+    }
+  }
+
   sendWidgetAction(name, param) {
     return this.rerenderResult(() => {
       const widget = this._findAncestorWithProperty(name);
```
ZogStriP, line 335: Shouldn't this method use the new `callWidgetFunction`?

jjaffeux: Yes, maybe... In one hand you DRY on the other hand you avoid some indirection of if call had found a widget then return else... Not sure it's very worth it.

jjaffeux: I made the change 👨‍🎨

PR #10005, opened 2020-06-09
DEV: allows to prepend widgets to post nav controls

jjaffeux:


app/assets/javascripts/discourse/app/widgets/post-menu.js
```diff
@@ -552,8 +552,18 @@ export default createWidget("post-menu", {
...
+    const extraControls = applyDecorators(this, "extra-controls", attrs, state);
+    const beforeExtraControls = applyDecorators(
+      this,
+      "before-extra-controls",
+      attrs,
+      state
+    );
+
+    let controlsButtons = beforeExtraControls.concat(visibleButtons);
+    controlsButtons = controlsButtons.concat(extraControls);
```
ZogStriP, line 564: Can we do this instead?

```suggestion
    const controlsButtons = [...beforeExtraControls, ...visibleButtons, ...extraControls];
```

PR #10007, opened 2020-06-09
FIX: Use include-subcategories filter in report export

udan11:


app/controllers/export_csv_controller.rb
```diff
@@ -18,7 +18,7 @@ def export_entity
   def export_params
     @_export_params ||= begin
       params.require(:entity)
-      params.permit(:entity, args: [:name, :start_date, :end_date, :category_id, :group_id, :trust_level]).to_h
+      params.permit(:entity, args: [:name, :start_date, :end_date, :category_id, :group_id, :trust_level, :include_subcategories]).to_h
```
ZogStriP, line 21: Might be worth extracting this array into a constant.

app/jobs/regular/export_csv_file.rb
```diff
@@ -221,6 +221,9 @@ def report_export
       if @extra[:group_id].present?
         @extra[:filters][:group] = @extra[:group_id].to_i
       end
+      if @extra[:include_subcategories].present?
+        @extra[:filters][:'include-subcategories'] = !!ActiveRecord::Type::Boolean.new.cast(@extra[:include_subcategories])
```
ZogStriP, line 225: Why are you using dash "`include-subcategories`" instead of underscore "`include_subcategories`" on the server-side?

udan11: That is the name of the filter. The client uses 'include-subcategories' which is changed to 'include_subcategories' when a request is made and then the server renames this back.

I tried to follow what we have now, which is not the best. I will change it.

PR #10025, opened 2020-06-11
Do not re-register items in inject-discourse-objects

featheredtoast:
It appears that this breaks some test assumptions in some of our plugins
when qunit is hitting category routes.

This might be an issue with the plugin, but the warnings/errors point here. Adding the following allows the tests to pass.


eviltrout:
This is a weird fix - why would the objects be registered twice? Can you trace it and figure out why that is happening? I would much prefer to fix that.

featheredtoast:
Oh hm, looks like the test failures were due to some earlier loading issues in this particular plugin, and doesn't have anything to do in core. Resulting failures were red herrings - apologies for the noise!

PR #10028, opened 2020-06-11
 FIX: Hide bootbox after a negative answer

nbianca:
Starting to reply and then editing a post was not possible because of a
bootbox which kept showing up.


ZogStriP:
I don't think the PR title is correct?

nbianca:
Oops, that's right. I think I choose another branch before I selected the right one and the title stuck somehow. :woman_facepalming:

PR #10032, opened 2020-06-11
FEATURE: Check if selectable avatars exist before enabling them

nbianca:


config/locales/server.en.yml
```diff
@@ -2253,6 +2253,7 @@ en:
       max_username_length_exists: "You cannot set the maximum username length below the longest username (%{username})."
       max_username_length_range: "You cannot set the maximum below the minimum."
       invalid_hex_value: "Color values have to be 6-digit hexadecimal codes."
+      empty_selectable_avatars: "You must first upload selectable avatars before enabling this setting."
```
ZogStriP, line 2256: Let's enforce at least 2 avatars in the list otherwise there's no point in enabling this feature to select just 1 avatar.

coding-horror:
Well you _could_ use the same avatar for everyone, but there's no need to
have a "selection" at that point.. probably not worth coding for

ZogStriP:
If they want to use the same avatar for everyone, then they should use the "default avatars" site setting.

"selectable avatars"  should only be available where the users have a choice.

coding-horror:
Aha yes! 👍

PR #10034, opened 2020-06-11
DEV: allow to have duplicate topic titles if category is different

sbernhard:


coding-horror:
Is there discussion about this on meta.discourse.org? Because the URL slugs
would still be nearly identical, which is a bad thing, unless the category
is in the slug

sbernhard:
See https://meta.discourse.org/t/could-we-check-the-duplicated-topic-title-for-the-each-category-only/98790/6

Same is true if allow duplicate titles is completely turned on. So, the limitation on the different category would still be a improvement.

eviltrout:
This breaks backwards compatibility because the boolean setting is now a string. Anyone who changed it will have a bool in the database. 

At the very least we'll need a migration to migrate the old setting to the new one.

We should also investigate if any plugins use the setting as that code will be broken too.

sbernhard:
That's true. Of course a migration can be added or would you prefer to have a new setting to configure this like:
Allow duplicate titles in different categories

eviltrout:
The secondary setting does seem much easier.

sbernhard:
Yes, much easier but the naming of the setting and that a user knows how to switch which setting is more difficult. Any suggestion is highly appreciated 😀

sbernhard:
Added a new setting. 

Looks like the tests are failing because of another reason. Can you verfiy @eviltrout ?

config/locales/server.en.yml
```diff
@@ -1449,6 +1449,7 @@ en:
     category_search_priority_very_high_weight: "Weight applied to ranking for very high category search priority."
     allow_uncategorized_topics: "Allow topics to be created without a category. WARNING: If there are any uncategorized topics, you must recategorize them before turning this off."
     allow_duplicate_topic_titles: "Allow topics with identical, duplicate titles."
+    allow_duplicate_topic_titles_at_different_category: "Allow topics with identical, duplicate titles if the category is different. (precondition: allow_duplicate_topic_titles=false)."
```
eviltrout, line 1452:
```suggestion
    allow_duplicate_topic_titles_at_different_category: "Allow topics with identical, duplicate titles if the category is different. allow_duplicate_topic_titles must be false."
```

config/site_settings.yml
```diff
@@ -716,6 +716,7 @@ posting:
     default: true
     refresh: true
   allow_duplicate_topic_titles: false
+  allow_duplicate_topic_titles_at_different_category: false
```
eviltrout, line 719:
```suggestion
  allow_duplicate_topic_titles_category: false
```
This is a bit long. How about ^

sbernhard:
All tests are finally green @eviltrout

PR #10036, opened 2020-06-12
FIX: allows PM owner to remove any user if >= TL2

jjaffeux:
- `< TL2 user` can remove itself:
Screenshot 2020-06-12 at 10 35 50

- `>= TL2 user` can remove everyone:
Screenshot 2020-06-12 at 10 35 58


ZogStriP:
What happens if OP remove themselves from the PM? :thinking:

jjaffeux:
> What happens if OP remove themselves from the PM? 🤔

replaced by system user

Screenshot 2020-06-12 at 12 11 59

So we keep trace of the OP being crated by this user, but we replace its presence in the topic by system.

PR #10044, opened 2020-06-14
FIX: Set default value for poll result field

nbianca:


ZogStriP:
There should probably be a test to ensure this does work as expected 😉

SamSaffron:
reasonably low risk and a nice fix, but still lets wait post 2.5

PR #10053, opened 2020-06-16
FEATURE: allows published pages to be public

jjaffeux:
This commit also introduces the `show_public_pages_with_login_required` site settings, which will allow to bypass `login_required`.


config/locales/client.en.yml
```diff
@@ -2513,11 +2513,14 @@ en:
         publish: "Publish"
         description: "When a topic is published as a page, its URL can be shared and it will displayed with custom styling."
         slug: "Slug"
+        public: "Public"
+        public_description: "A public page won't follow the visibility restrictions applied to the associated topic."
```
eviltrout, line 2517:
```suggestion
        public_description: "People can see the page even if the associated topic is private."
```

config/locales/server.en.yml
```diff
@@ -2131,6 +2131,7 @@ en:
     returning_user_notice_tl: "Minimum trust level required to see returning user post notices."
     returning_users_days: "How many days should pass before a user is considered to be returning."
     enable_page_publishing: "Allow staff members to publish topics to new URLs with their own styling."
+    show_public_pages_with_login_required: "Allows to display public published pages when login is required and no user is connected."
```
eviltrout, line 2134:
```suggestion
    show_public_pages_with_login_required: "Anonymous users can see published pages, even when login is required."
```

config/site_settings.yml
```diff
@@ -947,6 +947,8 @@ posting:
     max: 36500
   enable_page_publishing:
     default: false
+  show_public_pages_with_login_required:
```
eviltrout, line 950:
```suggestion
  show_published_pages_login_required:
```
A minor change but I think it should have "published" in there and we can omit things like "with"

app/controllers/published_pages_controller.rb
```diff
@@ -5,14 +5,30 @@ class PublishedPagesController < ApplicationController
...
   def show
     params.require(:slug)
 
     pp = PublishedPage.find_by(slug: params[:slug])
     raise Discourse::NotFound unless pp
 
-    guardian.ensure_can_see!(pp.topic)
+    return if enforce_login_required!
+
+    if !pp.public
```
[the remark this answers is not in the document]
jjaffeux: Indeed will look at implementing this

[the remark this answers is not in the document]
jjaffeux: After trying to implement this, I consider it kinda complicated given we have to handle the /login logic and the custom exception to avoid including ember. I will postpone this for now, and think of a better solution.

PR #10057, opened 2020-06-16
ARIA - add aria-label on topic input and post textarea

featheredtoast:
Add reply form role with label as create/reply, and aria-labels on text inputs

Regarding d-textarea - Is this the only/best way to add html attributes to predefined components, until Ember Octane and Glimmer components are more widely used?


eviltrout:
> Regarding d-textarea 

Seems good to me!

PR #10060, opened 2020-06-17
make unix domain sockets listening example match web.socketed.template.yml

adrelanos:
I cannot find any google results for `"server unix:/var/www/discourse/tmp/sockets/thin.0.sock;"`. Therefore I think it is not actually used by anyone in public writing about discourse in that way.

It would be more useful to use the same as [web.socketed.template.yml](https://github.com/discourse/discourse_docker/blob/master/templates/web.socketed.template.yml). Would make the example a bit easier to understand.

### DIFF
--- a/config/nginx.sample.conf
+++ b/config/nginx.sample.conf
@@ -4,10 +4,8 @@ types {
 }
 
 upstream discourse {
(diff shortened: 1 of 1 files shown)


eviltrout:
Hey if you sign the CLA I can take a look at this!

adrelanos:
Done.

PR #10065, opened 2020-06-17
FIX: Hide the post history for TL4

nbianca:
TL4 users should not see the edit history if `edit_history_visible_to_public` site setting is disabled.


ZogStriP:
LGTM 👍 

I wonder, are there any translations that we need to fix as well?

nbianca:
The translation already matches the behavior. :blush:

PR #10076, opened 2020-06-18
FIX: Do not change tracked categories for staged users

nbianca:


nbianca:
Is the only place I could find.

PR #10077, opened 2020-06-18
FIX: Use correct URL for unsubscribe

nbianca:


spec/mailers/subscription_mailer_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+require "rails_helper"
+
+describe SubscriptionMailer do
+  fab!(:user) { Fabricate(:user) }
+
+  subject { SubscriptionMailer.confirm_unsubscribe(user) }
+
+  it "contains the right URL" do
+    expect(subject.body).to include("http://test.localhost/email/unsubscribe/#{UnsubscribeKey.last.key}")
```
ZogStriP, line 11: Could this be something like this instead?

```suggestion
    expect(subject.body).to include("#{Discourse_base_url}/email/unsubscribe/#{UnsubscribeKey.last.key}")
```

SamSaffron:
This fix looks good to me, but it is a bit unclear what it is fixing can you expand? I do support merging this asap.

PR #10078, opened 2020-06-18
FIX: Allow multiple hyphens in host

cdeleeuwe:
Currently having multiple hyphens (-) in the host, throws an "invalid" error. For example, currently ```deploy-preview-1--example.netlify.app```, would not be considered valid.


eviltrout:
Hi there! Thanks for this PR. Could you add a unit test for this functionality please?

cdeleeuwe:
@eviltrout I have added the unit test. Thanks a lot for reviewing. If I should add anything else, let me know :)

PR #10079, opened 2020-06-18
FEATURE: Support converting HEIF images to JPEG

pmusaraj:
This is an optional feature (default off) to convert HEIF images (like live photos from iOS devices, for example) to be converted to jpegs when uploaded. Browsers do not support rendering HEIFs at this time. 

Requires allowing `heic` and/or `heif` in the `authorized_extensions` site setting. 

Developers on non-Mac machines will need to ensure ImageMagick includes HEIC support, otherwise the specs will not pass. See also https://github.com/techAPJ/install-rails/pull/21.


ZogStriP:
I wonder, can we do this on the client-side? How CPU-intensive is it? Any ideas?

pmusaraj:
@ZogStriP There is a library that can do this client-side, see https://github.com/alexcorvi/heic2any for an implementation but this seems considerably trickier. Heic2any does not retain metadata, which makes it less useful for converting files at upload. The library maintainer suggests it be used for display purposes only. But, while storing HEICs is more efficient (better compression), it would complicate lots of things in our upload pipeline, since HEICs aren't recognized as images (no thumbnail or optimized image generation, markup would need to be converted from an attachment to an image, etc). 

Also, the client-side library is quite large, at 1.15Mbs minified.

PR #10083, opened 2020-06-18
FIX:Add migration to delete tracking state for staged users

nbianca:


db/migrate/20200618175923_delete_tracking_state_for_staged_users.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class DeleteTrackingStateForStagedUsers < ActiveRecord::Migration[6.0]
+  def change
```
ZogStriP, line 4: I think this should be split into `up`/`down` methods with the `down` method throwing an "NotReversibleMigration` error.

SamSaffron:
change looks very safe to me!

PR #10084, opened 2020-06-18
FEATURE: Don't display muted/ignored users under "who liked"

eviltrout:
Previously, if you clicked on the heart icon below a post
it would show you the avatar for a user even if you ignored or muted
them.

This commit will instead display a (?) icon. The count of likes will
remain correct, but you needn't be reminded of the person you
preferred not to see.


coding-horror:
I was thinking a blank grey circle would be better than a ? icon which might call attention to it, LIKE THE RIDDLER

PR #10085, opened 2020-06-18
DEV: move user references deletion code to `before_destroy`.

vinothkannans:
Moving the `delete_source_user_references` method code from user merger service to user model.


eviltrout:
Is there a context for this change?

eviltrout:
Nevermind I got it!

PR #10089, opened 2020-06-19
FIX: Hide PM tags if the site setting is disabled

nbianca:


app/controllers/list_controller.rb
```diff
@@ -144,6 +144,10 @@ def group_topics
 
   def self.generate_message_route(action)
     define_method("#{action}") do
+      if action == :private_messages_tag && !guardian.can_tag_pms?
+        raise Discourse::NotFound
```
eviltrout, line 148: I don't see any tests for this guard - it seems fairly important!

PR #10093, opened 2020-06-20
FIX: Make migrations from S3 more robust; fix bare URL migration

johnsonm:
Fix handling of bare URLs to (more) correctly process URLs present
in s3 or a clone when migrating to local, including changing to an
`upload:` psuedo-protocol URL, tagging audio and video files, using
the `name` syntax, and in the process not creating non-URLs
that point to nothing.

Fix `batch_migrate_from_s3` to take both a max number of posts to
actually modify in a migration (the original but failed intent of
limit) and a limit of total posts to consider for migration (for
cheaper database queries).  Because a primary goal of the `max`
limit is debugging, also use it to cause verbose logging output.

Print diagnostic messages for conditions that might require manual
intervention, such as file too large (if upload sizes were reduced
in site settings) or failure to download (which can happen on an
intermittent basis and might require re-running the migration).

Finally, in the case of a truly unexpected error, print the error
and stop processing, so that error cascades do not become hidden
failures stopping a successful migration, but can instead be resolved.

### DIFF
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -225,7 +225,7 @@ def migrate_from_s3(max: nil, limit: nil)
       end
 
       if updated
-        post.save!
+        post.save!(validate: false)
         post.rebake!
         posts_modified += 1
         if max
--- a/lib/tasks/uploads.rake
+++ b/lib/tasks/uploads.rake
@@ -99,6 +99,12 @@ task "uploads:batch_migrate_from_s3", [:max, :limit] => :environment do |_, args
   ENV["RAILS_DB"] ? migrate_from_s3(max: args[:max], limit: args[:limit]) : migrate_all_from_s3(max: args[:max], limit: args[:limit])
 end
 
+def wait_for_queue_tasks(max_queue_tasks)
+  while Sidekiq::Stats.new.enqueued > max_queue_tasks
+    sleep 1
+  end
+end
+
 def guess_filename(url, raw)
   begin
     uri = URI.parse("http:#{url}")
@@ -130,6 +136,7 @@ def migrate_from_s3(max: nil, limit: nil)
     exit 1
   end
 
+  max_queue_tasks = ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"] ? ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"].to_i : 0
   posts_modified = 0
   # treat "0" the same as nil input; 0 means no limit and no max
   limit = limit.to_i if limit && limit != "0"
@@ -238,6 +245,8 @@ def migrate_from_s3(max: nil, limit: nil)
           puts "Modification limit #{max} reached, halting migration"
           return
         end
+        # don't spam the default queue, limit rate of modification
+        wait_for_queue_tasks max_queue_tasks
       else
         putc "."
       end
@@ -939,6 +948,7 @@ def determine_upload_security_and_posts_to_rebake(uploads_to_update, mark_secure
 end
 
 def inline_uploads(post)
+  max_queue_tasks = ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"] ? ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"].to_i : 0
   replaced = false
 
   original_raw = post.raw
@@ -974,10 +984,12 @@ def inline_uploads(post)
     post.save_custom_fields
     post.save!(validate: false)
     post.rebake!
+    wait_for_queue_tasks max_queue_tasks
   end
 end
 
 def inline_img_tags(post)
+  max_queue_tasks = ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"] ? ENV["DISCOURSE_MIGRATION_MAX_ENQUEUED"].to_i : 0
   replaced = false
 
   original_raw = post.raw
@@ -1015,6 +1027,7 @@ def inline_img_tags(post)
     post.save_custom_fields
     post.save!(validate: false)
     post.rebake!
+    wait_for_queue_tasks max_queue_tasks
   end
 end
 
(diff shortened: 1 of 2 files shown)


johnsonm:
The only failures I've seen with this code running so far on my live site are:
* Not migrating videos larger than the current max file size; not a bug.
* Transient failures from digital ocean spaces that are resolved by the workaround I have at https://github.com/johnsonm/discourse/commit/7dfac12a2ea6ec04ba4e0616b4e0dbd1d806cff7 — I can pull that commit into this PR or not, as you wish. I don't know the right number of retries in general, and I didn't implement exponential backoff on the retries yet since it was intended as a quick hack to see if it would work around my problem.

I think this is ready for review, and valuable generally.

johnsonm:
I've run into the need to retry download enough that it's worth pulling in that workaround. It would be better in theory to do exponential backoff, but as far as I can tell this simple workaround hasn't failed yet. Good enough, I hope. It can always be improved later to be exponential, but really any S3 clone shouldn't be that unreliable.

johnsonm:
It looks like there might be inconsistent behavior of the upload mocks. They aren't acting like I would have thought; the hash _sometimes_ changes and sometimes doesn't. This is the source of the occasional test failures, it seems.

johnsonm:
https://meta.discourse.org/t/requesting-help-with-failure-finding-discourseredis-connector-running-tests/155599 addressed and test failures understood and resolved. I believe this is ready to merge.

lib/tasks/uploads.rake
```diff
@@ -174,40 +230,156 @@ def migrate_from_s3(limit: nil)
...
+      profiles_migrated += 1
+    end
+    if max && profiles_migrated >= max
+      puts "Modification limit #{max} reached, halting migration"
+      return
+    end
+    wait_for_queue_tasks max_queue_tasks
+  end
+
+  puts "All profiles uploads migrated.  Migrating other non-post uploads..."
```
CvX, line 332: The full list (I hope it's full 😅) of non-post upload types is for example in the `ShrinkUploadedImage` class:

https://github.com/discourse/discourse/blob/630ec7b49cd98358736c5445302c509c62c68653/lib/shrink_uploaded_image.rb#L165-L173

This section of the script would break all those types of upload references (other than the profile uploads handled above)

lib/tasks/uploads.rake
```diff
@@ -362,6 +534,52 @@ def clean_up_uploads
   puts "Done!"
 end
 
+################################################################################
+#                         report missing uploads                               #
+################################################################################
+
+# This might be used to repair after various failures
+task "uploads:report_missing_uploads" => :environment do
```
CvX, line 542: @davidtaylorhq Can I get your eyes on this? 🙂 (the name of the task and what it does)

lib/tasks/uploads.rake
```diff
@@ -107,18 +119,40 @@ def guess_filename(url, raw)
...
+  file
+end
+
+def create_local_upload(upload, file, user_id)
+  filename = upload.original_filename
+  origin = upload.origin
+  upload.destroy
+
+  UploadCreator.new(file, filename, origin: origin).create_for(user_id)
+end
```
CvX, line 149: This re-creates uploads, preserving only the actual file and `user_id`, `filename`, `origin` attributes?

I'm fine with creating new upload records (there was even an idea to make them immutable, to have a permanent record of changes) but all original attributes should be preserved.

lib/tasks/uploads.rake
```diff
@@ -93,8 +93,20 @@ task "uploads:migrate_from_s3" => :environment do
   ENV["RAILS_DB"] ? migrate_from_s3 : migrate_all_from_s3
 end
 
-task "uploads:batch_migrate_from_s3", [:limit] => :environment do |_, args|
-  ENV["RAILS_DB"] ? migrate_from_s3(limit: args[:limit]) : migrate_all_from_s3(limit: args[:limit])
+# limit is the limit to consider migrating, max is the max to actually migrate
+# This is to manage resources during testing; limit 1000 max 1 is a reasonable test
```
CvX, line 97: > limit is the limit to consider migrating, max is the max to actually migrate

I _think_ I know what both of these do by reading the code, but this comment doesn't make it clear. `max` is the **number** of **uploads** to be migrated? And `limit` is used in SQL queries? (does that need to be configurable?)

lib/tasks/uploads.rake
```diff
@@ -93,8 +93,20 @@ task "uploads:migrate_from_s3" => :environment do
...
+  ENV["RAILS_DB"] ? migrate_from_s3(max: args[:max], limit: args[:limit]) : migrate_all_from_s3(max: args[:max], limit: args[:limit])
+end
+
+def wait_for_queue_tasks(max_queue_tasks)
+  while Sidekiq::Stats.new.enqueued > max_queue_tasks
+    sleep 1
+  end
+end
+
+def tag(post)
```
CvX, line 108: Naming is hard 😅 but it would be better to use something that isn't a model name in Discourse. 😃

lib/tasks/uploads.rake
CvX: From meta:
> It feels like it makes sense to start with posts, since you want to rebake posts to change the URL in the cooked post after each migration; migrating all the uploads and then starting the rebakes would leave the site very broken for potentially a long time for a large site

You can process Uploads one by one, and rebake posts in-place, like [downsize_uploads](https://github.com/discourse/discourse/blob/630ec7b49cd98358736c5445302c509c62c68653/script/downsize_uploads.rb)/[ShrinkUploadedImage](https://github.com/discourse/discourse/blob/630ec7b49cd98358736c5445302c509c62c68653/lib/shrink_uploaded_image.rb) does.

---

Ideally we would extract the code from those two files into a class that'd be responsible for updating all objects related to the given upload (posts, user profiles, avatars, etc.)

CvX:
Closing in favor of #11703, though this branch will eventually be used as the starting point of the S3 migration script rewrite.

PR #10097, opened 2020-06-22
FIX: CookText may be gone before promise resolves

CvX:
Bug introduced in 293467a


romanrizzi:
Hey @CvX 👋 

It looks like `cook-text` can no longer resolve uploads URLs correctly after this change. Mind taking a look when you can?

PR #10099, opened 2020-06-22
DEV: Make component-test `afterEach` async aware

CvX:
Before this fix, if a test case was async, `afterEach` callback would be executed immediately, without waiting for the test to finish. 😬


SamSaffron:
Seems very safe to merge since it only impacts tests

PR #10105, opened 2020-06-22
DEV: Refactor basic-topic-list to use topic-list-item on mobile

davidtaylorhq:
Previously, basic-topic-list had its own implementation of topic-list-item on mobile, which made it more difficult to maintain and extend. The visible difference was that the basic-topic-list implementation had no large avatar on the left. This commit adds a new hideMobileAvatar parameter to topic-list-item and topic-list, and sets it to `true` for the basic-topic-list component.


davidtaylorhq:
Yes it should definitely wait 👍

PR #10106, opened 2020-06-22
UX: Make youtube embeds full width by default

xfalcox:
Before

image

After
image

### DIFF
--- a/plugins/lazy-yt/plugin.rb
+++ b/plugins/lazy-yt/plugin.rb
@@ -22,8 +22,11 @@ class Onebox::Engine::YoutubeOnebox
 
   def to_html
     if video_id && !params['list']
+
+      size_restricted = [params['width'], params['height']].any?
       video_width = (params['width'] && params['width'].to_i <= 695) ? params['width'] : 690 # embed width
       video_height = (params['height'] && params['height'].to_i <= 500) ? params['height'] : 388 # embed height
+      size_tags = ["width=\"#{video_width}\"", "height=\"#{video_height}\""]
 
       og = get_opengraph.data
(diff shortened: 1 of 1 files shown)


SamSaffron:
Is this going to work on mobile?

xfalcox:
> Is this going to work on mobile?

It does work @SamSaffron, but the ratio gets busted.

I pushed a new one, that while more complex, makes it fluid and more adaptable.

plugins/lazy-yt/plugin.rb
```diff
@@ -35,11 +38,15 @@ def to_html
...
-           data-height="#{video_height}"
+           #{size_restricted ? size_tags.map { |t| "data-#{t}" }.join(' ') : ""}
            data-parameters="#{embed_params}">
         <a href="https://www.youtube.com/watch?v=#{video_id}" target="_blank">
-          <img class="ytp-thumbnail-image" src="#{thumbnail_url}" width="#{video_width}" height="#{video_height}" title="#{escaped_title}">
+          <img class="ytp-thumbnail-image"
+               src="#{thumbnail_url}"
+               #{size_restricted ? size_tags.join(' ') : ""}
+               width="#{video_width}"
+               height="#{video_height}"
```
davidtaylorhq, line 48: If size_restricted is true, I think this will leave us with duplicate width/height attributes?

eviltrout:
How risky is this change? Should it be moved post release?

plugins/lazy-yt/plugin.rb
xfalcox: Oops, forgot to delete those lines. Thanks for the catch!

Fixed in 8a68db4

xfalcox:
> How risky is this change? Should it be moved post release?

I'd say it is safe enough to make the cut.

PR #10109, opened 2020-06-23
Adding before-create-topic-button plugin-outlet

marielaSAG:
on tag pages

### DIFF
--- a/app/assets/javascripts/discourse/app/templates/tags/show.hbs
+++ b/app/assets/javascripts/discourse/app/templates/tags/show.hbs
@@ -36,10 +36,10 @@
         {{/if}}
         
         {{plugin-outlet name="before-create-topic-button"
-        args=(hash
-          canCreateTopic=canCreateTopic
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/discourse/app/templates/tags/show.hbs
```diff
@@ -34,6 +34,13 @@
             }}
           {{/unless}}
         {{/if}}
+        
+        {{plugin-outlet name="before-create-topic-button"
+        args=(hash
```
eviltrout, line 39: This is super minor but the indentation here is off. Could you line up `args` with `plugin-outlet`?

marielaSAG: Sure, thanks, I did that.

eviltrout: Great, we've tagged it for merging after the next release.

SamSaffron:
Thanks, merged it in!

PR #10110, opened 2020-06-23
DEV: If there's a reviewable targeting an upload, we don't remove it.

romanrizzi:


SamSaffron:
Do we have any other use for this except for the AV plugin? If we do not yet I say we just wait on 2.6 and go with UploadReferences

romanrizzi:
We do not. Let's go with UploadReferences.

PR #10116, opened 2020-06-24
DEV: Add skip seed flag

oblakeerickson:
Sometimes we don't want seed data to run like when running our post_deploy tasks. The seed data has already ran so we don't need to run it a second time.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -157,10 +157,13 @@ task 'multisite:migrate' => ['db:load_config', 'environment', 'set_locale'] do |
   SeedFu.seed(seed_paths, /001_refresh/)
 
   execute_concurently(concurrency, exceptions) do |db|
+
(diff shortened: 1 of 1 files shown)


tgxworld:
This looks good to me :+1:

lib/tasks/db.rake
```diff
@@ -157,8 +157,10 @@ task 'multisite:migrate' => ['db:load_config', 'environment', 'set_locale'] do |
   SeedFu.seed(seed_paths, /001_refresh/)
 
   execute_concurently(concurrency, exceptions) do |db|
-    puts "Seeding #{db}"
-    SeedFu.seed(seed_paths)
+    if ENV['SKIP_SEED'] != '1'
+      puts "Seeding #{db}"
+      SeedFu.seed(seed_paths)
```
tgxworld, line 162: Actually can we fold this into the `Discourse.skip_post_deploy_migrations` check below?

PR #10119, opened 2020-06-24
Notifications filter

Ahmedgagan:
notifications-filter bug solved 
issue: https://meta.discourse.org/t/notifications-unread-only-filter/37621/32


app/assets/javascripts/discourse/app/routes/user-notifications.js
```diff
@@ -13,14 +16,17 @@ export default DiscourseRoute.extend(ViewingActionType, {
     }
   },
 
-  model() {
+  model(params) {
     const username = this.modelFor("user").get("username");
-
+    this._params = params;
```
eviltrout, line 21: Why is this being set? I don't see it referenced elsewhere.

app/assets/javascripts/discourse/app/widgets/user-notifications-large.js
```diff
@@ -30,9 +30,8 @@ createWidget("large-notification-item", {
 
 export default createWidget("user-notifications-large", {
   html(attrs) {
-    const notifications = attrs.notifications;
```
eviltrout, line 33: I don't think the changes in this file are worth comitting.

Ahmedgagan:
Hey, @eviltrout I did the changes as specified. just tests are left

app/assets/javascripts/discourse/app/routes/user-notifications.js
```diff
@@ -2,6 +2,9 @@ import DiscourseRoute from "discourse/routes/discourse";
 import ViewingActionType from "discourse/mixins/viewing-action-type";
 
 export default DiscourseRoute.extend(ViewingActionType, {
+  controllerName: "user-notifications",
```
davidtaylorhq, line 5: Is this needed? I think it should automatically use this controller because it has the same name 🤔

app/assets/javascripts/discourse/app/templates/user/notifications-index.hbs
```diff
@@ -8,6 +8,9 @@
   </div>
 {{/if}}
 
+{{notifications-filter value=filter onChange=(action (mut filter))}}
+<span class="user-notifications-filter-separator"></span>
```
davidtaylorhq, line 12: It would be better if we can position this using CSS without adding a separator element. How about wrapping the notifications-filter in a `<div>`, and then setting it to 100% width with a border-bottom.

app/assets/javascripts/discourse/app/widgets/user-notifications-large.js
davidtaylorhq: @Ahmedgagan just bumping this comment ^^

app/controllers/notifications_controller.rb
```diff
@@ -37,18 +37,28 @@ def index
...
+        notifications = Notification.where(user_id: user.id, read: false)
+          .visible
+          .includes(:topic)
+          .order(created_at: :desc)
+      else
+        notifications = Notification.where(user_id: user.id)
+          .visible
+          .includes(:topic)
+          .order(created_at: :desc)
+      end
```
davidtaylorhq, line 55: I think we can reduce some repetition here by doing something like

```suggestion
      notifications = Notification.visible
                     .where(user_id: user.id)
                     .includes(:topic)
                     .order(created_at: :desc)
     
     if params[:filter] == "read"
       notifications = notifications.where(read: true)
     elsif params[:filter] == "unread"
       notifications = notifications.where(read: false)
     end
```

app/assets/javascripts/discourse/app/templates/user/notifications-index.hbs
Ahmedgagan: Sure, ill try this.

app/assets/javascripts/discourse/app/widgets/user-notifications-large.js
Ahmedgagan: Yes, i already changed this in my last commit.

app/assets/javascripts/discourse/app/routes/user-notifications.js
Ahmedgagan: If i remove this it gives this error
```
Assertion Failed: You're not allowed to have more than one controller property map to the same query param key, but both `userNotifications:filter` and `user-notifications:filter` map to `filter`. You can fix this by mapping one of the controller properties to a different query param key via the `as` config option, e.g. `filter: { as: 'other-filter' }
```

app/assets/javascripts/discourse/app/templates/user/notifications-index.hbs
Ahmedgagan: Sure, ill try this out

PR #10121, opened 2020-06-24
Update category_featured_topic.rb

LukeClancy:
Only made one change, catching ActiveRecord::RecordNotUnique as well as the pre-existing PG::UniqueViolation

In my experience, does not catch the PG::UniqueViolation as ActiveRecord::RecordNotUnique is on top of it, so you should have both. I ran into this problem when creating large amounts of topics at the same time (I usually find this downstream in the source code when I pull it and change it there after each git clone)


eviltrout:
Thanks, this seems good.

PR #10125, opened 2020-06-25
FIX: Do not highlight large code blocks

nbianca:


SamSaffron:
This is good, though we should probably add a test

nbianca:
I did not succeed at first to do that because highlight.js is loaded at run-time. I gave it another try and found out a test which was doing something similar, so I used that as a base.

test/javascripts/components/highlighted-code-test.js
```diff
@@ -20,3 +20,22 @@ componentTest("highlighting code", {
...
   }
 });
+
+componentTest("large code blocks are not highlighted", {
+  template: "{{highlighted-code lang='ruby' code=code}}",
+
+  beforeEach() {
+    Discourse.HighlightJSPath =
+      "assets/highlightjs/highlight-test-bundle.min.js";
+    this.set("code", "puts a\n".repeat(15000));
```
ZogStriP, line 30: The `"puts a\n".repeat(15000)` might be time-consuming / CPU-intensive. I think we should only do it once and put it in a `const`?

PR #10126, opened 2020-06-25
FIX: published-page-header should be a sibling to published-page-body

tshenry:
According to the [Page Publishing Meta topic](https://meta.discourse.org/t/page-publishing/151971), the spec for the `published-page-header` element is "title and authorship information." This change will move `published-page-body` to be a sibling of `published-page-header` since it does not seem correct to include `published-page-body` as a child.

**Before:**
Screen Shot 2020-06-25 at 11 05 04 AM

**After:**
Screen Shot 2020-06-25 at 11 06 21 AM


coding-horror:
This should probably be backported as well?

tshenry:
> This should probably be backported as well?

Done!

PR #10131, opened 2020-06-26
FIX: count new and unread respects muted categories

lis2:
countCategoryByState function should take categories muted by the user into consideration

### DIFF
--- a/app/assets/javascripts/discourse/app/models/topic-tracking-state.js
+++ b/app/assets/javascripts/discourse/app/models/topic-tracking-state.js
@@ -410,6 +410,7 @@ const TopicTrackingState = EmberObject.extend({
 
   countCategoryByState(fn, categoryId, tagId) {
     const subcategoryIds = this.getSubCategoryIds(categoryId);
+    const mutedCategoryIds = this.currentUser.muted_category_ids;
     return _.chain(this.states)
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/discourse/app/models/topic-tracking-state.js
```diff
@@ -410,14 +410,17 @@ const TopicTrackingState = EmberObject.extend({
...
       .filter(fn)
       .filter(
         topic =>
           topic.archetype !== "private_message" &&
           !topic.deleted &&
           (!categoryId || subcategoryIds.has(topic.category_id)) &&
-          (!tagId || (topic.tags && topic.tags.indexOf(tagId) > -1))
+          (!tagId || (topic.tags && topic.tags.indexOf(tagId) > -1)) &&
+          (!mutedCategoryIds ||
+            mutedCategoryIds.indexOf(topic.category_id) === -1)
```
SamSaffron, line 423: I do not think this is correct, this is only meant to apply to new topics, not to unread. You need to amend it so this change only impacts "countNew". Even if a category is muted you may have specific topics in said category you are tracking or watching.

SamSaffron:
looks good to me now! merge away

PR #10133, opened 2020-06-26
FIX: Improve category hashtag lookup

udan11:
This commit improves support for sub-sub-categories and does not include
the ID of the category in the slug, which fixes the composer preview.

### DIFF
--- a/app/models/concerns/category_hashtag.rb
+++ b/app/models/concerns/category_hashtag.rb
@@ -7,19 +7,8 @@ module CategoryHashtag
 
   class_methods do
     def query_from_hashtag_slug(category_slug)
-      parent_slug, child_slug = category_slug.split(SEPARATOR, 2)
-
-      categories = Category.where(slug: parent_slug)
-
-      if child_slug
-        Category.where(slug: child_slug, parent_category_id: categories.select(:id)).first
-      else
-        categories.where(parent_category_id: nil).first
-      end
+      slug_path = category_slug.split(SEPARATOR)
+      Category.find_by_slug_path(slug_path)
     end
-  end
-
-  def hashtag_slug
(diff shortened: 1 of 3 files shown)


app/models/concerns/category_hashtag.rb
```diff
@@ -7,19 +7,24 @@ module CategoryHashtag
...
-  end
 
-  def hashtag_slug
-    full_slug.split("-").last(2).join(SEPARATOR)
+      last_category_id = nil
+
+      slug_path.each do |slug|
+        query = Category.select(:id).where(slug: slug)
+        query = query.where(parent_category_id: last_category_id) if last_category_id.present?
+        return nil if query.blank?
```
ZogStriP, line 23: When would `query` be `blank?`

app/models/concerns/category_hashtag.rb
```diff
@@ -7,19 +7,24 @@ module CategoryHashtag
...
+        slug_path.map! { |slug| CGI.escape(slug) }
       end
-    end
-  end
 
-  def hashtag_slug
-    full_slug.split("-").last(2).join(SEPARATOR)
+      last_category_id = nil
+
+      slug_path.each do |slug|
```
ZogStriP, line 20: Can we do one SQL query instead of `slug_path.size` queries?

app/models/concerns/category_hashtag.rb
udan11: Sure, that was needed for three level mentions.

app/models/concerns/category_hashtag.rb
udan11: Sure, that was needed for three level mentions.

SamSaffron:
Very welcome fix! thanks @udan11 !

PR #10135, opened 2020-06-26
FIX: Correct version comparison logic when comparing stable to beta

davidtaylorhq:
For example, version 1.3.0 should be considered higher than 1.3.0.beta3. So `Discourse.has_needed_version?('1.3.0', '1.3.0.beta3')` should return true

### DIFF
--- a/lib/version.rb
+++ b/lib/version.rb
@@ -21,12 +21,14 @@ module Discourse
 
     (0..[current_split.size, needed_split.size].max).each do |idx|
       current_str = current_split[idx] || ''
+      needed_str = needed_split[idx] || ''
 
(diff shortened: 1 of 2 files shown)


CvX:
Can we replace all this custom logic with just `Gem::Version`?

```ruby
Gem::Version.new('1.3.0') > Gem::Version.new('1.3.0.beta3')
=> true
```

davidtaylorhq:
😂 well yes, I guess we can! I assumed that we had a custom implementation because our `.beta#` syntax was a little weird.But it looks like Gem::Version works just fine.

Updated the PR

SamSaffron:
do we need to backport this?

